PR #882, opened 2019-07-26
Publish CoordinatedReactiveControl extension

mathbagu:

N/A

Feature

N/A

There is an extension to control coordinately the reactive power. This extension could be used in load-flows to find the reactive power setpoints of generators with the same regulating terminal. 

**Other information**:

### DIFF
--- a/iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/CoordinatedReactiveControlXmlSerializer.java
+++ b/iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/CoordinatedReactiveControlXmlSerializer.java
@@ -1,0 +1,70 @@
+/**
+ * Copyright (c) 2019, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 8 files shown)


iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/CoordinatedReactiveControlXmlSerializer.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override
+    public InputStream getXsdAsStream() {
+        return getClass().getResourceAsStream("/xsd/coordinatedReactiveControl.xsd");
+    }
+
+    @Override
+    public String getNamespaceUri() {
+        return "http://www.itesla_project.eu/schema/iidm/ext/coordinated_reactive_control/1_0";
```
geofjamg: A general remark for future extension, we should stop using http://www.itesla_project.eu/ namespace. We have to keep it for old namespace including one providing from internal repo for backward compatibility but new extensions should use http://www.powsybl.org/...

mathbagu: I will change the namespace. Do you think it’s better to use http or https?

PR #892, opened 2019-08-16
Config

jonenst:
**Other information**:
Needs discussion: maven cli tests now behave differently from IDE tests...

NO

feature

The PlatformConfig from defaultConfig() has the hardcoded behavior of looking into powsybl.config.dirs (or ~/.itools if not defined) and looking into the environnement

Allow to choose a PlatformConfigProvider using system variables (-Dpowsybl.config.provider). It should implement a getConfig method that is called (once, it is cached) when using defaultConfig(). By default it keeps the old behavior.

NO

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@
 /commons/target/
 /computation-local/target/
 /computation/target/
+/config-classic/target/
 /config-test/target/
 /contingency/target/
 /contingency/contingency-api/target/
--- a/action/action-simulator/pom.xml
+++ b/action/action-simulator/pom.xml
@@ -121,6 +121,12 @@
         </dependency>
         <dependency>
             <groupId>${project.groupId}</groupId>
+            <artifactId>powsybl-config-test</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
             <artifactId>powsybl-tools</artifactId>
             <version>${project.version}</version>
             <scope>test</scope>
--- a/config-classic/pom.xml
+++ b/config-classic/pom.xml
@@ -1,0 +1,70 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (c) 2019, RTE (http://www.rte-france.com)
+    This Source Code Form is subject to the terms of the Mozilla Public
+    License, v. 2.0. If a copy of the MPL was not distributed with this
+    file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>com.powsybl</groupId>
+        <artifactId>powsybl-core</artifactId>
+        <version>3.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>powsybl-config-classic</artifactId>
+    <name>Config Classic</name>
+    <description>Classic Config Provider</description>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifestEntries>
+                            <Automatic-Module-Name>com.powsybl.config.classic</Automatic-Module-Name>
+                        </manifestEntries>
+                    </archive>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
+    <dependencies>
+        <!-- Compilation dependencies -->
+        <dependency>
+            <groupId>com.google.auto.service</groupId>
+            <artifactId>auto-service</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>powsybl-commons</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
+
+        <!-- Test dependencies -->
+        <dependency>
+            <groupId>com.google.jimfs</groupId>
+            <artifactId>jimfs</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+</project>
+
--- a/pom.xml
+++ b/pom.xml
@@ -222,6 +222,15 @@
                     <generateBackupPoms>false</generateBackupPoms>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                  <systemPropertyVariables>
+                    <powsybl.config.provider>test</powsybl.config.provider>
+                  </systemPropertyVariables>
+                </configuration>
+            </plugin>
         </plugins>
         <pluginManagement>
             <plugins>
--- a/pom.xml
+++ b/pom.xml
@@ -58,6 +58,7 @@
         <module>commons</module>
         <module>computation</module>
         <module>computation-local</module>
+        <module>config-classic</module>
         <module>config-test</module>
         <module>contingency</module>
         <module>distribution-core</module>
@@ -220,15 +221,6 @@
                 <version>${maven.versions.version}</version>
                 <configuration>
                     <generateBackupPoms>false</generateBackupPoms>
-                </configuration>
-            </plugin>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-surefire-plugin</artifactId>
-                <configuration>
-                  <systemPropertyVariables>
-                    <powsybl.config.provider>test</powsybl.config.provider>
-                  </systemPropertyVariables>
                 </configuration>
             </plugin>
         </plugins>
--- a/sensitivity-api/pom.xml
+++ b/sensitivity-api/pom.xml
@@ -76,6 +76,12 @@
         </dependency>
         <dependency>
             <groupId>${project.groupId}</groupId>
+            <artifactId>powsybl-config-test</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
             <artifactId>powsybl-tools</artifactId>
             <version>${project.version}</version>
             <type>test-jar</type>
--- a/afs/afs-ws/afs-ws-server/pom.xml
+++ b/afs/afs-ws/afs-ws-server/pom.xml
@@ -105,6 +105,12 @@
             <type>test-jar</type>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>powsybl-config-test</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/afs/afs-ws/afs-ws-server/src/test/java/com/powsybl/afs/ws/server/AppStorageServerTest.java
+++ b/afs/afs-ws/afs-ws-server/src/test/java/com/powsybl/afs/ws/server/AppStorageServerTest.java
@@ -55,6 +55,7 @@ public class AppStorageServerTest extends AbstractAppStorageTest {
                                .loadPomFromFile("pom.xml")
                                .importRuntimeDependencies()
                                .resolve("org.mockito:mockito-all",
+                                        "com.powsybl:powsybl-config-test",
                                         "com.powsybl:powsybl-afs-mapdb")
                                .withTransitivity()
                                .asFile();
--- a/commons/src/main/java/com/powsybl/commons/config/PlatformConfig.java
+++ b/commons/src/main/java/com/powsybl/commons/config/PlatformConfig.java
@@ -18,7 +18,6 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.*;
 import java.util.ServiceLoader;
-import java.util.stream.Collectors;
 
 /**
  *
@@ -27,8 +26,6 @@ import java.util.stream.Collectors;
 public class PlatformConfig {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(PlatformConfig.class);
-
-    private static final String KEY = "powsybl.config.provider";
 
     private static PlatformConfig defaultConfig;
 
@@ -67,30 +64,18 @@ public class PlatformConfig {
 
     public static synchronized PlatformConfig defaultConfig() {
         if (defaultConfig == null) {
-            String configName = System.getProperty(KEY);
-            if (configName != null) {
-                List<PlatformConfigProvider> platformConfigProviders = Lists
-                        .newArrayList(ServiceLoader.load(PlatformConfigProvider.class));
-                Optional<PlatformConfigProvider> foundProvider = platformConfigProviders.stream()
-                        .filter(platformConfigProvider -> configName.equals(platformConfigProvider.getName()))
-                        .findFirst();
-                if (!foundProvider.isPresent()) {
-                    if (LOGGER.isErrorEnabled()) {
-                        List<String> available = platformConfigProviders.stream().map(PlatformConfigProvider::getName)
-                                .collect(Collectors.toList());
-                        LOGGER.error("Requested platform configuration provider {} = {} not found; available: {}", KEY,
-                                configName, available);
-                    }
-                } else {
-                    LOGGER.info("Using platform configuration provider {} = {}", KEY, configName);
-                    defaultConfig = foundProvider.get().getPlatformConfig();
-                }
+            List<PlatformConfigProvider> providers = Lists.newArrayList(ServiceLoader.load(PlatformConfigProvider.class));
+            if (providers.isEmpty()) {
+                LOGGER.error("Platform configuration provider not found. For tests, consider using TestPlatformConfigProvider in powsybl-config-test. Otherwise, consider using ClassicPlatformConfigProvider from powsybl-config-classic.");
+                throw new PowsyblException("Platform configuration provider not found");
             }
-            if (defaultConfig == null) {
-                LOGGER.info("Using default platform configuration provider {}",
-                        ClassicPlatformConfigProvider.class.getSimpleName());
-                defaultConfig = new ClassicPlatformConfigProvider().getPlatformConfig();
+            if (providers.size() > 1) {
+                LOGGER.error("Multiple platform configuration providers found: {}", providers);
+                throw new PowsyblException("Multiple platform configuration providers found");
             }
+            PlatformConfigProvider p = providers.get(0);
+            LOGGER.info("Using platform configuration provider {}", p.getName());
+            defaultConfig = p.getPlatformConfig();
         }
         return defaultConfig;
     }
@@ -123,4 +108,5 @@ public class PlatformConfig {
     public Optional<ModuleConfig> getOptionalModuleConfig(String name) {
         return getRepository().getModuleConfig(name);
     }
+
 }
--- a/commons/src/main/java/com/powsybl/commons/config/PlatformConfig.java
+++ b/commons/src/main/java/com/powsybl/commons/config/PlatformConfig.java
@@ -8,6 +8,7 @@ package com.powsybl.commons.config;
 
 import com.google.common.base.Supplier;
 import com.google.common.base.Suppliers;
+import com.google.common.collect.Lists;
 import com.powsybl.commons.PowsyblException;
 import com.powsybl.commons.io.FileUtil;
 import org.slf4j.Logger;
@@ -16,6 +17,8 @@ import org.slf4j.LoggerFactory;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.*;
+import java.util.ServiceLoader;
+import java.util.stream.Collectors;
 
 /**
  *
@@ -24,6 +27,8 @@ import java.util.*;
 public class PlatformConfig {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(PlatformConfig.class);
+
+    private static final String KEY = "powsybl.config.provider";
 
     private static PlatformConfig defaultConfig;
 
@@ -62,7 +67,30 @@ public class PlatformConfig {
 
     public static synchronized PlatformConfig defaultConfig() {
         if (defaultConfig == null) {
-            defaultConfig = new ClassicPlatformConfigProvider().getPlatformConfig();
+            String configName = System.getProperty(KEY);
+            if (configName != null) {
+                List<PlatformConfigProvider> platformConfigProviders = Lists
+                        .newArrayList(ServiceLoader.load(PlatformConfigProvider.class));
+                Optional<PlatformConfigProvider> foundProvider = platformConfigProviders.stream()
+                        .filter(platformConfigProvider -> configName.equals(platformConfigProvider.getName()))
+                        .findFirst();
+                if (!foundProvider.isPresent()) {
+                    if (LOGGER.isErrorEnabled()) {
+                        List<String> available = platformConfigProviders.stream().map(PlatformConfigProvider::getName)
+                                .collect(Collectors.toList());
+                        LOGGER.error("Requested platform configuration provider {} = {} not found; available: {}", KEY,
+                                configName, available);
+                    }
+                } else {
+                    LOGGER.info("Using platform configuration provider {} = {}", KEY, configName);
+                    defaultConfig = foundProvider.get().getPlatformConfig();
+                }
+            }
+            if (defaultConfig == null) {
+                LOGGER.info("Using default platform configuration provider {}",
+                        ClassicPlatformConfigProvider.class.getSimpleName());
+                defaultConfig = new ClassicPlatformConfigProvider().getPlatformConfig();
+            }
         }
         return defaultConfig;
     }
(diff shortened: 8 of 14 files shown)


jonenst:
> For me, using a system variable for choosing the PlatformConfigProvider is a bad solution. You should rely only on service loader and classpath configuration. PlatformConfigProvider implementation should be discovered at runtime using service loader and without any system variable:
    if none is found use default platform config provider (normal behaviour)
    if one is found, use it (for instance to run unit tests)
    if more than one is found, throws an exception

We had a discussion about these ideas and concluded against using them. We can still discuss. The argument against it were:
- The classpath is easy to get wrong. A jvm system property is a lot easier to get right. For example, lookt at the very well known error using slf4j: "Multiple bindings were found on the class path."

However, if we choose to implement your solution, maybe we should use slf4j behavior's of not throwing an exception for multiple implementations ? "Even when multiple bindings are present, SLF4J will pick one logging framework/implementation and bind with it. The way SLF4J picks a binding is determined by the JVM and for all practical purposes should be considered random. As of version 1.6.6, SLF4J will name the framework/implementation class it is actually bound to."

Or at least explicitly find a good reason not to do like slf4j.


- The classpath is harder to change than a jvm system property.

I'm not sure about this one anymore. For the classpath, because of transitive dependencies, you either have to fix projects far away from the failing tests (hard or impossible), or use exclusions (ugly). However, the system property is maybe even worse because IDEs do not pick the maven system property configuration.. This leads many projects to warn their developpers "the tests run in maven CLI are not the same as the tests run in your IDE". If we can avoid that, we should.

- The classpath is harder to debug than a jvm property. Both in IDEs and in Maven, it's hard to inspect the whole class path and dependency graph..

What do you think ?

> Also, instead of an EmptyPlatformConfig we could have a ClassLoaderPlatformConfig, that would allow developpers to add an src/test/resources/config.json to setup platform config using a json file for unit tests. That would be really convenient because in actual code we need to define platform config data programmatically in unit tests.

This means that all the tests in one maven project have the same config. I'm not sure if it's enough to solve the problem of having to define data programmatically in the tests.

If it's enough, an alternative would be to configure the target/ folder as a powsybl.config.dirs of the StandardPlatformConfigLoader but I agree that the classpath approach is simpler and superior.

sylvlecl:
I think enforcing to have only one config provider in the classpath will not be very convenient :
 * you will not be able to define a provider based on another implementation (since it will need a dependency to the other provider, and then you will have 2 providers in classpath)
 * You will not be able either to use 2 differents implementations in different unit tests of one module.
 * you will **have** to include the "standard" dependency (or your custom one) in any packaging modules (itools or application), and **only** in them. It will not be obvious for the application builders.

I think we need to have some priority system, to at least authorize having the default implementation and another one in the classpath. In that case the other one will be used.

But it's probably not enough, because of the first reason mentioned above.

geofjamg:
> 
> 
> I think enforcing to have only one config provider in the classpath will not be very convenient :
> 
>     * you will not be able to define a provider based on another implementation (since it will need a dependency to the other provider, and then you will have 2 providers in classpath)
> 
>     * You will not be able either to use 2 differents implementations in different unit tests of one module.
> 
>     * you will **have** to include the "standard" dependency (or your custom one) in any packaging modules (itools or application), and **only** in them. It will not be obvious for the application builders.
> 
> 
> I think we need to have some priority system, to at least authorize having the default implementation and another one in the classpath. In that case the other one will be used.
> 
> But it's probably not enough, because of the first reason mentioned above.

But what could be the use case having 2 differents providers in your unit test? 
I do not see any issue including the dependency in any modules. You do it with mockito, slf4j-simple, jimfs etc.

geofjamg:
> 
> 
> > For me, using a system variable for choosing the PlatformConfigProvider is a bad solution. You should rely only on service loader and classpath configuration. PlatformConfigProvider implementation should be discovered at runtime using service loader and without any system variable:
> > if none is found use default platform config provider (normal behaviour)
> > if one is found, use it (for instance to run unit tests)
> > if more than one is found, throws an exception
> 
> We had a discussion about these ideas and concluded against using them. We can still discuss. The argument against it were:
> 
>     * The classpath is easy to get wrong. A jvm system property is a lot easier to get right. For example, lookt at the very well known error using slf4j: "Multiple bindings were found on the class path."
> 
> 
> However, if we choose to implement your solution, maybe we should use slf4j behavior's of not throwing an exception for multiple implementations ? "Even when multiple bindings are present, SLF4J will pick one logging framework/implementation and bind with it. The way SLF4J picks a binding is determined by the JVM and for all practical purposes should be considered random. As of version 1.6.6, SLF4J will name the framework/implementation class it is actually bound to."
> 
> Or at least explicitly find a good reason not to do like slf4j.
> 
>     * The classpath is harder to change than a jvm system property.
> 
> 
> I'm not sure about this one anymore. For the classpath, because of transitive dependencies, you either have to fix projects far away from the failing tests (hard or impossible), or use exclusions (ugly). However, the system property is maybe even worse because IDEs do not pick the maven system property configuration.. This leads many projects to warn their developpers "the tests run in maven CLI are not the same as the tests run in your IDE". If we can avoid that, we should.
> 
>     * The classpath is harder to debug than a jvm property. Both in IDEs and in Maven, it's hard to inspect the whole class path and dependency graph..
> 
> 
> What do you think ?
> 
> > Also, instead of an EmptyPlatformConfig we could have a ClassLoaderPlatformConfig, that would allow developpers to add an src/test/resources/config.json to setup platform config using a json file for unit tests. That would be really convenient because in actual code we need to define platform config data programmatically in unit tests.
> 
> This means that all the tests in one maven project have the same config. I'm not sure if it's enough to solve the problem of having to define data programmatically in the tests.
> 
> If it's enough, an alternative would be to configure the target/ folder as a powsybl.config.dirs of the StandardPlatformConfigLoader but I agree that the classpath approach is simpler and superior.

Do you agree that having a different behaviour in Maven, and IDE is a showstopper? It would be so a mess.
I totally disagree in your first statement about classpath eaysier to get wrong than system variable. Where does it come from?
Slf4j is a very sucessfull libray and its design is well knwon to be very good. What is the problem with the warning about multiple impl? If you have this warn, you just have to fix it and change your classpath to have only one impl. Pretty simple, clear and well design.
A general comment, are you able to find a well knwon libray that rely on system var to configure unit testing? Personnaly I cannot. More generally widely used design are often best...

sylvlecl:
I have an additional issue to bring to the discussion :)
But first, I just want to make clear that I am not defeinding the system property approach.

The `ServiceLoader` approach does not integrate very well with dependency injection frameworks. It's hard to inject some "external behaviour" inside services instantiated this way. Typically, if you use powsybl and spring-boot to build an application, you will not be able to inject application components (say datasource) inside the config provider. The example use-case, here, is to have a database-backed platform config.

Do we want to support such a feature ?

geofjamg:
> 
> 
> I have an additional issue to bring to the discussion :)
> But first, I just want to make clear that I am not defending the system property approach.
> 
> The `ServiceLoader` approach does not integrate very well with dependency injection frameworks. It's hard to inject some "external behaviour" inside services instantiated this way. Typically, if you use powsybl and spring-boot to build an application, you will not be able to inject application components (say datasource) inside the config provider. The example use-case, here, is to have a database-backed platform config.
> 
> Do we want to support such a feature ?

For sure database platform config is a valid use case, but why do you need multiple PlatformConfigProvider for that?
I agree with your concern about dependency injection and service loader but it is a more general topic that only platform config and I really doubt this PR could solve it. I also doubt we could solve it at short/middle term. I would require a deep redesign of the code.

What I really like in this PR is that with a simple developement it could solve several current issues.
  - platform config for unit testing. Today it is impossible to test code like MyConfig.load(). With this PR modified like described in my first comment (add a test dependency in each module) it would solve it in a clean way.
  - Your use case about database instead file could be done easily to.

Platform configuration is a complex topic and I think that we must be pragmatic and not try to solve all issues / use cases in this PR, otherwise we will just never make any progress on it.

jonenst:
> Do you agree that having a different behaviour in Maven, and IDE is a showstopper? It would be so a mess.

I agree that we should try really hard to avoid this, that's why I have mentioned the problem.

> I totally disagree in your first statement about classpath eaysier to get wrong than system variable. Where does it come from?

the classpath is the result of a pretty complex computation (maven dependency mediation) on a pretty big graph (hundreds of deps). The system variable is a manually written string. From my experience, it happens more frequently that people get something bad in their classpath (causing things like slf4j's "Multiple bindings were found on the class path." than that they write the wrong value in their system property. Forcing a system variable is often possible in all kinds of setups (maven cli, IDEs, ...). Forcing the classpath is harder. However, what will be extremely common though is that they will forget to set the system property (like when running the test in the IDE) and we had not taken this into account when we said that it's easier to get wrong, so because of this case I take back what I said: it's easy to forget to set a system variable.

> Slf4j is a very sucessfull libray and its design is well knwon to be very good. What is the problem with the warning about multiple impl? If you have this warn, you just have to fix it and change your classpath to have only one impl. Pretty simple, clear and well design.

So should we do that when there are multiple implementations, we just log a warning and use a random one like SLF4J instead of throwing like you proposed initially ?

> A general comment, are you able to find a well knwon libray that rely on system var to configure unit testing? Personnaly I cannot.

Doesn't seem crazy to me... setting the xmx for tests, configuring some loggers for tests.. For example: https://github.com/spring-projects/spring-framework/blob/master/import-into-eclipse.md "If attempting to run all JUnit tests from within the IDE, you will likely need to set the following VM options to avoid out of memory errors: -XX:MaxPermSize=2048m -Xmx2048m -XX:MaxHeapSize=2048m"

Log4j2 uses this mechanism: https://logging.apache.org/log4j/2.x/manual/extending.html  -DLoggerContextFactory=... and -Dlog4j.configurationFactory

> More generally widely used design are often best...

Yes I try as much as I can to do this too.


Using a system property to choose one of the implementations in the class path instead of expecting the classpath to contain only one implementation has the additional benefit to allow to rerun a program with a different PlatformConfigProvider. It could be useful if users report bugs and you want to disable their config, you can tell them to start the program with -Dpowsybl.config.provider=empty ?


In the end, I think there are 2 different aspects to this problem:
- do we allow only one implementation in the classpath ? I think this is too rigid. One good example is that this prevents inheritance between PlatformConfigProvider (@sylvlecl  's argument). We don't have this restrictions for all the other classes that we load through serviceloader.
- If we allow more than one, how do we choose the one we want ? For all other things we load through service loader, we use PlatformConfig (e.g. componentDefaultConfig:  LoadFlowFactory: com.rte_france.powsybl.hades2.Hades2Factory). but like "who watches the watcher?" we have the problem of "who configures the config". A system variable is easy to force once you know about it but also easy to forget. A classpath element is not forgotten but if it has the wrong value it's less flexible to change.


What about looking for a system property first, and only if missing look for a hardcoded file in the classpath containing the same thing as the system property ? (Not sure, I can't tell if this is crazy or not)

geofjamg:
> 
> 
> > Do you agree that having a different behaviour in Maven, and IDE is a showstopper? It would be so a mess.
> 
> I agree that we should try really hard to avoid this, that's why I have mentioned the problem.
>
>>> We should not "try hard to", we should do it. Most of the developpers are IDE users. I personnally click "run test" on my IDE hundred times a day but on the contrary I very rarely run tests through maven (just one time before commit). In IntelliJ you can run unit test at different level: method, class but also package and each time you run at a new level, a new test runner is created and you have to add variables again. And everybody will most of the time forget to add theses variables.

> > I totally disagree in your first statement about classpath eaysier to get wrong than system variable. Where does it come from?
> 
> the classpath is the result of a pretty complex computation (maven dependency mediation) on a pretty big graph (hundreds of deps). The system variable is a manually written string. From my experience, it happens more frequently that people get something bad in their classpath (causing things like slf4j's "Multiple bindings were found on the class path." than that they write the wrong value in their system property. Forcing a system variable is often possible in all kinds of setups (maven cli, IDEs, ...). Forcing the classpath is harder. However, what will be extremely common though is that they will forget to set the system property (like when running the test in the IDE) and we had not taken this into account when we said that it's easier to get wrong, so because of this case I take back what I said: it's easy to forget to set a system variable.
>
>>> Classpath could be complex, to manage yes but most of the time it is not. In the case of Slf4j it is not tricky at all because impl are always added in final modules. I never had any trouble to use it or debug bad configuration.
> 
> > Slf4j is a very sucessfull libray and its design is well knwon to be very good. What is the problem with the warning about multiple impl? If you have this warn, you just have to fix it and change your classpath to have only one impl. Pretty simple, clear and well design.
> 
> So should we do that when there are multiple implementations, we just log a warning and use a random one like SLF4J instead of throwing like you proposed initially ?
> 
> > A general comment, are you able to find a well knwon libray that rely on system var to configure unit testing? Personnaly I cannot.
> 
> Doesn't seem crazy to me... setting the xmx for tests, configuring some loggers for tests.. For example: https://github.com/spring-projects/spring-framework/blob/master/import-into-eclipse.md "If attempting to run all JUnit tests from within the IDE, you will likely need to set the following VM options to avoid out of memory errors: -XX:MaxPermSize=2048m -Xmx2048m -XX:MaxHeapSize=2048m"
>
>>> JVM parameters is a totally different thing, it is not applicative.

> 
> Log4j2 uses this mechanism: https://logging.apache.org/log4j/2.x/manual/extending.html -DLoggerContextFactory=... and -Dlog4j.configurationFactory
>
> > More generally widely used design are often best...
> 
> Yes I try as much as I can to do this too.
> 
> Using a system property to choose one of the implementations in the class path instead of expecting the classpath to contain only one implementation has the additional benefit to allow to rerun a program with a different PlatformConfigProvider. It could be useful if users report bugs and you want to disable their config, you can tell them to start the program with -Dpowsybl.config.provider=empty ?
> 
> In the end, I think there are 2 different aspects to this problem:
> 
>     * do we allow only one implementation in the classpath ? I think this is too rigid. One good example is that this prevents inheritance between PlatformConfigProvider (@sylvlecl  's argument). We don't have this restrictions for all the other classes that we load through serviceloader.
> 
>>> Inheritance is a bad example, because in that case you can just create an abstract class not annotated for service loader and inherit from it so that you won't have 2 PlatformConfigProvider in your classpath. I really don't understand why it is too rigid. We don't have to compare it with other service loader usage, it depends of the design you want to have. I you want a singleton (and a singleton is fine for many use cases), there is no problem to enforce a single impl in the classpath.  
>
>     * If we allow more than one, how do we choose the one we want ? For all other things we load through service loader, we use PlatformConfig (e.g. componentDefaultConfig:  LoadFlowFactory: com.rte_france.powsybl.hades2.Hades2Factory). but like "who watches the watcher?" we have the problem of "who configures the config". A system variable is easy to force once you know about it but also easy to forget. A classpath element is not forgotten but if it has the wrong value it's less flexible to change.
> 
> 
> What about looking for a system property first, and only if missing look for a hardcoded file in the classpath containing the same thing as the system property ? (Not sure, I can't tell if this is crazy or not)
>
>>> Not clear to me how this proposition could solve Maven/IDE issue which is the major issue of the actual design. There is no perfect solution. Even if I have a clear preference on classpath based design because I think there are more modern and widely used, I totally respect other other way to see and solve this kind of problem but one thing is sure for me: we must not have to add manual config to IDE to run a unit test and have the same result that with Maven. It is not acceptable and I personnaly never approved a change that will break this rule. I would just lead me to waste time in my day to day work (and make errors).

jonenst:
> > What about looking for a system property first, and only if missing look for a hardcoded file in the classpath containing the same thing as the system property ?

> Not clear to me how this proposition could solve Maven/IDE issue

This proposition was that we don't set the system property at all for the tests (the system property is only used by users running programs, if they want. Or the system property could be removed entirely if we don't need the feature). Without a system property, all the info comes from the classpath so tests run the same in maven CLI and the IDE. But instead of coming from the META-INF/services/PlatformConfigProvider file, it comes from another one (we could use META-INF/powsybl-platformconfigprovider or something) 

This is basically saying that a module running tests can't easily control what META-INF/services/PlatformConfigProvider files they have in the classpath, but they can more easily control this other file.

I've never seen this idea implemented so it fails the "More generally widely used design are often best..." rule though :)

 -
Additionally, This is a good example of how the classpath can be hard to control:
https://www.slf4j.org/faq.html#excludingJCL

Another example: (from https://www.slf4j.org/codes.html ):
> cassandra-all version 0.8.1 declares both log4j and slf4j-log4j12 as compile-time dependencies. Thus, when you include cassandra-all as a dependency in your project, the cassandra-all declaration will cause both slf4j-log4j12.jar and log4j.jar to be pulled in as dependencies. In case you do not wish to use log4j as the the SLF4J backend, you can instruct Maven to exclude these two artifacts

-
Side note: It's really sad that IDE correctly import the classpath from maven but incorrectly import system property.. What a nuisance..

sylvlecl:
What about keeping the service-based approach, but with an ordering system as proposed by @mathbagu on monday ?

I think it would solve most problems : the problem of enforcing the use of one implementation for tests, and the problem of not being able to inherit from one implementation for another one.

It's simple to implement with a `getOrder` or `getPriority` method which could have a default implementation. Value will be low for default implementation, higher for the default test implementation.

It's similar to the solution adopted in spring to order collections of beans with `@Order` annotation.

However, it would not solve the issue I mentioned above with external application components interoperability. I guess for this we will still need some static variable mechanism.

sylvlecl:
PS:
We may have common static methods to look up those kind of ordered or named services.
For that purpose, those services could implement general purpose interfaces `OrderedService` or`NamedService`.

jonenst:
> What about keeping the service-based approach, but with an ordering system as proposed by @mathbagu on monday ?

What I don't like about this approach is that it can get abused and people start using order=1000 and then order=10000 etc..

sylvlecl:
It sounds to me like a very theoretical downside to be compared with a very tangible added value :)

What exactly is wrong in your example ? What will be the actual cost of having arbitrary values such as 10000 ?

Note: it would also be possible to provide abstract implementations to let the user decide of the actual order.

jonenst:
Let's go forward with the classpath solution ?

> So, in all powsybl modules that needs to run unit tests accessing platform config (almost all...), we only have to add a dependency on a Maven artifact that contains the in EmptyPlatformConfigProvider.

Is this OK? or can we avoid to duplicate this dependency everywhere ?

If we put the dependency in the top pom (with scope=test), it's included everywhere but the project hosting the class can't use the top pom as its parent or else it would be self referencing in the dependencies.

jonenst:
Side Note: in a somewhat shocking plot twist, it turns out that intellij by default correctly imports maven's systempropertyvariables and passes it to it's junit runner ( cf https://www.jetbrains.com/help/idea/running-tests.html and I tested manually using idea-IC-192.6262.58  ), so importing a project with a systempropertyvariable defined for tests works out of the box even when clicking on the little green arrow inside intellij..

In eclipse, the testng runner has this feature too ( https://testng.org/doc/eclipse.html ),  but not the junit runner apparently :(

Does this change anything for you ? I assume it doesn't.

jonenst:
@geofjamg Did you have a solution in mind for the classpath based config ? Currently the config API uses Path objects and hence need a proper filesystem in 2 places:
- ```PlatformConfig.getConfigDir()``` returns a Path object where modules can look for their files (e.g. import-post-processor.js )
- ```ModuleConfig.getPathProperty()``` and its variants where the content of the config file is parsed into a java Path object using the passed in FileSystem object.

I think the long term plan should be to remove filesystems entirely from the config API (breaking change). Did you want to do this now or in another PR ?

An alternative would be to make the classpath based config only usable during tests and in this case we can use jimfs to implement everything.

An even simpler alternative is just to use EmptyPlatformConfigProvider and make the tests work with an empty config: most tests should run with the empty config and can call defaultConfig() but tests requiring specific config must pass it explicitly and can't use defaultConfig()

jonenst:
Note: Can paths be replaced with (using Files.InputStream newInputStream(Path path, OpenOption... o): 
- directly an input stream ? But this opens the file right away.
- a Supplier<InputStream> ? But this uses default OpenOptions.
- a Function<List<OpenOption>, inputStream> ?

Or maybe all of them so that the API is more flexible ?

jonenst:
So I pushed a new implementation called TestPlatformConfigProvider. It uses google reflections https://github.com/ronmamo/reflections and jimfs to provide a platformconfig initiliazed with the values of a folder in src/test/resources/

For now I have not implemented a new system to use this in all the modules so it's with the system properties. (Also there's a hack to use it everywhere automatically that requires jimfs and reflections to be in scope=provided in powsybl-commons but this will be be removed eventually)

jonenst:
Just pushed a version implementings @geofjamg initial proposal:
- powsybl-commons has the interface PlatformConfigProvider
- powsybl-config-standard has the StandardPlatformConfigProvider
- powsybl-config-test has the TestPlatformConfigProvider

PlatformConfig.defaultConfig() throws if the classpath doesn't contain exactly one serviceloader implementation of PlatformConfigProvider. This has 2 consequences:
- exceptions for users upgrading to this version of powsybl
- impossible to test PlatformConfig.getDefault() in powsybl-commons 

To discuss..

jonenst:
Google reflections doesn't work on java9+ https://github.com/ronmamo/reflections/issues/202

config-standard/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+        <dependency>
+            <groupId>com.google.auto.service</groupId>
+            <artifactId>auto-service</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.powsybl</groupId>
```
mathbagu: Same remark

config-standard/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+            <artifactId>powsybl-commons</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
```
mathbagu: Dependencies should be sorted alphabetically (groupId+artefactId)

config-standard/src/main/java/com/powsybl/configstandard/StandardPlatformConfigProvider.java
```diff
@@ -0,0 +1,97 @@
...
+ * $HOME/.itools) and reads configuration from yaml, xml or java properties
+ * files. The config dir names can use the keywords from {@link PlatformEnv}
+ * (e.g. app.root, user.home). It also uses
+ * {@link EnvironmentModuleConfigRepository} to read configuration from
+ * environment variables.
+ *
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ */
+@AutoService(PlatformConfigProvider.class)
+public class StandardPlatformConfigProvider implements PlatformConfigProvider {
```
mathbagu: See remarks about the name of this class

config-standard/src/test/java/com/powsybl/configstandard/StandardConfigTest.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Test
+    public void testNoUserHome() {
+        assertEquals(Arrays.asList("/.itools"), getAbsolutePaths(null));
+    }
+
+    @Test
+    public void testEmptyConfigDirs() {
+        // "FIXME: should return the same as with null but currently returns the working
```
mathbagu: Do we keep this fixme o we fix it right now?

config-test/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+        <dependency>
+            <groupId>com.google.auto.service</groupId>
+            <artifactId>auto-service</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.powsybl</groupId>
```
mathbagu: Same remark here

pom.xml
```diff
@@ -58,6 +58,8 @@
         <module>commons</module>
         <module>computation</module>
         <module>computation-local</module>
+        <module>config-test</module>
+        <module>config-standard</module>
```
mathbagu: Sort module by alphabetical order

action/action-simulator/pom.xml
```diff
@@ -99,6 +99,12 @@
             <artifactId>jimfs</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.powsybl</groupId>
```
mathbagu: Fix the groupId (see previous remarks)

sensitivity-api/pom.xml
```diff
@@ -52,6 +52,12 @@
             <artifactId>jimfs</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.powsybl</groupId>
```
mathbagu: Fix the groupId (see previous remarks)

config-classic/ClassicPlatformConfigProvider.java
```diff
@@ -0,0 +1,97 @@
...
+ * (e.g. app.root, user.home). It also uses
+ * {@link EnvironmentModuleConfigRepository} to read configuration from
+ * environment variables.
+ *
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ */
+@AutoService(PlatformConfigProvider.class)
+public class ClassicPlatformConfigProvider implements PlatformConfigProvider {
+
+    private static final String NAME = "standard";
```
MioRtia: Perhaps you should change the name as `classic` here too since you changed the class' name?

jonenst: Done thanks

config-classic/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifestEntries>
+                            <Automatic-Module-Name>com.powsybl.config-classic</Automatic-Module-Name>
```
MioRtia: I checked because I was not sure and it seems that dashes (`-`) are not allowed in automatic module names and should be replaced by dots? (cf. http://tutorials.jenkov.com/java/modules.html)

config-test/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifestEntries>
+                            <Automatic-Module-Name>com.powsybl.config-test</Automatic-Module-Name>
```
MioRtia: Same remark as above

.gitignore
```diff
@@ -31,6 +31,7 @@
 /commons/target/
 /computation-local/target/
 /computation/target/
+/config-standard/target/
```
MioRtia: It has been renamed as `config-classic`

PR #901, opened 2019-08-23
Import qPercent (coordinated reactive control) for synchronous machines

MioRtia:

No

Feature

CGMES `qPercent` attribute for synchronous machines is not taken into account during import.

It is taken into account

No


MioRtia:
> No need to add a unit test?

Ah good point, I forgot to push the commit!

PR #908, opened 2019-08-29
Correct handling of missing attributes in EI during CGMES import

MioRtia:

No

Bug fix

Inconsistent info can be imported from CGMES files when `regulationStatus` and `regulationTarget` are missing from `EquivalentInjection` outside boundaries.

When `regulationStatus` and `regulationTarget` are missing from `EquivalentInjection` outside boundaries, a warning is written (consistent with CGMES documentation: `EquivalentInjection.regulationStatus and EquivalentInjection.regulationTarget
attributes if the EquivalentInjection is connected to a non-Boundary node.` and the voltage regulation is considered as off for the associated IIDM generator.

No

**Other information**:


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/EquivalentInjectionConversion.java
```diff
@@ -36,14 +38,19 @@ public void convert() {
             targetP = -f.p();
             targetQ = -f.q();
         }
+
         boolean regulationCapability = p.asBoolean("regulationCapability", false);
-        boolean regulationStatus = p.asBoolean("regulationStatus", regulationCapability);
+        boolean regulationStatus = p.asBoolean("regulationStatus", false) && regulationCapability;
+        if (!p.containsKey("regulationStatus") || !p.containsKey(REGULATION_TARGET)) {
+            context.missing(String.format("Missing regulationStatus or regulationTarget for EI %s. Voltage regulation is considered as off.", id));
```
annetill: Maybe a clearest error message with EquivalentInjection instead of EI.

PR #909, opened 2019-08-30
Refactor merged line creation in UcteImporter to solve inconsistent data

murgeyseb:

Bug fix

Currently, when importing merged line from UCTE files, the ID of the line is generated based on the IDs of both dangling lines that compose it.
In order to always have the same ID, independently from the order of the dangling lines, they are sorted before creating the merged line ID.

However, there is no consistency between the order of the dangling lines in the generated ID and side one and side two terminals of the line.

The reordering is still done, but now the created line as consistent connection sides.


ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
```diff
@@ -867,76 +864,84 @@ private void mergeXnodeDanglingLines(UcteNetwork ucteNetwork, Network network) {
 
         Set<DanglingLine> danglingLinesToProcess = Sets.newHashSet(network.getDanglingLines());
         while (!danglingLinesToProcess.isEmpty()) {
-            DanglingLine dl1 = danglingLinesToProcess.iterator().next();
-            DanglingLine dl2 = getMatchingDanglingLine(dl1, danglingLinesByXnodeCode);
+            DanglingLine dlToProcess = danglingLinesToProcess.iterator().next();
+            DanglingLine dlMatchingDlToProcess = getMatchingDanglingLine(dlToProcess, danglingLinesByXnodeCode);
```
mathbagu: It's a detail, but the name of this variable should be `matchingDlToProcess`

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
```diff
@@ -867,76 +864,84 @@ private void mergeXnodeDanglingLines(UcteNetwork ucteNetwork, Network network) {
...
+        // lexical sort to always end up with same merge line id
+        String mergeLineId = dlAtSideOne.getId() + " + " + dlAtSideTwo.getId();
+
+        // create XNODE merge extension
+        float rdp = (float) (dlAtSideOne.getR() / (dlAtSideOne.getR() + dlAtSideTwo.getR()));
+        float xdp = (float) (dlAtSideOne.getX() / (dlAtSideOne.getX() + dlAtSideTwo.getX()));
+        double xnodeP1 = dlAtSideOne.getP0();
+        double xnodeQ1 = dlAtSideOne.getQ0();
+        double xnodeP2 = dlAtSideTwo.getP0();
+        double xnodeQ2 = dlAtSideTwo.getQ0();
```
mathbagu: Maybe you should respect your own style by removing this variable used once and call `dlAtSideOne.getP0()` (and the others) in the adder.

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
murgeyseb: Well, you can also consider that it is the dangling line that matches the dangling line to process.
Which one do you prefer ?

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
murgeyseb: They are used twice (also in the merged Xnode extension). Do I really change it ?

mathbagu: I didn't notice that these variables were used twice. Keep it like this

PR #917, opened 2019-09-03
CGMES correction: rho and alpha of PTC depend on the side where the PTC is

### LINKED ISSUE #905: Wrong conversion of PTC

The conversion of symmetrical and asymmetrical PTC from CGMES to IIDM has to be reviewed. 

1) The side where the PTC is located is not taken into account. I think that it is missing because rho and alpha have to be changed when we move the PTC from side 2 to side 1 in IIDM.  

2) The computation of rho is wrong for asymmetrical PTC. In the code, we have:
double alpha = Math.atan2(dy, 1 + dx);
double rho = Math.hypot(dy, 1 + dx);

1) We have to introduce alpha = - alpha and rho = 1 / rho when the PTC is located on side 2.

2) We have to do:
double alpha = Math.atan2(dy, 1 + dx);
double rho = 1 / Math.hypot(dy, 1 + dx);

IOP June wrong on TransnetBW/REN/REE and on ELES.

  - PowSyBl Version: ...
  - OS Version: ...

MioRtia:

Fixes #905

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/PhaseTapChangerConversion.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/PhaseTapChangerConversion.java
@@ -290,15 +290,19 @@ public class PhaseTapChangerConversion extends AbstractIdentifiedObjectConversio
             int n = step - neutralStep;
             double dx = (n * du - du0) * Math.cos(theta);
             double dy = (n * du - du0) * Math.sin(theta);
-            double alpha = Math.atan2(dy, 1 + dx);
-            double rho = Math.hypot(dy, 1 + dx);
+            double angle = Math.atan2(dy, 1 + dx);
+            double ratio = Math.hypot(dy, 1 + dx);
 
             // CGMES uses ratio to define the relationship between voltage ends while IIDM uses rho
             // ratio and rho as complex numbers are reciprocals. Given V1 and V2 the complex voltages at end 1 and end 2 of a branch we have:
             // V2 = V1 * rho and V2 = V1 / ratio
             // This is why we have: rho=1/ratio and alpha=-angle
-            alphas.add(-alpha);
-            rhos.add(1 / rho);
+            double alpha = -angle;
+            double rho = 1 / ratio;
+
+            // In IIDM, all PTC must be side one
+            alphas.add(side == 1 ? alpha : -alpha);
+            rhos.add(side == 1 ? rho : 1 / rho);
 
             LOG.debug("ACTUAL    n,dx,dy,alpha,rho  {} {} {} {} {}", n, dx, dy, alpha, rho);
         }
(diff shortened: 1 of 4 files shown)


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/PhaseTapChangerConversion.java
```diff
@@ -313,33 +317,37 @@ private void fillAlphaRhoListsSymmetrical(
...
+                alphas.add(side == 1 ? alpha : -alpha);
+                rhos.add(1.0);
             }
         } else {
             for (int step = lowStep; step <= highStep; step++) {
                 int n = step - neutralStep;
                 double dy = (n * du / 2 - du0) * Math.sin(theta);
-                double alpha = 2 * Math.asin(dy);
-                double rho = 1.0;
+                double angle = 2 * Math.asin(dy);
```
annetill: I think it is better to leave ratio = 1.0; for a better understanding of the code.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/PhaseTapChangerConversion.java
```diff
@@ -313,33 +317,37 @@ private void fillAlphaRhoListsSymmetrical(
...
 
                 // CGMES uses ratio to define the relationship between voltage ends while IIDM uses rho
                 // ratio and rho as complex numbers are reciprocals. Given V1 and V2 the complex voltages at end 1 and end 2 of a branch we have:
                 // V2 = V1 * rho and V2 = V1 / ratio
                 // This is why we have: rho=1/ratio and alpha=-angle
-                alphas.add(-alpha);
-                rhos.add(1 / rho);
+                double alpha = -angle;
```
annetill: Maybe rho = 1 / ratio ; ?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/PhaseTapChangerConversion.java
```diff
@@ -313,33 +317,37 @@ private void fillAlphaRhoListsSymmetrical(
...
+
+                // In IIDM, all PTC must be side one
+                alphas.add(side == 1 ? alpha : -alpha);
+                rhos.add(1.0);
```
annetill: Prefer rho instead of 1.0

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/PhaseTapChangerConversion.java
```diff
@@ -313,33 +317,37 @@ private void fillAlphaRhoListsSymmetrical(
...
 
-                LOG.debug("ACTUAL    n,dy,alpha,rho  {} {} {} {}", n, dy, alpha, rho);
+                LOG.debug("ACTUAL    n,dy,alpha,rho  {} {} {} {}", n, dy, alpha, 1.0);
```
annetill: Prefer rho instead of 1.0

PR #922, opened 2019-09-05
Remove deprecated sonar conf. Try to use the default conf.

jonenst:

Fix coverage report on sonarcloud
- sonar.jacoco.reportPath (merged exec) was deprecated and replaced with sonar.jacoco.reportPaths (multiple execs)
- sonar.jacoco.reportPaths (multiple execs) is deprecated and replaced with sonar.coverage.jacoco.xmlReportPaths (xml reports)

The coverage report dropped 10% after https://github.com/powsybl/powsybl-core/pull/893 .
This may be because we use a doubly deprecated jacoco-merge.exec system.

Try to use the xml reports:
https://sonarcloud.io/documentation/analysis/coverage/

> sonar.jacoco.reportPaths    Deprecated. Use sonar.coverage.jacoco.xmlReportsPath. Path to JaCoCo reports in binary format. Supported only for Java

>  sonar.coverage.jacoco.xmlReportPaths | Comma-delimited list of paths to JaCoCo XML coverage reports.

NO

### DIFF
--- a/distribution-core/pom.xml
+++ b/distribution-core/pom.xml
@@ -463,26 +463,6 @@
             <build>
                 <plugins>
                     <plugin>
-                        <groupId>org.apache.maven.plugins</groupId>
-                        <artifactId>maven-antrun-plugin</artifactId>
-                        <version>${maven.antrun.version}</version>
-                        <executions>
-                            <execution>
-                                <phase>verify</phase>
-                                <goals>
-                                    <goal>run</goal>
-                                </goals>
-                                <configuration>
-                                    <target name="copy jacoco reports">
-                                        <copy todir="${project.build.directory}/jacoco" flatten="true">
-                                            <fileset dir=".." includes="**/jacoco*.exec" />
-                                        </copy>
-                                    </target>
-                                </configuration>
-                            </execution>
-                        </executions>
-                    </plugin>
-                    <plugin>
                         <groupId>org.jacoco</groupId>
                         <artifactId>jacoco-maven-plugin</artifactId>
                         <executions>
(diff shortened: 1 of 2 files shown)


distribution-core/pom.xml
```diff
@@ -462,26 +462,6 @@
             </activation>
             <build>
                 <plugins>
-                    <plugin>
-                        <groupId>org.apache.maven.plugins</groupId>
-                        <artifactId>maven-antrun-plugin</artifactId>
-                        <version>${maven.antrun.version}</version>
```
mathbagu: To be confirmed, but it should be possible to remove this property in the parent pom.xml

pom.xml
```diff
@@ -109,11 +109,12 @@
...
 
-        <sonar.jacoco.reportPaths>
-            ../target/jacoco-merge.exec,
-            ../../target/jacoco-merge.exec,
-            ../../../target/jacoco-merge.exec
-        </sonar.jacoco.reportPaths>
+        <sonar.coverage.jacoco.xmlReportPaths>
+            ../distribution-core/target/site/jacoco-aggregate/jacoco.xml,
+            ../../distribution-core/target/site/jacoco-aggregate/jacoco.xml,
+            ../../../distribution-core/target/site/jacoco-aggregate/jacoco.xml
```
mathbagu: What is the most robust to a change: `../target/site/jacoco-aggregate/jacoco.xml` vs `../distribution-core/target/site/jacoco-aggregate/jacoco.xml`

distribution-core/pom.xml
jonenst: removed

pom.xml
jonenst: As discussed, the report is generated by jacoco:report-aggregate defined in the pom.xml of distribution-core. Therefore by default, the report is generated in distribution-core/target/...

We could change the outputDirectory of the jacoco:report-aggregate to output it in the target of the top project, but we prefer to use the default conf of jacoco:report-aggregate to have less surprise for developers looking for these reports

Issue #915, opened 2019-09-02
New contingency list API


geofjamg:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
Contingency API could be improved and could be more user friendly.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**
Rename `ContingenciesProvider` to `ContingencyList`
Allow better programmatic initialization:
```java
// manual list
ContingencyList l = ContingencyList.of(Contingency.line("L1"), Contingency.generator("L1"));

// automatic list
Network n = ...
ContingencyList l2 = ContingencyList.lines(n);

// automatic list with filter
ContingencyList l3 = ContingencyList.lines(n, l -> l.getTerminal1().getVoltageLevel().getNominalV() > 300);
```

* **What is the motivation / use case for changing the behavior?**
Having a more friendly contingency API.

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

mathbagu:
For N-K contingencies, I propose to have something like this:
```
ContingencyList l4 = ContingencyList.of(
    Contingency.builder()
        .line("L1")
        .generator("G2")
        .build(),
    Contingency.lines("L1", "L2", "L3")
);
```

mathbagu:
It could be great to be able to load a contingencyList from a file (groovy or Json) like this:
`ContigencyList.load(file)`

Depending on the extension, it could use the relevant parser.

geofjamg:
> It could be great to be able to load a contingencyList from a file (groovy or Json) like this: `ContigencyList.load(file)`
> 
> Depending on the extension, it could use the relevant parser.

That would be indeed a nice feature. But in that case we also should be able to save it?

PR #925, opened 2019-09-06
Add helper to create contingencies or contingency lists more easily

mathbagu:

#915 

Feature

The contingency API is not always user friendly to use.

The new API offer lot of helper functions to create contingencies or contingency lists.

**Other information**:

### DIFF
--- a/contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyList.java
+++ b/contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyList.java
@@ -1,0 +1,59 @@
+/**
+ * Copyright (c) 2020, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package com.powsybl.contingency;
+
+import com.powsybl.iidm.network.Network;
+import org.apache.commons.io.FilenameUtils;
+
+import java.io.BufferedInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+/**
+ * @author Mathieu Bague <mathieu.bague@rte-france.com>
+ */
+public interface ContingencyList {
+
+    /**
+     * Return the name of this contingency list
+     */
+    String getName();
+
+    /**
+     * Return a list of contingency for the given network.
+     */
+    List<Contingency> getContingencies(Network network);
+
+    /**
+     * Load a {@link ContingencyList} from a path
+     *
+     * @param path The file to load
+     *
+     * @return a contingency list
+     */
+    static ContingencyList load(Path path) {
+        try (InputStream stream = new BufferedInputStream(Files.newInputStream(path))) {
+            String filename = path.getFileName().toString();
+            ContingencyListLoader loader = ContingencyLists.getLoader(FilenameUtils.getExtension(filename));
+            return loader.load(filename, stream);
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
+
+    /**
+     * Create a contingency list from a set of contingencies
+     */
+    static ContingencyList of(Contingency... contingencies) {
+        return new DefaultContingencyList(contingencies);
+    }
+}
--- a/contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyList.java
+++ b/contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyList.java
@@ -43,6 +43,22 @@ public interface ContingencyList {
     static ContingencyList load(Path path) {
         try (InputStream stream = new BufferedInputStream(Files.newInputStream(path))) {
             String filename = path.getFileName().toString();
+            return load(filename, stream);
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
+
+    /**
+     * Load a {@link ContingencyList} from an input stream of given file name
+     *
+     * @param filename The input file name
+     * @param stream The input stream to load
+     *
+     * @return a contingency list
+     */
+    static ContingencyList load(String filename, InputStream stream) {
+        try {
             ContingencyListLoader loader = ContingencyLists.getLoader(FilenameUtils.getExtension(filename));
             return loader.load(filename, stream);
         } catch (IOException e) {
--- a/contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyList.java
+++ b/contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyList.java
@@ -59,7 +59,7 @@ public interface ContingencyList {
      */
     static ContingencyList load(String filename, InputStream stream) {
         try {
-            ContingencyListLoader loader = ContingencyLists.getLoader(FilenameUtils.getExtension(filename));
+            ContingencyListLoader loader = ContingencyListLoaderProvider.getLoader(FilenameUtils.getExtension(filename));
             return loader.load(filename, stream);
         } catch (IOException e) {
             throw new UncheckedIOException(e);
--- a/contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyList.java
+++ b/contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyList.java
@@ -7,7 +7,6 @@
 
 package com.powsybl.contingency;
 
-import com.powsybl.commons.Versionable;
 import com.powsybl.iidm.network.Network;
 import org.apache.commons.io.FilenameUtils;
 
@@ -22,7 +21,12 @@ import java.util.List;
 /**
  * @author Mathieu Bague <mathieu.bague@rte-france.com>
  */
-public interface ContingencyList extends Versionable {
+public interface ContingencyList {
+
+    /**
+     * Return the name of this contingency list
+     */
+    String getName();
 
     /**
      * Return a list of contingency for the given network.
--- a/contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyList.java
+++ b/contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyList.java
@@ -7,6 +7,7 @@
 
 package com.powsybl.contingency;
 
+import com.powsybl.commons.Versionable;
 import com.powsybl.iidm.network.Network;
 import org.apache.commons.io.FilenameUtils;
 
@@ -21,12 +22,7 @@ import java.util.List;
 /**
  * @author Mathieu Bague <mathieu.bague@rte-france.com>
  */
-public interface ContingencyList {
-
-    /**
-     * Return the name of this contingency list
-     */
-    String getName();
+public interface ContingencyList extends Versionable {
 
     /**
      * Return a list of contingency for the given network.
--- a/contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyLists.java
+++ b/contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyLists.java
@@ -1,0 +1,33 @@
+/**
+ * Copyright (c) 2020, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package com.powsybl.contingency;
+
+import com.powsybl.commons.util.ServiceLoaderCache;
+
+import java.util.Objects;
+
+/**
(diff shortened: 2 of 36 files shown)


jonenst:
Proposals:
- The change on checkSidedContingency changes should go in a different commit ?
- Rename ContingencyList to ContingencyListProvider ? or ContingencyListSupplier ? or ContingenciesSupplier ? Or something else ?
- remove or fix the commented code ?

geofjamg:
>     * Rename ContingencyList to ContingencyListProvider ? or ContingencyListSupplier ? or ContingenciesSupplier ? Or something else ?
> 

See related issue, the idea of the PR is to have a user friendly API. ContingencyList is more natural and business oriented that ContingencyListProvider which is a technical and really not friendly term.

mathbagu:
I think it could be interesting to have a `ContingencyList.load(Path)` to create a ContingencyList from a file. But for groovy scripts, we also need a network to create the contingencies (and retrieve their types).
Do we have to create two different methods:
```
static ContingencyList load(Path file);
static ContingencyList load(Network network, Path file);
```

jonenst:
> ContingencyList is more natural and business oriented that ContingencyListProvider which is a technical and really not friendly term.

Can we keep ContingenciesProvider for the interface and put all the new static methods in the ContingencyList utility class (or Contingencies utility class) ?

This improves things for the users, and keeps things clear for the developers (not having an object named ContingencyList that is not a list, but a list provider)

geofjamg:
> > ContingencyList is more natural and business oriented that ContingencyListProvider which is a technical and really not friendly term.
> 
> Can we keep ContingenciesProvider for the interface and put all the new static methods in the ContingencyList utility class (or Contingencies utility class) ?
> 
> Why not but in any case we should rename  ContingenciesProvider to ContingencyListProvider. And ok for an additional ContingencyList as an utility class.  @mathbagu also agree with this proposition?

mathbagu:
Just to be sure I understand well the proposal:
- we rename `ContingenciesProvider` to `ContingencyListProvider`
- the ContingencyList interface is converted to an utility class
- the `DefaultContingencyList` is renamed to `DefaultContingencyListProvider`

geofjamg:
> Just to be sure I understand well the proposal:
> 
>     * we rename `ContingenciesProvider` to `ContingencyListProvider`
> 
>     * the ContingencyList interface is converted to an utility class
> 
>     * the `DefaultContingencyList` is renamed to `DefaultContingencyListProvider`
> 
> 
> A thumb-up is sufficient if you agree

Maybe another question: 
Would it be a good idea to model a contingency list with a proper object instead of a list.
So instead of 
```java
class ContingencyListProvider {
    List<Contingency> getContingencies(Network network);
}
```
we would have:
```java
class ContingencyListProvider {
    ContingencyList getContingencies(Network network);
}
```

And `ContingencyList` would be a wrapper around a List<Contingency> but **it would allow to also name the contingency** which would be very convenient we it comes to integrate it in a platform where we want to manager several contingency lists.

Something like:

```java
class ContingencyList {
  
    // all static initializers of(...)

   private final String name;

   private final List<Contingency> contingencies;
)
```

And of course `ContingencyList` with this design would not be only a utility class.

sylvlecl:
I would think it makes sense to have an interface `ContingencyList`, additionally to the `ContingencyListProvider` interface: it would leave the floor to have more structured representations of lists than today.

For example, today there is no model for a generic list of "All 400 kV lines", which looks like a missing feature. Another missing feature would be serialization.

In particular, for our third-party project, we are planning to have a structured visualization of the list of contingencies : it will require to have a structured model of named lists.

An alternative is to have this structured model as an implementation of `ContingencyListProvider`, but I don't think it's the most powerful design. We need to define carefully the responsibility of the provider, of the provider factory, and of the list. In the current state, the provider has 2 responsibilities : 
1. reading a file 
2. creating a list of actual contingencies for a specific network

I think it's important to separate those 2 steps : we should be able to read a model from a file (and write back to a file), and then instantiate actual contingencies based on this model.
For this reason it may be a good idea to have:
```java
interface ContingencyListProvider {
    ContingencyList getContingencies(Network network);
}
```
```java
interface ContingencyList {
    String getName();
    List<Contingency> getContingencies(Network network);
}
```

It may look like the 2 methods are redundant, but the first one could use the network at the step of model creation from a file (say groovy), and the second one could use the network at the step of instantiating contingencies from the list model.

We could provide a kind of java DSL to create programmatically this model. The model could look like an AST, easily serializable.

jonenst:
```java
interface ContingencyListProvider {
    ContingencyList getContingencies(Network network);
}

interface ContingencyList {
    String getName();
    List<Contingency> getContingencies(Network network);
}
```
I think that this is confusing:
- The second one is called ContingencyList but is not a list. (same remark as before)
- The names are too similar. Maybe your ContigencyList that "read a model from a file (and write back to a file)," should be called ContingencyModel ?
- should there be a method like this somewhere ?
```java
List<Contingency> getContingencies(ContingencyList model)
```

sylvlecl:
> * The names are too similar. Maybe your ContigencyList that "read a model from a file (and write back to a file)," should be called ContingencyModel ?

Hmmm I don't think it's usual to have "model" in a class name (a large part of classes are actually models of some business domain object), and it should be plural. Possibly `Contingencies` ? But what it creates is indeed a `List` eventually, and it represents a "list" of contingencies in the natural language sense.

> * should there be a method like this somewhere ?

Only the implementation will be able to create the actual contingencies and it will need the network for this. What I imagine here is a list based on criteria such as "nominal voltage > 200kV".

For example, it could be nice to create generic contingencies lists in a way similar to:
```java
ContingencyList contingencies = N1Contingencies.of(Lines
        .withCountry(Country.FR)
        .withNominalVoltageGreaterThan(200));
```

contingency/contingency-api/src/test/java/com/powsybl/contingency/ContingencyTest.java
```diff
@@ -29,16 +28,19 @@
 
     @Test
     public void test() {
-        ContingencyElement element1 = new BranchContingency("line");
-        ContingencyElement element2 = new GeneratorContingency("generator");
-        Contingency contingency = new Contingency("contingency", element1, element2);
+        Contingency contingency = Contingency.builder("contingency")
+                .branch("line")
```
geofjamg, line 32: Maybe we should add a prefix like 'add' to better figure out that we add somthing to the builder and don't replace (like in guava ImmutableList) ?
```java
Contingency contingency = Contingency.builder("contingency")
    .addBranch("")
    .addGenerator("")
   .build()            
```

contingency/contingency-api/src/main/java/com/powsybl/contingency/tasks/LineTripping.java
```diff
@@ -0,0 +1,35 @@
...
+
+package com.powsybl.contingency.tasks;
+
+import com.powsybl.commons.PowsyblException;
+import com.powsybl.iidm.network.Network;
+
+/**
+ * @author Mathieu Bague <mathieu.bague@rte-france.com>
+ */
+public class LineTripping extends BranchTripping {
```
geofjamg, line 16: Do we really need to subclass BranchTripping as we do not add anything from base class (just more precise error message) ?

murgeyseb:
Recently, DanglingLineTripping has been added in contingency API, and should be added in this PR.

sylvlecl (commented):
Mostly minor comments + one more important design question about contingencies validation.

Otherwise the API looks good !

contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyLists.java
```diff
@@ -0,0 +1,33 @@
...
+package com.powsybl.contingency;
+
+import com.powsybl.commons.util.ServiceLoaderCache;
+
+import java.util.Objects;
+
+/**
+ * @author Mathieu Bague <mathieu.bague@rte-france.com>
+ */
+final class ContingencyLists {
```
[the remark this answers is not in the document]
tadam50: We could keep it private and rename it to ContingencyListLoaderProvider ?

contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyList.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+public interface ContingencyList {
+
+    /**
+     * Return the name of this contingency list
+     */
+    String getName();
+
+    /**
+     * Return a list of contingency for the given network.
```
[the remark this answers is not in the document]
tadam50: Will be done in different PR linked with new issue '#1532'

contingency/contingency-dsl/src/main/java/com/powsybl/contingency/dsl/GroovyContingencyList.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * @author Mathieu Bague <mathieu.bague@rte-france.com>
+ */
+public class GroovyContingencyList implements ContingencyList {
+
+    private final String name;
+
+    private final GroovyCodeSource codeSource;
+
+    GroovyContingencyList(String name, InputStream stream) {
```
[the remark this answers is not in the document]
tadam50: User will only use "loader" mechanism, in order to hide complexicity

contingency/contingency-api/src/main/java/com/powsybl/contingency/json/JsonContingencyListLoader.java
```diff
@@ -0,0 +1,38 @@
...
+    public String getFormat() {
+        return "json";
+    }
+
+    @Override
+    public ContingencyList load(String name, InputStream stream) throws IOException {
+        ObjectMapper objectMapper = JsonUtil.createObjectMapper();
+        objectMapper.registerModule(new ContingencyJsonModule());
+
+        return objectMapper.readValue(stream, DefaultContingencyList.class);
```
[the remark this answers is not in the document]
tadam50: Will be done in different PR linked with new issue '#1533' based on your comment

contingency/contingency-api/src/main/java/com/powsybl/contingency/LineContingency.java
```diff
@@ -0,0 +1,50 @@
+/**
+ * Copyright (c) 2017, All partners of the iTesla project (http://www.itesla-project.eu/consortium)
```
mathbagu, line 2: Fix the copyright

contingency/contingency-api/src/main/java/com/powsybl/contingency/TwoWindingsTransformerContingency.java
```diff
@@ -0,0 +1,50 @@
+/**
+ * Copyright (c) 2017, All partners of the iTesla project (http://www.itesla-project.eu/consortium)
```
mathbagu, line 2: Fix the copyright

contingency/contingency-api/src/main/java/com/powsybl/contingency/json/DefaultContingencyListSerializer.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+public class DefaultContingencyListSerializer extends StdSerializer<DefaultContingencyList> {
+
+    public DefaultContingencyListSerializer() {
+        super(DefaultContingencyList.class);
+    }
+
+    @Override
+    public void serialize(DefaultContingencyList contingencyList, JsonGenerator jsonGenerator, SerializerProvider serializerProvider) throws IOException {
+        jsonGenerator.writeStartObject();
```
geofjamg, line 28: A version is missing

contingency/contingency-api/src/main/java/com/powsybl/contingency/json/DefaultContingencyListDeserializer.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public DefaultContingencyListDeserializer() {
+        super(DefaultContingencyList.class);
+    }
+
+    public DefaultContingencyList deserialize(JsonParser parser, DeserializationContext ctx) throws IOException {
+        String name = null;
+        List<Contingency> contingencies = Collections.emptyList();
+
+        while (parser.nextToken() != JsonToken.END_OBJECT) {
```
geofjamg, line 36: Version check is missing

contingency/contingency-api/src/main/java/com/powsybl/contingency/json/JsonContingencyListLoader.java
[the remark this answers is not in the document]
geofjamg: @sylvlecl is right and it has to be done in this PR otherwise we won't be backward compatible when adding new contingency list type. The solution with @c is really ugly (I known this is what jackson do for that). We should add a type field in the list to store the contingency list type.

When we will add for example an automatic contingency list based on filtering criteria (like country, voltage level), serialization will be json too. How we will be able to deserialize it ? Same contingency loader right because associated to the json extension? So we need to add a type field? But in that case it is not really expandable as it will be hardcoded in the Json contingency list loader the mapping between the type and the contingency list subclass?

geofjamg (changes requested):
I am wondering if we missed something in this PR related to different kinds of contingency list serialized in the same format (json). How to we manage that, on the loading part? Does it has to be extendable? 
Also whereever there is a serializer we should have a version to allow breaking change and backward compatibility in next releases.

Issue #831, opened 2019-06-13
Addition of a storage version in NodeInfo serialization within MapDB


niconoir:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
Currently, no information is serialized in NodeInfo within MapDB about how the data have been serialize and how to deserialize them.

* **What is the expected behavior?**
A new information (storage version) has to be serialized in NodeInfo serialization within MapDB in order to know how the data have been serialized and how to deserialize them.

* **What is the motivation / use case for changing the behavior?**
NodeInfo class will evolve in order to add new informations about the AFS node (see #827 for example). Thus, serialization of NodeInfo within MapDB will evolve, and we need to know how NodeInfo we are reading in MapDB has been serialized in order to deserialize it properly. A storage version included in the NodeInfo serialization will allow us to keep track of the different ways the NodeInfo have been serialized and to deserialize them properly. This will allow us to keep a backward compatibility and deserialize data from older versions from MapDB.
N.B: When this feature will be implemented, the data previously stored in MapDB will not be compatible with the new version.

geofjamg:
AFS is not part anymore of core

PR #926, opened 2019-09-09
Addition of a storage version in NodeInfo serialization within MapDB (#831)

niconoir:

#831


niconoir:
Yes, we can anticipate changes in other afs components storage. But in this case, do we consider afs as a whole and use only one storage version common to all afs components, or do we use a different storage version for each component ?

niconoir:
I added a storage version number, common to all afs elements. I did not add the storage version for List or Set serializers, since the storage version will already be contained in each serialized element of the list (or set).

afs/afs-mapdb-storage/src/main/java/com/powsybl/afs/mapdb/storage/NamedLinkSerializer.java
```diff
@@ -22,12 +22,14 @@
...
     @Override
     public void serialize(DataOutput2 out, NamedLink namedLink) throws IOException {
+        out.writeInt(MapDbStorageConstants.STORAGE_VERSION);
         UuidSerializer.INSTANCE.serialize(out, namedLink.getNodeUuid());
         out.writeUTF(namedLink.getName());
     }
 
     @Override
     public NamedLink deserialize(DataInput2 input, int available) throws IOException {
+        input.readInt(); // Storage version is retrieved here
```
geofjamg: Maybe we could get the version and throw an exception if not version 0 to not forget to update the code the day we switch to version 1?

niconoir: I'm not sure this would ensure the code will be properly updated, since the first developer to move from version 0 to version 1 will correctly do the update (and remove exception throwing), but following developers using version 1 too may then not be aware they have to update their code. Proper unit tests may be sufficient to verify code update was properly made ?

geofjamg: > 
> 
> I'm not sure this would ensure the code will be properly updated, since the first developer to move from version 0 to version 1 will correctly do the update (and remove exception throwing), but following developers using version 1 too may then not be aware they have to update their code. Proper unit tests may be sufficient to verify code update was properly made ?

Let's keep it as it is and see how to handle migration when we will need to. 
Could you update the migration guide to warm users that they have to destroy their db when upgrading?
https://github.com/powsybl/powsybl-core/wiki/Migration-guide-v3.1.0

PR #929, opened 2019-09-10
Add methods to retrieve the HVDCLine from the converter stations

mathbagu:

N/A

Feature

It's impossible to easily retrieve the HVDC line from the converter station

It's now possible thanks to a new getter.

**Other information**:


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/HvdcLineImpl.java
```diff
@@ -152,6 +157,11 @@ public HvdcLineImpl setActivePowerSetpoint(double activePowerSetpoint) {
         return this;
     }
 
+    @Override
+    public AbstractHvdcConverterStation<?> getConverterStation(Side side) {
+        return (side == Side.ONE) ? getConverterStation1() : getConverterStation2();
```
geofjamg: Is it necessary as it is already defined as a default method in HvdcLine interface?

mathbagu: If you want to remove one day, the default implementation in `HvdcLine` interface, you have to implement this method in all the sub classes. I can also consider there is only one implementation of IIDM for now.
What do you prefer?

PR #932, opened 2019-09-11
Bug fix in getConnectableBus() in calculated BBranch topology from NB network

MioRtia:

No

Bug fix

There are several issues with calculated bus branch topology from a node breaker network:
- `connectableBus` of an injection can be `null` (which is **contrary** to specification)
- for an unconnected injection, `connectableBus` depends on the swiches' traversing order in a voltage level since it is reinitialized at each iteration in switch even if a `connectableBus` is found
- for an unconnected injection, `connectableBus` was actually always the first bus of the voltage level

- `connectableBus` is never null: if none is found, the first bus of the voltage level is returned
- for an unconnected injection, `connectableBus` is independant of the switches' order (if found, it is not reinitialized) and is not always the first bus of the voltage level

No

**Other information**:

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevel.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevel.java
@@ -387,6 +387,9 @@ class NodeBreakerVoltageLevel extends AbstractVoltageLevel {
             // node associated to a bus
             BusExt[] connectableBus2 = new BusExt[1];
             graph.traverse(node, (v1, e, v2) -> {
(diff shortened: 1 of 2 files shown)


geofjamg:
> 
> 
> @geofjamg @mathbagu There is a second case that @zamarrenolm showed me: if there is no bus in a voltage level (every switch is open), the connectable bus cannot be found and the method returns an exception.

It is another issue, more difficult to solve. Let's keep it for another PR.

MioRtia:
I created two issues (#938 and #939) to trace the issues discussed in this PR. For now, I kept the first proposition (`getConnectableBus` returns the first bus it finds) but it can be changed in another PR. @mathbagu do you agree with this?

PR #911, opened 2019-08-30
Replace Identifiable.getProperties by getProperty & setProperty

tadam50:
Signed-off-by: Thomas ADAM <tadam@silicom.fr>

**Please check if the PR fulfills these requirements** *(please use `'[x]'` to check the checkboxes, or submit the PR and then click the checkboxes)*
- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**Does this PR already have an issue describing the problem ?** *If so, link to this issue using `'#XXX'` and skip the rest*
No

**What kind of change does this PR introduce?** *(Bug fix, feature, docs update, ...)*
Change access to properties fields of Identifiable class

**What is the current behavior?** *(You can also link to an open issue here)*
Identifiable.getProperties is used to set new properties

**What is the new behavior (if this is a feature change)?**
Identifiable.getProperties method is no longer accessible, replaced by getProperty & setProperty

**Does this PR introduce a breaking change or deprecate an API?** *If yes, check the following:*
- [x] The *Breaking Change* or *Deprecated* label has been added
- [ ] The migration guide has been updated in the github wiki *(What changes might users need to make in their application due to this PR?)*

**Other information**:

(if any of the questions/checkboxes don't apply, please delete them entirely)

PR #941, opened 2019-09-17
Fix IIDM Groovy property access

geofjamg:

No

A regression fix in PR #911

It is not possible anymore to set a non `java.lang.String` property using Groovy syntax:
```groovy
s.ok = true
```
It only works with `java.lang.String`

It works for any Java type.

**Other information**:


mathbagu:
> Everything works fine with this correction

Thank you for your feedback!

PR #946, opened 2019-09-18
generic mechanism for naming the tap changer attribute in transformers

zamarrenolm:

In the future three winding transformers will have phase and ratio tap changers at any one of the three legs.

The method `getTapChangerAttribute` in the `AbstractTapChanger` class builds a proper name for the attribute depending on the type of tap changer and its location in the parent transformer. 

For two winding transformers the name of the attributes must be "`ratioTapChanger`" and "`phaseTapChanger`".

Currently three winding transformers can only have `ratioTapChanger2` and `ratioTapChanger3`. In the future they should be able to return `ratioTapChanger1`, `phaseTapChanger1`, `ratioTapChanger2`, `phaseTapChanger2` and `ratioTapChanger3`, `phaseTapChanger3`.

Because the attribute name is now composed, a `Supplier<String>` has been used, following the same optimization already made for combining the tap changer attribute name and its properties.

It is not a breaking change, it only affects the implementation of IIDM, not the public API.

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/PhaseTapChangerImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/PhaseTapChangerImpl.java
@@ -61,7 +61,7 @@ class PhaseTapChangerImpl extends AbstractTapChanger<TwoWindingsTransformerImpl,
         ValidationUtil.checkPhaseTapChangerRegulation(parent, regulationMode, getRegulationValue(), isRegulating(), getRegulationTerminal(), getNetwork());
         RegulationMode oldValue = this.regulationMode;
         this.regulationMode = regulationMode;
-        notifyUpdate(() -> getTapChangerAttribute().get() + ".regulationMode", oldValue, regulationMode);
+        notifyUpdate(() -> getTapChangerAttribute() + ".regulationMode", oldValue, regulationMode);
         return this;
     }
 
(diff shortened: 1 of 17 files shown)


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/PhaseTapChangerImpl.java
```diff
@@ -121,7 +121,7 @@ public void allocateVariantArrayElement(int[] indexes, final int sourceIndex) {
     }
 
     @Override
-    protected String getTapChangerAttribute() {
-        return "phaseTapChanger";
+    protected Supplier<String> getTapChangerAttribute() {
+        return () -> "phase" + parent.getTapChangerAttribute();
```
mathbagu: It's hard to say but is it necessary to return a supplier here, specially if this method is always called from a supplier. Suppliers have been added to avoid string creation by concatenation, but with this solution we have two levels of `Supplier`s.

From my point of view, it's equivalent to keep a `String` here.

zamarrenolm: I added the `supplier` to follow the same optimization that was made when building the string for the `notifyUpdate` methods in the `NetworkListener`. 

I agree that if `getTapChangerAttribute` is called only in the context of `notifyUpdate` methods that already contain the `Supplier` it is redundant.

If there is no other use for this method, I agree on keeping the return value as a string.

Please confirm and I will make the change.

mathbagu: I’m pretty sure that these methods are used only in that context. So I suggest to change the return type to String.
If we encounter any performance issue, we’ll fix it.

zamarrenolm: Agree. I will make the change

PR #947, opened 2019-09-18
Implementation of Identifiable::getNetwork()

tadam50:

No

Adding following method into Identifiable interface:
Network getNetwork()

Identifiable interface do not offer getter on Network used

Identifiable interface offer getter on Network used

**Other information**:

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiable.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiable.java
@@ -38,6 +38,9 @@ abstract class AbstractIdentifiable<I extends Identifiable<I>> extends AbstractE
         return name != null ? name : id;
     }
 
+    @Override
+    public abstract NetworkImpl getNetwork();
+
     protected abstract String getTypeDescription();
 
     @Override
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/MergedBus.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/MergedBus.java
@@ -88,6 +88,11 @@ class MergedBus extends AbstractIdentifiable<Bus> implements CalculatedBus {
     }
 
     @Override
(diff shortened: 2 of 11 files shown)


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiable.java
```diff
@@ -38,6 +38,9 @@ public String getName() {
         return name != null ? name : id;
     }
 
+    @Override
+    public abstract NetworkImpl getNetwork();
```
geofjamg: Is it really useful?

mathbagu: Probably not except if we need a NetworkImpl to notify. @tadam50 could you confirm this?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/MergedBus.java
```diff
@@ -87,6 +87,11 @@ public void invalidate() {
         buses.clear();
     }
 
+    @Override
+    public NetworkImpl getNetwork() {
+        return (NetworkImpl) getVoltageLevel().getSubstation().getNetwork();
```
mathbagu: In AbstractBus, you use voltageLevel.getNetwork(). Do the same thing here?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiable.java
tadam50: I need NetworkImpl in order to notify changes on setProperty method

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/MergedBus.java
tadam50: My mistake, i will change that

PR #951, opened 2019-09-20
[WIP] Three windings transformer model update 

zamarrenolm:
This PR updates the IIDM model for three windings transformers. It also contains a refactored version of the conversion from CGMES data.

Details about the changes in the three windings transformer model:

- A ratio and phase tap changer may be present at any leg.
- A phase angle clock attribute is stored for each leg. Phase angle clock is taken into account by some TSOs when computing power flow results.
- A shunt admittance is defined explicitly at each leg. 

For two windings transformers, the phase angle clock attribute can also be defined for each side.

The old and new version of the CGMES conversion are available in the current code. The refactored version is activated by the flag `EXTENDED_CGMES_CONVERSION` in the class `Conversion` of the `cgmes/cgmes-conversion` module.

The IIDM API for two and three winding transformers has been extended, but has been kept compatible with previous versions. It should not introduce a breaking change.


zamarrenolm:
Added code to verify that CGMES conversion using old and new (refactored) code give the same IIDM Network. Some unit tests fail, they must be reviewed.

zamarrenolm:
Minor adjustments in current and refactored conversion have been made to pass added verification that IIDM networks obtained should be the same.

PR #658, opened 2019-02-28
Tests independent of local TableFormatterConfig

MioRtia:
#650 
Breaking change: validation methods are not static anymore but must be used on the final static instance `INSTANCE` of each validation class.

PR #952, opened 2019-09-20
Revert "Tests independent of local TableFormatterConfig (#658)"

jonenst:
This reverts commit a5999c533a54294c741e4cb9611530e443109ba0.

Now that we can use a test config during tests, we don't have to pass PlatformConfig
everywhere and pollute the API

NO

Refactoring

Apis have been polluted to make the tests independant of local configurations

Apis are clean again

NO


jonenst:
Ok reverted the static vs INSTANCE changes, but I kept the ValidationWriterFactory revert of adding a method with the explicit TableFormatterConfig, so that's still a breaking change.

jonenst:
I just realized that changing a method from instance to static is source compatible but not binary compatible, but binary compatibility is still important I think

jonenst:
I have restored what I deemed important. I still left out:
- the LoadFlowActionSimulatorTest artificial test to improved coverage which was only asserting same on constructor args and getters
- the new method of the ValidationWriterFactory that takes a tableformatterconfig. That's a breaking change, but are other people implementing this interface and/or using the new method ?

To me this is ready for merge

jonenst:
Note: the coveralls decrease is expected, I removed artificial tests for LoadFlowActionSimulator.java that were just asserting that the constructor was storing the inputs

Issue #900, opened 2019-08-23
Change of Shunt Compensator IIDM model


MioRtia:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
Only one type of shunt compensator is represented. Some are not, e.g. non linear shunt compensators. Furthermore, regulating controls are not taken in account with current modeling.

* **What is the expected behavior?**
To be able to model different sort of shunt compensators and to add regulating control

* **What is the motivation / use case for changing the behavior?**
CGMES models shunt compensators with different behaviours and regulating controls.

* **Please tell us about your environment:**
  - PowSyBl Version: 3.0.0-SNAPSHOT

geofjamg:
I would go for a solution where model is dissociated to the equipment (which will be responsible mainly for connectivity)

```java
ShuntCompensator sc = ...
ShuntCompensatorLinearModel model = sc.getModel(ShuntCompensatorLinearModel.class);
double b = model.getbPerSection();
```

PR #958, opened 2019-09-30
Add regulation in shunt (IIDM/XIIDM)

MioRtia:

#900 : add regulation for shunts

**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**
New features:
- regulation of shunt is added (cf. http://www.powsybl.org/docs/iidm/model/shuntCompensator.html)
- for homogenization purposes, `RatioTapChanger.targetDeadband` and `PhaseTapChanger.targetDeadband` are now mandatory when the tap changer is regulating (same behavior than shunts)
- XIIDM serialization/deserialization of these evolutions
- implement new utility methods in `IidmXmlUtil`
- implement new utility test methods in `AbstractXmlConverterTest`

- [x] The migration guide has been updated in the github wiki *(What changes might users need to make in their application due to this PR?)

### DIFF
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
@@ -83,4 +83,80 @@ public interface ShuntCompensator extends Injection<ShuntCompensator> {
      */
     double getCurrentB();
 
+    /**
+     * Get the terminal used for regulation.
+     */
+    default Terminal getRegulatingTerminal() {
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * Set the terminal used for regulation.
+     * If null is passed as regulating terminal, the regulation is considered local.
+     */
+    default ShuntCompensator setRegulatingTerminal(Terminal regulatingTerminal) {
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * Get the shunt compensator's regulating status.
+     * <p>
+     * Depends on the working variant.
+     * @see VariantManager
+     */
+    default boolean isVoltageRegulatorOn() {
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * Set the shunt compensator's regulating status.
+     * <p>
+     * Depends on the working variant.
+     * @see VariantManager
+     */
+    default ShuntCompensator setVoltageRegulatorOn(boolean voltageRegulatorOn) {
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * Get the shunt compensator's voltage target in kV if it exists. Else return NaN.
+     * <p>
+     * Depends on the working variant.
+     * @see VariantManager
+     */
+    default double getTargetV() {
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * Set the shunt compensator's voltage target in kV.
+     * <p>
+     * Depends on the working variant.
+     * @see VariantManager
+     */
+    default ShuntCompensator setTargetV(double targetV) {
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * Get the shunt compensator's deadband (in kV) used to avoid excessive update of discrete control while regulating.
+     * This attribute is necessary only if the shunt compensator is regulating.
+     * <p>
+     * Depends on the working variant.
+     * @see VariantManager
+     */
+    default double getTargetDeadband() {
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * Set the shunt compensator's deadband (in kV) used to avoid excessive update of discrete control while regulating.
+     * This attribute is necessary only if the shunt compensator is regulating. It must be positive.
+     * <p>
+     * Depends on the working variant.
+     * @see VariantManager
+     */
+    default ShuntCompensator setTargetDeadband(double targetDeadband) {
+        throw new UnsupportedOperationException();
+    }
 }
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorAdder.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorAdder.java
@@ -32,6 +32,22 @@ public interface ShuntCompensatorAdder extends InjectionAdder<ShuntCompensatorAd
 
     ShuntCompensatorAdder setCurrentSectionCount(int currentSectionCount);
 
+    default ShuntCompensatorAdder setRegulatingTerminal(Terminal regulatingTerminal) {
+        throw new UnsupportedOperationException();
+    }
+
+    default ShuntCompensatorAdder setVoltageRegulatorOn(boolean voltageRegulatorOn) {
+        throw new UnsupportedOperationException();
+    }
+
+    default ShuntCompensatorAdder setTargetV(double targetV) {
+        throw new UnsupportedOperationException();
+    }
+
+    default ShuntCompensatorAdder setTargetDeadband(double targetDeadband) {
+        throw new UnsupportedOperationException();
+    }
+
     ShuntCompensator add();
 
 }
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChanger.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChanger.java
@@ -9,11 +9,13 @@ package com.powsybl.iidm.network.impl;
 import com.powsybl.commons.util.trove.TBooleanArrayList;
 import com.powsybl.iidm.network.Terminal;
 import com.powsybl.iidm.network.ValidationException;
+import com.powsybl.iidm.network.ValidationUtil;
 import com.powsybl.iidm.network.impl.util.Ref;
 import gnu.trove.list.array.TDoubleArrayList;
 import gnu.trove.list.array.TIntArrayList;
 
 import java.util.List;
+import java.util.Objects;
 
 /**
  *
@@ -29,6 +31,8 @@ abstract class AbstractTapChanger<H extends TapChangerParent, C extends Abstract
 
     protected final List<S> steps;
 
+    private final String type;
+
     protected TerminalExt regulationTerminal;
 
     // attributes depending on the variant
@@ -41,12 +45,12 @@ abstract class AbstractTapChanger<H extends TapChangerParent, C extends Abstract
 
     protected AbstractTapChanger(Ref<? extends VariantManagerHolder> network, H parent,
                                  int lowTapPosition, List<S> steps, TerminalExt regulationTerminal,
-                                 int tapPosition, boolean regulating, double targetDeadband) {
+                                 int tapPosition, boolean regulating, double targetDeadband, String type) {
         this.network = network;
         this.parent = parent;
         this.lowTapPosition = lowTapPosition;
         this.steps = steps;
-        steps.stream().forEach(s -> s.setParent(this));
+        steps.forEach(s -> s.setParent(this));
         this.regulationTerminal = regulationTerminal;
         int variantArraySize = network.get().getVariantManager().getVariantArraySize();
         this.tapPosition = new TIntArrayList(variantArraySize);
@@ -57,6 +61,7 @@ abstract class AbstractTapChanger<H extends TapChangerParent, C extends Abstract
             this.regulating.add(regulating);
             this.targetDeadband.add(targetDeadband);
         }
+        this.type = Objects.requireNonNull(type);
     }
 
     protected abstract NetworkImpl getNetwork();
@@ -146,10 +151,8 @@ abstract class AbstractTapChanger<H extends TapChangerParent, C extends Abstract
     }
 
     public C setTargetDeadband(double targetDeadband) {
-        if (!Double.isNaN(targetDeadband) && targetDeadband < 0) {
-            throw new ValidationException(parent, "Unexpected value for target deadband of phase tap changer: " + targetDeadband);
-        }
         int variantIndex = network.get().getVariantIndex();
+        ValidationUtil.checkTargetDeadband(parent, type, this.regulating.get(variantIndex), targetDeadband);
         double oldValue = this.targetDeadband.set(variantIndex, targetDeadband);
         String variantId = network.get().getVariantManager().getVariantId(variantIndex);
         parent.getNetwork().getListeners().notifyUpdate(parent.getTransformer(), () -> getTapChangerAttribute() + ".targetDeadband", variantId, oldValue, targetDeadband);
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/GeneratorImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/GeneratorImpl.java
@@ -140,7 +140,7 @@ class GeneratorImpl extends AbstractConnectable<Generator> implements Generator,
 
     @Override
     public GeneratorImpl setRegulatingTerminal(Terminal regulatingTerminal) {
-        ValidationUtil.checkRegulatingTerminal(this, (TerminalExt) regulatingTerminal, getNetwork());
+        ValidationUtil.checkRegulatingTerminal(this, regulatingTerminal, getNetwork());
         Terminal oldValue = this.regulatingTerminal;
         this.regulatingTerminal = regulatingTerminal != null ? (TerminalExt) regulatingTerminal : getTerminal();
         notifyUpdate("regulatingTerminal", oldValue, regulatingTerminal);
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/RatioTapChangerAdderImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/RatioTapChangerAdderImpl.java
@@ -183,9 +183,7 @@ class RatioTapChangerAdderImpl implements RatioTapChangerAdder {
                     + highTapPosition + "]");
         }
         ValidationUtil.checkRatioTapChangerRegulation(parent, regulating, regulationTerminal, targetV, getNetwork());
-        if (!Double.isNaN(targetDeadband) && targetDeadband < 0) {
-            throw new ValidationException(parent, "Unexpected value for target deadband of ratio tap changer: " + targetDeadband);
-        }
+        ValidationUtil.checkTargetDeadband(parent, "ratio tap changer", regulating, targetDeadband);
         RatioTapChangerImpl tapChanger
                 = new RatioTapChangerImpl(parent, lowTapPosition, steps, regulationTerminal, loadTapChangingCapabilities,
                                           tapPosition, regulating, targetV, targetDeadband);
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/RatioTapChangerImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/RatioTapChangerImpl.java
@@ -32,7 +32,7 @@ class RatioTapChangerImpl extends AbstractTapChanger<RatioTapChangerParent, Rati
     RatioTapChangerImpl(RatioTapChangerParent parent, int lowTapPosition,
                         List<RatioTapChangerStepImpl> steps, TerminalExt regulationTerminal, boolean loadTapChangingCapabilities,
                         int tapPosition, boolean regulating, double targetV, double targetDeadband) {
-        super(parent.getNetwork().getRef(), parent, lowTapPosition, steps, regulationTerminal, tapPosition, regulating, targetDeadband);
+        super(parent.getNetwork().getRef(), parent, lowTapPosition, steps, regulationTerminal, tapPosition, regulating, targetDeadband, "ratio tap changer");
         this.loadTapChangingCapabilities = loadTapChangingCapabilities;
         int variantArraySize = network.get().getVariantManager().getVariantArraySize();
         this.targetV = new TDoubleArrayList(variantArraySize);
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
@@ -8,6 +8,7 @@ package com.powsybl.iidm.network.impl;
 
 import com.powsybl.iidm.network.ShuntCompensatorAdder;
 import com.powsybl.iidm.network.ValidationUtil;
+import com.powsybl.iidm.network.Terminal;
 
 /**
  *
@@ -22,6 +23,14 @@ class ShuntCompensatorAdderImpl extends AbstractInjectionAdder<ShuntCompensatorA
     private int maximumSectionCount;
 
     private int currentSectionCount;
+
+    private double targetV = Double.NaN;
+
+    private double targetDeadband = Double.NaN;
+
+    private TerminalExt regulatingTerminal;
+
+    private boolean voltageRegulatorOn = false;
 
     ShuntCompensatorAdderImpl(VoltageLevelExt voltageLevel) {
         this.voltageLevel = voltageLevel;
@@ -56,15 +65,43 @@ class ShuntCompensatorAdderImpl extends AbstractInjectionAdder<ShuntCompensatorA
     }
 
     @Override
+    public ShuntCompensatorAdder setRegulatingTerminal(Terminal regulatingTerminal) {
+        this.regulatingTerminal = (TerminalExt) regulatingTerminal;
+        return this;
+    }
+
+    @Override
+    public ShuntCompensatorAdder setVoltageRegulatorOn(boolean voltageRegulatorOn) {
+        this.voltageRegulatorOn = voltageRegulatorOn;
+        return this;
+    }
+
+    @Override
+    public ShuntCompensatorAdder setTargetV(double targetV) {
+        this.targetV = targetV;
+        return this;
+    }
+
+    @Override
+    public ShuntCompensatorAdder setTargetDeadband(double targetDeadband) {
+        this.targetDeadband = targetDeadband;
+        return this;
+    }
+
+    @Override
     public ShuntCompensatorImpl add() {
         String id = checkAndGetUniqueId();
         TerminalExt terminal = checkAndGetTerminal();
         ValidationUtil.checkbPerSection(this, bPerSection);
         ValidationUtil.checkSections(this, currentSectionCount, maximumSectionCount);
+        ValidationUtil.checkRegulatingTerminal(this, regulatingTerminal, getNetwork());
+        ValidationUtil.checkVoltageControl(this, voltageRegulatorOn, targetV);
+        ValidationUtil.checkTargetDeadband(this, "shunt compensator", voltageRegulatorOn, targetDeadband);
         ShuntCompensatorImpl shunt
                 = new ShuntCompensatorImpl(getNetwork().getRef(),
-                                           id, getName(), bPerSection, maximumSectionCount,
-                                           currentSectionCount);
+                id, getName(), bPerSection, maximumSectionCount,
+                currentSectionCount, regulatingTerminal == null ? terminal : regulatingTerminal,
+                voltageRegulatorOn, targetV, targetDeadband);
         shunt.addTerminal(terminal);
         voltageLevel.attach(terminal, false);
         getNetwork().getIndex().checkAndAdd(shunt);
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorImpl.java
@@ -6,10 +6,13 @@
  */
 package com.powsybl.iidm.network.impl;
 
+import com.powsybl.commons.util.trove.TBooleanArrayList;
 import com.powsybl.iidm.network.ConnectableType;
 import com.powsybl.iidm.network.ShuntCompensator;
 import com.powsybl.iidm.network.ValidationUtil;
+import com.powsybl.iidm.network.Terminal;
 import com.powsybl.iidm.network.impl.util.Ref;
+import gnu.trove.list.array.TDoubleArrayList;
 import gnu.trove.list.array.TIntArrayList;
 
 /**
@@ -26,22 +29,42 @@ class ShuntCompensatorImpl extends AbstractConnectable<ShuntCompensator> impleme
     /* the maximum number of section */
     private int maximumSectionCount;
 
+    /* the regulating terminal */
+    private TerminalExt regulatingTerminal;
+
     // attributes depending on the variant
 
     /* the current number of section switched on */
     private final TIntArrayList currentSectionCount;
 
+    /* the regulating status */
+    private final TBooleanArrayList voltageRegulatorOn;
+
+    /* the target voltage value */
+    private final TDoubleArrayList targetV;
+
+    /* the target deadband */
+    private final TDoubleArrayList targetDeadband;
+
     ShuntCompensatorImpl(Ref<? extends VariantManagerHolder> network,
                          String id, String name, double bPerSection, int maximumSectionCount,
-                         int currentSectionCount) {
+                         int currentSectionCount, TerminalExt regulatingTerminal, boolean voltageRegulatorOn,
+                         double targetV, double targetDeadband) {
         super(id, name);
         this.network = network;
         this.bPerSection = bPerSection;
         this.maximumSectionCount = maximumSectionCount;
+        this.regulatingTerminal = regulatingTerminal;
         int variantArraySize = network.get().getVariantManager().getVariantArraySize();
         this.currentSectionCount = new TIntArrayList(variantArraySize);
+        this.voltageRegulatorOn = new TBooleanArrayList(variantArraySize);
+        this.targetV = new TDoubleArrayList(variantArraySize);
+        this.targetDeadband = new TDoubleArrayList(variantArraySize);
         for (int i = 0; i < variantArraySize; i++) {
             this.currentSectionCount.add(currentSectionCount);
+            this.voltageRegulatorOn.add(voltageRegulatorOn);
+            this.targetV.add(targetV);
+            this.targetDeadband.add(targetDeadband);
         }
     }
 
@@ -104,11 +127,76 @@ class ShuntCompensatorImpl extends AbstractConnectable<ShuntCompensator> impleme
     }
 
     @Override
+    public TerminalExt getRegulatingTerminal() {
+        return regulatingTerminal;
+    }
+
+    @Override
+    public ShuntCompensatorImpl setRegulatingTerminal(Terminal regulatingTerminal) {
+        ValidationUtil.checkRegulatingTerminal(this, regulatingTerminal, getNetwork());
+        Terminal oldValue = this.regulatingTerminal;
+        this.regulatingTerminal = regulatingTerminal != null ? (TerminalExt) regulatingTerminal : getTerminal();
+        notifyUpdate("regulatingTerminal", oldValue, this.regulatingTerminal);
+        return this;
+    }
+
+    @Override
+    public boolean isVoltageRegulatorOn() {
+        return voltageRegulatorOn.get(network.get().getVariantIndex());
+    }
+
+    @Override
+    public ShuntCompensatorImpl setVoltageRegulatorOn(boolean voltageRegulatorOn) {
+        int variantIndex = network.get().getVariantIndex();
+        ValidationUtil.checkVoltageControl(this, voltageRegulatorOn, targetV.get(variantIndex));
+        boolean oldValue = this.voltageRegulatorOn.set(variantIndex, voltageRegulatorOn);
+        String variantId = network.get().getVariantManager().getVariantId(variantIndex);
+        notifyUpdate("voltageRegulatorOn", variantId, oldValue, voltageRegulatorOn);
+        return this;
+    }
+
+    @Override
+    public double getTargetV() {
+        return targetV.get(network.get().getVariantIndex());
+    }
+
+    @Override
+    public ShuntCompensatorImpl setTargetV(double targetV) {
+        int variantIndex = network.get().getVariantIndex();
+        ValidationUtil.checkVoltageControl(this, voltageRegulatorOn.get(variantIndex), targetV);
+        double oldValue = this.targetV.set(variantIndex, targetV);
+        String variantId = network.get().getVariantManager().getVariantId(variantIndex);
+        notifyUpdate("targetV", variantId, oldValue, targetV);
+        return this;
+    }
+
+    @Override
+    public double getTargetDeadband() {
+        return targetDeadband.get(network.get().getVariantIndex());
+    }
+
+    @Override
+    public ShuntCompensatorImpl setTargetDeadband(double targetDeadband) {
+        int variantIndex = network.get().getVariantIndex();
+        ValidationUtil.checkTargetDeadband(this, "shunt compensator", this.voltageRegulatorOn.get(variantIndex), targetDeadband);
+        double oldValue = this.targetDeadband.set(variantIndex, targetDeadband);
+        String variantId = network.get().getVariantManager().getVariantId(variantIndex);
+        notifyUpdate("targetDeadband", variantId, oldValue, targetDeadband);
+        return this;
+    }
+
+    @Override
     public void extendVariantArraySize(int initVariantArraySize, int number, int sourceIndex) {
         super.extendVariantArraySize(initVariantArraySize, number, sourceIndex);
         currentSectionCount.ensureCapacity(currentSectionCount.size() + number);
+        voltageRegulatorOn.ensureCapacity(voltageRegulatorOn.size() + number);
+        targetV.ensureCapacity(targetV.size() + number);
+        targetDeadband.ensureCapacity(targetDeadband.size() + number);
         for (int i = 0; i < number; i++) {
             currentSectionCount.add(currentSectionCount.get(sourceIndex));
+            voltageRegulatorOn.add(voltageRegulatorOn.get(sourceIndex));
+            targetV.add(targetV.get(sourceIndex));
+            targetDeadband.add(targetDeadband.get(sourceIndex));
         }
     }
 
@@ -116,6 +204,9 @@ class ShuntCompensatorImpl extends AbstractConnectable<ShuntCompensator> impleme
     public void reduceVariantArraySize(int number) {
         super.reduceVariantArraySize(number);
         currentSectionCount.remove(currentSectionCount.size() - number, number);
+        voltageRegulatorOn.remove(voltageRegulatorOn.size() - number, number);
+        targetV.remove(targetV.size() - number, number);
+        targetDeadband.remove(targetDeadband.size() - number, number);
     }
 
     @Override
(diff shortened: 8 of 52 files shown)


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChanger.java
```diff
@@ -145,13 +145,12 @@ public double getTargetDeadband() {
     }
 
     public C setTargetDeadband(double targetDeadband) {
-        if (!Double.isNaN(targetDeadband) && targetDeadband < 0) {
-            throw new ValidationException(parent, "Unexpected value for target deadband of phase tap changer: " + targetDeadband);
-        }
+        double realTargetDeadband = Double.isNaN(targetDeadband) ? 0 : targetDeadband;
```
mathbagu: As the target deadband is required only if the shunt is regulating (regarding to the javadoc you wrote), why don't you accept NaN? Maybe the `checkTargetDeadband` method should take two parameters?

If you set the targetDeadband to NaN and call the getTargetDeadband then, you will have 0 ? It's strange to me...

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/RatioTapChangerAdderImpl.java
```diff
@@ -172,12 +172,10 @@ public RatioTapChanger add() {
...
         }
         ValidationUtil.checkRatioTapChangerRegulation(parent, loadTapChangingCapabilities, regulating, regulationTerminal, targetV, getNetwork());
-        if (!Double.isNaN(targetDeadband) && targetDeadband < 0) {
-            throw new ValidationException(parent, "Unexpected value for target deadband of ratio tap changer: " + targetDeadband);
-        }
+        ValidationUtil.checkTargetDeadband(parent, targetDeadband);
         RatioTapChangerImpl tapChanger
                 = new RatioTapChangerImpl(parent, lowTapPosition, steps, regulationTerminal, loadTapChangingCapabilities,
-                                          tapPosition, regulating, targetV, targetDeadband);
+                                          tapPosition, regulating, targetV, Double.isNaN(targetDeadband) ? 0 : targetDeadband);
```
mathbagu: To be fixed depending on what we decide about NaN values for targetDeadband.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
```diff
@@ -22,6 +23,14 @@
 
     private int currentSectionCount;
 
+    private double targetV = Double.NaN;
+
+    private double targetDeadband = 0;
```
mathbagu: To be fixed depending on what we decide about NaN values for targetDeadband.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
```diff
@@ -78,4 +78,78 @@
...
+     * Depends on the working variant.
+     * @see VariantManager
+     */
+    default ShuntCompensator setTargetV(double targetV) {
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * Get the shunt compensator's deadband (in kV) used to avoid excessive update of discrete control while regulating if it exists.
+     * Else returns NaN.
```
mathbagu: To be fixed depending on what we decide about NaN values for targetDeadband.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
```diff
@@ -54,16 +63,44 @@ public ShuntCompensatorAdder setCurrentSectionCount(int currentSectionCount) {
...
+        ValidationUtil.checkRegulatingTerminal(this, regulatingTerminal, getNetwork());
+        ValidationUtil.checkVoltageControl(this, regulating, targetV);
+        ValidationUtil.checkTargetDeadband(this, targetDeadband);
         ShuntCompensatorImpl shunt
                 = new ShuntCompensatorImpl(getNetwork().getRef(),
-                                           id, getName(), bPerSection, maximumSectionCount,
-                                           currentSectionCount);
+                id, getName(), bPerSection, maximumSectionCount,
+                currentSectionCount, regulatingTerminal == null ? terminal : regulatingTerminal,
+                regulating, targetV, Double.isNaN(targetDeadband) ? 0 : targetDeadband);
```
mathbagu: To be fixed depending on what we decide about NaN values for targetDeadband.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorImpl.java
```diff
@@ -107,19 +130,88 @@ public double getMaximumB() {
...
     }
 
+    @Override
+    public TerminalExt getRegulatingTerminal() {
+        return regulatingTerminal;
+    }
+
+    @Override
+    public ShuntCompensatorImpl setRegulatingTerminal(Terminal regulatingTerminal) {
+        ValidationUtil.checkRegulatingTerminal(this, (TerminalExt) regulatingTerminal, getNetwork());
```
mathbagu: Are you sure about this control: should we get a valid terminal:
```
this.regulatingTerminal = regulatingTerminal != null ? (TerminalExt) regulatingTerminal : getTerminal();
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorImpl.java
```diff
@@ -107,19 +130,88 @@ public double getMaximumB() {
...
+        ValidationUtil.checkVoltageControl(this, regulating.get(variantIndex), targetV);
+        double oldValue = this.targetV.set(variantIndex, targetV);
+        String variantId = network.get().getVariantManager().getVariantId(variantIndex);
+        notifyUpdate("targetV", variantId, oldValue, targetV);
+        return this;
+    }
+
+    @Override
+    public double getTargetDeadband() {
+        return targetDeadband.get(network.get().getVariantIndex());
```
mathbagu: Regarding to the javadoc, it should return NaN sometimes?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ValidationUtil.java
```diff
@@ -49,18 +49,28 @@ static void checkActivePowerLimits(Validable validable, double minP, double maxP
...
+            throw new ValidationException(validable, "Unexpected value for target deadband: " + targetDeadband);
+        }
+    }
+
+    static boolean checkVoltageControl(Validable validable, Boolean voltageRegulatorOn, double voltageSetpoint) {
         if (voltageRegulatorOn == null) {
             throw new ValidationException(validable, "voltage regulator status is not set");
         }
-        if (voltageRegulatorOn) {
+        if (Boolean.TRUE.equals(voltageRegulatorOn)) {
```
mathbagu: Why? the previous version was ok

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
```diff
@@ -30,18 +30,28 @@ protected String getRootElementName() {
...
+        XmlUtil.writeOptionalBoolean("regulating", sc.isRegulating(), false, context.getWriter());
+        XmlUtil.writeDouble("targetV", sc.getTargetV(), context.getWriter());
+        XmlUtil.writeOptionalDouble("targetDeadband", sc.getTargetDeadband(), 0, context.getWriter());
         writeNodeOrBus(null, sc.getTerminal(), context);
         writePQ(null, sc.getTerminal(), context.getWriter());
     }
 
+    @Override
+    protected void writeSubElements(ShuntCompensator sc, VoltageLevel vl, NetworkXmlWriterContext context) throws XMLStreamException {
+        if (sc.getRegulatingTerminal().getConnectable() != sc) {
```
mathbagu: Maybe `sc.getRegulatingTerminal() != sc.getTerminal()` is more understandable?

iidm/iidm-xml-converter/src/main/resources/xsd/iidm.xsd
```diff
@@ -260,9 +260,15 @@
...
             <xs:extension base="iidm:Injection">
+                <xs:sequence>
+                    <xs:element name="regulatingTerminal" minOccurs="0" type="iidm:TerminalRef"/>
+                </xs:sequence>
                 <xs:attribute name="bPerSection" use="required" type="xs:double"/>
                 <xs:attribute name="maximumSectionCount" use="required" type="xs:int"/>
                 <xs:attribute name="currentSectionCount" use="required" type="xs:int"/>
+                <xs:attribute name="regulating" use="optional" type="xs:boolean"/>
+                <xs:attribute name="targetV" use="optional" type="xs:double"/>
+                <xs:attribute name="targetDeadband" use="optional" type="xs:double"/>
```
mathbagu: Are all of these attributes really optional or do you set them optional for backward compatibility?

iidm/iidm-xml-converter/src/test/java/com/powsybl/iidm/xml/ShuntCompensatorXmlTest.java
```diff
@@ -0,0 +1,29 @@
+/**
+ * Copyright (c) 2019, All partners of the iTesla project (http://www.itesla-project.eu/consortium)
```
mathbagu: Change this copyright

iidm/iidm-test/src/main/java/com/powsybl/iidm/network/test/ShuntTestCaseFactory.java
```diff
@@ -0,0 +1,81 @@
+/**
+ * Copyright (c) 2019, All partners of the iTesla project (http://www.itesla-project.eu/consortium)
```
mathbagu: Change this copyright

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ValidationUtil.java
MioRtia: Yes, not really sure why I changed it... Sonar considered it as "risky" since it can return a `NullPointerException` but the `null` case is handled above. I will reverse it.

iidm/iidm-xml-converter/src/main/resources/xsd/iidm.xsd
MioRtia: `targetV` and `targetDeadband` are really optional (they can be undefined if `regulating` is `false`). for `regulating`, I can put it as required but it means that all previous files would not be readable.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorImpl.java
MioRtia: Not sure, the check is taking place if the terminal is not null anyway.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
```diff
@@ -52,9 +62,15 @@ protected ShuntCompensator readRootElementAttributes(ShuntCompensatorAdder adder
         double bPerSection = XmlUtil.readDoubleAttribute(context.getReader(), "bPerSection");
         int maximumSectionCount = XmlUtil.readIntAttribute(context.getReader(), "maximumSectionCount");
         int currentSectionCount = XmlUtil.readIntAttribute(context.getReader(), "currentSectionCount");
+        boolean regulating = XmlUtil.readBoolAttribute(context.getReader(), "regulating");
```
yichen88: In those existing xml, this filed is not exist. 
Would it throw exception while using new code to read an old xml?

iidm/iidm-xml-converter/src/test/resources/LccRoundTripRef.xml
```diff
@@ -22,8 +22,8 @@
                 <iidm:switch id="DISC_BBS1_BK3" name="Disconnector" kind="DISCONNECTOR" retained="false" open="false" node1="0" node2="5"/>
                 <iidm:switch id="BK3" name="Breaker" kind="BREAKER" retained="true" open="false" node1="5" node2="6"/>
             </iidm:nodeBreakerTopology>
-            <iidm:shunt id="C2_Filter1" name="Filter 3"  bPerSection="3.0E-5" maximumSectionCount="1" currentSectionCount="1" node="4" q="12.5"/>
-            <iidm:shunt id="C2_Filter2" name="Filter 4" bPerSection="4.0E-5" maximumSectionCount="1" currentSectionCount="1" node="6" q="12.5"/>
+            <iidm:shunt id="C2_Filter1" name="Filter 3"  bPerSection="3.0E-5" maximumSectionCount="1" currentSectionCount="1" regulating="false" node="4" q="12.5"/>
```
yichen88: double spaces before `bPerSection`

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
MioRtia: Yes, this is why this PR must wait the merge of the evolution for versionning XIIDM, this will be in IIDM 1.1

yichen88: FYI, there is a branch about v1.1
https://github.com/powsybl/powsybl-core/tree/my_iidms

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChanger.java
```diff
@@ -54,8 +56,9 @@ protected AbstractTapChanger(Ref<? extends VariantManagerHolder> network, H pare
         for (int i = 0; i < variantArraySize; i++) {
             this.tapPosition.add(tapPosition);
             this.regulating.add(regulating);
-            this.targetDeadband.add(targetDeadband);
+            this.targetDeadband.add(Double.isNaN(targetDeadband) ? 0 : targetDeadband);
         }
+        this.type = type;
```
mathbagu: Could it be null?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChanger.java
```diff
@@ -54,8 +56,9 @@ protected AbstractTapChanger(Ref<? extends VariantManagerHolder> network, H pare
         for (int i = 0; i < variantArraySize; i++) {
             this.tapPosition.add(tapPosition);
             this.regulating.add(regulating);
-            this.targetDeadband.add(targetDeadband);
+            this.targetDeadband.add(Double.isNaN(targetDeadband) ? 0 : targetDeadband);
```
mathbagu: Could probably be cached in a local variable.

Looking to the code below, it seems there is a `checkDeadband` method that throws when value is NaN

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/RatioTapChangerImpl.java
```diff
@@ -31,7 +31,7 @@
     RatioTapChangerImpl(RatioTapChangerParent parent, int lowTapPosition,
                         List<RatioTapChangerStepImpl> steps, TerminalExt regulationTerminal, boolean loadTapChangingCapabilities,
                         int tapPosition, boolean regulating, double targetV, double targetDeadband) {
-        super(parent.getNetwork().getRef(), parent, lowTapPosition, steps, regulationTerminal, tapPosition, regulating, targetDeadband);
+        super(parent.getNetwork().getRef(), parent, lowTapPosition, steps, regulationTerminal, tapPosition, regulating, targetDeadband, "phase tap changer");
```
mathbagu: Typo: "ratio tap changer"?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractTransformerXml.java
```diff
@@ -46,7 +46,7 @@ protected static void writeTapChangerStep(TapChangerStep<?> tcs, XMLStreamWriter
     protected static void writeTapChanger(TapChanger<?, ?> tc, XMLStreamWriter writer) throws XMLStreamException {
         writer.writeAttribute(ATTR_LOW_TAP_POSITION, Integer.toString(tc.getLowTapPosition()));
         writer.writeAttribute(ATTR_TAP_POSITION, Integer.toString(tc.getTapPosition()));
-        XmlUtil.writeDouble(TARGET_DEADBAND, tc.getTargetDeadband(), writer);
+        XmlUtil.writeOptionalDouble(TARGET_DEADBAND, tc.getTargetDeadband(), 0, writer);
```
mathbagu: I think you have to change this since I merged #1054 ?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractTransformerXml.java
```diff
@@ -73,7 +73,7 @@ protected static void writeRatioTapChanger(String name, RatioTapChanger rtc, Net
     protected static void readRatioTapChanger(String elementName, RatioTapChangerAdder adder, Terminal terminal, NetworkXmlReaderContext context) throws XMLStreamException {
         int lowTapPosition = XmlUtil.readIntAttribute(context.getReader(), ATTR_LOW_TAP_POSITION);
         int tapPosition = XmlUtil.readIntAttribute(context.getReader(), ATTR_TAP_POSITION);
-        double targetDeadband = XmlUtil.readOptionalDoubleAttribute(context.getReader(), TARGET_DEADBAND);
+        double targetDeadband = XmlUtil.readOptionalDoubleAttribute(context.getReader(), TARGET_DEADBAND, 0);
```
mathbagu: Do you have to check the version?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractTransformerXml.java
```diff
@@ -147,7 +147,7 @@ protected static void writePhaseTapChanger(String name, PhaseTapChanger ptc, Net
     protected static void readPhaseTapChanger(String name, PhaseTapChangerAdder adder, Terminal terminal, NetworkXmlReaderContext context) throws XMLStreamException {
         int lowTapPosition = XmlUtil.readIntAttribute(context.getReader(), ATTR_LOW_TAP_POSITION);
         int tapPosition = XmlUtil.readIntAttribute(context.getReader(), ATTR_TAP_POSITION);
-        double targetDeadband = XmlUtil.readOptionalDoubleAttribute(context.getReader(), TARGET_DEADBAND);
+        double targetDeadband = XmlUtil.readOptionalDoubleAttribute(context.getReader(), TARGET_DEADBAND, 0);
```
mathbagu: Same remark here

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
```diff
@@ -30,18 +32,28 @@ protected String getRootElementName() {
 
     @Override
     protected boolean hasSubElements(ShuntCompensator sc) {
-        return false;
+        return !Objects.equals(sc, sc.getRegulatingTerminal().getConnectable());
```
mathbagu: We already had this discussion: why don't use simple address comparison?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
```diff
@@ -30,18 +32,28 @@ protected String getRootElementName() {
...
     }
 
     @Override
     protected void writeRootElementAttributes(ShuntCompensator sc, VoltageLevel vl, NetworkXmlWriterContext context) throws XMLStreamException {
         XmlUtil.writeDouble("bPerSection", sc.getbPerSection(), context.getWriter());
         context.getWriter().writeAttribute("maximumSectionCount", Integer.toString(sc.getMaximumSectionCount()));
         context.getWriter().writeAttribute("currentSectionCount", Integer.toString(sc.getCurrentSectionCount()));
+        context.getWriter().writeAttribute("regulating", Boolean.toString(sc.isRegulating()));
+        XmlUtil.writeDouble("targetV", sc.getTargetV(), context.getWriter());
+        XmlUtil.writeOptionalDouble("targetDeadband", sc.getTargetDeadband(), 0, context.getWriter());
```
mathbagu: Do we have to check the version

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractTransformerXml.java
MioRtia: Don't think so, using `writeDouble` instead of `writeOptionalDouble` was a mistake I think...

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
```diff
@@ -83,4 +83,77 @@ default double getMaximumB() {
...
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * Get the shunt compensator's regulating status.
+     * <p>
+     * Depends on the working variant.
+     * @see VariantManager
+     */
+    default boolean isRegulating() {
```
annetill: Just a question: regulations in IIDM have a different modeling depending on the object. Here you have chosen ````Regulating```` whereas for generators we use ````VoltageRegulatorOn````. The regulating terminal is needed of course, but you did not put it in the documentation page (we have it for generators). Why ? 
Generators: https://www.powsybl.org/docs/iidm/model/generator.html
Shunts: https://github.com/powsybl/powsybl.github.io/pull/114/files

MioRtia: > Just a question: regulations in IIDM have a different modeling depending on the object. Here you have chosen Regulating whereas for generators we use VoltageRegulatorOn.

Contrary to shunts which can only be regulated in voltage or off regulation, generators are always regulating and can be regulated in voltage or in reactive power so an attribute `regulating` for generators does not really make sense.

> The regulating terminal is needed of course, but you did not put it in the documentation page (we have it for generators). Why ?

I 've forgotten it, I will add it to the documentation.

annetill: I am not sure to well understand your answer. Generators can be equiped of automatic generation control and of automatic voltage regulator. The first one is not modeled in IIDM. The second one is a voltage regulation (the first one) and a generator can be in or off voltage regulation.

MioRtia: In IIDM, when a generator is off voltage regulation, it is required to be on reactive power regulation. It can not be entirely out of any regulation (in this model). This is why it does not really make sense to have a `regulating` attribute. Shunts, however, can only be on voltage regulation or off regulation altogether. 
Note that I would personally prefer having an enum of `RegulationMode` for generators rather than a `voltageRegulatorOn` attribute, it would be clearer in my opinion...

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ValidationUtil.java
```diff
@@ -52,18 +52,31 @@ static void checkActivePowerLimits(Validable validable, double minP, double maxP
...
-            }
+            return false;
+        }
+        return true;
+    }
+
+    static void checkVoltageControl(Validable validable, Boolean voltageRegulatorOn, double voltageSetpoint, double reactivePowerSetpoint) {
+        if (checkVoltageControl(validable, voltageRegulatorOn, voltageSetpoint) && Double.isNaN(reactivePowerSetpoint)) {
+            throw createInvalidValueException(validable, reactivePowerSetpoint, "reactive power setpoint", "voltage regulator is off");
         }
```
mathbagu: I'm not sure to understand why these method is split in two parts...

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractTransformerXml.java
mathbagu: Are you sure we don't want to serialize `targetDeadband` if the value is `0`? Is it really a good idea?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ValidationUtil.java
MioRtia: I am only using `checkVoltageControl(Validable, Boolean, double)` for shunts since they don't have `reactivePowerSetpoint` and it was to avoid duplication. But for two lines, I could do a specific method I guess if this is too confusing.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractTransformerXml.java
MioRtia: In 1.0, `targetDeadband` was optionally serialized if not defined (i.e. at default value which was considered during calculations as if it was equal to 0). We can make it systematically serialized if it is less confusing?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
mathbagu: @MioRtia I think I agree with @annetill. For me, it makes the API more complicated to understand if the name of the attributes/method used for the voltage regulation are different for 2 equipments.

MioRtia: I agree but it is already the case, voltage regulation for ratio tap changers is indicated by an attribute called `regulating`. However, if you think having `voltageRegulatorOn` as attribute is clearer, I will change it.

MioRtia: I still think that having a boolean attribute for generators set as false to indicate that the generator has reactive power regulation is not very user-friendly and I would rather change that. What do you think?

annetill: Yes, same comments for ratio tap changers. As I said, maybe we have to take some minutes to think about the regulations modeling in IIDM. I am talking about the voltage regulation on a local point in the network and on active power regulation (or current, we need to precise) for phase tap changers. And what about Frequency control ? And what the rest ?

MioRtia: I replaced `regulating` by `voltageRegulatorOn` which is indeed clearer about the kind of regulation that is taking place. The modeling of regulation in general is a broader question that will take place, I made an issue about it (cf #1145 )

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
```diff
@@ -54,16 +63,44 @@ public ShuntCompensatorAdder setCurrentSectionCount(int currentSectionCount) {
...
         ValidationUtil.checkSections(this, currentSectionCount, maximumSectionCount);
+        ValidationUtil.checkRegulatingTerminal(this, regulatingTerminal, getNetwork());
+        ValidationUtil.checkVoltageControl(this, voltageRegulatorOn, targetV);
+        ValidationUtil.checkTargetDeadband(this, "shunt compensator", targetDeadband);
         ShuntCompensatorImpl shunt
                 = new ShuntCompensatorImpl(getNetwork().getRef(),
-                                           id, getName(), bPerSection, maximumSectionCount,
-                                           currentSectionCount);
+                id, getName(), bPerSection, maximumSectionCount,
+                currentSectionCount, regulatingTerminal == null ? terminal : regulatingTerminal,
```
annetill: We need to document that if the regulating terminal is null, then the regulation becomes local. Is it the same assumption for SVC, generators, tap changers ?

MioRtia: It is the same for generators and SVC but not for tap changers.

mathbagu: Maybe we should fix the bahavior for tap changers to be consistent everywhere. At least we should create an issue to report that.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorImpl.java
```diff
@@ -102,19 +125,87 @@ public double getCurrentB() {
...
+        return voltageRegulatorOn.get(network.get().getVariantIndex());
+    }
+
+    @Override
+    public ShuntCompensatorImpl setVoltageRegulatorOn(boolean voltageRegulatorOn) {
+        int variantIndex = network.get().getVariantIndex();
+        ValidationUtil.checkVoltageControl(this, voltageRegulatorOn, targetV.get(variantIndex));
+        boolean oldValue = this.voltageRegulatorOn.set(variantIndex, voltageRegulatorOn);
+        String variantId = network.get().getVariantManager().getVariantId(variantIndex);
+        notifyUpdate("regulating", variantId, oldValue, voltageRegulatorOn);
```
mathbagu: Fix the name of this property: in GeneratorImpl, we use `voltageRegulatorOn`. If we consider regulating is better, change the name of the property everywhere.

To be honest, I think the property name should be the same than the name of the method / attribute. What do you think?

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/ShuntCompensatorTest.java
```diff
@@ -25,31 +26,48 @@
...
         ShuntCompensator shuntCompensator = voltageLevel
-                                        .newShuntCompensator()
-                                            .setId("shunt")
-                                            .setName("shuntName")
-                                            .setConnectableBus("busA")
-                                            .setbPerSection(5.0)
-                                            .setCurrentSectionCount(6)
-                                            .setMaximumSectionCount(10)
-                                        .add();
+                .newShuntCompensator()
```
mathbagu: Coding style: I think in the rest of the code, we would put `newShuntCompensator` on the previous line.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractTransformerXml.java
mathbagu: If I understand well, the `targetDeadband` attribute doesn't exist in fist versions of powsybl. We add this attribute recently (end of June 2019). You add this attribute as optional in the XSD to manage the compatibility.

Looking to the master, the attribute is read from the file but never write, right? The allowed values are `NaN` or a value greater than `0`. How do you distinguish that the value was NaN or 0 if you don't export it when it's equal to 0?

Another thing you can do is to convert NaN values to 0. In that case you have to fix the adder.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorImpl.java
MioRtia: Yes, I think I've forgotten to change this in the last commit.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractTransformerXml.java
MioRtia: Changed the behavior of `targetDeadband`: it is by default undefined (`Double.NaN`) and is required when the equipment is regulating. For backward compatibility, `targetDeadband` is not serialized in 1.0 when it equals `0` and **in 1.0**, if it is absent when the equipment is regulating, it is considered as equal to 0.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractTransformerXml.java
```diff
@@ -70,13 +85,19 @@ protected static void writeRatioTapChanger(String name, RatioTapChanger rtc, Net
...
         int tapPosition = XmlUtil.readIntAttribute(context.getReader(), ATTR_TAP_POSITION);
-        double targetDeadband = XmlUtil.readOptionalDoubleAttribute(context.getReader(), TARGET_DEADBAND);
+        double[] targetDeadband = new double[1];
         boolean regulating = XmlUtil.readOptionalBoolAttribute(context.getReader(), ATTR_REGULATING, false);
+        targetDeadband[0] = XmlUtil.readOptionalDoubleAttribute(context.getReader(), TARGET_DEADBAND);
+        IidmXmlUtil.runUntilMaximumVersion(IidmXmlVersion.V_1_0, context, () -> {
+            if (regulating && Double.isNaN(targetDeadband[0])) {
+                targetDeadband[0] = 0;
+            }
+        });
```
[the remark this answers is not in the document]
mathbagu: Why don't you do exactly the same thing for both tap changers?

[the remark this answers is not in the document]
MioRtia: A bit complicated but considering how `targetDeadband` has been changed, I don't think we can do better... To sum up, when reading in XIIDM 1.0:
- if `targetDeadband` is present, it is normally read
- if `targetDeadband` is absent and the tap changer is not regulating, it is considered as `NaN` (like in XIIDM 1.1)
- if `targetDeadband` is absent **but the tap changer is regulating** (forbidden case in IIDM 1.1 but allowed in IIDM 1.0), it is considered as `0`

I tried to do the same for ratio and phase tap changers alike, did I miss something?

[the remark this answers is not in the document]
MioRtia: Maybe a warning should be written in the third case.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
```diff
@@ -23,25 +23,41 @@
...
+                IidmXmlUtil.ErrorMessage.NOT_DEFAULT_NOT_SUPPORTED, IidmXmlVersion.V_1_1, context);
+        IidmXmlUtil.writeDoubleAttributeFromMinimumVersion(ROOT_ELEMENT_NAME, "targetDeadband",
+                sc.getTargetDeadband(), IidmXmlUtil.ErrorMessage.NOT_DEFAULT_NOT_SUPPORTED, IidmXmlVersion.V_1_1, context);
         writeNodeOrBus(null, sc.getTerminal(), context);
         writePQ(null, sc.getTerminal(), context.getWriter());
     }
 
+    @Override
+    protected void writeSubElements(ShuntCompensator sc, VoltageLevel vl, NetworkXmlWriterContext context) throws XMLStreamException {
+        if (sc != sc.getRegulatingTerminal().getConnectable()) {
```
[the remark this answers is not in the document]
MioRtia: In `AbstractIdentifiableXml`, we see that this method is always called even if `hasSubElements` return `false`: this method is only used to write XML start and end elements.

mathbagu:
Could you extract read/write target deadband to a private method to make the code easier to read.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChanger.java
```diff
@@ -41,12 +45,12 @@
 
     protected AbstractTapChanger(Ref<? extends VariantManagerHolder> network, H parent,
                                  int lowTapPosition, List<S> steps, TerminalExt regulationTerminal,
-                                 int tapPosition, boolean regulating, double targetDeadband) {
+                                 int tapPosition, boolean regulating, double targetDeadband, String type) {
```
agnesLeroy: Was it really necessary to add the type here? Depending on which java class you're in you basically know the type of equipment considered, or am I wrong?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChanger.java
```diff
@@ -41,12 +45,12 @@
...
         this.network = network;
         this.parent = parent;
         this.lowTapPosition = lowTapPosition;
         this.steps = steps;
-        steps.stream().forEach(s -> s.setParent(this));
+        steps.forEach(s -> s.setParent(this));
```
agnesLeroy: Why do you remove the stream() here?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/GeneratorImpl.java
```diff
@@ -140,7 +140,7 @@ public TerminalExt getRegulatingTerminal() {
 
     @Override
     public GeneratorImpl setRegulatingTerminal(Terminal regulatingTerminal) {
-        ValidationUtil.checkRegulatingTerminal(this, (TerminalExt) regulatingTerminal, getNetwork());
+        ValidationUtil.checkRegulatingTerminal(this, regulatingTerminal, getNetwork());
```
agnesLeroy: Why do you remove the (TerminalExt) cast?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorAdder.java
```diff
@@ -32,6 +32,22 @@
...
+    default ShuntCompensatorAdder setVoltageRegulatorOn(boolean voltageRegulatorOn) {
+        throw new UnsupportedOperationException();
+    }
+
+    default ShuntCompensatorAdder setTargetV(double targetV) {
+        throw new UnsupportedOperationException();
+    }
+
+    default ShuntCompensatorAdder setTargetDeadband(double targetDeadband) {
+        throw new UnsupportedOperationException();
```
agnesLeroy: Here do you throw exceptions because with older versions of IIDM the new attributes are not supported?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/util/IidmXmlUtil.java
```diff
@@ -103,26 +105,125 @@ public static double readDoubleAttributeFromMinimumVersion(String rootElementNam
...
+                                                         ErrorMessage type, IidmXmlVersion minVersion, NetworkXmlWriterContext context,
+                                                         Runnable write) {
         if (context.getVersion().compareTo(minVersion) >= 0) {
-            XmlUtil.writeDouble(attributeName, value, context.getWriter());
+            write.run();
         } else {
-            assertMinimumVersionIfNotDefault(!Objects.equals(value, defaultValue), rootElementName, attributeName, type, minVersion, context);
+            assertMinimumVersionIfNotDefault(isNotDefaultValue, rootElementName, attributeName, type, minVersion, context);
         }
     }
```
agnesLeroy: It seems to me that this piece of development is necessary to the addition of regulations on shunts, because it's the first case like this maybe, but it's actually a modification that could impact other equipments... Maybe it should have been pushed separately? Anyway now that it's here, it should probably be mentioned in the PR's main text just to keep track of it, what do you think?

iidm/iidm-xml-converter/src/test/java/com/powsybl/iidm/xml/ShuntCompensatorXmlTest.java
```diff
@@ -0,0 +1,31 @@
...
+    public void test() throws IOException {
+        Network network = ShuntTestCaseFactory.create();
+        ShuntCompensator sc = network.getShuntCompensator("SHUNT");
+        sc.setProperty("test", "test");
+        roundTripXmlTest(network,
+                NetworkXml::writeAndValidate,
+                NetworkXml::read,
+                getVersionDir(IidmXmlConstants.CURRENT_IIDM_XML_VERSION) + "/shuntRoundTripRef.xml");
+    }
+}
```
agnesLeroy: Shouldn't there be more tests here, to check an exception is thrown In case you try to import a shunt in IIDM 1_0 with regulations?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/GeneratorImpl.java
MioRtia: `checkRegulatingTerminal` has changed its signature: from `checkRegulatingTerminal(Validable, TerminalExt, Network)` to `checkRegulatingTerminal(Validable, Terminal, Network)` (cf. #1125). Hence, casting as `TerminalExt` is not necessary anymore.

agnesLeroy: Ok, thanks for the explanation!

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/util/IidmXmlUtil.java
MioRtia: Git is really weird in its displaying of differences, the differences here do not apply to a pre-existing method, it is a new one. It should not result in any behavior change (unless I made a mistake) which is why I did not mention it. Basically, I refactored `writeDoubleAttributeFromMinimumVersion` by implementing a private intermediary method that takes a boolean which will be the result of `!Objects.equals(value, defaultValue)` for double values.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorAdder.java
MioRtia: No, `default` implementation in Java interfaces are just here to prevent breaking changes in implementations of this interface. The exception is just a warning that by default, this method is not implemented.

MioRtia:
@agnesLeroy thanks for your inputs! Some answers below:

> One of them concerns the fact that your changes impact the IIDM xml Utils, and this should probably be mentioned in the PR's description : the review would have been easier if I were aware that changes had been made elsewhere than specifically in the shunts. The same goes for the tap changers, which are also impacted since the check on target dead band is now shared with the shunt compensators.

I agree, I updated the PR message accordingly. 

> I also have a question about the documentation : have you modified the website documentation about shunt compensators accordingly, so that when we release the version we can simultaneously update the website?

Yes, the associated PR for documentation is powsybl/powsybl.github.io#114

iidm/iidm-xml-converter/src/test/java/com/powsybl/iidm/xml/ShuntCompensatorXmlTest.java
MioRtia: Good point, I will add it

iidm/iidm-xml-converter/src/test/resources/V1_1/LccRoundTripRef.xml
```diff
@@ -21,7 +21,7 @@
                 <iidm:switch id="DISC_BBS1_BK3" name="Disconnector" kind="DISCONNECTOR" retained="false" open="false" node1="0" node2="5"/>
                 <iidm:switch id="BK3" name="Breaker" kind="BREAKER" retained="true" open="false" node1="5" node2="6"/>
             </iidm:nodeBreakerTopology>
-            <iidm:shunt id="C2_Filter1" name="Filter 3" bPerSection="3.0E-5" maximumSectionCount="1" currentSectionCount="1" node="4" q="12.5"/>
+            <iidm:shunt id="C2_Filter1" name="Filter 3"  bPerSection="3.0E-5" maximumSectionCount="1" currentSectionCount="1" node="4" q="12.5"/>
```
mathbagu: Strange to have a space there. If I remember well, @annetill fixed it in a previous PR

PR #976, opened 2019-10-14
Cgmes conversion t2x regulating control

marqueslanauja:

Refactoring

`RatioTapChangerAdder`: Regulating control is created with all the information available.
`PhaseTapChangeAdder`: Regulating control is created with all the information available.
`End of the cgmes conversion process`: The remote control data is added.
 The control is localized if the regulating terminal is not found in the IIDM model.
LoadTapChangingCapabilities is set to false if there is not ratio voltage control.

`RatioTapChangerAdder`: Regulating control is initialized.
`PhaseTapChangerAdder`: Regulating control is initialized.
`TwoWindingsTransformerAdder`: The relationship "IIDM transformerId - regulationControlId " is recorded in the context.
`End of the cgmes conversion process`: Control information is added for each two windings transformer in IIDM model.
 The control is never localized.
The LoadTapChangingCapabilities contains always the value of the cgmes ltcFlag attribute.
Only one regulating control enabled is allowed. If both (ratio and phase) are enabled the ratio is disabled.

**Other information**:

`Conversion`: ratioTapChanger and phaseTapChanger data has been cached. (Also used in new two and three windings conversion).
`RegulatingControlMapping`: phase regulating control methods has been moved to RegulatingControlMappingForTransformers.
`RegulatingControlMappingForGenerators`: isControlModeVoltage method has been moved to RegulatingControlMapping as it is also used in RegulatingControlMappingForTransformers.
`RegulatingControlMappingForTransformers`: New class for regulatingControl in transformers.
`PhaseTapChangerConversion`: Set regulation mode to FIXED_TAP if ltcFlag = 0 (and comments) has been moved to RegulatingControlMappingForTransformers.
`CgmesConformity1ModifiedConversionTest`: Test has been adjusted to the new behaviour of LoadTapChangingCapabilities attribute.

The check in the setRegulating method allowing only one control enabled will be included in the two windings transformer API PR.


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForTransformers.java
```diff
@@ -0,0 +1,391 @@
...
+    private RegulatingControlPhase buildRegulatingControlPhase(String regulatingControlId,
+        boolean tapChangerControlEnabled, boolean ltcFlag) {
+        RegulatingControlPhase rtc = new RegulatingControlPhase();
+        rtc.regulatingControlId = regulatingControlId;
+        rtc.tapChangerControlEnabled = tapChangerControlEnabled;
+        rtc.ltcFlag = ltcFlag;
+        return rtc;
+    }
+
+    public static void initializeRatioTapChanger(RatioTapChangerAdder adder) {
```
MioRtia: I don't think this method is useful as it is already the behavior of unset `RatioTapChangerAdder`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForTransformers.java
```diff
@@ -0,0 +1,391 @@
...
+        adder.setRegulationTerminal(null);
+        adder.setTargetV(Double.NaN);
+        adder.setTargetDeadband(Double.NaN);
+        adder.setRegulating(false);
+    }
+
+    public static void initializePhaseTapChanger(PhaseTapChangerAdder adder) {
```
MioRtia: I don't think this method is useful as it is already the behavior of unset `PhaseTapChangerAdder`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForTransformers.java
```diff
@@ -0,0 +1,391 @@
...
+    }
+
+    public static class RegulatingControlPhase extends RegulatingControlReference {
+        boolean ltcFlag;
+    }
+
+    private static class RegulatingControlReference {
+        String regulatingControlId;
+        boolean tapChangerControlEnabled;
+    }
```
MioRtia: I feel like there are too many transition classes to hold the data and this is too confusing. Could you only use `RegulatingControl`, a class named `RegulatingControlRatio` and a class named `RegulatingControlPhase`? There seems to be quite a lot of redundant information.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForTransformers.java
```diff
@@ -0,0 +1,391 @@
...
+
+    private RegulatingControlPhase buildRegulatingControlPhase(String regulatingControlId,
+        boolean tapChangerControlEnabled, boolean ltcFlag) {
+        RegulatingControlPhase rtc = new RegulatingControlPhase();
+        rtc.regulatingControlId = regulatingControlId;
+        rtc.tapChangerControlEnabled = tapChangerControlEnabled;
+        rtc.ltcFlag = ltcFlag;
+        return rtc;
+    }
```
MioRtia: Could you merge these two methods? It will increase the readability.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/TwoWindingsTransformerConversion.java
```diff
@@ -134,8 +163,34 @@ private void addTapChangers(TwoWindingsTransformer tx) {
...
+        setRegulatingControlContext(tx, rtc, ptc);
+    }
+
+    private void setRegulatingControlContext(TwoWindingsTransformer tx, String rtcId, String ptcId) {
+        RegulatingControlRatio rcRtc = null;
+        if (rtcId != null) {
+            PropertyBag tc = powerTransformerRatioTapChanger.get(rtcId);
+            rcRtc = context.regulatingControlMapping().forTransformers().buildRegulatingControlRatio(rtcId, tc);
+        } else {
+            rcRtc = context.regulatingControlMapping().forTransformers().buildEmptyRegulatingControlRatio();
```
MioRtia: Is it really useful? I feel like this should just be a null `rcRtc` rather than building a whole new empty object.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/TwoWindingsTransformerConversion.java
```diff
@@ -134,8 +163,34 @@ private void addTapChangers(TwoWindingsTransformer tx) {
...
+        }
+
+        RegulatingControlPhase rcPtc = null;
+        if (ptcId != null) {
+            PropertyBag tc = powerTransformerPhaseTapChanger.get(ptcId);
+            rcPtc = context.regulatingControlMapping().forTransformers().buildRegulatingControlPhase(tc);
+        } else {
+            rcPtc = context.regulatingControlMapping().forTransformers().buildEmptyRegulatingControlPhase();
```
MioRtia: Is it really useful? I feel like this should just be a null `rcPtc` rather than building a whole new empty object.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForTransformers.java
```diff
@@ -0,0 +1,391 @@
...
+            throw new CgmesModelException("Transformer already added, Transformer id : " + transformerId);
+        }
+
+        RegulatingControlForTwoWindingsTransformer rc = new RegulatingControlForTwoWindingsTransformer();
+        rc.ratioTapChanger = rcRtc;
+        rc.phaseTapChanger = rcPtc;
+        t2xMapping.put(transformerId, rc);
+    }
+
+    public RegulatingControlRatio buildEmptyRegulatingControlRatio() {
```
MioRtia: Is this method really useful? (see below)

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForTransformers.java
```diff
@@ -0,0 +1,391 @@
...
+        String tculControlMode, boolean tapChangerControlEnabled) {
+        RegulatingControlRatio rtc = new RegulatingControlRatio();
+        rtc.id = id;
+        rtc.regulatingControlId = regulatingControlId;
+        rtc.tculControlMode = tculControlMode;
+        rtc.tapChangerControlEnabled = tapChangerControlEnabled;
+        return rtc;
+    }
+
+    public RegulatingControlPhase buildEmptyRegulatingControlPhase() {
```
MioRtia: Is this method really useful? (see below)

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/TwoWindingsTransformerConversion.java
```diff
@@ -82,11 +89,33 @@ private void addTapChangers(TwoWindingsTransformer tx) {
         String ptc1 = end1.getId(ptcPropertyName);
         String ptc2 = end2.getId(ptcPropertyName);
 
+        // unused, will be delete in the full conversion
```
MioRtia: If it is unused, I think you can delete it now.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/conformity/modified/CgmesConformity1ModifiedConversionTest.java
```diff
@@ -139,7 +139,7 @@ public void microBEInvalidRegulatingControl() {
 
         RatioTapChanger rtc = network.getTwoWindingsTransformer("_e482b89a-fa84-4ea9-8e70-a83d44790957").getRatioTapChanger();
         assertNotNull(rtc);
-        assertFalse(rtc.hasLoadTapChangingCapabilities());
+        assertTrue(rtc.hasLoadTapChangingCapabilities());
```
MioRtia: Just for me, could you explain why it this attribute is `true` now when it was `false` before? The regulating control is supposed to be invalid here and the default value for `loadTapChangingCapabilities` should be `false`.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/RatioTapChangerConversion.java
```diff
@@ -87,8 +87,11 @@ public void convert() {
         }
 
         rtca.setLoadTapChangingCapabilities(ltcFlag);
```
MioRtia: Is it something that should remain in `RatioTapChangerConversion`?

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/conformity/modified/CgmesConformity1ModifiedConversionTest.java
```diff
@@ -166,7 +166,7 @@ public void microBEMissingRegulatingControl() {
 
         RatioTapChanger rtc = network.getTwoWindingsTransformer("_b94318f6-6d24-4f56-96b9-df2531ad6543").getRatioTapChanger();
         assertNotNull(rtc);
-        assertFalse(rtc.hasLoadTapChangingCapabilities());
+        assertTrue(rtc.hasLoadTapChangingCapabilities());
```
MioRtia: Just for me, could you explain why it this attribute is true now when it was false before? The regulating control is supposed to be missing here and the default value for `loadTapChangingCapabilities` shoud be `false`.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForTransformers.java
marqueslanauja: Deleted

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForTransformers.java
marqueslanauja: Deleted

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForTransformers.java
marqueslanauja: i tried to clarify it selecting a more appropiate class name

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/TwoWindingsTransformerConversion.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForTransformers.java
marqueslanauja: Deleted

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/TwoWindingsTransformerConversion.java
marqueslanauja: I changed the comment. It is used for debugging in current cgmes conversion. It will be not needed in the new cgmes conversion.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForTransformers.java
marqueslanauja: Done.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/TwoWindingsTransformerConversion.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForTransformers.java
marqueslanauja: Deleted

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/conformity/modified/CgmesConformity1ModifiedConversionTest.java
marqueslanauja: In this unitary test the `ltcFlag` = `true` and the regulating control is invalid. 
In the current version under these circumstances the `loadTapChangingCapabilities` is forced to `false`.
In this PR i changed the behaviour allowing the option `ltcFlag` = `true` and regulating control invalid (See reference below) so according to the new code implementation the `loadTapChangingCapabilities` in the unitary test should be `true`.

Implementation Guidelines – CGMES v2.4.15  page 9

ltcFlag = false and TapChangerControl not present

A real and fixed tap that is not controlled but can be moved
under load, e.g. manually. Power flow cannot be set to change
the tap for voltage control during the calculation. Optimal
power flow might have access to these taps and change in order
to optimize. Also State Estimator may estimate the tap position
to find a better solution for the system state.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/RatioTapChangerConversion.java
marqueslanauja: I will depend on the final behaviour of the `loadTapchangingCapabilities` attribute. If finally `ltcFlag` and `tapChangerControl` are not related it is ok.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/conformity/modified/CgmesConformity1ModifiedConversionTest.java
marqueslanauja: Same case as in `microBEInvalidRegulatingControl`.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/conformity/modified/CgmesConformity1ModifiedConversionTest.java
MioRtia: Okay.

MioRtia:
I tried to make `RegulatingControlMappingForTransformers` more readable, it was too complicated for me.
Furthermore, I corrected a bug (already present in the refactoring of generators): a regulating control object must not be deleted when set once as it can be used by several different equipment, it must only be set as **correctly set** for this equipment.

MioRtia:
> I have checked the code after your refactoring and i think that the behavior of the code is not the same. In the previous version a `ratioTapChanger` control was only disabled if there was a `phaseTapChanger` control well defined and enabled.
> In the new version to deactivate the control only the enabled attribute is checked. Could happen that at the end the `phaseTapChanger` will not be defined. That will happen if the `controlMode` or the `regulating terminal` is not well defined.
> If you agree with this the refactoring code is ok.

Yes, you're right, I modified it: I'm creating first the ptc and only if the created object is regulating, will I disable the rtc. If this is okay for you, we will merge this refactoring.

marqueslanauja:
> > I have checked the code after your refactoring and i think that the behavior of the code is not the same. In the previous version a `ratioTapChanger` control was only disabled if there was a `phaseTapChanger` control well defined and enabled.
> > In the new version to deactivate the control only the enabled attribute is checked. Could happen that at the end the `phaseTapChanger` will not be defined. That will happen if the `controlMode` or the `regulating terminal` is not well defined.
> > If you agree with this the refactoring code is ok.
> 
> Yes, you're right, I modified it: I'm creating first the ptc and only if the created object is regulating, will I disable the rtc. If this is okay for you, we will merge this refactoring.

Is ok for me.

PR #980, opened 2019-10-15
Hybridperfvisitor

jonenst:

NO

performance improvement

To sum a full binary tree of 16 million zeroes, it takes 2s. 

With this patch it takes 0.2s

**Other information**:
Performance is on back to that of v2.5.0 (for trees that are not too deep)

NO


time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/NodeCalc.java
```diff
@@ -14,17 +14,19 @@
...
+import java.util.Deque;
 import java.util.Objects;
 
 /**
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
  */
 public interface NodeCalc {
 
-    <R, A> List<NodeCalc> acceptIterate(NodeCalcVisitor<R, A> visitor, A arg);
+    <R, A> R accept(NodeCalcVisitor<R, A> visitor, A arg, int depth);
```
geofjamg: That would be the best place to explain the difference between accept, acceptIterate and acceptHandle.

PR #991, opened 2019-10-30
Groovy25

jonenst:

Feature, Upgrade, Cleanup

* Powsybl is using an old groovy version (2.4.0 was released in Jan 2015). Latest stable is groovy 2.5 (2.5.0 was released in may 2018).
* Powsybl contains groovy modules (e.g. groovy-ant, groovy-bsf, groovy-sql..) that don't work because they are missing dependencies
* Powsybl code that uses the META-INF/services/org.codehaus.groovy.runtime.ExtensionModule can not be packaged as a java9 JPMS module.

Groovy is updated to 2.5.8.
groovy-all is replaced with the multiple groovy module jar, only the ones that we use and that work.
META-INF/services/org.codehaus.groovy.runtime.ExtensionModule are moved to META-INF/groovy/org.codehaus.groovy.runtime.ExtensionModule (possible since groovy 2.5)

Here's a list of groovy-modules whose classes are no longer included. Note that the modules would not have been working anyway if their dependencies were missing (they were missing unless included manually or by chance because they were optional)
    
    groovy-ant
    groovy-bsf
    groovy-groovydoc
    groovy-jmx
    groovy-jsr223
    groovy-servlet
    groovy-sql
    groovy-test
    groovy-testng


geofjamg:
Maybe we should mention it in migration guide because it will forces upstream project to also migrate to Groovy 2.5 if they use it internally for other purposes than powsybl core ones.

jonenst:
Added https://github.com/powsybl/powsybl-core/wiki/Migration-guide-v3.1.0

Everyone is ok with the removal of the groovy modules ?

geofjamg:
> 
> 
> Added https://github.com/powsybl/powsybl-core/wiki/Migration-guide-v3.1.0
> 
> Everyone is ok with the removal of the groovy modules ?

ok

PR #994, opened 2019-10-31
PST : change tap in relative

hamou-amroun:

No

Feature : define new methods in groovy that permits to setTapPosition of a TwoWindingsTransformer by delta

SetTapPosition of a TwoWindingsTransformer in absolut

SetTapPosition of a TwoWindingsTransformer in relative

**Other information**:


action/action-util/src/main/java/com/powsybl/action/util/PhaseShifterDeltaTapTask.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Override
+    public void modify(Network network, ComputationManager computationManager) {
+        Objects.requireNonNull(network);
+        TwoWindingsTransformer phaseShifter = network.getTwoWindingsTransformer(phaseShifterId);
+        if (phaseShifter == null) {
+            throw new PowsyblException("Transformer '" + phaseShifterId + "' not found");
+        }
+        PhaseTapChanger phaseTapChanger = phaseShifter.getPhaseTapChanger();
+        Objects.requireNonNull(phaseTapChanger);
```
yichen88: Not needed, as would be checked in next line.

action/action-util/src/main/java/com/powsybl/action/util/PhaseShifterDeltaTapTask.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.powsybl.commons.PowsyblException;
+import com.powsybl.computation.ComputationManager;
+import com.powsybl.contingency.tasks.ModificationTask;
+import com.powsybl.iidm.network.Network;
+import com.powsybl.iidm.network.PhaseTapChanger;
+import com.powsybl.iidm.network.TwoWindingsTransformer;
+
+import java.util.Objects;
+
```
yichen88: Add author info

action/action-util/src/main/java/com/powsybl/action/util/PhaseShifterDeltaTapTask.java
```diff
@@ -0,0 +1,53 @@
+/**
+ * Copyright (c) 2017, RTE (http://www.rte-france.com)
```
yichen88: Wrong year and same in other file

action/action-dsl/src/test/java/com/powsybl/action/dsl/ActionDslLoaderTest.java
```diff
@@ -151,6 +186,73 @@ private void addPhaseShifter() {
...
     }
+
+    public static class DeltaTapData {
+
+        private int iniTapPosition;
+        private int expectedTapPosition;
+        private int deltaTap;
+        private String testName;
+
+        public DeltaTapData() {
```
yichen88: Remove this and do not `public` if possible.

action/action-util/src/main/java/com/powsybl/action/util/PhaseShifterDeltaTapTask.java
hamou-amroun: Done !

action/action-util/src/main/java/com/powsybl/action/util/PhaseShifterDeltaTapTask.java
hamou-amroun: Done !

action/action-util/src/main/java/com/powsybl/action/util/PhaseShifterDeltaTapTask.java
hamou-amroun: Done !

action/action-dsl/src/test/java/com/powsybl/action/dsl/ActionDslLoaderTest.java
hamou-amroun: Done !

action/action-dsl/src/test/java/com/powsybl/action/dsl/ActionDslLoaderTest.java
```diff
@@ -151,6 +186,70 @@ private void addPhaseShifter() {
...
+                    .setX(2.0)
+                    .setG(3.0)
+                    .setB(4.0)
+                    .setAlpha(5.0)
+                    .setRho(6.0)
+                .endStep()
                 .add();
     }
+
+    private static final class DeltaTapData {
```
sylvlecl: Not very important for such a test class, but setters don't seem useful:
we can make this class immutable (remove setters and have final fields).

action/action-dsl/src/test/java/com/powsybl/action/dsl/ActionDslLoaderTest.java
```diff
@@ -89,6 +92,38 @@ public void testFixTapDslExtension() {
...
+                new DeltaTapData(1, 3, 3, "deltaTap3"),
+                new DeltaTapData(1, 3, 10, "deltaTap10"),
+                new DeltaTapData(1, 0, -1, "deltaTapMinus1"),
+                new DeltaTapData(1, 0, -2, "deltaTapMinus2"),
+                new DeltaTapData(1, 0, -10, "deltaTapMinus10")
+        );
+    }
+
+    @Test
+    public void testDeltaTapDslExtension() {
```
sylvlecl: Note : maybe we should add a dummy unit test for the behaviour on invalid phase shifter ID, or on transformers which do not have a phase shifter.

action/action-dsl/src/test/java/com/powsybl/action/dsl/ActionDslLoaderTest.java
```diff
@@ -89,6 +93,66 @@ public void testFixTapDslExtension() {
...
+        for (DeltaTapData data : provideParams()) {
+            ActionDb actionDb = new ActionDslLoader(new GroovyCodeSource(getClass().getResource("/actions2.groovy"))).load(network);
+            Action deltaTapAction = actionDb.getAction(data.getTestName());
+            assertNotNull(deltaTapAction);
+            assertEquals(data.getDeltaTap(), ((PhaseShifterDeltaTapTask) deltaTapAction.getTasks().get(0)).getTapDelta());
+            addPhaseShifter(data.getIniTapPosition());
```
mathbagu: Typo: getIni**t**TapPosition or getIni**tial**TapPosition

action/action-dsl/src/test/java/com/powsybl/action/dsl/ActionDslLoaderTest.java
```diff
@@ -89,6 +93,66 @@ public void testFixTapDslExtension() {
...
+        ActionDb actionDb = new ActionDslLoader(new GroovyCodeSource(getClass().getResource("/actions2.groovy"))).load(network);
+        Action deltaTapAction = actionDb.getAction("InvalidTransformerId");
+        assertNotNull(deltaTapAction);
+        assertEquals(-10, ((PhaseShifterDeltaTapTask) deltaTapAction.getTasks().get(0)).getTapDelta());
+        try {
+            deltaTapAction.run(network, null);
+        } catch (PowsyblException ex) {
+            assertNotNull(ex);
+            assertEquals(ex.getMessage(), "Transformer 'NHV1_NHV2_1' not found");
+        }
```
mathbagu: Use `@Rule` instead?

action/action-dsl/src/test/java/com/powsybl/action/dsl/ActionDslLoaderTest.java
```diff
@@ -89,6 +93,66 @@ public void testFixTapDslExtension() {
...
+        ActionDb actionDb = new ActionDslLoader(new GroovyCodeSource(getClass().getResource("/actions2.groovy"))).load(network);
+        Action deltaTapAction = actionDb.getAction("TransformerWithoutPhaseShifter");
+        assertNotNull(deltaTapAction);
+        assertEquals(-10, ((PhaseShifterDeltaTapTask) deltaTapAction.getTasks().get(0)).getTapDelta());
+        try {
+            deltaTapAction.run(network, null);
+        } catch (PowsyblException ex) {
+            assertNotNull(ex);
+            assertEquals(ex.getMessage(), "Transformer 'NGEN_NHV1' is not a phase shifter");
+        }
```
mathbagu: Same remark here

action/action-dsl/src/test/java/com/powsybl/action/dsl/ActionDslLoaderTest.java
```diff
@@ -151,6 +215,54 @@ private void addPhaseShifter() {
...
+        private final String testName;
+
+        private DeltaTapData(int iniTapPosition, int expectedTapPosition, int deltaTap, String testName) {
+            this.iniTapPosition = iniTapPosition;
+            this.expectedTapPosition = expectedTapPosition;
+            this.deltaTap = deltaTap;
+            this.testName = testName;
+        }
+
+        private int getIniTapPosition() {
```
mathbagu: Even for test classes, I think we have to make method or field names human readable. I'm not sure **ini** is the standard word or abbreviation for initial

action/action-dsl/src/test/resources/actions2.groovy
```diff
@@ -40,3 +40,63 @@ action ('fixedTap') {
         phaseShifterFixedTap('NGEN_NHV1', 1)
     }
 }
+
+action ('deltaTap0') {
+    tasks {
+        phaseShifterDeltaTap('NGEN_NHV1', 0)
```
mathbagu: Is `phaseShifterTap()` a better name? The "fixed" equivalent exist, so maybe we can avoid it (not sure of this proposition)
If you change this keyword, also change the corresponding tasks (java & groovy)

action/action-util/src/main/java/com/powsybl/action/util/PhaseShifterDeltaTapTask.java
```diff
@@ -0,0 +1,55 @@
...
+            throw new PowsyblException("Transformer '" + phaseShifterId + "' is not a phase shifter");
+        }
+        adjustTapPosition(phaseTapChanger);
+        phaseTapChanger.setRegulating(false);
+        phaseTapChanger.setRegulationMode(PhaseTapChanger.RegulationMode.FIXED_TAP);
+    }
+
+    private void adjustTapPosition(PhaseTapChanger phaseTapChanger) {
+        phaseTapChanger.setTapPosition(Math.min(Math.max(phaseTapChanger.getTapPosition() + tapDelta,
+                phaseTapChanger.getLowTapPosition()), phaseTapChanger.getHighTapPosition()));
```
mathbagu: To be discussed: do we prefer this implementation or to throw an exception if the target tap is out of range? Personally I prefer to throw, or at least to have a warning.

sylvlecl: I agree that powsybl user should have a way to know that an action has not fully applied : maybe we could answer this need in further evolutions by returning a richer model of the result of an action application ? Plain logging may not be enough for a proper use in an application.

However, I think throwing here will be too constraining for the user : 
one of the goals of the "delta" implementation is for the user to be able to specify an action which will have more or less the same effect independently from the underlying network state. Throwing depending on the position of the initial tap compared to the limits removes this "feature", and from our business process point of view, it will often prevent the action to be computed without an easy fix.

In our business process, this case is simply handled by the computation possibly saying to the user that the remedial action has not been efficient (not solved constraints). If instead he receives a "not computed", he will not know if the partial application would have been enough or not.

PR #998, opened 2019-11-04
Not using an InMemoryPlatformConfig object or a mock PlatformConfig for empty platformconfig in tests but powsybl-config-test

MioRtia:

**Other information**:


jonenst:
Do you know if the different constructors that take a PlatformConfig argument were added just for tests ? Can we change them to take a more precise *Parameters object ? We could do this in a different PR

MioRtia:
> Do you know if the different constructors that take a PlatformConfig argument were added just for tests ? Can we change them to take a more precise *Parameters object ? We could do this in a different PR

Yes, I think this one of the next steps. In my opinion, we should do the following:
- remove constructors with mock/empty `PlatformConfig` in tests
- remove constructors with configured `PlatformConfig` in tests and replace them by tests with a config file (this will probably require some modules to move around)
- remove the constructors with `PlatformConfig`

Each step should be a PR to avoid errors and confusions due to too many changes. Furthermore, I am not sure how many changes the second step will make, it might even require several PR...

PR #1007, opened 2019-11-07
Add less restrictive task creation

pl-buiquang:

- [ x] Tests for the changes have been added (for bug fixes / features)

Add a less restrictive way to create tasks.

Task must be related to a project file.

Task only need a name and an project reference.

### DIFF
--- a/afs/afs-ws/afs-ws-server/src/main/java/com/powsybl/afs/ws/server/AppStorageServer.java
+++ b/afs/afs-ws/afs-ws-server/src/main/java/com/powsybl/afs/ws/server/AppStorageServer.java
@@ -7,10 +7,7 @@
 package com.powsybl.afs.ws.server;
 
 import com.google.common.io.ByteStreams;
-import com.powsybl.afs.AfsException;
-import com.powsybl.afs.AppFileSystem;
-import com.powsybl.afs.ProjectFile;
-import com.powsybl.afs.TaskMonitor;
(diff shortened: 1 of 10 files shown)


afs/afs-ws/afs-ws-server/src/main/java/com/powsybl/afs/ws/server/AppStorageServer.java
```diff
@@ -595,13 +592,25 @@ public Response getNodeInfo(@ApiParam(value = "File system name") @PathParam("fi
     @Produces(MediaType.APPLICATION_JSON)
     @Path("fileSystems/{fileSystemName}/tasks")
     public Response startTask(@PathParam("fileSystemName") String fileSystemName,
-                              @QueryParam("projectFileId") String projectFileId) {
+                              @QueryParam("projectFileId") String projectFileId,
+                              @QueryParam("projectId") String projectId,
```
geofjamg: Is project id really necessary as it from a project file it is possible to reach the project by getting parents ?

pl-buiquang: Yes, but this parameter is to be used only with the next one (name).

I added two parameters to the only existing one (projectFileId), to keep backward compatibility with previous way to create a task.
So now you have to provide either a projectFileId or a projectId + a name.

PR #1008, opened 2019-11-08
BBranch calculated connectable bus from NB view can be null

### LINKED ISSUE #939: getConnectableBus() returns exception if there is not bus in voltage level
Bit of both

From a node breaker network, if there is a voltage level without bus, getConnectableBus() returns an exception "should not happen".

To be defined: should we accept null if no bus is found? Or should we create a fictitious bus in this case? If so, how do we discriminate fictitious bus from real ones?

It makes the import of the CGMES files of some contries to fail.

### LINKED ISSUE #981: connectableBus throws exception in topology calculated from node-breaker all switches open
Report a bug

Exception is thrown when `BusView` of a `Terminal` is queried for `calculatedBus` in some situations.
Specifically: the voltage level has topology kind of node-breaker and all switches are opened. In this situation no bus is calculated, and the search for a `calculatedBus` fails, throwing an exception.

A minimal example has been created as a unit test case in the branch `issue_nb_connectableBus` (module `iidm-impl`, class `NodeBreakerTest`, method `issueConnectableBus`). The following graph shows the export of the voltage level topology:

Screen Shot 2019-10-15 at 23 02 19

A bus should be calculated in the voltage level. If no bus is calculated by design, then `null` could be a valid return value from the method `connectableBus`.

  - OS Version: *

It seems the current behavior is related to change in commit 0e2578716fc913b0d8e7f779278dc5316a201d44 where the condition for throwing the exception was changed (fixed).

MioRtia:

fix #981 
fix #939

**Other information**:


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevel.java
```diff
@@ -401,8 +401,8 @@ BusExt getConnectableBus(int node) {
             // if nothing found, just take the first bus
             if (connectableBus2[0] == null) {
                 Iterator<CalculatedBus> it = getBuses().iterator();
```
geofjamg: I don't remember why the original check is done using iterator but as getBuses() returns a Collection it would be more clear to use isEmpty()

zamarrenolm: agree on using `isEmpty`

PR #1018, opened 2019-11-15
iAL generator modifications support

olperr1:

No

A feature

This PR adds a new iAL task allowing to change variables on generators.


action/action-util/src/main/groovy/com/powsybl/action/util/GeneratorModificationTaskExtension.groovy
```diff
@@ -0,0 +1,124 @@
...
+            modifs.setTargetP(targetP);
+            modifs.setpDelta(pDelta);
+            modifs.setTargetV(targetV);
+            modifs.setTargetQ(targetQ);
+            modifs.setVoltageRegulatorOn(voltageRegulatorOn);
+            modifs.setConnected(connected);
+            return modifs;
+        }
+
+        void minP(Double minP) {
```
yichen88: Not needed setter/getter methods on groovy fields, so just remove `private`. It looks more like Groovy-Style.

action/action-util/src/main/java/com/powsybl/action/util/GeneratorModificationTask.java
```diff
@@ -0,0 +1,162 @@
...
+
+        if (modifs.getTargetP() != null) {
+            setTargetpWithinBoundaries(g, modifs.getTargetP(), skipOtherConnectionChange);
+        } else if (modifs.getpDelta() != null) {
+            setTargetpWithinBoundaries(g, g.getTargetP() + modifs.getpDelta(), skipOtherConnectionChange);
+        }
+
+        if ((modifs.getTargetV() != null) && (g.isVoltageRegulatorOn())) {
+            // The target voltage can be defined only when the generator is in voltage regulation mode.
+            g.setTargetV(modifs.getTargetV());
```
sylvlecl: Open question : not sure it's useful to be that strict : maybe if a subsequent action connects this generator we'll be suprised the target V is not updated ?

action/action-util/src/main/java/com/powsybl/action/util/GeneratorModificationTask.java
```diff
@@ -0,0 +1,162 @@
...
+        }
+        if ((modifs.getTargetQ() != null) && (!g.isVoltageRegulatorOn())) {
+            // The target reactive can be defined only when the generator is NOT in voltage regulation mode.
```
sylvlecl: Same open question as above.

action/action-util/src/main/java/com/powsybl/action/util/GeneratorModificationTask.java
```diff
@@ -0,0 +1,162 @@
...
+                GeneratorUtil.connectGenerator(g);
+            }
+        } else {
+            if (g.getTerminal().isConnected()) {
+                g.getTerminal().disconnect();
+            }
+        }
+    }
+
+    private void setTargetpWithinBoundaries(Generator g, double targetP, boolean skipConnect) {
```
sylvlecl: Detail: maybe setTarget**P**

action/action-util/src/main/java/com/powsybl/action/util/GeneratorModificationTask.java
```diff
@@ -0,0 +1,162 @@
...
+            changeConnectionState(g, true);
+        }
+        g.setTargetP(Math.min(g.getMaxP(), Math.max(g.getMinP(), targetP)));
+    }
+
+    public static class Modifs {
+        private Double minP;
+        private Double maxP;
+        private Double targetP;
+        private Double pDelta;
```
sylvlecl: Open question : maybe `deltaP` would sound more consistent with other properties ? or `deltaTargetP` for a little more precision ?

action/action-util/src/main/java/com/powsybl/action/util/GeneratorUtil.java
```diff
@@ -0,0 +1,36 @@
...
+package com.powsybl.action.util;
+
+import com.powsybl.iidm.network.Bus;
+import com.powsybl.iidm.network.Generator;
+import com.powsybl.iidm.network.Terminal;
+
+/**
+ * @author Olivier Perrin <olivier.perrin at rte-france.com>
+ */
+final class GeneratorUtil {
```
sylvlecl: Maybe this would deserve to be public.

action/action-util/src/main/java/com/powsybl/action/util/GeneratorUtil.java
```diff
@@ -0,0 +1,36 @@
...
+        Terminal t = g.getTerminal();
+        t.connect();
+        if (g.isVoltageRegulatorOn()) {
+            Bus bus = t.getBusView().getBus();
+            if (bus != null) {
+                // set voltage setpoint to the same as other generators connected to the bus
+                double targetV = bus.getGeneratorStream().findFirst().map(Generator::getTargetV).orElse(Double.NaN);
+                // if no other generator connected to the bus, set voltage setpoint to network voltage
+                if (Double.isNaN(targetV) && !Double.isNaN(bus.getV())) {
+                    g.setTargetV(bus.getV());
```
sylvlecl: I know it's not yours, but it looks like the method does not do what the comment says : 
it does not set the target to the same as other generators target when it finds one.

In the case where the voltage is lower than the target voltage of connected generators, the action may not do what's expected.

Also we can wonder if it is desirable in that case to ignore the targetV that was possibly defined in the generator modification.

sylvlecl: Maybe we should add a specific unit test for this, once we have decided how it should behave.

action/action-dsl/src/test/java/com/powsybl/action/dsl/GeneratorModificationActionTest.java
```diff
@@ -0,0 +1,147 @@
...
+import org.junit.rules.ExpectedException;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * @author Olivier Perrin <olivier.perrin at rte-france.com>
+ */
+public class GeneratorModificationActionTest {
```
sylvlecl: Maybe we could add a dummy test with an invalid group ID.

action/action-util/src/main/groovy/com/powsybl/action/util/GeneratorModificationTaskExtension.groovy
olperr1: Without these setters, the DSL spec doesn't work.

action/action-util/src/main/java/com/powsybl/action/util/GeneratorModificationTask.java
geofjamg: Agree, targetV can be set even if only used later (or never...). I would also remove the  g.isVoltageRegulatorOn() check.

action/action-util/src/main/java/com/powsybl/action/util/GeneratorModificationTask.java
geofjamg: yes, deltaTargetP seems more consistent with all properties

action/action-util/src/main/java/com/powsybl/action/util/GeneratorUtil.java
geofjamg: Agreed

action/action-util/src/main/java/com/powsybl/action/util/GeneratorUtil.java
geofjamg: > I know it's not yours, but it looks like the method does not do what the comment says :
> it does not set the target to the same as other generators target when it finds one.

We could rename  it "setTargetVoltageFromState()" ?

geofjamg: > In the case where the voltage is lower than the target voltage of connected generators, the action may not do what's expected.

You talk about calculated voltage?

action/action-util/src/main/groovy/com/powsybl/action/util/GeneratorModificationTaskExtension.groovy
```diff
@@ -0,0 +1,124 @@
...
+
+import com.google.auto.service.AutoService
+import com.powsybl.action.dsl.spi.DslTaskExtension
+import com.powsybl.commons.PowsyblException
+import com.powsybl.contingency.tasks.ModificationTask
+
+/**
+ * @author Olivier Perrin <olivier.perrin at rte-france.com>
+ */
+@AutoService(DslTaskExtension.class)
```
geofjamg: A general cosmetic remark. Checktyle seems  not to be activated on Groovy file as  a lot a formatting rules are not respected in this file without any build issue.

action/action-util/src/main/java/com/powsybl/action/util/GeneratorModificationTask.java
```diff
@@ -0,0 +1,162 @@
...
+        }
+        if (modifs.getMaxP() != null) {
+            g.setMaxP(modifs.getMaxP());
+        }
+
+        if (modifs.getVoltageRegulatorOn() != null) {
+            g.setVoltageRegulatorOn(modifs.getVoltageRegulatorOn());
+        }
+
+        boolean skipOtherConnectionChange = false;
```
geofjamg: Same kind of remark than @sylvlecl. It is not because  generator is  disconnected that we should not be able to set targetP.

olperr1: When ``targetP`` is provided in a ``generatorModification`` task, we want to connect the generator if it isn't already in a connected state. But if in a same task we specify ``connected false`` and a ``targetP`` value, we want the final generator state to be disconnected.
The ``skipOtherConnectionChange`` is to ensure that the ``targetP`` modification will not reconnect the generator if we disconnected it in the same task.

action/action-util/src/main/java/com/powsybl/action/util/GeneratorUtil.java
olperr1: > Also we can wonder if it is desirable in that case to ignore the targetV that was possibly defined in the generator modification.

In ``GeneratorModificationTask.modify(Network, ComputerManager)``, "targetV" is modified after the generator connection if a target V value is specified. Thus, it won't be ignored.

I will add a test for this specific case.

sylvlecl: > 
> 
> > In the case where the voltage is lower than the target voltage of connected generators, the action may not do what's expected.
> 
> You talk about calculated voltage?

Yes : if calculated voltage is lower than target because generators are at their maximum reactive power, it would seem mor logical to re-use the target instead of the computed value, which may be too low (even lower than operational limits).

olperr1: >>>In the case where the voltage is lower than the target voltage of connected generators, the action may not do what's expected.
>>>
>>You talk about calculated voltage?
>>
>Yes : if calculated voltage is lower than target because generators are at their maximum reactive power, it would seem mor logical to re-use the target instead of the computed value, which may be too low (even lower than operational limits).

I think there is a missing instruction here. If others generators are found, no targetV is set to the current generator, which is pretty weird.

Maybe adding something like
```java
if (!Double.isNaN(targetV)) {
   g.setTargetV(targetV);
}
```
before the other test would solve the problem: we use the targetV of other generators, if any, or the computed voltage if none.

This would be compliant with the comments in the method.

sylvlecl: Agreed!

action/action-util/src/main/groovy/com/powsybl/action/util/GeneratorModificationTaskExtension.groovy
olperr1: I removed the trailing semi-colons (unnecessary in Groovy) and added a single empty line between each method to ensure the ``EmptyLineSeparator`` rules.
Tell me if some other things bother you.

PR #1020, opened 2019-11-15
Improve computation exceptions logging.

### LINKED ISSUE #1015: [Computation API] Computation exception message improvement

Feature.

Currently, the message of `ComputationException`s is always `null`, which does not look well and useful in logs:
```
16:35:14.998 [pool-8-thread-197] ERROR c.p.c.slurm.SlurmComputationManager - com.powsybl.computation.ComputationException
com.powsybl.computation.ComputationException: null
```

Also:
 * it's not possible to create an exception without a cause, while it can be useful (for instance after checking command exit code)
 * in case there is a single cause exception, it is not registered as the cause of the created exception, therefore it does not appear as such in logs

It would be useful to :
 * be able to create computation exceptions without a cause (with a plain message)
 * in case there is a single cause exception, have it as the cause of the created exception
 * in case there are multiple cause exceptions, have their messages concatenated in the created exception message

Having a better logging of computation exceptions.

  - PowSyBl Version: 3.0.0

sylvlecl:

fix #1015 

Between feature and bug.

- [X] The migration guide has been updated in the github wiki *(What changes might users need to make in their application due to this PR?)*

See [here](https://github.com/powsybl/powsybl-core/wiki/Migration-guide-v3.1.0#computation-exceptions-evolutions) for migration guide.

### DIFF
--- a/computation/src/main/java/com/powsybl/computation/ComputationException.java
+++ b/computation/src/main/java/com/powsybl/computation/ComputationException.java
@@ -6,66 +6,87 @@
  */
 package com.powsybl.computation;
 
+import com.google.common.collect.ImmutableMap;
 import com.powsybl.commons.PowsyblException;
 import com.powsybl.commons.compress.ZipPackager;
 
 import java.util.*;
 
+import static java.util.Objects.requireNonNull;
+
 /**
- * An immutable class contains all types of log collected.
+ * A special exception containing information about an error that occurred during
+ * an externally executed computation.
+ *
+ * In particular, it may contain:
+ * <ul>
+ *     <li>Standard output of executed commands</li>
+ *     <li>Standard error of executed commands</li>
+ *     <li>Other logs of executed commands as bytes, which may represent plain text files or archives
+ *     (for instance, it may contain outputs of other sub-commands of the computation)</li>
+ * </ul>
+ * As the computation may consist of multiple command executions, multiple files of each type may be registered.
+ *
+ * <p>In order to create a {@link ComputationException}, you will need to use a {@link ComputationExceptionBuilder}.
  *
  * @author Yichen TANG <yichen.tang at rte-france.com>
+ * @author Sylvain LECLERC <sylvain.leclerc at rte-france.com>
  */
 public final class ComputationException extends PowsyblException {
 
-    private final Map<String, String> outMsgByLogFileName;
+    private final Map<String, String> outLogs;
+    private final Map<String, String> errLogs;
+    private final Map<String, byte[]> fileBytes;
 
-    private final Map<String, String> errMsgByLogFileName;
-
-    private final Map<String, byte[]> bytesByFileName;
-
-    private final List<Exception> exceptions;
-
-    public ComputationException(ComputationException wrappedException, Exception e) {
-        this(wrappedException.getOutLogs(), wrappedException.getErrLogs(), wrappedException.getFileBytes(), Arrays.asList(wrappedException, e));
+    ComputationException(Map<String, String> outMap, Map<String, String> errMap, Map<String, byte[]> fileBytesMap) {
+        outLogs = ImmutableMap.copyOf(requireNonNull(outMap));
+        errLogs = ImmutableMap.copyOf(requireNonNull(errMap));
+        fileBytes = ImmutableMap.copyOf(requireNonNull(fileBytesMap));
     }
 
-    ComputationException(Map<String, String> outMap, Map<String, String> errMap, Map<String, byte[]> fileBytesMap, List<Exception> exceptions) {
-        outMsgByLogFileName = Collections.unmodifiableMap(Objects.requireNonNull(outMap));
-        errMsgByLogFileName = Collections.unmodifiableMap(Objects.requireNonNull(errMap));
-        bytesByFileName = Collections.unmodifiableMap(Objects.requireNonNull(fileBytesMap));
-        this.exceptions = Collections.unmodifiableList(Objects.requireNonNull(exceptions));
+    ComputationException(String message, Map<String, String> outMap, Map<String, String> errMap, Map<String, byte[]> fileBytesMap) {
+        super(message);
+        outLogs = ImmutableMap.copyOf(requireNonNull(outMap));
+        errLogs = ImmutableMap.copyOf(requireNonNull(errMap));
+        fileBytes = ImmutableMap.copyOf(requireNonNull(fileBytesMap));
+    }
+
+    ComputationException(Throwable cause, Map<String, String> outMap, Map<String, String> errMap, Map<String, byte[]> fileBytesMap) {
+        super(cause);
+        outLogs = ImmutableMap.copyOf(requireNonNull(outMap));
+        errLogs = ImmutableMap.copyOf(requireNonNull(errMap));
+        fileBytes = ImmutableMap.copyOf(requireNonNull(fileBytesMap));
+    }
+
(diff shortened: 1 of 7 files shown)


sylvlecl:
Please note that the coverage decrease from coveralls looks wrong.
It was around 0 before last merge of the master (and actually there is not enough code in that PR to reduce the covergae that much ... ).

computation/src/main/java/com/powsybl/computation/ComputationException.java
```diff
@@ -6,68 +6,90 @@
...
 
     private final Map<String, String> outMsgByLogFileName;
-
     private final Map<String, String> errMsgByLogFileName;
-
     private final Map<String, byte[]> bytesByFileName;
 
-    private final List<Exception> exceptions;
+    ComputationException(Map<String, String> outMap, Map<String, String> errMap, Map<String, byte[]> fileBytesMap) {
+        super();
```
mathbagu: It's not necessary to call super() without parameters.

computation/src/main/java/com/powsybl/computation/ComputationExceptionBuilder.java
```diff
@@ -133,4 +157,23 @@ private ComputationExceptionBuilder readFileToMap(@Nullable Path path, Map<Strin
...
+        if (message != null && cause != null) {
+            exception = new ComputationException(message, cause, outMsgByLogFileName, errMsgByLogFileName, bytesByFileName);
+        } else if (message != null) {
+            exception = new ComputationException(message, outMsgByLogFileName, errMsgByLogFileName, bytesByFileName);
+        } else if (cause != null) {
+            exception = new ComputationException(cause, outMsgByLogFileName, errMsgByLogFileName, bytesByFileName);
+        } else {
+            exception = new ComputationException(outMsgByLogFileName, errMsgByLogFileName, bytesByFileName);
+        }
+        return exception;
```
mathbagu: To be discussed: is it necessary to handle all the possibilities? What if `ComputationException` accept Nullable message or cause?
You could simply write `new ComputationException(message, cause, outMsgByLogFileName, ...);`

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/distributed/SecurityAnalysisExecutionHandler.java
```diff
@@ -74,12 +74,31 @@
         void customizeOptions(Path workinDir, SecurityAnalysisCommandOptions options);
     }
 
+    /**
+     * Defines the creation of computation exceptions, in particular log reading.
+     */
+    @FunctionalInterface
+    interface ExceptionHandler {
+        ComputationException createComputationException(Path workinDir, Exception cause);
```
mathbagu: Typo: workin**g**Dir

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/distributed/SecurityAnalysisExecutionHandler.java
```diff
@@ -74,12 +74,31 @@
...
+     * @param input The execution input data.
+     */
     public SecurityAnalysisExecutionHandler(ResultReader<R> reader,
                                             OptionsCustomizer optionsCustomizer,
+                                            ExceptionHandler exceptionHandler,
                                             int executionCount,
                                             SecurityAnalysisExecutionInput input) {
         this.reader = requireNonNull(reader);
         this.optionsCustomizer = optionsCustomizer;
+        this.exceptionHandler = exceptionHandler;
```
mathbagu: Is null value OK? (same question for the customizer)

mathbagu:
@yichen88 Please fix the DCO by a force push

computation/src/main/java/com/powsybl/computation/ComputationExceptionBuilder.java
sylvlecl: The `cause` could be nullable, but it's not that easy for the `message` if we want to keep the usual behaviour :
`new Throwable(null, cause)` and `new Throwable(cause)` are not equivalent, the second one takes the description of the cause as its message while the first one really does not have any message.
And we cannot move the else/if switches to the constructor and choose here which overload of the `super` constructor is called, since it must be the first line of the constructor.

So in order to keep the usual construction possibilities offered by exceptions in the JDK, I think it's better to either:
* keep it this way
* make only the `cause` nullable
* make both nullable in the constructor, but handle the case null message + non null cause in the builder, to fill the message with the cause, as is usually done by `Throwable(cause)`
What possibility would you prefer?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/distributed/SecurityAnalysisExecutionHandler.java
sylvlecl: I will remove the null possibility : it's correctly handled but not used.

computation/src/main/java/com/powsybl/computation/ComputationException.java
```diff
@@ -6,68 +6,89 @@
...
+ * As the computation may consist of multiple command executions, multiple files of each type may be registered.
+ *
+ * <p>In order to create a {@link ComputationException}, you will need to use a {@link ComputationExceptionBuilder}.
  *
  * @author Yichen TANG <yichen.tang at rte-france.com>
+ * @author Sylvain LECLERC <sylvain.leclerc at rte-france.com>
  */
 public final class ComputationException extends PowsyblException {
 
     private final Map<String, String> outMsgByLogFileName;
```
geofjamg: Would you mean outMsgLogByFileName instead of outMsgByLogFileName?

computation/src/main/java/com/powsybl/computation/ComputationException.java
```diff
@@ -6,68 +6,89 @@
...
+        errMsgByLogFileName = ImmutableMap.copyOf(requireNonNull(errMap));
+        bytesByFileName = ImmutableMap.copyOf(requireNonNull(fileBytesMap));
     }
 
     /**
      * Returns a map which log file name is {@literal key}, and standard output message is {@literal value}
-     * @return
+     * @return a map which log file name is {@literal key}, and standard output message is {@literal value}
      */
     public Map<String, String> getOutLogs() {
```
geofjamg: I would have chose getter name consistent with attribute name.

computation/src/main/java/com/powsybl/computation/ComputationExceptionBuilder.java
```diff
@@ -16,66 +16,106 @@
...
+ *
  * @author Yichen TANG <yichen.tang at rte-france.com>
+ * @author Sylvain LECLERC <sylvain.leclerc at rte-france.com>
  */
 public class ComputationExceptionBuilder {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ComputationExceptionBuilder.class);
 
+    private String message;
+    private Throwable cause;
```
geofjamg: Could be final as it is set in builder constructor

computation/src/main/java/com/powsybl/computation/ComputationExceptionBuilder.java
geofjamg: I would keep it as it is. It make sense to me to stick to existing Exception constructors.

mathbagu:
@sylvlecl The DCO has to be solved or you will have to play with security settings...

PR #1031, opened 2019-11-26
[WIP]Add new api in bus

yichen88:

Feature

From a bus, it could not get connected terminals on it. But only the number of terminal connected on.

Get connected terminals.

**Other information**:


geofjamg:
Could you add a unit test?

geofjamg:
@yichen88 is there a way a add another unit test to reach 80% code coverage?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/CalculatedBusImpl.java
```diff
@@ -24,11 +25,14 @@
 
     private final List<NodeTerminal> terminals;
 
+    private final List<Integer> nodes;
```
sylvlecl: It's probably better for memory usage to keep it as `TIntArrayList` here and maybe transform it only in the getter.

@mathbagu, @geofjamg : I wonder if we should worry about the additional memory consumption : but alternatives would imply to kind of recompute this list after the fact, which is not great.

geofjamg: We should, there is a huge number of nodes in a real  node/breaker network and this code will cause of lot of boxing in addition to additional memory consumption.
At least we could convert from TIntArrayList to List<Integer> in getNodes() to only penalize few codes that will use this new method.
Or maybe we could replace getNodes() by getNodeCount + getNode(index) to wrap internal TIntArrayList and avoid copy.

mathbagu: Is it really a good idea to expose getNodes in the public API? Why do you need this? This is not available in all implementation: for instance, configured bus don’t have this list. I think it’s only valid in N/B voltage levels

geofjamg: > 
> 
> Is it really a good idea to expose getNodes in the public API? Why do you need this? This is not available in all implementation: for instance, configured bus don’t have this list. I think it’s only valid in N/B voltage levels

You are right, it has nothing to do in Bus interface but it could be in the node/breaker view of the voltage level.

sylvlecl: We would like, for buses calculated from a node breaker view, to be able to know what's part of the bus, in particular which nodes. Today the API does not allow this.

About the API, if we want to make it more clean for bus breaker topologies we can:
* subclass `Bus` for the node breaker view, to add access to those more specific data
* add methods in `NodeBreakerView`, which would look like `int[] getNodes(Bus bus)`
* or leave it as proposed in the PR : there are already methods which throw for bus breaker voltage levels (all methods of the node breaker view)

About performances:
* I would be ok for `int getNodeCount` and `int[] getNodes`, it's consistent with other classes
* if we want to optimize for memory, there is still the possibility to recompute the list on request, but if we do this, we must make sure it's the same "visitor" that what used for the initial computation which takes care of this

geofjamg:
Why not just moving getNodeCount and  getNodes() in the NodeBreakerView of the terminal and  node in Bus like:
```java
int nodeCount = generator.getTerminal().getNodeBreakerView().getNodeCount(busId);
int node0 = generator.getTerminal().getNodeBreakerView().getNode(busId, 0);
```

Issue #1039, opened 2019-11-28
Identifiable objects cannot hold properties with type other than String, this break groovy scripts


pl-buiquang:
* **Do you want to request a *feature* or report a *bug*?**
bug, regression

* **What is the current behavior?**
Now when a groovy script assign to an object undefined properties a typed value (double or date) it will be casted into a String.
When reusing this property, we loose the type. This lead to runtime error in the script if we expect to use the property in a case when the typed value is needed.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
Create a modificationScript.
Get any object from the network (eg. a generator)
Set a temporary value within this object : `network.getGenerator('SOME_ID').tmpValue = 100`
Use this temporary value to value a defined property : `network.getGenerator('SOME_ID').maxP = network.getGenerator('SOME_ID').tmpValue`
This will throw a RuntimeException since the String can't be casted into a Double.

* **What is the expected behavior?**
Before commit 3f4bd8604d0e0733e68e15b10d7557051c34f754, the properties exposed was a Map<Object, Object> (inside Properties) and therefore allowed groovy scripts (via the IdentifiableExtension) to overload objects with typed properties.

* **What is the motivation / use case for changing the behavior?**
This is regression in our mapping scripts, that heavily use the dynamic properties of objects and notably to store computation results (Double) or computed dates.

* **Please tell us about your environment:**
  - PowSyBl Version: 3.1.0-SNAPSHOT
  - OS Version: Linux 70cc24868cda 3.10.0-693.11.6.el7.x86_64

PR #1040, opened 2019-11-28
Readd Properties to default AbstractIdentifiable

pl-buiquang:

https://github.com/powsybl/powsybl-core/issues/1039


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiable.java
```diff
@@ -60,17 +64,19 @@ public boolean hasProperty(String key) {
 
     @Override
     public String getProperty(String key) {
-        return properties.get(key);
+        Object val = properties.get(key);
+        return val != null ? val.toString() : null;
```
mathbagu: Could the property key really be null? In getPropertyNames() it seems you assume that it’s never null.

pl-buiquang: no the property key is indeed assumed to be null.
Though the value associated to the key may be null.

geofjamg: I think @mathbagu just want you to add a null check on the key as it is not done inside properties.get().

mathbagu: I think the properties class throw the exception (NullPointerException) so it should be ok like this.

PR #1042, opened 2019-12-02
Add methods for CGMES validation

MioRtia:

Feature.
Some methods are added to allow access to CGMES conversion context and CGMES model profiles.

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
@@ -188,11 +188,11 @@ public class CgmesImport implements Importer {
                                 p,
                                 STORE_CGMES_MODEL_AS_NETWORK_EXTENSION_PARAMETER,
                                 defaultValueConfig))
-                .setStoreCgmesTerminalMappingAsNetworkExtension(
+                .setStoreCgmesConversionContextAsNetworkExtension(
                         ConversionParameters.readBooleanParameter(
                                 getFormat(),
                                 p,
-                                STORE_CGMES_TERMINAL_MAPPING_AS_NETWORK_EXTENSION_PARAMETERT,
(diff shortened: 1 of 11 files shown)


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
```diff
@@ -103,11 +103,6 @@ public String getFormat() {
     public Network importData(ReadOnlyDataSource ds, NetworkFactory networkFactory, Properties p) {
         CgmesModel cgmes = CgmesModelFactory.create(ds, boundary(p), tripleStore(p));
         Network network = new Conversion(cgmes, config(p), activatedPostProcessors(p), networkFactory).convert();
```
zamarrenolm: As we have moved the management of the extensions inside the conversion object, we could return directly the created the network without storing it in an intermediate variable

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
```diff
@@ -187,8 +182,19 @@ private String tripleStore(Properties p) {
...
-                                defaultValueConfig)
-                );
+                                defaultValueConfig))
+                .setStoreCgmesModelAsNetworkExtension(
+                        ConversionParameters.readBooleanParameter(
+                                getFormat(),
+                                p,
+                                STORE_CGMES_MODEL_AS_NETWORK_EXTENSION_PARAMETER,
+                                defaultValueConfig))
+                .setStoreCgmesTerminalMappingAsNetworkExtension(
```
zamarrenolm: why not store the whole conversion context as a separate network extension? I see that it is good to have separate extensions for CGMES model and for conversion context, but I think we could just save the whole context for debugging/validation purposes, instead of creating one extension for each piece of the context. What do you think?

MioRtia: Yes, I agree, I made changes in this direction.

MioRtia:
@annetill if this is okay for you, we can merge this

Issue #1043, opened 2019-12-02
IEEE import precision


annetill:
* **Do you want to request a *feature* or report a *bug*?**

Maybe a bug.

* **What is the current behavior?**

I have analyzed the IEEE case with 14 buses. I have made an import and then run a loadflow (Hades2, https://rte-france.github.io/hades2/index.html). Some buses seems to have more precise results than others. For example, the buses with almost identical results (less than 10^-5) for voltage are Buses 1, 2, 3, 6 and 8. It corresponds to buses with generators or synchronous compensator. Buses 4, 7, 11, 12, 13 ,14 are really less precise (less than 10^-2) for voltages. 

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**

Maybe more precise voltages for buses 4, 7, 11, 12, 13 ,14. I Wonder if we well convert in IIDM the reactances and the resistances.

* **What is the motivation / use case for changing the behavior?**

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

zamarrenolm:
The IEEE14 test case included in main resources for IEEE CDF model has been obtained from the Power System Test Case Archive of the University of Washington: https://labs.ece.uw.edu/pstca/pf14/pg_tca14bus.htm.

The page has a disclaimer that states: _The data has not been run through a power flow to check the answers_. 

Also, we must take into account that the voltages given in the CDF format have only 3 decimal places of precision (voltage modules are given in per-unit and angles given in degrees).

University of Illinois has published the same test case solved in different formats (not IEEE CDF). The files are available at https://icseg.iti.illinois.edu/ieee-14-bus-system/.

The IEEE14 test case given by the University of Illinois in PSS/E Raw Data format has been verified against HELM-Flow (giving the same solution). All bus balance mismatches given in the Raw data file are `< 0.01` MVA (with flows computed in HELM-Flow after import). In the Raw file, voltage modules are given with 5 decimal places in per-unit, and angles with 4 decimals in degrees.

[...]

PR #1047, opened 2019-12-03
use double instead of float in IEEE CDF model; add a solved case test

zamarrenolm:

This change addresses the issue #1043.

To be able to validate solved cases in IEEE CDF format, the reactive generation given at bus must be preserved even if the generator is regulating voltage or voltage and angle. The reactive generation given in the CDF file is stored as `targetQ` in the converted `IIDM Generator`. An alternative to this could be to put the active/reactive load and generation of each bus as flows at the terminal related to the equipment created.

This change completes the verification of the IEEE importer and proposes to store the parameters and values of the IEEE CDF model using the type `double` instead of `float`.

See #1043.


annetill:
I have made a comparison for 14 bus IEEE network. The voltages are closed to the results in "ieee14cdf-solved.txt". It means that we assume that voltages are equals with a 10^-3 precision.

PR #1050, opened 2019-12-04
Cgmes conversion fix alpha

marqueslanauja:

Bug fix

`Step X` is calculated using `alpha` and `alphaMax` associated to `rho`

`Step X` is calculated using `alpha` and `alphaMax` associated to `ratio`

**Other information**:

Fix the current cgmes conversion before the PR associated to the new conversion.
The error was founded comparing results from both conversions


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/PhaseTapChangerConversion.java
```diff
@@ -350,15 +351,16 @@ private void addSteps(
 
         for (int i = 0; i < alphas.size(); i++) {
             double alpha = alphas.get(i);
+            double alphaRatio = -alpha;
```
annetill: I am not sure to understand the bug : at this stage of the code, the tables ````rhos```` and ````alphas```` are in the IIDM convention. In fact, you want to go back to CGMES convention to call the functions ````getStepXforAsymmetrical```` and ````getStepXforSymmetrical````. Am I right ?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/PhaseTapChangerConversion.java
```diff
@@ -350,15 +351,16 @@ private void addSteps(
...
             double rho =  rhos.get(i);
             double x = 0.0;
             if (!xStepRangeIsConsistent || alphaMax == 0) {
                 x = tx.getX();
             } else {
                 if (asymmetrical()) {
-                    x = getStepXforAsymmetrical(xStepMin, xStepMax, alpha, alphaMax, theta);
+                    x = getStepXforAsymmetrical(xStepMin, xStepMax, alphaRatio, alphaMax, theta);
```
annetill: The sign changes are a bit confusing for me and I wonder if the best way to handle this bug is with your solution or if it is maybe better to change the functions ````getStepXforAsymmetrical/Symetrical```` and assume that they are in IIDM convention ?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/PhaseTapChangerConversion.java
marqueslanauja: You are right. The functions `getStepXforAsymmetrical` and `getStepXforSymmetrical` are coded using CGMES convention. 
Instead of changing the `getStepXforAsymmetrical` and `getStepXforSymmetrical`  to the IIDM convection i propose to spend a couple of hours to refactorize this code following a schema closer to the new CGMES conversion. It will be a small change.
In this schema i will work with CGMES convention (angles and ratios) until we set this information to IIDM. At this point i will change to IIDM convention and i will move to side 1 if it is necessary.
I know that the current conversion will be changed by the new one but at least after the refactorization will be easier to compare both. 
Are you agree?

marqueslanauja:
> I have the same questions as @annetill. I thought the sign changes were done line 273 and line 302. Why do we need to change the sign convention again?
The problem is that in the functions  `getStepXforAsymmetrical` and `getStepXforSymmetrical` the alpha and alphaMax arguments are expected to be associated to ratio (CGMES convention) and not to rho (IIDM convention). As the sign has been changed before calculating the stepX  adjustment it is necessary to change it again to call this functions. 
As I answered to Anne in a previous comment i propose to refactorize this code changing from CGMES convention to IIDM convention just before setting the rho to IIDM, so only one sign change will be necessary and the call to this functions will be ok.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/PhaseTapChangerConversion.java
MioRtia: I agree, it will more comprehensive I think.

annetill: To conclude: there is a bug (which I understand well) and you propose a quick solution but you propose to refactorize this code in a next PR ?

marqueslanauja: I have done a refactorization of this part of the code to be more clear. We are testing now the refactored code with the conformity cases. After the test i will update this PR. The idea is to include the refactorization in this PR.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/PhaseTapChangerConversion.java
```diff
@@ -335,41 +309,62 @@ private void addSteps(
...
+        LOG.debug("ACTUAL    angleMax {}", angleMax);
         LOG.debug("ACTUAL    xStepMin, xStepMax {}, {}", xStepMin, xStepMax);
 
-        // rho0 adjustment computed the same way that is done in TwoWindingsTransformer,
-        // using factor rho0square as a float
-        double rho0 = tx.getRatedU2() / tx.getRatedU1();
-        double rho0square = rho0 * rho0;
-        LOG.debug("ACTUAL    u2,u1,rho0square {}, {}, {}", tx.getRatedU2(), tx.getRatedU1(), rho0square);
+        // Take structural ratio
+        double ratio0 = tx.getRatedU2() / tx.getRatedU1();
```
annetill: Why ````ratio0 = tx.getRatedU2() / tx.getRatedU1()```` ? I expected the inverse here.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/PhaseTapChangerConversion.java
```diff
@@ -242,45 +242,36 @@ private double theta() {
 
     private void fillAlphasRhos(
```
annetill: Maybe we have to rename the method ?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/PhaseTapChangerConversion.java
```diff
@@ -242,45 +242,36 @@ private double theta() {
...
             double du0, double du, double theta,
-            List<Double> alphas,
-            List<Double> rhos) {
+            List<Double> angles,
+            List<Double> ratios) {
         if (asymmetrical()) {
-            fillAlphaRhoListsAsymmetrical(du0, du, theta, alphas, rhos);
+            fillAlphaRhoListsAsymmetrical(du0, du, theta, angles, ratios);
```
annetill: Maybe we have to rename the method ?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/PhaseTapChangerConversion.java
```diff
@@ -242,45 +242,36 @@ private double theta() {
...
         } else if (symmetrical()) {
-            fillAlphaRhoListsSymmetrical(du0, du, theta, alphas, rhos);
+            fillAlphaRhoListsSymmetrical(du0, du, theta, angles, ratios);
```
annetill: Maybe we have to rename the method ?

cgmes/cgmes-conformity/src/test/java/com/powsybl/cgmes/conformity/test/CgmesConformity1NetworkCatalog.java
```diff
@@ -966,8 +966,8 @@ private static void addPhaseTapChanger(
             PhaseTapChanger.RegulationMode mode, boolean regulating,
             double regulationValue, double targetDeadband) {
         LOG.debug("EXPECTED tx {}", tx.getId());
-        double rho0 = tx.getRatedU2() / tx.getRatedU1();
-        double rho02 = rho0 * rho0;
+        double ratio0 = tx.getRatedU2() / tx.getRatedU1();
```
annetill: The ratio in CGMES is primary nominal voltage / seconday nominal voltage, no ?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/PhaseTapChangerConversion.java
marqueslanauja: Done. Sorry

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/PhaseTapChangerConversion.java
marqueslanauja: Done. Sorry

cgmes/cgmes-conformity/src/test/java/com/powsybl/cgmes/conformity/test/CgmesConformity1NetworkCatalog.java
marqueslanauja: It is the structural ratio at end2 (after transmission impedance) so in this case the ratio is secondary nominal voltage / primary nominal voltage.
The structural ratio at end1 (before transmission impedance) is primary nominal voltage / secondary nominal voltage as you said but in the current CGMES conversion the structural ratio is always considered at end2 (after transmission impedance)

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/PhaseTapChangerConversion.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/PhaseTapChangerConversion.java
marqueslanauja: It is the structural ratio at end2 (after transmission impedance) so is ok.

marqueslanauja: > It is the structural ratio at end2 (after transmission impedance) so is ok.

The structural ratio at end2 (after transmission impedance) is secondary nominal voltage / primary nominal voltage.
The structural ratio at end1 (before transmission impedance) is primary nominal voltage / secondary nominal voltage as you said but in the current CGMES conversion the structural ratio is always considered at end2 (after transmission impedance)

annetill: Yes you are right, it is the ratio we use to "transport" the transmission impedance from side 1 to side 2.

annetill:
We can merge this PR.

PR #1059, opened 2019-12-06
Equivalent shunts cgmes

### LINKED ISSUE #1056: CGMES EquivalentShunt not converted in IIDM

Feature.

It seems that an EquivalentShunt from a CGMES file is not converted in a IIDM Shunt Compensator.

I think we have to convert the CGMES EquivalentShunt into Shunt Compensator.

I think it could explain voltage differences between what Hades2 computes through Powsybl and the SV values of PSE CGMES file (see June interoperability dataset). 

  - PowSyBl Version: ...
  - OS Version: ...

MioRtia:

fix #1056


zamarrenolm:
Because equivalent shunts can not have regulating capabilities, they could also be converted to loads if IIDM was able to handle constant impedance loads. 

For reference: CIMdesk paints equivalent injections and shunts using the same symbol that applies to energy consumers (loads). In CIMdesk, only _real_ shunt compensators (not equivalents) are painted using the capacitor/reactor symbol.

MioRtia:
> Because equivalent shunts can not have regulating capabilities, they could also be converted to loads if IIDM was able to handle constant impedance loads.
> 
> For reference: CIMdesk paints equivalent injections and shunts using the same symbol that applies to energy consumers (loads). In CIMdesk, only _real_ shunt compensators (not equivalents) are painted using the capacitor/reactor symbol.

Interesting take, I think the solution in the PR is good for a first fix but maybe we will have to think about how we modelize equivalent equipment in IIDM.

PR #1060, opened 2019-12-06
Update modification date on object changes

pl-buiquang:

feature

the modification date of the afs object isn't updated on object changes (content or dep changes)

the modification date is updated over various changes depending of the object type

### DIFF
--- a/security-analysis/security-analysis-afs-local/src/main/java/com/powsybl/security/afs/local/LocalSecurityAnalysisRunningService.java
+++ b/security-analysis/security-analysis-afs-local/src/main/java/com/powsybl/security/afs/local/LocalSecurityAnalysisRunningService.java
@@ -78,6 +78,7 @@ public class LocalSecurityAnalysisRunningService implements SecurityAnalysisRunn
                         return null;
                     }).join();
         } finally {
+            runner.updateModificationTime();
             runner.stopTask(taskId);
         }
     }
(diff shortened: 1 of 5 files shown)


security-analysis/security-analysis-afs-local/src/main/java/com/powsybl/security/afs/local/LocalSecurityAnalysisRunningService.java
```diff
@@ -78,6 +78,7 @@ public void run(SecurityAnalysisRunner runner) {
                         return null;
                     }).join();
         } finally {
+            runner.updateModificationTime();
```
sylvlecl: I wonder : do we want to update modification time even when no result is written ?
Maybe it's only the "writeResult" which should update the modification time ?

From a broader point of view, maybe the writeBinaryData method should actually be responsible for updating the modification date ?
(not part of this PR of course)

pl-buiquang: hm.. I guess your approach seems better. the AppStorage should be in charge of updating modification time when it modifies data (writeBinaryData, operation on timeseries and dependencies)

pl-buiquang: then the updateModificationTime method should not even be accessible since appstorage is  the sole responsable for updating afs objects

sylvlecl: "Pas faux" :)

pl-buiquang: what do you think @geofjamg ?

mathbagu:
This PR has to be moved to powsybl-afs

PR #1062, opened 2019-12-06
Refactor scripting module to remove dependency to AFS

mathbagu:
Signed-off-by: Mathieu BAGUE <mathieu.bague@rte-france.com>

N/A

Refactoring of the scripting modules, specially to remove the AFS dependency.

The scripting module depends on the AFS modules. If we want to extract all AFS modules in a separated repository, we have to remove this bound.

The powsybl-scripting module doesn't depend on AFS anymore. AFS is loaded as an extension if the powsybl-afs-scripting module is in the classpath. The loadFlow extension follow the same idea, with a new powsybl-loadflow-scripting module.

**Other information**:


scripting/scripting-afs/src/main/groovy/com/powsybl/scripting/groovy/InitPowsybl.groovy
```diff
@@ -1,35 +0,0 @@
...
-    protected InitPowsybl(final Groovysh shell) {
-        super(shell, ':init_powsybl', ':powsybl')
-    }
-
-    Object execute(List args) {
-        // force logback to error level
-        Logger rootLogger = (Logger) LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME);
-        rootLogger.setLevel(Level.ERROR);
-
-        DefaultComputationManagerConfig config = DefaultComputationManagerConfig.load();
```
geofjamg: In the new InitPowsybl, this piece of code has been removed and it is not possible anymore to use afs in Groovy shell. Is it on purpose? Maybe we should load  here all Groovy scripting extensions?

mathbagu: you right, I miss something ^^'. I will fix it

geofjamg:
As it is only code refactoring I would not care about 70% code coverage which is too low but not really related to this  work.

PR #1063, opened 2019-12-09
Extend task with optional related computation future

pl-buiquang:

feature

The task is loosely link to a real computation process

The user can set a CompletableFuture to the task object. This allow cancellation of the process via the task manager.


afs/afs-core/src/main/java/com/powsybl/afs/TaskMonitor.java
```diff
@@ -140,20 +190,79 @@ public boolean equals(Object obj) {
...
     void removeListener(TaskListener listener);
 
+    /**
+     * Update the future of the computation process monitored by this task
+     *
+     * @param taskId
+     * @param future
+     * @throws NotACancelableTaskMonitor in case the task monitor is operating as a remote task monitor
+     */
+    void updateTaskCancelableFuture(UUID taskId, Future future) throws NotACancelableTaskMonitor;
```
sylvlecl: Maybe simply `updateTaskFuture` ? The cancellable word sounds redundant.

afs/afs-core/src/main/java/com/powsybl/afs/TaskMonitor.java
```diff
@@ -35,16 +51,31 @@
...
+
+        public Task(String name,
+                    String message,
+                    long revision,
+                    String projectId) {
+            this(name, message, revision, projectId, false);
+        }
+
         @JsonCreator
         public Task(@JsonProperty("name") String name,
```
sylvlecl: Should we categorize this as breaking change, since old JSON messages will not be readable anymore ?

afs/afs-core/src/test/java/com/powsybl/afs/LocalTaskMonitorTest.java
```diff
@@ -91,11 +97,36 @@ public void onEvent(TaskEvent event) {
...
+            monitor.updateTaskCancelableFuture(task.getId(), null);
+            assertEquals(1, events.size());
+            assertEquals(new TaskCancelableStatusChangeEvent(task.getId(), 3L, false), events.pop());
+            AtomicInteger taskCounter = new AtomicInteger(0);
+            AtomicBoolean interrupted = new AtomicBoolean(false);
+            CompletableFuture<Integer> dummyTaskProcess = CompletableFuture.supplyAsync(() -> {
+                for (int i = 0; i < 1000; i++) {
+                    taskCounter.incrementAndGet();
+                    try {
+                        Thread.sleep(100);
```
sylvlecl: Maybe we could simply use a `TaskCounter` or a `CompletableFuture` to wait until it gets interrupted ? (see for example `CompletableFutureTaskTest#whenCancelDuringExecutionThenThrowAndInterrupt`)
The test relies on an arbitrary value which may not be necessary.

Not blocking though.

afs/afs-core/src/main/java/com/powsybl/afs/TaskMonitor.java
pl-buiquang: yeah ok, but maybe we should configure jackson to allow for unknown properties DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES <- false ?
I don't know how to do this in non spring apps.. :/

sylvlecl: I think for this use case it's enough to mention the change in the release note, just in case. I don't expect this feature to be widely used.

Otherwise, I guess the configuration would happen in `JacksonDecoder` class (where we use the default powsybl `ObjectMapper` at the moment).

pl-buiquang: @geofjamg would it be ok to add `.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)` to the default objectMapper ?

pl-buiquang: as it allows to extend dtos without breaking the app

geofjamg: > @geofjamg would it be ok to add `.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)` to the default objectMapper ?

I would say no because of unexpected effects on other codes. Most of the time you want to be warmed in that case because it is an input error.

pl-buiquang: ok, it wasn't added anyway.

mathbagu:
This PR has to be moved to powsybl-afs

PR #1068, opened 2019-12-10
Add clear result method to security analysis

pl-buiquang:

**Other information**:
Add a clear result method to security analysis (to be used to automatically clean executable project files).


security-analysis/security-analysis-afs/src/main/java/com/powsybl/security/afs/SecurityAnalysisRunner.java
```diff
@@ -130,6 +130,12 @@ public void writeResult(SecurityAnalysisResult result) {
         storage.flush();
     }
 
+    public void clearResults(){
```
geofjamg: Should we keep the  s as we don't have  it in any other methods (and it is probably a mistake...)

PR #1071, opened 2019-12-11
Merging view : refactoring Network.BusView & Network.BusBreakerView

tadam50:

No

Try to fit IIDM API for Network.BusView & Network.BusBreakerView

Network.BusView & Network.BusBreakerView adapters are not nested classes of the relevant views

Network.BusView & Network.BusBreakerView adapters are nested classes of the relevant views

**Other information**:


iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergingView.java
```diff
@@ -37,17 +37,116 @@
     /** To listen events from merging network */
     private final NetworkListener listener = new MergingNetworkListener();
 
-    private final NetworkBusBreakerViewAdapter busBreakerView;
+    static class BusBreakerViewAdapter implements Network.BusBreakerView {
```
mathbagu: Why this class is private package? Private could be enough?
It seems it's required by unit tests. It's usually a bad idea to make the model weaker just for unit test

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergingView.java
```diff
@@ -37,17 +37,116 @@
...
+
+        @Override
+        public int getSwitchCount() {
+            return (int) getSwitchStream().count();
+        }
+    }
+
+    private final BusBreakerViewAdapter busBreakerView;
+
+    static class BusViewAdapter implements Network.BusView {
```
mathbagu: Same question here

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergingView.java
tadam50: My mistake, i forgot to put private after to move BusBreakerViewAdapter class into MergingView class

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergingView.java
tadam50: Same answer

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergingView.java
```diff
@@ -37,17 +37,116 @@
...
+        public Bus getBus(final String id) {
+            final List<Network.BusView> views = index.getNetworkStream()
+                    .map(Network::getBusView).collect(Collectors.toList());
+            for (final Network.BusView bv : views) {
+                final Bus bus = bv.getBus(id);
+                if (Objects.nonNull(bus)) {
+                    return index.getBus(bus);
+                }
+            }
+            return null;
```
mathbagu: Should be implemented using lambdas:
```
return index.getNetworkStream()
        .map(n -> n.getBusView().getBus(id))
        .findFirst()
        .map(index::getBus)
        .orElse(null);
```
Not totally sure of my proposal, so I let you test it

tadam50: It works ... 
I just add a filter on nullalbe bus

PR #1073, opened 2019-12-11
Update XIIDM serialization/deserialization of ThreeWindingsTransformer

### LINKED ISSUE #1070: Adapt XIIDM serialization of T3W to new IIDM T3W model
Feature

When an attribute of the new IIDM T3W model is not set to default value, an exception is thrown during XIIDM serialization.

A new serialization should exist in XIIDM 1.1 version for T3W.

  - PowSyBl Version: 3.1.0-SNAPSHOT

MioRtia:

fix #1070


iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ThreeWindingsTransformerXml.java
```diff
@@ -154,8 +188,44 @@ protected void readSubElements(ThreeWindingsTransformer tx, NetworkXmlReaderCont
...
+                        break;
+                }
+                break;
+            case V_1_0:
+                consumer.accept(tx, context);
+                break;
+            default:
+                throw new PowsyblException("XIIDM version " + context.getVersion().toString(".") + " is not supported for ThreeWindingsTransformer.");
+        }
+    }
```
mathbagu: I'm not sure about the code design...To avoid code duplication, we could introduce a function that use a predicate to know if the attribute should be read (or write)? We can also use readOptionalDoubleAttribute for new attributes to reduce the complexity of the code...

I think we have to create a design, developer could follow for the new changes in the format. I'm not sure switch/case is the best design.

The default case should not throw PowsyblException but AssertionError: it should not happen?!

MioRtia: I agree with you proposition, I was not so sure myself about the way I coded it.

Actually, after some thinking, I think the default case should be the same as the most recent version (the serialization/reading of an equipment does not change for more recent version of IIDM unless explicitly specified), what do you think?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractTransformerXml.java
```diff
@@ -20,6 +22,10 @@
  */
 abstract class AbstractTransformerXml<T extends Connectable, A extends IdentifiableAdder<A>> extends AbstractConnectableXml<T, A, Substation> {
 
+    private interface StepConsumer {
```
mathbagu: Probably also static?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractTransformerXml.java
```diff
@@ -196,4 +186,28 @@ protected static void readPhaseTapChanger(TwoWindingsTransformer twt, NetworkXml
             adder.add();
         }
     }
+
+    protected static void readPhaseTapChanger(TwoWindingsTransformer twt, NetworkXmlReaderContext context) throws XMLStreamException {
+        readPhaseTapChanger("phaseTapChanger", twt.newPhaseTapChanger(), twt.getTerminal1(), context);
```
mathbagu: Should we introduce a constant for phaseTapChanger?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractTransformerXml.java
MioRtia: Nested interfaces are apparently static by default (https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.5.1) so the modifier is redundant.

> A member interface is implicitly static (§9.1.1). It is permitted for the declaration of a member interface to redundantly specify the static modifier.

PR #1079, opened 2019-12-12
Merging view : HvdcLine adapter implementation

tadam50:

No

New adaptated classes are implemented.
Each method adaptated not implemented yet throw an exception

Merging view module implementation need to implement hvdcline adapter

Merging view module implementation step : hvdcline adapter implementation needed
HvdcLine creation are not allowed between two merged network, exception will be thrown

**Other information**:


iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/HvdcLineAdderAdapter.java
```diff
@@ -0,0 +1,180 @@
...
+import com.powsybl.iidm.network.HvdcLineAdder;
+import com.powsybl.iidm.network.Network;
+import com.powsybl.iidm.network.util.Identifiables;
+
+import java.util.Objects;
+
+/**
+ * @author Thomas Adam <tadam at silicom.fr>
+ */
+public class HvdcLineAdderAdapter implements HvdcLineAdder {
```
mathbagu: Why the design of this adder is completely different from the others: why don't we have a delegate here?

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/HvdcLineAdderAdapter.java
```diff
@@ -0,0 +1,180 @@
...
+        checkAndSetUniqueId();
+        return index.getHvdcLine(addLine(n1));
+    }
+
+    private Network checkAndGetNetwork1() {
+        if (Objects.isNull(converterStationId1)) {
+            throw new PowsyblException("Side 1 converter station is not set");
+        }
+        final Network network = index.getNetworkStream()
+                                     .filter(n -> n.getHvdcConverterStation(converterStationId1) != null)
```
mathbagu: The merging view should implement `getHvdcConverterStation(String)`: why don't you use it?

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/HvdcLineAdderAdapter.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    private Network checkAndGetNetwork2() {
+        if (Objects.isNull(converterStationId2)) {
+            throw new PowsyblException("Side 2 converter station is not set");
+        }
+        final Network network = index.getNetworkStream()
+                                     .filter(n -> n.getHvdcConverterStation(converterStationId2) != null)
+                                     .findFirst()
+                                     .orElse(null);
```
mathbagu: Same remark here

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/VoltageLevelAdapter.java
```diff
@@ -511,7 +511,22 @@ public StaticVarCompensatorAdder newStaticVarCompensator() {
...
+    }
+
+    @Override
+    public LccConverterStationAdder newLccConverterStation() {
+        return new LccConverterStationAdderAdapter(getDelegate().newLccConverterStation(), getIndex());
+    }
+
+    @Override
+    public Iterable<LccConverterStation> getLccConverterStations() {
+        return getLccConverterStationStream().collect(Collectors.toSet());
```
mathbagu: Is the set needed? Is the order important?

tadam50: My mistake, i willl change it to List

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/HvdcLineAdderAdapter.java
tadam50: We need to search which network contains "converterStationId1" & "converterStationId1"
In order to know, if we are creating HvdcLine between 2 networks or not.
For now, no creation are allowed between 2 networks
So, i need to get reference on networks contains into mergingview

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/HvdcLineAdderAdapter.java
tadam50: I think, the delegate cannot be used here
Which network will be used to create the HvdcLine ?
The network to use cannot be known before `setConverterStationId` methods

mathbagu: Right... My bad, it miss something when I did the review

PR #1082, opened 2019-12-12
use -T 2.0C for tests in github actions for speed

jonenst:

NO

Feature

tests take 7-9 minutes on github actions workers

Hopefully bring it down to 4-5 minutes ?

NO


geofjamg:
Results are not really convincing, I would close this PR and let it single core.

mathbagu:
@jonenst maybe you should also enable multithreading on sonar step?

jonenst:
If you look into the details, the mac time went from 6min to 3min. So it's promising. But the linux and windows build times didn't go down. I'm looking into it.

(also, the mac build times are consistently faster. Maybe we should do the sonar analysis there ? but we have only 5 workers)

jonenst:
> maybe you should also enable multithreading on sonar step?
It doesn't seem to be supported ? Jul 2018 https://community.sonarsource.com/t/parallel-analysis/1152

geofjamg:
> If you look into the details, the mac time went from 6min to 3min. So it's promising. But the linux and windows build times didn't go down. I'm looking into it.
> 
> (also, the mac build times are consistently faster. Maybe we should do the sonar analysis there ? but we have only 5 workers)

MacOS could be in the future dropped from push build (very few added value in addition to Linux one), so it is clearly not here that we would like to improve performance.

jonenst:
If the mac build is faster and "equivalent" to the linux build, why not use it instead ?

mathbagu:
@jonenst This PR is flagged WIP. Are you working on this? Do we merge it? Do we close it?

PR #1085, opened 2019-12-13
spurious share/java/apache-jena-libs-3.0.0.pom file in itools distribution #753

### LINKED ISSUE #753: spurious share/java/apache-jena-libs-3.0.0.pom file in itools distribution
bug

the itools distribution has a spurious file target/powsybl/share/java/apache-jena-libs-3.0.0.pom

$ mvn package
$ ls distribution-core/target/powsybl/share/java/apache-jena-libs-3.0.0.pom

No spurious file in the distribution. Right now the share/java only has jars and this one file.

cleanup this unused file.
We need to understand why this file is present, and prevent other types of spurious files to end up in the distribution. Is it because of a badly configured dependency in our poms ? or is it because itools-packager should filter these out ?

  - PowSyBl Version: 2.4.0

frannicot:

 Fix #753

Bug fix

#753 

No more spurious share/java/apache-jena-libs-3.0.0.pom file in the distribution
JAR files only in share/java/ directory

No


mathbagu:
@jonenst do you think this fix is OK? In [Jena documentation](https://jena.apache.org/download/maven.html#specifying-jena-as-a-dependency), they advice to add jena-libs as dependency like it was done before.

Maybe the fix should be done in the itools packager, to filter the type of dependency to not include pom file.

jonenst:
In the jena docs, they also say:
"Note the use of <type>pom</type> above. This does not work in all tools. An alternative is to depend on jena-tdb, which will pull in the other artifacts."

I'm the who advised frannicot to use this method instead of modifying itools-packager. It seemed simpler than thinking about whether we want to do a whitelist (only allow jars)  or a blacklist (reject poms)

PR #1087, opened 2019-12-13
CI github actions: cache .m2 for speed and stability

jonenst:
Looking at several builds, it looks like this is a 1 minute improvement on average on speed. It should also improve stability because sometimes maven downloads fail.

NO

Feature

maven dependencies are downloaded from the internet every time

maven dependencies are cached (exec 150Mb wildfly distribution)

**Other information**:
This could bring a lesser improvement if we enabled multithreaded builds

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -19,6 +19,16 @@ jobs:
         with:
           java-version: 1.8
 
+      - uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: ${{ matrix.os }}-maven
+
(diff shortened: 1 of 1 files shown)


mathbagu:
Before doing any optimization, shouldn’t you move AFS modules in a specific repository as discussed/decided?
For release branch, I afraid that caching can have side effect. How to be sure I download the release from the internet, and not by a bad build from a repository?

geofjamg:
I would close this issue. There is no performance issue anymore with core repo and this optimization could have drawbacks.

PR #1094, opened 2019-12-20
Add getNodes(int node) in NodeBreakerView

yichen88:

#1031 

feature

**Other information**:

### DIFF
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/VoltageLevel.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/VoltageLevel.java
@@ -7,7 +7,6 @@
 package com.powsybl.iidm.network;
 
 import com.powsybl.iidm.network.util.ShortIdDictionary;
-import org.apache.commons.lang3.NotImplementedException;
 
 import java.io.IOException;
 import java.io.PrintStream;
@@ -371,9 +370,7 @@ public interface VoltageLevel extends Container<VoltageLevel> {
         /**
          * Get the list of nodes in bus.
          */
-        default int[] getNodes(String busId) {
-            throw new NotImplementedException("Not implemented yet");
-        }
+        int[] getNodes(String busId);
 
         /**
          * Set the number of node.
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/VoltageLevel.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/VoltageLevel.java
@@ -7,6 +7,7 @@
 package com.powsybl.iidm.network;
 
 import com.powsybl.iidm.network.util.ShortIdDictionary;
+import org.apache.commons.lang3.NotImplementedException;
 
 import java.io.IOException;
 import java.io.PrintStream;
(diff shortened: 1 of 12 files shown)


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/VoltageLevel.java
```diff
@@ -367,6 +368,13 @@
          */
         int[] getNodes();
 
+        /**
+         * Get the list of nodes in bus.
+         */
+        default int[] getNodes(String busId) {
```
mathbagu: It could be strange to mention busId in the NodeBreakerView as there is no bus there. Do you add this method in the correct view?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusExt.java
```diff
@@ -7,19 +7,33 @@
...
+     */
     Stream<TerminalExt> getConnectedTerminalStream();
 
+    /**
+     * Get nodes in this bus.
+     * @return Returns the nodes in array.
+     */
+    default int[] getNodes() {
+        throw new NotImplementedException("Not implemented yet");
+    }
```
mathbagu: This method should probably not implemented like this here: this is the implementation, you should keep it unimplemented?

mathbagu:
I propose to change a little bit this PR to avoid introducing concepts in VoltageLevel views that have no sense (bus in Node/Breaker view, or nodes in Bus/Breaker or Bus Views).
If you move both `getConnectedTerminalStream` and `getConnectedTerminals` methods from BusExt (implementation) to Bus (API), you will be able to get the nodes from a bus.
```
        int nodes[] = bus.getConnectedTerminalStream()
                .mapToInt(t -> t.getNodeBreakerView().getNode())
                .distinct()
                .toArray();
```

Thus, it's possible to do a for-loop over the buses of the Bus/Breaker or the Bus views.
```

**Comment by yichen88** [comment]
> I propose to change a little bit this PR to avoid introducing concepts in VoltageLevel views that have no sense (bus in Node/Breaker view, or nodes in Bus/Breaker or Bus Views).
> If you move both `getConnectedTerminalStream` and `getConnectedTerminals` methods from BusExt (implementation) to Bus (API), you will be able to get the nodes from a bus.
> 
> ```
>         int nodes[] = bus.getConnectedTerminalStream()
>                 .mapToInt(t -> t.getNodeBreakerView().getNode())
>                 .distinct()
>                 .toArray();
> ```
> 
> Thus, it's possible to do a for-loop over the buses of the Bus/Breaker or the Bus views.
> 
> @yichen88 Could you tell me if it works regarding to your needs?

No, because i need those switch's node too. Via terminal, i could only get node of Load/Generator etc...

**Comment by yichen88** [comment]
Could we change this implements into 2 steps?
1. `int[] getNodes(int node)` returns connected nodes.
```java
assertArrayEquals(new int[]{1, 6, 5}, vl.getNodeBreakerView().getNodes(1));
        assertArrayEquals(new int[]{1, 6, 5}, vl.getNodeBreakerView().getNodes(6));
        assertArrayEquals(new int[]{1, 6, 5}, vl.getNodeBreakerView().getNodes(5));
```
2. A new utils class `VoltageLevels` to find which bus is the node belongs to.

mathbagu:
Another proposition have been made in the branch [voltage-by-bus](https://github.com/powsybl/powsybl-core/compare/voltages-by-bus)

MioRtia:
Resolved by PR #1170

PR #1096, opened 2019-12-26
Fix two issues in iterative proportional scalable scaling algorithm.

murgeyseb:

Bug fix

Currently, iterative scaling algorithm on proportional scalable does not work as expected and can even throw assertion errors or consider non saturated scalables as saturated

Iterative scaling algorithm works as expected


murgeyseb:
> I understand the fixes, but if the current unit test did not allow to identify such bugs, we clearly need to update them.
> Probably for instance having a test with an "asked" value different from 100, and a unit test with more than 2 iterations.

I added a test for the two iterations process and with a total value > 100. It should cover these issues now

PR #590, opened 2019-01-30
[WIP]Add concurrent SARBuilder

yichen88:
**Please check if the PR fulfills these requirements** *(please use `'[x]'` to check the checkboxes, or submit the PR and then click the checkboxes)*
- [] The commit message follows our guidelines
- [] Tests for the changes have been added (for bug fixes / features)
- [] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** *(Bug fix, feature, docs update, ...)*
Bug.

**What is the current behavior?** *(You can also link to an open issue here)*
The security analysis result builder is not safe for concurrent use. In particular, the way it used in `SecurityAnalysisImpl` may malfunction as soon as contingencies are computed in parallel.

**What is the new behavior (if this is a feature change)?**
The builder is safe for concurrent use (addition of results for multiple contingencies in parallel).

**Does this PR introduce a breaking change or deprecate an API?** *If yes, check the following:*
- [ ] The *Breaking Change* or *Deprecated* label has been added
- [ ] The migration guide has been updated in the github wiki *(What changes might users need to make in their application due to this PR?)*

PR #1097, opened 2020-01-02
[WIP]Fix SecurityAnalysisImpl in multi-thread.

yichen88:
1. cloneVariant() should not be executed in parallel.
2. One result builder per contingency, at the end do merger.
3. Fix bug: clone working variant instead of init variant.

#590 

Bug fix

Throws exception using result builder in parallel.

**Other information**:

### DIFF
--- a/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisImpl.java
+++ b/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisImpl.java
@@ -10,7 +10,6 @@ import com.powsybl.computation.ComputationManager;
 import com.powsybl.contingency.ContingenciesProvider;
 import com.powsybl.contingency.Contingency;
 import com.powsybl.iidm.network.Network;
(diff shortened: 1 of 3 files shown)


yichen88:
> After this pull request in its current state:
> 
> * The checkAll() method is called N times for the precontingency state whereas before it was called only once. It think we can and should avoid it.
> * The contingencies modifications are all applied sequentially and synchronously. Why this change ? What if one contingency is expensive to apply (I don't think this can happen, can it )?

Because modifications happens in an tmp variant while clone() variant could not be invoked concurrently.
> * @mathbagu's comment, it's true that there is a problem with memory for big networks and big contingency lists, it should be fixed.

mathbagu:
> Because modifications happens in an tmp variant while clone() variant could not be invoked concurrently.

Not exactly: cloning a variant is not thread-safe if you create / delete a variant. But it's allowed to replace and existing variant. What we propose here is:
- creating a variant per worker
- for each contingency
  - clone (replace) the current variant by the initial state
  - applying the contingency
  - run a load-flow
- remove all variants

PR #1119, opened 2020-01-17
Allow to choose the name of the default iidm implementation instead o…

jonenst:
…f the hardcoded 'Default'

This allows other implementations to be used by code that is using the the iidm api without specifying the implementation. For examples, all the tests in iidm-impl.

NO

Feature

The iidm implementation is chosen by serviceLoader.load.filter(name == "Default")

The iidm implementation is chosen by 
String defaultName = ... ;//from platformConfig iidm.default
serviceLoader.load.filter(name == defaultName)

If the iidm is module is not present, "Default" is used
if the iidm module is present and the "default" property is not present, an exception is thrown
if the iidm module is present and the default value is set to a missing implementation in the classpath an exception is thrown, 

No

### DIFF
--- a/commons/src/main/java/com/powsybl/commons/config/PlatformConfigNamedProvider.java
+++ b/commons/src/main/java/com/powsybl/commons/config/PlatformConfigNamedProvider.java
@@ -6,12 +6,17 @@
  */
 package com.powsybl.commons.config;
 
+import com.google.common.collect.Lists;
 import com.powsybl.commons.PowsyblException;
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.ServiceLoader;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 /**
@@ -57,6 +62,8 @@ public interface PlatformConfigNamedProvider {
         private static final String DEFAULT_SERVICE_IMPL_NAME_PROPERTY = "default-impl-name";
         private static final String LEGACY_SERVICE_IMPL_NAME_PROPERTY = "default";
 
+        private static final Map<Class<? extends PlatformConfigNamedProvider>, List<? extends PlatformConfigNamedProvider>> PROVIDERS = new ConcurrentHashMap<>();
+
         /**
          * Find the default provider configured in the standard field of
          * {@code moduleName} in {@code platformConfig} among the {@code providers}
@@ -65,9 +72,10 @@ public interface PlatformConfigNamedProvider {
          * @return the provider
          */
         public static <T extends PlatformConfigNamedProvider> T findDefault(String moduleName,
-                List<T> providers, PlatformConfig platformConfig) {
-            return find(null, moduleName, Arrays.asList(DEFAULT_SERVICE_IMPL_NAME_PROPERTY),
-                    providers, platformConfig);
+                Class<T> clazz, PlatformConfig platformConfig) {
+            return find(null, moduleName,
+                    Arrays.asList(DEFAULT_SERVICE_IMPL_NAME_PROPERTY), clazz,
+                    platformConfig);
         }
 
         /**
@@ -77,9 +85,10 @@ public interface PlatformConfigNamedProvider {
          * @return the provider
          */
         public static <T extends PlatformConfigNamedProvider> T find(String name, String moduleName,
-                List<T> providers, PlatformConfig platformConfig) {
-            return find(name, moduleName, Arrays.asList(DEFAULT_SERVICE_IMPL_NAME_PROPERTY),
-                    providers, platformConfig);
+                Class<T> clazz, PlatformConfig platformConfig) {
+            return find(name, moduleName,
+                    Arrays.asList(DEFAULT_SERVICE_IMPL_NAME_PROPERTY), clazz,
+                    platformConfig);
         }
 
         /**
@@ -93,9 +102,10 @@ public interface PlatformConfigNamedProvider {
          */
         @Deprecated
         public static <T extends PlatformConfigNamedProvider> T findDefaultBackwardsCompatible(
-                String moduleName, List<T> providers, PlatformConfig platformConfig) {
-            return find(null, moduleName, Arrays.asList(DEFAULT_SERVICE_IMPL_NAME_PROPERTY,
-                    LEGACY_SERVICE_IMPL_NAME_PROPERTY), providers, platformConfig);
+                String moduleName, Class<T> clazz, PlatformConfig platformConfig) {
+            return find(null, moduleName,
+                    Arrays.asList(DEFAULT_SERVICE_IMPL_NAME_PROPERTY,
+                    LEGACY_SERVICE_IMPL_NAME_PROPERTY), clazz, platformConfig);
         }
 
         /**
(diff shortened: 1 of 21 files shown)


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/NetworkFactory.java
```diff
@@ -47,7 +48,11 @@ static NetworkFactory find(String name) {
      * @return default {@code NetworkFactory} instance.
      */
     static NetworkFactory findDefault() {
-        return find(NetworkFactoryConstants.DEFAULT);
+        String defaultName = PlatformConfig.defaultConfig()
+                .getOptionalModuleConfig("iidm")
+                .map(module -> module.getStringProperty("default"))
```
MioRtia: Should you name the property `default`? Isn't `impl` more understandable for example?

jonenst: I chose this to be consistent with the loadflow.default configuration

geofjamg: As we known that IIDM "branding" should one day disappear and just call it "the powsybl network model", I would have named it:

```yaml
network:
    default-impl-name: xxx
```

jonenst:
```yaml
network-model:
   default: xxx
```
?

If we use default-impl-name, we should update the loadflow.default config for consistency

MioRtia: I actually prefer @geofjamg's proposition, given than we will then update the `loadflow` config names to be consistent.

commons/src/main/java/com/powsybl/commons/config/PlatformConfigNamedProvider.java
```diff
@@ -0,0 +1,197 @@
...
+    String getName();
+
+    /**
+     * Get the Provider name used for identifying this provider in the
+     * PlatformConfig. Defaults to getName(). Override this method only getName() is
+     * already implemented and returns the wrong name.
+     *
+     * @return the name
+     */
+    default String getPlatformConfigName() {
```
MioRtia: Not sure I understand how this method is different from `getName()`? It is never overriden in the code, isn't it? In which case would it be overriden?

commons/src/main/java/com/powsybl/commons/config/PlatformConfigNamedProvider.java
```diff
@@ -0,0 +1,197 @@
...
+    /**
+     * Get the name.
+     *
+     * @return the name
+     */
+    String getName();
+
+    /**
+     * Get the Provider name used for identifying this provider in the
+     * PlatformConfig. Defaults to getName(). Override this method only getName() is
```
MioRtia: Small typo: 
> Override this method only **if** getName() is

commons/src/main/java/com/powsybl/commons/config/PlatformConfigNamedProvider.java
jonenst: This is done to automatically reuse the implement of "getName" methods from other interfaces (e.g. Versionable ) in most cases, but still allow to have different names in case you need it. It is not currently needed in our code now.

commons/src/main/java/com/powsybl/commons/config/PlatformConfigNamedProvider.java
```diff
@@ -0,0 +1,197 @@
...
+         *
+         * @return the provider
+         */
+        @Deprecated
+        public static <T extends PlatformConfigNamedProvider> T findBackwardsCompatible(String name,
+                String moduleName, Class<T> clazz, PlatformConfig platformConfig) {
+            return find(name, moduleName,
+                    Arrays.asList(DEFAULT_SERVICE_IMPL_NAME_PROPERTY,
+                            LEGACY_SERVICE_IMPL_NAME_PROPERTY),
+                    clazz, platformConfig);
```
mathbagu: Coding style: Indent this code like in the previous method

commons/src/main/java/com/powsybl/commons/config/PlatformConfigNamedProvider.java
```diff
@@ -0,0 +1,197 @@
...
+        }
+
+        private static Optional<String> getOptionalFirstProperty(ModuleConfig moduleConfig,
+                List<String> propertyNames) {
+            return propertyNames.stream().map(moduleConfig::getOptionalStringProperty)
+                    .filter(Optional::isPresent).map(Optional::get).findFirst();
```
mathbagu: Coding style: To increase the readability, we should write 1 stream function per line

commons/src/main/java/com/powsybl/commons/config/PlatformConfigNamedProvider.java
```diff
@@ -0,0 +1,197 @@
...
+                k -> Lists.newArrayList(ServiceLoader.load(clazz)));
+            return find(name, moduleName, propertyNames, providers, platformConfig);
+        }
+
+        // package private for tests
+        static <T extends PlatformConfigNamedProvider> T find(String name,
+                String moduleName, List<String> propertyNames, List<T> providers,
+                PlatformConfig platformConfig) {
+            Objects.requireNonNull(providers);
+            Objects.requireNonNull(platformConfig);
```
mathbagu: We discussed about this few days ago... You propose to remove this assertion, because, if the name is not null we don't need platformConfig. Did you change your mind?

commons/src/main/java/com/powsybl/commons/config/PlatformConfigNamedProvider.java
```diff
@@ -0,0 +1,197 @@
...
+                // no information to select the implementation but only one provider, so we can
+                // use it by default
+                // (that is be the most common use case)
+                provider = providers.get(0);
+            } else {
+                if (providers.size() > 1 && finalName == null) {
+                    // several providers and no information to select which one to choose, we can
+                    // only throw
+                    // an exception
+                    List<String> loadFlowNames = providers.stream()
```
geofjamg: Should be renamed?

commons/src/main/java/com/powsybl/commons/config/PlatformConfigNamedProvider.java
jonenst: Yes, I find it more robust to always require non null. This avoids problems when the code "works" when the name is not null, but suddenly name becomes null and it doesn't work anymore.

We can rework the API in the future in a separate PR to not give this API where parameters are used depending on whether other parameters are null

commons/src/main/java/com/powsybl/commons/config/PlatformConfigNamedProvider.java
jonenst: done. I used clazz.getSimpleName() in the message because this is expected to be used on our classes which have very specific names (e.g. LoadFlowProvider)

commons/src/main/java/com/powsybl/commons/config/PlatformConfigNamedProvider.java
jonenst: ok

commons/src/main/java/com/powsybl/commons/config/PlatformConfigNamedProvider.java
jonenst: ok

commons/src/main/java/com/powsybl/commons/config/PlatformConfigNamedProvider.java
jonenst: done

commons/src/main/java/com/powsybl/commons/config/PlatformConfigNamedProvider.java
MioRtia: ok

PR #1125, opened 2020-01-20
Tck

jonenst:

NO

Feature

iidm-impl tests can't be reused

iidm-impl tests can be reused

NO

**Other information**:
The tests are extracted to a new module iidm/iidm-tck. tck means Technology Compatibility Kit. The name "tck" is used by java itself and by know java projects (e.g. https://github.com/reactive-streams/reactive-streams-jvm/tree/master/tck ).
In a second pass, it could be valuable to choose more precisely which tests that originally came from iidm-impl should be in this tck.  Here all the tests that could be written to use only the iidm-api have been moved without much more thought.

This PR requires #1119 to be run in other projects


iidm/iidm-tck/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>powsybl-iidm-test</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>compile</scope>
```
mathbagu: Is it necessary to specify this scope? it's the default value, isn't it?

iidm/iidm-tck/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-all</artifactId>
+            <scope>compile</scope>
```
mathbagu: Same remark here

iidm/iidm-tck/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
```
mathbagu: Is the default scope OK there?

iidm/iidm-tck/src/main/java/com/powsybl/iidm/network/tck/AbstractNodeBreakerInternalConnectionsTest.java
```diff
@@ -13,29 +13,25 @@
...
-        assertEquals(3, internalConnections.get(2).getNode2());
-        assertEquals(5, internalConnections.get(3).getNode1());
-        assertEquals(2, internalConnections.get(3).getNode2());
-        assertEquals(9, internalConnections.get(4).getNode1());
-        assertEquals(2, internalConnections.get(4).getNode2());
-        assertEquals(8, internalConnections.get(5).getNode1());
-        assertEquals(1, internalConnections.get(5).getNode2());
+        int[] expecteds1 = new int[] {7, 6, 4, 5, 9, 8 };
+        int[] expecteds2 = new int[] {0, 3, 3, 2, 2, 1 };
+        for (int i = 0; i < 5; i++) {
```
mathbagu: i < 6?

iidm/iidm-tck/src/main/java/com/powsybl/iidm/network/tck/AbstractTapChangerTest.java
```diff
@@ -220,18 +222,7 @@ public void testTapChangerSetterGetterInMultiVariants() {
...
-        assertEquals(1, ratioTapChanger.getTapPosition());
-        assertTrue(ratioTapChanger.isRegulating());
-        assertEquals(10.0, ratioTapChanger.getTargetV(), 0.0);
-        assertEquals(1, ratioTapChangerInLeg2.getTapPosition());
-        assertTrue(ratioTapChangerInLeg2.isRegulating());
-        assertEquals(10.0, ratioTapChangerInLeg2.getTargetV(), 0.0);
-        assertEquals(3, ratioTapChangerInLeg3.getTapPosition());
-        assertFalse(ratioTapChangerInLeg3.isRegulating());
-        assertEquals(11.0, ratioTapChangerInLeg3.getTargetV(), 0.0);
+        assertKnownState(phaseTapChanger, ratioTapChanger, ratioTapChangerInLeg2, ratioTapChangerInLeg3);
```
mathbagu: Could you certify this is equivalent?

iidm/iidm-tck/src/main/java/com/powsybl/iidm/network/tck/AbstractNodeBreakerInternalConnectionsTest.java
jonenst: good catch, ```i < expecteds1.length``` ?

iidm/iidm-tck/src/main/java/com/powsybl/iidm/network/tck/AbstractTapChangerTest.java
jonenst: If I did this correctly, it is equivalent (asserKnownState is juste a newly extracted method to avoid code duplication detected by sonar).

iidm/iidm-tck/pom.xml
jonenst: It's necessary because junit is declared with scope test in the pom of powsybl-core.

Note: I'm not sure it's correct. It seems to work though.

iidm/iidm-tck/src/main/java/com/powsybl/iidm/network/tck/AbstractNodeBreakerInternalConnectionsTest.java
mathbagu: why not

iidm/iidm-tck/pom.xml
jonenst: good catch, it was actually unneeded

PR #1129, opened 2020-01-22
Allow to write IIDM-XML files in previous versions

MioRtia:

No

Feature

IIDM-XML files can only be written in current IIDM-XML version (i.e. 1.1)

IIDM-XML files can now be written in current and previous IIDM-XML versions (i.e. 1.0 and 1.1)

No

**Other information:**
:warning: Network extensions **cannot** for the moment be written in previous versions, it will be implemented in another PR

### DIFF
--- a/iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/export/ExportOptions.java
+++ b/iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/export/ExportOptions.java
@@ -45,15 +45,15 @@ public class ExportOptions extends AbstractOptions<ExportOptions> {
     }
 
     public ExportOptions(boolean withBranchSV, boolean indent, boolean onlyMainCc, TopologyLevel topologyLevel, boolean throwExceptionIfExtensionNotFound) {
+        this(withBranchSV, indent, onlyMainCc, topologyLevel, throwExceptionIfExtensionNotFound, null);
+    }
+
+    public ExportOptions(boolean withBranchSV, boolean indent, boolean onlyMainCc, TopologyLevel topologyLevel, boolean throwExceptionIfExtensionNotFound, String version) {
         this.withBranchSV = withBranchSV;
         this.indent = indent;
         this.onlyMainCc = onlyMainCc;
         this.topologyLevel = Objects.requireNonNull(topologyLevel);
         this.throwExceptionIfExtensionNotFound = throwExceptionIfExtensionNotFound;
-    }
-
-    public ExportOptions(boolean withBranchSV, boolean indent, boolean onlyMainCc, TopologyLevel topologyLevel, boolean throwExceptionIfExtensionNotFound, String version) {
-        this(withBranchSV, indent, onlyMainCc, topologyLevel, throwExceptionIfExtensionNotFound);
         this.version = version;
     }
 
--- a/iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/export/ExportOptions.java
+++ b/iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/export/ExportOptions.java
@@ -39,6 +39,8 @@ public class ExportOptions extends AbstractOptions<ExportOptions> {
 
     private boolean throwExceptionIfExtensionNotFound = false;
 
+    private String version;
+
     public ExportOptions() {
     }
 
@@ -48,6 +50,11 @@ public class ExportOptions extends AbstractOptions<ExportOptions> {
         this.onlyMainCc = onlyMainCc;
         this.topologyLevel = Objects.requireNonNull(topologyLevel);
         this.throwExceptionIfExtensionNotFound = throwExceptionIfExtensionNotFound;
+    }
+
+    public ExportOptions(boolean withBranchSV, boolean indent, boolean onlyMainCc, TopologyLevel topologyLevel, boolean throwExceptionIfExtensionNotFound, String version) {
+        this(withBranchSV, indent, onlyMainCc, topologyLevel, throwExceptionIfExtensionNotFound);
+        this.version = version;
     }
 
     @Override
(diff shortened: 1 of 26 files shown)


iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/export/ExportOptions.java
```diff
@@ -50,6 +52,11 @@ public ExportOptions(boolean withBranchSV, boolean indent, boolean onlyMainCc, T
         this.throwExceptionIfExtensionNotFound = throwExceptionIfExtensionNotFound;
     }
 
+    public ExportOptions(boolean withBranchSV, boolean indent, boolean onlyMainCc, TopologyLevel topologyLevel, boolean throwExceptionIfExtensionNotFound, String version) {
+        this(withBranchSV, indent, onlyMainCc, topologyLevel, throwExceptionIfExtensionNotFound);
+        this.version = version;
+    }
```
mathbagu: Implement this in the opposite way: the constructor with the biggest number of parameter should contain the logic. The other constructor(s) should call it with default values

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractConnectableXml.java
```diff
@@ -153,31 +151,47 @@ public static void readCurrentLimits(Integer index, Supplier<CurrentLimitsAdder>
...
+                        .endTemporaryLimit();
             }
         });
         adder.add();
     }
 
+    /**
+     * @deprecated Use {@link #writeCurrentLimits(Integer, CurrentLimits, XMLStreamWriter, IidmXmlVersion)} instead.
+     */
+    @Deprecated
```
mathbagu: Not sure this should be deprecated: for most people, export network using the current version is OK?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractNetworkXmlContext.java
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * @author Miora Ralambotiana <miora.ralambotiana at rte-france.com>
+ */
+public abstract class AbstractNetworkXmlContext<T> extends AbstractConverterContext<T> {
+
+    private final IidmXmlVersion version;
+
+    AbstractNetworkXmlContext(Anonymizer anonymizer, IidmXmlVersion version) {
+        super(anonymizer);
+        this.version = version;
```
mathbagu: Assert that it's not null

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ThreeWindingsTransformerXml.java
```diff
@@ -22,6 +22,11 @@
 
     static final String ROOT_ELEMENT_NAME = "threeWindingsTransformer";
 
+    private static final String RATIO_TAP_CHANGER_1 = "ratioTapChanger1";
```
mathbagu: Maybe you should also introduce RATIO_TAP_CHANGER_2 and RATIO_TAP_CHANGER_3 even there is no duplication yet

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/util/IidmXmlUtil.java
```diff
@@ -7,17 +7,29 @@
...
+import java.util.Objects;
 
 /**
  * @author Miora Ralambotiana <miora.ralambotiana at rte-france.com>
  */
 public final class IidmXmlUtil {
 
+    public interface IidmXmlRunnable {
+        void run() throws XMLStreamException;
+    }
```
mathbagu: Why don't you use the default Runnable? Maybe for exception handling? In that case, I think we have an UncheckedXmlStreamException

iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/export/ExportOptions.java
```diff
@@ -39,15 +39,22 @@
 
     private boolean throwExceptionIfExtensionNotFound = false;
 
+    private String version;
+
     public ExportOptions() {
     }
 
     public ExportOptions(boolean withBranchSV, boolean indent, boolean onlyMainCc, TopologyLevel topologyLevel, boolean throwExceptionIfExtensionNotFound) {
+        this(withBranchSV, indent, onlyMainCc, topologyLevel, throwExceptionIfExtensionNotFound, null);
```
mathbagu: To be discussed: I'm not sure we should allow null values

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractConnectableXml.java
```diff
@@ -153,31 +151,39 @@ public static void readCurrentLimits(Integer index, Supplier<CurrentLimitsAdder>
...
+                        .setAcceptableDuration(acceptableDuration)
+                        .setValue(value)
+                        .setFictitious(fictitious)
+                        .endTemporaryLimit();
             }
         });
         adder.add();
     }
 
     public static void writeCurrentLimits(Integer index, CurrentLimits limits, XMLStreamWriter writer) throws XMLStreamException {
```
mathbagu: To be discussed: I fact, I have a doubt now... This method is not really public, but private package... maybe you were right yesterday when I asked you to remove the deprecated. I think we should even break the API and make this method private package.

Is this method still used?

iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/export/ExportOptions.java
MioRtia: In my opinion, `null` for version would mean the same as "default" version i.e. the most recent version (which is implemented in `NetworkXml`).

The logic here of using `null` for undefined rather than directly a defined value is that if we use `ExportOptions` for an another serialization of IIDM in the future (IIDM-JSON for example), the most recent version of this serialization does not have to be the same as IIDM-XML most recent version. In this case, keeping it undefined and letting the converter itself assert the defaut version is the best solution I think.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractConnectableXml.java
MioRtia: Currently, it is only used in some extensions relying on `CurrentLimits`. Not sure either if this should remain public or not but I think both positions are valid.

PR #1138, opened 2020-01-31
Support writing IIDM-XML network extensions in previous versions

MioRtia:

Feature

Network extensions can only be serialized in their latest version.

A user can choose the version in which they want to serialize their network extensions.

No


iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/export/ExportOptions.java
```diff
@@ -153,6 +152,18 @@ public ExportOptions setVersion(String version) {
...
+            throw new PowsyblException(extensionName + " is not an extension you have passed in the extensions list to export.");
+        }
+        extensionsVersions.put(extensionName, extensionVersion);
+        return this;
+    }
+
+    public Optional<String> getExtensionVersion(String extensionName) {
+        return Optional.ofNullable(extensionsVersions.get(extensionName));
+    }
+
```
mathbagu: I don't understand how these two methods will be used. Could you explain it to me?

MioRtia: `addExtensionVersion ` is to add a particular **extension's version** in which you want the extension to be serialized e.g. if you want to serialize  `CoordinatedReactiveControl` in its 1.0 version, you can use `options.addExtensionVersion("coordinatedReactiveControl", "1.0");`. It is different from the **IIDM-XML version** which is by default 1.1.

`getExtensionVersion` is to get the extension's version in which you want the extension to be serialized if it has been defined (return an empty `Optional` if it has not been defined).

mathbagu:
Could please add:
- a method in the Context to get the version for an extension (`String getExtensionVersion(String name);`)
- an example of configuration where you specify the version for a specific extension (what happen if the version is not supported by the serializer ?)

commons/src/main/java/com/powsybl/commons/extensions/ExtensionXmlSerializer.java
```diff
@@ -69,4 +70,13 @@ default String getName() {
     default String getVersion() {
         return "1.0";
     }
+
+    /**
+     * Check that a given extension version exists.
+     */
+    default void checkExtensionVersionExists(String extensionVersion) {
```
mathbagu: Maybe replace "exists" by "Supported"?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/NetworkXml.java
```diff
@@ -215,6 +223,27 @@ private static ExtensionXmlSerializer getExtensionXmlSerializer(ExportOptions op
         return extensionXmlSerializer;
     }
 
+    private static String getNamespaceUri(ExtensionXmlSerializer extensionXmlSerializer, ExportOptions options) {
+        return getNamespaceUri(extensionXmlSerializer, options,
+                options.getVersion() == null ? IidmXmlConstants.CURRENT_IIDM_XML_VERSION : IidmXmlVersion.of(options.getVersion(), "."));
+    }
```
mathbagu: Rewrite this method in two lines to make it simpler:
```
String version = options.getVersion() == null ? IidmXmlConstants.CURRENT_IIDM_XML_VERSION : IidmXmlVersion.of(options.getVersion(), ".");
return getNamespaceUri(extensionXmlSerializer, options, version);
```

iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/export/ExportOptions.java
agnesLeroy: Then if for some reason someone calls addExtensionVersion twice with two different versions for the same extension, that extension will be serialized twice, one for each version?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/NetworkXml.java
agnesLeroy: Why is the test done on the version on the IidmXmlVersion here?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/extensions/AbstractVersionableNetworkExtensionXmlSerializer.java
```diff
@@ -25,6 +25,8 @@
  */
 public abstract class AbstractVersionableNetworkExtensionXmlSerializer<T extends Extendable, E extends Extension<T>> implements ExtensionXmlSerializer<T, E> {
 
+    private static final String INCOMPATIBILITY_NETWORK_VERSION_MESSAGE = "IIDM-XML version of network (";
```
agnesLeroy: Why create this variable?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/XMLExporter.java
```diff
@@ -63,6 +67,8 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(XMLExporter.class);
 
+    private static final Supplier<ExtensionProviders<ExtensionXmlSerializer>> EXTENSIONS_SUPPLIER = Suppliers.memoize(() -> ExtensionProviders.createProvider(ExtensionXmlSerializer.class, "network"));
```
agnesLeroy: Why do you use memoize here?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/extensions/AbstractVersionableNetworkExtensionXmlSerializer.java
MioRtia: It is a string used 3 times (error's message), Sonar considers not creating a variable as a code smell.

iidm/iidm-xml-converter/src/test/java/com/powsybl/iidm/xml/IdentifiableExtensionXmlSerializerTest.java
```diff
@@ -191,26 +194,48 @@ public void testTerminalExtension() throws IOException {
...
+        new XMLExporter().export(network, properties, dataSource);
+        try (InputStream is = new ByteArrayInputStream(dataSource.getData(null, "xiidm"))) {
+            assertNotNull(is);
+            compareXml(getClass().getResourceAsStream(getVersionDir(IidmXmlVersion.V_1_1) + "eurostag-tutorial-example1-with-loadMockExt-1_1.xml"),
+                    is);
+        }
+    }
+
     @Test
     public void testThrowErrorUncompatibleExtensionVersion() {
```
agnesLeroy: Uncompatible should be Incompatible

agnesLeroy: What is the difference between the two resource files in this test? Why do you use MemDataSource?

iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/export/ExportOptions.java
MioRtia: Good observation, I didn't think about this case! With the current implementation, the last call to the method is kept... Maybe an exception should be thrown or at least a warning should be written? This case should not happen (the version of an extension is unique for each file/export).

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/XMLExporter.java
MioRtia: This supplier provides in runtime the XML serializers of network extensions available in the dependencies of the module in which it is called. `memoize` allows to use lazy loading: the serializers are only loaded when first called (if never called, they are never loaded and once loaded, they are kept in memory).

iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/export/ExportOptions.java
agnesLeroy: I believe an exception would be better than a warning, since this would correspond to an erroneous use of the feature.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/XMLExporter.java
agnesLeroy: Ok, thank you for the explanation!

iidm/iidm-xml-converter/src/test/java/com/powsybl/iidm/xml/IdentifiableExtensionXmlSerializerTest.java
MioRtia: > Why didn't you use the same name basis as the two other tests below?

For this test specifically, the version of the extension is supported, just not compatible with the given network version.

> What is the difference between the two resource files in this test? Why do you use MemDataSource?

I'm assuming you are talking about the test above. The difference between the two resource files is the version of the `loadMock` extension (1.2 for the imported file and 1.1 for the reference of the exported file). I am using `MemDataSource` because when exporting, I don't want the file to be physically on my machine but to be in memory (which `MemDataSource` allows to do).

MioRtia: Comments have been added in tests to explain more in detail what is done

iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/export/ExportOptions.java
```diff
@@ -153,6 +152,20 @@ public ExportOptions setVersion(String version) {
         return this;
     }
 
+    public ExportOptions addExtensionVersion(String extensionName, String extensionVersion) {
```
mathbagu: Could you add documentation about the usage of this method. I'm sorry but I never remember how and when to use this

iidm/iidm-xml-converter/src/test/java/com/powsybl/iidm/xml/IdentifiableExtensionXmlSerializerTest.java
```diff
@@ -200,17 +203,51 @@ public void testTerminalExtension() throws IOException {
     }
 
     @Test
-    public void testThrowErrorUncompatibleExtensionVersion() {
+    public void testNotLatestVersionTerminalExtension() throws IOException {
+        // import XIIDM file with loadMock v1.2
+        Network network = NetworkXml.read(getClass().getResourceAsStream("/V1_1/eurostag-tutorial-example1-with-loadMockExt-1_2.xml"));
```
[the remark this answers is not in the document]
MioRtia: Yes, I will make the change you suggested and introduce the new method in a new PR to prevent confusion in this one.

PR #1142, opened 2020-02-03
Proposal to fix issue #1136 (regulating terminal is at busbar section)

### LINKED ISSUE #1136: Referred regulating terminals do not exist in exported IIDM XML
Bug

There is a problem related to regulating terminals linked to busbar sections in node-breaker models when exported to IIDM XML using `BUS_BREAKER` or `BUS_BRANCH` topology levels.
The resulting .xiidm file contains a reference to a regulating terminal that is not output to the file. 
The output is correct if the topology level selected for output is `NODE_BREAKER`. 

Start from an IIDM network where there is a voltage regulation that has a regulating terminal pointing to a busbar section (a node-breaker network definition is required). 
Export to IIDM XML using `BUS_BREAKER` or `BUS_BRANCH` topology level.
Trying to load the generated .xiidm file fails because there is a terminal ref for which no actual terminal could be found in the model.
If the network is exported using `NODE_BREAKER` topology level the output is correct, and the network can be loaded correctly from the resulting .xiidm file.

One possible solution would be to change the terminal reference exported to the bus that corresponds to the original busbar section at the topology level that is being exported.

  - PowSyBl Version: 3.2.0-SNAPSHOT.
  - OS Version: *

zamarrenolm:

A proposal to fix #1136 with the following approach:
When exporting a node-breaker network to bus-breaker or bus-branch topology levels, terminal references that point to busbar sections (elements that are only available at node-breaker level) are changed by references to the current bus of the regulating terminal, obtained from the proper topology view.

### DIFF
--- a/iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/TerminalRefXml.java
+++ b/iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/TerminalRefXml.java
@@ -10,6 +10,8 @@ package com.powsybl.iidm.xml;
 import com.powsybl.commons.PowsyblException;
 import com.powsybl.iidm.network.*;
 
+import java.util.concurrent.atomic.AtomicReference;
+
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamWriter;
 
@@ -32,7 +34,17 @@ public final class TerminalRefXml {
             throw new PowsyblException("Oups, terminal ref point to a filtered equipment " + c.getId());
         }
         writer.writeEmptyElement(namespace, elementName);
-        writer.writeAttribute("id", context.getAnonymizer().anonymizeString(c.getId()));
+        String terminalRefId = c.getId();
+        if (c instanceof BusbarSection) {
+            // Found remote terminal that refers to a busbar section
+            // Choose a bus for the terminal ref depending on export topology level
+            if (context.getOptions().getTopologyLevel() == TopologyLevel.BUS_BREAKER) {
+                terminalRefId = ((BusbarSection) c).getTerminal().getBusBreakerView().getConnectableBus().getId();
+            } else if (context.getOptions().getTopologyLevel() == TopologyLevel.BUS_BRANCH) {
+                terminalRefId = ((BusbarSection) c).getTerminal().getBusView().getConnectableBus().getId();
+            }
+        }
+        writer.writeAttribute("id", context.getAnonymizer().anonymizeString(terminalRefId));
         if (c.getTerminals().size() > 1) {
             if (c instanceof Injection) {
                 // nothing to do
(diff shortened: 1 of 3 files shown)


MioRtia:
Just a question to be sure I have understood the issue correctly: in this case, is the CGMES case node breaker or bus branch? If it is node breaker and that the terminalRef is correct in XIIDM export in node breaker, I don't think the correction should be in the CGMES import but in XIIDM converter.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/TerminalRefXml.java
```diff
@@ -32,7 +34,17 @@ public static void writeTerminalRef(Terminal t, NetworkXmlWriterContext context,
...
         writer.writeEmptyElement(namespace, elementName);
-        writer.writeAttribute("id", context.getAnonymizer().anonymizeString(c.getId()));
+        String terminalRefId = c.getId();
+        if (c instanceof BusbarSection) {
+            // Found remote terminal that refers to a busbar section
+            // Choose a bus for the terminal ref depending on export topology level
+            if (context.getOptions().getTopologyLevel() == TopologyLevel.BUS_BREAKER) {
+                terminalRefId = ((BusbarSection) c).getTerminal().getBusBreakerView().getConnectableBus().getId();
+            } else if (context.getOptions().getTopologyLevel() == TopologyLevel.BUS_BRANCH) {
+                terminalRefId = ((BusbarSection) c).getTerminal().getBusView().getConnectableBus().getId();
```
MioRtia: Not sure if we should link it to a bus... Maybe it would be better to link to a connectable in the same bus (is there a case where there is no other connectable in the same bus?). It would be more consistent with how we have handled terminals in the past: they are references for equipment, not to electrical buses.

zamarrenolm: I proposed to use a bus because it is the usual convention when defining regulation points in analysis tools; also, it is the "point" of the network that is more closely related to the original terminal (the busbar section).

In fact when we read the terminal reference, the related terminal is obtained from the first connected terminal available at the bus.

To take into account your suggestion we should just use `findConnectedTerminal` in the `writeTerminalRef` method instead of using it in the `readTerminalRef`.

zamarrenolm: By the way, the `findConnectedTerminal` here is implemented and "complex" (uses the visitor) because there is no function in the API like `getConnectedTerminals`. It is only available in the default implementation, inside the class `AbstractBus`.

I think we should provide this method at the API level. What do you think @mathbagu, @MioRtia  ?

zamarrenolm: In fact it is somewhat dangerous linking the terminal to the "first" connected terminal and leave it there forever. 

What happens to the regulating control if the equipment associated to that terminal is disconnected ? the generator (or the regulating equipment) may still be able to "reach" the bus linked to the original busbar section, but not the terminal that we have selected to "replace" it.

MioRtia: > By the way, the findConnectedTerminal here is implemented and "complex" (uses the visitor) because there is no function in the API like getConnectedTerminals. It is only available in the default implementation, inside the class AbstractBus.
I think we should provide this method at the API level. What do you think @mathbagu, @MioRtia ?

I agree.

> In fact it is somewhat dangerous linking the terminal to the "first" connected terminal and leave it there forever.
What happens to the regulating control if the equipment associated to that terminal is disconnected? The generator (or the regulating equipment) may still be able to "reach" the bus linked to the original busbar section, but not the terminal that we have selected to "replace" it.

You are right but I would rather do that than change the definition of regulation terminal.
However, we may have a broader reflection about how we modelize regulation points. Only using bus would be confusing in node/breaker but maybe we should use nodes/bus according to the view? @mathbagu @geofjamg what do you think?

MioRtia: > However, we may have a broader reflection about how we modelize regulation points. Only using bus would be confusing in node/breaker but maybe we should use nodes/bus according to the view?

That might mean changing the API though...

mathbagu: In the XML, when we refer to a terminal, in our mind we refer to the bus of this terminal. All the IIDM model is built using this and I start to think this is a problem, but the IDs of the nodes are generated, it's impossible to refer to this bus directly.
Depending on the level of details, we shoud refer to a node (even there is no terminal) or a bus... I don't know what is the best way to implement this, but I think we should try to make it better. I think we should refer to a bus by its name in Bus/Breaker VL. For NBVL, we should do it inderectly using the node... and don't use the terminal anymore, specially because there isn't a terminal everywhere, and the regulation point is definitely the bus, not a terminal...

MioRtia: I made the issue #1143 in that sense: it would be an IIDM evolution (in IIDM 1.1 or later). However, I think we should keep the first workaround (finding an equivalent terminal) to solve the issue when exporting in IIDM 1.0. 
I also think that exposing the method `findConnectedTerminal` to API level is a good idea.

geofjamg: Terminal does not really exist in XIIDM even if tag is named terminalRef...

Regulating terminal are exported in XIIDM using a side of an equipment because it was the 
simplest solution for having a stable configuration of regulating terminal.
Unfortunatly it does not work for in this case but for me having a regulating terminal exported has "something which a result of a calculation (like a bus)" would be a severe design issue.
It has to be stable  and as a general rule calculated view should not be exported or referenced in the XIIDM (or at least should not be mandatory and could be discarded without any consequence).
I agree with @mathbagu proposition that depending of node/breaker or bus/breaker topology the terminal could be exported as a reference to a node or a bus. It would be the most robust and consistant (with the rest of the model) solution. 

in node/breaker
```xml
<regulationPoint voltageLevelId="AAA" node="3"/> 
```
or in bus/breaker
```xml
<regulationPoint voltageLevelId="AAA"  busId="TOTO"/>
```

Same in java

```java
...newGenerator()
...
setNode(3)
...
setRegulatingVoltageLevel("AAA")
setRegulationNode(7)
..
.add()
```

geofjamg: > 
> 
> In the XML, when we refer to a terminal, in our mind we refer to the bus of this terminal.

No, we don't refer to a terminal we refer to a side of an equipment, it is totally different. Even if it has some weakness like shown by this issue.

geofjamg: > 
> 
> > By the way, the findConnectedTerminal here is implemented and "complex" (uses the visitor) because there is no function in the API like getConnectedTerminals. It is only available in the default implementation, inside the class AbstractBus.
> > I think we should provide this method at the API level. What do you think @mathbagu, @MioRtia ?
> 
> I agree.
> 
> > In fact it is somewhat dangerous linking the terminal to the "first" connected terminal and leave it there forever.
> > What happens to the regulating control if the equipment associated to that terminal is disconnected? The generator (or the regulating equipment) may still be able to "reach" the bus linked to the original busbar section, but not the terminal that we have selected to "replace" it.
> 
> You are right but I would rather do that than changing the definition of regulation terminal.
> However, we may have a broader reflection about how we modelize regulation points. Only using bus would be confusing in node/breaker but maybe we should use nodes/bus according to the view? @mathbagu @geofjamg what do you think?

Yes!

PR #1148, opened 2020-02-07
Fix calculated/merged bus name

geofjamg:

No

Bug fix

Calculated buses in node/breaker view and merged buses in bus/breaker view are not named correctly. Calculated/merged bus name is based on voltage level id like their ids but are not based on voltage level name.

If voltage level name exists, calculated/merged bus name is derived from it. If voltage level name does not exist, name is set to null.

**Other information**:


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusBreakerVoltageLevel.java
```diff
@@ -254,7 +254,8 @@ private void updateCache() {
                     }, encountered);
                     if (isBusValid(busSet)) {
                         String mergedBusId = BusBreakerVoltageLevel.this.id + "_" + busNum++;
-                        MergedBus mergedBus = new MergedBus(mergedBusId, busSet);
+                        String mergedBusName = BusBreakerVoltageLevel.this.name != null ? BusBreakerVoltageLevel.this.name + "_" + busNum++ : null;
```
mathbagu: It seems you increment the busNum twice: the first time to generate the busId, the second time for the busName.
I think you should use busNum without increment there.

geofjamg: Fixed!

Issue #828, opened 2019-06-11
IIDM extension design issue


geofjamg:
* **Do you want to request a *feature* or report a *bug*?**
A bug or at least a design flaw

* **What is the current behavior?**
Network model has been split in API and implementation part so that we can have different implementations for different purposes but a single API. So processing on top of network api like merging, import, export, simulation could be reused for any implementation.

Unfortunately this has not been done for extensions. 

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

Example to add an XNode extension to a dangling line:
```java
DanglingLine l = ...
l.addExtension(Xnode.class, new Xnode(l, "FR"));
```
So here we directly instantiate the extension and add it the dangling line.
What about if DanglingLine is not coming from in memory default implementation but from a persistent connected to a db one?

* **What is the expected behavior?**
As for network API extension could be split into API and implementations and provide a builder like way to create it in order to hide the instantiation. 

```java
DanglingLine l = ...
Xnode xNode = l.newExtensionAdder(XnodeAdder.class)
   .withCode("FR")
   .add();
```
And depending of impl, correct Adder will be used. For this we could imagine a plugin archi like

```java
interface ExtensionAdder<E extends Extension> {

   E add();
}

interface ExtensionAdderFactory<A extends ExtensionAdder> {

   // infos to link with correct Network impl ????

    A createAdder();
}
```

* **What is the motivation / use case for changing the behavior?**

To allow alternative IIDM implementation to use extensions.

* **Please tell us about your environment:**
  - PowSyBl Version: 2.5.0-SNAPSHOT
  - OS Version: -

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

MioRtia:
I agree with the principle but `ExtensionAdderProvider` strucks me as an odd class name.

MioRtia:
Maybe `ExtensionAdderFactory` can be more understandable than `ExtensionAdderProvider`?

geofjamg:
> Maybe `ExtensionAdderFactory` can be more understandable than `ExtensionAdderProvider`?

Ok for `ExtensionAdderFactory`

jonenst:
What's the relationship between ```l.newExtensionAdder``` and ```ExtensionAdderFactory.createAdder``` ?

geofjamg:
[...]

PR #1152, opened 2020-02-10
Allow to have different implementations of extensions

jonenst:

#828

Feature

It's not possible to have multiple implementations of an extension

it is possible to have multiple implementations of an extension

NO

**Other information**:

### DIFF
--- a/commons/src/main/java/com/powsybl/commons/extensions/AbstractExtensionAdder.java
+++ b/commons/src/main/java/com/powsybl/commons/extensions/AbstractExtensionAdder.java
@@ -1,0 +1,41 @@
+/**
+ * Copyright (c) 2020, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.commons.extensions;
+
+import java.util.Objects;
+
+/**
+ * A base class for implementations of {@link ExtensionAdder} that holds the
+ * extendable to be able build and then add the extension to the extendable.
+ * This class calls {@link #createExtension} that must be overriden by
+ * subclasses to create the extension.
+ *
+ * @author Jon Harper <jon.harper at rte-france.com>
+ */
+public abstract class AbstractExtensionAdder<T extends Extendable<T>, E extends Extension<T>>
+        implements ExtensionAdder<T, E> {
+
+    private final T extendable;
+
+    protected AbstractExtensionAdder(T extendable) {
+        this.extendable = Objects.requireNonNull(extendable);
+    }
+
+    /**
+     * Creates the extension.
+     *
+     * @return the extension
+     */
+    protected abstract E createExtension(T extendable);
+
+    @Override
+    public T add() {
+        extendable.addExtension(getExtensionClass(), createExtension(extendable));
+        return extendable;
+    }
+
+}
--- a/commons/src/main/java/com/powsybl/commons/extensions/Extendable.java
+++ b/commons/src/main/java/com/powsybl/commons/extensions/Extendable.java
@@ -49,4 +49,30 @@ public interface Extendable<O> {
      * @return all extensions associated to this extendable object.
      */
     <E extends Extension<O>> Collection<E> getExtensions();
+
+    /**
+     * Returns a name that is used to find matching {@link ExtensionAdderProvider}s
+     * when selecting implementations of extensions in {@link #newExtension}. This
+     * is meant to be overriden by extendables when multiple implementations exist.
+     *
+     * @return the name
+     */
+    default String getImplementationName() {
+        return "Default";
+    }
+
+    /**
+     * Returns an extensionAdder to build and add an extension for this extendable.
+     *
+     * The extension implementation is selected at runtime based on matching the
+     * {@link #getImplementationName} of this extendable to the
+     * {@link ExtensionAdderProvider#getImplementationName} of a provider.
+     * Implementations are loaded with java's {@link java.util.ServiceLoader} using
+     * the ExtensionAdderProvider interface.
+     *
+     * @param type The interface of the ExtensionAdder
+     * @return the adder
+     */
+    <E extends Extension<O>, B extends ExtensionAdder<O, E>> B newExtension(Class<B> type);
+
 }
--- a/commons/src/main/java/com/powsybl/commons/extensions/ExtensionAdderProvider.java
+++ b/commons/src/main/java/com/powsybl/commons/extensions/ExtensionAdderProvider.java
@@ -1,0 +1,48 @@
+/**
+ * Copyright (c) 2020, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.commons.extensions;
+
+/**
+ * A provider used through java's {@link java.util.ServiceLoader}. It will
+ * provide an {@link ExtensionAdder} to add an extension to an extendable.
+ * {@link #getImplementationName} is used to find providers corresponding to the
+ * implementation of the {@link Extendable}. {@link #getAdderClass} is used to
+ * specify the adder class.
+ *
+ * @author Jon Harper <jon.harper at rte-france.com>
+ *
+ * @param <T> The extendable
+ * @param <E> The extension
+ * @param <B> The extensionBuilder
+ */
+public interface ExtensionAdderProvider<T extends Extendable<T>, E extends Extension<T>, B extends ExtensionAdder<T, E>> {
+
+    /**
+     * Returns a name that is used to select this provider when searching for
+     * implementations of extension builders in {@link Extendable#newExtension}.
+     *
+     * @return the name
+     */
+    String getImplementationName();
+
+    /**
+     * Returns the builder class provided by this Provider.
+     *
+     * @return the class
+     */
+    // Class<? super B> to allow providers of generic builders to return the class
+    // of the raw type. This has the bad effect of allowing the super classes but we
+    // consider the trade-off acceptable.
+    Class<? super B> getAdderClass();
+
+    /**
+     * returns an new empty ExtensionAdder for this extendable.
+     *
+     * @param extendable the extendable on which the adder will add the extension
+     */
+    B newAdder(T extendable);
+}
--- a/commons/src/main/java/com/powsybl/commons/extensions/ExtensionAdderProviders.java
+++ b/commons/src/main/java/com/powsybl/commons/extensions/ExtensionAdderProviders.java
@@ -1,0 +1,95 @@
+/**
+ * Copyright (c) 2020, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.commons.extensions;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.ServiceLoader;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+import org.apache.commons.lang3.tuple.Pair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Suppliers;
+import com.powsybl.commons.PowsyblException;
+
+/**
+ * A utility class to help finding providers using ServiceLoader.
+ *
+ * @author Jon Harper <jon.harper at rte-france.com>
+ *
+ */
+//Don't bother with generics because serviceloader doesn't return them
+//and we put them in a cache where we can't propagate the generic types.
+@SuppressWarnings("rawtypes")
+public final class ExtensionAdderProviders {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ExtensionAdderProviders.class);
+
+    private static final Supplier<ConcurrentMap<String, List<ExtensionAdderProvider>>> ADDER_PROVIDERS = Suppliers
+            .memoize(() -> groupProvidersByName(ServiceLoader.load(ExtensionAdderProvider.class)))::get;
+
+    private static final ConcurrentMap<Pair<String, Class>, ExtensionAdderProvider> CACHE = new ConcurrentHashMap<>();
+
+    //package private for tests
+    static ConcurrentMap<String, List<ExtensionAdderProvider>> groupProvidersByName(Iterable<ExtensionAdderProvider> i) {
+        return StreamSupport.stream(i.spliterator(), false).collect(Collectors
+                .groupingByConcurrent(ExtensionAdderProvider::getImplementationName));
+    }
+
+    private ExtensionAdderProviders() {
+    }
+
+    private static <O, E extends Extension<O>, B extends ExtensionAdder<O, E>> ExtensionAdderProvider findProvider(
+            String name, Class<B> type,
+            ConcurrentMap<String, List<ExtensionAdderProvider>> providersMap) {
+
+        List<ExtensionAdderProvider> providersForName = providersMap.get(name);
+        if (providersForName == null) {
+            providersForName = Collections.emptyList();
+        }
+        List<ExtensionAdderProvider> providers = providersForName.stream()
+                .filter(s -> type.isAssignableFrom(s.getAdderClass()))
+                .collect(Collectors.toList());
+
+        if (providers.isEmpty()) {
+            LOGGER.error(
+                    "ExtensionAdderProvider not found for ExtensionAdder {} for implementation {}",
+                    type.getSimpleName(), name);
+            throw new PowsyblException("ExtensionAdderProvider not found");
+        }
+
+        if (providers.size() > 1) {
+            LOGGER.error(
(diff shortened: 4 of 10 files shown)


jonenst:
Coverage is done is #1153

commons/src/main/java/com/powsybl/commons/extensions/AbstractExtensionAdder.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @author Jon Harper <jon.harper at rte-france.com>
+ */
+//TODO can we use T extends Extendable<T> here ? Then we can remove the cast in add()
+public abstract class AbstractExtensionAdder<T, E extends Extension<T>> implements ExtensionAdder<T, E> {
+
+    protected T extendable;
+
+    protected AbstractExtensionAdder(T extendable) {
+        this.extendable = extendable;
```
mathbagu: 2 questions: do we allow null values and is extendable final?

commons/src/main/java/com/powsybl/commons/extensions/Extendable.java
```diff
@@ -49,4 +49,39 @@
...
+    /**
+     * Returns a name that is used to find matching {@link ExtensionAdderProvider}s
+     * when selecting implementations of extensions in {@link #newExtension}. This
+     * is meant to be overriden by extendables when multiple implementations exist.
+     *
+     * @return the name
+     */
+    default String getImplementationName() {
+        return "Default";
+    }
```
mathbagu: It's strange to have this in `commons`. Is it specific to IIDM extensions?

commons/src/main/java/com/powsybl/commons/extensions/Extendable.java
```diff
@@ -49,4 +49,39 @@
...
+    /**
+     * Returns an extensionAdder to build and add an extension for this extendable.
+     *
+     * The extension implementation is selected at runtime based on matching the
+     * {@link #getImplementationName} of this extendable to the
+     * {@link ExtensionAdderProvider#getImplementationName} of a provider.
+     * Implementations are loaded with java's {@link java.util.ServiceLoader} using
+     * the ExtensionAdderProvider interface.
+     *
+     * This is not meant to be overriden.
```
mathbagu: Could you declare this method final?

commons/src/main/java/com/powsybl/commons/extensions/ExtensionAdderProviders.java
```diff
@@ -0,0 +1,73 @@
+package com.powsybl.commons.extensions;
```
mathbagu: License is missing

commons/src/main/java/com/powsybl/commons/extensions/ExtensionAdderProviders.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+//Don't bother with generics because serviceloader doesn't return them
+//and we put them in a cache where we can't propagate the generic types.
+@SuppressWarnings("rawtypes")
+final class ExtensionAdderProviders {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ExtensionAdderProviders.class);
+
+    private ExtensionAdderProviders() {
+    }
```
mathbagu: Sonar should warn, but static variables should be declared before this constructor

commons/src/main/java/com/powsybl/commons/extensions/Extendable.java
jonenst: No, you can't have a final method in an interface

commons/src/main/java/com/powsybl/commons/extensions/Extendable.java
jonenst: Not particularly. The getImplementationName method is only compared to the getImplementationName method of an ExtensionAdderProvider (when you call Extendable#newExtensionAdder). It's never compared to the NetworkFactoryServiceImpl from the iidm default implementation (although it happens to be the same constant "Default")

This is used when you have multiple implementations of an extendable, and you want the matching implementation of the extension. So this means that for all existing extendables today where we want multiple implementations, we can just create a Provider with the name "Default". But if we want we can also give the existing extendable an implementation name and create a Provider with this name.  For new extendable implementations we have to override and choose a new name (and use it in their providers).

commons/src/main/java/com/powsybl/commons/extensions/AbstractExtensionAdder.java
jonenst: There is no problem with forbidding null values (this is always called by doing extendable.newExtensionAdder(..)) so the extendable is never null.
We can also make the extendable final, even private if we want (and pass it to the createExtension method.

You think we should do these changes ?

commons/src/main/java/com/powsybl/commons/extensions/ExtensionAdderProviders.java
jonenst: Oops good catch

jonenst:
Will add a few tests in this PR by extracting the static state of ExtendableProviders

commons/src/main/java/com/powsybl/commons/extensions/ExtensionAdderProviders.java
jonenst: done

jonenst:
I added tests in the same module as the code instead of relying on cross module testing. Also this improves the test coverage.

One problem with the current design is that it allows a compilation (with a warning) of the following code:
```java
public void bad(VoltageLevel vl) {
  vl.newExtension(ConnectablePositionBuilder.class).add();//bad: compiles but shouldn't
}
```

The previous API was better (although not perfect)
```java
public void bad(VoltageLevel vl) {
  vl.addExtension(ConnectablePosition.class, new ConnectablePosition<>());//good: compile error
  vl.addExtension(ConnectablePosition.class, new ConnectablePosition());//bad, compiles but shouldn't. But you are using raw types 
}
```

jonenst:
The previous API also allowed to compile the following:
```
        vl.getExtension(ConnectablePosition.class); // should not compile; there is a warning though
```

I think we have to use the same trick as Collections.emptyList() vs Collections.EMPTY_LIST to allow the type checker to check the generic types

jonenst:
@mathbagu What do you think of my proposed "clazz" static method ? It fixes the holes in the previous and new APIS.

It is only needed for generic extensions and generic extension adders, but should we say that every extension and extension adder should implement it ?

jonenst:
We can also discuss to find a better solution to the asymetry introduced by the new API:
```java
// before it was symmetric:
   extendable.addExtension(ExtensionAdder.class, ...);
   extendable.getExtension(Extension.class);

//now it is assymetric
  extendable.newExtension(ExtensionAdder.class);
  extendable.getExtension(Extension.class);
```

commons/src/main/java/com/powsybl/commons/extensions/Extendable.java
mathbagu: But you agree that we do not want someone to overwrite it?

commons/src/main/java/com/powsybl/commons/extensions/AbstractExtensionAdder.java
mathbagu: I think it's cleaner if you declare it `private`, `final` and if you assert the extendable is not null

commons/src/main/java/com/powsybl/commons/extensions/AbstractExtensionAdder.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * A base class for implementations of {@link ExtensionAdder} that holds the
+ * extendable to be able build and then add the extension to the extendable.
+ * This class calls {@link Extendable#createExtension} that must be overriden by
+ * subclasses to create the extension.
+ *
+ * @author Jon Harper <jon.harper at rte-france.com>
+ */
+//TODO can we use T extends Extendable<T> here ? Then we can remove the cast in add()
```
mathbagu: Make a decision: if you can, do it

commons/src/main/java/com/powsybl/commons/extensions/ExtensionAdderProvider.java
```diff
@@ -0,0 +1,49 @@
...
+ * implementation of the {@link Extendable}. {@link #getAdderClass} is used to
+ * specify the adder class.
+ *
+ * @author Jon Harper <jon.harper at rte-france.com>
+ *
+ * @param <T> The extendable
+ * @param <E> The extension
+ * @param <B> The extensionBuilder
+ */
+//TODO can we use T extends Extendable<T> here ? It would make calling newAdder safer.
```
mathbagu: Same remark

commons/src/test/java/com/powsybl/commons/extensions/ExtensionAdderProvidersTest.java
```diff
@@ -0,0 +1,355 @@
...
+    private static class SimpleExtensionImpl extends AbstractExtension<SimpleExtendable>
+            implements SimpleExtension {
+    }
+
+    private static class SimpleExtensionImpl2 extends AbstractExtension<SimpleExtendable>
+            implements SimpleExtension {
+    }
+
+    private static interface SimpleExtensionAdder extends ExtensionAdder<SimpleExtendable, SimpleExtension> {
+        default Class<SimpleExtension> getExtensionClass() {
```
mathbagu: It's dirty but I assume we can write dirty code in tests... Default should not be used to avoid duplications

jonenst:
> The previous API also allowed to compile the following:
> 
> ```java
>         vl.getExtension(ConnectablePosition.class); // should not compile; there is a warning though
> ```
> 
> I think we have to use the same trick as Collections.emptyList() vs Collections.EMPTY_LIST to allow the type checker to check the generic types


This is only in eclipse. Using maven (ie javac) or intellij, we get the expected compile error.

commons/src/main/java/com/powsybl/commons/extensions/AbstractExtensionAdder.java
```diff
@@ -0,0 +1,41 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.commons.extensions;
+
+import java.util.Objects;
+
+/**
+ * A base class for implementations of {@link ExtensionAdder} that holds the
+ * extendable to be able build and then add the extension to the extendable.
+ * This class calls {@link Extendable#createExtension} that must be overriden by
```
geofjamg: #createExtension ?

commons/src/test/java/com/powsybl/commons/extensions/ExtensionAdderProvidersTest.java
```diff
@@ -0,0 +1,355 @@
...
+
+import org.apache.commons.lang3.tuple.Pair;
+import org.junit.Test;
+
+import com.google.auto.service.AutoService;
+import com.powsybl.commons.PowsyblException;
+
+public class ExtensionAdderProvidersTest {
+
+    private static interface SimpleExtendable extends Extendable<SimpleExtendable> {
```
geofjamg: useless static?

commons/src/test/java/com/powsybl/commons/extensions/ExtensionAdderProvidersTest.java
```diff
@@ -0,0 +1,355 @@
...
+
+    private static class SimpleExtendableImpl2
+            extends AbstractExtendable<SimpleExtendable> implements SimpleExtendable {
+        @Override
+        public String getImplementationName() {
+            return "Custom";
+        }
+    }
+
+    private static interface SimpleExtension extends Extension<SimpleExtendable> {
```
geofjamg: useless static?

commons/src/test/java/com/powsybl/commons/extensions/ExtensionAdderProvidersTest.java
```diff
@@ -0,0 +1,355 @@
...
+
+    private static class SimpleExtensionImpl extends AbstractExtension<SimpleExtendable>
+            implements SimpleExtension {
+    }
+
+    private static class SimpleExtensionImpl2 extends AbstractExtension<SimpleExtendable>
+            implements SimpleExtension {
+    }
+
+    private static interface SimpleExtensionAdder extends ExtensionAdder<SimpleExtendable, SimpleExtension> {
```
geofjamg: useless static?

commons/src/test/java/com/powsybl/commons/extensions/ExtensionAdderProvidersTest.java
```diff
@@ -0,0 +1,355 @@
...
+            return SimpleExtensionAdderImpl2.class;
+        }
+
+        @Override
+        public SimpleExtensionAdderImpl2 newAdder(SimpleExtendable extendable) {
+            return new SimpleExtensionAdderImpl2(extendable);
+        }
+    }
+
+    private static interface GenericExtendable<G extends GenericExtendable<G>> extends Extendable<G> {
```
geofjamg: useless static?

commons/src/test/java/com/powsybl/commons/extensions/ExtensionAdderProvidersTest.java
```diff
@@ -0,0 +1,355 @@
...
+    }
+
+    private static class GenericExtendableImpl2<G extends GenericExtendable<G>>
```
geofjamg: Is it used somewhere?

commons/src/test/java/com/powsybl/commons/extensions/ExtensionAdderProvidersTest.java
```diff
@@ -0,0 +1,355 @@
...
+            extends AbstractExtendable<G> implements GenericExtendable<G> {
+        @Override
+        public String getImplementationName() {
+            return "Custom";
+        }
+    }
+
+    private static interface SpecificExtendable
```
geofjamg: useless static?

commons/src/test/java/com/powsybl/commons/extensions/ExtensionAdderProvidersTest.java
```diff
@@ -0,0 +1,355 @@
...
+            extends AbstractExtension<G>
+            implements GenericExtension<G> {
+    }
+
+    private static class GenericExtensionImpl2<G extends GenericExtendable<G>>
+            extends AbstractExtension<G>
+            implements GenericExtension<G> {
+    }
+
+    private static interface GenericExtensionAdder<G extends GenericExtendable<G>> extends ExtensionAdder<G, GenericExtension<G>> {
```
geofjamg: useless static?

commons/src/test/java/com/powsybl/commons/extensions/ExtensionAdderProvidersTest.java
```diff
@@ -0,0 +1,355 @@
...
+        } catch (PowsyblException e) {
+            assertTrue("Should throw Missing Provider exception",
+                    e.getMessage().contains("not found"));
+        }
+    }
+
+    @Test
+    public void testMissingAlternate() {
+        try {
+            List<ExtensionAdderProvider> listProviders = Arrays
```
geofjamg: Unused variable?

commons/src/test/java/com/powsybl/commons/extensions/ExtensionAdderProvidersTest.java
```diff
@@ -0,0 +1,355 @@
...
+        } catch (PowsyblException e) {
+            assertTrue("Should throw Missing Provider exception",
+                    e.getMessage().contains("not found"));
+        }
+    }
+
+    @Test
+    public void testMissingAlternate2() {
+        try {
+            List<ExtensionAdderProvider> listProviders = Arrays
```
geofjamg: Unused variable?

commons/src/test/java/com/powsybl/commons/extensions/ExtensionAdderProvidersTest.java
jonenst: Yes

commons/src/main/java/com/powsybl/commons/extensions/AbstractExtensionAdder.java
jonenst: yes good catch

commons/src/test/java/com/powsybl/commons/extensions/ExtensionAdderProvidersTest.java
jonenst: Good catch the test was not testing what I wanted to test

commons/src/test/java/com/powsybl/commons/extensions/ExtensionAdderProvidersTest.java
jonenst: Good catch the test was not testing what I wanted to test

commons/src/test/java/com/powsybl/commons/extensions/ExtensionAdderProvidersTest.java
jonenst: Good catch added a test

commons/src/main/java/com/powsybl/commons/extensions/ExtensionAdderProviders.java
```diff
@@ -0,0 +1,95 @@
...
+                    type.getSimpleName(), name);
+            throw new PowsyblException("ExtensionAdderProvider not found");
+        }
+
+        if (providers.size() > 1) {
+            LOGGER.error(
+                    "Multiple ExtensionAdderProviders found for ExtensionAdder {} for implementation {} : {}",
+                    type.getSimpleName(), name, providers);
+            throw new PowsyblException(
+                    "Multiple platform configuration providers found");
```
mathbagu: Maybe a copy/past error: are you sure about this message?

jonenst: Oops yes

jonenst: fixed

PR #1154, opened 2020-02-11
Fix LoadFlowParameters copy.

### LINKED ISSUE #1151: [Load flow] Parameters copy does not work with extensions

Bug.

`LoadFlowParameters.copy()` does not copy extensions. For instance, the use of this in `SecurityAnalysisImpl` causes post contingency violations to not use extensions.

A copy of parameters should have exactly the same behaviour as the original. Also, modifying the copy should not modify the original.

An easy solution will be to implement the copy by serializing and reading again the parameters, since extensions are usually serializable.

Otherwise, there is no easy technical solution for this, as extensions cannot be copied themselves.
We would need to choose if:
     - we want to support copy, and then impose that extensions must be copyable 
     - not support it and deprecate or remove the `copy` method

  - PowSyBl Version: 3.1.0

yichen88:
Copy extensions via serialize instance.
Fix usage in SecurityAnalysisImpl.

fix #1151 

**Other information**:


loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowParameters.java
```diff
@@ -181,7 +186,19 @@ public LoadFlowParameters setSpecificCompatibility(boolean specificCompatibility
...
+        byte[] bytes = writeInMemory();
+        try (ByteArrayInputStream bais = new ByteArrayInputStream(bytes)) {
+            return JsonLoadFlowParameters.read(bais);
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
+
+    private byte[] writeInMemory() {
+        try (ByteArrayBuilder byteArrayBuilder = new ByteArrayBuilder()) {
```
mathbagu: Why do you use `ByteArrayBuilder`?  Why don't you use `ByteArrayOutputStream`?

pom.xml
```diff
@@ -136,6 +136,16 @@
...
                 <version>${blazegraph.version}</version>
             </dependency>
+            <dependency>
+                <groupId>com.fasterxml.jackson.core</groupId>
+                <artifactId>jackson-core</artifactId>
+                <version>${jackson.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>com.fasterxml.jackson.core</groupId>
+                <artifactId>jackson-annotations</artifactId>
```
mathbagu: I cannot see where this dependency is used. Is it really needed?

This dependency should be declared before jackson-core.

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowParameters.java
yichen88: From doc, it seems like it is a better usage here.
> Since version 1.5, also implements OutputStream to allow efficient aggregation of output content as a byte array, similar to how ByteArrayOutputStream works, but somewhat more efficiently for many use cases.

pom.xml
sylvlecl: The jackson-annotations and jackson-core are pulled by jackson-databind (hence the order, but we can change it).

In case we don't impose here the version used, blazegraph pulls its own version of those dependencies, which is inconsistent with our version of jackson-databind : it throws unknown methods exceptions at runtime in cgmes unit tests.

As it could happen with dependencies other than blazegraph, it's better to impose our own version of those 2 dependencies a the parent pom level.

sylvlecl: It's now correctly sorted!

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowParameters.java
mathbagu: We could probably keep `ByteArrayBuilder` then...

mathbagu:
@yichen88 @sylvlecl Sonar is not happy with the code coverage. Maybe we should remove the copy constructor to solve this as we know it's buggy.
What do you think?

PR #1155, opened 2020-02-11
Modify the sensitivity API to allow for post-contingency sensitivity analyses

agnesLeroy:

New feature: modification of the sensitivity api to make it possible to perform post-contingency sensitivity analyses based on Hades2. The HADES2 sensitivity provider used with Powsybl offers the possibility to calculate the sensitivity on a set of contingencies besides the N state. The computation is launched only once, but the solver itself modifies the matrix for each state of the network to output a full set of results. 

At the moment it is only possible to run single sensitivity analyses through Powsybl. It returns, besides its status and some stats on the computation itself, all the sensitivity values for each factor (combination of a monitoredBranch and a specific equipment or group of equipments) on the network in state N.

With this change it is possible to run sensitivity analyses on state N and post-contingency states, based on the Hades2 provider, by passing a list of contingencies as an additional argument. It runs a single calculation, but that calculation will compute the sensitivity factors for the N state and on a set of N-1 network states. The full set of results then consists of :
- the list of sensitivity values in N
- the lists of sensitivity values for each N-1 situation
- some metadata (status, stats, logs)

### DIFF
--- a/sensitivity-api/pom.xml
+++ b/sensitivity-api/pom.xml
@@ -93,6 +93,12 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.powsybl</groupId>
+            <artifactId>powsybl-contingency-api</artifactId>
+            <version>${project.version}</version>
+            <scope>compile</scope>
+        </dependency>
     </dependencies>
 
 </project>
--- a/sensitivity-api/pom.xml
+++ b/sensitivity-api/pom.xml
@@ -45,6 +45,11 @@
             <artifactId>powsybl-loadflow-api</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>powsybl-contingency-api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
 
         <!-- Test dependencies -->
         <dependency>
@@ -92,12 +97,6 @@
             <artifactId>powsybl-iidm-impl</artifactId>
             <version>${project.version}</version>
             <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>com.powsybl</groupId>
-            <artifactId>powsybl-contingency-api</artifactId>
-            <version>${project.version}</version>
-            <scope>compile</scope>
         </dependency>
     </dependencies>
 
--- a/sensitivity-api/src/main/java/com/powsybl/sensitivity/SensitivityComputation.java
+++ b/sensitivity-api/src/main/java/com/powsybl/sensitivity/SensitivityComputation.java
@@ -7,6 +7,7 @@
 package com.powsybl.sensitivity;
 
 import com.powsybl.commons.Versionable;
+import com.powsybl.contingency.ContingenciesProvider;
 
 import java.util.concurrent.CompletableFuture;
 
@@ -17,7 +18,7 @@ import java.util.concurrent.CompletableFuture;
  *     Sensitivity computation is used to assess the impact of a small modification
  *     of a network variables on the value of network functions.
  *     This computation can be assimilated to a partial derivative computed on a given
- *     network state.
+ *     network state and on that state modified based on a list of contingencies, if specified.
  * </p>
  * <p>
  *     PTDFs used in Flowbased methodology for example are sensitivity computation
@@ -28,12 +29,24 @@ import java.util.concurrent.CompletableFuture;
  */
 public interface SensitivityComputation extends Versionable {
     /**
-     * Run an asynchronous sensitivity computation job using given parameters and input provider
+     * Run an asynchronous systematic sensitivity computation job using given parameters and input provider
+     *
+     * @param factorsProvider sensitivity factors provider for the computation
+     * @param contingenciesProvider contingencies provider for the computation
+     * @param workingStateId id of the network base state for the computation
+     * @param sensiParameters sensitivity computation parameters
+     * @return the sensitivity computation results in N and N-1
+     */
+    CompletableFuture<SensitivityComputationResults> run(SensitivityFactorsProvider factorsProvider, ContingenciesProvider contingenciesProvider, String workingStateId, SensitivityComputationParameters sensiParameters);
+
+
+    /**
+     * Run an asynchronous single sensitivity computation job using given parameters and input provider
      *
      * @param factorsProvider sensitivity factors provider for the computation
      * @param workingStateId id of the network base state for the computation
      * @param sensiParameters sensitivity computation parameters
-     * @return the sensitivity computation results
+     * @return the sensitivity computation results on the given situation
      */
     CompletableFuture<SensitivityComputationResults> run(SensitivityFactorsProvider factorsProvider, String workingStateId, SensitivityComputationParameters sensiParameters);
 
--- a/sensitivity-api/src/main/java/com/powsybl/sensitivity/SensitivityComputation.java
+++ b/sensitivity-api/src/main/java/com/powsybl/sensitivity/SensitivityComputation.java
@@ -39,7 +39,7 @@ public interface SensitivityComputation extends Versionable {
     CompletableFuture<SensitivityComputationResults> run(SensitivityFactorsProvider factorsProvider, String workingStateId, SensitivityComputationParameters sensiParameters);
 
     /**
-     * Run an asynchronous systematic sensitivity computation job using given parameters and input provider
+     * Run an asynchronous sensitivity analysis job using given parameters and input provider
      *
      * @param factorsProvider sensitivity factors provider for the computation
      * @param contingenciesProvider contingencies provider for the computation
--- a/sensitivity-api/src/main/java/com/powsybl/sensitivity/SensitivityComputation.java
+++ b/sensitivity-api/src/main/java/com/powsybl/sensitivity/SensitivityComputation.java
@@ -29,18 +29,6 @@ import java.util.concurrent.CompletableFuture;
  */
 public interface SensitivityComputation extends Versionable {
     /**
-     * Run an asynchronous systematic sensitivity computation job using given parameters and input provider
-     *
-     * @param factorsProvider sensitivity factors provider for the computation
-     * @param contingenciesProvider contingencies provider for the computation
-     * @param workingStateId id of the network base state for the computation
-     * @param sensiParameters sensitivity computation parameters
-     * @return the sensitivity computation results in N and N-1
-     */
-    CompletableFuture<SensitivityComputationResults> run(SensitivityFactorsProvider factorsProvider, ContingenciesProvider contingenciesProvider, String workingStateId, SensitivityComputationParameters sensiParameters);
-
-
-    /**
      * Run an asynchronous single sensitivity computation job using given parameters and input provider
      *
      * @param factorsProvider sensitivity factors provider for the computation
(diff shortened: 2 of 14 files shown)


sensitivity-api/pom.xml
```diff
@@ -93,6 +93,12 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.powsybl</groupId>
```
MioRtia: Replace `com.powsybl` by `${project.groupId}`

sensitivity-api/pom.xml
```diff
@@ -93,6 +93,12 @@
...
+            <artifactId>powsybl-contingency-api</artifactId>
+            <version>${project.version}</version>
+            <scope>compile</scope>
```
MioRtia: Why is it in `compile` scope?

sensitivity-api/src/main/java/com/powsybl/sensitivity/SensitivityComputation.java
```diff
@@ -28,12 +29,24 @@
...
+     * Run an asynchronous systematic sensitivity computation job using given parameters and input provider
      *
      * @param factorsProvider sensitivity factors provider for the computation
+     * @param contingenciesProvider contingencies provider for the computation
      * @param workingStateId id of the network base state for the computation
      * @param sensiParameters sensitivity computation parameters
-     * @return the sensitivity computation results
+     * @return the sensitivity computation results in N and N-1
+     */
+    CompletableFuture<SensitivityComputationResults> run(SensitivityFactorsProvider factorsProvider, ContingenciesProvider contingenciesProvider, String workingStateId, SensitivityComputationParameters sensiParameters);
```
MioRtia: add a default implementation to prevent breaking change here

sensitivity-api/src/main/java/com/powsybl/sensitivity/SensitivityComputationResults.java
```diff
@@ -41,23 +51,34 @@
     @JsonProperty("values")
     private final List<SensitivityValue> sensitivityValues;
 
+    @JsonProperty("contingencies present")
+    private final boolean contingenciesPresent;
```
MioRtia: Not sure having this field is a good idea. It could lead to inconsistent data (e.g. empty contingencies map and `contingenciesPresent` as `true` or non-empty contingencies map and `contingenciesPresent` as `false`. Why do you need this field? Isn't it enough to check that `sensitivityValueContingencies` is empty or not?

sensitivity-api/src/main/java/com/powsybl/sensitivity/SensitivityComputationResults.java
```diff
@@ -90,61 +111,142 @@ public String getLogs() {
...
         return returnValue.get();
     }
+
+    /**
+     * Get the status of the presence of contingencies
+     *
+     * @return true if the computation contains contingencies, false otherwise
+     */
+    public boolean contingenciesArePresent() {
+        return contingenciesPresent;
```
MioRtia: Same remark as above. This method can return `!sensitivityValuesContingencies.isEmpty()`

sensitivity-api/src/main/java/com/powsybl/sensitivity/SensitivityComputationResults.java
```diff
@@ -90,61 +111,142 @@ public String getLogs() {
...
+    }
+
+    /**
+     * Get a collection of all the sensitivity values for all contingencies.
+     *
+     * @return a collection of all the sensitivity values for all contingencies.
+     */
+    public Map<String, List<SensitivityValue>> getSensitivityValuesContingencies() {
+        return sensitivityValuesContingencies;
```
MioRtia: Return a copy of the map and not the map itself to prevent unsafe changes:
```java
return Collections.unmodifiableMap(sensitivityValuesContingencies);
```

sensitivity-api/src/main/java/com/powsybl/sensitivity/SensitivityComputationResults.java
```diff
@@ -41,23 +51,34 @@
...
+                                         @JsonProperty("values") List<SensitivityValue> sensitivityValues,
+                                         @JsonProperty("contingencies present") boolean contingenciesPresent,
+                                         @JsonProperty("contingencyId values") Map<String, List<SensitivityValue>> sensitivityValuesContingencies) {
         this.ok = ok;
         this.metrics = Objects.requireNonNull(metrics);
         this.logs = Objects.requireNonNull(logs);
-        this.sensitivityValues = Objects.requireNonNull(values);
+        this.sensitivityValues = Objects.requireNonNull(sensitivityValues);
+        this.contingenciesPresent = contingenciesPresent;
+        this.sensitivityValuesContingencies = sensitivityValuesContingencies;
```
MioRtia: Don't set directly the map but set as copy to prevent unsafe changes:
```java
this.sensitivityValuesContingencies = Collections.unmodifiableMap(sensitivityValuesContingencies);
```

sensitivity-api/src/main/java/com/powsybl/sensitivity/SensitivityComputationResults.java
```diff
@@ -90,61 +111,142 @@ public String getLogs() {
...
+     * Get a collection of all the sensitivity values associated with given variable
+     * for a specific contingency.
+     *
+     * @param variable sensitivity variable
+     * @param contingencyId the ID of the considered contingency
+     * @return a collection of all the sensitivity values associated with given variable
+     */
+    public Collection<SensitivityValue> getSensitivityValuesByVariable(SensitivityVariable variable, String contingencyId) {
+        return sensitivityValuesContingencies.get(contingencyId).stream().filter(sensitivityValue -> sensitivityValue.getFactor().getVariable().equals(variable)).collect(Collectors.toList());
+    }
```
MioRtia: Indent to make it more readable:
```java
sensitivityValuesContingencies.get(contingencyId).stream()
     .filter(sensitivityValue -> sensitivityValue.getFactor().getVariable().equals(variable))
     .collect(Collectors.toList());
```

sensitivity-api/src/main/java/com/powsybl/sensitivity/SensitivityComputationResults.java
```diff
@@ -90,61 +111,142 @@ public String getLogs() {
...
+     * @return the sensitivity value associated with given function and given variable
+     */
+    public SensitivityValue getSensitivityValue(SensitivityFunction function, SensitivityVariable variable, String contingencyId) {
+        Optional<SensitivityValue> returnValue;
+        returnValue = sensitivityValuesContingencies.get(contingencyId).stream().filter(sensitivityValue -> sensitivityValue.getFactor().getFunction().equals(function)
+                && sensitivityValue.getFactor().getVariable().equals(variable)).findAny();
+        if (!returnValue.isPresent()) {
+            throw new NoSuchElementException(String.format("Sensitivity value not found for function %s and variable %s at contingencyId %s.", function.getId(), variable.getId(), contingencyId));
+        }
+        return returnValue.get();
```
MioRtia: You can simplify this code:
```java
return sensitivityValuesContingencies.get(contingencyId).stream()
    .filter(sensitivityValue -> sensitivityValue.getFactor().getFunction().equals(function)
                && sensitivityValue.getFactor().getVariable().equals(variable))
    .findFirst()
    .orElseThrow(() -> new NoSuchElementException(String.format("Sensitivity value not found for function %s and variable %s at contingencyId %s.", function.getId(), variable.getId(), contingencyId)));
```

sensitivity-api/src/main/java/com/powsybl/sensitivity/SensitivityComputationResults.java
```diff
@@ -90,61 +111,142 @@ public String getLogs() {
...
+     * @param contingencyId the ID of the considered contingency
+     * @return the sensitivity value associated with given function and given variable
+     */
+    public SensitivityValue getSensitivityValue(SensitivityFactor factor, String contingencyId) {
+        Optional<SensitivityValue> returnValue;
+        returnValue = sensitivityValuesContingencies.get(contingencyId).stream().filter(sensitivityValue -> sensitivityValue.getFactor().equals(factor)).findAny();
+        if (!returnValue.isPresent()) {
+            throw new NoSuchElementException(String.format(VALUE_NOT_FOUND, factor.getFunction().getId(), factor.getVariable().getId()));
+        }
+        return returnValue.get();
```
MioRtia: Same as above, this code can be simplified. (Code of other similar methods of this class can be simplified as well)

sensitivity-api/src/main/java/com/powsybl/sensitivity/SensitivityComputationResults.java
```diff
@@ -41,23 +51,34 @@
...
                                          @JsonProperty("logs") String logs,
-                                         @JsonProperty("values") List<SensitivityValue> values) {
+                                         @JsonProperty("values") List<SensitivityValue> sensitivityValues,
+                                         @JsonProperty("contingencies present") boolean contingenciesPresent,
+                                         @JsonProperty("contingencyId values") Map<String, List<SensitivityValue>> sensitivityValuesContingencies) {
         this.ok = ok;
         this.metrics = Objects.requireNonNull(metrics);
         this.logs = Objects.requireNonNull(logs);
-        this.sensitivityValues = Objects.requireNonNull(values);
+        this.sensitivityValues = Objects.requireNonNull(sensitivityValues);
```
MioRtia: Don't set directly the list but set as copy to prevent unsafe changes:
```java
this.sensitivityValues = Collections.unmodifiableList(Objects.requireNonNull(sensitivityValues));
```

sensitivity-api/src/main/java/com/powsybl/sensitivity/converter/SensitivityComputationResultExporter.java
```diff
@@ -30,7 +30,7 @@
     String getComment();
 
     /**
-     * Export a result of a sensitivity computation
+     * Export a result of a sensitivity computation without contingency
```
MioRtia: `CsvSensitivityComputationResultExporter` implements this interface and seems to be able to export a result of a sensitivity computation with contingencies, why is "without contingency" specified here?

sensitivity-api/src/main/java/com/powsybl/sensitivity/SensitivityComputationResults.java
```diff
@@ -41,23 +51,34 @@
     @JsonProperty("values")
     private final List<SensitivityValue> sensitivityValues;
 
+    @JsonProperty("contingencies present")
+    private final boolean contingenciesPresent;
+
+    @JsonProperty("contingencies values")
```
MioRtia: I know that JSON format allows whitespaces in properties but shouldn't we use camel case instead to be consistent with other results serialization?

geofjamg:
A general comment about sensitivity: is systematic sensitivity the correct name? Same for sensitivity computation for me the right term is sensitivity analysis ( most common in power system domain). For systematic sensitivity it is not very meaningful and not common. Probably "post contingency sensitivity analysis" is more correct.

agnesLeroy:
@geofjamg I agree, sensitivity analysis seems indeed much more common than sensitivity computation. I also agree that the word "systematic" sounds poorly chosen. On the other hand, "post-contingency sensitivity analysis" does not reflect the fact that we are dealing with successive sensivitiy analyses. Maybe a good name would be just that : successive sensitivity analyses ?

sensitivity-api/src/main/java/com/powsybl/sensitivity/SensitivityComputation.java
agnesLeroy: I had thought of handling that on the implementation side, is it wrong?

geofjamg:
> Maybe a good name would be just that : successive sensitivity analyses ?

It seems to me that the important thing is we want to analyze sensitivity on post contingency states. The systematic aspect (meaning that we want to analyze for several contingencies) or whatever synonym we can found is not really important. Is is just a matter of parameter and one contingency analysis can be done by just passing a list of one contingency.  

It is exactly like loadflow (also called powerflow) versus security analysis (sometimes called contingency analysis). What is important here is that we want to analyse the security of the grid => calculate constraint on post contingency states not that we want to simulate the security for one or many contingencies.

sensitivity-api/src/main/java/com/powsybl/sensitivity/SensitivityComputationResults.java
```diff
@@ -89,61 +105,144 @@ public String getLogs() {
...
+     * @param variable sensitivity variable
+     * @param contingencyId the ID of the considered contingency
+     * @return the sensitivity value associated with given function and given variable
+     */
+    public SensitivityValue getSensitivityValue(SensitivityFunction function, SensitivityVariable variable, String contingencyId) {
+        return sensitivityValuesContingencies.get(contingencyId).stream()
+                .filter(sensitivityValue -> sensitivityValue.getFactor().getFunction().equals(function)
+                        && sensitivityValue.getFactor().getVariable().equals(variable))
+                .findFirst()
+                .orElseThrow(() -> new NoSuchElementException(String.format("Sensitivity value not found for function %s and variable %s at contingencyId %s.", function.getId(), variable.getId(), contingencyId)));
```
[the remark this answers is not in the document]
agnesLeroy: Done

PR #1156, opened 2020-02-12
Handle switches missing equipment container in CGMES conversion

zamarrenolm:

No

Bug fix

Equipment container for switch is required to decide if switch should be converted to a low impedance line during CGMES conversion. If equipment container referenced object does not exist in the CGMES model a null pointer exception is raised.
Switches are converted to low impedance lines when any of the voltage levels of the end points are different of the voltage level of the equipment container.

If the equipment container is not found, the voltage level of the equipment container can not be obtained. The result will be that the switch will be converted to a low impedance line in this situation.
An error is written in the log.


MioRtia:
I agree with this PR, is it also possible to add a test case for this correction?

PR #1161, opened 2020-02-17
Delete node count

### LINKED ISSUE #781: Clarify graph API regarding vertex indices and its use in node breaker topology

API improvement.

See PR #775:
the node breaker topology uses the `UndirectGraph` interface to represent the topology. When creating it, it starts by setting a "node count" before actually adding those nodes.
A problem arises when nodes indices are not contiguous, for example after removing some equipments. In that  case, the index of some vertices can overflow the allocated array.
The PR fixes this by ensuring that the array is extended in that case, but the node breaker API remains ambiguous.

The API must be clarified:
    - Are vertices indices required to be contiguous ?
    - Should the setNodeCount method be deleted, considered as handled by the graph implementation? Otherwise, what should it do ? If only allocation with a "guess", rename it ? If we remove it, should we add a method to ensure memory efficiency once the voltage level is "stabilized" ? (compact)

Clarify the API to avoid bugs when using it, for example by underallocating vertices, or by looping on vertices indices from 0 to N whereas they are not contiguous.

PowSyBl Version: 2.5.0-SNAPSHOT

Follows bug fix in PR #775 .

### LINKED ISSUE #1118: Delete nodeCount attribute of NodeBreakerView network
Evolution? but it will also solve a bug

When a `Connectable` is removed from a voltage level, the vertex is removed if it is isolated. When we export the network to an XML, and import the file, the `nodeCount` can be lesser than the index of a vertex. Then, an exception is thrown.

The `nodeCount` attribute will be deleted (not present anymore in the XIIDM file.

This attribute was supposed to be used to allocate a certain amount of vertices but it is not so useful in the end... Deleting it would solve the issue.

  - PowSyBl Version: 3.2.0-SNAPSHOT

MioRtia:

fix #781 
fix #1118 

Bug fix

**Other information**:
In this proposition, I kept the same behaviour for `NodeBreakerVoltageLevel.cleanTopology()` than `NodeBreakerVoltageLevel.clean()` used to have **with the same issues**: hence, when the topology is cleaned and intermediate nodes are removed, they cannot be used again e.g. if a node/breaker voltage level only used nodes 1, 2 and 4 and is cleaned, attempting to use node 3 will throw a `PowsyblException` with the message `Vertex 3 not found`. Is this the behavior we want?

### DIFF
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/VoltageLevel.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/VoltageLevel.java
@@ -405,7 +405,7 @@ public interface VoltageLevel extends Container<VoltageLevel> {
         Stream<InternalConnection> getInternalConnectionStream();
 
         /**
-         * Remove <b>all</b> the internal connections between node1 and node2 if they exist.
+         * Remove <b>all</b> the internal connections between node1 and node2 (not orientated) if they exist.
          */
         default void removeInternalConnections(int node1, int node2) {
             throw new UnsupportedOperationException();
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/VoltageLevel.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/VoltageLevel.java
@@ -406,9 +406,9 @@ public interface VoltageLevel extends Container<VoltageLevel> {
         Stream<InternalConnection> getInternalConnectionStream();
 
         /**
-         * Remove the internal connection between node1 and node2 if it exists.
+         * Remove <b>all</b> the internal connections between node1 and node2 if they exist.
          */
-        default void removeInternalConnection(int node1, int node2) {
+        default void removeInternalConnections(int node1, int node2) {
             throw new UnsupportedOperationException();
         }
 
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/VoltageLevel.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/VoltageLevel.java
@@ -360,18 +360,17 @@ public interface VoltageLevel extends Container<VoltageLevel> {
         /**
          * Get the count of used nodes (nodes attached to an equipment, a switch or an internal connection).
          *
-         * @deprecated Depending on your use case, use {@link #getMaximumNodeIndex()} or {@link #getNodeCapacity()}.
+         * @deprecated Use {@link #getMaximumNodeIndex()} instead.
          */
         @Deprecated
         default int getNodeCount() {
             throw new UnsupportedOperationException();
         }
 
+        /**
+         * Get the highest index of used nodes (i.e. attached to an equipment, a switch or an internal connection) in the voltage level.
+         */
         default int getMaximumNodeIndex() {
-            throw new UnsupportedOperationException();
-        }
-
-        default int getNodeCapacity() {
             throw new UnsupportedOperationException();
         }
 
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/VoltageLevel.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/VoltageLevel.java
@@ -358,19 +358,27 @@ public interface VoltageLevel extends Container<VoltageLevel> {
         }
 
         /**
-         * Get the number of node.
+         * Get the count of used nodes (nodes attached to an equipment, a switch or an internal connection).
+         *
+         * @deprecated Depending on your use case, use {@link #getMaximumNodeIndex()} or {@link #getNodeCapacity()}.
          */
-        int getNodeCount();
+        @Deprecated
+        default int getNodeCount() {
+            throw new UnsupportedOperationException();
+        }
+
+        default int getMaximumNodeIndex() {
+            throw new UnsupportedOperationException();
+        }
+
+        default int getNodeCapacity() {
+            throw new UnsupportedOperationException();
+        }
 
         /**
          * Get the list of nodes.
          */
         int[] getNodes();
-
-        /**
-         * Set the number of node.
-         */
-        NodeBreakerView setNodeCount(int count);
 
         /**
          * Get a builder to create a new switch.
@@ -396,6 +404,13 @@ public interface VoltageLevel extends Container<VoltageLevel> {
          * Get internal connection stream.
          */
         Stream<InternalConnection> getInternalConnectionStream();
+
+        /**
+         * Remove the internal connection between node1 and node2 if it exists.
+         */
+        default void removeInternalConnection(int node1, int node2) {
+            throw new UnsupportedOperationException();
+        }
 
         /**
          * Get a builder to create a new breaker.
@@ -1035,5 +1050,4 @@ public interface VoltageLevel extends Container<VoltageLevel> {
      * @param writer a writer
      */
     void exportTopology(Writer writer) throws IOException;
-
 }
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusBreakerVoltageLevel.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusBreakerVoltageLevel.java
@@ -340,8 +340,8 @@ class BusBreakerVoltageLevel extends AbstractVoltageLevel {
     @Override
     public Iterable<Terminal> getTerminals() {
         return FluentIterable.from(graph.getVerticesObj())
-                             .transformAndConcat(ConfiguredBus::getTerminals)
-                             .transform(Terminal.class::cast);
+                .transformAndConcat(ConfiguredBus::getTerminals)
+                .transform(Terminal.class::cast);
     }
 
     @Override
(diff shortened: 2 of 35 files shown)


iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/util/IidmXmlUtil.java
```diff
@@ -21,6 +21,8 @@
  */
 public final class IidmXmlUtil {
 
+    private static final String FOR_IIDM_XML_VERSION = " for IIDM-XML version ";
```
mathbagu: Create a complete message instead and use it in Message.format()

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevel.java
```diff
@@ -872,13 +863,23 @@ private void checkTerminal(TerminalExt terminal) {
...
     }
 
+    private void addNodes(int node) {
+        int oldCount = graph.getVertexCount();
+        if (node >= oldCount) {
+            for (int i = oldCount; i < node + 1; i++) {
+                graph.addVertex();
+            }
+        }
+    }
```
mathbagu: I'm not sure it's what we discussed. Adding a vertex will create a terminal. The idea is to create a terminal only if you need to attach something. As you remove the call to `clean()`, the unnecessary terminals are not removed.

It's compliant with the current version of powsybl, but I thought we wanted to go further.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevel.java
```diff
@@ -166,6 +166,8 @@ public Switch add() {
             }
             SwitchImpl aSwitch = new SwitchImpl(NodeBreakerVoltageLevel.this, id, getName(), kind, open, retained, fictitious);
             getNetwork().getIndex().checkAndAdd(aSwitch);
+            addNodes(node1);
```
mathbagu: I think I will rename this method to `reserveNodes` or `ensureNode`. The **s** in the name of the method seems strange for someone who reads the code. You have to go in the implementation to understand why you put an **s**.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusBreakerVoltageLevel.java
```diff
@@ -757,7 +752,7 @@ public void detach(final TerminalExt terminal) {
     }
 
     @Override
-    public void clean() {
+    public void cleanTopology() {
```
mathbagu: Maybe we could keep the name of the old method?

mathbagu:
@MioRtia I think it's strange to not be able to reuse a vertex. How can I replace an injection by another? If it's supported by the API (remove + add), it's OK for me to keep it like this

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevel.java
MioRtia: Okay, I was not really sure about the conclusion of what had been decided last time... If we want to create a vertex only each time a Terminal is created, we might not need a `clean` method (the vertex will be deleted when not used and created again if reused).

MioRtia:
> @MioRtia I think it's strange to not be able to reuse a vertex. How can I replace an injection by another? If it's supported by the API (remove + add), it's OK for me to keep it like this

With this proposition, when an injection (for example) is removed and another replaces it, there is no issue. There is an issue however, when `cleanTopology` is called between the two actions. I'm wondering how useful should this method be? As I said above, if we don't want to create "superfluous" vertices, it can be deleted (it will avoid most of issues I think).

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/VoltageLevel.java
```diff
@@ -358,20 +358,28 @@
...
+         *
+         * @deprecated Depending on your use case, use {@link #getMaximumNodeIndex()} or {@link #getNodeCapacity()}.
          */
-        int getNodeCount();
+        @Deprecated
+        default int getNodeCount() {
+            throw new UnsupportedOperationException();
+        }
+
+        default int getMaximumNodeIndex() {
```
mathbagu: Please add Javadoc on this method to well understand what is expected

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/VoltageLevel.java
```diff
@@ -358,20 +358,28 @@
...
+            throw new UnsupportedOperationException();
+        }
+
+        default int getNodeCapacity() {
```
mathbagu: Please add Javadoc on this method to well understand what is expected

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevel.java
```diff
@@ -647,6 +653,21 @@ public int getNode2() {
                     });
         }
 
+        @Override
+        public void removeInternalConnections(int node1, int node2) {
+            int[] internalConnectionsToBeRemoved = Arrays.stream(graph.getEdges())
+                    .filter(e -> graph.getEdgeObject(e) == null)
+                    .filter(e -> graph.getEdgeVertex1(e) == node1 && graph.getEdgeVertex2(e) == node2)
```
mathbagu: I think you should also get edges between node2 and node1 (the graph is not directed)

math/src/main/java/com/powsybl/math/graph/UndirectedGraph.java
```diff
@@ -49,6 +56,10 @@
      */
     int getVertexCount();
 
+    default int getMaximumVertexIndex() {
```
mathbagu: Add a Javadoc also there

math/src/main/java/com/powsybl/math/graph/UndirectedGraphImpl.java
```diff
@@ -129,6 +129,24 @@ public int addVertex() {
         return v;
     }
 
+    @Override
+    public void addVertexIfNotPresent(int v) {
+        if (v < vertices.size()) {
+            if (removedVertices.contains(v)) {
```
mathbagu: Are you sure this is efficient: should we store the removedVertices in another container (ordered) like a set?

math/src/main/java/com/powsybl/math/graph/UndirectedGraphImpl.java
```diff
@@ -129,6 +129,24 @@ public int addVertex() {
...
+                vertices.set(v, new Vertex<>());
+                removedVertices.remove(v);
+            }
+        } else {
+            for (int i = vertices.size(); i < v; i++) {
+                vertices.add(null);
+                removedVertices.add(i);
```
mathbagu: Maybe we should also rename `removedVertices` to availableVertices`

math/src/main/java/com/powsybl/math/graph/UndirectedGraphImpl.java
```diff
@@ -149,11 +167,29 @@ public V removeVertex(int v) {
         return obj;
     }
 
+    private void cleanVertices(int v) {
+        vertices.remove(v);
+        removedVertices.remove(v);
+        if (removedVertices.contains(v - 1)) {
+            cleanVertices(v - 1);
+        }
```
mathbagu: I'm not sure it is really effecient to have a recursion here where a for-loop looks OK

math/src/main/java/com/powsybl/math/graph/UndirectedGraphImpl.java
```diff
@@ -149,11 +167,29 @@ public V removeVertex(int v) {
...
 
+    @Override
+    public int getMaximumVertexIndex() {
+        for (int i = vertices.size(); i > 0; i--) {
+            if (!removedVertices.contains(i - 1)) {
+                return i - 1;
+            }
+        }
+        return -1;
+    }
```
mathbagu: I'm not sure to understand what this method does:
`if (!removedVertices.contains(i - 1))` seems equivalent to `if (vertices.at(i) != null` ?

Is it really possible to have null vertices at the end of the list. The cleanVertices should avoid that situation?

MioRtia: Yes, this was coded before I thought of doing `cleanVertices`, i will modify it

math/src/main/java/com/powsybl/math/graph/UndirectedGraph.java
MioRtia: In the end I deleted the method since it basically `getVertexCapacity() - 1`

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/VoltageLevel.java
MioRtia: At the end, I deleted this method since it always returns `getMaximumNodeIndex() + 1`. It was useful before having `cleanVertices`.

math/src/main/java/com/powsybl/math/graph/UndirectedGraphImpl.java
MioRtia: At the end, I deleted this method as it is redundant with `getNodeCapacity`

PR #1162, opened 2020-02-17
Merging view step : to create Line between two different networks

tadam50:

No

New feature for non destructive network merge, in order to create a line between 2 different Network

Merging view does not take into account of line creation between two different networks

Creation of a line between two different network is implemented.
One FIXME left : computation of q0, p0 and ucte xnode code in LineAdderAdapter

**Other information**:


iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/LineAdderAdapter.java
```diff
@@ -52,24 +57,83 @@
...
+        Line newLine;
         final Network n1 = checkAndGetNetwork1();
         final Network n2 = checkAndGetNetwork2();
         if (n1 == n2) {
             newLine = index.getLine(addLine(n1));
         } else {
+            // FIXME : how to compute P0, Q0 & ucteXnodeCode ?
+            ucteXnodeCode = "";
+            p0 = 0.0;
+            q0 = 0.0;
```
mathbagu: The P0 and Q0 are equal to the half of the sum of P1 and P2 (respectively Q1 and Q2). The sign depends on the side.
If P1 and P2 are NaN, you cannot initialize these values. You probably have to listen (NetworkListener) an event and recompute them each time P1, P2, Q1, Q2 are updated.

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergedLine.java
```diff
@@ -169,6 +171,11 @@ public String getId() {
         return id;
     }
 
+    public MergedLine setId(String id) {
+        this.id = id;
+        return this;
+    }
```
mathbagu: Id is not mutable

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/LineAdderAdapter.java
tadam50: So, I keep Line creation with q0 & p0 equal to 0.0 (NaN are not allowed)
I implement NetworkListener::onUpdate with variantId into MergingNetworkListener and compute P0 & Q0 if P1, P2, Q1 or Q2 are updated.

How i set ucteXnodeCode ?

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergedLine.java
tadam50: In LineAdderAdapter, in case of Line creation from both network.
A MergedLine is created with generated id from both DangingLine.
But i need to overrride generated id with the requested id by LineAdderAdapter::setId.

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/LineAdderAdapter.java
```diff
@@ -52,24 +57,84 @@
...
+            // Return the merged line as the new line
+            newLine = index.getMergedLineByCode(ucteXnodeCode)
+                           .setId(id);
         }
         return newLine;
     }
 
+    private void checkAndSetUniqueId() {
+        if (this.id == null) {
+            throw new PowsyblException(getClass().getSimpleName() + " id is not set");
```
mathbagu: I'm not sure this is a good idea to use `getClass().getSimpleName()` to get "Line"

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/LineAdderAdapter.java
```diff
@@ -52,24 +57,84 @@
...
+        DanglingLineAdder adder = vl.newDanglingLine()
+                    .setId(id)
+                    .setName(name)
+                    .setP0(p0)
+                    .setQ0(q0)
+                    .setR(r)
+                    .setX(x)
+                    .setG(g)
+                    .setB(b)
+                    .setUcteXnodeCode(ucteXnodeCode)
```
mathbagu: You should make this function static and pass `ucteXnodeCode` as parameter?

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergedLine.java
```diff
@@ -95,6 +97,26 @@ private void mergeProperties(DanglingLine dl1, DanglingLine dl2) {
         });
     }
 
+    void computeAndSetP0() {
+        double p1 = dl1.getTerminal().getP();
+        double p2 = dl2.getTerminal().getP();
+        if (!Double.isNaN(p1) && !Double.isNaN(p2)) {
+            double p0 = (p1 + p2) / 2.0d;
+            dl1.setP0(p0);
+            dl2.setP0(-p0); // sign depends of side
```
mathbagu: No, it depends on the transit flow: `P1 ---->-----DL1.P0 ---->----- DL2.P0 ---->---- P2`

The sign of DL1.P0 is the same as P2, and respectively the sign of DL2.P0 is the same than P1

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergedLine.java
mathbagu: I still disagree: I have to look deeper in the code to understand why the MergedLine can be created without having the final ID

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergingView.java
```diff
@@ -672,7 +672,7 @@ public int getDanglingLineCount() {
     @Override
     public DanglingLine getDanglingLine(final String id) {
         final DanglingLine dl = index.get(n -> n.getDanglingLine(id), index::getDanglingLine);
-        return index.isMerged(dl) ? dl : null;
+        return !index.isMerged(dl) ? dl : null;
```
mathbagu: Change a little bit the code:
`return index.isMerged(dl) ? null : dl`

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergingViewIndex.java
```diff
@@ -272,14 +276,15 @@ Identifiable getIdentifiable(final Identifiable identifiable) {
     }
 
     boolean isMerged(final DanglingLine dl) {
-        return !mergedLineCached.containsKey(dl.getUcteXnodeCode());
+        Objects.requireNonNull(dl, "DanglingLine is null");
+        return mergedLineCached.containsKey(dl.getUcteXnodeCode());
```
mathbagu: See my remarks in the other PR.

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/LineAdderAdapter.java
mathbagu: @tadam50 The xnode code is not a mandatory information so you can just skip it. The idea we have is to use both Xnode and MergedXnode extensions and initialize the attributes from these extensions, so not directly when the object is created.
As the name of the XNode is not mandatory, you have to find another way to cache the merged lines

About P and Q, the solution I suggested is ok.

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergedLine.java
tadam50: setId method must be private-package

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/LineAdderAdapter.java
tadam50: Issue #1183 created to manage Xnode

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/LineAdderAdapter.java
```diff
@@ -52,24 +57,84 @@
...
+            // MergedLine.id is forced here
+            // Return the merged line as the new line
+            newLine = index.getMergedLineByCode(ucteXnodeCode)
+                           .setId(id);
         }
         return newLine;
     }
 
+    private void checkAndSetUniqueId() {
+        if (this.id == null) {
```
mathbagu: Remove this in this expression: it's useless

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/LineAdderAdapter.java
mathbagu: Please take this comment into account.

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergedLine.java
```diff
@@ -95,6 +97,33 @@ private void mergeProperties(DanglingLine dl1, DanglingLine dl2) {
         });
     }
 
+    void computeAndSetP0() {
+        double p1 = dl1.getTerminal().getP();
+        double p2 = dl2.getTerminal().getP();
+        if (!Double.isNaN(p1) && !Double.isNaN(p2)) {
+            double p0 = (p1 + p2) / 2.0d;
```
mathbagu: I think there is a mistake here: the sum should use absolute value: p1 and p2 have opposite sign and their values are close. So if I make the sum, I will have a value close to 0...
Example:
```
p1 = -605
p2 = 600
--> losses = 5
p1 + p2 / 2.0 = -2,5
```

Another way (simplest) way to compute the P0 and Q0 is to use the losses:
```
double losses = (p1 + p2) / 2.0
dl1.P0 = (P1 + 1/2 losses) * sign(P2)
dl2.P0 = (P2 + 1/2 losses) * sign(P1)
```

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergedLine.java
```diff
@@ -95,6 +97,33 @@ private void mergeProperties(DanglingLine dl1, DanglingLine dl2) {
...
+        }
+    }
+
+    void computeAndSetQ0() {
+        double q1 = dl1.getTerminal().getQ();
+        double q2 = dl2.getTerminal().getQ();
+        if (!Double.isNaN(q1) && !Double.isNaN(q2)) {
+            double q0 = (q1 + q2) / 2.0d;
+            dl1.setQ0(getSign(q2) * q0);
+            dl2.setQ0(getSign(q1) * q0);
```
mathbagu: Same mistake there

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergingNetworkListener.java
```diff
@@ -35,8 +35,24 @@ public void onRemoval(final Identifiable identifiable) {
...
+            final MergedLine mergedLine = index.getMergedLineByCode(ucteCode);
+            if (mergedLine != null) {
+                if (attribute.contains("p")) {
+                    mergedLine.computeAndSetP0();
+                } else if (attribute.contains("q")) {
+                    mergedLine.computeAndSetQ0();
+                }
+            }
+        }
+    }
```
mathbagu: I thought we discussed about that: is it OK for you to move this code to a new dedicated listener (DanglingLinePowerListener) and attach it in the constructor of the MergingView?

PR #1169, opened 2020-02-19
SecurityAnalysisImpl multithread fixes alternative implementations

jonenst:

#1097 #1117 

**Other information**:
This is my proposed implementation. I tried to keep it as simple as I can while still beeing correct.
If we decide to go this way, I think we can do more refactoring to make this code more readable (extracting methods will allow to reduce heavy indentation; removing the synchronize on the ResultBuilder and merging the results at the end). This builds upon the work of yichen in #1097 #1117, thanks.

More tests are needed to prove that the many bugs fixed by this PR are real and correctly fixed and we don't have regressions..

The diff is best understood using a whitespace agnostic algorithm, as most of the code only change by indentation https://github.com/powsybl/powsybl-core/pull/1169/files?w=1


security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisImpl.java
```diff
@@ -88,48 +89,76 @@ public boolean removeInterceptor(SecurityAnalysisInterceptor interceptor) {
...
+                        // variants.
+                        future = CompletableFuture.completedFuture(null).thenComposeAsync(aVoid -> {
+                            boolean previousMultiThreadAcces = network.getVariantManager().isVariantMultiThreadAccessAllowed();
+                            network.getVariantManager().allowVariantMultiThreadAccess(true);
+                            CompletableFuture<Void>[] futures = new CompletableFuture[contingencies.size()];
+                            for (int i = 0; i < contingencies.size(); i++) {
+                                Contingency contingency = contingencies.get(i);
+
+                                // run one loadflow per contingency
+                                futures[i] = CompletableFuture
```
sylvlecl: I think sonar cognitive complexity analyzer must be crying :)

It was already the case before the PR, but I think we should stick to the rule that lambda should be as much as possible one-liners, in particular for complex, asynchronous code. 
I think we should simply create more functions here.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisImpl.java
```diff
@@ -88,48 +89,76 @@ public boolean removeInterceptor(SecurityAnalysisInterceptor interceptor) {
...
-                                        resultBuilder.endContingency();
-                                        network.getVariantManager().removeVariant(postContStateId);
-
-                                        return null;
-                                    }, computationManager.getExecutor());
+                        int workerCount = Math.min(computationManager.getResourcesStatus().getAvailableCores(), contingencies.size());
+                        BlockingQueue<Integer> queue = new ArrayBlockingQueue<>(workerCount);
+                        for (int i = 0; i < workerCount; i++) {
+                            String postContStateId = hash + "_" + i;
+                            network.getVariantManager().cloneVariant(VariantManagerConstants.INITIAL_VARIANT_ID, postContStateId);
```
sylvlecl: Here we should clone the `workingStateId` used for pre contingency computation.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisImpl.java
```diff
@@ -88,48 +89,76 @@ public boolean removeInterceptor(SecurityAnalysisInterceptor interceptor) {
...
+                                                throw new UncheckedInterruptedException(e);
+                                            }
+                                        }, computationManager.getExecutor())
+                                        .thenCompose(workerId -> {
+                                            String postContStateId = hash + "_" + workerId;
+                                            return CompletableFuture
+                                                    .runAsync(() -> {
+                                                        network.getVariantManager().setWorkingVariant(postContStateId);
+                                                        // apply the contingency on the network
+                                                        contingency.toTask().modify(network, computationManager);
```
sylvlecl: agreed with @mathbagu : Here we should clone the variant used for pre contingency computation, in order to restore the "N" situation before applying the contingency.

jonenst:
> I think you forget to reset the working variant at the begin or after each contingency simulation. The initialization is done only when you create the worker but it’s not well placed

Good catch. We don't have an API for creating the variants without copying the data so right now the first loadflows copy the data twice.

jonenst:
I also added the change to clone from workingStateId  instead of INITIAL_VARIANT_ID. It's a big change but as discussed orally the previous behavior was probably a bug

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisImpl.java
jonenst: Yes, we should. I didn't refactor to allow to compare to the previous version. If we decide to go this way, we should refactor

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisImpl.java
```diff
@@ -88,48 +89,76 @@ public boolean removeInterceptor(SecurityAnalysisInterceptor interceptor) {
...
+                        BlockingQueue<Integer> queue = new ArrayBlockingQueue<>(workerCount);
+                        for (int i = 0; i < workerCount; i++) {
+                            String postContStateId = hash + "_" + i;
+                            network.getVariantManager().cloneVariant(VariantManagerConstants.INITIAL_VARIANT_ID, postContStateId);
+                            queue.add(i); //Will always work because we are initializing the queue and the initial capacity is workerCount
                         }
+                        // use completedFuture(null).thenCompose so that more of the
+                        // execution is handled by the whenComplete block removing the
+                        // variants.
+                        future = CompletableFuture.completedFuture(null).thenComposeAsync(aVoid -> {
```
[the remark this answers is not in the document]
jonenst: yes, I refrained from refactoring as much as I could to keep the code comparable to the previous code

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisImpl.java
jonenst: Yes, I didn't change the original behavior in the first proposal because it was not related. It's now included

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisImpl.java
jonenst: fixed

jonenst:
> It may have a small performance disadvantage compared to @yichen88 approach, which is the higher number of tasks submitted to the executor (say 10000 if we have 10000 contingencies), but I am not sure it is really significant.

Yes and no. To me, the only problem is when you are submitting an arbitrary big number of tasks to an executor that can not keep up, you get memory problems. And that's when you use backpressure to limit the task producer. But in this case where the number of tasks is the number of contingencies, I think that's the way we are supposed to do it. The performance could be even better than writing an adhoc throttling of the tasks.

jonenst:
From #1117 

> Remaining question is:
how should we choose "X" ? Number of cores of the runtime ? An additional parameter 

> Number of cores of the runtime 
I assume you mean the number of cores in the local jvm Runtime.getRuntime().availableProcessors(); ? 
Good for LocalComputationManager, bad for a distributed computation manager because too low

> computationManager.available-cores
Good for LocalComputationManager, bad for a distributed computation manager because too big (imagine you have a cluster with 2000 declared cores)

>  An additional parameter 
Good for everything but what would be the default value ?

Maybe we should know the size of the pool behind computationManager.getExecutor, and only use half of it ?

For info, for a full french network, we use about 3 mega bytes of data per variant, and it takes 5ms to clone it.

Also, the SlurmComputationManager returns null to getExecutor, we should we do ?

jonenst:
> Yes and no. To me, the only problem is when you are submitting an arbitrary big number of tasks to an executor that can not keep up, you get memory problems. And that's when you use backpressure to limit the task producer. But in this case where the number of tasks is the number of contingencies, I think that's the way we are supposed to do it. The performance could be even better than writing an adhoc throttling of the tasks.

Wait, I have a doubt, If we have many submitted tasks that block (in the queue.take method); can the executor still use the thread for the blocking occured for another task ? If no, then submitting all those tasks that block will kill the Executor.

sylvlecl:
> If no, then submitting all those tasks that block will kill the Executor

Indeed the tasks will block some threads in the waiting state if there are more running tasks than variants. Those blocked threads may cause other unrelated tasks submitted to that executor to wait, too.

jonenst:
> > Yes and no. To me, the only problem is when you are submitting an arbitrary big number of tasks to an executor that can not keep up, you get memory problems. And that's when you use backpressure to limit the task producer. But in this case where the number of tasks is the number of contingencies, I think that's the way we are supposed to do it. The performance could be even better than writing an adhoc throttling of the tasks.
> 
> Wait, I have a doubt, If we have many submitted tasks that block (in the queue.take method); can the executor still use the thread for the blocking occured for another task ? If no, then submitting all those tasks that block will kill the Executor.

Yes thinking about it some more, there's a problem with my initial approach. If we have an executor with T=50 threads max, a number of variants V=10 and a number of contingencies C=100; then with my approach, the executor will always have 10 threads running (one on each variant), and 40 threads blocked by tasks that do nothing (they wait for variants to become available). The executor becomes totally unusable (50 threads are used) even though 40 of its 50 threads block doing nothing.

A solution to this problem is to throttle (block) the thread submitting the tasks. However, if we do this, we can't have this thread be a thread of the same executor, because if the thread blocks and waits for tasks to have been executed, but at the same time tasks are not beeing executed because the executor is full of these threads waiting for tasks to be executed, then there is a deadlock. Using a different executor solves the problem (but we don't have another executor in the ComputationManager interface).

Another approach is to not do 1 task = 1 contingency; instead one task = 1 variant, and the task handles many contingencies one after another until all contingencies have been handled. This has the downside of having a fixed number of tasks for a variable amount of work, so tasks will be arbitrarily long (vs my initial approach of having a variable number of tasks that execute in a fixed time). Arbitrarily long tasks is a problem for scheduling other computations (especially in the case where the number of variants is greater or equal to the number of threads of the executor. in this case the whole executor becomes unusable for the rest of the system for the whole duration of the SecurityAnalysis computation).

sylvlecl:
> Arbitrarily long tasks is a problem for scheduling other computations

I understand but I don't think it's a problem in our case : although not stated in the API doc, I think the role of this executor is to execute computation tasks, by which I imply long running tasks.

A possibility would be for the computation manager to provide several executors, like for instance rx java framework which provides `Schedulers.io()` and `Schedulers.computation()` ?

jonenst:
I implemented throttling of the thread generating the tasks, and running it on a separate thread pool. I didn't use the static ForkJoinPool.commonPool() because it's meant for computation, instead I created a static pool for SecurityAnalysisImpl.

jonenst:
> A possibility would be for the computation manager to provide several executors, like for instance rx java framework which provides `Schedulers.io()` and `Schedulers.computation()` ?

Yes, using rxjava or taking ideas from rxjava is good. It seems important to have a bounded compute pool (threads never block), and an unbounded pool where threads can block (and with an unbounded pool you don't have to consider dependencies between tasks, you can't get a deadlock; a downside is that you can get a very high number of threads). Or we can replace the one unbounded pool by a series of bounded pools where tasks may only depend on (=block on the completion of) task in an another pool (with no cycles between dependencies across pools obviously..)

jonenst:
> > Arbitrarily long tasks is a problem for scheduling other computations
> 
> I understand but I don't think it's a problem in our case : although not stated in the API doc, I think the role of this executor is to execute computation tasks, by which I imply long running tasks.

I think there's a difference between computationManager.execute() and computationManager.getExecutor().execute(). For example in MpiComputationManager, the former is executed on "computationExecutor", the latter on "applicationExecutor". I think that what you are saying is true for computationManager.execute() but not for computationManager.getExecutor().

For info, LocalComputationManager uses 2 types of executors:
- the one given as an argument in the constructor (default ForkjoinPool.commonPool()) for 2 use cases:
  - returned by LocalComputatoinManager.getExecutor
  - used by execute to create 1 thread that starts many threads for the computations and waits for their completion
- a new (unbounded) one for each call to localComputationManager.execute(). With N command execution in the execute() call, N threads are immediately created on this new unbounded Executor. The first thing that these threads do is to block if more than available-core threads have already started across all those unbouned executors created by execute() calls. (so there is the same design flaw as my initial proposal)

And SlurmComputationManager returns null for getExecutor()..

jonenst:
So I pushed a version where:
- one "scheduler" thread is used for each security analysis. This thread clones the variants, submits the loadflow tasks, submits the violation checking tasks. This thread is in a static fixed thread pool for SecurityAnalysisImpl ( size controllable from PlatformConfig "security-analysis-impl.scheduler-pool-size", default 10 ). 
- MAX_VARIANTS_PER_ANALYSIS variants are created per analysis (controllable from PlatformConfig "security-analysis-impl.max-variants-per-analysis", default 10) (less if availableCores or number of contingencies is lower).

I think we can discuss more and adapt this solution to fit our needs. What do you think ?

Among the things we can change, I see:
- Should we guarantee that the executor returned by ComputationManager.getExecutor() is different than the one used for ComputationManager.execute() ? If yes, then we can submit tasks on ComputationManager.getExecutor() that will block waiting for tasks submitted to  ComputationManager.execute() finish without risking deadlocks. And so we can remove the scheduler thread pool that I just added to SecurityAnalsysImpl.
- should the MAX_VARIANTS_PER_ANALYSIS become a parameter of securityAnalsysImpl objects instead of using a static (probably).

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisImpl.java
```diff
@@ -69,70 +97,120 @@ public boolean removeInterceptor(SecurityAnalysisInterceptor interceptor) {
...
+    private void setContingencyOkAndCheckViolations(String postContStateId,
+            SecurityAnalysisResultBuilder resultBuilder, Contingency contingency,
+            LoadFlowResult lfResult) {
+        network.getVariantManager().setWorkingVariant(postContStateId);
+        synchronized (resultBuilder) {
+            resultBuilder.contingency(contingency)
+                    .setComputationOk(lfResult.isOk());
+            violationDetector.checkAll(contingency, network, resultBuilder::addViolation);
+            resultBuilder.endContingency();
+        }
```
mathbagu: I'm not sure this is a really good API design. The contingency method should create a PostContingencyResultBuilder that is thread safe. The endContingency should add the PostContingencyResult to the SAResult.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisImpl.java
```diff
@@ -69,70 +97,120 @@ public boolean removeInterceptor(SecurityAnalysisInterceptor interceptor) {
...
-                                        network.getVariantManager().cloneVariant(VariantManagerConstants.INITIAL_VARIANT_ID, postContStateId);
-                                        network.getVariantManager().setWorkingVariant(postContStateId);
+    private CompletableFuture<Void> submitAllLoadFlows(String workingStateId,
+            ContingenciesProvider contingenciesProvider,
+            LoadFlowParameters postContParameters,
+            SecurityAnalysisResultBuilder resultBuilder) {
+        List<Contingency> contingencies = contingenciesProvider.getContingencies(network);
+        List<String> variantIds = makeWorkingVariantsNames(contingencies.size());
+        BlockingQueue<String> queue = new ArrayBlockingQueue<>(variantIds.size(), false, variantIds);
+        network.getVariantManager().cloneVariant(workingStateId, variantIds);
```
mathbagu: You should not create a variant for all variant, but create 1 variant per worker. This is a regression on what you did yesterday?

mathbagu:
@jonenst said that:

> Before merging, options parsing must be fixed. Currently it requires
> max-variants-per-analysis : "20"
> # should be just 20, no quotes

We should a unit test to assert that it works properly

mathbagu:
@jonenst said that:

> Before merging, options parsing must be fixed. Currently it requires
> max-variants-per-analysis : "20"
> # should be just 20, no quotes

We should a unit test to assert that it works properly

mathbagu:
@geofjamg @sylvlecl The sonar issue is considered as [a bug](https://sonarcloud.io/project/issues?id=com.powsybl%3Apowsybl-core&pullRequest=1169&resolved=false&types=BUG). Do you think we have to fix it before the merge?

sylvlecl:
> 
> @geofjamg @sylvlecl The sonar issue is considered as [a bug](https://sonarcloud.io/project/issues?id=com.powsybl%3Apowsybl-core&pullRequest=1169&resolved=false&types=BUG). Do you think we have to fix it before the merge?

I think we can either overlook this, as our use is valid, or synchronize on "this", which will have the same effect,  to satisfy sonar. It makes the intent a little less clear, since what we want is really to synchronize the result builder use.
Maybe the best tradeoff is to synchronize the method `setContingencyOkAndCheckViolations`.

jonenst:
Can't a SecurityAnalysisImpl  be used to run multiple security analyses on the same network in parallel? In this case synchronizing on "this" instead of the builder is pessimistic

sylvlecl:
> 
> 
> Can't a SecurityAnalysisImpl be used to run multiple security analyses on the same network in parallel? In this case synchronizing on "this" instead of the builder is pessimistic

You're right that in theory it could. If someone does this however, he gets exposed to other risks, like changing the multithread mode while the other security analysis is running, or cloning variants while others are being used by the other security analysis ...
Maybe this should be discouraged.

PR #1201, opened 2020-03-03
Cgmes update simple update sv file

kaltakovae:

Bug fix

State Variable Adder does not include complete information.

State Variable Adder includes complete information about cgmes model's State Variables.


triple-store/triple-store-api/src/main/java/com/powsybl/triplestore/api/PropertyBag.java
```diff
@@ -195,10 +197,19 @@ public boolean isClassProperty(String name) {
         return classPropertyNames.contains(name);
     }
 
+    public void setMultivaluedProperty(List<String> multiValuedPropertyNames) {
+        this.multiValuedPropertyNames = Objects.requireNonNull(multiValuedPropertyNames);
```
MioRtia: set this as a copy to be protected against unsafe changes:
```java
this.multiValuedPropertyNames = new ArrayList<>(Objects.requireNonNull(multiValuePropertyNames));
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -17,42 +26,122 @@
...
+        this.originalTopologicalIslands = cgmes.topologicalIslands();
+        this.originalSVcontext = originalSVcontext();
+        this.boundaryNodesFromDanglingLines = boundaryNodesFromDanglingLines();
     }
 
-    public static void add(Network n, CgmesModel cgmes) {
-        // TODO Add full model data with proper profile (StateVariables)
-        // FullModel is defined in ModelDescription:
-        // http://iec.ch/TC57/61970-552/ModelDescription/1#
+    private String originalSVcontext() {
```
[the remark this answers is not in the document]
mathbagu: I don't think so because it uses the `cgmes` attribute.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -61,30 +150,51 @@ public static void add(Network n, CgmesModel cgmes) {
...
             if (p != null) {
-                powerFlows.add(createPowerFlowProperties(cgmes, s.getTerminal()));
+                powerFlows.add(p);
             }
         }
-        cgmes.add(CgmesSubset.STATE_VARIABLES, "SvPowerFlow", powerFlows);
 
+        // PowerFlow at boundaries set as it was in original cgmes.
+        for (PropertyBag terminal : originalTerminals) {
+            Objects.requireNonNull(terminal);
```
mathbagu: I don't think this is really needed.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -100,20 +210,109 @@ public static void add(Network n, CgmesModel cgmes) {
...
+                // need to check if connectable was already added
+                if (!addedConnectables.contains(c.getId())) {
+                    addedConnectables.add(c.getId());
+                    if (c.getId() != null) {
+                        PropertyBag p = new PropertyBag(SV_SVSTATUS_PROPERTIES);
+                        for (Terminal t : c.getTerminals()) {
+                            if (t == null) {
+                                continue;
+                            }
+                            p.put(IN_SERVICE, String.valueOf(t.isConnected()));
```
mathbagu: Consider replacing `String.valueOf` by `Boolean.toString`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -100,20 +210,109 @@ public static void add(Network n, CgmesModel cgmes) {
...
+                }
+            });
+        }
+        cgmes.add(originalSVcontext, "SvStatus", svStatus);
+    }
+
+    private Map<String, String> boundaryNodesFromDanglingLines() {
+        Map<String, String> nodesFromLines = new HashMap<>();
+        List<String> boundaryNodes = new ArrayList<>();
+        cgmes.boundaryNodes().forEach(node -> boundaryNodes.add(node.getId("Node")));
```
mathbagu: Use the stream API to collect boundadyNodes:
```
List<String> boundaryNodes = cgmes.boundaryNodes().map(X::getId).collect(Collectors.toList());
```
Replace X by the name of the class of `node`.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -127,22 +326,115 @@ private static PropertyBag createPowerFlowProperties(CgmesModel cgmes, Terminal
...
+            cgmes.add(originalSVcontext, CgmesNames.FULL_MODEL, fullModelSV);
+        }
+    }
+
+    private String getMultivaluedProperty(PropertyBags pb, String multivaluedPropertyName) {
+        // for properties such as "md:Model.DependentOn" which might have arbitrary
+        // number of values, SPARQL will return multiple result sets. We will loop
+        // through all and collect all values for multiValued property.
+        List<String> list = new ArrayList<>();
+        pb.forEach(m -> list.add(m.getId(multivaluedPropertyName)));
```
mathbagu: Use the stream API to initialize `list`. If there is no stream() method in PropertyBags, you should consider to add one.

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesSubset.java
```diff
@@ -71,11 +72,26 @@ public String getIdentifier() {
...
 
     private static boolean isBoundary(String contextName) {
         return contextName.contains("_BD") || contextName.contains("BOUNDARY");
     }
+
+    private final String identifier;
+    private final String validName0;
+    private final String validName1;
+    private String profileName;
+    private final Map<String, String> profile = ImmutableMap.of(
```
mathbagu: I think you should make this map static. Change also the case of `profile` to upper case.

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesSubset.java
```diff
@@ -71,11 +72,26 @@ public String getIdentifier() {
...
     }
 
     private static boolean isBoundary(String contextName) {
         return contextName.contains("_BD") || contextName.contains("BOUNDARY");
     }
+
+    private final String identifier;
+    private final String validName0;
+    private final String validName1;
+    private String profileName;
```
mathbagu: should probably be final

cgmes/cgmes-model/src/main/resources/CIM14.sparql
```diff
@@ -74,6 +81,10 @@ OPTIONAL { GRAPH ?graphSV {
         cim:SvPowerFlow.Terminal ?Terminal ;
         cim:SvPowerFlow.p ?p ;
         cim:SvPowerFlow.q ?q
+} OPTIONAL {?SvStatus
+		a cim:SvStatus ;
+		cim:SvStatus.ConductingEquipment ?ConductingEquipment ;
+		cim:SvStatus.inService ?inService .
```
mathbagu: Fix indentation: do not use tabs

cgmes/cgmes-model/src/main/resources/CIM16.sparql
```diff
@@ -144,6 +167,10 @@ OPTIONAL { GRAPH ?graphSV {
         cim:SvPowerFlow.Terminal ?Terminal ;
         cim:SvPowerFlow.p ?p ;
         cim:SvPowerFlow.q ?q
+} OPTIONAL {?SvStatus
+		a cim:SvStatus ;
+		cim:SvStatus.ConductingEquipment ?ConductingEquipment ;
+		cim:SvStatus.inService ?inService .
```
mathbagu: Same remark about tabs

cgmes/cgmes-model/src/main/resources/CIM16.sparql
```diff
@@ -199,6 +226,16 @@ OPTIONAL { GRAPH ?graphSVT {
 }}
 }
 
+# query: topologicalIslands
+SELECT *
+{ GRAPH ?graph { 
+	?TopologicalIsland
+		a cim:TopologicalIsland ;
+		cim:IdentifiedObject.name ?name ;
```
mathbagu: Same remakr

triple-store/triple-store-api/src/main/java/com/powsybl/triplestore/api/PropertyBag.java
```diff
@@ -173,9 +174,10 @@ public final boolean equals(Object obj) {
 
     public boolean isResource(String name) {
         // TODO do not rely on property name, use metadata or answer based on value?
-        return name.equals("TopologicalNode") || name.equals("Terminal")
-                || name.equals("ShuntCompensator") || name.equals("TapChanger")
-                || resourceNames.contains(name);
+        List<String> list = Stream.of("TopologicalNode", "Terminal", "ShuntCompensator",
+            "TapChanger", "ConductingEquipment", "Model.DependentOn", "TopologicalNodes",
+            "AngleRefTopologicalNode").collect(Collectors.toList());
```
mathbagu: Replace `Stream.of` by `Arrays.asList`. This list should be static and final ?

triple-store/triple-store-api/src/main/java/com/powsybl/triplestore/api/PropertyBag.java
```diff
@@ -195,10 +197,19 @@ public boolean isClassProperty(String name) {
...
+
+    public boolean isMultivaluedProperty(String name) {
+        return multiValuedPropertyNames.contains(name);
+    }
+
     private final List<String> propertyNames;
     private final boolean removeInitialUnderscoreForIdentifiers;
     private List<String> resourceNames = new ArrayList<>();
     private List<String> classPropertyNames = new ArrayList<>();
+    private List<String> multiValuedPropertyNames = new ArrayList<>();
```
mathbagu: All these lists should be final?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -17,42 +26,122 @@
...
+
+        addStatusToCgmes();
+    }
+
+    private void addVoltagesForTopologicalNodes() {
         PropertyBags voltages = new PropertyBags();
-        for (Bus b : n.getBusBreakerView().getBuses()) {
+        // add voltages for TpNodes existing in the Model
+        for (PropertyBag tn : cgmes.topologicalNodes()) {
+            Bus b = n.getBusBreakerView().getBus(tn.getId(CgmesNames.TOPOLOGICAL_NODE));
```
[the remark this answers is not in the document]
kaltakovae: Correct, NodeBreaker view require further investigation to map correctly Topological Nodes. It will be done with the next iteration.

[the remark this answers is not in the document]
MioRtia: Okay!

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -100,20 +210,109 @@ public static void add(Network n, CgmesModel cgmes) {
                 tapSteps.add(p);
             }
         }
-        cgmes.add(CgmesSubset.STATE_VARIABLES, "SvTapStep", tapSteps);
+
+        for (ThreeWindingsTransformer t : n.getThreeWindingsTransformers()) {
+            PropertyBag p = new PropertyBag(svTapStepProperties);
+            Collections.unmodifiableList(Arrays.asList(t.getLeg1(), t.getLeg2(), t.getLeg3())).forEach(leg -> {
```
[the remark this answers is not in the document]
kaltakovae: you're right, we don't keep the object here, the list is already immutable. I'll clean out the 'unmodifiable'

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -100,20 +210,109 @@ public static void add(Network n, CgmesModel cgmes) {
...
+        // create SvStatus, iterate on Connectables, check Terminal status, add
+        // to SvStatus
+        PropertyBags svStatus = new PropertyBags();
+        Set<String> addedConnectables = new HashSet<>();
+        for (VoltageLevel v : n.getVoltageLevels()) {
+            for (Connectable<?> c : v.getConnectables()) {
+                // need to check if connectable was already added
+                if (!addedConnectables.contains(c.getId())) {
+                    addedConnectables.add(c.getId());
+                    if (c.getId() != null) {
```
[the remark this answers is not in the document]
kaltakovae: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -100,20 +210,109 @@ public static void add(Network n, CgmesModel cgmes) {
...
+                        PropertyBag p = new PropertyBag(SV_SVSTATUS_PROPERTIES);
+                        for (Terminal t : c.getTerminals()) {
+                            if (t == null) {
```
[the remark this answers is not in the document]
kaltakovae: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -127,22 +326,115 @@ private static PropertyBag createPowerFlowProperties(CgmesModel cgmes, Terminal
...
+                getMultivaluedProperty(originalFullModel, "DependentOn"));
+            newModelDescription.put(CgmesNames.PROFILE, originalFullModel.get(0).getId("profile"));
+            newModelDescription.put(CgmesNames.MODELING_AUTHORITY_SET,
+                originalFullModel.get(0).getId("modelingAuthoritySet"));
+            fullModelSV.add(newModelDescription);
+            cgmes.add(originalSVcontext, CgmesNames.FULL_MODEL, fullModelSV);
+        }
+    }
+
+    private String getMultivaluedProperty(PropertyBags pb, String multivaluedPropertyName) {
```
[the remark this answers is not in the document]
kaltakovae: Done

triple-store/triple-store-api/src/main/java/com/powsybl/triplestore/api/PropertyBag.java
kaltakovae: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
kaltakovae: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
kaltakovae: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
kaltakovae: Done. With the only difference: we we used arrow operator, not double colon
List<String> boundaryNodes = cgmes.boundaryNodes().stream().map(n -> n.getId("Node")).collect(Collectors.toList());

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
kaltakovae: Done. PropertyBags.pluckLocals(String property) does what we need, I missed it initially.

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesSubset.java
kaltakovae: profile is used in the constructor of the enum

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesSubset.java
kaltakovae: Done

cgmes/cgmes-model/src/main/resources/CIM14.sparql
kaltakovae: Done

cgmes/cgmes-model/src/main/resources/CIM16.sparql
kaltakovae: Done

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesSubset.java
kaltakovae: Done

cgmes/cgmes-model/src/main/resources/CIM16.sparql
kaltakovae: Done

triple-store/triple-store-api/src/main/java/com/powsybl/triplestore/api/PropertyBag.java
kaltakovae: Done

triple-store/triple-store-api/src/main/java/com/powsybl/triplestore/api/PropertyBag.java
kaltakovae: yes, we set them as finals

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -100,20 +221,103 @@ public static void add(Network n, CgmesModel cgmes) {
...
+
+    private void addStatusToCgmes() {
+        // create SvStatus, iterate on Connectables, check Terminal status, add
+        // to SvStatus
+        PropertyBags svStatus = new PropertyBags();
+        Set<String> addedConnectables = new HashSet<>();
+        for (VoltageLevel v : network.getVoltageLevels()) {
+            for (Connectable<?> c : v.getConnectables()) {
+                // need to check if connectable was already added
+                if (!addedConnectables.contains(c.getId())) {
```
MioRtia: @mathbagu We discussed it but I don't really like that `getConnectables()` does not ensure that the elements are distinct
@kaltakovae Do you have a case in which the same `Connectable` is met several times? In any case, I think you should drop this check, I will make another PR to ensure the list/stream is distinct. Besides, you can use the stream API here.

kaltakovae: Currently we observe multiple duplicates, e.g. for each Line or Transformer getConnectables() duplicates results. We've tested with MicroGrid and HOPS test cases.
But, even if getConnectables() results in distincts, we still need the check for duplicates here:
Without this check any Connectables which are in different VoltageLevels (e.g Transformers) will be duplicated or triplicated.

MioRtia: Thank you for your input! I made a PR to check for duplicates in a same voltage level. As for your code, thank your for the clarification! I think you can optimize it as such:
```java
PropertyBags svStatus = new PropertyBags();
        network.getVoltageLevelStream()
                .flatMap(VoltageLevel::getConnectableStream)
                .distinct()
                .forEach(c -> {
                    PropertyBag p = new PropertyBag(SV_SVSTATUS_PROPERTIES);
                    for (Terminal t : ((Connectable<?>) c).getTerminals()) {
                        p.put(IN_SERVICE, Boolean.toString(t.isConnected()));
                    }
                    p.put(CgmesNames.CONDUCTING_EQUIPMENT, c.getId());
                    svStatus.add(p);
                });
```

zamarrenolm: About finding duplicates:
There is no `getConnectables` at `Network` level, so to obtain all connectables of the network we must iterate first over the voltage levels, then its connectables.
This means that we have to account for potential duplicates: lines and 2w transformers will be found twice, once for every end, and 3w transformers will appear three times during the iteration.
So, unless we have a `getConnectables` at `Network` level, we have to take into account for potential duplicates. 
An alternative could be to iterate over specific types of equipment (`getLines`, `getTransformers`, `getGenerators`, `getLoads`, ...) but that would mean to have a long list of method calls, and leave the risk of omitting new types when they are added to the Network (in the future). I mean, leaving a long list of specific dependencies when a generic treatment over all kind of equipment can be done.

kaltakovae: @MioRtia  Thank you Miora, this mathes perfectly to what we wanted. I've done the change.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -7,86 +7,205 @@
...
+                p.put(CgmesNames.ANGLE, fs(0.0));
+                p.put(CgmesNames.VOLTAGE, fs(0.0));
+                p.put(CgmesNames.TOPOLOGICAL_NODE, entry.getValue());
+            }
+            voltages.add(p);
+        });
+        cgmes.add(originalSVcontext, "SvVoltage", voltages);
+    }
+
+    private void addPowerFlowToCgmes() {
```
MioRtia: To lower this method's complexity, create an intermediary method:
```java
private <I extends Injection> void addInjectionPowerFlowToCgmes(PropertyBags powerFlows, Iterable<I> injectionStream) {
        injectionStream.forEach(i -> {
            PropertyBag p = createPowerFlowProperties(i.getTerminal());
            if (p != null) {
                powerFlows.add(p);
            } else if (i instanceof Load) {
                // FIXME CGMES SvInjection objects created as loads
                LOG.error("No SvPowerFlow created for load {}", i.getId());
            }
        });
    }
```

You can then call:
```java
PropertyBags powerFlows = new PropertyBags();
addInjectionPowerFlowToCgmes(powerFlows, network.getLoads());
addInjectionPowerFlowToCgmes(powerFlows, network.getGenerators());
addInjectionPowerFlowToCgmes(powerFlows, network.getShuntCompensators());
addInjectionPowerFlowToCgmes(powerFlows, network.getStaticVarCompensators());
addInjectionPowerFlowToCgmes(powerFlows, network.getBatteries());
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -127,22 +325,106 @@ private static PropertyBag createPowerFlowProperties(CgmesModel cgmes, Terminal
...
+                    pbs.add(pb);
+                    byTopologicalIslandId.put(island, pbs);
+                }
+            });
+            // now we can process all TPNodes from each island, and put them in one
+            // multivaluedProperty.
+            PropertyBags topologicalIslands = new PropertyBags();
+            byTopologicalIslandId.values().forEach(island -> {
+                PropertyBag topologicalIsland = new PropertyBag(SV_TOPOLOGICALISLAND_PROPERTIES);
+                topologicalIsland.setClassPropertyNames(Arrays.asList(CgmesNames.NAME));
```
MioRtia: You can replace `Arrays.asList(CgmesNames.NAME)` by `Collections.singletonList(...)`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -127,22 +325,106 @@ private static PropertyBag createPowerFlowProperties(CgmesModel cgmes, Terminal
...
+                topologicalIsland.setMultivaluedProperty(Arrays.asList("TopologicalNodes"));
```
MioRtia: You can replace `Arrays.asList(TopologicalNodes)` by `Collections.singletonList(...)`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -127,22 +325,106 @@ private static PropertyBag createPowerFlowProperties(CgmesModel cgmes, Terminal
...
+        // number of values, SPARQL will return multiple result sets.
+        // All are equal, except the multiValued property value.
+        if (!originalFullModel.isEmpty()) {
+            PropertyBag newModelDescription = new PropertyBag(SV_FULLMODEL_PROPERTIES);
+            newModelDescription
+                .setClassPropertyNames(
+                    Arrays.asList(CgmesNames.SCENARIO_TIME, CgmesNames.CREATED, CgmesNames.DESCRIPTION,
+                        CgmesNames.VERSION, CgmesNames.DEPENDENT_ON, CgmesNames.PROFILE,
+                        CgmesNames.MODELING_AUTHORITY_SET));
+            newModelDescription.setMultivaluedProperty(Arrays.asList(CgmesNames.DEPENDENT_ON));
```
MioRtia: You can replace `Arrays.asList(CgmesNames.DEPENDENT_ON)` by `Collections.singletonList(...)`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -127,22 +325,106 @@ private static PropertyBag createPowerFlowProperties(CgmesModel cgmes, Terminal
...
-        return iidmBusId;
-    }
+    private CgmesModel cgmes;
+    private Network network;
+    private int cimVersion;
+    private PropertyBags originalTerminals;
+    private PropertyBags originalFullModel;
+    private PropertyBags originalTopologicalIslands;
+    private String originalSVcontext;
+    private Map<String, String> boundaryNodesFromDanglingLines;
```
MioRtia: All these fields can be final.

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesModel.java
```diff
@@ -29,6 +29,8 @@
 
     Properties getProperties();
 
+    PropertyBags fullModel(String cgmesProfile);
```
MioRtia: Add a default implementation to prevent breaking changes

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesModel.java
```diff
@@ -130,10 +132,16 @@
 
     PropertyBags dcTerminalsTP();
 
+    PropertyBags topologicalIslands();
+
+    PropertyBags graph();
```
MioRtia: Same remark

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesModel.java
```diff
@@ -130,10 +132,16 @@
...
+
     void clear(CgmesSubset subset);
 
     void add(CgmesSubset subset, String type, PropertyBags objects);
 
+    void add(String context, String type, PropertyBags objects);
```
MioRtia: Same remark

triple-store/triple-store-api/src/main/java/com/powsybl/triplestore/api/PropertyBag.java
```diff
@@ -184,24 +183,36 @@ public String namespacePrefix(String name) {
     }
 
     public void setResourceNames(List<String> resourceNames) {
-        this.resourceNames = Objects.requireNonNull(resourceNames);
+        this.resourceNames.addAll(Objects.requireNonNull(resourceNames));
```
MioRtia: If you want to set `this.resourceNames`, you will need to clear it before calling `addAll`:
```java
this.resourceNames.clear();
this.resourceNames.addAll(...);
```

triple-store/triple-store-api/src/main/java/com/powsybl/triplestore/api/PropertyBag.java
```diff
@@ -184,24 +183,36 @@ public String namespacePrefix(String name) {
...
     }
 
     public void setClassPropertyNames(List<String> classPropertyNames) {
-        this.classPropertyNames = Objects.requireNonNull(classPropertyNames);
+        this.classPropertyNames.addAll(Objects.requireNonNull(classPropertyNames));
```
MioRtia: Same remark

triple-store/triple-store-api/src/main/java/com/powsybl/triplestore/api/PropertyBag.java
```diff
@@ -184,24 +183,36 @@ public String namespacePrefix(String name) {
...
     }
 
     public boolean isClassProperty(String name) {
         return classPropertyNames.contains(name);
     }
 
+    public void setMultivaluedProperty(List<String> multiValuedPropertyNames) {
+        this.multiValuedPropertyNames.addAll(Objects.requireNonNull(multiValuedPropertyNames));
```
MioRtia: Same remark

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -7,86 +7,205 @@
...
+        this.boundaryNodesFromDanglingLines = boundaryNodesFromDanglingLines();
     }
 
-    public static void add(Network n, CgmesModel cgmes) {
-        // TODO Add full model data with proper profile (StateVariables)
-        // FullModel is defined in ModelDescription:
-        // http://iec.ch/TC57/61970-552/ModelDescription/1#
+    private String originalSVcontext() {
+        PropertyBags pbs = cgmes.graph();
+        PropertyBag defaultSvContext = new PropertyBag(Arrays.asList(CgmesNames.GRAPH));
```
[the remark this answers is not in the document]
kaltakovae: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -7,86 +7,205 @@
...
+            }
         }
-        cgmes.add(CgmesSubset.STATE_VARIABLES, "SvVoltage", voltages);
+        cgmes.add(originalSVcontext, "SvVoltage", voltages);
+    }
 
+    private void addVoltagesForBoundaryNodes() {
+        PropertyBags voltages = new PropertyBags();
+        // add voltages for TpNodes existing in the Model's boundaries
+        boundaryNodesFromDanglingLines.entrySet().forEach(entry -> {
```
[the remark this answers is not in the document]
kaltakovae: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
kaltakovae: Done.
Thank you Miora, that's really good, I'll try to use this pattern in future.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
kaltakovae: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
kaltakovae: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
kaltakovae: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
kaltakovae: Done

triple-store/triple-store-api/src/main/java/com/powsybl/triplestore/api/PropertyBag.java
kaltakovae: Done

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesModel.java
kaltakovae: Done

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesModel.java
kaltakovae: Done

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesModel.java
kaltakovae: Done

triple-store/triple-store-api/src/main/java/com/powsybl/triplestore/api/PropertyBag.java
kaltakovae: Done

triple-store/triple-store-api/src/main/java/com/powsybl/triplestore/api/PropertyBag.java
kaltakovae: Done

annetill:
Do we need a last functional validation or it is okay as it is ?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -100,20 +203,99 @@ public static void add(Network n, CgmesModel cgmes) {
                 tapSteps.add(p);
             }
         }
-        cgmes.add(CgmesSubset.STATE_VARIABLES, "SvTapStep", tapSteps);
+
+        for (ThreeWindingsTransformer t : network.getThreeWindingsTransformers()) {
+            PropertyBag p = new PropertyBag(svTapStepProperties);
+            Arrays.asList(t.getLeg1(), t.getLeg2(), t.getLeg3()).forEach(leg -> {
+                if (leg.getPhaseTapChanger() != null) {
```
mathbagu: You should use `leg.hasPhaseTapChanger()` that is more human readable.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -100,20 +203,99 @@ public static void add(Network n, CgmesModel cgmes) {
...
+                    p.put(tapChangerPositionName, is(leg.getPhaseTapChanger().getTapPosition()));
+                    p.put(CgmesNames.TAP_CHANGER, cgmes.phaseTapChangerForPowerTransformer(t.getId()));
+                    tapSteps.add(p);
+                } else if (leg.getRatioTapChanger() != null) {
```
mathbagu: Same improvement using `leg.hasRatioTapChanger()`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -100,20 +203,99 @@ public static void add(Network n, CgmesModel cgmes) {
...
+        cgmes.add(originalSVcontext, "SvTapStep", tapSteps);
+    }
+
+    private void addStatusToCgmes() {
+        // create SvStatus, iterate on Connectables, check Terminal status, add
+        // to SvStatus
+        PropertyBags svStatus = new PropertyBags();
+        network.getVoltageLevelStream()
+            .flatMap(VoltageLevel::getConnectableStream)
+            .distinct()
```
mathbagu: It could be a good idea to add this convenient feature to the network.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -100,20 +203,99 @@ public static void add(Network n, CgmesModel cgmes) {
...
+            .forEach(c -> {
+                PropertyBag p = new PropertyBag(SV_SVSTATUS_PROPERTIES);
+                for (Terminal t : ((Connectable<?>) c).getTerminals()) {
+                    p.put(IN_SERVICE, Boolean.toString(t.isConnected()));
```
mathbagu: Not sure this is OK: If a line is connected on side 1 but disconnected on side 2, I'm not sure the IN_SERVICE should be true.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -100,20 +203,99 @@ public static void add(Network n, CgmesModel cgmes) {
...
+                }
+                p.put(CgmesNames.CONDUCTING_EQUIPMENT, c.getId());
+                svStatus.add(p);
+            });
+
+        // SvStatus at boundaries set as it was in original cgmes.
+        for (PropertyBag terminal : originalTerminals) {
+            Objects.requireNonNull(terminal);
```
mathbagu: Are you sure this check is needed: in which case the terminal is null?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -127,22 +309,106 @@ private static PropertyBag createPowerFlowProperties(CgmesModel cgmes, Terminal
         return p;
     }
 
+    // added TopologicalIsland as it was in cgmes : original topology is
+    // preserved.
+    private void addTopologicalIslandsToCgmes() {
+        if (!originalTopologicalIslands.isEmpty()) {
```
mathbagu: This test is not necessary, the for-each at line 323 seems sufficient. If it's empty the `byTopologicalIslandId will be empty, and the for-each loop at line 336 won't loop.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
mathbagu: @MioRtia @zamarrenolm What do you think?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
kaltakovae: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
kaltakovae: Removed unnecessary check

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
kaltakovae: The hasPhaseTapChanger() is not available for Leg. Theses methods are in TapChangerParent interface. Would you like us to modify iidm Leg - add hasPhaseTapChanger() ? or make the method in TapChangerParent available for Leg?

kaltakovae: Alternatively, I can add private methods like this to StateVariableAdder:
    private boolean hasRatioTapChanger(Leg leg) {
        return leg.getRatioTapChanger() != null;
    }

    private boolean hasPhaseTapChanger(Leg leg) {
        return leg.getPhaseTapChanger() != null;
    }

MioRtia: I don't think it is a good idea to make this evolution in this PR... Keep the code here as it is for now, I am opening an issue to add convenient methods such as `hasPhaseTapChanger()` on APIs.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
MioRtia: Same remark

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
MioRtia: I will write an issue related to that

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
MioRtia: > Alternatively, I can add private methods like this to StateVariableAdder:
> private boolean hasRatioTapChanger(Leg leg) {
> return leg.getRatioTapChanger() != null;
> }
> 
> ```
> private boolean hasPhaseTapChanger(Leg leg) {
>     return leg.getPhaseTapChanger() != null;
> }
> ```

This is a good idea, it will make it easier to "clean" the code, once the API change will be implemented.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
MioRtia: I agree with @mathbagu, not sure what is put in the property bag, it always is the connectivity status of the last terminal, right? Is it what we want?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
kaltakovae: @Miora what do you think if we use this version:
    private boolean hasPhaseTapChanger(Object leg) {
        if (leg instanceof Leg) {
            return ((Leg) leg).getPhaseTapChanger() != null;
        } else if (leg instanceof TwoWindingsTransformer) {
            return ((TwoWindingsTransformer) leg).getPhaseTapChanger() != null;
        }
        return false;
    }

We could use it for both 2 and 3 wing transformes in addTapStepToCgmes()

MioRtia: it works for me!

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
zamarrenolm: The definition for `inService` attribute for a `ConductingEquipment` states that it is: _The in service status as a result of topology processing_.

For me: if a conducting equipment with multiple terminals has at least one terminal connected, it should be considered in-service. At least it makes sense from a modeling point of view, and its impact in the network could be calculated.

annetill: In fact Miora seems to say that we only keep the status of the last terminal we have checked. So what is the good solution to put ````inService```` when at least one terminal is connected ?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
kaltakovae: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
kaltakovae: We will change it for "any" terminal, instead of "last"

kaltakovae: Done

PR #1202, opened 2020-03-04
Introduction of non linear shunt compensators

### LINKED ISSUE #87: Add a gPerSection parameter in the ShuntCompensator class
Add an optional gPerSection attribute to the ShuntCompensator class.

**IIDM v1.0**
```
<iidm:shunt id="SC" name="SC" bPerSection="1.58025E-4" maximumSectionCount="10" currentSectionCount="0" connectableBus="SC_bus"/>
```

**IIDM v1.1**
```
<iidm:shunt id="SC" name="SC" bPerSection="1.58025E-4" gPerSection="1e-4" maximumSectionCount="10" currentSectionCount="0" connectableBus="SC_bus"/>
```

### LINKED ISSUE #900: Change of Shunt Compensator IIDM model

MioRtia:

fix #900 
fix #87 

Yes.

**Other information**:
This PR only deals with the change of API. XIIDM serialization and deserialization will be done in another PR.

### DIFF
--- a/ampl-converter/src/main/java/com/powsybl/ampl/converter/AmplNetworkWriter.java
+++ b/ampl-converter/src/main/java/com/powsybl/ampl/converter/AmplNetworkWriter.java
@@ -20,6 +20,7 @@ import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Set;
 
+import com.powsybl.commons.PowsyblException;
 import com.powsybl.iidm.network.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -1310,6 +1311,9 @@ public class AmplNetworkWriter {
                      new Column("sections count"))) {
             List<String> skipped = new ArrayList<>();
             for (ShuntCompensator sc : network.getShuntCompensators()) {
+                if (ShuntCompensatorModelType.NON_LINEAR.equals(sc.getModelType())) {
+                    throw new PowsyblException("Non linear shunt compensator not yet supported");
+                }
                 Terminal t = sc.getTerminal();
                 Bus bus = AmplUtil.getBus(t);
                 String busId = null;
@@ -1336,11 +1340,11 @@ public class AmplNetworkWriter {
                 double vb = t.getVoltageLevel().getNominalV();
                 double zb = vb * vb / AmplConstants.SB;
                 double b1 = 0;
-                double b2 = sc.getbPerSection() * sc.getMaximumSectionCount() * zb;
+                double b2 = sc.getModel(ShuntCompensatorLinearModel.class).getbPerSection() * sc.getModel().getMaximumSectionCount() * zb;
                 double minB = Math.min(b1, b2);
                 double maxB = Math.max(b1, b2);
                 double b = sc.getCurrentB() * zb;
-                int points = sc.getMaximumSectionCount() < 1 ? 0 : sc.getMaximumSectionCount() - 1;
+                int points = sc.getModel().getMaximumSectionCount() < 1 ? 0 : sc.getModel().getMaximumSectionCount() - 1;
                 formatter.writeCell(variantIndex)
                         .writeCell(num)
                         .writeCell(busNum)
--- a/ampl-converter/src/main/java/com/powsybl/ampl/converter/AmplNetworkWriter.java
+++ b/ampl-converter/src/main/java/com/powsybl/ampl/converter/AmplNetworkWriter.java
@@ -1340,11 +1340,11 @@ public class AmplNetworkWriter {
                 double vb = t.getVoltageLevel().getNominalV();
                 double zb = vb * vb / AmplConstants.SB;
                 double b1 = 0;
-                double b2 = sc.getModel(ShuntCompensatorLinearModel.class).getbPerSection() * sc.getModel().getMaximumSectionCount() * zb;
+                double b2 = sc.getModel(ShuntCompensatorLinearModel.class).getbPerSection() * sc.getMaximumSectionCount() * zb;
                 double minB = Math.min(b1, b2);
                 double maxB = Math.max(b1, b2);
                 double b = sc.getCurrentB() * zb;
-                int points = sc.getModel().getMaximumSectionCount() < 1 ? 0 : sc.getModel().getMaximumSectionCount() - 1;
+                int points = sc.getMaximumSectionCount() < 1 ? 0 : sc.getMaximumSectionCount() - 1;
                 formatter.writeCell(variantIndex)
                         .writeCell(num)
                         .writeCell(busNum)
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
@@ -17,19 +17,6 @@ package com.powsybl.iidm.network;
 public interface ShuntCompensator extends Injection<ShuntCompensator> {
 
     /**
-     * Get the maximum section count.
-     */
-    int getMaximumSectionCount();
-
-    /**
-     * Set the maximum number of section.
-     *
-     * @param maximumSectionCount the maximum number of section
-     * @return the shunt compensator to chain method calls.
-     */
-    ShuntCompensator setMaximumSectionCount(int maximumSectionCount);
-
-    /**
      * Get the current section count.
      * <p>
      * It is expected to be greater than one and lesser than or equal to the
@@ -53,35 +40,32 @@ public interface ShuntCompensator extends Injection<ShuntCompensator> {
     ShuntCompensator setCurrentSectionCount(int currentSectionCount);
 
     /**
-     * Get the susceptance per section in S.
-     */
-    double getbPerSection();
-
-    /**
-     * Set the susceptance per section in S.
-     *
-     * @param bPerSection the susceptance per section
-     * @return the shunt compensator to chain method calls.
-     */
-    ShuntCompensator setbPerSection(double bPerSection);
-
-    /**
-     * Get the susceptance for the maximum section count.
-     *
-     * @deprecated Use {@link #getbPerSection()} * {@link #getMaximumSectionCount()} instead.
-     */
-    @Deprecated
-    default double getMaximumB() {
-        return getbPerSection() * getMaximumSectionCount();
-    }
-
-    /**
-     * Get the susceptance for the current section counts.
+     * Get the susceptance for the current section count.
      * <p>
      * Depends on the working variant.
      * @see VariantManager
      */
     double getCurrentB();
+
+    /**
+     * Get the conductance for the current section count.
+     * <p>
+     * Depends on the working variant.
+     * @see VariantManager
+     */
+    double getCurrentG();
+
+    /**
+     * Get the model type of the shunt compensator (linear or non-linear)
+     */
+    ShuntCompensatorModelType getModelType();
+
+    /**
+     * Get the shunt model.
+     */
+    ShuntCompensatorModel getModel();
+
+    <M extends ShuntCompensatorModel> M getModel(Class<M> modelType);
 
     /**
      * Get the terminal used for regulation.
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
@@ -28,6 +28,11 @@ public interface ShuntCompensator extends Injection<ShuntCompensator> {
     int getCurrentSectionCount();
 
     /**
+     * Get the maximum number of sections.
+     */
+    int getMaximumSectionCount();
+
+    /**
      * Change the number of section.
      *
      * <p>
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorAdder.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorAdder.java
@@ -26,9 +26,9 @@ package com.powsybl.iidm.network;
  */
 public interface ShuntCompensatorAdder extends InjectionAdder<ShuntCompensatorAdder> {
 
-    ShuntCompensatorAdder setbPerSection(double bPerSection);
+    ShuntCompensatorLinearModelAdder newLinearModel();
 
-    ShuntCompensatorAdder setMaximumSectionCount(int maximumSectionCount);
+    ShuntCompensatorNonLinearModelAdder newNonLinearModel();
 
     ShuntCompensatorAdder setCurrentSectionCount(int currentSectionCount);
 
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorModel.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorModel.java
@@ -1,0 +1,30 @@
+/**
+ * Copyright (c) 2020, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.iidm.network;
+
+/**
+ * @author Miora Ralambotiana <miora.ralambotiana at rte-france.com>
+ */
+public interface ShuntCompensatorModel {
+
+    /**
+     * Get the maximum number of sections.
+     */
+    int getMaximumSectionCount();
+
+    /**
+     * Get the susceptance in S of the section with a given section number if it exists.
+     * Throw an exception if such a section does not exist.
+     */
+    double getB(int sectionNum);
+
+    /**
+     * Get the conductance in S of the section with a given section number if it exists.
+     * Throw an exception if such a section does not exist.
+     */
+    double getG(int sectionNum);
+}
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorModel.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorModel.java
@@ -12,11 +12,6 @@ package com.powsybl.iidm.network;
 public interface ShuntCompensatorModel {
 
     /**
-     * Get the maximum number of sections.
-     */
-    int getMaximumSectionCount();
-
-    /**
      * Get the susceptance in S of the section with a given section number if it exists.
      * Throw an exception if such a section does not exist.
      */
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorNonLinearModel.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorNonLinearModel.java
@@ -1,0 +1,63 @@
+/**
+ * Copyright (c) 2020, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.iidm.network;
+
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * @author Miora Ralambotiana <miora.ralambotiana at rte-france.com>
+ */
+public interface ShuntCompensatorNonLinearModel extends ShuntCompensatorModel {
+
+    interface Section {
+
+        /**
+         * Get the susceptance in S of the section.
+         */
+        double getB();
+
+        /**
+         * Get the conductance in S of the section.
+         */
+        double getG();
+    }
+
+    /**
+     * Get the maximum susceptance for a section in S.
+     */
+    double getMaximumB();
+
+    /**
+     * Get the maximum conductance for a section in S.
+     */
+    double getMaximumG();
+
+    /**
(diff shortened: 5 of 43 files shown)


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorLinearModelImpl.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Override
+    public void checkCurrentSection(int currentSectionCount) {
+        checkCurrentSection(shuntCompensator, currentSectionCount);
+    }
+
+    @Override
+    public void checkCurrentSection(Validable validable, int currentSectionCount) {
+        ValidationUtil.checkSections(validable, currentSectionCount, maximumSectionCount);
+    }
```
mathbagu: Are these two methods really public?

iidm/iidm-scripting/src/main/groovy/com/powsybl/iidm/network/scripting/ShuntCompensatorExtension.groovy
```diff
@@ -0,0 +1,62 @@
...
+class ShuntCompensatorExtension {
+
+    static double getMaximumB(ShuntCompensator self) {
+        if (ShuntCompensatorModelType.LINEAR == self.getModelType()) {
+            // forward bPerSection * maximumSectionCount for linear shunts
+            return self.getModel(ShuntCompensatorLinearModel.class).getbPerSection() * self.getModel().getMaximumSectionCount()
+        } else if (ShuntCompensatorModelType.NON_LINEAR == self.getModelType()) {
+            return self.getModel(ShuntCompensatorNonLinearModel.class).getMaximumB()
+        }
+        throw new PowsyblException("Unexpected shunt model type")
```
mathbagu: Add the model type in the message

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorModel.java
```diff
@@ -0,0 +1,30 @@
...
+     * Throw an exception if such a section does not exist.
+     */
+    double getB(int sectionNum);
+
+    /**
+     * Get the conductance in S of the section with a given section number if it exists.
+     * Throw an exception if such a section does not exist.
+     */
+    double getG(int sectionNum);
+}
```
agnesLeroy: I went to the website documentation about shunt compensators to better understand your changes, but it was not enough for me to understand... I guess you have edited that page of the website in a branch to update that doc for the next release: maybe we should discuss it.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorNonLinearModel.java
```diff
@@ -0,0 +1,63 @@
...
+     * If a section already exists for this section number, respectively replace its susceptance and conductance with the given susceptance and conductance.
+     */
+    ShuntCompensatorNonLinearModel addOrReplaceSection(int sectionNum, double b, double g);
+
+    /**
+     * Remove the section associated with a given section number if it exists <b>and</b> the current section count is different of the given section number.
+     * Else, throw an exception.
+     */
+    ShuntCompensatorNonLinearModel removeSection(int sectionNum);
+}
```
agnesLeroy: We could imagine having only this model and using it to create either linear or non-linear shunts, or am I wrong?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
```diff
@@ -53,35 +40,32 @@
...
-     * Get the susceptance for the current section counts.
-     * <p>
-     * Depends on the working variant.
-     * @see VariantManager
+     * Get the shunt model.
      */
-    double getCurrentB();
+    ShuntCompensatorModel getModel();
+
+    <M extends ShuntCompensatorModel> M getModel(Class<M> modelType);
```
mathbagu: Should you add a default behaviour? Or do we decide that the API breaks

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorAdder.java
```diff
@@ -26,9 +26,9 @@
  */
 public interface ShuntCompensatorAdder extends InjectionAdder<ShuntCompensatorAdder> {
 
-    ShuntCompensatorAdder setbPerSection(double bPerSection);
+    ShuntCompensatorLinearModelAdder newLinearModel();
 
-    ShuntCompensatorAdder setMaximumSectionCount(int maximumSectionCount);
+    ShuntCompensatorNonLinearModelAdder newNonLinearModel();
```
mathbagu: To be discussed: should we design the API like we did for the extension:
```
newModel(ShuntCompensatorLinearModelAdder.class);
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractShuntCompensatorModel.java
```diff
@@ -0,0 +1,49 @@
...
+ * @author Miora Ralambotiana <miora.ralambotiana at rte-france.com>
+ */
+abstract class AbstractShuntCompensatorModel implements ShuntCompensatorModelHolder {
+
+    protected ShuntCompensatorImpl shuntCompensator;
+
+    @Override
+    public void setShuntCompensator(ShuntCompensatorImpl shuntCompensator) {
+        this.shuntCompensator = Objects.requireNonNull(shuntCompensator);
+    }
```
mathbagu: I'm not sure about this: the model should be attached to the ShuntCompensator in the constructor, like it's done for tap changer

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractShuntCompensatorModel.java
```diff
@@ -0,0 +1,49 @@
...
+            throw new IllegalArgumentException("shunt compensator model type is null");
+        }
+        if (modelType.isInstance(this)) {
+            return modelType.cast(this);
+        }
+        throw new ValidationException(shuntCompensator, "incorrect shunt compensator model type " +
+                modelType.getName() + ", expected " + getClass());
+    }
+
+    void checkCurrentSection(int currentSectionCount) {
```
mathbagu: Not private-package, but protected instead?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractShuntCompensatorModel.java
```diff
@@ -0,0 +1,49 @@
...
+        checkCurrentSection(shuntCompensator, currentSectionCount);
+    }
+
+    abstract void checkCurrentSection(Validable validable, int currentSectionCount);
```
mathbagu: Same question there

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
```diff
@@ -46,16 +43,106 @@ protected String getTypeDescription() {
...
+                this.g = g;
+                return this;
+            }
+
+            @Override
+            public ShuntCompensatorNonLinearModelAdder endSection() {
+                ValidationUtil.checkSectionNumber(ShuntCompensatorAdderImpl.this, sectionNum);
+                if (sections.containsKey(sectionNum)) {
+                    throw new ValidationException(ShuntCompensatorAdderImpl.this, "a section is already defined at this number");
+                }
```
mathbagu: It's strange to not replace the section whereas it's totally allowed later? It's not really important but it's questioning me

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
```diff
@@ -92,16 +179,19 @@ public ShuntCompensatorAdder setTargetDeadband(double targetDeadband) {
     public ShuntCompensatorImpl add() {
         String id = checkAndGetUniqueId();
         TerminalExt terminal = checkAndGetTerminal();
-        ValidationUtil.checkbPerSection(this, bPerSection);
-        ValidationUtil.checkSections(this, currentSectionCount, maximumSectionCount);
+        if (model == null) {
+            throw new ValidationException(this, "the shunt compensator model has not been defined");
+        }
+        model.getModel(AbstractShuntCompensatorModel.class).checkCurrentSection(this, currentSectionCount);
```
mathbagu: model.getModel() is really stange design: is the variable name is the good one? The adder set this variable with a model implementation, so I don't understand why the model is able to return a model...

I think you look to the ReactiveLimitsHolder? The design is a bit different because we instanciate ReactiveLimitsHolderImpl?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorModelHolder.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+interface ShuntCompensatorModelHolder extends ShuntCompensatorModel {
+
+    ShuntCompensatorModelType getType();
+
+    ShuntCompensatorModel getModel();
+
+    <M extends ShuntCompensatorModel> M getModel(Class<M> modelType);
+
+    void setShuntCompensator(ShuntCompensatorImpl shuntCompensator);
```
mathbagu: Not sure this method should be available

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
MioRtia: For me, it would be clearer to break the API because I think keeping a mix between the old and new APIs would lead to more errors.

agnesLeroy: Why do you create getModel() and getModelType? Couldn't the model be able to inform about its type?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorAdder.java
agnesLeroy: Do you think that choosing the same design here could improve the user experience?

MioRtia: It is mostly to keep an easy way to add a custom model... but doesn't it make the API more complex? (as in, it may make it more difficult to understand what "model" refers to)? Maybe we should keep `newLinearModel`, `newNonLinearModel` and add `newModel(...)`?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
MioRtia: Not sure if choosing to replace a section in a built model and creating the same section twice while building the whole model is the same thing? I think the second case is more probably an error but I may be mistaken. We can delete this validation condition and consider it is allowed to rewrite a section while building the model.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
MioRtia: Yes, I may have not named the objects well, I will change a bit the design in my next commit

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractShuntCompensatorModel.java
MioRtia: It doesn't work as tapchangers are created when the transformer is already built, which is not the case for models (they are created at the same time since it does not make sense to create a shunt without model).

MioRtia: I can check however that the shunt is only defined once.

ampl-converter/src/main/java/com/powsybl/ampl/converter/AmplNetworkWriter.java
```diff
@@ -1336,11 +1340,11 @@ private void writeShunts(AmplExportContext context) throws IOException {
                 double vb = t.getVoltageLevel().getNominalV();
                 double zb = vb * vb / AmplConstants.SB;
                 double b1 = 0;
-                double b2 = sc.getbPerSection() * sc.getMaximumSectionCount() * zb;
+                double b2 = sc.getModel(ShuntCompensatorLinearModel.class).getbPerSection() * sc.getModel().getMaximumSectionCount() * zb;
```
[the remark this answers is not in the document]
MioRtia: Indeed, I will put in on the Shunt API.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractShuntCompensatorModel.java
MioRtia: After changing a bit the implementation, I deleted this method.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorModelHolder.java
MioRtia: See comment above

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorModel.java
MioRtia: I am keeping your comment for now since I have not yet submitted a documentation PR as the APIs are still in construction. I will get back to you after the APIs are fixed.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
MioRtia: Yes but when calling it (from `ShuntCompensator`), we do not have this information, hence these two methods to retrieve it.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorAdder.java
MioRtia: After some thinking, I do not think it is a good idea to implement a `newModel(...)` method, I am not sure we want to extend the possibilities of shunt models.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
agnesLeroy: Maybe  getModelType should be renamed getShuntType then?

MioRtia: Isn't it a bit redundant to put "shuntType" in a method that only exists for shunts? (I put `getModelType()` because `getType()` already exists and returns which kind of connectable it is).

iidm/iidm-scripting/src/main/groovy/com/powsybl/iidm/network/scripting/ShuntCompensatorExtension.groovy
```diff
@@ -0,0 +1,63 @@
...
+/**
+ *
+ * @author Miora Ralambotiana <miora.ralambotiana at rte-france.com>
+ */
+class ShuntCompensatorExtension {
+
+    static double getMaximumB(ShuntCompensator self) {
+        if (ShuntCompensatorModelType.LINEAR == self.getModelType()) {
+            // forward bPerSection * maximumSectionCount for linear shunts
+            return self.getModel(ShuntCompensatorLinearModel.class).getbPerSection() * self.getModel().getMaximumSectionCount()
```
mathbagu: Is it a bad idea to keep the getMaximumB in all shunt model and provide a convenient function?

MioRtia: The issue is that we used to have `getMaximumB` for linear shunts which returned the susceptance of the highest section (not the actual maximum susceptance) so if we put it in the shunt model, we would have several possibilities:
- either keep this behavior for linear shunt but for non linear shunt it returns the actual maximum susceptance
- or change the behavior and in any case, it returns the actual maximum susceptance
- or keep this behavior for linear shunt and for non linear shunt it returns the susceptance of the highest section too

In any case, I find it too confusing... What do you think?

MioRtia:
> Could you please check again if everything in the merging view is implemented? I think we should have adapter for the model since you have a backward link. Is it required to have this link?

You're right... I will think about it but I think I will have to implement an adapter for the shunt models.

iidm/iidm-scripting/src/main/groovy/com/powsybl/iidm/network/scripting/ShuntCompensatorExtension.groovy
agnesLeroy: My opinion would be that it should return the actual maximum susceptance regardless of the model, as indicated by its name. Maybe another function could return the susceptance of the highest section for linear shunts if needed.

MioRtia: I agree with you, but isn't it deceitful to change the behavior of the method? Or I guess we could say that the previous one was buggy as when it was first implemented, it didn't take into account the case where the susceptance per section was negative...

MioRtia:
> > Could you please check again if everything in the merging view is implemented? I think we should have adapter for the model since you have a backward link. Is it required to have this link?
> 
> You're right... I will think about it but I think I will have to implement an adapter for the shunt models.

@mathbagu After checking, I don't think it is necessary to have adapters for the model as there is no backward link in the API. However there is a backward link in the model adders so implementing adapters for them should be enough I think.

mathbagu:
> > > Could you please check again if everything in the merging view is implemented? I think we should have adapter for the model since you have a backward link. Is it required to have this link?
> > 
> > 
> > You're right... I will think about it but I think I will have to implement an adapter for the shunt models.
> 
> @mathbagu After checking, I don't think it is necessary to have adapters for the model as there is no backward link in the API. However there is a backward link in the model adders so implementing adapters for them should be enough I think.

If there is no backward link, you right, it should not be needed!

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
```diff
@@ -92,16 +179,22 @@ public ShuntCompensatorAdder setTargetDeadband(double targetDeadband) {
...
         ValidationUtil.checkRegulatingTerminal(this, regulatingTerminal, getNetwork());
         ValidationUtil.checkVoltageControl(this, voltageRegulatorOn, targetV);
         ValidationUtil.checkTargetDeadband(this, "shunt compensator", voltageRegulatorOn, targetDeadband);
         ShuntCompensatorImpl shunt
                 = new ShuntCompensatorImpl(getNetwork().getRef(),
-                id, getName(), isFictitious(), bPerSection, maximumSectionCount,
+                id, getName(), isFictitious(), model,
                 currentSectionCount, regulatingTerminal == null ? terminal : regulatingTerminal,
                 voltageRegulatorOn, targetV, targetDeadband);
+        model.setShuntCompensator(shunt);
```
mathbagu: I think this should be done in the constructor, what do you think?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorImpl.java
```diff
@@ -47,13 +39,12 @@
...
     ShuntCompensatorImpl(Ref<? extends VariantManagerHolder> network,
-                         String id, String name, boolean fictitious, double bPerSection, int maximumSectionCount,
+                         String id, String name, boolean fictitious, ShuntCompensatorModelWrapper model,
                          int currentSectionCount, TerminalExt regulatingTerminal, boolean voltageRegulatorOn,
                          double targetV, double targetDeadband) {
         super(id, name, fictitious);
         this.network = network;
-        this.bPerSection = bPerSection;
-        this.maximumSectionCount = maximumSectionCount;
+        this.model = model;
```
mathbagu: Attach the model this this shunt compensator, like it's done between HvdcLine and HvdcConverterStation

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorModel.java
MioRtia: I have created a PR for documentation (and assigned you as reviewer). Could you please take a look at it to see if it is clearer now?

agnesLeroy: @MioRtia I just approved that PR, it looks much clearer to me with your changes.

PR #1204, opened 2020-03-04
Fix bug: check target deadband when setting regulation status to prevent inconsistent data

MioRtia:

No

Bug fix

target deadband is checked when setting target deadband and in adders but not when setting regulation status which can lead to inconsistent data (regulation status set to `true` but undefined target deadband).

target deadband is now checked when setting regulation status

No, it follows the breaking change of #958

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/PhaseTapChangerImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/PhaseTapChangerImpl.java
@@ -58,7 +58,6 @@ class PhaseTapChangerImpl extends AbstractTapChanger<PhaseTapChangerParent, Phas
     @Override
     public PhaseTapChangerImpl setRegulating(boolean regulating) {
         ValidationUtil.checkPhaseTapChangerRegulation(parent, getRegulationMode(), getRegulationValue(), regulating, getRegulationTerminal(), getNetwork());
-        ValidationUtil.checkTargetDeadband(parent, type, regulating, getTargetDeadband());
 
         Set<TapChanger> tapChangers = new HashSet<TapChanger>();
         tapChangers.addAll(parent.getAllTapChangers());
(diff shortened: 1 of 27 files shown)


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/PhaseTapChangerImpl.java
```diff
@@ -58,6 +58,7 @@ protected NetworkImpl getNetwork() {
     @Override
     public PhaseTapChangerImpl setRegulating(boolean regulating) {
         ValidationUtil.checkPhaseTapChangerRegulation(parent, getRegulationMode(), getRegulationValue(), regulating, getRegulationTerminal(), getNetwork());
+        ValidationUtil.checkTargetDeadband(parent, type, regulating, getTargetDeadband());
```
agnesLeroy: This seems to imply that the target deadband should always be set before using setRegulating. Should we propose a function to set the PST as regulating, together with the definition of the target deadband?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/RatioTapChangerImpl.java
```diff
@@ -57,6 +57,7 @@ protected NetworkImpl getNetwork() {
     @Override
     public RatioTapChangerImpl setRegulating(boolean regulating) {
         ValidationUtil.checkRatioTapChangerRegulation(parent, regulating, regulationTerminal, getTargetV(), getNetwork());
+        ValidationUtil.checkTargetDeadband(parent, type, regulating, getTargetDeadband());
```
agnesLeroy: Same comment

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/PhaseTapChangerImpl.java
MioRtia: It is the same principle as `regulationValue`, `regulationMode` and `regulationTerminal`: `targetDeadband` must be defined before setting regulation status to `true`. To do otherwise, a method with all these parameters should be implemented since they are all dependent on each other and it increases risks of confusion when using the API. The pattern `setAttribute(attribute)` has been chosen to be user-friendly even if the drawback is to be careful of the set order.

MioRtia: I am adding some tests to check the exception is thrown if setters are not called in the right order.

PR #1205, opened 2020-03-05
Psse raw file loading

geofjamg:

No

feature

This PR aims at loading a simple PSS/E file (IEEE 14 bus) into an in memory data model.

**Other information**:


mathbagu:
@geofjamg If you agree with my fixes, I think we can merge this PR.

geofjamg:
> @geofjamg If you agree with my fixes, I think we can merge this PR.

Agreed, thanks for the fix!

psse/psse-model/src/main/java/com/powsybl/psse/model/PsseTransformer.java
```diff
@@ -0,0 +1,618 @@
...
+        private double x12;
+
+        @Parsed(index = 2)
+        private double sbase12 = Double.NaN;
+
+        @Parsed(index = 3)
+        private double r23 = 0;
+
+        @Parsed(index = 4)
+        private double x23 = Double.NaN;
```
annetill: Why do you prefer to set the default value to NaN here whereas no default value is allowed as for ````x12```` ?

psse/psse-model/src/main/java/com/powsybl/psse/model/PsseTransformer.java
```diff
@@ -0,0 +1,618 @@
...
+
+        @Parsed(index = 5)
+        private double sbase23 = Double.NaN;
+
+        @Parsed(index = 6)
+        private double r31 = 0;
+
+        @Parsed(index = 7)
+        private double x31 = Double.NaN;
```
annetill: Same comment, no default value is allowed for ````x31````.

psse/psse-model/src/main/java/com/powsybl/psse/model/PsseTransformer.java
```diff
@@ -0,0 +1,618 @@
...
+
+        @Parsed(index = 0)
+        private double r12 = 0;
+
+        @Parsed(index = 1)
+        @Validate
+        private double x12;
+
+        @Parsed(index = 2)
+        private double sbase12 = Double.NaN;
```
annetill: The default value is the whole system base in MVA, how to set it correctly ?

psse/psse-model/src/main/java/com/powsybl/psse/model/PsseTransformer.java
```diff
@@ -0,0 +1,618 @@
...
+
+        @Parsed(index = 3)
+        private double r23 = 0;
+
+        @Parsed(index = 4)
+        private double x23 = Double.NaN;
+
+        @Parsed(index = 5)
+        private double sbase23 = Double.NaN;
```
annetill: The default value is the whole system base in MVA, how to set it correctly ?

psse/psse-model/src/main/java/com/powsybl/psse/model/PsseTransformer.java
```diff
@@ -0,0 +1,618 @@
...
+
+        @Parsed(index = 6)
+        private double r31 = 0;
+
+        @Parsed(index = 7)
+        private double x31 = Double.NaN;
+
+        @Parsed(index = 8)
+        private double sbase31 = Double.NaN;
```
annetill: The default value is the whole system base in MVA, how to set it correctly ?

psse/psse-model/src/main/java/com/powsybl/psse/model/PsseTransformer.java
```diff
@@ -0,0 +1,618 @@
...
+
+        public void setAnstar(double anstar) {
+            this.anstar = anstar;
+        }
+    }
+
+    public static class ThirdRecord {
+
+        @Parsed(index = 0)
+        private double windv = Double.NaN;
```
annetill: Here too the default value setting is more complex and need linked between objects. You put Nan to remember that a more complex treatement has to be done ?

psse/psse-model/src/main/java/com/powsybl/psse/model/PsseTransformer.java
```diff
@@ -0,0 +1,618 @@
...
+        private double ratc = 0;
+
+        @Parsed(index = 6)
+        private int cod = 0;
+
+        @Parsed(index = 7)
+        private int cont = 0;
+
+        @Parsed(index = 8)
+        private double rma = Double.NaN;
```
annetill: Same remarks as before: the default value setting is more complex and we have to keep that in mind for later.

psse/psse-model/src/main/java/com/powsybl/psse/model/PsseRawReader.java
```diff
@@ -0,0 +1,183 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ */
+public class PsseRawReader {
+
+    private static final Logger LOGGGER = LoggerFactory.getLogger(PsseRawReader.class);
```
annetill: Loggger ?

psse/psse-model/src/main/java/com/powsybl/psse/model/PsseTransformer.java
annetill: After the second record ?

psse/psse-model/src/main/java/com/powsybl/psse/model/PsseTransformer.java
annetill: At the end of all the records ?

psse/psse-model/src/main/java/com/powsybl/psse/model/PsseTransformer.java
```diff
@@ -0,0 +1,618 @@
...
+
+        @Parsed(index = 9)
+        private double rmi = Double.NaN;
```
annetill: Same remark.

psse/psse-model/src/main/java/com/powsybl/psse/model/PsseTransformer.java
```diff
@@ -0,0 +1,618 @@
...
+
+        @Parsed(index = 10)
+        private double vma = Double.NaN;
```
annetill: Same remark, we have to keep in mind.

psse/psse-model/src/main/java/com/powsybl/psse/model/PsseTransformer.java
```diff
@@ -0,0 +1,618 @@
...
+
+        @Parsed(index = 11)
+        private double vmi = Double.NaN;
```
annetill: Same remark, we have to keep in mind.

geofjamg:
> I think that small fixes are need for some default values setting. A more general question is about the treatment of default values that depends on other psse objects (as the sbase management) and about the "switch case" default value setting that also depends on other objects. We are going to do that after the raw reading or it sould be in this PR ?

For these cases, I let NaN as a value and default value decision could be taken in the conversion part.

Or that could also be done in a kind of post processin of data model loading. Maybe it would be better?

psse/psse-model/src/main/java/com/powsybl/psse/model/PsseTransformer.java
geofjamg: No default value is allowed in case of 3 windings transformers but this field is not present in case of 2 windings transformers.

psse/psse-model/src/main/java/com/powsybl/psse/model/PsseTransformer.java
geofjamg: Either at the conversion level or in a data model loading post processing.

psse/psse-model/src/main/java/com/powsybl/psse/model/PsseTransformer.java
geofjamg: Same

psse/psse-model/src/main/java/com/powsybl/psse/model/PsseRawReader.java
geofjamg: Fixed

psse/psse-model/src/main/java/com/powsybl/psse/model/PsseNonTransformerBranch.java
```diff
@@ -0,0 +1,285 @@
...
+    private double gj = 0;
+
+    @Parsed(index = 12)
+    private double bj = 0;
+
+    @Parsed(index = 13)
+    private int st = 1;
+
+    @Parsed(index = 14)
+    private int met = 0;
```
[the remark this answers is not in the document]
geofjamg: Fixed

annetill:
Hi Geoffroy, I have seen your fixes. About the linked default values we have to decide if you want to do it during the conversion or as a post-processing after the raw data loading. But, whatever we decide, we can also merge this PR as it is from now and do the rest in an other PR. What do you prefer ?

geofjamg:
> Hi Geoffroy, I have seen your fixes. About the linked default values we have to decide if you want to do it during the conversion or as a post-processing after the raw data loading. But, whatever we decide, we can also merge this PR as it is from now and do the rest in an other PR. What do you prefer ?

I started to work on post processor to set related default values. 
Not sure I can terminate that work today. I'm busy with other tasks. 
As you want, but maybe yes we can merge it as it is and create another PR for missing default values. @mathbagu too ok?

mathbagu:
Agreed!

PR #1207, opened 2020-03-06
Change the design of the SecurityAnalysisResultBuilder

mathbagu:

N/A

Refactoring

The `SecurityAnalysisResultBuilder` is not thread-safe by design. In the `SecurityAnalysisImpl`, we have to create a dedicated lock to avoid `IllegalStateException` (cf. #1169).

When a pre-contingency or a post-contingency result is created, we do not change the state of the `SecurityAnalysisResultBuilder` but we create independant sub builders. 

**Other information**:


mathbagu:
@jonenst thinks that the `addViolations(LimitViolationDetector)` should not be in the builder API. If I remove this method, I have to change a little bit the method `setPreContigencyOkAndCheckViolations` from
```
    private void setPreContigencyOkAndCheckViolations(SecurityAnalysisResultBuilder resultBuilder) {
        resultBuilder.preContingency()
                .setComputationOk(true)
                .addViolations(violationDetector)
                .endPreContingency();
    }
```
to
```
    private void setPreContigencyOkAndCheckViolations(SecurityAnalysisResultBuilder resultBuilder) {
        SecurityAnalysisResultBuilder.PreContingencyResultBuilder builder = resultBuilder.preContingency()
                .setComputationOk(true);
        violationDetector.checkAll(network, builder::addViolation);
        builder.endPreContingency();
    }
```

I on't remember why all the methods of the `LimitViolationDetector` API take a `Consumer` instead of simply returning a list of violations. If we change this API too (eventually in other PR), it could be possible to write:
```
    private void setPreContigencyOkAndCheckViolations(SecurityAnalysisResultBuilder resultBuilder) {
        List<LimitViolation> violations = violationDetector.checkAll(network);
        return resultBuilder.preContingency()
                .setComputationOk(true)
                .addViolations(violations)
                .endPreContingency();
    }
```
Do you agree with this proposal?

sylvlecl:
> I on't remember why all the methods of the `LimitViolationDetector` API take a `Consumer` instead of simply returning a list of violations. If we change this API too (eventually in other PR), it could be possible to write:

Reasons were flexibility and performance considerations. Flexibility obviously because you can pass any consumer, not just List::add, performance because if every method of the detector returns a list, then in order to build the whole network list you need to build and concatenate multiple small lists (one for each equipment), which seems an unnecessary burden.
I would be more in favor of proposing a utility method which builds and returns a list based on the detector, or adding new method(s) to the detector which would return lists, but not deleting the consumer based methods.

PR #1214, opened 2020-03-10
WIP : Merging view : implement getSynchronousComponent & getConnectedComponent

tadam50:

No

Using ComponentsManager::update method with MergingView  

ComponentsManager::update method with MergingView is not implemented

ComponentsManager::update method with MergingView can be used with a MergingView

**Other information**:


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
```diff
@@ -636,162 +633,39 @@ public BusViewImpl getBusView() {
...
+        protected ComponentImpl createComponent(int num, int size) {
+            return new ComponentImpl(num, size, ((NetworkImpl) network).ref);
         }
 
         @Override
-        protected void setComponentNumber(BusExt bus, int num) {
+        protected void setComponentNumber(Bus bus, int num) {
             Objects.requireNonNull(bus);
-            bus.setSynchronousComponentNumber(num);
+            ((BusExt) bus).setSynchronousComponentNumber(num);
```
mathbagu: Maybe the cast is not necessary: setConnectedComponentNumber is visible in the public API?

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/BusAdapter.java
```diff
@@ -168,51 +168,77 @@ public double getQ() {
...
-        throw MergingView.createNotImplementedException();
+    public void setSynchronousComponentNumber(int synchronousComponentNumber) {
+        getDelegate().setSynchronousComponentNumber(synchronousComponentNumber);
     }
 
     @Override
-    public Iterable<DanglingLine> getDanglingLines() {
-        throw MergingView.createNotImplementedException();
+    public int getSynchronousComponentNumber() {
+        return getDelegate().getSynchronousComponentNumber();
```
mathbagu: I'm not sure it's working. I think we should not change the connected/synchronous component number of the underlying bus. If I access to the underlying network, the component numbers could be wrong, because there are valid only for a CGM point of view, not an IGM.

I think the BusAdapter should handle its own copy of the component numbers. Same remark for the TerminalAdapter

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergingView.java
```diff
@@ -41,6 +41,48 @@ static PowsyblException createNotImplementedException() {
...
+
+        @Override
+        protected ComponentAdapter createComponent(int num, int size) {
+            return new ComponentAdapter(index, num, size);
+        }
+
+        @Override
+        protected void setComponentNumber(Bus bus, int num) {
+            Objects.requireNonNull(bus);
+            ((BusAdapter) bus).setConnectedComponentNumber(num);
```
mathbagu: I'm not sure the cast is needed (see above)

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergingView.java
```diff
@@ -41,6 +41,48 @@ static PowsyblException createNotImplementedException() {
...
+
+        @Override
+        protected ComponentAdapter createComponent(int num, int size) {
+            return new ComponentAdapter(index, num, size);
+        }
+
+        @Override
+        protected void setComponentNumber(Bus bus, int num) {
+            Objects.requireNonNull(bus);
+            ((BusAdapter) bus).setSynchronousComponentNumber(num);
```
mathbagu: Same remark

iidm/iidm-mergingview/src/test/java/com/powsybl/iidm/mergingview/BusAdapterTest.java
```diff
@@ -72,12 +73,53 @@ public void testSetterGetter() {
...
+        assertSame(busCalc2, generator1.getTerminal().getBusBreakerView().getBus());
+
+        assertTrue(busCalc1.isInMainConnectedComponent());
+        assertTrue(busCalc2.isInMainConnectedComponent());
+        assertEquals(0, busCalc1.getConnectedComponent().getNum());
+        assertEquals(0, busCalc2.getConnectedComponent().getNum());
+    }
+
+    @Test
+    public void testComponentVisitor() {
```
mathbagu: To be sure that everything is working, you have to merge two networks:
- the two network should have 2 connected components with two different sizes
- once merged, the two smallest connected components should be the main connected component.

In that case, from a CGM point of view, the main component is not the same than from the IGM point of view. This is exacly why I think we should add components in BusAdapter and TerminalAdapter.

But, if the topology changes, in the underlying network, you have to update both connected components (IGM and CGM).

PR #1218, opened 2020-03-10
Ensure that voltageLevel.getConnectables() and associated return collections of distinct elements

MioRtia:

Bug fix

`voltageLevel.getConnectables()` and `voltageLevel.getConnectableStream()` do not check if they have duplicates in their respective returned iterable and stream.

The returned iterable and stream contain distinct elements.

No

**Other information**:
It generated issues in voltage level with an internal transformer (for example) which:
- is not forbidden
- has been seen to happen by @agnesLeroy during UCTE imports

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractVoltageLevel.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractVoltageLevel.java
@@ -8,6 +8,7 @@ package com.powsybl.iidm.network.impl;
 
 import com.google.common.collect.FluentIterable;
 import com.google.common.collect.Lists;
+import com.google.common.primitives.Ints;
 import com.powsybl.iidm.network.*;
 
 import java.util.List;
@@ -123,11 +124,12 @@ abstract class AbstractVoltageLevel extends AbstractIdentifiable<VoltageLevel> i
     }
 
     @Override
-    public <T extends Connectable> FluentIterable<T> getConnectables(Class<T> clazz) {
+    public <T extends Connectable> Iterable<T> getConnectables(Class<T> clazz) {
         Iterable<Terminal> terminals = getTerminals();
         return FluentIterable.from(terminals)
                 .transform(Terminal::getConnectable)
-                .filter(clazz);
+                .filter(clazz)
+                .toSet();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractVoltageLevel.java
```diff
@@ -123,41 +124,45 @@ public VoltageLevel setHighVoltageLimit(double highVoltageLimit) {
     }
 
     @Override
-    public <T extends Connectable> FluentIterable<T> getConnectables(Class<T> clazz) {
+    public <T extends Connectable> Iterable<T> getConnectables(Class<T> clazz) {
```
agnesLeroy: Why do you change FluentIterable into Iterable here?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractVoltageLevel.java
```diff
@@ -123,41 +124,45 @@ public VoltageLevel setHighVoltageLimit(double highVoltageLimit) {
...
                 .filter(clazz::isInstance)
-                .map(clazz::cast);
+                .map(clazz::cast)
+                .distinct();
     }
 
     @Override
     public <T extends Connectable> int getConnectableCount(Class<T> clazz) {
-        return getConnectables(clazz).size();
+        return Ints.checkedCast(getConnectableStream(clazz).count());
```
agnesLeroy: Do I get it right that using count() is a better practice than using size(), and that count() returns a long so you need to cast it to an int?

MioRtia: No, I only used `count()` because `size()` does not exist for streams or for iterables (only for fluentiterables)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractVoltageLevel.java
MioRtia: `FluentIterable` extends `Iterable` and is more powerful (can chain more easily like streams for example). However, when ensuring its elements are distinct by using `toSet()`, I am returning a `ImmutableSet` which extends `Iterable` too but not `FluentIterable`

PR #1222, opened 2020-03-12
Set rdp to 0 when R1+R2 = 0 in mergedXnode (resp. xdp)

agnesLeroy:

This is a bug fix in the UCTE import/export feature.

In a mergedXnode, rdp is calculated as R1/(R1+R2).
When R1+R2 = 0, rdp is thus NaN. When converting a UCTE network to IIDM, for a mergedXnode with R (or X) = 0 on both sides, rdp is not written because NaN values are not serialized.
When importing the generated xiidm file into Powsybl, since rdp is a required attribute, a nullPointerException is raised. This also holds for xdp.

With this change, when R1+R2 = 0, rdp is set to 0 (resp. xdp). This fixes the isssue mentionned above. Besides, when splitting the merged Xnode again, we recover R1 & R2 = 0 (resp. X).


MioRtia:
@geofjamg Should we also change `MergedXNode` to forbid creation with `rdp` and `xdp` as `NaN`?

agnesLeroy:
@geofjamg what do you think of @MioRtia's question above?

geofjamg:
> @geofjamg what do you think of @MioRtia's question above?

That would be safer indeed.
But in that case you have to adapt ucte part.

PR #1229, opened 2020-03-17
[WIP] SecurityAnalysisResultBuilder, modify API to make it inheritable in other projects

jonenst:

NO

Bug Fix

After #1207 , it is no longer possible to implement inheritance on SecurityAnalysisResultBuilder.
the protected getContext() method was removed, we need it back
the CRTP on the builder types can't be override by subclasses

It is possible to subclass SecurityAnalysisResultBuilder properly

NO

**Other information**:
I am not satisfied with the current design :( What can be done to improve it ?
- In powsybl-core, we get this AbstractSecurityAnalysisResultBuilder class with very long templated arguments; and we get the empty class SecurityAnalsysisResultBuilder
- in projects subclassing it, they must subclass the new AbstractSecurityAnalysisResultBuilder, not the SecurityAnalsysisResultBuilder. If they want to implement shared functionnality in the subbuilders, they can't create a base class because they must inherit from the Base*Builder
- if we have code that works on SecurityAnalysisResultBuilder, it won't work on objects inheriting from AbstractBaseSecurityAnalysisResultBuilder. Instead, this code needs to know that it should use AbstractBaseSecurityAnalysisResultBuilder<? extends AbstractBaseSecurityAnalysisResultBuilder> . It's complex..

### DIFF
--- a/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/AbstractBaseSecurityAnalysisResultBuilder.java
+++ b/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/AbstractBaseSecurityAnalysisResultBuilder.java
@@ -1,0 +1,161 @@
+/**
+ * Copyright (c) 2018, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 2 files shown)


mathbagu:
@jonenst Can we close this PR? I think it will be totally replaced by Yichen's one?

sylvlecl:
Fixed by #1247

PR #1232, opened 2020-03-18
Create convenient methods to run through connectables from network

### LINKED ISSUE #1228: Create convenient methods to run through connectables from network
Feature

Connectables can only be run through at voltage level's level.

This should be possible at network's level without using `getVoltageLevelStream()`

Convenience

  - PowSyBl Version: 3.3.0-SNAPSHOT
  - OS Version: *

MioRtia:

fix #1228


iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergingViewIndex.java
```diff
@@ -132,6 +132,29 @@ Identifiable getIdentifiable(final Identifiable identifiable) {
         return getIdentifiableStream().collect(Collectors.toList());
     }
 
+    /** @return all Adapters according to all Connectables of a given type */
+    <C extends Connectable> Collection<C> getConnectables(Class<C> clazz) {
+        // Search Connectables of a given type into merging & working networks
+        if (clazz.isAssignableFrom(Line.class)) {
+            return getLines().stream().filter(clazz::isInstance).map(clazz::cast).collect(Collectors.toList());
+        } else if (clazz.isAssignableFrom(DanglingLine.class)) {
+            return getDanglingLines().stream().filter(clazz::isInstance).map(clazz::cast).collect(Collectors.toList());
```
mathbagu: Why calling getLines() or getDanglingLines() is not sufficient there?

MioRtia: We can not being sure that the returned stream will be of `clazz` instance. For example, if `class` is `TieLine.class`, `TieLine.class` is assignable from `Line.class` but the returned stream must still be filtered and cast to return only `TieLine`s.

mathbagu: Looking to the NetworkImpl, `getLines()` returns both Line and TieLine objects. The `NetworkImpl::getConnectableStream(Line.class)` return lines and tie lines:
```
    @Override
    public <C extends Connectable> Stream<C> getConnectableStream(Class<C> clazz) {
        return index.getAll().stream().filter(clazz::isInstance).map(clazz::cast);
    }

In case of the merging view, the getLine() should return lines, tie tiles and merged lines. I think the getLines() method is OK
For the danglingLines() method, it should return all the unmerged dangling lines. Once again, I think the getDanglingLines() method is OK

Last remark, this `TieLine.class.isAssignableFrom(Line.class)` test returns `false`. So I still think you can simply call `getLines()` and `getDanglingLines()`.
```

MioRtia: > Last remark, this TieLine.class.isAssignableFrom(Line.class) test returns false. So I still think you can simply call getLines() and getDanglingLines().

Okay, I corrected it in the last commit.

> In case of the merging view, the getLine() should return lines, tie tiles and merged lines. I think the getLines() method is OK
For the danglingLines() method, it should return all the unmerged dangling lines. Once again, I think the getDanglingLines() method is OK

I agree. However, when trying to retrieve a subclass of `Line` (in the example in the tests added in the last commit, `MergedLine`), it may be necessary to call `getLines` first and then filter and cast. What do you think?

PR #1233, opened 2020-03-18
Add nodeBreakerView.hasAttachedEquipment(node) method + Ensure an IIDM node number is valid before trying to use it in CGMES conversion

zamarrenolm:

Bug fix for conversion of isolated connectivity nodes in CGMES.

CGMES data contains an isolated connectivity node (no terminals attached to it, no related equipment).
Space for the connectivity node is reserved at IIDM node-breaker graph (a vertex is reserved).
During the conversion, the reserved vertex is never used (no equipment is attached to it).
After all equipment has been converted, voltages and angles from SV are put on the resulting IIDM data.
The CGMES SV data contains voltage and angle values for the connectivity node.
But the corresponding node in IIDM node-breaker graph is not valid.
An exception is raised.

Before trying to use the node from IIDM node-breaker graph, it is verified that the node number is valid.
To do it, we check that the nodes returned by the node-breaker view in IIDM contain the node number reserved.
There is no severe performance penalty on adding the check. It has been tested on a network with more than 14000 values for voltages. Without the check, all voltages and angles were set in 1400 ms. After adding the check, the cost is 1600 ms.

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/NodeConversion.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/NodeConversion.java
@@ -7,7 +7,6 @@
 
 package com.powsybl.cgmes.conversion.elements;
 
-import java.util.Arrays;
 import java.util.Objects;
 
 import org.slf4j.Logger;
@@ -72,10 +71,6 @@ public class NodeConversion extends AbstractIdentifiedObjectConversion {
             .newSubstation()
             .setId(context.namingStrategy().getId("Substation", substationId))
             .setName(substationName)
-            // TODO(mathbagu): Country should be optional. This will be done in another PR.
-            // A non-null country code must be set
-            // This is an arbitrary country code, Bangladesh code BD also matches with
-            // BounDary
             .setCountry(boundaryCountryCode())
             .add()
             .newVoltageLevel()
@@ -168,17 +163,15 @@ public class NodeConversion extends AbstractIdentifiedObjectConversion {
         VoltageLevel.NodeBreakerView topo = vl.getNodeBreakerView();
         String connectivityNode = id;
         int iidmNode = context.nodeMapping().iidmNodeForConnectivityNode(connectivityNode, vl);
-        if (!iidmNodeIsValid(topo, iidmNode)) {
+        if (!topo.hasAttachedEquipment(iidmNode)) {
             LOG.error("ConnectivityNode {} with voltage and angle is not valid in IIDM", connectivityNode);
             return;
         }
         // To obtain a bus for which we want to set voltage:
         // If there no Terminal at this IIDM node,
         // then find from it the first connected node with a Terminal
-        Terminal t = topo.getTerminal(iidmNode);
-        if (t == null) {
-            t = new FirstTerminalTraverser(topo, iidmNode).firstTerminal();
-        }
+        Terminal t = topo.getOptionalTerminal(iidmNode)
+                .orElseGet(() -> new FirstTerminalTraverser(topo, iidmNode).firstTerminal());
         if (t == null) {
             LOG.error("Can't find a Terminal to obtain a Bus to set Voltage, Angle. ConnectivityNode {}", id);
             return;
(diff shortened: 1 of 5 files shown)


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/NodeConversion.java
```diff
@@ -68,34 +69,34 @@ private VoltageLevel newBoundarySubstationVoltageLevel() {
...
-                .setNominalV(nominalVoltage)
-                .setTopologyKind(context.nodeBreaker() ? TopologyKind.NODE_BREAKER : TopologyKind.BUS_BREAKER)
-                .add();
+            .newSubstation()
+            .setId(context.namingStrategy().getId("Substation", substationId))
+            .setName(substationName)
+            // TODO(mathbagu): Country should be optional. This will be done in another PR.
+            // A non-null country code must be set
+            // This is an arbitrary country code, Bangladesh code BD also matches with
+            // BounDary
```
mathbagu: Not related to this PR, but this TODO could be removed.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/NodeConversion.java
```diff
@@ -186,12 +191,19 @@ public void setVoltageAngleNodeBreaker() {
                 return;
             }
             LOG.warn(
-                    "Can't find a calculated Bus to set Voltage, Angle, but found a configured Bus {}. Connectivity node {}",
-                    bus, id);
+                "Can't find a calculated Bus to set Voltage, Angle, but found a configured Bus {}. Connectivity node {}",
```
mathbagu: Not related to this PR, but CalculatedBus and ConfiguredBus are implementation specific and should not appear in this message

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/NodeConversion.java
```diff
@@ -186,12 +191,19 @@ public void setVoltageAngleNodeBreaker() {
...
+                bus, id);
         }
         setVoltageAngle(bus);
     }
 
+    private boolean iidmNodeIsValid(VoltageLevel.NodeBreakerView topo, int iidmNode) {
+        // A node is valid only if it has some equipment attached
+        // The list of valid nodes is obtained from the node-breaker view
+        // We assume the returned list is sorted
```
mathbagu: Not sure we should assume that, or we should update the documentation.
Anyway, is it the best way to check if a terminal is associated to this node: maybe we should add a convenient method to get an `Optional<Terminal>` from a node?

MioRtia:
You can now use the method `getOptionalTerminal(node)` merged since PR #1235

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/NodeConversion.java
MioRtia: It is not enough because a node is still valid if an internal connection or a switch is attached to it (with no terminal)... I added a convenient method to check if there is any equipment attached to the node at NodeBreakerView level to delete this method.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/NodeConversion.java
MioRtia: What error message do you suggest?

MioRtia:
@zamarrenolm I change your PR following fixes in VoltageLevel API, do you agree with this correction?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/NodeConversion.java
mathbagu: Somethink like that:
> Can't find a bus from the Bus View to set Voltage and Angle, we use the bus {} from the Bus/Breaker view. Connectivity node {}

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/NodeConversion.java
```diff
@@ -68,34 +68,30 @@ private VoltageLevel newBoundarySubstationVoltageLevel() {
...
-                // This is an arbitrary country code, Bangladesh code BD also matches with
-                // BounDary
-                .setCountry(boundaryCountryCode())
-                .add()
-                .newVoltageLevel()
-                .setId(context.namingStrategy().getId("VoltageLevel", vlId))
-                .setName(vlName)
-                .setNominalV(nominalVoltage)
-                .setTopologyKind(context.nodeBreaker() ? TopologyKind.NODE_BREAKER : TopologyKind.BUS_BREAKER)
-                .add();
```
mathbagu: @zamarrenolm you should configure your IDE to set to 8 spaces the continuous indentation.

zamarrenolm:
> @zamarrenolm I changed your PR following fixes in VoltageLevel API, do you agree with this correction?

Totally agree. I have checked it with the test case that was giving me problems.

PR #1235, opened 2020-03-19
Add UndirectedGraph.vertexExists(node) method + NodeBreakerView.getOptionalTerminal(node)

MioRtia:

No

Feature

If a node in `NodeBreakerView` does not have an attached network component, the methode `getTerminal(node)` throws an exception. 

The method `getOptionalTerminal` now exists and can return an empty Optional if there is no attached network component to a valid node.

No

**Other information**:

Linked to #1233

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevel.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevel.java
@@ -594,6 +594,14 @@ class NodeBreakerVoltageLevel extends AbstractVoltageLevel {
         }
 
         @Override
+        public Optional<Terminal> getOptionalTerminal(int node) {
+            if (graph.vertexExists(node)) {
+                return Optional.ofNullable(graph.getVertexObject(node));
+            }
+            return Optional.empty();
+        }
+
+        @Override
         public Terminal getTerminal1(String switchId) {
             return getTerminal(getNode1(switchId));
         }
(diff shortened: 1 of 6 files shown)


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevel.java
```diff
@@ -590,7 +590,10 @@ public int getNode2(String switchId) {
 
         @Override
         public Terminal getTerminal(int node) {
-            return graph.getVertexObject(node);
+            if (graph.vertexExists(node)) {
+                return graph.getVertexObject(node);
+            }
+            return null;
```
mathbagu: Is it OK to add a convenient `public Optional<Terminal> getOptionalTerminal(int node)`. Returning null could introduce a lot of breaking changes in the client code. I think the optional version is safer.

math/src/main/java/com/powsybl/math/graph/UndirectedGraphImpl.java
```diff
@@ -149,6 +159,12 @@ public void addVertexIfNotPresent(int v) {
         notifyListener();
     }
 
+    @Override
+    public boolean vertexExists(int v) {
+        checkInvalidVertex(v);
+        return v < vertices.size() && vertices.get(v) != null;
+    }
```
mathbagu: I think you should not change the existing code: it's not necessary to extract the `checkInvalidVertex` in a method, consider duplicate the test here

PR #1238, opened 2020-03-20
Change to CGMES import of SVC regulation

MioRtia:

Bug fix?

When the SVC's regulating control is disabled, the SVC does not regulate even if its `controlEnabled` property is set to true.

When the SVC's regulating control is disabled and its `controlEnabled` property is set to true, the SVC regulates with local default regulation.

No


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForStaticVarCompensators.java
```diff
@@ -103,6 +97,12 @@ private boolean setRegulatingControl(RegulatingControl control, StaticVarCompens
         StaticVarCompensator.RegulationMode regulationMode;
 
         boolean okSet = false;
+        if (!control.enabled && rc.controlEnabledProperty) {
+            context.fixed("SVCControlEnabledStatus", String.format("Regulating control of %s is disabled but controlEnabled property is set to true." +
```
annetill: Maybe you should add in the comment that it is the property of the equipment.

annetill:
Thanks for your detailed answer. I understand that this PR only deals with SVC and nothing has to be changed in the CGMES conversion for generators and tap changers in case of having ````controlEnabled```` for the equipment  ````regulatingControl```` disabled.

For SVC, I understand that, in case of a disabled/wrong regulating control:
* The IIDM regulation status is given by the SSH profile ````cim:RegulatingCondEq.controlEnabled```` ;
** In case of a ````true```` status, we take the mode from the ````EQ```` profile using field ````StaticVarCompensator.sVCControlMode```` ;
*** In case of `````voltage````, we take the set point from the ````EQ```` profile field ````voltageSetPoint`````;
*** In case of `````reactivePower````, we take the set point from the ````SSH```` profile field ````StaticVarCompensator.q`````;

Is it what you mean ? If yes, we will modify this PR in consequence.

Thanks for your review !

MioRtia:
Thank you both for your inputs @annetill @zamarrenolm! 

> CGMES documentation is not clear about the priority to use regulation control or the attributes defined at equipment level in Static Var Compensators. My suggestion would be regulation control should be considered first. If disabled or no info available, then we should rely on sVCControlMode and voltageSetPoint or q attributes, depending if control mode is voltage or reactivePower

Isn't it what is done in this PR? The regulation from EQ is only activated when the regulating control is missing or disabled **and** controlEnabled is set to true. Or is there something I am missing?

annetill:
Do you rely on field ````SVCControlMode```` from EQ profile ? If yes, just wait for Luma's answer.

MioRtia:
> Do you rely on field `SVCControlMode` from EQ profile ? If yes, just wait for Luma's answer.

Yes, I do, this is checked in the `setDefaultRegulatingControl` method.

zamarrenolm:
> > Do you rely on field `SVCControlMode` from EQ profile ? If yes, just wait for Luma's answer.
> 
> Yes, I do, this is checked in the `setDefaultRegulatingControl` method.

Yes you are right, I've reviewed the code and follows my recommendation. No additional changes are required.

Only additional thing is that I am not sure that `controlEnabled` property should be checked for `true` when we are setting the default control data, as it is an attribute directly related to Regulating Controls. Even if `controlEnabled` is `false` (it refers to its regulation control) we should use and apply info from equipment definition.

What do you think?

PR #1243, opened 2020-03-24
Identifiable.name is optional + introduce getOptionalName(), getNameOrId() + deprecate getName()

### LINKED ISSUE #178: Identifiable getName
The AbstractIdentifiable::getName implementation does not fit with the description of the API:
```
Get an the (optional) name  of the object.
```
The implementation returns the ID if the name is null:
```
return name != null ? name : id;
```

Is it really needed?

MioRtia:

fix #178 

**Other information**:
The lack of coverage is not really linked to this PR, but to the code corrected by this PR... Should the tests be added here (bearing in mind, this is not easily tested)? @mathbagu


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiable.java
```diff
@@ -42,7 +42,12 @@ public String getId() {
     }
 
     @Override
-    public String getName() {
```
agnesLeroy: I think we should make getName deprecated to avoid a breaking change.
Then in a later release we could remove this getName, make getOptionalName deprecated and recreate a getName that behaves like getOptionalName. What do you think?

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/VariantManagerImplTest.java
```diff
@@ -48,7 +49,12 @@ public String getId() {
...
         @Override
-        public String getName() {
+        public Optional<String> getOptionalName() {
+            return Optional.empty();
+        }
+
+        @Override
+        public String getNameOrId() {
             return id;
         }
```
agnesLeroy: same remark here

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiable.java
MioRtia: Agree with you. I deprecated it and made it call `getNameOrId()` as default so we don't need to implement it in implementation classes.

PR #1244, opened 2020-03-25
Support XIIDM serialization/deserialization with new shunt API

MioRtia:

No

Support new shunt API IIDM-XML serialization

Only linear shunts IIDM-XML serialization is supported

Linear **and** non linear shunts IIDM-XML serialization should be supported

**Other information**:
I am not sure this is how we want to do things...


iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
```diff
@@ -34,13 +40,30 @@ protected boolean hasSubElements(ShuntCompensator sc) {
...
+            try {
+                XmlUtil.writeDouble(B_PER_SECTION, sc.getModel(ShuntCompensatorLinearModel.class).getbPerSection(), context.getWriter());
+                context.getWriter().writeAttribute(MAXIMUM_SECTION_COUNT, Integer.toString(sc.getMaximumSectionCount()));
+            } catch (XMLStreamException e) {
+                throw new UncheckedXmlStreamException(e);
+            }
+        });
+        IidmXmlUtil.runFromMinimumVersion(IidmXmlVersion.V_1_2, context, () -> {
+            try {
+                context.getWriter().writeAttribute(MODEL, sc.getModelType().name());
```
mathbagu: I think I wouldn't serialize the type: it's a redundant information with the model you will have to check. As you do not use the same XML element name, it's not necessary.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
```diff
@@ -54,35 +77,84 @@ protected void writeRootElementAttributes(ShuntCompensator sc, VoltageLevel vl,
 
     @Override
     protected void writeSubElements(ShuntCompensator sc, VoltageLevel vl, NetworkXmlWriterContext context) throws XMLStreamException {
+        IidmXmlUtil.runFromMinimumVersion(IidmXmlVersion.V_1_2, context, () -> {
+            try {
+                writeModel(sc, context);
+            } catch (XMLStreamException e) {
+                throw new UncheckedXmlStreamException(e);
+            }
```
mathbagu: To make the code simpler, maybe you should move the try catch in the writeModel method.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
```diff
@@ -54,35 +77,84 @@ protected void writeRootElementAttributes(ShuntCompensator sc, VoltageLevel vl,
...
+                adder.newLinearModel()
+                        .setMaximumSectionCount(currentSectionCount + 1) // maximumSectionCount must be > 0
+                        .setbPerSection(Double.MIN_VALUE)
+                        .add();
+            } else if (ShuntCompensatorModelType.NON_LINEAR.equals(modelType)) { // default value for non linear shunt (always overwritten)
+                adder.newNonLinearModel()
+                        .beginSection()
+                            .setSectionNum(currentSectionCount)
+                            .setB(Double.MIN_VALUE)
+                        .endSection()
```
mathbagu: Is it a bit strange? Maybe we should create a default Linear Model, like we did for ReactiveLimits? And allow to change the model of a shunt after its creation.

MioRtia: It may be a good solution but there are two questions:
 1. What value for `bPerSection` do we want?
 2. Does that mean that we are allowing to modify the shunt model of a built shunt? (we wanted to prevent that at first, given that a modification of shunt model may not really make sense...)

MioRtia: Ok, I put another solution that only requires default values for non linear shunt.... what do you think?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
MioRtia: I changed `linearModel` and `nonLinearModel` rather than this attribute (it simplifies the file)

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
```diff
@@ -22,67 +22,53 @@
...
-
     @Override
     protected String getRootElementName() {
         return ROOT_ELEMENT_NAME;
     }
 
     @Override
     protected boolean hasSubElements(ShuntCompensator sc) {
-        return sc != sc.getRegulatingTerminal().getConnectable();
+        return false;
```
mathbagu, line 32: Is there a bug there (and elsewhere)? If an identifiable has properties, then he has sub elements.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
```diff
@@ -22,67 +22,53 @@
...
-        IidmXmlUtil.writeDoubleAttributeFromMinimumVersion(ROOT_ELEMENT_NAME, "targetV", sc.getTargetV(),
-                IidmXmlUtil.ErrorMessage.NOT_DEFAULT_NOT_SUPPORTED, IidmXmlVersion.V_1_2, context);
-        IidmXmlUtil.writeDoubleAttributeFromMinimumVersion(ROOT_ELEMENT_NAME, "targetDeadband",
-                sc.getTargetDeadband(), IidmXmlUtil.ErrorMessage.NOT_DEFAULT_NOT_SUPPORTED, IidmXmlVersion.V_1_2, context);
         writeNodeOrBus(null, sc.getTerminal(), context);
         writePQ(null, sc.getTerminal(), context.getWriter());
     }
 
-    @Override
-    protected void writeSubElements(ShuntCompensator sc, VoltageLevel vl, NetworkXmlWriterContext context) throws XMLStreamException {
```
mathbagu, line 56: Everytime we writeSubElements, we should write properties. I think we make this bug everywhere?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
```diff
@@ -22,67 +22,53 @@
...
+                    + ". IIDM-XML version should be >= 1.2");
         }
+        IidmXmlUtil.assertMinimumVersionIfNotDefault(sc.isVoltageRegulatorOn(), ROOT_ELEMENT_NAME, "voltageRegulatorOn",
+                IidmXmlUtil.ErrorMessage.NOT_DEFAULT_NOT_SUPPORTED, IidmXmlVersion.V_1_2, context);
+        IidmXmlUtil.assertMinimumVersionIfNotDefault(!Double.isNaN(sc.getTargetV()), ROOT_ELEMENT_NAME, "targetV",
+                IidmXmlUtil.ErrorMessage.NOT_DEFAULT_NOT_SUPPORTED, IidmXmlVersion.V_1_2, context);
+        IidmXmlUtil.assertMinimumVersionIfNotDefault(!Double.isNaN(sc.getTargetDeadband()), ROOT_ELEMENT_NAME, "targetDeadband",
+                IidmXmlUtil.ErrorMessage.NOT_DEFAULT_NOT_SUPPORTED, IidmXmlVersion.V_1_2, context);
+        IidmXmlUtil.assertMinimumVersionIfNotDefault(sc.getRegulatingTerminal().getConnectable() != sc, ROOT_ELEMENT_NAME, "regulatingTerminal",
+                IidmXmlUtil.ErrorMessage.NOT_DEFAULT_NOT_SUPPORTED, IidmXmlVersion.V_1_2, context);
```
mathbagu, line 48: in IIDM v1.2, we should not use this class anymore?

iidm/iidm-xml-converter/src/main/resources/xsd/iidm_V1_2.xsd
```diff
@@ -260,21 +261,43 @@
             </xs:extension>
         </xs:complexContent>
     </xs:complexType>
-    <xs:complexType name="ShuntCompensator">
+    <xs:complexType name="AbstractShuntCompensator">
```
mathbagu, line 264: As this class doesn't exist in Java, we should remove it from the XSD. What do you think?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
MioRtia: No, in `AbstractIdentifiableXml`, we have it checks `if (hasSubElements || identifiable.hasProperty())`, not only what `hasSubElements` returns.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
MioRtia: It is written in `super.readSubElements(...)` from `AbstractIdentifiableXml`

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
MioRtia: These lines do not check the version, they check if the shunt is regulating/has a target value/has a regulating terminal/etc... and if it does, it throws an exception because it is not supported in serialization in version older than v1.2.

iidm/iidm-xml-converter/src/main/resources/xsd/iidm_V1_2.xsd
MioRtia: I can simply call it `ShuntCompensator`? However, the xsd is not equivalent to Java structure (the Java classes `NonLinearShuntCompensator` and `LinearShuntCompensator` do not exist for example).

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
mathbagu: Hum... So you check that the Shunt (in V1.2) is compatible with an old version (the new attributes have the default value)? Do I understand well?

iidm/iidm-xml-converter/src/main/resources/xsd/iidm_V1_2.xsd
mathbagu: ShuntCompensator seems more user friendly: keep in mind that the XSD is for the human.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
MioRtia: Yes!

PR #1247, opened 2020-03-27
Improve security analysis interceptor

yichen88:
Add two more methods in interceptor and two contexts correspondent.

Make interceptors protected

Improve interceptor

**Other information**:


security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResultBuilder.java
```diff
@@ -26,8 +28,8 @@
 public class SecurityAnalysisResultBuilder {
 
     private final LimitViolationFilter filter;
-    private final RunningContext context;
-    private final List<SecurityAnalysisInterceptor> interceptors;
+    protected final RunningContext context;
```
sylvlecl: Since we should not anymore subclass the builder, is it necessary / useful to make it protected?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResultBuilder.java
```diff
@@ -26,8 +28,8 @@
...
+    protected final List<SecurityAnalysisInterceptor> interceptors;
```
sylvlecl: Same question.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResultBuilder.java
```diff
@@ -62,13 +64,23 @@ public PreContingencyResultBuilder preContingency() {
...
         return new PostContingencyResultBuilder(contingency);
     }
 
+    /**
+     * Initiates the creation of a post-contingency result
+     * @param contingency the contingency for which a result should be created
+     * @param contingencyContextExtensions extensions for {@link com.powsybl.security.interceptors.ContingencyContext}
+     * @return a {@link PostContingencyResultBuilder} instance
+     */
+    public PostContingencyResultBuilder contingency(Contingency contingency, Collection<Extension<ContingencyContext>> contingencyContextExtensions) {
```
sylvlecl: Maybe it would be better to add the extensions throught the context ?
Like
```
builder.addContextExtension(class, ext)
```

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/interceptors/SecurityAnalysisInterceptor.java
```diff
@@ -15,9 +16,44 @@
...
+     */
     void onPreContingencyResult(RunningContext context, LimitViolationsResult preContingencyResult);
 
+    /**
+     * Deprecated. Use {@link #onPostContingencyResult(ContingencyContext context, PostContingencyResult postContingencyResult)}
+     * @param context
+     * @param postContingencyResult
+     */
+    @Deprecated
     void onPostContingencyResult(RunningContext context, PostContingencyResult postContingencyResult);
```
sylvlecl: Should we provide a default implementation so that new implementations do not have to implement this deprecated method ?

yichen88: The new InterceptorImpl should extends `DefaultSecurityAnalysisInterceptor`

security-analysis/security-analysis-api/src/test/java/com/powsybl/security/SecurityAnalysisResultBuilderTest.java
```diff
@@ -33,29 +37,46 @@ public void failedResult() {
...
         SecurityAnalysisResultBuilder builder = new SecurityAnalysisResultBuilder(new LimitViolationFilter(),
-                new RunningContext(network, network.getVariantManager().getWorkingVariantId()));
+                new RunningContext(network, network.getVariantManager().getWorkingVariantId()), Collections.singleton(securityAnalysisInterceptorMock));
 
         VoltageLevel vl = network.getVoltageLevel("VLHV1");
         vl.getBusView().getBusStream().forEach(b -> b.setV(410));
 
+        SecurityAnalysisResultBuilder.PreContingencyResultBuilder preContingencyResultBuilder = builder.preContingency();
+        SecurityAnalysisResultBuilder securityAnalysisResultBuilder = preContingencyResultBuilder.setComputationOk(true)
+                .addViolationContextExtension(Security.checkLimits(network).get(0), preVioContextExt)
```
mathbagu, line 51: From my point of view, this is not a good design. With this one, we fill the SecurityAnalysisResultBuilder with a lot of information that are used only once.

The use case is:
- the SecurityAnalysisResultBuilder is responsible on creating the contingencies and the violations.
- the SecurityAnalysisInterceptors are responsible on adding the additional data to the contingencies and the violations

The SecurityAnalysisInterceptors need a way to access to the network or to another data source (the result of the simulation). I think we should keep only the RunningContext with extensions. However, the Interceptors should have a new notification method to distinguish the violation creation in pre and post contingency state:
```
public void onLimitViolation(RunningContext, Contingency contingency, LimitViolation violation);
```

With the contingency name, and the violation, the extension in the running context should be able to retrieve the relevant information from an additional datasource. In that case, I'm not convinced we should keep ViolationContext and ContingencyContext, but only RunningContext, that is global to the simulation. Separate everything in really small classes make the API hard to understand.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/interceptors/SecurityAnalysisInterceptor.java
mathbagu: You cannot assume that: a developper could inherit from the interface directly. If you deprecate a method, you have to provide a default implementation to make it possible to remove it without breaking things.
Without a default implementation, I need to implement it in my subclass so removing this method from the API will break my code.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResultBuilder.java
```diff
@@ -94,20 +97,58 @@ public SecurityAnalysisResult build() {
...
+            if (!violations.contains(limitViolation)) {
+                this.addViolation(limitViolation);
+            }
+            extensionsByViolation.computeIfAbsent(limitViolation, lv -> new ArrayList<>());
+            extensionsByViolation.computeIfPresent(limitViolation, (lv, list) -> {
+                list.add(extension);
+                return list;
+            });
+            return (B) this;
+        }
```
mathbagu, line 135: We should not create a ViolationContext each time a violation is created. I think this is the responisibility of the extension to offer a way to retrieve relevant information from the violation (and the current contingency). 

See below for more details.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResultBuilder.java
```diff
@@ -132,19 +173,33 @@ public SecurityAnalysisResultBuilder endPreContingency() {
...
+
+        public PostContingencyResultBuilder addContingencyContextExtension(Extension<ContingencyContext> extension) {
+            contingencyContextExtensions.add(Objects.requireNonNull(extension));
+            return this;
+        }
+
+        public PostContingencyResultBuilder addContingencyContextExtensions(List<Extension<ContingencyContext>> extensions) {
+            Objects.requireNonNull(extensions).forEach(this::addContingencyContextExtension);
+            return this;
```
mathbagu, line 190: Same remark here

security-analysis/security-analysis-api/src/test/java/com/powsybl/security/SecurityAnalysisResultBuilderTest.java
sylvlecl: The design with more specific `ViolationContext` and `ContingencyContext` comes from a comment from @jonenst about not changing the currently immutable class. If we keep only one `RunningContext` but make it very "mutable" by adding and removing extensions, it will not be possible to ensure its consistence, in particular in case of concurrent use.

I think we could keep only the `RunningContext` class as you propose, but in that case it's important to create a new one for each notification of the interceptor, so that it stays thread safe.

mathbagu: Then, don't make the RunningContext mutable using the extension mechanism and create subclasses? I think creating a lot of small objects with very short life cycle is not a good idea. You can also create a decorator to hide methods you don't want someone use (probably removing an extension). There is a lot of way to solve this, isn't it?

sylvlecl: But then again you leave the implentation deal with a big unique stateful object, which is not convenient for thread safety, and may prevent to use the result builder "on the go": you will need to read all the result in order to be able to create your context.

It will not either be more performant I think, because you will probably need to create maps and search in them frequently, whereas small objects carry just the information you need.

mathbagu:
@yichen88 Here the final specification to solve this issue.

First of all, we decide to revert the PR #1207 (see. #1257). Once it will be merged, I will patch your branch to reintroduce the modifications I made in #1207 as a starting point.

We will introduce a single interface for the context:
```
public interface SecurityAnalysisResultContext {
    Network getNetwork();
} 
```

One instance of the SecurityAnalysisResultContext would be created for each result (or violation) that would be created. The `SecurityAnalysisResultContext` instances will be passed to the different builders and then used by the interceptors. So you have to change a little bit the design of both `SecurityAnalysisResultBuilder` and `SecurityAnalysisInterceptor` class

# SecurityAnalysisResultBuilder

**- Constructor of SecurityAnalysisResultBuilder**
A `SecurityAnalysisResultContext` will be passed as parameter to the constructor of the builder, in replacement of the current RunningContext. This context will be passed to the `SecurityAnalysisInterceptor::onSecurityAnalysisResult` method.

**- SecurityAnalysisResultBuilder::preContingency()**
A `SecurityAnalysisResultContext` will be passed as parameter to the `preContingency()` method, and forward to the interceptors when the `PreContingencyResult` is build.

**- SecurityAnalysisResultBuilder::contingency()**
A `SecurityAnalysisResultContext` will be passed as parameter to the `contingency()` method, and forward to the interceptors when the `PostContingencyResult` is build.

**- PreContingencyResultBuilder::addViolation()**
A `SecurityAnalysisResultContext` will be passed as parameter to the `addViolation()` and `addViolations()` methods and forward to the interceptors when the `onLimitViolation` is called. Note that there will be two different `onLimitViolation` method in the interceptor (see below).

**Example of usage:**
```
SecurityAnalysisResultContext context = new SecurityAnalysisResultContext(network);
SecurityAnalysisResultBuilder builder = createBuilder(context);
builder.preContingency(context)
    .setComputationOk(true)
    .addViolations(violations, context)
    .build();
for (Contingency contingency : contingencies) {
    builder.contingency(contingency, context)
        .setComputationOk(true)
        .addViolations(getViolations(contingency), context)
        .build();
}
SecurityAnalysisResult result = builder.build();
```

# SecurityAnalysisInterceptor
The new SecurityAnalysisInterceptor interface looks like this:
```
public interface SecurityAnalysisInterceptor {

    void onPreContingencyResult(LimitViolationsResult preContingencyResult, SecurityAnalysisResultContext context);

    void onPostContingencyResult(PostContingencyResult postContingencyResult, SecurityAnalysisResultContext context);

    void onSecurityAnalysisResult(SecurityAnalysisResult result, SecurityAnalysisResultContext context);

    void onLimitViolation(LimitViolation limitViolation, SecurityAnalysisResultContext context);

    void onLimitViolation(Contingency contingency, LimitViolation limitViolation, SecurityAnalysisResultContext context);
}
```

You probably notice that the parameters' order changed. To make the API backward compatible, I propose to
- deprecated the old method, and give them a default behavior (i.e. call the new method)
- The RunningContext should be a subclass of the new SecurityAnalysisResultContext
These two changes should be sufficient, hopefully.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResultBuilder.java
```diff
@@ -26,179 +26,177 @@
...
+        /**
+         * Finalize the creation of the PreContingencyResult instance
+         * @return the parent {@link SecurityAnalysisResultBuilder} instance.
+         */
+        public SecurityAnalysisResultBuilder endPreContingency() {
+            List<LimitViolation> filteredViolations = filterViolations(resultContext, filter, interceptors);
             LimitViolationsResult res = new LimitViolationsResult(computationOk, filteredViolations);
-            interceptors.forEach(i -> i.onPreContingencyResult(context, res));
-            preContingencyResult = res;
+//            interceptors.forEach(i -> i.onPreContingencyResult(context, res));
```
yichen88, line 159: !!

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/interceptors/SecurityAnalysisResultContextImpl.java
```diff
@@ -0,0 +1,28 @@
...
+package com.powsybl.security.interceptors;
+
+import com.powsybl.iidm.network.Network;
+
+import java.util.Objects;
+
+/**
+ * @author Yichen TANG <yichen.tang at rte-france.com>
+ */
+public class SecurityAnalysisResultContextImpl implements SecurityAnalysisResultContext {
```
sylvlecl, line 16: Would slightly prefer to name it `DefaultSecurityAnalysisResultContext` because the "Impl" suffix sounds as if it's the only implementation of the context.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/interceptors/SecurityAnalysisInterceptor.java
```diff
@@ -15,9 +18,80 @@
...
+        onSecurityAnalysisResult(result, context);
+    }
+
+    /**
+     * Callback after the pre-contingency result is built.
+     * @param preContingencyResult
+     * @param context
+     */
+    default void onPreContingencyResult(LimitViolationsResult preContingencyResult, SecurityAnalysisResultContext context) {
+        onPreContingencyResult(new RunningContext(context.getNetwork(), VariantManagerConstants.INITIAL_VARIANT_ID), preContingencyResult);
```
sylvlecl, line 57: I finally don't think it's a good idea to add those default implementations, because they will not have the same behaviour as the original methods (because passed variant ID is not the same).
I think it's less error-prone to force the implementations to adapt on migration.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResultBuilder.java
```diff
@@ -57,18 +57,26 @@ private void addPostContingencyResult(PostContingencyResult result) {
      * @return a {@link PreContingencyResultBuilder} instance.
      */
     public PreContingencyResultBuilder preContingency() {
-        return new PreContingencyResultBuilder();
+        return new PreContingencyResultBuilder(context);
+    }
+
+    public PreContingencyResultBuilder preContingency(SecurityAnalysisResultContext context) {
```
sylvlecl, line 63: Maybe we should rename the argument to avoid confusion with the class field.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResultBuilder.java
```diff
@@ -80,49 +88,71 @@ public SecurityAnalysisResult build() {
...
+
         public B addViolation(LimitViolation violation) {
+            addViolation(violation, resultContext);
+            return (B) this;
+        }
+
+        public B addViolation(LimitViolation violation, SecurityAnalysisResultContext context) {
+            Objects.requireNonNull(context);
+            interceptors.forEach(i -> i.onLimitViolation(violation, context));
             violations.add(Objects.requireNonNull(violation));
```
sylvlecl, line 124: Better to move the non null check at the method start now.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/interceptors/DefaultSecurityAnalysisInterceptor.java
```diff
@@ -23,12 +25,33 @@ public void onPreContingencyResult(RunningContext context, LimitViolationsResult
     }
 
     @Override
-    public void onPostContingencyResult(RunningContext context, PostContingencyResult postContingencyResult) {
+    public void onSecurityAnalysisResult(RunningContext context, SecurityAnalysisResult result) {
```
sylvlecl, line 28: It should be removed.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResultBuilder.java
yichen88:
```java
        /**
         * Adds a {@link LimitViolation} to the builder.
         * The default result context would be supplied to interceptors.
         * @param violation
         * @return
         */
        public B addViolation(LimitViolation violation) {
            addViolation(violation, resultContext);
            return (B) this;
        }
```
I am not sure for this part.

sylvlecl: It looks good to me :) As a convenience we allow the implementation to provide only a context at the beginning, and we pass it along to all calls.
It ensures that a context is always given to the interceptors.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResultBuilder.java
```diff
@@ -26,179 +26,198 @@
...
 
-    private class PostContingencyResultBuilder extends LimitViolationsResultBuilder {
+    public class PostContingencyResultBuilder extends AbstractLimitViolationsResultBuilder<PostContingencyResultBuilder> {
 
         private final Contingency contingency;
 
         PostContingencyResultBuilder(Contingency contingency) {
+            super(context);
+            this.contingency = Objects.requireNonNull(contingency);
+        }
```
mathbagu, line 195: We should remove this method, and pass the context in the postContingency() method, to have a symmetrical design

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResultBuilder.java
```diff
@@ -26,179 +26,198 @@
...
+            return (B) this;
+        }
 
-    private class LimitViolationsResultBuilder implements ResultBuilder {
+        /**
+         * Initiates a result builder with a {@link SecurityAnalysisResultContext}.
+         * @param resultContext The context would be used when creation result or as default context when a limit violation added.
+         */
+        public AbstractLimitViolationsResultBuilder(SecurityAnalysisResultContext resultContext) {
+            this.resultContext = resultContext;
```
mathbagu, line 128: It seems that the context is never null. Is it a coincidence or a choice. I think we should assert that it's not null there and in the SecurityAnalysisResultBuilder constructor.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResultBuilder.java
```diff
@@ -26,179 +26,198 @@
...
+
+        PostContingencyResultBuilder(Contingency contingency, SecurityAnalysisResultContext resultContext) {
+            super(resultContext);
             this.contingency = Objects.requireNonNull(contingency);
         }
 
-        void endContingency() {
+        @Override
+        public PostContingencyResultBuilder addViolation(LimitViolation violation, SecurityAnalysisResultContext limitViolationContext) {
+            Objects.requireNonNull(limitViolationContext);
```
mathbagu, line 204: Look, we are not consistent

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/interceptors/DefaultSecurityAnalysisInterceptor.java
```diff
@@ -18,17 +20,28 @@
 public class DefaultSecurityAnalysisInterceptor implements SecurityAnalysisInterceptor {
```
mathbagu, line 20: Sonar require an explanation when a method is let empty. Please every method in this class

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/interceptors/SecurityAnalysisInterceptor.java
```diff
@@ -6,18 +6,114 @@
...
+     * @param contingency
+     * @param limitViolation
+     * @param context
+     */
+    default void onLimitViolation(Contingency contingency, LimitViolation limitViolation, SecurityAnalysisResultContext context) {
+
+    }
 
-    void onSecurityAnalysisResult(RunningContext context, SecurityAnalysisResult result);
 }
```
mathbagu, line 119: Fix javadoc and explain why method are let empty with a comment

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResultBuilder.java
yichen88: Made it private and check non null on two usages

PR #1248, opened 2020-03-30
Enable the configuration YML to be updated by a JSON file

### LINKED ISSUE #743: LF and SA parameter extensions updates from JSON files

Bug.

The loadflow and security-analysis commands offer the possibility to update default parameters from a JSON file (option --parameters-file). The file correctly updates properties values for "top level" properties (the ones in LoadFlowParameters for example), but on the other side it completely resets the content of specific parameters defined in extensions.

That is because the JSON parser overwrites the whole extension object, instead of only updating its fields.

For example with hades2 load flow:

1. define in your configuration some default values for its parameters:

```yaml
hades2-default-parameters:
        balanceType: PROPORTIONAL_TO_GENERATION_P_MAX
        svcRegulationOn: true
        countriesToBalance: FR
        dcMode: false

```

2. override one of those values in the JSON file:

```json
  "extensions" : {
    "adn-loadflow-parameters" : {
      "dcMode": false
    }
  }
```

3. run the computation : other parameters will be reset (countriesToBalance etc.)

The behaviour should be homogeneous : ideally extensiosn should only be updated, not completely overwritten.

Being able, for one commande line execution, to set some security analysis parameters specific to an implementation without resetting the default values defined in configuration.

PowSyBl Version: 2.4.1

Etienn3L:

fix #743 

Bug fix.

For loadflow and security analysis computations :
The configuration provided by the configuration file (yml) is completely overwritten when using a JSON file and the missing values in the json file are set to default.

Now the json file just updates the fields and keep the configuration provided by the yml file.

**Other information**:


contingency/contingency-api/src/test/java/com/powsybl/contingency/json/ContingencyJsonTest.java
```diff
@@ -123,6 +123,11 @@ public DummyExtension deserialize(JsonParser jsonParser, DeserializationContext
             return new DummyExtension();
         }
 
+        @Override
+        public DummyExtension deserializeAndUpdate(JsonParser jsonParser, DeserializationContext deserializationContext, DummyExtension parameters) throws IOException {
```
sylvlecl, line 127: I think it's better to leave the default implementation here.

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/json/LoadFlowParametersDeserializer.java
```diff
@@ -71,16 +69,18 @@ public LoadFlowParameters deserialize(JsonParser parser, DeserializationContext
 
                 case "extensions":
                     parser.nextToken();
-                    extensions = JsonUtil.readExtensions(parser, deserializationContext, JsonLoadFlowParameters.getExtensionSerializers());
+                    if (parameters.getExtensions().size() > 0) {
```
sylvlecl, line 72: As we are in an update operation here (since we have an initial value of parameters, I think it's simpler to call `updateExtensions` in any case. If there is no extension, it will have the same behaviour anyway.

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/tools/RunLoadFlowTool.java
```diff
@@ -160,6 +161,7 @@ public void run(CommandLine line, ToolRunningContext context) throws Exception {
         }
 
         LoadFlowParameters params = LoadFlowParameters.load();
+        // System.out.print("parametre avant fusion : ");
```
sylvlecl, line 164: Don't forget to delete it ;)

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/SecurityAnalysisParametersDeserializer.java
```diff
@@ -51,16 +49,17 @@ public SecurityAnalysisParameters deserialize(JsonParser parser, Deserialization
 
                 case "extensions":
                     parser.nextToken();
-                    extensions = JsonUtil.readExtensions(parser, deserializationContext, JsonSecurityAnalysisParameters.getExtensionSerializers());
+                    if (parameters.getExtensions().size() > 0) {
```
sylvlecl, line 52: Same as for load flow parameters : I think it's simpler to call the "update" event when there is no extension.

security-analysis/security-analysis-api/src/test/resources/ActivePowerExtension.json
```diff
@@ -0,0 +1,11 @@
+{
+  "version" : "1.0",
```
sylvlecl, line 2: (I think you probably committed this by mistake)

sensitivity-api/src/test/java/com/powsybl/sensitivity/json/JsonSensitivityComputationParametersTest.java
```diff
@@ -89,6 +89,11 @@ public DummyExtension deserialize(JsonParser jsonParser, DeserializationContext
             return new DummyExtension();
         }
 
+        @Override
+        public DummyExtension deserializeAndUpdate(JsonParser jsonParser, DeserializationContext deserializationContext, DummyExtension parameters) throws IOException {
```
sylvlecl, line 93: Probably better to leave the default implementation instead.

dynamic-simulation/dynamic-simulation-api/src/test/java/com/powsybl/dynamicsimulation/json/JsonDynamicSimulationParametersTest.java
```diff
@@ -89,6 +89,11 @@ public DummyExtension deserialize(JsonParser jsonParser, DeserializationContext
             return new DummyExtension();
         }
 
+        @Override
+        public DummyExtension deserializeAndUpdate(JsonParser jsonParser, DeserializationContext deserializationContext, DummyExtension parameters) throws IOException {
```
sylvlecl, line 93: Same : I think it's better to leave the default implementation here.

security-analysis/security-analysis-api/src/test/java/com/powsybl/security/json/JsonSecurityAnalysisParametersTest.java
```diff
@@ -61,12 +63,61 @@ public void readError() throws IOException {
...
+        SecurityAnalysisParameters parameters = new SecurityAnalysisParameters();
+        DummyExtension extension = new DummyExtension();
+        extension.setParameterBoolean(false);
+        extension.setParameterString("Bonjour");
+        extension.setParameterDouble(2.8);
+        DummyExtension oldExtension = new DummyExtension(extension);
+        parameters.addExtension(DummyExtension.class, extension);
+        JsonSecurityAnalysisParameters.update(parameters, getClass().getResourceAsStream("/SecurityAnalysisParametersExtensionUpdate.json"));
+        DummyExtension updatedExtension = parameters.getExtension(DummyExtension.class);
+        assertEquals(updatedExtension.isParameterBoolean(), oldExtension.isParameterBoolean());
```
sylvlecl, line 77: Detail for the assertions:
the first parameter should be the "expected" value, so here it would be more logical to invert the 2 parameters (same for other assertions).

MioRtia:
Hello, thank you for your contribution! Could you please take a more specific name for this PR? "**Json parser**" is not really sufficient to explain what is implemented in this PR. It will be the commit name when we merge it, that's why it is really important to have a comprehensive name.

commons/src/main/java/com/powsybl/commons/json/JsonUtil.java
```diff
@@ -217,6 +217,52 @@ public static void writeOptionalIntegerField(JsonGenerator jsonGenerator, String
...
+            }
+        } else {
+            if (extensionsNotFound != null) {
+                extensionsNotFound.add(extensionName);
+            }
+            skip(parser);
+            return null;
+        }
+    }
+
```
MioRtia, line 265: I think it can be refactored with `readExtensions` and `readExtension`, it looks similar.

commons/src/test/java/com/powsybl/commons/extensions/ExtensionTest.java
```diff
@@ -109,4 +108,27 @@ public void testWriteJson() throws IOException {
             assertTrue(notFound.contains("BarExt"));
         }
     }
+
+    @Test
+    public void testUpdateAndDeserialize() throws IOException {
+        Foo foo = new Foo();
+        FooExt fooExt = new FooExt(false, "Bonjour");
```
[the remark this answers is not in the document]
Etienn3L: Ok I'll change that

commons/src/main/java/com/powsybl/commons/json/JsonUtil.java
```diff
@@ -217,6 +217,48 @@ public static void writeOptionalIntegerField(JsonGenerator jsonGenerator, String
...
+        while (parser.nextToken() != JsonToken.END_OBJECT) {
+            Extension<T> extension = updateExtension(parser, context, supplier, extensionsNotFound, toUpdatedParameters);
+            if (extension != null) {
+                extensions.add(extension);
+            }
+        }
+        return extensions;
+    }
+
+    public static <T extends Extendable, E extends Extension<T>> E updateExtension(JsonParser parser, DeserializationContext context,
```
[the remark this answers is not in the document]
Etienn3L: the only method calling this one have already the requirements (non null) and there is no require non null on the method "readExtension" which is similar, maybe requirements can be add to both methods ?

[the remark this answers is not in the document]
Etienn3L: Or both methods can be private ?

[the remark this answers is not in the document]
MioRtia: Yes, if this method is only called in `updateExtensions`, it might be better to keep it private. I think `readExtension` is public because it is called elsewhere (not in this repository but in another project using powsybl-core) but if `Objects.requireNonNull` is not called then, it is by mistake I think.

commons/src/main/java/com/powsybl/commons/extensions/ExtensionJsonSerializer.java
```diff
@@ -23,4 +23,8 @@
     void serialize(E extension, JsonGenerator jsonGenerator, SerializerProvider serializerProvider) throws IOException;
 
     E deserialize(JsonParser jsonParser, DeserializationContext deserializationContext) throws IOException;
+
+    default E deserializeAndUpdate(JsonParser jsonParser, DeserializationContext deserializationContext, E parameters) throws IOException {
+        return deserialize(jsonParser, deserializationContext);
```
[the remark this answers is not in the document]
sylvlecl: You're right that it's error-prone. I think the options are:
* do it this way, and existing extensions will keep behaving as today, BUT new extensions could "forget" to actually implement this
* throw an exception as default implementation, but I think it's not a good option since it will compile but fail at runtime
* do not implement it, and make it a breaking change, but in that case it's gonna be inconvenient for users

Maybe the best option is to keep the current approach but insist in javadoc that the method must be implemented for updates to work correctly ?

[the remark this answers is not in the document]
sylvlecl: I added some javadoc to insist on the need to implement the method for the feature to work properly.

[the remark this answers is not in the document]
MioRtia: I agree it should also be documented in the migration guide for people/projects with custom extensions in my opinion.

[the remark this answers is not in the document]
sylvlecl: I have added a section in the migration guide !

PR #1249, opened 2020-03-31
Issue 1083 ucte internal equivalent nodes

### LINKED ISSUE #1083: UCTE internal equivalent nodes lost during UCTE roundtrip

Bug

Currently, it is hard-coded in PowSyBl that Xnodes are "Equivalent" nodes in UCTE format, and all other nodes (inside actual countries) are "Real" nodes.
However, some files may define nodes inside a country that are tagged as "Equivalent" nodes. It is the case for example for XNodes between german areas.

This information is lost during a Convergence round trip from UCTE back to UCTE.

Following UCTE file illustrates the issue.
Initial file:
```
##C 2007.05.01
##N
##ZFR
FFFFFF11              1 0        3.96708 0.00000 0.00000 0.00000
##L
##T
##R
```
becomes after UCTE import/export:
```
##C 2007.05.01
##N
##ZFR
FFFFFF11              0 0        3.96708 0.00000 0.00000 0.00000
##L
##T
##R
```

Equivalent nodes tag should be kept at UCTE import/export

Current behaviour is not in line with UCTE-DEF specification.

  - PowSyBl Version: 3.1.0-SNAPSHOT
  - OS Version: Linux

jlabous:

fix #1083


ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
```diff
@@ -93,6 +93,7 @@ private static void createBuses(UcteNetworkExt ucteNetwork, UcteVoltageLevel uct
                     .add();
 
             addGeographicalNameProperty(ucteNode, bus);
+            bus.setFictitious(UcteNodeStatus.EQUIVALENT.equals(ucteNode.getStatus()));
```
mathbagu, line 96: Replace the equals by a ==

PR #1253, opened 2020-04-01
Matpower importer

CBiasuzzi:

feature

adds a Network importer for the MatPower case format file (.m, .mat)

**Other information**:


matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerNetworkFactory.java
```diff
@@ -0,0 +1,77 @@
...
+
+import com.powsybl.commons.datasource.ResourceDataSource;
+import com.powsybl.commons.datasource.ResourceSet;
+import com.powsybl.iidm.network.Network;
+import com.powsybl.iidm.network.NetworkFactory;
+
+/**
+ * @author Christian Biasuzzi <christian.biasuzzi@techrain.eu>
+ */
+public final class MatpowerNetworkFactory {
```
mathbagu, line 17: What is the license of the different cases (i.e. compatible with ours)?

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
```diff
@@ -0,0 +1,469 @@
...
+@AutoService(Importer.class)
+public class MatpowerImporter implements Importer {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(MatpowerImporter.class);
+
+    private static final String FORMAT = "MATPOWER";
+
+    private static final String[] EXTENSIONS = {"m", "mat"};
+
+    public static final LocalDate DEFAULTDATE = LocalDate.of(2020, Month.JANUARY, 1);
```
mathbagu, line 47: The default date should be the _now()_

Coding style: a general remark about constant. The order we have is public first, then private constant.

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
```diff
@@ -0,0 +1,469 @@
...
+
+    private static final Parameter IGNORE_BASE_VOLTAGE_PARAMETER = new Parameter("ignore-base-voltage",
+            ParameterType.BOOLEAN,
+            "Ignore base voltage specified in the file",
+            Boolean.TRUE);
```
mathbagu, line 52: Is it really interesting to ignore the base voltage? Do you have use cases where these values are inconsistent?

By convention, you should prefix this parameter name with a prefix to avoid clash. Have a look to XmlImporter, XmlExporter or UcteExporter: I would propose `matpower.import.ignore-base-voltage`.

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
```diff
@@ -0,0 +1,469 @@
...
+        return branch.getRatio() != 0;
+    }
+
+    private static void createSubstationMapping(MatpowerModel model, ContainersMapping containersMapping) {
+        UndirectedGraph<String, Object> sGraph = new Pseudograph<>(Object.class);
+        for (String voltageLevelId : containersMapping.voltageLevelIdToBusNums.keySet()) {
+            sGraph.addVertex(voltageLevelId);
+        }
+        for (MBranch branch : model.getBranches()) {
+            if (isTransformer(branch)) {
```
mathbagu, line 96: Question: is there non impedant branches in MatPower format? In that case, could their ends be in the same substation or not?

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
```diff
@@ -0,0 +1,469 @@
...
+    private static void createVoltageLevelMapping(MatpowerModel model, ContainersMapping containersMapping) {
+        UndirectedGraph<Integer, Object> vlGraph = new Pseudograph<>(Object.class);
+        for (MBus mBus : model.getBuses()) {
+            vlGraph.addVertex(mBus.getNumber());
+        }
+        for (MBranch mBranch : model.getBranches()) {
+            if (mBranch.getR() == 0 && mBranch.getX() == 0) {
+                vlGraph.addEdge(mBranch.getFrom(), mBranch.getTo());
+            }
+        }
```
mathbagu, line 119: To be check, regarding my previous comment

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
```diff
@@ -0,0 +1,469 @@
...
+            String voltageLevelId = "VL" + busNums.iterator().next();
+            containersMapping.voltageLevelIdToBusNums.put(voltageLevelId, busNums);
+            for (int busNum : busNums) {
+                containersMapping.busNumToVoltageLevelId.put(busNum, voltageLevelId);
+            }
+        }
+    }
+
+    private static String getBusId(int busNum) {
+        return "B" + busNum;
```
mathbagu, line 130: Use `BUS` instead?

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
```diff
@@ -0,0 +1,469 @@
...
+            String busId = getBusId(mGen.getNumber());
+            String genId = createGenId(mGen, voltageLevel.getNetwork());
+            Generator generator = voltageLevel.newGenerator()
+                    .setId(genId)
+                    .setConnectableBus(busId)
+                    .setBus(isInService(mGen) ? busId : null)
+                    .setTargetV(mGen.getVoltageMagnitudeSetpoint() * voltageLevel.getNominalV())
+                    .setTargetP(mGen.getRealPowerOutput())
+                    .setTargetQ(mGen.getReactivePowerOutput())
+                    .setVoltageRegulatorOn(true)
```
mathbagu, line 169: All generators are regulating? Could getVoltageMagnitudeSetpoint() be 0, in that case, maybe we should set voltageRegulatorOn to false.

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
```diff
@@ -0,0 +1,469 @@
...
+                    .setMaxP(mGen.getMaximumRealPowerOutput())
+                    .setMinP(mGen.getMinimumRealPowerOutput())
+                    .add();
+
+            generator.newMinMaxReactiveLimits()
+                    .setMinQ(mGen.getMinimumReactivePowerOutput())
+                    .setMaxQ(mGen.getMaximumReactivePowerOutput())
+                    .add();
```
mathbagu, line 177: Should be move in the else block of the next `if` statement

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
```diff
@@ -0,0 +1,469 @@
...
+            }
+        });
+    }
+
+    private static Bus createBus(MBus mBus, VoltageLevel voltageLevel) {
+        String busId = getBusId(mBus.getNumber());
+        LOGGER.debug("Creating bus {}", busId);
+        Bus bus = voltageLevel.getBusBreakerView().newBus()
+                .setId(busId)
+                .setName(busId)
```
mathbagu, line 201: To be consistent with generators (and maybe the rest of this importer), you should set the name for all the identifiables to the ID, or never set it.

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
```diff
@@ -0,0 +1,469 @@
...
+        if (mBus.getRealPowerDemand() != 0 || mBus.getReactivePowerDemand() != 0) {
+            String busId = getBusId(mBus.getNumber());
+            String loadId = busId + "-L";
+            LOGGER.debug("Creating load {}", loadId);
+            voltageLevel.newLoad()
+                .setId(loadId)
+                .setConnectableBus(busId)
+                .setBus(busId)
+                .setP0(mBus.getRealPowerDemand())
+                .setQ0(mBus.getReactivePowerDemand())
```
mathbagu, line 243: Is it OK regarding the sign convention? (same question for all equipments)

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
```diff
@@ -0,0 +1,469 @@
...
+
+    private static String createGenId(MGen mGen, Network network) {
+        String genIdPrefix = "G" + mGen.getNumber() + "-";
+        int uniqueGenSuffix = 0;
+        String genId;
+        do {
+            genId = genIdPrefix + uniqueGenSuffix++;
+        } while  (network.getIdentifiable(genId) != null);
+        return genId;
+    }
```
mathbagu, line 256: Why don't you use the ensureUniqueId feature of `IdentifiableAdder`?

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
```diff
@@ -0,0 +1,469 @@
...
+                .setConnectableBus(busId)
+                .setBus(busId)
+                .setP0(mBus.getRealPowerDemand())
+                .setQ0(mBus.getReactivePowerDemand())
+                .add();
+        }
+    }
+
+    private static String createGenId(MGen mGen, Network network) {
+        String genIdPrefix = "G" + mGen.getNumber() + "-";
```
mathbagu, line 249: The naming convention is different for generator: for generator you use a prefix, but for loads and shunts you use a post-fix

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
```diff
@@ -0,0 +1,469 @@
...
+        } while  (network.getIdentifiable(genId) != null);
+        return genId;
+    }
+
+    private static void createShuntCompensator(MBus mBus, VoltageLevel voltageLevel, PerUnitContext perUnitContext) {
+        if (mBus.getShuntSusceptance() != 0) {
+            String busId = getBusId(mBus.getNumber());
+            String shuntId = busId + "-SH";
+            LOGGER.debug("Creating shunt {}", shuntId);
+            double zb = Math.pow(voltageLevel.getNominalV(), 2) / perUnitContext.getBaseMva();
```
mathbagu, line 263: I'm not sure Math.pow is really efficient to compute a square... use `voltageLevel.getNominalV() * voltageLevel.getNominalV() / perUnitContext.getBaseMva())`

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
```diff
@@ -0,0 +1,469 @@
...
+    }
+
+    private static String getBranchId(char type, int from, int to, Network network) {
+        String id;
+        int uniqueCircuit = 1;
+        do {
+            id = "" + type + from + "-" + to + "-" + uniqueCircuit++;
+        } while (network.getIdentifiable(id) != null);
+        return id;
+    }
```
mathbagu, line 284: Same question about ensureUniqueId

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
```diff
@@ -0,0 +1,469 @@
...
+    private static boolean isInService(MBranch branch) {
+        return Math.abs(branch.getStatus()) > 0;
+    }
+
+    private static boolean isInService(MGen generator) {
+        return generator.getStatus() > 0;
+    }
+
+    private static void createLine(MBranch branch, ContainersMapping containerMapping, Network network, PerUnitContext perUnitContext) {
+        String lineId = getBranchId('L', branch.getFrom(), branch.getTo(), network);
```
mathbagu, line 295: For branches (lines and transformers), you use a prefix also. See above, try to be consistent

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
```diff
@@ -0,0 +1,469 @@
...
+        LOGGER.debug("Creating line {}", lineId);
+        String bus1Id = getBusId(branch.getFrom());
+        String bus2Id = getBusId(branch.getTo());
+        String voltageLevel1Id = containerMapping.busNumToVoltageLevelId.get(branch.getFrom());
+        String voltageLevel2Id = containerMapping.busNumToVoltageLevelId.get(branch.getTo());
+        VoltageLevel voltageLevel2 = network.getVoltageLevel(voltageLevel2Id);
+        double zb = Math.pow(voltageLevel2.getNominalV(), 2) / perUnitContext.getBaseMva();
```
mathbagu, line 302: Same remark

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
```diff
@@ -0,0 +1,469 @@
...
+        return Math.abs(branch.getStatus()) > 0;
+    }
+
+    private static boolean isInService(MGen generator) {
+        return generator.getStatus() > 0;
+    }
+
+    private static void createLine(MBranch branch, ContainersMapping containerMapping, Network network, PerUnitContext perUnitContext) {
+        String lineId = getBranchId('L', branch.getFrom(), branch.getTo(), network);
+        LOGGER.debug("Creating line {}", lineId);
```
mathbagu, line 296: To be consistent, maybe you should also log bus1 and bus2 like for transformers

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
```diff
@@ -0,0 +1,469 @@
...
+
+    @Override
+    public Network importData(ReadOnlyDataSource dataSource, NetworkFactory networkFactory, Properties parameters) {
+        Objects.requireNonNull(dataSource);
+        Objects.requireNonNull(networkFactory);
+        Network network = networkFactory.createNetwork(dataSource.getBaseName(), FORMAT);
+
+        // no info abount time & date from the matpower file, set a  default
+        ZonedDateTime caseDateTime = DEFAULTDATE.atStartOfDay(ZoneOffset.UTC.normalized());
+        network.setCaseDate(new DateTime(caseDateTime.toInstant().toEpochMilli(), DateTimeZone.UTC));
```
mathbagu, line 425: See above, the default date is now. For round trip, you have to set-up a post-processor maybe to fix the date. I'm pretty sure this is the reason why you fix the date!?

matpower/matpower-model/src/main/java/com/powsybl/matpower/model/MBusTypeConversion.java
```diff
@@ -0,0 +1,24 @@
...
+public class MBusTypeConversion extends ObjectConversion<MBus.Type> {
+    @Override
+    protected MBus.Type fromString(String s) {
+        return MBus.Type.values()[Integer.parseInt(s.trim()) - 1];
+    }
+
+    @Override
+    public String revert(MBus.Type type) {
+        return Integer.toString(type.ordinal() + 1);
+    }
```
mathbagu, line 23: Why don't you use the fromInt() and getValue() methods? If there are useless, remove them?

matpower/matpower-model/src/main/java/com/powsybl/matpower/model/MatpowerBinReader.java
```diff
@@ -0,0 +1,129 @@
...
+
+import static com.powsybl.matpower.model.MatpowerReader.MATPOWER_SUPPORTED_VERSION;
+
+/**
+ * @author Christian Biasuzzi <christian.biasuzzi@techrain.eu>
+ */
+public class MatpowerBinReader {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(MatpowerBinReader.class);
+    public static final String MATPOWER_STRUCT_NAME = "mpc";
```
mathbagu, line 31: See my previous comment about constant order.

matpower/matpower-model/src/main/java/com/powsybl/matpower/model/MatpowerBinReader.java
```diff
@@ -0,0 +1,129 @@
...
+
+    public MatpowerModel read(InputStream iStream, String caseName) throws IOException {
+        MatpowerModel model = null;
+        try (MatFile mat = Mat5.newReader(Sources.wrapInputStream(iStream)).setEntryFilter(entry -> entry.getName().equals(MATPOWER_STRUCT_NAME)).readMat()) {
+            if (mat.getNumEntries() == 0) {
+                throw new IllegalStateException("not a MATPOWER stream: a structure named '" + MATPOWER_STRUCT_NAME + "' is expected.");
+            }
+            Struct mpcStruct = mat.getStruct(MATPOWER_STRUCT_NAME);
+            Set<String> mpcNames = new HashSet<>(Arrays.asList("version", "baseMVA", "bus", "gen", "branch"));
```
mathbagu, line 40: This is not the proper way to initialize a set from a list of constants. Use `Sets.of`

matpower/matpower-model/src/main/java/com/powsybl/matpower/model/MatpowerBinWriter.java
```diff
@@ -0,0 +1,131 @@
...
+        return struct;
+    }
+
+    public void write(File oFile) throws IOException {
+        try (Struct struct = fillMatStruct(Mat5.newStruct())) {
+            try (MatFile matFile = Mat5.newMatFile().addArray(MatpowerBinReader.MATPOWER_STRUCT_NAME, struct)) {
+                Mat5.writeToFile(matFile, oFile);
+            }
+        }
+    }
```
mathbagu, line 109: It's old school code to use I/O ? Does a version with Path exists?

We should have an overload with a Path that call the version with the OutputStream (Files.newOutputStream(path))

matpower/matpower-model/src/main/java/com/powsybl/matpower/model/MatpowerBinWriter.java
```diff
@@ -0,0 +1,131 @@
...
+            try (Struct struct = fillMatStruct(Mat5.newStruct())) {
+                try (MatFile matFile = Mat5.newMatFile().addArray(MatpowerBinReader.MATPOWER_STRUCT_NAME, struct)) {
+                    ByteBuffer bBuffer = getByteBuffer(matFile);
+                    channel.write(bBuffer);
+                }
+            }
+        }
+    }
+
+    private ByteBuffer getByteBuffer(MatFile matFile) throws IOException {
```
mathbagu, line 122: Could be static

matpower/matpower-model/src/main/java/com/powsybl/matpower/model/MatpowerModel.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public String getVersion() {
+        return version;
+    }
+
+    public void setVersion(String version) {
+        this.version = version;
+    }
+
+    private String version;
```
mathbagu, line 35: Put these methods after the constructor

matpower/matpower-model/src/main/java/com/powsybl/matpower/model/MatpowerModel.java
```diff
@@ -0,0 +1,64 @@
...
+public class MatpowerModel {
+
+    private final String caseName;
+    private Double baseMva;
+
+    private List<MBus> buses = new ArrayList<>();
+
+    private List<MGen> generators = new ArrayList<>();
+
+    private List<MBranch> branches = new ArrayList<>();
```
mathbagu, line 25: Should be final

matpower/matpower-model/src/main/java/com/powsybl/matpower/model/MatpowerReader.java
```diff
@@ -0,0 +1,154 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(MatpowerReader.class);
+
+    enum MatpowerSection {
+        BUS,
+        BRANCH,
+        GENERATOR
+    }
+
+    public static final String MATPOWER_SUPPORTED_VERSION = "2";
```
mathbagu, line 37: See my comment about constants

matpower/matpower-model/src/main/java/com/powsybl/matpower/model/MatpowerReader.java
```diff
@@ -0,0 +1,154 @@
...
+
+    private String processCaseName(String str) {
```
mathbagu, line 39: To be check if some or all of these functions could be static

matpower/matpower-model/src/main/java/com/powsybl/matpower/model/MatpowerReader.java
```diff
@@ -0,0 +1,154 @@
...
+                    lines.add(line);
+                }
+            }
+        }
+
+        return model;
+    }
+
+    private void processBaseMva(String line, MatpowerModel model) {
+        Double baseMva = Double.parseDouble(processMatlabAssignment(line));
```
mathbagu, line 105: Use a double instead of a Double

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerNetworkFactory.java
CBiasuzzi: test cases are the same tests as the ones used in ieee-cdf module (cdf files converted to the matpower format by a MATLAB tool cdf2mpc), so their licenses should be OK, I suppose ...

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
CBiasuzzi: thanks for your feedback,
- fixed the default date: now it's now()
- aligned the constant declarations order to the coding style

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
CBiasuzzi: again, I used the same test cases as the ieee-cdf importer, I thought it 'd be useful to provide the same options...
I've changed the prefix for the parameter as you suggested

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
CBiasuzzi: my general remark for all the ID convention related feedbacks&suggestions: I tried to be as consistent as I could 
with respect to the existing cdf importer approach (ref. the iidm in the ieee test resources ) in order to have the same generated IDs
(and make it possible to directly compare the iidm, for example)

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
CBiasuzzi: it seems OK (consistent with the ieee-cdf importer)

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
CBiasuzzi: you're right! done.

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
CBiasuzzi: done.

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
CBiasuzzi: done: I've fixed the date in the test, before the check takes place.

matpower/matpower-model/src/main/java/com/powsybl/matpower/model/MatpowerBinReader.java
CBiasuzzi: done.

matpower/matpower-model/src/main/java/com/powsybl/matpower/model/MatpowerBinReader.java
CBiasuzzi: isn't Sets.of java9?  I changed it with a guava method, instead.

matpower/matpower-model/src/main/java/com/powsybl/matpower/model/MatpowerBinWriter.java
CBiasuzzi: done.

matpower/matpower-model/src/main/java/com/powsybl/matpower/model/MatpowerBinWriter.java
CBiasuzzi: done:  changed this class' methods to static

matpower/matpower-model/src/main/java/com/powsybl/matpower/model/MatpowerModel.java
CBiasuzzi: done

matpower/matpower-model/src/main/java/com/powsybl/matpower/model/MatpowerModel.java
CBiasuzzi: done

matpower/matpower-model/src/main/java/com/powsybl/matpower/model/MatpowerReader.java
CBiasuzzi: done

matpower/matpower-model/src/main/java/com/powsybl/matpower/model/MatpowerReader.java
CBiasuzzi: done:  changed this class' methods to static

matpower/matpower-model/src/main/java/com/powsybl/matpower/model/MatpowerReader.java
CBiasuzzi: done

matpower/matpower-model/src/main/java/com/powsybl/matpower/model/MBusTypeConversion.java
CBiasuzzi: fixed

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
CBiasuzzi: there are not, apparently. fixed

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
CBiasuzzi: fixed

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
CBiasuzzi: fixed: used a prefix, consistently

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
CBiasuzzi: done

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
CBiasuzzi: done

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
CBiasuzzi: fixed: used a prefix, consistently

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
CBiasuzzi: fixed

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
CBiasuzzi: fixed: used a prefix, consistently

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
CBiasuzzi: fixed

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
CBiasuzzi: fixed: used a prefix, consistently

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
CBiasuzzi: fixed

matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerImporter.java
CBiasuzzi: fixed: used a prefix, consistently

geofjamg:
Thanks for this impressive work. 

The only problem I have with this PR is regarding .m import. I think we should not support this. .m files are is a data file but a code file. The parser you wrote only write for a well defined arbitrary indentation. What about if arrays would not be formatted one row per line? And if arrays would be created by more complex code like a for loop? The only way to process a .m file would to to run it and get "state" at the end. Trying to parse a .m file to get data is really not robust enough.

Fortunately, this is the purpose of .mat (binary) file which is a data file and the right way to import Matpower data in a Java program.

Personally I would:
 - remove from your PR everything regarding .m support
 - provide a script that convert a .m to .mat using Mathlab / Octave. That would require to install for instance Octave and Matpower toolbox to use it.  
 - Maybe we could recycle what you did with .m parsing in a .m -> .mat pure Java converter provided "as it is" but not in powsybl-core. Why not in powsybl-incubator? If someone wants to use it and it doesn't work we won't provide any support.

CBiasuzzi:
thanks for the review, just a clarification:
the .m parser is for .m files exported by MATPOWER itself ... 
Apparently, MATPOWER is consistent when it writes .m files, in its export functions (i.e. cdf2mpc, used to convert the original ieee  tests to MATPOWER  and the savecase(".m") function). 
Also, the .m files in the MATPOWER distribution seem to be well formatted  and indented.
So it's true that the .m importer is not robust enough to interpret a generic .m file, but it should be OK for .m files exported by the dedicated MATPOWER functions. With that said, I wonder if we could keep it in the core alongside the existing .mat  ...  maybe with the above notes in a README?

sylvlecl:
@mathbagu and @geofjamg : @CBiasuzzi proposed to move the .m text reader to test jar, so that it can still be used for unit tests purpose.

Is this alternative approach OK with you ? It looks good to me.
This way the code is still available for who might be interested in it, but not delivered in applications. It also avoids, for unit tests, to store binary .mat files.

geofjamg:
> @mathbagu and @geofjamg : @CBiasuzzi proposed to move the .m text reader to test jar, so that it can still be used for unit tests purpose.
> 
> Is this alternative approach OK with you ? It looks good to me.
> This way the code is still available for who might be interested in it, but not delivered in applications. It also avoids, for unit tests, to store binary .mat files.

I would still remove it even from test. For testing I would just use a jackson serialization of the data model (there is only beans so nothing to configure to annotate I guess, object mapper should work for free). So tests would load data from json, convert to binary and then run converter.

PR #1258, opened 2020-04-03
Fix SynchronousComponent buses filtering

tadam50:

No

Bug fix

The class ConnectedComponentImpl do not improve/override to super class ComponentImpl
Into synchronous component the bus filter used is the same as connected component

Add new abstract method into ComponentImpl in order to define the filtering lambda in getBuses() & getBusStream
Add a new class SynchronousComponentImpl as ConnectedComponentImpl

**Other information**:

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractComponentImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractComponentImpl.java
@@ -6,19 +6,21 @@
  */
 package com.powsybl.iidm.network.impl;
 
-import com.google.common.collect.Iterables;
 import com.powsybl.iidm.network.Bus;
 import com.powsybl.iidm.network.Component;
 import com.powsybl.iidm.network.impl.util.Ref;
 
 import java.util.Objects;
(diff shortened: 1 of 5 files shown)


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractComponentImpl.java
```diff
@@ -44,11 +46,15 @@ public int getSize() {
...
+                            .collect(Collectors.toList());
     }
 
     @Override
     public Stream<Bus> getBusStream() {
-        return networkRef.get().getBusView().getBusStream().filter(bus -> bus.getConnectedComponent() == ComponentImpl.this);
+        return networkRef.get().getBusView().getBusStream().filter(filterOnThis());
     }
+
+    protected abstract Predicate<Bus> filterOnThis();
```
mathbagu, line 59: We should rename this method to give it a better name like getBusFilter() or getBusPredicate()

MioRtia: I did it, if it is okay for you, you can merge.

PR #1259, opened 2020-04-05
Fix NullPointerException when removing a voltage level with internal …

geofjamg:
…connections

No

Bug fix

A NullPointerException is thrown when we remove a voltage level that contains internal connections. This is because internal connections are modeled as a edge with no switch object attached.

A null check of switch object of the edge has been added and we can now safely remove voltage levels with internal connections.

**Other information**:

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevel.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevel.java
@@ -1080,8 +1080,10 @@ class NodeBreakerVoltageLevel extends AbstractVoltageLevel {
 
     private void removeAllSwitches() {
         for (SwitchImpl s : graph.getEdgesObject()) {
-            getNetwork().getIndex().remove(s);
-            getNetwork().getListeners().notifyRemoval(s);
+            if (s != null) {
+                getNetwork().getIndex().remove(s);
+                getNetwork().getListeners().notifyRemoval(s);
+            }
         }
         graph.removeAllEdges();
         switches.clear();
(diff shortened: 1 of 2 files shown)


geofjamg:
> `graph.removeAllEdges()` is called in `removeAllSwitches()`, meaning internal connections are removed as well in this method. Is it what we want?

Yes, should you probably only remove edge with s != null in the loop instead of the removeAllEdges. This method is strange because it has been set private and is only called when voltage level is deleted... so it not really part of the API and I think the primary intend what to clean all edges. In that case maybe what is wrong is not the implementation but the name.

geofjamg:
> > `graph.removeAllEdges()` is called in `removeAllSwitches()`, meaning internal connections are removed as well in this method. Is it what we want?
> 
> Yes, should you probably only remove edge with s != null in the loop instead of the removeAllEdges. This method is strange because it has been set private and is only called when voltage level is deleted... so it not really part of the API and I think the primary intend what to clean all edges. In that case maybe what is wrong is not the implementation but the name.

Looking again at the code I would just rename it: removeAllEdges

MioRtia:
> > > `graph.removeAllEdges()` is called in `removeAllSwitches()`, meaning internal connections are removed as well in this method. Is it what we want?
> > 
> > 
> > Yes, should you probably only remove edge with s != null in the loop instead of the removeAllEdges. This method is strange because it has been set private and is only called when voltage level is deleted... so it not really part of the API and I think the primary intend what to clean all edges. In that case maybe what is wrong is not the implementation but the name.
> 
> Looking again at the code I would just rename it: removeAllEdges

I agree with you

geofjamg:
> > > > `graph.removeAllEdges()` is called in `removeAllSwitches()`, meaning internal connections are removed as well in this method. Is it what we want?
> > > 
> > > 
> > > Yes, should you probably only remove edge with s != null in the loop instead of the removeAllEdges. This method is strange because it has been set private and is only called when voltage level is deleted... so it not really part of the API and I think the primary intend what to clean all edges. In that case maybe what is wrong is not the implementation but the name.
> > 
> > 
> > Looking again at the code I would just rename it: removeAllEdges
> 
> I agree with you

Done!

PR #1261, opened 2020-04-05
Add sub-network reduction predicate

geofjamg:

No

feature

A new predicate for network reduction is added. It allows to specify a center voltage level to keep and a voltage level depth to keep around the center voltage level.

**Other information**:

### DIFF
--- a/iidm/iidm-reducer/src/main/java/com/powsybl/iidm/reducer/SubNetworkPredicate.java
+++ b/iidm/iidm-reducer/src/main/java/com/powsybl/iidm/reducer/SubNetworkPredicate.java
@@ -58,12 +58,12 @@ public class SubNetworkPredicate implements NetworkPredicate {
             }
 
             @Override
-            public void visitLine(Line line, Line.Side side) {
+            public void visitLine(Line line, Branch.Side side) {
                 visitBranch(line, side);
             }
 
             @Override
-            public void visitTwoWindingsTransformer(TwoWindingsTransformer transformer, TwoWindingsTransformer.Side side) {
+            public void visitTwoWindingsTransformer(TwoWindingsTransformer transformer, Branch.Side side) {
                 visitBranch(transformer, side);
             }
 
--- a/iidm/iidm-reducer/src/main/java/com/powsybl/iidm/reducer/SubNetworkPredicate.java
+++ b/iidm/iidm-reducer/src/main/java/com/powsybl/iidm/reducer/SubNetworkPredicate.java
@@ -44,7 +44,7 @@ public class SubNetworkPredicate implements NetworkPredicate {
         voltageLevelIds.add(vl.getId());
 
         vl.visitEquipments(new DefaultTopologyVisitor() {
-            public void visitBranch(Branch branch, Branch.Side side) {
+            private void visitBranch(Branch branch, Branch.Side side) {
                 switch (side) {
                     case ONE:
                         traverse(branch.getTerminal2().getVoltageLevel(), depth + 1, maxDepth, voltageLevelIds);
--- a/iidm/iidm-reducer/src/main/java/com/powsybl/iidm/reducer/SubNetworkPredicate.java
+++ b/iidm/iidm-reducer/src/main/java/com/powsybl/iidm/reducer/SubNetworkPredicate.java
@@ -1,0 +1,115 @@
+/**
+ * Copyright (c) 2020, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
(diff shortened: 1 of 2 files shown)


iidm/iidm-reducer/src/main/java/com/powsybl/iidm/reducer/SubNetworkPredicate.java
```diff
@@ -0,0 +1,115 @@
...
+ * A network reducer predicate that allow reduction based on a center voltage level and all other voltage level neighbors
+ * within a specified depth.
+ *
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ */
+public class SubNetworkPredicate implements NetworkPredicate {
+
+    private String voltageLevelId;
+
+    private int maxDepth;
```
mathbagu, line 26: Declare these variables final.

iidm/iidm-reducer/src/main/java/com/powsybl/iidm/reducer/SubNetworkPredicate.java
```diff
@@ -0,0 +1,115 @@
...
+        if (delegate != null) {
+            return;
+        }
+        VoltageLevel vl = network.getVoltageLevel(voltageLevelId);
+        if (vl == null) {
+            throw new PowsyblException("Voltage level '" + voltageLevelId + "' not found");
+        }
+        Set<String> voltageLevelIds = new LinkedHashSet<>();
+        traverse(vl, 0, maxDepth, voltageLevelIds);
+        delegate = new IdentifierNetworkPredicate(voltageLevelIds);
```
mathbagu, line 48: I think it's not a good design: the delegate is in fact final (you skip this part if the delegate is already initialized), but initialized after the constructor. I propose to make it delegate final, and initialize the SubNetworkPredicate using a VoltageLevel instead of its ID.

If you want to reuse the SubAreaNetwork for different VL or networks, the `init()` method should return the predicate and you should use a local variable.

iidm/iidm-reducer/src/main/java/com/powsybl/iidm/reducer/SubNetworkPredicate.java
```diff
@@ -0,0 +1,115 @@
...
+
+    private static void traverse(VoltageLevel vl, int depth, int maxDepth, Set<String> voltageLevelIds) {
+        if (voltageLevelIds.contains(vl.getId()) || depth > maxDepth) {
+            return;
+        }
+
+        voltageLevelIds.add(vl.getId());
+
+        vl.visitEquipments(new DefaultTopologyVisitor() {
+            public void visitBranch(Branch branch, Branch.Side side) {
```
mathbagu, line 59: Make this method private

iidm/iidm-reducer/src/main/java/com/powsybl/iidm/reducer/SubNetworkPredicate.java
```diff
@@ -0,0 +1,115 @@
...
+                        traverse(transformer.getLeg3().getTerminal().getVoltageLevel(), depth + 1, maxDepth, voltageLevelIds);
+                        break;
+                    case THREE:
+                        traverse(transformer.getLeg1().getTerminal().getVoltageLevel(), depth + 1, maxDepth, voltageLevelIds);
+                        traverse(transformer.getLeg2().getTerminal().getVoltageLevel(), depth + 1, maxDepth, voltageLevelIds);
+                        break;
+                    default:
+                        throw new IllegalStateException("Unknown side: " + side);
+                }
+            }
```
mathbagu, line 100: You should visit HVDC lines also?

iidm/iidm-reducer/src/main/java/com/powsybl/iidm/reducer/SubNetworkPredicate.java
geofjamg: Ok done

iidm/iidm-reducer/src/main/java/com/powsybl/iidm/reducer/SubNetworkPredicate.java
geofjamg: Done

iidm/iidm-reducer/src/main/java/com/powsybl/iidm/reducer/SubNetworkPredicate.java
geofjamg: Good question. In my own use case, I didn't want to traverse HVDC lines. I can imagine in some other use case, it would make sens.
What should I do?
  - let as it is and if somebody wants one day, we will consider to add it.
  - optionally add it right now

mathbagu: It makes sense to do it now. Do we consider it's an option enabled by a boolean or do we consider it always traverse the HVDC lines

PR #1262, opened 2020-04-06
Validations around HVDC lines in IIDM

annetill:

The documentation in the website will follow in the Following days.

No issue.

It adds some validations around HVDC lines in the internal model.

No check about the sign of active power setpoint, no hupper limits for powerFactor and lossFactor. These two factors are ratios (whitout units) but could be set in percents without new checks.

The active power setpoint cannot be negative. Ratios could not be more than 3. This number is arbitrary.

**Other information**:

I think that the AMPL conversion has to be fixed. I have made a fix for the CGMES conversion for active power setpoint. The factors management have to be reviewed.

### DIFF
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
@@ -33,6 +33,14 @@ public final class ValidationUtil {
         }
     }
 
+    public static void checkHvdcActivePowerSetpoint(Validable validable, double activePowerSetpoint) {
+        if (Double.isNaN(activePowerSetpoint)) {
+            throw createInvalidValueException(validable, activePowerSetpoint, "active power setpoint");
+        } else if (activePowerSetpoint < 0) {
+            throw createInvalidValueException(validable, activePowerSetpoint, "active power setpoint should not be negative");
+        }
+    }
+
     public static void checkActivePowerLimits(Validable validable, double minP, double maxP) {
         if (minP > maxP) {
             throw new ValidationException(validable, "invalid active limits [" + minP + ", " + maxP + "]");
@@ -365,6 +373,8 @@ public final class ValidationUtil {
     public static void checkPowerFactor(Validable validable, double powerFactor) {
         if (Double.isNaN(powerFactor)) {
             throw new ValidationException(validable, "power factor is invalid");
+        } else if (Math.abs(powerFactor) > 3) {
+            throw new ValidationException(validable, "power factor is invalid, it should be a ratio");
         }
     }
 
@@ -385,6 +395,8 @@ public final class ValidationUtil {
             throw new ValidationException(validable, "loss factor is invalid");
         } else if (lossFactor < 0) {
             throw new ValidationException(validable, "loss factor must be >= 0");
+        } else if (lossFactor > 3) {
+            throw new ValidationException(validable, "loss factor must be a ratio");
         }
     }
 }
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
@@ -111,6 +111,14 @@ public final class ValidationUtil {
         }
     }
 
+    public static void checkHvdcMaxP(Validable validable, double maxP) {
+        if (Double.isNaN(maxP)) {
+            throw createInvalidValueException(validable, maxP, "maximum P");
+        } else if (maxP < 0) {
+            throw createInvalidValueException(validable, maxP, "maximum P");
+        }
+    }
+
     public static void checkRegulatingTerminal(Validable validable, Terminal regulatingTerminal, Network network) {
         if (regulatingTerminal != null && regulatingTerminal.getVoltageLevel().getNetwork() != network) {
             throw new ValidationException(validable, "regulating terminal is not part of the network");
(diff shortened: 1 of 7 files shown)


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -365,6 +381,8 @@ public static void checkConvertersMode(Validable validable, HvdcLine.ConvertersM
     public static void checkPowerFactor(Validable validable, double powerFactor) {
         if (Double.isNaN(powerFactor)) {
             throw new ValidationException(validable, "power factor is invalid");
+        } else if (Math.abs(powerFactor) > 3) {
+            throw new ValidationException(validable, "power factor is invalid, it should be a ratio");
```
marqueslanauja, line 385: The power factor is defined as the ratio of the active power to the apparent power
`powerFactor = P/ Math.hypot(P, Q)`
so the powerFactor is a number in the closed interval [-1,1], `-1 <= powerFactor <= 1`

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -385,6 +403,8 @@ public static void checkLossFactor(Validable validable, float lossFactor) {
             throw new ValidationException(validable, "loss factor is invalid");
         } else if (lossFactor < 0) {
             throw new ValidationException(validable, "loss factor must be >= 0");
+        } else if (lossFactor > 3) {
```
marqueslanauja, line 406: The lossFactor is defined and calculated as a percentage, so `0 <= lossFactor <= 100`

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
annetill: Nice thanks, I was not sure about the hupper limit. If it is 1, it is really easiest ! I am going to change the PR.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
annetill: In fact, I ask @mathbagu and @geofjamg and I have checked in some private parts and the ````lossFactor````is not a percentage, it is a ratio. So:
- What is the hupper limit for you ? 1 or more (a ratio could be greater than 1, as a percentage can be greater than 100) ?
- You have to fix your PR about the CGMES conversion of HVDC (sorry for that) because the web site documentation was wrong.
-

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -365,6 +381,8 @@ public static void checkConvertersMode(Validable validable, HvdcLine.ConvertersM
     public static void checkPowerFactor(Validable validable, double powerFactor) {
         if (Double.isNaN(powerFactor)) {
             throw new ValidationException(validable, "power factor is invalid");
+        } else if (Math.abs(powerFactor) > 1) {
+            throw new ValidationException(validable, "power factor is invalid, it should be between -1 and 1");
```
mathbagu, line 385: If the power factor is between -1 and 1, we should also check the lower bound

annetill: I cannot make it with Math.abs ?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
marqueslanauja: According with the master version the lossFactor is calculated and expressed as a percentage and to be coherent with the code implementation should be a number in the closed interval [0,100]. (See code lines 179 - 186 and 41 - 93 of the  `DcLineSegmentConversion.java` class). 

It is a valid way to express the loss factor and in the IIDM model there are other attributes expressed in the same way, e.g. the step correction in `ratioTapChanger` and `phaseTapChanger` but we are in the IIDM model so RTE can choose a different way to express it.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
mathbagu: You right

marqueslanauja: I prefer to use  the standard definition of power factor `P/sqrt(P*P + Q*Q)`. You can check the wikipedia definition (https://en.wikipedia.org/wiki/Power_factor).

In our particular case the active power `P` is always positive so the power factor will be in the interval [0,1] but I prefer to check it between -1 and 1 as we already have an additional check for the active power to be positive.  This is my particular opinion.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
annetill: In fact, the DCLineSegmentConversion.Java has been modified with percentage last year by Miora and me. Miora and me have been influenced by the documentation. You are right about RTC and PTC.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
annetill: Finally, do you agree with this part of the code ? I "accidentaly" merge this PR minutes ago…

MioRtia: Yes, I've just finished to review it, it is okay for me, we can leave it merged as it is!

marqueslanauja: Totally agree.

PR #1263, opened 2020-04-06
merging inconsistent tie-line element names

benrejebmoh:
Signed-off-by: benrejebmoh <mohamed.ben-rejeb@rte-france.com>

Bug fix

The element name information is lost in UCTE export when the element names in two sides of tie-line are inconsistent

beTestGridForMerging:
BBBBBB11 XXXXXX11 1 9 0.0000 0.2000 0.000000   5051 ABCDE

frTestGridForMerging:
FFFFFF11 XXXXXX11 1 9 0.0000 0.2000 0.000000   5051 ZZZZZ

current behavior: merged file uxTestGridForMerging:
##L
BBBBBB11 XXXXXX11 1 9 0.0000 0.2000 0.000000   5051 
FFFFFF11 XXXXXX11 1 9 0.0000 0.2000 0.000000   5051 

new behavior: merged file uxTestGridForMerging:
we keep the original properties of each tieline , post-fixing the name of the properties with _1 or _2
when creating danglinglines.
in UCTE export for geographical name we use the merged properties but for element name we use the original tieline element name propertiies suffixed with _1 and _2

The result for the use case above will be:
##L
BBBBBB11 XXXXXX11 1 9 0.0000 0.2000 0.000000   5051 ABCDE
FFFFFF11 XXXXXX11 1 9 0.0000 0.2000 0.000000   5051 ZZZZZ

**Other information**:


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
```diff
@@ -1013,7 +1015,11 @@ private void mergeProperties(DanglingLine dl1, DanglingLine dl2, Properties prop
                 LOGGER.warn("Inconsistencies of property '{}' between both sides of merged line. Side 2 is empty, keeping side 1 value '{}'", prop, dl1.getProperty(prop));
                 properties.setProperty(prop, dl1.getProperty(prop));
             } else {
-                LOGGER.error("Inconsistencies of property '{}' between both sides of merged line. '{}' on side 1 and '{}' on side 2. Removing the property of merged line", prop, dl1.getProperty(prop), dl2.getProperty(prop));
+                if (!prop.equals(ELEMENT_NAME_PROPERTY)) {
+                    LOGGER.error("Inconsistencies of property '{}' between both sides of merged line. '{}' on side 1 and '{}' on side 2. Removing the property of merged line", prop, dl1.getProperty(prop), dl2.getProperty(prop));
+                } else {
+                    properties.setProperty(prop, String.format("%s,%s", dl1.getProperty(prop), dl2.getProperty(prop)));
```
mathbagu, line 1021: Replace String.format by a simple concatenation (performance issue)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
```diff
@@ -1013,7 +1015,11 @@ private void mergeProperties(DanglingLine dl1, DanglingLine dl2, Properties prop
...
+                }
```
mathbagu, line 1022: I'm not sure we should make a specific case for element names. Maybe we should decide the strategy, and apply it to all properties.
@murgeyseb What do you think?

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteExporter.java
```diff
@@ -475,10 +475,17 @@ private static void convertTieLine(UcteNetwork ucteNetwork, TieLine tieLine, Uct
         // FIXME(mathbagu) REAL or EQUIVALENT ?
         UcteElementStatus status = tieLine.getTerminal1().isConnected() && tieLine.getTerminal2().isConnected() ? EQUIVALENT_ELEMENT_IN_OPERATION : EQUIVALENT_ELEMENT_OUT_OF_OPERATION;
 
+        String elementName = tieLine.getProperty(ELEMENT_NAME_PROPERTY_KEY, null);
+        String elementName1 = elementName;
+        String elementName2 = elementName;
+        if (elementName != null && elementName.contains(",")) {
+            elementName1 = elementName.split(",")[0];
+            elementName2 = elementName.split(",")[1];
+        }
```
mathbagu, line 484: The property separator should be a constant, just in case this is not well chosen and we have to change it.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
murgeyseb: Yes, I think we could keep the same behaviour for Element name, and change the parsing of the property in the exporter to create a unique element name if needed, using the same strategy as before (if equals use it, if one empty use the other, otherwise no element name)

benrejebmoh: But if we don't make a specific case for element names this strategy of concatenating the two  side's property will impcact all types of property. for example the geographical name it will create other problems. Should i delete this condition here ? but make another one in exporter saying if geographival name contains the separator "," --> don't export it

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteExporter.java
```diff
@@ -478,10 +502,23 @@ private static void convertTieLine(UcteNetwork ucteNetwork, TieLine tieLine, Uct
...
         UcteElementStatus status = tieLine.getTerminal1().isConnected() && tieLine.getTerminal2().isConnected() ? EQUIVALENT_ELEMENT_IN_OPERATION : EQUIVALENT_ELEMENT_OUT_OF_OPERATION;
 
+        String elementName = tieLine.getProperty(ELEMENT_NAME_PROPERTY_KEY, null);
+        String elementName1 = elementName;
+        String elementName2 = elementName;
+        if (elementName != null && elementName.contains(PROPERTY_SEPARATOR)) {
+            elementName1 = elementName.split(PROPERTY_SEPARATOR, -1)[0];
+            if (elementName1.isEmpty()) {
+                elementName1 = null;
+            }
```
mathbagu, line 512: This is not necessary: empty is a good enough to be serialized.

mathbagu:
See with @murgeyseb, it's OK to lose the geographicalName when there are inconsistencies between the two IGMs

PR #1265, opened 2020-04-07
[WIP] add post processor for ucte exporter

### LINKED ISSUE #1115: Generate a custom comment in UCTE files header
Feature

UCTE exporter generate a UCTE file containing no file description in the generated comment header. However, ENTSO-E quality of data sets and calculations says (page 7 of the document https://docstore.entsoe.eu/Documents/Publications/SOC/Continental_Europe/150420_quality_of_datasets_and_calculations_3rd_edition.pdf):

> The comment block (##C) at the top of data should describe at least the provider of the file.

Though it is important that a PowSyBl user can add a comment in the header part of UCTE file exported to define the provider of the file.

UcteExporter should give the user the possibility of setting a comment for the generated file (via network property, exporter property, or any relevant combination of these).

Generate UCTE files of industrial quality.

jlabous:

fix #1115


ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteExportScriptPostProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+            .orElse(defaultScript);
+    }
+
+    @Override
+    public String getName() {
+        return NAME;
+    }
+
+    @Override
+    public void process(Network network, UcteNetwork ucteNetwork) throws Exception {
```
geofjamg, line 68: @mathbagu @murgeyseb 
Why is it a specific UCTE post processor? Why not like IIDM import post processor an IIDM export post processor and an "Object" to access original data model?

```java
public void process(Network network, String format, Object nativeDataModel);
```

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteExportPostProcessor.java
```diff
@@ -0,0 +1,16 @@
+package com.powsybl.ucte.converter;
```
mathbagu, line 1: Add missing license header

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteExportPostProcessor.java
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ *
+ * @author Jérémy LABOUS <jlabous@silicom.fr>
+ */
+public interface UcteExportPostProcessor {
+
+    String getName();
+
+    void process(Network network, UcteNetwork ucteNetwork) throws Exception;
```
mathbagu, line 14: Not sure this method should throw something

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteExportScriptPostProcessor.java
mathbagu: It could be done like you suggest. In our first approach we just want to have a post-processor to the conversion of an IIDM model to a UCTE model. We can consider this need is generic and move this to the IIDM API.

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteExporter.java
```diff
@@ -86,6 +86,13 @@ public void export(Network network, Properties parameters, DataSource dataSource
 
         UcteNetwork ucteNetwork = createUcteNetwork(network, namingStrategy);
 
+        UcteExportPostProcessor processor = new UcteExportScriptPostProcessor();
+        try {
+            processor.process(network, ucteNetwork);
+        } catch (Exception e) {
+            throw new PowsyblException("Error during post processing");
+        }
```
mathbagu, line 94: @geofjamg This code would be duplicate in each exporter class, right?

ucte/ucte-converter/src/test/java/com/powsybl/ucte/converter/UcteExporterTest.java
```diff
@@ -36,12 +40,12 @@
      * @param filePath path of the file relative to resources directory
      * @return imported network
      */
-    private static Network loadNetworkFromResourceFile(String filePath) {
+    public static Network loadNetworkFromResourceFile(String filePath) {
```
mathbagu, line 43: There is no need tio make this method public, private package seems ok

ucte/ucte-converter/src/test/java/com/powsybl/ucte/converter/UcteExporterTest.java
```diff
@@ -36,12 +40,12 @@
...
         ReadOnlyDataSource dataSource = new ResourceDataSource(FilenameUtils.getBaseName(filePath), new ResourceSet(FilenameUtils.getPath(filePath), FilenameUtils.getName(filePath)));
         return new UcteImporter().importData(dataSource, NetworkFactory.findDefault(), null);
     }
 
-    private static void testExporter(Network network, String reference) throws IOException {
+    public static void testExporter(Network network, String reference) throws IOException {
```
mathbagu, line 48: Same remark here

mathbagu:
@jlabous: after a small telco with Geoffroy, we think that it's a good idea to make this mechanism less specific to the UCTE. We could have the same kind of use cases for other supported formats.
What we propose is :
- Move the interface in iidm-converter-api module, and integrate it like it's done for the equivalent ImportPostProcessor in `Importers`.
- Move the implementation in the iidm-scripting module that already depends on groovy. The new process method would look like this:
`void process(Network network, String format, Object nativeModel, ComputationManager computationManager)`

We decide to add the computationManager, in the case someone want to run a computation

If something seems unclear for you, we could do a telco on slack.

ucte/ucte-converter/pom.xml
```diff
@@ -53,6 +53,11 @@
             <artifactId>powsybl-iidm-converter-api</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>powsybl-iidm-scripting</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
mathbagu, line 60: This is not the correct way to do it: the post-processor should be discovered at runtime using the ServiceLoader. You should have a look to the `Importers` class to see how it works. The idea is to have something similar for exporters.

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteExporter.java
```diff
@@ -50,18 +53,25 @@
 
     public static final String NAMING_STRATEGY = "ucte.export.naming-strategy";
 
+    public static final String POST_PROCESSOR = "ucte.post-processor.script";
+
```
mathbagu, line 57: To be removed

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteExporter.java
```diff
@@ -50,18 +53,25 @@
...
     private static final Parameter NAMING_STRATEGY_PARAMETER = new Parameter(NAMING_STRATEGY, ParameterType.STRING, "Default naming strategy for UCTE codes conversion", "Default");
 
+    private static final Parameter POST_PROCESSOR_PARAMETER = new Parameter(POST_PROCESSOR, ParameterType.STRING, "Post processor script for ucte export", null);
+
```
mathbagu, line 61: To be removed

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteExporter.java
```diff
@@ -50,18 +53,25 @@
...
     private static final Supplier<List<NamingStrategy>> NAMING_STRATEGY_SUPPLIERS
             = Suppliers.memoize(() -> new ServiceLoaderCache<>(NamingStrategy.class).getServices())::get;
 
     private final ParameterDefaultValueConfig defaultValueConfig;
 
+    private final PlatformConfig config;
+
```
mathbagu, line 68: To be removed

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteExporter.java
```diff
@@ -86,6 +96,13 @@ public void export(Network network, Properties parameters, DataSource dataSource
 
         UcteNetwork ucteNetwork = createUcteNetwork(network, namingStrategy);
 
+        String postProcessorScript = ConversionParameters.readStringParameter(getFormat(), parameters, POST_PROCESSOR_PARAMETER, defaultValueConfig);
+        if (postProcessorScript != null) {
+            Path script = config.getConfigDir().resolve(postProcessorScript);
+            ExportPostProcessor processor = new GroovyScriptExportPostProcessor(script);
+
+            processor.process(network, null, ucteNetwork, null);
+        }
```
mathbagu, line 105: The list of the PostProcessor to run should be passed as parameters to this method or discovered automatically. Passing these information to the method is a breaking change, so be carful: you have to manage the compatibility with a default behavior to the new method. But I think it's the proper way to do it... otherwise we could not pass an explicit list of post-processors
@geofjamg What do you think?

geofjamg: Yes, we can pass the list of post processor to the export method and with default methods it should be easy to get backward compatibility.

iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/export/Exporter.java
```diff
@@ -43,4 +55,25 @@
...
     void export(Network network, Properties parameters, DataSource dataSource);
 
+    /**
+     * Conversion from iidm format.
+     *
+     * @param network the iidm model
+     * @param parameters some properties to configure the export
+     * @return the native data model
+     */
+    default Object conversion(Network network, Properties parameters) {
```
geofjamg, line 65: Splitting the export into network -> nativeDataModel then nativeDataModel -> file is probably not a good idea as we can imagine in some cases no need to intermediate native data model. Also for most of the use cases we don't care about of the native data model and we don't want to see it in the export interface.
I would simply return it in the orginal export method and everybody can ignore it (and also it can be null)

jlabous: If we did like this, the export method generate the export file and return the nativeDataModel witch can be use by the postprocessor, but the postprocessor have to export another file (or override).
Is it really what we expect ?

geofjamg: You just have to pass the post processors to the export method and let the implementation call them (or not, just do it for UCTE exporter) at the right place in the export code.

PR #1270, opened 2020-04-10
Issue 890 ucte power plant type management

### LINKED ISSUE #890: Fix incorrect management of power plant type in UCTE import/export
*feature*

Currently, the kind of a generation unit in UCTE import is lost, due to an unconsistency between generation kind listed in UCTE-DEF format and IIDM ones. This can be impacting in the round-trip process, because initial data is lost and not usable for the export process.

The expected behaviour is such that a roundtrip with a UCTE-DEF formatted file in the platform keep any power plant type information in the initial file.

Avoid losing/modifying information available in UCTE-DEF input files

  - PowSyBl Version: 2.6.0

jlabous:

fix #890


ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
```diff
@@ -181,8 +187,14 @@ private static void createGenerator(UcteNode ucteNode, VoltageLevel voltageLevel
...
             switch (ucteNode.getPowerPlantType()) {
                 case C:
+                    voltageLevel.getNetwork().setProperty(UCTE_POWERPLANT_TYPE_KEY + generatorId, ucteNode.getPowerPlantType().toString());
+                    break;
                 case G:
+                    voltageLevel.getNetwork().setProperty(UCTE_POWERPLANT_TYPE_KEY + generatorId, ucteNode.getPowerPlantType().toString());
+                    break;
                 case L:
+                    voltageLevel.getNetwork().setProperty(UCTE_POWERPLANT_TYPE_KEY + generatorId, ucteNode.getPowerPlantType().toString());
+                    break;
```
mathbagu, line 197: The simplest way to implement that, is to add a property to the generator whatever the type of the power plant. Add the property after the creation of the generator:
```
generator.setProperty(UCTE_POWER_PLANT_TYPE, ucte.getPowerPlantType());
```

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteExporter.java
```diff
@@ -702,7 +702,7 @@ private static void setSwitchCurrentLimit(UcteLine ucteLine, Switch sw) {
         }
     }
 
-    private static UctePowerPlantType energySourceToUctePowerPlantType(EnergySource energySource) {
+    private static UctePowerPlantType energySourceToUctePowerPlantType(EnergySource energySource, String uctePowerPlantType) {
```
mathbagu, line 705: Change the signature of this method to pass the generator instead of these two values.
Look for the property. If it's defined, use it. If the property is not defined, use the existing code without modifications.

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteExporter.java
```diff
@@ -702,8 +702,11 @@ private static void setSwitchCurrentLimit(UcteLine ucteLine, Switch sw) {
         }
     }
 
-    private static UctePowerPlantType energySourceToUctePowerPlantType(EnergySource energySource) {
-        switch (energySource) {
+    private static UctePowerPlantType energySourceToUctePowerPlantType(Generator generator) {
+        if (generator.hasProperty(UcteImporter.UCTE_POWERPLANT_TYPE_KEY + generator.getId())) {
+            return UctePowerPlantType.valueOf(generator.getProperty(UcteImporter.UCTE_POWERPLANT_TYPE_KEY + generator.getId()));
+        }
```
mathbagu, line 708: There is no good reason to make the property name dependent on the ID of the generator.

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
```diff
@@ -45,6 +45,12 @@
 
     private static final String[] EXTENSIONS = {"uct", "UCT"};
 
+    /**
+     * Key to store ucte power plant type in properties
+     * Concat this key with generator id
+     */
+    public static final String UCTE_POWERPLANT_TYPE_KEY = "UctePowerPlantType_";
```
mathbagu, line 52: This property should be declared with the others: as it's used in both importer and exporter, the right place is UcteConstants.

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
```diff
@@ -181,8 +187,11 @@ private static void createGenerator(UcteNode ucteNode, VoltageLevel voltageLevel
         if (ucteNode.getPowerPlantType() != null) {
             switch (ucteNode.getPowerPlantType()) {
                 case C:
+                    break;
                 case G:
+                    break;
                 case L:
+                    break;
```
mathbagu, line 194: Look like a regression

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
```diff
@@ -219,6 +228,9 @@ private static void createGenerator(UcteNode ucteNode, VoltageLevel voltageLevel
                 .setMinQ(-ucteNode.getMinimumPermissibleReactivePowerGeneration())
                 .setMaxQ(-ucteNode.getMaximumPermissibleReactivePowerGeneration())
                 .add();
+        if (ucteNode.getPowerPlantType() != null) {
+            generator.setProperty(UCTE_POWERPLANT_TYPE_KEY + generatorId, ucteNode.getPowerPlantType().toString());
+        }
```
mathbagu, line 233: From my point of view, I think you should always add the property

PR #1276, opened 2020-04-15
Add splitShuntAdmittance in threeWindingsTransformer. Only one flag "twtSplitShuntAdmittance" for twoWindings and threeWindingsTransformers

marqueslanauja:

Feature

There is no possibility to split the shunt admittance in `threeWindingsTransformers` on Load Flow execution.

A new Load Flow parameter has been added so it is possible to configure the shunt admittance split in `threeWindingsTransformer `on Load Flow execution.

**Other information**:

`SplitShuntAdmittance `has not been extended to the ampl-converter package as this package is not using the current IIDM API of the `threeWindingTransformer`. We can make a new PR to update the API and add the `splitShuntAdmittance` parameter.


sensitivity-api/src/test/resources/SensitivityComputationParametersError.json
```diff
@@ -1,12 +0,0 @@
-{
-  "version" : "1.0",
```
annetill, line 2: It is normal that you have delete this file?

marqueslanauja: I found that it is not used.

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowParameters.java
```diff
@@ -84,14 +85,15 @@ protected static void load(LoadFlowParameters parameters, PlatformConfig platfor
...
-                    parameters.setT2wtSplitShuntAdmittance(config.getBooleanProperty("t2wtSplitShuntAdmittance", config.getBooleanProperty("specificCompatibility", DEFAULT_T2WT_SPLIT_SHUNT_ADMITTANCE)));
-                });
+            .ifPresent(config -> {
+                parameters.setVoltageInitMode(config.getEnumProperty("voltageInitMode", VoltageInitMode.class, DEFAULT_VOLTAGE_INIT_MODE));
+                parameters.setTransformerVoltageControlOn(config.getBooleanProperty("transformerVoltageControlOn", DEFAULT_TRANSFORMER_VOLTAGE_CONTROL_ON));
+                parameters.setNoGeneratorReactiveLimits(config.getBooleanProperty("noGeneratorReactiveLimits", DEFAULT_NO_GENERATOR_REACTIVE_LIMITS));
+                parameters.setPhaseShifterRegulationOn(config.getBooleanProperty("phaseShifterRegulationOn", DEFAULT_PHASE_SHIFTER_REGULATION_ON));
+                // keep old tag name "specificCompatibility" for compatibility
+                parameters.setT2wtSplitShuntAdmittance(config.getBooleanProperty("t2wtSplitShuntAdmittance", config.getBooleanProperty("specificCompatibility", DEFAULT_T2WT_SPLIT_SHUNT_ADMITTANCE)));
```
annetill, line 94: I understand that it is more coherent with our IIDM modeling to have 2 parameters to mention that the shunt as to be split, one for T2W and one for T3W. But for the powerflow engines I know, these two parameters should be identical (false for both or true for both). So I am not sure, I need to ask my colleagues.

annetill: It is maybe dangerous to have 2 parameters as in the use cases I have in mind these two parameters should be equal. Do you have in mind a use case where these two parameters could be different ?

marqueslanauja: No, but I will review the IOP cases to check it

marqueslanauja:
Migration Guide 3.4.0 has been updated.

Issue #1167, opened 2020-02-19
[UCTE] Assert that our import/export follow all ENTSO-E rules


mathbagu:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
The UCTE importer/exporter follows the rules defined in the [UCTE-DEF](https://cimug.ucaiug.org/Groups/Model%20Exchange/UCTE-format.pdf) specification format. But this specification is really old and a new ENTSO-E document seems to be the reference now. 

* **What is the motivation / use case for changing the behavior?**
If some TSO/RSC exchange files following the new rules, we should be able to correctly handle these files. For now, I see at least differences in the country code (D2, D4, D6, KS). The rule for order code has been updated in #1128 and #1166.

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

murgeyseb:
[QUALITY OF DATASETS AND CALCULATIONS 20130125.pdf](https://github.com/powsybl/powsybl-core/files/4510384/QUALITY.OF.DATASETS.AND.CALCULATIONS.20130125.pdf)

alicecaron:
Fixed by [MR](https://github.com/powsybl/powsybl-core/pull/1567)

PR #1288, opened 2020-04-20
[WIP] Create UCTE report to reduce log verbosity

geofjamg:

**Other information**:

### DIFF
--- a/ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
+++ b/ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
@@ -297,7 +297,7 @@ public class UcteImporter implements Importer {
                                                           UcteNodeCode nodeCode1, UcteNodeCode nodeCode2,
                                                           UcteVoltageLevel ucteVoltageLevel1, UcteVoltageLevel ucteVoltageLevel2,
                                                           boolean connected, double z) {
-        LOGGER.info("Create coupler '{}' from low impedance line ({})", ucteLine.getId(), z);
+        LOGGER.trace("Create coupler '{}' from low impedance line ({})", ucteLine.getId(), z);
 
         if (ucteVoltageLevel1 != ucteVoltageLevel2) {
             throw new UcteException("Nodes coupled with a low impedance line are expected to be in the same voltage level");
--- a/ucte/ucte-network/src/main/java/com/powsybl/ucte/network/UcteElement.java
+++ b/ucte/ucte-network/src/main/java/com/powsybl/ucte/network/UcteElement.java
@@ -148,7 +148,7 @@ public class UcteElement implements UcteRecord {
     private static final float MIN_X = 0.05f;
 
     @Override
-    public void fix() {
+    public void fix(UcteReport report) {
         switch (status) {
             case EQUIVALENT_ELEMENT_IN_OPERATION:
             case EQUIVALENT_ELEMENT_OUT_OF_OPERATION:
--- a/ucte/ucte-network/src/main/java/com/powsybl/ucte/network/UcteNetworkImpl.java
+++ b/ucte/ucte-network/src/main/java/com/powsybl/ucte/network/UcteNetworkImpl.java
@@ -112,18 +112,22 @@ public class UcteNetworkImpl implements UcteNetwork {
 
     @Override
     public void fix() {
+        UcteReport report = new UcteReport();
+
         for (UcteNode node : nodes.values()) {
-            node.fix();
+            node.fix(report);
         }
         for (UcteLine line : lines.values()) {
-            line.fix();
+            line.fix(report);
         }
         for (UcteTransformer transfo : transformers.values()) {
-            transfo.fix();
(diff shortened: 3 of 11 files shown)


mathbagu:
I disagree with the systematic change of log level to debug or trace. When data are not valid, it's definitely an error or a warning (not a debug or trace). If a user don't want to see these logs, the good way to do is to configure the log for this specific class, depending on the use case.

Anyway I think it's a nice feature to have a synthetic report at the end of the import. Most of the case, we don't the detail but only the synthesis. I will try to have a look to each log and ask myself if the log is relevant, if every interesting information is logged in the message.

geofjamg:
It is still a warning, but synthetic. If you want more details for the warning you just activate debug level.

geofjamg:
@mathbagu if you want we can postpone the discussion to how to reduce log verbosity to another issue/PR (it is a common issue with CGMES import either) by just reverting logs gravity (so debug -> warn back) and just let the synthetic report in addition to the individual warnings.
On my side I will be able to configure log level to error in ucte model to avoid this hundreds of warning (even I am convinced it is not the right thing to do because I could miss some important warning in the middle of not very serious one like all of fix() methods).
Do you agree with this proposition?

mathbagu:
@geofjamg: as a quick fix, I totally agree with your proposal!

From my point of view, in some use cases (not only to debug a file), someone (me for instance) would like to have all the detailed information on issues. This is the reason why I propose to introduce an option `verbose` in our importers (UCTE and CGMES). If this option is set to `false` (the default value), only the synthetic report is logged. If this option is set to `true`, each detailed message is loggged.

We also have an issue about checking issues in UCTE importer (#1167). In this issue, we want to record all inconsistencies. I thought we can have two implementation of the Report class you made:
- the synthetic one could be the implementation you propose
- the complete one could be the fix for #1167.
The choice of which implementation to use could be also a configuration property.

geofjamg:
Done, I revert all warn -> debug gravity changes. I only kept the info -> debug one because a low impedance line creation is a not an info (like all the other equipment creation logs).

mathbagu:
Thanks for the changes. I think you forgot to revert changes in UcteElement (some logs are still in debug).

@geofjamg What do you think about these two different implementations:
- V1: the UcteReport is a plugin, and user can select the implementation to use (synthetic or verbose, or a custom one) : https://github.com/powsybl/powsybl-core/commit/e7f9d6f077776a8985dfa0369d0374b438846aad
- V2: the UcteReport has a new verbose attribute to log details if the property is set to true (false by default): https://github.com/powsybl/powsybl-core/commit/756e9bc2e8f7178f7f3b5946d9076a1c354ccbc5

geofjamg:
> Thanks for the changes. I think you forgot to revert changes in UcteElement (some logs are still in debug).
> 
> @geofjamg What do you think about these two different implementations:
> 
>     * V1: the UcteReport is a plugin, and user can select the implementation to use (synthetic or verbose, or a custom one) : [e7f9d6f](https://github.com/powsybl/powsybl-core/commit/e7f9d6f077776a8985dfa0369d0374b438846aad)
> 
>     * V2: the UcteReport has a new verbose attribute to log details if the property is set to true (false by default): [756e9bc](https://github.com/powsybl/powsybl-core/commit/756e9bc2e8f7178f7f3b5946d9076a1c354ccbc5)

I really like the first one as it is more extensible than the second one. I would go for first.
I could also imagine a `UcteComposteReport` that would allows conbinning several type of reports?

mathbagu:
I don't know if it's planned, but currently there is a `Context` that logs this kind of changes. We could easily change the design to delegate to a CgmesReport.

@geofjamg as a plugin, how to you create the sub UcteReport for the composite implementation? With another configuration module which list the implementation to create?
```
ucte-composite-report:
    reports: "default,synthetic"
```

geofjamg:
Yes why not

mathbagu:
> Yes why not but with a yaml array

You right :+1:

geofjamg:
So @mathbagu we can merge your first branch in this PR and continue the dev?

mathbagu:
@geofjamg sorry for the delay: I pushed my 1st proposal. Coud you have a check in UcteNode.java: I add 2 todos because we check for inconsistency and log them, but we don't fix the value. I think something is missing but I was not totally sure.

geofjamg:
> @geofjamg sorry for the delay: I pushed my 1st proposal. Coud you have a check in UcteNode.java: I add 2 todos because we check for inconsistency and log them, but we don't fix the value. I think something is missing but I was not totally sure.

@mathbagu this PR is a very useful refactoring of existing code and is fine for me. I would remove the 2 TODO of this PR to just let the "refactoring part" and let the data fixes for another PR.

geofjamg:
I just figured out there is a serious design issue in the `UcteReport` design. As `UcteReport` is a the plugin interface, it is supposed to be stateless (as there is only one instance). In the synthetic implementation, log count is stored in the `UcteSyntheticReport` instance, which could lead to concurrency issue. We should add a `UcteReportProvider` class that will be the plugin entry point and that will allow to provide a new `UcteReport` instance at each conversion.

PR #1251, opened 2020-03-31
Regulating controls: terminal topological node can be obtained through connectivity node

zamarrenolm:
Signed-off-by: Luma Zamarreño <zamarrenolm@aia.es>

**Please check if the PR fulfills these requirements** *(please use `'[x]'` to check the checkboxes, or submit the PR and then click the checkboxes)*
- [X] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** *(Bug fix, feature, docs update, ...)*
For some regulating controls, the regulating terminal in IIDM must be obtained from the topological node associated to the CGMES terminal. The topological node is not always directly expressed as an attribute of the terminal. Sometimes, it must be obtained through the connectivity node linked to the terminal.

**What is the current behavior?** *(You can also link to an open issue here)*
Some regulating control terminals were lost because no topological node could be found. The CGMES data did not contain the explicit relationship between terminal and topological node. 

**What is the new behavior (if this is a feature change)?**
A topological node can be found from the terminal based on the explicit attribute in the TP data (`Terminal.TopologicalNode`) or through the connectivity node linked to the terminal in EQ: `Terminal.ConnectivityNode`, then using `ConnectivityNode.TopologicalNode` from TP.

The SPARQL query for obtaining regulating control  data (`query: regulatingControls`) is kept simple, and the complexity of dealing with alternative ways of obtaining a topological node from a terminal is kept in the `query: terminals`.

PR #1292, opened 2020-04-21
Check on qPercent is deleted to prevent regression in CGMES import

MioRtia:

Prevent regression

If qPercent has an invalid value during CGMES import, an exception is thrown.

If qPercent has an invalid value during CGMES import, the value is ignored.

No

**Other information**:
This bug was not seen before as some regulating controls went undetected in node/breaker (cf PR #1251 )

@mathbagu This should be in release 3.3.0 as it prevents a regression in CGMES import

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForGenerators.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForGenerators.java
@@ -112,13 +112,9 @@ public class RegulatingControlMappingForGenerators {
 
         // add qPercent as an extension
         if (!Double.isNaN(qPercent)) {
-            if (qPercent < 0.0 || qPercent > 100) {
-                context.ignored(QPERCENT, () -> String.format("Unexpected value for qPercent of generator %s: %s", gen.getId(), qPercent));
-            } else {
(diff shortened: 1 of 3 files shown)


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForGenerators.java
```diff
@@ -112,9 +112,13 @@ private boolean setRegulatingControlVoltage(String controlId,
 
         // add qPercent as an extension
         if (!Double.isNaN(qPercent)) {
-            gen.newExtension(CoordinatedReactiveControlAdder.class)
-                    .withQPercent(qPercent)
-                    .add();
+            if (qPercent < 0.0 || qPercent > 100) {
+                context.ignored(QPERCENT, () -> String.format("Unexpected value for qPercent of generator %s: %s", gen.getId(), String.valueOf(qPercent)));
```
mathbagu, line 116: Not sure valueOf is required. I think String.format is able to convert the value to a string

MioRtia:
> I think that all the code we have could be useful in the future and it could be also useful to have the information during the import. Maybe could you just log the result of the check without throughing an exception ?

Agreed. Is `DEBUG` an acceptable log level for you?

annetill:
Yes.

marqueslanauja:
I reviewed and it is ok.

PR #1296, opened 2020-04-27
Consider only .xml files that contain valid namespaces (RDF, CIM)

zamarrenolm:

Bug fix

In CGMES importer, all files inside a `DataSource` that match the filename pattern (`*<CGMES_SUBSET>*.XML`) are considered for import.

CGMES Importer will try to load only the files that match the filename pattern **AND** that contain XML text with references to valid namespaces. The RDF namespace is required. Either the CIM14 or CIM16 namespace must be present.

RDF namespace: http://www.w3.org/1999/02/22-rdf-syntax-ns#
CIM14 namespace: http://iec.ch/TC57/2009/CIM-schema-cim14#
CIM16 namespace: http://iec.ch/TC57/2013/CIM-schema-cim16#


zamarrenolm:
> @zamarrenolm Maybe you should also add a file that is not XML to complete the tests?

Added a file that does not have .xml extension. 
In the set of invalid files there was already a file with .xml extension but invalid XML content.
The unit test now checks that all invalid files are present in the data source but are ignored as CIM content.

PR #1301, opened 2020-04-29
Add method to write a unique subset

mathbagu:

N/A

Feature

It's not possible to write some of the subsets of a TripleStore

The new `write(DataSource, CgmesSubset)` method allow to write a unique subset.

**Other information**:


cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesModel.java
```diff
@@ -166,6 +166,8 @@ static String baseName(ReadOnlyDataSource ds) {
 
     void write(DataSource ds);
 
+    void write(DataSource ds, CgmesSubset subset);
```
MioRtia, line 169: Not sure it is used anywhere else but it may be safer to add a default implementation

triple-store/triple-store-api/src/main/java/com/powsybl/triplestore/api/TripleStore.java
```diff
@@ -41,6 +41,13 @@
      */
     void write(DataSource ds);
 
+    /**
+     * Write the contents of a context in the given data source.
+     * @param ds the output data source
+     * @param contextName the context to write
+     */
+    void write(DataSource ds, String contextName);
```
MioRtia, line 49: Same remark

PR #1304, opened 2020-05-03
IEEE: fix generator active power limits default value

geofjamg:

No

Bug fix

There is no generator active power limits in IEEE CDF format. Min/max value is set to +-Double.MAX_VALUE. The problem is that in case of loadflow and distributed slack, the computed factor are totally wrong (because we at some point, we sum all participation factors that are closed to double max value...).

We set min/max active power to -+9999 like in UCTE importer.

**Other information**:


ieee-cdf/ieee-cdf-converter/src/test/resources/ieee9cdf.xiidm
```diff
@@ -5,7 +5,7 @@
             <iidm:busBreakerTopology>
                 <iidm:bus id="B1" name="BUS-1   100" v="1.04" angle="0.0"/>
             </iidm:busBreakerTopology>
-            <iidm:generator id="B1-G" energySource="OTHER" minP="-1.7976931348623157E308" maxP="1.7976931348623157E308" voltageRegulatorOn="true" targetP="71.64102" targetV="1.04" targetQ="27.04592" bus="B1" connectableBus="B1">
+            <iidm:generator id="B1-G" energySource="OTHER" minP="-9999.0" maxP="9999.0" voltageRegulatorOn="true" targetP="71.64102" targetV="1.04" targetQ="27.04592" bus="B1" connectableBus="B1">
                 <iidm:minMaxReactiveLimits minQ="-99990.0" maxQ="999900.0"/>
```
annetill, line 9: These values of minQ and maxQ are really strange. It seems to be default values. What have we planned to do with default values management in IIDM ? Do we need a global strategy ?

geofjamg: This is min max double value. This is probably a bad design and we should reactive limit should be optional in the IIDM API to better handle to no reactive limit case.
Anyway it is not really the same topic as this PR, as reactive limits in a loadflow are not used to dispatch everything out of the limit like it is the case with active power.

PR #1311, opened 2020-05-12
Move the component management in IIDM API

mathbagu:

N/A

Refactoring

The component management (connected and synchronous) is in the implementation and cannot be shared between implementations

The implementation has been moved to the API

**Other information**:

### DIFF
--- a/iidm/iidm-api/pom.xml
+++ b/iidm/iidm-api/pom.xml
@@ -64,6 +64,11 @@
             <artifactId>powsybl-commons</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>powsybl-math</artifactId>
+            <version>${project.version}</version>
+        </dependency>
         
         <!-- Test dependencies -->
         <dependency>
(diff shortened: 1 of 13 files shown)


mathbagu:
You right, it's probably not a good idea to expose the setters from the API. I'm not really convinced by what your propose... creating the BusExt interface and force implementation to inherit from it to be usable with our class is equivalent to keep the implementation in the implementation...

What do you think if the component manager is responsible to maintain the mapping between the buses and the component number. Instead of setting the number on each bus, it could fill a map?

geofjamg:
> You right, it's probably not a good idea to expose the setters from the API. I'm not really convinced by what your propose... creating the BusExt interface and force implementation to inherit from it to be usable with our class is equivalent to keep the implementation in the implementation...
> 
> What do you think if the component manager is responsible to maintain the mapping between the buses and the component number. Instead of setting the number on each bus, it could fill a map?

The mistake is this PR is to suppose that the components impl is so generic that it could be part of the API. For instance we don't use it in the network-store and we probably won't ever use it. The persistent impl of IIDM is really very different from original one and using an "external" component calculation algo is not relevant. It has to be more specific.

So whatever the chosen design it has nothing to do in iidm-api. This is not because you want to reuse it in merging view that it should be part of the API. It is fundamentally part of "some" implementation not necessarily all so it is a utility . Wherever we put this code, I don't care (iidm-util or elsewhere) but not in the API.

Regarding the "external" storage of component to bus mapping with a map, why not but the risk is to degrade performance (direct attribute access => map lookup). But maybe it would not be noticeable. We have to test it to be sure before taking decision.

PR #1313, opened 2020-05-13
Remove skip-postproc from LoadFlow and DynamicSimulation tools

mathbagu:

N/A

Bug

There is skip-postproc option to skip the Import postprocessors. But when this option is set, the ImportConfig is created without loading the configuration (default constructor).

The option has been removed: the ImportConfig is always loaded from the configuration file.

**Other information**:


sylvlecl:
Ok for this change but I don't fully understand what problem it will solve:
is it in case we add new properties inside the `ImportConfig` in the future?
For now there is only those post-processors, so it would seem that it behaves as expected. It's not the case ?

mathbagu:
> Ok for this change but I don't fully understand what problem it will solve:
> is it in case we add new properties inside the `ImportConfig` in the future?
> For now there is only those post-processors, so it would seem that it behaves as expected. It's not the case ?

My proposal is to remove the `skip-postproc` option from all the simulation commands of iTools. If we keep it, we should add this opiton for all the iTools command that import a network. In the PR #1265 we introduce post processors for the export.

You right, for now the ImportConfig contains only the list of the activated post-processor. But we could imagine to add options in the future, options that are global to all the importers. In that case, the code we have currently is not ready for that. It should load the default config and clear the list of the post-processors instead.

Anyway, if you agree with the proposal, we can merge it. If you disagree and we decide to keep the `skip-postproc`, we should add it for all the iTools command

@sylvlecl what do you decide?






The main idea is to respect 1 feature per tool. Today, there is an inconsistency in the

sylvlecl:
Ok it's more clear to me, thank you!
Indeed it's not consistent today, and may cause bugs in the future the way it's implemented.

Issue #1315, opened 2020-05-18
The Bus definition in BusView is not consistent


mathbagu:
* **Do you want to request a *feature* or report a *bug*?**
Bug/Feature

* **What is the current behavior?**
The definition of a bus in the BusView is not consistent in both VoltageLevel implementations. In a Node/Breaker VL, there is a bus if :
- there is at least one busbar sections and at least one feeder
- there is at least one branch and at least 2 feeders
In the Bus/Breaker VL implementation, there is a bus if there is at least one branch.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

In CGMES, there are some use-cases, where there is no busbar section in a Node/Breaker VL, but a branch is connected. In that case, there is no bus in the BusView, whereas there would be one if the voltage level was described in a Bus/Breaker topology.

Another issue with the current definition of a Bus concerns the DanglingLine. Currently they are considered as feeders, but not as branches. So in this kind of configuration, there is no bus in Bus/Breaker voltage level:
```
[LOAD] --- (BUS) --- DL --- [Xnode]
```

So, in the merging view, when we merge this VL with a similar one, we don't have a bus for the MergedLine.

* **What is the expected behavior?**
The proposition is:
- DanglingLine should be consider as a branch like the Lines or TwoWindingsTransformer
- The Bus definition of the NBVL::BusView should be changed to:
    - at least one busbar section and at least one feeder (unchanged)
    - at least one branch

* **What is the motivation / use case for changing the behavior?**
To have the same number of buses in both topology and to have the same number of buses in the IGM and the merging view

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

geofjamg:
I am not sure it is a safe modification to have a bus with just a busbar/configured bus and a branch connected. The bus view is made for giving a topology nice for a power flow calculation and I suspect having a bus with just a branch would break loadflow compatibility. We should probably make a test for confirm or not.

Why not doing the inverse modification, so in a bus/.breaker topo if just one configured bus and a branch, not bus in the bus/view?

PR #1316, opened 2020-05-18
Fix inconsistency in the bus definition of the bus view

mathbagu:

#1315 

Bug fix

The definition of a bus is not equivalent in the bus view of the two type of voltage level topology.

With the proposal fix, there are more buses, but the number of buses is equivalent in the two implementations

**Other information**:


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevel.java
```diff
@@ -291,9 +291,17 @@ protected void updateCache(final Predicate<SwitchImpl> terminate) {
             CalculatedBus[] node2bus = new CalculatedBus[graph.getVertexCapacity()];
             boolean[] encountered = new boolean[graph.getVertexCapacity()];
             Arrays.fill(encountered, false);
-            for (int e : graph.getEdges()) {
-                traverse(graph.getEdgeVertex1(e), encountered, terminate, id2bus, node2bus);
-                traverse(graph.getEdgeVertex2(e), encountered, terminate, id2bus, node2bus);
+            if (graph.getEdgeCount() == 0) {
```
annetill, line 294: @mathbagu can you please clean this function for me ? I am not sure I am able to. Thanks !

contingency/contingency-api/src/test/java/com/powsybl/contingency/tasks/HvdcLineTrippingTest.java
```diff
@@ -65,7 +65,7 @@ public void lineTrippingTest() {
         contingency.toTask().modify(network, null);
 
         assertTrue(terminal1.isConnected());
-        assertFalse(terminal2.isConnected());
+        assertTrue(terminal2.isConnected());
```
mathbagu, line 68: I think this is a regression, so I have to check it: if you trigger an AC line, both side are disconnected. I used the same approach for HVDC: I disconnected the 2 converter stations, so both terminal should be disconnected.

annetill: Thanks @mathbagu Indeed you are right we have an issue here because the HVDC converter station is considered as a feeder and as a branch.

mathbagu:
I distinguish 5 different types of equiments: busbarSection, branch (Line, 2WT, 3WT), injections (all except the DL and the converterStations), ConverterStation and DanglingLine

In that case, there is no bus:
- [ ] 1 Injection
- [ ] 1+ SJB and 1 injection
- [ ] 1+ SJB and 1 branch
From my point of view, the busbar section should not be considered

In that case, there is a bus:
- [ ] 2 injections (all)
- [ ] 2 branches
- [ ] 1+ injection (all) and 1+ branch
**There is one bus if there are at least 2 feeders**

In the following case, it's arguable:
- [ ] 1 danglingLine : could be a bus if the danglingLine has been merged
- [ ] 1 AC/DC converter station: could be a bus if the converter is connected to a HVDC line
- [ ] 1 branch

mathbagu:
test1

test2-8

geofjamg:
Where does the expected diagram come from? You expect the load to be always connected to a bus even of not connected but it should not be the case. If there is nothing to feed the load, there is no voltage so there is no bus.

A bus means voltage, if no voltage you cannot have any bus.

I think the misunderstanding come from the bus definition in the bus/breaker model which is slightly different. It a mix between a busbar and a bus. Not really a busbar, not necessary a bus. This is very confusing.

But in the bus view, a bus is point in the diagram with a unique voltage. This is a very physical definition.

PR #1319, opened 2020-05-22
Addition of an alias system on Identifiables in IIDM

murgeyseb:
> This PR is a **proposal** and a **Work In Progress**

No

Feature

Currently, there is only one way to reference an Identifiable: its ID.
However, in some context it may be interesting to attach "secondary" IDs to the network devices and efficiently retrieve the device based on this new ID using the network API.

An example of such a need may be the usage of UCTE files. There are actually two ways to refer to network devices in UCTE files:
- node from / node to / order code triplet (the one chosen by PowSyBl)
- node from / node to / element name (used by some countries)

This PR proposes the addition of an alias mechanism on all IIDM Identifiable. The alias can be seen as secondary IDs and has similar rules:
- Aliases must be unique in the network
- Aliases must be different from any ID of the network
- Network API has a method that allow searching an Identifiable based on an alias

Open questions on my side:
- What if the triplet node from / node to / element name is not unique in the UCTE file ? Do we catch the exception ? Do we allow multiple identifiables referenced by an alias ?
- Would it better to replace the Exception thrown in default implementation by "do nothing" implementation, to avoid UCTE import crash using incomplete IIDM implementation ?

No

**Other information**:


geofjamg:
@murgeyseb do you really need to be able de change aliases after object creation? Why not an immutable approach like with the primary id where alias would be specified in the adder at object creation? That would simplify the design and would allow more easily to adapt this feature to IIDM network store implementation.

murgeyseb:
@geofjamg I tried this approach but the implementation was a bit more intrusive in other part of the code (it forces to modify all the Adders API and implementations for example). This was the first reason why I chose this approach in the first time. Moreover, I think it may also be interesting in some applications to be able to add aliases (remove it is a bit less interesting though).

However we can also try another immutable implementation of this feature.

geofjamg:
At least, if no clear identified use case, we should not support alias removal.
Do you remember what was intruisive in adder impl? To you have a branch with the try?

In addition to this PR, another one has to be done on network store side before merging this one to ensure that it is possible to adapt it with a persistent impl (which is really more complex especially when it comes to data indexing).

murgeyseb:
If you want to make it immutable, best is to add a method `addAlias(String)` to IdentifiableAdder. Then you have to modify all implementations of IdentifiableAdder::add() to modify the Constructor of the implemented object at least.
It may not be difficult, but I thought that aliases may be seen as Properties (though mutable), excepted that it would be used efficiently when "traversing" the Network object.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Network.java
```diff
@@ -661,12 +661,22 @@ default HvdcLine getHvdcLine(HvdcConverterStation converterStation) {
...
+    /**
+     * Get an equipment by its ID or eventually an alias.
+     *
+     * @param idOrAlias the id or an alias of the equipment
+     * @param searchInAlias if true then searches in the alias and the IDs. If false, searches only by ID.
+     */
+    default Identifiable<?> getIdentifiable(String idOrAlias, boolean searchInAlias) {
+        throw new UnsupportedOperationException();
+    }
+
```
mathbagu, line 679: Do we really need this method: what is the risk to change the behavior of the `getIdentifiable(String id)` if we consider aliases like unique IDs. At the end, all the user code would use this new method, don't they?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
```diff
@@ -1101,13 +1111,15 @@ private void replaceDanglingLineByLine(List<MergedLine> lines, Multimap<Boundary
             l.getTerminal1().setP(mergedLine.p1).setQ(mergedLine.q1);
             l.getTerminal2().setP(mergedLine.p2).setQ(mergedLine.q2);
             mergedLine.properties.forEach((key, val) -> l.setProperty(key.toString(), val.toString()));
+            mergedLine.aliases.forEach(l::addAlias);
```
mathbagu, line 1114: From my point of view, we should automatically add the ids of the half line when a TieLine is created, not only in the merge function. Is there a risk to do it automatically in the add() method of the TieLineAdder?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkIndex.java
```diff
@@ -50,14 +53,53 @@ void checkAndAdd(Identifiable<?> obj) {
...
+        } else if (idByAlias.containsKey(alias)) {
+            Identifiable<?> aliasConflict = objectsById.get(idByAlias.get(alias));
+            String message = String.format("Object (%s) with alias '%s' cannot be created because alias already refers to object (%s) with ID '%s'",
+                    obj.getClass(),
+                    alias,
+                    aliasConflict.getClass(),
+                    aliasConflict.getId());
+            throw new PowsyblException(message);
+        }
+        idByAlias.put(alias, obj.getId());
```
mathbagu, line 74: You should test if the object that already exists in the map is not the same than the `obj` argument, and silently ignore it. You could be possible to write something like this:
```
addAlias(obj, "alias"); // works because "alias" is not used
addAlias(obj, "alias"); // works because "alias" is already an alias of obj
addAlias(obj2, "alias"); // fails because "alias" is already an alias of obj, and obj != obj2
```

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AliasesXml.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * @author Sebastien Murgey <sebastien.murgey at rte-france.com>
+ */
+public final class AliasesXml {
+
+    static final String ALIAS = "alias";
+
+    public static void write(Identifiable<?> identifiable, NetworkXmlWriterContext context) throws XMLStreamException {
+        if (identifiable.hasAliases()) {
```
mathbagu, line 21: Remove this test: a for loop on an empty list is OK

iidm/iidm-xml-converter/src/test/java/com/powsybl/iidm/xml/TieLineWithAliasesXmlTest.java
```diff
@@ -0,0 +1,26 @@
...
+
+package com.powsybl.iidm.xml;
+
+import org.junit.Test;
+
+import java.io.IOException;
+
+/**
+ * @author Mathieu Bague <mathieu.bague@rte-france.com>
+ */
```
mathbagu, line 16: Impostor! :rofl:

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
```diff
@@ -258,6 +258,9 @@ private static void createDanglingLine(UcteLine ucteLine, boolean connected,
                 .setQ0(q0)
                 .setUcteXnodeCode(xnode.getCode().toString())
                 .add();
+        if (ucteLine.getElementName() != null && !ucteLine.getElementName().isEmpty()) {
+            dl.addAlias(String.format("%s %s %s", ucteLine.getId().getNodeCode1().toString(), ucteLine.getId().getNodeCode2().toString(), ucteLine.getElementName()));
```
mathbagu, line 262: Is really what you want? Maybe the elementName should be an alias, maybe "VL1 VL2 order" should be another one.
Maybe it could be nice to create a function for that

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Network.java
murgeyseb: I do not know if it may be interesting to search objects by ID uniquely.

geofjamg: As searching also with alias could impact performance, it might be interesting to keep the possibility to only seach with primary id

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
murgeyseb: Actually, the TieLineAdder does not have access to the aliases of the half lines, only the IDs. So a dedicated line for merging aliases is still needed.

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
murgeyseb: As "node1/node1/order" is already the id, no need of it as an alias.
The element name cannot be an alias itself, cause TSOs usually use it as a kind of order code (you can have same element name for different lines)

mathbagu: I know, it is the reason why I suggest "VL1 VL2 co" to complement the "node1 node2 co". If I remember, that was a way to reference equipments in UCTE process. Maybe it's not the case anymore.

murgeyseb: Good remark, it's true that sometimes the node code is replaced by * in the input.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Network.java
```diff
@@ -261,7 +261,7 @@ static PrettyNetworkFactory with(String name) {
     /**
      * Get a substation voltage level.
      *
-     * @param id the id of the substation voltage level
+     * @param id the id or ana alias of the substation voltage level
```
mathbagu, line 264: There is a small typo here
> or an**a** alias

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiable.java
```diff
@@ -24,6 +24,8 @@
 
     protected boolean fictitious = false;
 
+    protected Set<String> aliases = new TreeSet<>();
```
mathbagu, line 27: Is there a reason why you choose a TreeSet instead of a HashSet. Does the order important for you? Maybe for the XML round trip?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
```diff
@@ -966,6 +966,11 @@ private void mergeDanglingLines(List<MergedLine> lines, DanglingLine dl1, Dangli
         if (dl1 != null) {
             MergedLine l = new MergedLine();
             l.id = dl1.getId().compareTo(dl2.getId()) < 0 ? dl1.getId() + " + " + dl2.getId() : dl2.getId() + " + " + dl1.getId();
+            l.aliases = new HashSet<>();
+            l.aliases.add(dl1.getId());
+            l.aliases.add(dl2.getId());
+            l.aliases.addAll(dl1.getAliases());
+            l.aliases.addAll(dl2.getAliases());
```
mathbagu, line 973: Should we change the mergeability check to consider also aliases?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkIndex.java
```diff
@@ -50,14 +53,54 @@ void checkAndAdd(Identifiable<?> obj) {
...
     }
 
-    Identifiable get(String id) {
+    boolean addAlias(Identifiable<?> obj, String alias) {
+        if (objectsById.containsKey(alias)) {
+            if (objectsById.get(alias).equals(obj)) {
+                // Silently ignore affecting the objects id to its own aliases
+                return false;
+            }
+            Identifiable<?> aliasConflict = objectsById.get(alias);
```
mathbagu, line 62: It's not efficient: you search for the object 3 times. If the map is huge, it could lead to bad performance issue.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkIndex.java
```diff
@@ -50,14 +53,54 @@ void checkAndAdd(Identifiable<?> obj) {
...
+                    alias,
+                    aliasConflict.getClass(),
+                    aliasConflict.getId());
+            throw new PowsyblException(message);
+        } else if (idByAlias.containsKey(alias)) {
+            if (objectsById.get(idByAlias.get(alias)).equals(obj)) {
+                // Silently ignore affecting the same alias twice to an object
+                return false;
+            }
+            Identifiable<?> aliasConflict = objectsById.get(idByAlias.get(alias));
```
mathbagu, line 74: Same kind of remark there

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkIndex.java
```diff
@@ -50,14 +53,54 @@ void checkAndAdd(Identifiable<?> obj) {
...
+            throw new PowsyblException(message);
+        }
+        idByAlias.put(alias, obj.getId());
+        return true;
+    }
+
+    public <I extends Identifiable<I>> void removeAlias(Identifiable<?> obj, String alias) {
+        if (!idByAlias.containsKey(alias)) {
+            throw new PowsyblException(String.format("No alias '%s' found in the network", alias));
+        } else if (!idByAlias.get(alias).equals(obj.getId())) {
```
mathbagu, line 89: Performance: you are searching for the same object multiple times.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkIndex.java
```diff
@@ -50,14 +53,54 @@ void checkAndAdd(Identifiable<?> obj) {
...
+            throw new PowsyblException(String.format("Alias '%s' do not correspond to object '%s'", alias, obj.getId()));
+        } else {
+            idByAlias.remove(alias);
+        }
+    }
+
+    Identifiable get(String idOrAlias) {
+        String id = idByAlias.getOrDefault(idOrAlias, idOrAlias);
```
mathbagu, line 97: That means it's not possible to use the ID of an object as one of its alias. I know it doesn't make any sense, but if it's allowed by the code, does it work properly?

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/AliasesTest.java
```diff
@@ -0,0 +1,92 @@
...
+        generator.addAlias("load1");
+    }
+
+    @Test(expected = PowsyblException.class)
+    public void failWhenRemovingNonExistingAlias() {
+        Network network = NetworkTest1Factory.create();
+        Load load = network.getLoad("load1");
+        load.removeAlias("Load alias");
+    }
+}
```
mathbagu, line 92: Could you please add a test where you try to add as an alias the ID of an identifiable, and ensure you can get the object.

iidm/iidm-xml-converter/src/test/java/com/powsybl/iidm/xml/TieLineWithAliasesXmlTest.java
```diff
@@ -0,0 +1,26 @@
...
+ * @author Sebastien Murgey <sebastien.murgey@rte-france.com>
+ */
+public class TieLineWithAliasesXmlTest extends AbstractXmlConverterTest {
+
+    @Test
+    public void test() throws IOException {
+        roundTripVersionedXmlTest("tielineWithAliases.xml", IidmXmlConstants.CURRENT_IIDM_XML_VERSION);
+
+        // Tests for backward compatibility
+        roundTripVersionedXmlFromMinToCurrentVersionTest("tielineWithAliases.xml", IidmXmlVersion.V_1_3);
```
mathbagu, line 24: Both are redundant?

ucte/ucte-util/pom.xml
```diff
@@ -0,0 +1,52 @@
...
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+</project>
```
mathbagu, line 52: Add an empty line

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiable.java
murgeyseb: I do not remember if it was for ensuring that roundtrip test work... Usually, I would sa that order is not important but it may be nice to keep aliases sorted, as you said for consistency in XML I/O. However, it may not impact a lot the performances as there is no reason, in current use case, that an Identifiable would have thousands of aliases...

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
murgeyseb: Should we enforce consistency (no duplicate aliases) or remove them in case of conflict during merging ?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkIndex.java
murgeyseb: In current implementation, ID is not in the alias list, but as it is the ID, it works to retrieve the object using the ID

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/AliasesTest.java
murgeyseb: The ID of the identifiable you are adding an alias to, or of another identifiable ?

iidm/iidm-xml-converter/src/test/java/com/powsybl/iidm/xml/TieLineWithAliasesXmlTest.java
murgeyseb: Not in next version of IIDM

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
mathbagu: What is the current behavior if you try to add an existing alias to another identifiable? If I remember well, an exception is thrown. So I think we should consider the network are not mergeable, specially if we consider alias as a kind of identifier.

iidm/iidm-xml-converter/src/test/java/com/powsybl/iidm/xml/TieLineWithAliasesXmlTest.java
mathbagu: You right, the current_version is ignored. I ask to @MioRtia if the expected behavior

MioRtia:
I have think of something with @mathbagu: would it be possible to adda type to aliases? Indeed, in CGMES for example, it would be useful to add as aliases the ID of sub-elements (such as current limits, tap changers, etc) but in the current state, there is no way to have the information of where the aliases come from. Maybe as an optional attribute, the type can be a string which by default can be null?

murgeyseb:
> I have think of something with @mathbagu: would it be possible to adda type to aliases? Indeed, in CGMES for example, it would be useful to add as aliases the ID of sub-elements (such as current limits, tap changers, etc) but in the current state, there is no way to have the information of where the aliases come from. Maybe as an optional attribute, the type can be a string which by default can be null?
> 
> **EDIT** It is just an idea, not a blocking issue

It could be an interesting addition, but would you want to provide e research criteria (for example search t2wt with given alias of phase tap changer ?) What about aliases unicity ? unique pair type/alias.
I think it is definitely something to keep for later.

PR #1321, opened 2020-05-26
New extension "VoltagePerReactivePowerControl" for StaticVarCompensator

annetill:

Yes, here: https://github.com/powsybl/powsybl-open-loadflow/issues/97

The possibility to store a slope for StaticVarCompensator that allow them to have remote regulating capabilities in reactive power.

No slope are store, StaticVarCompensator regulate voltage like generators.

The slope could be used by loadflow engine.

**Other information**:

This PR has to be finished :-)


iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/VoltagePerReactivePowerControl.java
```diff
@@ -0,0 +1,25 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.iidm.network.extensions;
+
+import com.powsybl.commons.extensions.Extension;
+import com.powsybl.iidm.network.StaticVarCompensator;
+
+/**
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
```
mathbagu, line 13: Fix the author (same remark for all the files)

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/VoltagePerReactivePowerControlAdderImpl.java
```diff
@@ -0,0 +1,35 @@
...
+    private double slope;
+
+    protected VoltagePerReactivePowerControlAdderImpl(StaticVarCompensator svc) {
+        super(svc);
+    }
+
+    @Override
+    protected VoltagePerReactivePowerControlImpl createExtension(StaticVarCompensator svc) {
+        return new VoltagePerReactivePowerControlImpl(svc, slope);
+    }
```
mathbagu, line 28: There is something strange in the design: it's not related to your PR, but why the adder take one `Extendable` instance and the createExtension could take another one?
@MioRtia : what do you think?

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/VoltagePerReactivePowerControlImpl.java
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ */
+public class VoltagePerReactivePowerControlImpl extends AbstractExtension<StaticVarCompensator> implements VoltagePerReactivePowerControl {
+
+    private double slope;
+
+    public VoltagePerReactivePowerControlImpl(StaticVarCompensator svc, double slope) {
+        super(svc);
+        this.slope = slope;
```
mathbagu, line 21: Is there any restriction for this value? Do we have to implement a `checkSlope` method to ensure this value is in range, and use it in the constructor and the setter?

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/VoltagePerReactivePowerControl.java
annetill: Indeed it is from @geofjamg I just have moved some code.

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/VoltagePerReactivePowerControlImpl.java
annetill: Yes Indeed it should be positive in the load sign convention.

iidm/iidm-extensions/src/test/java/com/powsybl/iidm/network/extensions/VoltagePerReactivePowerControlXmlSerializerTest.java
```diff
@@ -0,0 +1,55 @@
...
+    @Test
+    public void test() throws IOException {
+        Network network = SvcTestCaseFactory.create();
+        network.setCaseDate(DateTime.parse("2019-05-27T12:17:02.504+02:00"));
+        StaticVarCompensator svc = network.getStaticVarCompensator("SVC2");
+        assertNotNull(svc);
+
+        svc.newExtension(VoltagePerReactivePowerControlAdder.class).withSlope(0.5).add();
+
+        // FIXME
```
annetill, line 40: @mathbagu Help, it is not working and I think it is just because I am not familiar with xml conversion. Thanks !

annetill: The error I get is:
````com.powsybl.commons.exceptions.UncheckedSaxException: org.xml.sax.SAXParseException; lineNumber: 24; columnNumber: 60; cvc-complex-type.2.4.c : Le caractère générique concordant est strict, mais aucune déclaration ne peut être trouvée pour l'élément 'vprpc:voltagePerReactivePowerControl'.````

Comment (zamarrenolm) [reply iidm/iidm-extensions/src/test/java/com/powsybl/iidm/network/extensions/VoltagePerReactivePowerControlXmlSerializerTest.java]: The problem was that namespace was different in xsd and in the XML serializer class.
I removed the FIXME annotation. 
I also fixed the expected value in the final assert (should be 0.5, that is the value used when creating the extension).

Comment (annetill) [reply iidm/iidm-extensions/src/test/java/com/powsybl/iidm/network/extensions/VoltagePerReactivePowerControlXmlSerializerTest.java]: Thanks a lot @zamarrenolm ! Have you check the cgmes conversion by the way ?

Comment (zamarrenolm) [reply iidm/iidm-extensions/src/test/java/com/powsybl/iidm/network/extensions/VoltagePerReactivePowerControlXmlSerializerTest.java]: Yes, I saw that you are storing in the extension the `slope` attribute read in the `staticVarCompensators` SPARQL query, so for me it is ok.

File: iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/VoltagePerReactivePowerControlImpl.java [hunk 5]
```diff
@@ -0,0 +1,49 @@
...
+    public VoltagePerReactivePowerControl setSlope(double slope) {
+        this.slope = slope;
+        return this;
+    }
+
+    private static double checkSlope(StaticVarCompensator svc, double slope) {
+        if (Double.isNaN(slope)) {
+            throw new PowsyblException("Undefined value for slope");
+        }
+        if (slope < 0) {
```
mathbagu, line 44: There is a diffference in this check and in the check done in `StaticVarCompensatorConversion.java`
```
        if (slope > 0) {
            svc.newExtension(VoltagePerReactivePowerControlAdder.class).withSlope(slope);
        }
```
Does `0` a valid value?

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/VoltagePerReactivePowerControlImpl.java
```diff
@@ -0,0 +1,49 @@
...
+            LOGGER.debug("Slope value of svc {} should be positive: {}", svc.getId(), slope);
```
mathbagu, line 45: Is it OK or should we throw an exception? If we accept negative value, it will never be possible to change that. It's better to throw an exception, and then make the code more permissive. 
Moreover, in CGMES conversion, we do not create the extension if the slope is negative (or equal to 0)

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/VoltagePerReactivePowerControlAdderImpl.java
MioRtia: Hm, good question. In the code, `createExtension` is only called with the `extendable` field (from the constructor) as parameter. Maybe the method should not take any parameter and `extendable` should be protected? (@jonenst do you remember why you designed it this way?)

jonenst: I did it this way because it was more explicit for an implementer of an adder to take the extendable as a method argument from the method he is supposed to implement. The alternative (that you mentionned) is to have the extendable in a protected field of the base class and document that it should be used in the createExtension method

jonenst: I'm ok with both implementations if you want to change it

jonenst: I guess it's not so implicit with a protected field because you have to write the constructor calling super() with the extendable so you are hinted about it

MioRtia: Ok, thank you for the explanation! Both are okay for me, what do you think @mathbagu ? In any case, I think this PR is good to be merged, we can make another PR on this subject later.

mathbagu: The only thing I want we ensure, is that should be forbidden to pass a different instance. I think it could be the case with the abstract class we provide. It's not directly related to this PR, so I will merge it but the discussion is still open

PR #1322, opened 2020-05-26
Redesign IIDM shunt compensator allowing non linear shunts + updated conversions + updated serialization

MioRtia:

No

New feature: integration of non linear shunts + redesign of shunts

All shunts are linear.

There can be linear and non linear shunts, specified by their model.

Yes


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorModel.java
```diff
@@ -0,0 +1,28 @@
...
+ * @author Miora Ralambotiana <miora.ralambotiana at rte-france.com>
+ */
+public interface ShuntCompensatorModel {
+
+    /**
+     * Get the section susceptance in S of the section with a given section number if it exists.
+     * Return 0 if sectionIndex is equal to 0 (disconnected state).
+     * Throw an exception if such a section does not exist.
+     */
+    double getBSection(int sectionIndex);
```
annetill, line 19: This function will be shared by ````ShuntCompensatorLinearModel```` and ````ShuntCompensatorNonLinearModel````, right ? Why cannot we have simply getbPerSection(int sectionIndex) ?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorLinearModel.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * @author Miora Ralambotiana <miora.ralambotiana at rte-france.com>
+ */
+public interface ShuntCompensatorLinearModel extends ShuntCompensatorModel {
+
+    /**
+     * Get the susceptance per section in S.
+     */
+    double getbPerSection();
```
annetill, line 17: ````getbPerSection```` or ````getBPerSection```` ?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorNonLinearModel.java
```diff
@@ -0,0 +1,54 @@
...
+     * If a section already exists for this section number, respectively replace its susceptance and conductance with the given susceptance and conductance.
+     * Throw an exception if the section index equals to 0 (corresponds to disconnected state).
+     */
+    ShuntCompensatorNonLinearModel addOrReplaceSection(int sectionIndex, double b, double g);
+
+    /**
+     * Remove the section associated with a given section number if it exists <b>and</b> the current section count is different of the given section number.
+     * Else, throw an exception.
+     */
+    ShuntCompensatorNonLinearModel removeLastSection();
```
annetill, line 53: Why do you need this function ?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -228,26 +228,42 @@ public static void checkForecastDistance(Validable validable, int forecastDistan
...
+    }
+
+    public static void checkbPerSection(Validable validable, double bPerSection) {
+        checkSectionB(validable, bPerSection);
         if (bPerSection == 0) {
             throw new ValidationException(validable, "susceptance per section is equal to zero");
         }
     }
 
+    public static void checkSectionB(Validable validable, double sectionB) {
```
annetill, line 246: The name chosen is not so clear. See remark above.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -228,26 +228,42 @@ public static void checkForecastDistance(Validable validable, int forecastDistan
...
-            throw new ValidationException(validable, "susceptance per section is invalid");
+    public static void checkSectionNumber(Validable validable, int sectionNum) {
+        if (sectionNum < 0) {
+            throw new ValidationException(validable,
+                    "this number of section (" + sectionNum
+                            + ") should be greater than or equal to 0");
         }
+    }
+
+    public static void checkbPerSection(Validable validable, double bPerSection) {
```
annetill, line 239: I fall in this exception on some TSOs, included for our French network !

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
```diff
@@ -46,16 +45,120 @@ protected String getTypeDescription() {
...
+            @Override
+            public SectionAdder setG(double g) {
+                this.g = g;
+                return this;
+            }
+
+            @Override
+            public ShuntCompensatorNonLinearModelAdder endSection() {
+                if (sectionIndex == 0) {
+                    throw new ValidationException(ShuntCompensatorAdderImpl.this, "section 0 (b = 0.0, g = 0.0) should not be written, it corresponds to disconnected state");
```
annetill, line 117: I fall in this exception with NGESO network.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorModel.java
annetill: Do you have any idea ?

MioRtia:
@annetill Do we want `simulShunt` set as `false` or as `ŧrue` in default configuration? I set it as `false` but I am having second thoughts.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorModel.java
MioRtia: Yes, I agree with your proposition. It will probably be clearer.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorLinearModel.java
MioRtia: For consistence, we might as well rename as `getBPerSection`

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorNonLinearModel.java
MioRtia: Otherwise, it would be impossible to remove shunt section in non linear shunt (for development purposes for example)

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
MioRtia: Corrected in CGMES conversion

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
MioRtia: I will check but I think I will add a fix on CGMES converter here.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorNonLinearModel.java
annetill: I am not sure to get what you mean :-) The function is not to remove a given section, just the last one.

MioRtia: When you have a non linear shunt, what you can want to do is:
- adding a new section
- replacing the value of an existing section
- removing an existing section
However, when you remove an existing section, it cannot be one in the middle (you cannot have a "hole" in your section list), so it is necessarily the last section. It occurs to me that a method can however be added to remove the last 2, 3, etc sections of a non linear shunt and not only the last one... (for now, to do that we have to call removeLastSection 2, 3, etc times)

MioRtia: Side note, it is the same for adding a new section, you cannot add a section with a index strictly superior to the last index + 1 to avoid "holes"

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java
```diff
@@ -43,31 +46,48 @@ public void convert() {
...
-        if (bPerSection == 0) {
-            float bPerSectionFixed = Float.MIN_VALUE;
-            fixed(CgmesNames.B_PER_SECTION, "Can not be zero", bPerSection, bPerSectionFixed);
-            bPerSection = bPerSectionFixed;
+            for (PropertyBag sp : ss) {
+                int sectionNumber = sp.asInt(SECTION_NUMBER);
+                double b = ss.stream().filter(s -> s.asInt(SECTION_NUMBER) <= sectionNumber).map(s -> s.asDouble("b")).reduce(0.0, Double::sum);
+                if (sectionNumber == 0) {
+                    if (b != 0) {
+                        fixed("Shunt.section0", "Susceptance of section 0 should always be zero (disconnected state)", b, 0.);
```
annetill, line 72: Is it in the CGMES documentation that section 0 should have a b equal to zero ?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java
```diff
@@ -43,31 +46,48 @@ public void convert() {
...
+                    }
+                    continue;
+                }
+                if (b == 0) {
+                    double bFixed = Double.MIN_VALUE;
```
annetill, line 77: It is wierd that CGMES requires a b equal to zero and that we don't allow it in IIDM. No?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java
MioRtia: It doesn't but I don't know if it makes sense? 
Actually I think it is linked to your other remark: to be "realistic", it doesn't really matter if `b` is 0 at section 0 as long as there is a section where `b` is 0 (a section where the shunt can be considered as disconnected). Maybe we should go in that sense? However in CGMES datasets, even in explicit section 0, the susceptance is always 0 from what I've seen.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java
MioRtia: This check might actually not be useful for non linear shunt (see above). Maybe there are shunt that can be considered as disconnected in several positions.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java
annetill: From Implementation guide:
````Use the cim:ShuntCompensator.sections to control the impedance where cim:ShuntCompensator.sections=0 means zero admittance. For linear shunt this means multiplying with the per section admittance results in zero impedance. For nonlinear shunts the cim:NonlinearShuntCompensatorPoint.sectionNumber=0 shall have the corresponding admittance values equal zero.````

File: cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java (hunk 8 above)
Comment (annetill) [reply cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java:77]: The check has to be removed (see above).

File: iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorNonLinearModel.java (hunk 3 above)
Comment (annetill) [reply iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorNonLinearModel.java:53]: Thanks so Indeed this remove function is using by the function addOrReplace ? Why have you decided to put it in the interface and not just in the implementation?

File: iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java [hunk 9]
```diff
@@ -228,26 +228,42 @@ public static void checkForecastDistance(Validable validable, int forecastDistan
...
+
+    public static void checkLinearBPerSection(Validable validable, double bPerSection) {
+        checkBPerSection(validable, bPerSection);
         if (bPerSection == 0) {
             throw new ValidationException(validable, "susceptance per section is equal to zero");
         }
     }
 
+    public static void checkBPerSection(Validable validable, double sectionB) {
+        if (Double.isNaN(sectionB)) {
```
annetill, line 247: Yes great, so we can leave the check!

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowParameters.java
```diff
@@ -52,6 +52,7 @@
     public static final boolean DEFAULT_NO_GENERATOR_REACTIVE_LIMITS = false;
     public static final boolean DEFAULT_PHASE_SHIFTER_REGULATION_ON = false;
     public static final boolean DEFAULT_T2WT_SPLIT_SHUNT_ADMITTANCE = false;
+    public static final boolean DEFAULT_SIMUL_SHUNT = false;
```
annetill, line 55: Great!

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorNonLinearModel.java
MioRtia: No it is used when we really want to remove a section! addOrReplace can not remove a section, only replacing a value or adding a section.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
```diff
@@ -17,71 +17,79 @@
 public interface ShuntCompensator extends Injection<ShuntCompensator> {
 
     /**
-     * Get the maximum section count.
+     * Get the current count of sections in service.
+     * Please note sections can only be sequentially in service i.e. all sections from section 1 to section currentSectionCount are in service.
+     * <p>
+     * It is expected to be greater than one and lesser than or equal to the
```
geofjamg, line 23: Why? It should be possible to have 0 section activated?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
```diff
@@ -17,71 +17,79 @@
...
+     * maximum section count.
+     * <p>
+     * Depends on the working variant.
+     * @see VariantManager
+     */
+    int getCurrentSectionCount();
```
geofjamg, line 29: Maybe we could try to find a better naming than current. What I don't like with current is that in power system it is also a unit.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorModel.java
```diff
@@ -0,0 +1,28 @@
...
+package com.powsybl.iidm.network;
+
+/**
+ * @author Miora Ralambotiana <miora.ralambotiana at rte-france.com>
+ */
+public interface ShuntCompensatorModel {
+
+    /**
+     * Get the section susceptance in S of the section with a given section number if it exists.
+     * Return 0 if sectionIndex is equal to 0 (disconnected state).
```
geofjamg, line 16: If first section is section 0, we must have a non zero value as all section should have a b ou g?

ampl-converter/src/main/java/com/powsybl/ampl/converter/AmplNetworkReader.java
```diff
@@ -277,6 +278,9 @@ private Void readShunt(String[] tokens) {
             throw new AmplException("Invalid shunt compensator id '" + id + "'");
         }
 
+        if (ShuntCompensatorModelType.NON_LINEAR.equals(sc.getModelType())) {
```
geofjamg, line 281: You should use == for an enum

ampl-converter/src/main/java/com/powsybl/ampl/converter/AmplNetworkWriter.java
```diff
@@ -1310,6 +1311,9 @@ private void writeShunts(AmplExportContext context) throws IOException {
                      new Column("sections count"))) {
             List<String> skipped = new ArrayList<>();
             for (ShuntCompensator sc : network.getShuntCompensators()) {
+                if (ShuntCompensatorModelType.NON_LINEAR.equals(sc.getModelType())) {
```
geofjamg, line 1314: Same

cgmes/cgmes-conformity/src/test/java/com/powsybl/cgmes/conformity/test/CgmesConformity1NetworkCatalog.java
```diff
@@ -1028,6 +1034,42 @@ public static Network microType4BE() {
...
+        network.getVoltageLevel("_469df5f7-058f-4451-a998-57a48e8a56fe")
+                .newShuntCompensator()
+                    .setId("_002b0a40-3957-46db-b84a-30420083558f")
+                    .setName("BE_S2")
+                    .setConnectableBus(busBrussels380.getId())
+                    .setBus(busBrussels380.getId())
+                    .setCurrentSectionCount(1)
+                    .newNonLinearModel()
+                        .beginSection()
+                            .setSectionIndex(1)
```
geofjamg, line 1047: Is it really need to specify an index? It could be implicit (builder order)

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java
geofjamg: I don't understand the meaning of a zero admittance section. For me this should not be a requirement in IIDM but we need to accept it in order to keep the same number of section between CGMES and IIDM and then be able to update back the current section in case of regulation.
Why having a special meaning for section 0? section 0 is just the first one and could have an admittance. Also the shunt should accept 0 sections activated.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java
```diff
@@ -43,31 +46,43 @@ public void convert() {
...
+                if (sectionNumber == 0) {
+                    if (b != 0) {
+                        fixed("Shunt.section0", "Susceptance of section 0 should always be zero (disconnected state)", b, 0.);
+                    }
+                    continue;
+                }
+                modelAdder.beginSection()
+                        .setB(b)
+                        .setG(ss.stream().filter(s -> s.asInt(SECTION_NUMBER) <= sectionNumber).map(s -> s.asDouble("g")).reduce(0.0, Double::sum))
+                        .setSectionIndex(sp.asInt(SECTION_NUMBER))
```
geofjamg, line 79: Is it really necessary to keep the number, if we keep all of them in the right order?

cgmes/cgmes-conversion/src/test/resources/cim14/nordic32.xiidm
```diff
@@ -90,7 +90,7 @@
                 <iidm:bus id="_N1022____TN" v="110.60600280761719" angle="-17.621028900146484"/>
             </iidm:busBreakerTopology>
             <iidm:load id="_N1022____EC" name="N1022   " loadType="UNDEFINED" p0="280.0" q0="95.0" bus="_N1022____TN" connectableBus="_N1022____TN" p="280.0" q="95.0"/>
-            <iidm:shunt id="_N1022____SC" name="N1022   " bPerSection="0.002958579920232296" maximumSectionCount="1" currentSectionCount="1" voltageRegulatorOn="false" bus="_N1022____TN" connectableBus="_N1022____TN"/>
+            <iidm:linearShunt id="_N1022____SC" name="N1022   " bPerSection="0.002958579920232296" gPerSection="0.0" maximumSectionCount="1" currentSectionCount="1" voltageRegulatorOn="false" bus="_N1022____TN" connectableBus="_N1022____TN"/>
```
geofjamg, line 93: I would not change the top level equipment name. This is really confusing.
Why not:
```xml
<iidm:shunt ...>
    <iidm:shuntLinearModel ... />
```

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorNonLinearModel.java
```diff
@@ -0,0 +1,54 @@
...
+         */
+        double getG();
+    }
+
+    /**
+     * Get an optional of the section associated with a given section number if it exists.
+     * If such a section does not exist, return an empty optional.
+     *
+     */
+    Optional<Section> getSection(int sectionIndex);
```
geofjamg, line 35: I would model section in IIDM as a [0, maxSectionCount[ array and here avoid the optional.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorNonLinearModel.java
geofjamg: I don't understand why is a problem to remove a section that is not the last.
In any case if you remove a section it doesn't make sense anymore to update back the original model so we can allow everything.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorModelWrapper.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+package com.powsybl.iidm.network.impl;
+
+import com.powsybl.iidm.network.ShuntCompensatorModel;
+import com.powsybl.iidm.network.ShuntCompensatorModelType;
+
+/**
+ * @author Miora Ralambotiana <miora.ralambotiana at rte-france.com>
+ */
+interface ShuntCompensatorModelWrapper extends ShuntCompensatorModel {
```
geofjamg, line 15: It is not really a wrapper?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/LinearShuntXml.java
```diff
@@ -0,0 +1,99 @@
...
+ */
+public class LinearShuntXml extends AbstractConnectableXml<ShuntCompensator, ShuntCompensatorAdder, VoltageLevel> {
+
+    static final LinearShuntXml INSTANCE = new LinearShuntXml();
+
+    static final String ROOT_ELEMENT_NAME = "linearShunt";
+
+    @Override
+    protected String getRootElementName() {
+        return ROOT_ELEMENT_NAME;
```
geofjamg, line 29: As I said before, it seems to me very confusing to have a different root name when just changing the model. It is still a shunt compensator.

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowParameters.java
```diff
@@ -104,13 +105,20 @@ protected static void load(LoadFlowParameters parameters, PlatformConfig platfor
 
     private boolean t2wtSplitShuntAdmittance;
 
-    public LoadFlowParameters(VoltageInitMode voltageInitMode, boolean transformerVoltageControlOn,
-                              boolean noGeneratorReactiveLimits, boolean phaseShifterRegulationOn, boolean t2wtSplitShuntAdmittance) {
+    private boolean simulShunt;
```
geofjamg, line 108: Be careful, you forgot to take into account this new attribute in many places in this class.

loadflow/loadflow-validation/src/main/java/com/powsybl/loadflow/validation/ShuntCompensatorsValidation.java
```diff
@@ -85,11 +84,15 @@ public boolean checkShunts(ShuntCompensator shunt, ValidationConfig config, Vali
         Objects.requireNonNull(config);
         Objects.requireNonNull(shuntsWriter);
 
+        if (ShuntCompensatorModelType.NON_LINEAR.equals(shunt.getModelType())) {
```
geofjamg, line 87: == is more common for enum

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java
MioRtia: In what I implemented, section 0 is in fact the disconnected state (like linear shunts: when `sectionIndex == 0 `, b and g are also at 0 because they are proportional to `sectionIndex` ; this is also the convention adopted in CGMES). We can not do things like this but this would imply:
- that a load-flow can disconnect the shunt's terminal **or** we create a parameter to indicate if the shunt is in disconnected state or not (and in which case, we do not consider its b and g)
- having different definitions of what section 0 is for linear and non linear shunts (that might be leading to different behaviors after a load-flow)

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
MioRtia: By default for me when 0 section is activated, it means that there is zero susceptance and admittance (cf. above)

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
MioRtia: Maybe `getActivatedSectionCount()`?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorModel.java
MioRtia: cf answer above

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java
MioRtia: In my opinion, it makes it less efficient in cases when we are not sure in which order we receive the sections during the building (typically the case in CGMES). However, it makes the implementation less heavy (no need to use a map), what do you think?

cgmes/cgmes-conversion/src/test/resources/cim14/nordic32.xiidm
MioRtia: I started like this but it made the file bigger and less understandable? (particularly for linear shunts, it really make the XML object bigger and I didn't really think it was justified)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorModelWrapper.java
MioRtia: Yes, I'm not really sure how to name this interface...

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorNonLinearModel.java
MioRtia: Okay, just a question: when a section is removed, should all the cumulated susceptances and admittances of following sections be calculated again or does it just mean it will not be possible anymore to activate a certain section count to access a given cumulated susceptance and admittance?

For example, if we have:
```xml
<shunt id="SHUNT">
  <shuntNonLinearModel>
    <section b="1.0" g="1.0"/>
    <section b="2.0" g="2.0"/>
    <section b="4.0" g="5.0"/>
  </shuntNonLinearModel>
</shunt>
```
If the section section is removed, should we have:
```xml
<shunt id="SHUNT">
  <shuntNonLinearModel>
    <section b="1.0" g="1.0"/>
    <section b="4.0" g="5.0"/>
  </shuntNonLinearModel>
</shunt>
```
or:
```xml
<shunt id="SHUNT">
  <shuntNonLinearModel>
    <section b="1.0" g="1.0"/>
    <section b="3.0" g="4.0"/>
  </shuntNonLinearModel>
</shunt>
```
?

MioRtia: I guess my question should be: do you know the use case where a section is "removed" (for maintenance, etc) and what is the behavior of the shunt in this case?

cgmes/cgmes-conformity/src/test/java/com/powsybl/cgmes/conformity/test/CgmesConformity1NetworkCatalog.java
MioRtia: cf answer below

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
MioRtia: Should we also rename `getCurrentB()` and `getCurrentG()`? What do you propose for those?

annetill:
In the XIIDM file, we still have "currentSectionCount". As we don't keep "current" in the API, I imagine that you will remove it from the XIIDM too, no ?

MioRtia:
> In the XIIDM file, we still have "currentSectionCount". As we don't keep "current" in the API, I imagine that you will remove it from the XIIDM too, no ?

Yes, you're right, I forgot to change that as well.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
geofjamg: doc should be updated?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
```diff
@@ -17,71 +17,83 @@
...
+
+    /**
+     * Get the maximum number of sections that can be in service
      */
     int getMaximumSectionCount();
 
     /**
-     * Set the maximum number of section.
+     * Change the count of sections in service.
+     * Please note sections can only be sequentially in service i.e. all section from section 1 to section currentSectionCount are in service.
```
geofjamg, line 38: Same?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorNonLinearModel.java
```diff
@@ -0,0 +1,51 @@
...
+        Section setG(double g);
+    }
+
+    /**
+     * Get an optional of the section associated with a given section index if it exists.
+     *
+     * @param sectionIndex the index of the wanted section. Must be in [1;maximumSectionCount]. Else, throws a {@link ValidationException}.
+     *
+     */
+    Section getSection(int sectionIndex);
```
geofjamg, line 45: sectionCount?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
```diff
@@ -46,21 +44,106 @@ protected String getTypeDescription() {
...
+            ValidationUtil.checkMaximumSectionCount(ShuntCompensatorAdderImpl.this, maximumSectionCount);
+            model = new ShuntCompensatorLinearModelImpl(bPerSection, gPerSection, maximumSectionCount);
+            return ShuntCompensatorAdderImpl.this;
+        }
+    }
+
+    class ShuntCompensatorNonLinearModelAdderImpl implements ShuntCompensatorNonLinearModelAdder {
+
+        private final List<ShuntCompensatorNonLinearModelImpl.SectionImpl> sections = new ArrayList<>();
+        private boolean hasDefinedG = false;
```
geofjamg, line 85: Is it just for debugging?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
```diff
@@ -46,21 +44,106 @@ protected String getTypeDescription() {
...
-        return this;
+    public ShuntCompensatorNonLinearModelAdder newNonLinearModel() {
+        return new ShuntCompensatorNonLinearModelAdderImpl();
     }
 
     @Override
-    public ShuntCompensatorAdder setCurrentSectionCount(int currentSectionCount) {
-        this.currentSectionCount = currentSectionCount;
+    public ShuntCompensatorAdder setSectionCount(int sectionCount) {
+        this.currentSectionCount = sectionCount;
```
geofjamg, line 146: currentSectionCount => currentSectionCount

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorImpl.java
```diff
@@ -47,13 +39,12 @@
     private final TDoubleArrayList targetDeadband;
 
     ShuntCompensatorImpl(Ref<? extends VariantManagerHolder> network,
-                         String id, String name, boolean fictitious, double bPerSection, int maximumSectionCount,
+                         String id, String name, boolean fictitious, ShuntCompensatorModelWrapper model,
                          int currentSectionCount, TerminalExt regulatingTerminal, boolean voltageRegulatorOn,
```
geofjamg, line 43: currentSectionCount => sectionCount

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorImpl.java
```diff
@@ -79,51 +75,68 @@ public TerminalExt getTerminal() {
     }
 
     @Override
-    public double getbPerSection() {
-        return bPerSection;
+    public int getSectionCount() {
+        return currentSectionCount.get(network.get().getVariantIndex());
```
geofjamg, line 79: currentSectionCount => sectionCount

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorNonLinearModelImpl.java
```diff
@@ -0,0 +1,142 @@
...
+
+    private final List<SectionImpl> sections;
+
+    ShuntCompensatorNonLinearModelImpl(List<SectionImpl> sections) {
+        this.sections = sections;
+    }
+
+    @Override
+    public Section getSection(int sectionIndex) {
+        if (sectionIndex < 1 || sectionIndex > sections.size()) {
```
geofjamg, line 83: I still do not understand why section index start from 1, it is a bad convention.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorNonLinearModelImpl.java
```diff
@@ -0,0 +1,142 @@
...
+        }
+        return sectionCount == 0 || Double.isNaN(sections.get(sectionCount - 1).getG()) ? 0 : sections.get(sectionCount - 1).getG();
+    }
+
+    @Override
+    public double getBPerSection(int sectionIndex) {
+        if (sectionIndex < 1 || sectionIndex > sections.size()) {
+            throw new ValidationException(shuntCompensator, "invalid section index (must be in [1;maximumSectionCount]");
+        }
+        return sectionIndex == 1 ? sections.get(0).getB() : sections.get(sectionIndex - 1).getB() - sections.get(sectionIndex - 2).getB();
```
geofjamg, line 125: This is very dirty code just because of index > 0 convention ?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorNonLinearModelImpl.java
MioRtia: Method deleted (as every method using `sectionIndex` concept)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorNonLinearModelImpl.java
MioRtia: Method deleted (as every method using `sectionIndex` concept)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
MioRtia: No, this is a bit complicated for what it does, but it is to check that if the conductance of one section has been defined, all sections must have defined conductance (and vice versa, if a section with an undefined conductance has been added, all sections must have undefined conductances).

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractShuntCompensatorTest.java
```diff
@@ -268,18 +471,25 @@ public void testSetterGetterInMultiVariants() {
...
+
+    private ShuntCompensatorAdder createShuntAdder(String id, String name, int currentSectionCount, Terminal regulatingTerminal, boolean voltageRegulatorOn, double targetV, double targetDeadband) {
+        return voltageLevel.newShuntCompensator()
                 .setId(id)
                 .setName(name)
                 .setConnectableBus("busA")
-                .setbPerSection(bPerSection)
-                .setCurrentSectionCount(currentSectionCount)
-                .setMaximumSectionCount(maxSectionCount)
+                .setSectionCount(currentSectionCount)
```
marqueslanauja, line 489: Maybe change CurrentSectionCount to be coherent

MioRtia:
> Only a general comment. After reviewing the PR, I understood that the zero section is always defined. I would prefer to define it only when it was defined in the CGMES model. In my personal opinion IIDM model should always be as close as possible to the user defined model.

Thank you for your feedback! The "zero" section is not really "defined", it is just that when no section is in service, the shunt is considered as disconnected (even when this is not specified explicitly in CGMES). I think it would be weird to have a shunt with no disconnected state, what do you think?

PR #1325, opened 2020-05-27
check voltage level on both side of the line

### LINKED ISSUE #1242: [UCTE] Lines should have the same nominal voltage at each side
Bug

If a UCTE line is defined between two buses with different nominal voltages, the import works properly but it's an error.

Try to import a UCTE file with this kind of line:
```
ABCDEF61 AGHIJK31 1 0 0.5770 1.0950 7.576000    67
```

An exception should be thrown. The ENTSO-E Quality of dataset indicates:
> Both line connectivity nodes must have the same voltage level, i.e. the same 7th character.

The error should be thrown by the `fix()` method, so you have to override this method in the UcteLine class to check that the 7th characters of each nodes are the same.

  - PowSyBl Version: ...
  - OS Version: ...

jlabous:

fix #1242


ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
```diff
@@ -399,6 +399,9 @@ private static void createLines(UcteNetworkExt ucteNetwork, Network network) {
         for (UcteLine ucteLine : ucteNetwork.getLines()) {
             UcteNodeCode nodeCode1 = ucteLine.getId().getNodeCode1();
             UcteNodeCode nodeCode2 = ucteLine.getId().getNodeCode2();
+            if (!nodeCode1.getVoltageLevelCode().equals(nodeCode2.getVoltageLevelCode())) {
+                throw new UcteException("Line with two different nominal voltages");
+            }
```
mathbagu, line 404: I think this test should be done in the `UcteElement::fix` method.
@murgeyseb what do you think?
@jlabous You should give more information in the error message: with your proposal, if this exception is thrown, I don't know which line is the faulty one so it's impossible to debug/fix it.

ucte/ucte-converter/src/test/resources/differentLinesVoltage.uct
```diff
@@ -0,0 +1,12 @@
+##C 2007.05.01
+Test case for UCTE import issue in PowSyBl core package
+contact : sebastien.murgey@rte-france.com
```
mathbagu, line 3: Do not copy/paste test cases, but adapt it to the situation.

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
murgeyseb: Yes it could be done in UcteElement::fix, it is true. And the message should be more explicit.
But this method is supposed to fix the issue, not throw exception, but why not.

mathbagu: So maybe we can do the test in the constructor then?

PR #1326, opened 2020-05-27
Mergingview vl adapter

mathbagu:

Bug fix

The views in the VoltageLevelAdapter don't work properly: they don't manage the bus cache properly. Some of buses are temporary and should be invalidated when the topology change. They have to be cached in the VoltageLevel itself. Some of the buses are configured, and should be cached in the merging view

This PR try to change this behaviour.

**Other information**:
Before merging this PR, we have to fix the bus definition for the BusView, and consider the DanglingLine as a branch, not only a feeder.


iidm/iidm-mergingview/src/test/java/com/powsybl/iidm/mergingview/TopologyTest.java
```diff
@@ -0,0 +1,73 @@
...
+        MergingView cgm = MergingViewFactory.createCGM(null);
+        Network n1 = cgm.getNetwork("n1");
+        Network n2 = cgm.getNetwork("n2");
+        assertNotNull(n1);
+        assertNotNull(n2);
+
+        Line mergedLine = cgm.getLine("DL1 + DL2");
+        assertNotNull(mergedLine);
+        assertNotNull(mergedLine.getTerminal1().getBusView().getBus());
+        // FIXME(mathbagu) assertNotNull(mergedLine.getTerminal2().getBusView().getBus());
```
annetill, line 34: Hi @mathbagu ! I have tried to resolve this fixme and it is fix with this PR https://github.com/powsybl/powsybl-core/pull/1316/files

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/BusBreakerVoltageLevelAdapter.java
```diff
@@ -0,0 +1,211 @@
...
+
+    @Override
+    public BusViewExt getBusView() {
+        return busView;
+    }
+
+    @Override
+    public void invalidateCache() {
+        busView.invalidateCache();
+        System.out.println("BBKVLA::invalidate");
```
MioRtia, line 209: Is this `System.out.println` voluntary?

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/NodeBreakerVoltageLevelAdapter.java
```diff
@@ -0,0 +1,407 @@
...
+    @Override
+    public BusViewExt getBusView() {
+        return busView;
+    }
+
+    @Override
+    public void invalidateCache() {
+        busBreakerView.invalidateCache();
+        busView.invalidateCache();
+        System.out.println("INVALIDATE");
```
MioRtia, line 405: Same remark as before. Shouldn't a logger be used?

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/AbstractVoltageLevelAdapter.java
```diff
@@ -0,0 +1,390 @@
...
+        getDelegate().setHighVoltageLimit(highVoltageLimit);
+        return this;
+    }
+
+    @Override
+    public <T extends Connectable> T getConnectable(String id, Class<T> clazz) {
+        T connectable = getDelegate().getConnectable(id, clazz);
+        if (clazz == DanglingLine.class) {
+            DanglingLine dl = (DanglingLine) connectable;
+            return dl == null || getIndex().isMerged(dl) ? null : clazz.cast(getIndex().getDanglingLine(dl));
```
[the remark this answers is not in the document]
mathbagu: In the merging view, the dangling line doesn't exists anymore. So, if you get a DanglingLine from the delegate, and it's merged, you should ignore it (the mergedLine has its own ID).

But, you right, I should change the end of this method if the connectable is not null, to have the same error than in the iidm-impl

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/BusBreakerVoltageLevelAdapter.java
mathbagu: Just to debug, but you right, I have to remove it

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/NodeBreakerVoltageLevelAdapter.java
mathbagu: Same answer, just for the debug

iidm/iidm-mergingview/src/test/java/com/powsybl/iidm/mergingview/TopologyTest.java
annetill: Indeed, now the PR that resolves this test is #1340

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergedLine.java
```diff
@@ -493,12 +497,14 @@ public void remove() {
 
     @Override
     public <E extends Extension<Line>> E getExtension(final Class<? super E> type) {
-        throw MergingView.createNotImplementedException();
+        return null;
+        // throw MergingView.createNotImplementedException();
```
MioRtia, line 501: Just for me, could you explain why you commented the throwing of exception here? Was it only to test?

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergedLine.java
```diff
@@ -493,12 +497,14 @@ public void remove() {
...
     }
 
     @Override
     public <E extends Extension<Line>> E getExtensionByName(final String name) {
-        throw MergingView.createNotImplementedException();
+        return null;
+        // throw MergingView.createNotImplementedException();
```
MioRtia, line 507: Same question

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergedLine.java
```diff
@@ -508,7 +514,8 @@ public void remove() {
 
     @Override
     public <E extends Extension<Line>> Collection<E> getExtensions() {
-        throw MergingView.createNotImplementedException();
+        return Collections.emptyList();
+        // throw MergingView.createNotImplementedException();
```
MioRtia, line 518: Same question

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergedLine.java
mathbagu: I put two TODOs in these methods to explain why I prefer returning null or an empty list instead of throwing an exception.

PR #1329, opened 2020-05-27
Implement getLineStream(), getLines(), getDanglingLineStream() and getDanglingLines() in BusAdapter

MioRtia:

No

Feature

An exception is thrown when the user tries to retrieve lines and dangling lines connected to a bus in a merging view.

You can retrieve lines and dangling lines connected to a bus in a merging view (a merged dangling line is considered as a line).

No

### DIFF
--- a/iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/BusAdapter.java
+++ b/iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/BusAdapter.java
@@ -70,6 +70,32 @@ class BusAdapter extends AbstractIdentifiableAdapter<Bus> implements Bus {
     }
 
     @Override
+    public Iterable<DanglingLine> getDanglingLines() {
+        return getDanglingLineStream().collect(Collectors.toList());
+    }
+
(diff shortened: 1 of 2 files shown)


MioRtia:
> This is not OK to me: you have to wrap all the objects the different steams / iterables returns.

Ah yes,  good catch ><

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/BusAdapter.java
```diff
@@ -188,22 +189,28 @@ public int getConnectedTerminalCount() {
 
     @Override
     public Iterable<Line> getLines() {
-        throw MergingView.createNotImplementedException();
+        return getLineStream().collect(Collectors.toList());
```
mathbagu, line 192: I think it's not efficient to implement thinks like this: to create the Iterable, you will iterate over all the elements (collect), and then the client of the API will also do it, so the list is read twice.

I think it would be better to use Iterables.concat(). There is exactly the same issue in the MergingViewIndex... we should fix

PR #1330, opened 2020-05-27
Dynamic curves provider

mathbagu:

N/A

Feature

The `dynamic-simulation` command doesn't support curves file.

The curves file can be passed to the command, and loaded. Only groovy is supported yet.

**Other information**:
This is a breaking change because the tool is in its own maven module to make the code simpler. 

**Do not merge* this PR until the `DynamicSimulation` class has been updated to use CurvesProvider.


dynamic-simulation/dynamic-simulation-api/src/main/java/com/powsybl/dynamicsimulation/Provider.java
```diff
@@ -0,0 +1,37 @@
...
+package com.powsybl.dynamicsimulation;
+
+import com.powsybl.iidm.network.Network;
+
+import java.util.List;
+
+/**
+ * @author Mathieu Bague <mathieu.bague@rte-france.com>
+ */
+public interface Provider<T> {
```
zamarrenolm, line 17: What about naming this class as `DynamicSimulationDataProvider` or `DynamicSimulationInputsProvider` ?

mathbagu: With @agnesLeroy we think this could be moved in a more common module, so we wouldn't add a Dynamic prefix. Looking to the Javadoc, the good name should be `Supplier`. The `Supplier` class is a generic class that have only one method `get`. In our case, it's the same concept, except that the `get` method takes one argument.

dynamic-simulation/dynamic-simulation-tool/src/main/java/com/powsybl/dynamicsimulation/tool/DynamicSimulationTool.java
```diff
@@ -118,22 +120,30 @@ public void run(CommandLine line, ToolRunningContext context) throws Exception {
...
+
         DynamicSimulationParameters params = DynamicSimulationParameters.load();
         if (line.hasOption(PARAMETERS_FILE)) {
             Path parametersFile = context.getFileSystem().getPath(line.getOptionValue(PARAMETERS_FILE));
             JsonDynamicSimulationParameters.update(params, parametersFile);
         }
 
-        DynamicSimulationResult result = DynamicSimulation.run(network,
-            context.getShortTimeExecutionComputationManager(), params);
+        // TODO: use the curvesProvider
```
agnesLeroy, line 145: We should fix this TODO before merging this PR. @mathbagu did you leave it for @marcosmc to complete?

marcosmc: Done.

dynamic-simulation/dynamic-simulation-api/src/main/java/com/powsybl/dynamicsimulation/DynamicSimulation.java
```diff
@@ -71,6 +71,11 @@ public DynamicSimulationResult run(Network network, CurvesProvider curvesProvide
             return runAsync(network, curvesProvider, workingVariantId, computationManager, parameters).join();
         }
 
+        public DynamicSimulationResult run(Network network, CurvesProvider curvesProvider, ComputationManager computationManager,
+            DynamicSimulationParameters parameters) {
+            return run(network, curvesProvider, network.getVariantManager().getWorkingVariantId(), computationManager, parameters);
+        }
+
```
agnesLeroy, line 78: @marcosmc In this class we would like to have all the combinations of possible calls of run and runAsync. Could you add them? There are 11 possibilities aside from the one you've done here to provide. Let me know if it's unclear to you.

marcosmc: Done

dynamic-simulation/dynamic-simulation-api/src/main/java/com/powsybl/dynamicsimulation/DynamicSimulation.java
```diff
@@ -32,43 +32,70 @@ public Runner(DynamicSimulationProvider provider) {
...
         }
 
-        public CompletableFuture<DynamicSimulationResult> runAsync(Network network, String workingStateId,
-            ComputationManager computationManager, DynamicSimulationParameters parameters) {
-            Objects.requireNonNull(workingStateId);
-            Objects.requireNonNull(parameters);
-            return provider.run(network, computationManager, workingStateId, parameters);
+        public CompletableFuture<DynamicSimulationResult> runAsync(Network network, CurvesSupplier curvesSupplier, String workingVariantId,
+                                                                   ComputationManager computationManager, DynamicSimulationParameters parameters) {
+            return provider.run(network, curvesSupplier, computationManager, workingVariantId, parameters);
```
agnesLeroy, line 37: Here it seems computationManager and workingVariantId are inverted

dynamic-simulation/dynamic-simulation-api/src/main/java/com/powsybl/dynamicsimulation/DynamicSimulation.java
```diff
@@ -32,43 +32,70 @@ public Runner(DynamicSimulationProvider provider) {
...
-            return provider.run(network, computationManager, workingStateId, parameters).join();
+        public DynamicSimulationResult run(Network network, CurvesSupplier curvesSupplier, String workingVariantId, ComputationManager computationManager,
+                                           DynamicSimulationParameters parameters) {
+            return runAsync(network, curvesSupplier, workingVariantId, computationManager, parameters).join();
         }
 
-        public DynamicSimulationResult run(Network network, ComputationManager computationManager,
-            DynamicSimulationParameters parameters) {
+        public DynamicSimulationResult run(Network network, String workingVariantId, ComputationManager computationManager, DynamicSimulationParameters parameters) {
+            return runAsync(network, workingVariantId, computationManager, parameters).join();
```
agnesLeroy, line 76: Here shouldn't we return run(network, CurvesSuplier.empty(), workingVariantId, computationManager, parameters) instead?

dynamic-simulation/dynamic-simulation-api/src/main/java/com/powsybl/dynamicsimulation/DynamicSimulation.java
```diff
@@ -113,9 +140,9 @@ public static Runner find() {
         return find().runAsync(network);
     }
```
agnesLeroy, line 142: We also need all the run and runAsync combinations in this part of the file please.

agnesLeroy:
Now the code looks good to me, but it should be possible to improve the test coverage. @marcosmc can you have a look at it?

dynamic-simulation/dynamic-simulation-api/src/main/java/com/powsybl/dynamicsimulation/DynamicSimulationProvider.java
```diff
@@ -18,6 +18,6 @@
  */
 public interface DynamicSimulationProvider extends Versionable, PlatformConfigNamedProvider {
```
zamarrenolm, line 19: Just a question about the _semantics_ of the `Versionable` interface.

I assume the name and version strings returned by a `DynamicSimulationProvider` should be interpreted as the name and version of the external simulator that this provider is integrating in PowSyBl ?

For example for Dynawo, `getName()` would return `"Dynawo"` and `getVersion()` `"1.1.0"`. 

Or should it be related to a specific version of the provider, internal to powsybl, and used consistently for later potential updates of the same "kind" of provider ?

As an example: we start with a Dynawo provider with version `"1"` that connects only with Dynawo 1.1.0. Later we create an updated Dynawo provider, version `"2"`, that is able to connect with multiple versions of Dynawo, starting from 1.1.x and up to Dynawo 2.0.0.

How the version attribute is planned to be used in PowSyBl ?

mathbagu: No idea how we'll use it in powsybl. From my point of view, the provider should use this to ensure that it is compatible with the binary it will run. It's how we use it in RTE for Hades2.

But I don't know if we already discuss (and agree) on how it will be used. `Versionable` is a really old interface... coming from iTesla project.

PR #1335, opened 2020-06-05
Refactoring getLines(), getLineStream(), getDanglingLines(), getDanglingLineStream() in Merging View

MioRtia:

No

Refactoring

The lists returned by  `getLines()` and `getDanglingLines()` are read twice (once during the building, once during the using of the list). It is not efficient.

The list is only read when used.

No

### DIFF
--- a/iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergingView.java
+++ b/iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergingView.java
@@ -703,7 +703,7 @@ public final class MergingView implements Network {
 
     @Override
     public int getLineCount() {
-        return (int) index.getLineStream().count();
+        return index.getLineCount();
     }
 
     @Override
@@ -724,7 +724,7 @@ public final class MergingView implements Network {
 
     @Override
     public int getDanglingLineCount() {
-        return (int) index.getDanglingLineStream().count();
+        return index.getDanglingLineCount();
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergingView.java
```diff
@@ -693,17 +693,17 @@ public LineAdder newLine() {
...
     @Override
     public Stream<Line> getLineStream() {
-        return index.getLines().stream();
+        return index.getLineStream();
     }
 
     @Override
     public int getLineCount() {
-        return index.getLines().size();
+        return (int) index.getLineStream().count();
```
mathbagu, line 706: Is it the optimize way to implement this? Could we simply sum the lineCount of the underlying networks and add the number of mergedLine? I don't know if it's really better, but it avoids to iterate over all the lines

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergingView.java
```diff
@@ -714,17 +714,17 @@ public Line getLine(final String id) {
...
     @Override
     public Stream<DanglingLine> getDanglingLineStream() {
-        return index.getDanglingLines().stream();
+        return index.getDanglingLineStream();
     }
 
     @Override
     public int getDanglingLineCount() {
-        return index.getDanglingLines().size();
+        return (int) index.getDanglingLineStream().count();
```
mathbagu, line 727: Same kind of remark: we could sum the danglingLine count, avec remove 2 times the mergedLine. (1 merged Line = 2 danglingLine).

MioRtia:
> This PR is OK to me: I simply propose a possible optimization you should consider. I let you decide if we merge it like this, or if my proposal makes sense

I agree with your proposal, I will do the corrections and merge it.

PR #1355, opened 2020-06-17
add Voltage Extension class

EtienneLt:
Signed-off-by: Etienne Lesot <etienne.lesot@rte-france.com>

it adds a new extension for the limitviolations with voltage contraints. It is similar to Current Extension but for voltage.

**Other information**:

### DIFF
--- a/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/extensions/VoltageExtension.java
+++ b/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/extensions/VoltageExtension.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2018, RTE (http://www.rte-france.com)
+ * Copyright (c) 2020, RTE (http://www.rte-france.com)
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
@@ -10,16 +10,18 @@ import com.powsybl.commons.extensions.AbstractPrecontingencyValueExtension;
 import com.powsybl.security.LimitViolation;
 
 /**
- * @author Mathieu Bague <mathieu.bague at rte-france.com>
+ * Extension to handle pre-contingency voltage value for a voltage limit violation
+ *
+ * @author Olivier Bretteville <olivier.bretteville at rte-france.com>
  */
-public class CurrentExtension extends AbstractPrecontingencyValueExtension<LimitViolation> {
+public class VoltageExtension extends AbstractPrecontingencyValueExtension<LimitViolation> {
 
-    public CurrentExtension(double value) {
+    public VoltageExtension(double value) {
         super(value);
     }
 
     @Override
     public String getName() {
-        return "Current";
+        return "Voltage";
     }
 }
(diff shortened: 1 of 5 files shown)


security-analysis/security-analysis-api/src/main/java/com/powsybl/security/extensions/VoltageExtension.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.powsybl.commons.extensions.Extension;
+import com.powsybl.security.LimitViolation;
+
+/**
+ * Extension to handle pre-contingency voltage value for a voltage limit violation
+ *
+ * @author Olivier Bretteville <olivier.bretteville at rte-france.com>
+ */
+public class VoltageExtension implements Extension<LimitViolation> {
```
mathbagu, line 17: You should consider to inherit from AbstractExtension

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/VoltageExtensionSerializer.java
```diff
@@ -0,0 +1,65 @@
+/**
+ * Copyright (c) 2018, RTE (http://www.rte-france.com)
```
mathbagu, line 2: Change the copyright date.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/extensions/VoltageExtension.java
```diff
@@ -0,0 +1,53 @@
+/**
+ * Copyright (c) 2018, RTE (http://www.rte-france.com)
```
mathbagu, line 2: Change the copyright date

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/extensions/VoltageExtension.java
```diff
@@ -0,0 +1,39 @@
...
+import com.powsybl.security.LimitViolation;
+
+/**
+ * Extension to handle pre-contingency voltage value for a voltage limit violation
+ *
+ * @author Olivier Bretteville <olivier.bretteville at rte-france.com>
+ */
+public class VoltageExtension extends AbstractExtension<LimitViolation> {
+
+    private final float preContingencyValue;
```
mathbagu, line 18: @sylvlecl Should we use a `float` or a `double` there?

mathbagu:
@sylvlecl For `CurrentExtension` the security analysis API provides an interceptor that create this extension. Is it possible to do the same kind of interceptor for the `VoltageExtension`, or is it too specific to the simulator?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/extensions/VoltageExtension.java
sylvlecl: Ah, well spotted, we should use `double` as in the network and similar extensions.

sylvlecl:
It would clearly be more consistent, but unfortunately we cannot implement it as long as we cannot find the `Bus` which corresponds to a given `LimitViolation`. Today, it's still the id of the `VoltageLevel` which is reported in the violation. 

So I think this could only be done after a deeper evolution about buses identification.

PR #1356, opened 2020-06-17
Improvements of PSS/E importer:

JB-H:
- supporting both raw and RAW file extensions
- supporting STATUS = 0 of equipments
- modification of header detection
- reading Qmin and Qmax of generating units
- modification of targetQ setter
- better check of the number of windings of transformers

No

feature

- supporting both raw and RAW file extensions
- supporting STATUS = 0 of equipments
- modification of header detection
- reading Qmin and Qmax of generating units
- modification of targetQ setter
- better check of the number of windings of transformers

**Other information**:


psse/psse-converter/src/main/java/com/powsybl/psse/converter/PsseImporter.java
```diff
@@ -65,15 +66,31 @@ public String getComment() {
...
         try {
-            if (dataSource.exists(null, EXT)) {
-                try (BufferedReader reader = new BufferedReader(new InputStreamReader(dataSource.newInputStream(null, EXT)))) {
-                    String titleLine = reader.readLine();
-                    if (titleLine != null) {
-                        return titleLine.length() >= 38
-                                && titleLine.charAt(37) == '/';
+            String ext = findExtension(dataSource, false);
+            if (ext != null) {
+                try (BufferedReader reader = new BufferedReader(new InputStreamReader(dataSource.newInputStream(null, ext)))) {
```
geofjamg, line 87: Have you tried what you get in the case of a *.raw|RAW file but the content is not a raw file?

geofjamg: Ideally we should had unit tests to cover the different cases (real raw or RAW, fake one)

psse/psse-converter/src/main/java/com/powsybl/psse/converter/PsseImporter.java
```diff
@@ -65,15 +66,31 @@ public String getComment() {
...
 
+    private String findExtension(ReadOnlyDataSource dataSource, boolean throwException) throws IOException {
+        for (String ext : EXTS) {
+            if (dataSource.exists(null, ext)) {
+                return ext;
+            }
+        }
+        if (throwException) {
+            throw new PsseException("File " + dataSource.getBaseName()
+                    + "." + Joiner.on("|").join(EXTS) + " not found");
```
mathbagu, line 77: Use `String.join("|", EXTS)` instead of `Joiner.on`?

psse/psse-converter/src/main/java/com/powsybl/psse/converter/PsseImporter.java
mathbagu: raw is a really common extension, used for instance in photography. I think it's a good idea to check if the file is valid or not.

psse/psse-converter/src/main/java/com/powsybl/psse/converter/PsseImporter.java
```diff
@@ -300,62 +339,67 @@ public Network importData(ReadOnlyDataSource dataSource, NetworkFactory networkF
...
 
-            for (PsseTransformer psseTfo : psseModel.getTransformers()) {
-                createTransformer(psseTfo, containerMapping, perUnitContext, network);
+                for (PsseTransformer psseTfo : psseModel.getTransformers()) {
+                    createTransformer(psseTfo, containerMapping, perUnitContext, network);
+                }
+
+            } catch (IOException e) {
+                throw new UncheckedIOException(e);
             }
```
mathbagu, line 397: Is this catch OK? There is probably to many try/catch level where only one is sufficient?

psse/psse-converter/src/main/java/com/powsybl/psse/converter/PsseImporter.java
JB-H: I have added a test for a better check of the header

psse/psse-converter/src/main/java/com/powsybl/psse/converter/PsseImporter.java
JB-H: I modified PSSE and UCT import since they werer using the same approach

psse/psse-converter/src/main/java/com/powsybl/psse/converter/PsseImporter.java
JB-H: proposition of modification done

psse/psse-converter/src/main/java/com/powsybl/psse/converter/PsseImporter.java
geofjamg: I added an additional unit test to cover @mathbagu use case so a "fake" .raw file.

PR #1357, opened 2020-06-18
DanglingLine with generation attributes

annetill:

In UCTE format, Xnode, that are buses at boundary, can have voltage regulating capabilities. They can have a P and a Q for the load part and an other P and Q for the generator part (@murgeyseb for more details). 

In CGMES format, Equivalent Injection are used in IGM to described boundaries. They usually don't have voltage regulating capabilities excpect for HVDC modeling. 

DanglingLine can not support this. 

We will be able to model PV node at boundary with the new DanglingLine attributes, when running a powerflow on the IGM. No change on the CGM model at this stage.

We only support PQ nodes.

We will support PV nodes in the bounadries, but we will need to support this in the open loadflow.

**Other information**:


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DanglingLine.java
```diff
@@ -92,6 +92,79 @@
...
+     * @return this to allow method chaining
+     */
+    default DanglingLine setReactivePowerSetpoint(double reactivePowerSetpoint) {
+        return this;
+    }
+
+    /**
+     * Get the voltage regulation status.
+     */
+    default boolean isVoltageRegulationOn() {
```
MioRtia, line 138: Just a question, why do we use this and not something like `regulationType` that can be `OFF`, `VOLTAGE` and `REACTIVE`? Here when there is no voltage regulation, how can we know if there is an active or reactive regulation or not at all?

MioRtia: @zamarrenolm What do you think?

zamarrenolm: I think for dangling lines we should consider only voltage regulation, like in generators. In fact, we should duplicate the attributes from the IIDM generator into the dangling line, to highlight the fact that we are adding just an implicit/fictitious generator at the boundary end of the dangling line.

annetill:
@MioRtia Before opening the PR to other reviewers, could you please help me to do that object ````generator```` that will be inside the DanglingLine ? And after that, it seems that I have a low test coverage. I also have skiped the serialization (as usual). Thanks a lot for your help !

annetill:
Some unit tests are missing, maybe at least for the merging-view ?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Boundary.java
```diff
@@ -70,6 +71,12 @@ public void addLineAtNode(PropertyBag line, String node) {
         lines.add(line);
     }
 
+    public void addEquivalentInjectionAtNode(PropertyBag equivalentInjection, String node) {
+        List<PropertyBag> equivalentInjections;
+        equivalentInjections = nodesEquivalentInjections.computeIfAbsent(node, ls -> new ArrayList<>(2));
+        equivalentInjections.add(equivalentInjection);
```
mathbagu, line 77: Could be write in a single line: `nodesEquivalentInjections.computeIfAbsent(node, ls -> new ArrayList<>(2)).add(equivalentInjection);`

Why do you set the capacity to 2?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ACLineSegmentConversion.java
```diff
@@ -131,26 +131,43 @@ private void convertDanglingLine(int boundarySide) {
...
+                    .add();
+            dl.newGeneration()
+                    .setTargetP(0.0)
+                    .setTargetQ(0.0)
+                    .setTargetV(Double.NaN)
+                    .setVoltageRegulationOn(false)
+                    .add();
+        }
+        if (equivalentInjectionConversion != null) {
+            equivalentInjectionConversion.convertReactiveLimits(dl);
```
mathbagu, line 169: You should move this line at line 156.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/EquivalentInjectionConversion.java
```diff
@@ -21,55 +19,128 @@
...
+        Regulation regulation = getRegulation();
+        DanglingLine dl;
+        if (regulation.status) {
+            // If this equivalent injection is regulating voltage,
+            // map it over the dangling line 'virtual generator'
+            dl = adder
+                    .setP0(fother.p())
+                    .setQ0(fother.q())
+                    .add();
+            dl.newGeneration()
```
mathbagu, line 62: It's a bit weird: why the generation is not created using the adder, but using the DanglingLine?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DanglingLine.java
```diff
@@ -20,9 +20,106 @@
...
+        /**
+         * <p>Set the generator voltage target in Kv.</p>
+         * <p>Depends on the working variant.</p>
+         * @param targetV the generator voltage target
+         * @return this to allow method chaining
+         */
+        Generation setTargetV(double targetV);
+    }
+
+    interface GenerationAdder {
```
mathbagu, line 107: This class should be moved into the DanglingLineAdder?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DanglingLineAdder.java
```diff
@@ -21,6 +21,7 @@
  *</pre>
  *
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ * @author Anne Tilloy <anne.tilloy at rte-france.com>
```
mathbagu, line 24: To be reverted if you do not make any change

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineAdderImpl.java
```diff
@@ -12,6 +12,8 @@
 /**
  *
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ * @author Anne Tilloy <anne.tilloy at rte-france.com>
```
mathbagu, line 15: Same remark

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineImpl.java
```diff
@@ -6,18 +6,226 @@
...
 import gnu.trove.list.array.TDoubleArrayList;
 
 /**
  *
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
  */
-class DanglingLineImpl extends AbstractConnectable<DanglingLine> implements DanglingLine, CurrentLimitsOwner<Void> {
+class DanglingLineImpl extends AbstractConnectable<DanglingLine> implements DanglingLine, CurrentLimitsOwner<Void>, ReactiveLimitsOwner {
+
+    class GenerationImpl implements Generation, MultiVariantObject {
```
mathbagu, line 20: I think it's not necessary to inherit from MultiVariantObject. If the enclosing object is alreay a MultiVariantObject it could be ok

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineImpl.java
```diff
@@ -6,18 +6,226 @@
...
+            for (int index : indexes) {
+                targetP.set(index, targetP.get(sourceIndex));
+                targetQ.set(index, targetQ.get(sourceIndex));
+                voltageRegulationOn.set(index, voltageRegulationOn.get(sourceIndex));
+                targetV.set(index, targetV.get(sourceIndex));
+            }
+        }
+    }
+
+    class GenerationAdderImpl implements GenerationAdder {
```
mathbagu, line 177: To be moved into DanglingAdderImpl

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineImpl.java
```diff
@@ -31,15 +239,20 @@
 
     private final String ucteXnodeCode;
 
+    private GenerationImpl generation;
+
     private CurrentLimitsImpl limits;
 
+    private final ReactiveLimitsHolderImpl reactiveLimits;
```
mathbagu, line 246: Maybe the reactiveLimits should be hold by the generation object: does it make sense to have no generation but a reactiveLimits?

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
```diff
@@ -257,6 +247,23 @@ private static void createDanglingLine(UcteLine ucteLine, boolean connected,
                 .setQ0(q0)
                 .setUcteXnodeCode(xnode.getCode().toString())
                 .add();
+        dl.newGeneration()
+                .setTargetP(-targetP)
+                .setTargetQ(-targetQ)
+                .add();
```
mathbagu, line 253: You should create a generation only if necessary: i.e there is a voltage setpoint or a power generation?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/EquivalentInjectionConversion.java
MioRtia: I used this design (similar to reactive limits for example) because the dangling line object can be consistent without it (contrary to shunt model or T3W's legs for example). Do you think it would be better to put it in DLAdder?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineImpl.java
MioRtia: No, you are right, I am changing it

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Boundary.java
MioRtia: @annetill Is there a reason? I didn't touch your code here.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineImpl.java
MioRtia: I'll change it

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Boundary.java
MioRtia: Rectification: it seems to be @zamarrenolm's code. @zamarrenolm can you tell us why the list capacity is at 2 please?

annetill: Because in a boundary TN, you have 2 EquivalentInjections, one of each IGM.

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
MioRtia: @annetill Are these values used in loadflow even when there is no voltage setpoint and/or power generation?

annetill: If the generator is regulating voltage, I use the voltage setpoint, otherwise I use the generator targetQ. From CGMES files, I will most of the time use only P0 and Q0. From UCTE file, I will most of the time have both load part and generation part, as it is described in the input data.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineAdderImpl.java
```diff
@@ -12,9 +12,64 @@
...
+
+        @Override
+        public GenerationAdder setTargetV(double targetV) {
+            this.targetV = targetV;
+            return this;
+        }
+
+        @Override
+        public DanglingLineAdder add() {
+            ValidationUtil.checkActivePowerLimits(DanglingLineAdderImpl.this, minP, maxP);
```
mathbagu, line 67: In GeneratorAdderImpl, we check a lot of thing I would expect to find here such as:
 - ValidationUtil.checkMinP(this, minP);
- ValidationUtil.checkMaxP(this, maxP);
- ValidationUtil.checkActivePowerSetpoint(this, targetP);
- ValidationUtil.checkVoltageControl(this, voltageRegulatorOn, targetV, targetQ);

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineImpl.java
```diff
@@ -19,6 +17,166 @@
  */
 class DanglingLineImpl extends AbstractConnectable<DanglingLine> implements DanglingLine, CurrentLimitsOwner<Void> {
 
+    public static class GenerationImpl implements Generation, ReactiveLimitsOwner, Validable {
```
mathbagu, line 20: Same king of remark: why there is no integrity checks in the different setters?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineImpl.java
```diff
@@ -54,6 +215,11 @@
         this.g = g;
         this.b = b;
         this.ucteXnodeCode = ucteXnodeCode;
+        if (generationAdder != null) {
+            generation = new GenerationImpl(generationAdder, this);
+        } else {
+            generation = null;
+        }
```
mathbagu, line 222: That's strange: the constructor should have a Generation parameter instead of mixing data and adder?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineImpl.java
```diff
@@ -19,6 +17,166 @@
...
+
+        private final TDoubleArrayList targetP;
+
+        private final TDoubleArrayList targetQ;
+
+        private final TBooleanArrayList voltageRegulationOn;
+
+        private final TDoubleArrayList targetV;
+
+        GenerationImpl(DanglingLineAdderImpl.GenerationAdderImpl adder, DanglingLineImpl danglingLine) {
```
mathbagu, line 40: This is not a good design: you should keep only the second constructor. Remove the DanglingLineImpl and male the generation class not static to be able to access to the DanglingLine.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineImpl.java
```diff
@@ -188,13 +359,31 @@ public void extendVariantArraySize(int initVariantArraySize, int number, int sou
...
+            generation.targetQ.ensureCapacity(generation.targetQ.size() + number);
+            generation.voltageRegulationOn.ensureCapacity(generation.voltageRegulationOn.size() + number);
+            generation.targetV.ensureCapacity(generation.targetV.size() + number);
+            for (int i = 0; i < number; i++) {
+                generation.targetP.add(generation.targetP.get(sourceIndex));
+                generation.targetQ.add(generation.targetQ.get(sourceIndex));
+                generation.voltageRegulationOn.add(generation.voltageRegulationOn.get(sourceIndex));
+                generation.targetV.add(generation.targetV.get(sourceIndex));
+            }
+        }
```
mathbagu, line 373: Move the variant management in the generation: this is an encapsulation issue, the generation should be in charge of its data.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/DanglingLineXml.java
```diff
@@ -62,6 +84,11 @@ protected DanglingLineAdder createAdder(VoltageLevel vl) {
 
     @Override
     protected DanglingLine readRootElementAttributes(DanglingLineAdder adder, NetworkXmlReaderContext context) {
+        throw new UnsupportedOperationException();
+    }
```
mathbagu, line 88: Is it correct?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/DanglingLineXml.java
```diff
@@ -70,26 +97,106 @@ protected DanglingLine readRootElementAttributes(DanglingLineAdder adder, Networ
...
+    class TemporaryLimitXml {
+        private final String name;
+        private final double value;
+        private final boolean fictitious;
+
+        TemporaryLimitXml(String name, double value, boolean fictitious) {
+            this.name = name;
+            this.value = value;
+            this.fictitious = fictitious;
+        }
```
mathbagu, line 200: I'm surprised: this piece of code change a lot. To be checked, but maybe there is a design issue that makes this code complex.

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteExporter.java
```diff
@@ -282,6 +282,29 @@ private static void convertXNode(UcteNetwork ucteNetwork, DanglingLine danglingL
...
+            if (maxP != DEFAULT_POWER_LIMIT) {
+                ucteNode.setMaximumPermissibleActivePowerGeneration(-maxP);
+            }
+            if (minQ != -DEFAULT_POWER_LIMIT) {
+                ucteNode.setMinimumPermissibleReactivePowerGeneration(-minQ);
+            }
+            if (maxQ != DEFAULT_POWER_LIMIT) {
+                ucteNode.setMaximumPermissibleReactivePowerGeneration(-maxQ);
+            }
+        }
```
mathbagu, line 307: Should we export the reactive limits even if the voltage regulation is OFF?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineAdderImpl.java
MioRtia: Yes but in this case minP/maxP are not mandatory (which leads to no check), same goes for targets.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineImpl.java
MioRtia: The issue is `GenerationImpl` cannot be created if the `DanglingLineImpl` object has not being created... I am a bit bothered by it too but I did not have any better idea.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineImpl.java
MioRtia: Hm, this was what was done at first but I think you told me to change it? Did I miss something?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/DanglingLineXml.java
MioRtia: It changes a lot because by moving `GenerationAdder` in `DanglingLineAdder`, we now cannot call `add()` before reading sub elements (similar to what is done in `ShuntXml`)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineAdderImpl.java
MioRtia: I am not sure how generation data is actually used in CGMES/UCTE, @annetill do you have any input on this?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineImpl.java
MioRtia: See my remark on the comment above

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/DanglingLineXml.java
MioRtia: Do you prefer using attributes in danglingline like for threewindingstransformers?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineAdderImpl.java
annetill: Indeed from CGMES and from UCTE, we will create minimal generation part of a dangling line:
- Sometimes a generation that has only targetQ and targetP ;
- Sometimes a generation that has only targetP, targetV and voltageRegulationOn.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineImpl.java
MioRtia: Okay, I changed a bit the implementation in my last commit and put the same mecanism that with T3W's leg: a generation without a set `danglingline` is created at first and `danglingLine` is set when the object is created. It is a bit cleaner.

PR #1358, opened 2020-06-18
Export UCT - change of status

benrejebmoh:

BugFix

In the exported CGM, the status of all the tie-lines are changed from 0 = real element to 1 = equivalent element.
This is also a problem for all X-nodes, which are changed from real to equivalent node status
In also happens on the other way round: internal equivalent lines become real ones after the merging. 

Keeping Ucte element statut as property in import then use it in export.
For DanglingLines we keep both UcteNodeStatus and UcteElementStatus.
For XNode Creation in export if there is any incoherence between the two network, we use UcteNodeStatus.REAL as default value.
Example:
network1:
##ZXX
XXXXXX11 XNODE 1      0 0        3.96708 0.00000 0.00000 0.00000
network2:
##ZXX
XXXXXX11              1 0        3.96708 0.00000 0.00000 0.00000
Merged network will be:
##ZXX
XXXXXX11 XNODE 1      0 0        0.00000 0.00000 0.00000 0.00000

**Other information**:


mathbagu:
@benrejebmoh Are you sure this cannot be managed using the fictitious status of IIDM. This flag is now available on all equipments, so you should be able to initialize the fictitious attribute from UCTE, and be able to export it back?

benrejebmoh:
@mathbagu that could work for XNodes, but how i could manage UcteElementStatus otherwise?

mathbagu:
> @mathbagu that could work for UcteNodeStatus, but how i could manage UcteElementStatus otherwise?

Exactly the same: I'm not sure the fictitious flag is correctly initialized in the import, but it should be possible to retrieve the status using it:
- In/Out means connected/disconnected for a branch, and close/open for a switch (coupler)
- Fictitious for real/equivalent equipment.

Do I miss something?

benrejebmoh:
@mathbagu , 
Thank you for the explanation.
For the danglingLines i need to keep both node status and element status. that's why in createDanglingLine method you will find  addStatusProperty(ucteLine, dl) and addStatusProperty(xnode, dl).
In addition for tieLine creation i need to identify element status for each half (exactly like element name).
That's why i made the choice to work with properties for every thing related to status managment.
Indeed i can refactor my code to work with isFictious flag when it's possible and with properties otherwise.
What do you recommend?

mathbagu:
I don't know which is the best way to implement this:
- using a couple of properties
- adding a fictitious flag on both HalfLine of a TieLine
- using the MergedXnode extension

@sebamurg: any suggestion?

murgeyseb:
> I don't know which is the best way to implement this:
> 
>     * using a couple of properties
> 
>     * adding a fictitious flag on both HalfLine of a TieLine
> 
>     * using the MergedXnode extension
> 
> 
> @murgeyseb : any suggestion?

For consistency with elementName properties, I would prefer using properties for that, at least while we do not have dedicated extensions for storing UCTE dedicated data. I do not know who uses MergedXnode extension, because we do not use it in merging server, nor in FARAO.

benrejebmoh:
So is it ok?

murgeyseb:
After a short talk with Mathieu, it seems that it may be better to add a fictitious flag for half lines. I try to do a modification of the PR to take it into account

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/TieLineAdapter.java
```diff
@@ -34,4 +34,9 @@ public HalfLine getHalf1() {
     public HalfLine getHalf2() {
         return ((TieLine) getDelegate()).getHalf2();
     }
+
+    @Override
+    public HalfLine getHalf(Side side) {
+        return ((TieLine) getDelegate()).getHalf(side);
```
mathbagu, line 40: I wonder if we should override getDelegate() to return a TieLine and avoid casting everywhere

murgeyseb: As it is an AbstractAdapter<Line>, I do not think it is possible to change return type of getDelegate to TieLine.

PR #1359, opened 2020-06-19
IIDM operational limits 

### LINKED ISSUE #899: Operational Limits
Feature

Only some operational limits are imported from CGMES files

All operational limits should be imported from CGMES files. The corresponding objects must be created in IIDM if they don't exist.

To not lose information

Some changes in IIDM modeling might be necessary

MioRtia:

fix #899 

Feature : support all kinds of operational limits, not only current limits

Yes

**Other information**:
This is only the change of API, this does not take the new serialization into account nor the new conversions


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/LoadingLimits.java
```diff
@@ -0,0 +1,72 @@
+package com.powsybl.iidm.network;
```
mathbagu, line 1: Don't forget the license header

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/LoadingLimitsAdder.java
```diff
@@ -0,0 +1,37 @@
+/**
+ * Copyright (c) 2016, All partners of the iTesla project (http://www.itesla-project.eu/consortium)
```
mathbagu, line 2: Fix copyright information. I think you rename the CurrentLimitsAdder to LoadingLimitsAdder, but maybe you should fix it

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/OperationalLimits.java
```diff
@@ -0,0 +1,19 @@
+package com.powsybl.iidm.network;
```
mathbagu, line 1: Don't forget the license header

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/LoadingLimits.java
```diff
@@ -0,0 +1,82 @@
...
+         * @return the acceptable duration of the limit in second
+         */
+        int getAcceptableDuration();
+
+        /**
+         * Check if the temporary limit is a real limit corresponding to an overloading  protection or just an operating
+         * rule
+         * @return false if it is a real limit, false otherwise
+         */
+        boolean isFictitious();
```
annetill, line 45: I am really at ease with this function name choice. I imagine that we have it from the actual API. What do you think ?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Branch.java
```diff
@@ -57,16 +60,79 @@
...
-    CurrentLimits getCurrentLimits1();
+    /**
+     * @deprecated Use {@link #getOperationalLimits1(LimitType, Class)} instead.
+     */
+    @Deprecated
+    default CurrentLimits getCurrentLimits1() {
+        return getOperationalLimits1(LimitType.CURRENT, CurrentLimits.class);
+    }
+
+    default List<OperationalLimits> getOperationalLimits1() {
```
annetill, line 84: I am not sure to well understand: here we don't have the methods ````getApparentPowerLimits1()```` and ````getActivePowerLimits1()```` ?  So how do I do to get them ?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DanglingLine.java
```diff
@@ -98,8 +101,21 @@
...
-    CurrentLimitsAdder newCurrentLimits();
-
+    /**
+     * @deprecated Use {@link #getOperationalLimits(LimitType, Class)} instead.
+     */
+    @Deprecated
+    default CurrentLimits getCurrentLimits() {
+        return getOperationalLimits(LimitType.CURRENT, CurrentLimits.class);
+    }
+
```
annetill, line 111: Same remark, how do I access to apparent power limits and active power limits ?

annetill: Indeed you just leave ````getCurrentLimits()```` for compatibility but you will remove it ?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Switch.java
```diff
@@ -53,4 +56,44 @@
...
+    default List<OperationalLimits> getOperationalLimits1() {
+        return Collections.emptyList();
+    }
+
+    default <L extends OperationalLimits> L getOperationalLimits1(LimitType limitType, Class<L> limitClazz) {
+        return null;
+    }
+
+    default CurrentLimitsAdder newCurrentLimits1() {
+        throw new UnsupportedOperationException();
```
annetill, line 69: Mmmmm I think that we have a problem here. Do you have any idea of the solution ?

annetill: Do we have a side for switches ?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Injection.java
```diff
@@ -18,4 +21,27 @@
...
+    default List<OperationalLimits> getOperationalLimits() {
+        return Collections.emptyList();
+    }
+
+    default <L extends OperationalLimits> L getOperationalLimits(LimitType type, Class<L> limitClazz) {
+        return null;
+    }
+
+    default CurrentLimitsAdder newCurrentLimits() {
+        throw new UnsupportedOperationException();
```
annetill, line 33: Why do we have this exception ?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Branch.java
```diff
@@ -28,7 +31,7 @@
          * In particular, it determines the duration during which
          * the current current value may be sustained.
          */
-        CurrentLimits.TemporaryLimit getTemporaryLimit();
+        LoadingLimits.TemporaryLimit getTemporaryLimit();
```
mathbagu, line 34: Should we be more specific? The type of the limit is not CURRENT anymore?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Branch.java
```diff
@@ -57,16 +60,79 @@
 
     Side getSide(Terminal terminal);
 
-    CurrentLimits getCurrentLimits(Side side);
+    /**
+     * @deprecated Use {@link #getOperationalLimits1(LimitType, Class)} or {@link #getOperationalLimits2(LimitType, Class)} instead.
```
mathbagu, line 64: I don't know if it's possible, but to have the same bahaviour you have to use the new method, but the first parameter is `LimitType.CURRENT` and the second one is `CurrentLimits.class`

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Branch.java
```diff
@@ -57,16 +60,79 @@
...
+            return getOperationalLimits1(LimitType.CURRENT, CurrentLimits.class);
+        } else if (side == Side.TWO) {
+            return getOperationalLimits2(LimitType.CURRENT, CurrentLimits.class);
+        }
+        throw new AssertionError("Unexpected side: " + side);
+    }
 
-    CurrentLimits getCurrentLimits1();
+    /**
+     * @deprecated Use {@link #getOperationalLimits1(LimitType, Class)} instead.
```
mathbagu, line 77: Same remark

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Branch.java
mathbagu: This method will return a list of operational limit, that contains apparent limits, active power limits, and current limits.
In the API, we have to give a default behavior to the method to ensure backward compatibility. In previous IIDM version, only current limits was supported. So the implementation seems correct.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Branch.java
```diff
@@ -57,16 +60,79 @@
...
+    default CurrentLimits getCurrentLimits1() {
+        return getOperationalLimits1(LimitType.CURRENT, CurrentLimits.class);
+    }
+
+    default List<OperationalLimits> getOperationalLimits1() {
+        return Collections.singletonList(getCurrentLimits1());
+    }
+
+    default <L extends OperationalLimits> L getOperationalLimits1(LimitType limitType, Class<L> limitClazz) {
+        return limitType == LimitType.CURRENT && limitClazz == CurrentLimits.class ? (L) getCurrentLimits1() : null;
```
mathbagu, line 89: I don't know how it's possible, but do we really need to pass both parameters? For instance, in `ReactiveLimitsHolder`, we only pass the class:
```
<L extends ReactiveLimits> L getReactiveLimits(Class<L> type)
```

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Branch.java
```diff
@@ -57,16 +60,79 @@
...
+        throw new UnsupportedOperationException();
+    }
+
+    default ActivePowerLimitsAdder newActivePowerLimits1() {
+        throw new UnsupportedOperationException();
+    }
+
+    default VoltageLimitsAdder newVoltageLimits1() {
+        throw new UnsupportedOperationException();
+    }
```
mathbagu, line 104: Are all these type of limits are possible for lines and/or transformers? I thought apparentPower was for transformers: we do not have this information for lines.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Injection.java
```diff
@@ -18,4 +21,27 @@
...
+    }
+
+    default ActivePowerLimitsAdder newActivePowerLimits() {
+        throw new UnsupportedOperationException();
+    }
+
+    default VoltageLimitsAdder newVoltageLimits() {
+        throw new UnsupportedOperationException();
+    }
```
mathbagu, line 46: Some of the operation limits are weird for injections. In IIDM, we do not have sufficient information to compute them.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/LoadingLimits.java
mathbagu: I agree: we should rename this method. This method returns true if there is an protection automaton that monitor the equipment. I think we can find a better name.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/OperationalLimits.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+public interface OperationalLimits {
+    /**
+     * Get the operational limits' type (can be APPARENT_POWER, CURRENT or VOLTAGE)
+     */
+    LimitType getLimitType();
+
+    default void remove() {
+        // do nothing
+    }
```
mathbagu, line 21: I'm not convinced you have to provide a default method here: default method should not be used to factorize the code (use abstract class instead).

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Switch.java
```diff
@@ -53,4 +56,44 @@
...
+    }
+
+    default ApparentPowerLimitsAdder newApparentPowerLimits2() {
+        throw new UnsupportedOperationException();
+    }
+
+    default VoltageLimitsAdder newVoltageLimits2() {
+        throw new UnsupportedOperationException();
+    }
```
mathbagu, line 98: I would refactor a little bit the code: create a OperationLimitsHolder interface, to avoid to declare these methods everywhere?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -385,8 +385,9 @@ public static void checkConnected(Validable validable, Boolean connected) {
     }
 
     public static void checkPermanentLimit(Validable validable, double permanentLimit) {
+        // TODO: if (Double.isNaN(permanentLimit) || permanentLimit <= 0) {
         if (permanentLimit <= 0) {
-            throw new ValidationException(validable, "permanent limit must be > 0");
+            throw new ValidationException(validable, "permanent limit must defined and be > 0");
```
mathbagu, line 390: I think a word is missing:
> must **be** defined and **greater than** 0

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractBranch.java
```diff
@@ -80,54 +82,71 @@ public Side getSide(Terminal terminal) {
...
+    public VoltageLimitsAdder newVoltageLimits2() {
+        return operationalLimitsHolder2.newVoltageLimits();
+    }
+
+    OperationalLimitsHolderImpl getLimitsHolder1() {
+        return operationalLimitsHolder1;
+    }
+
+    OperationalLimitsHolderImpl getLimitsHolder2() {
+        return operationalLimitsHolder2;
```
mathbagu, line 149: Is it really useful to have these getters?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractLoadingLimits.java
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * @author Miora Ralambotiana <miora.ralambotiana at rte-france.com>
+ */
+abstract class AbstractLoadingLimits<L extends AbstractLoadingLimits<L>> implements LoadingLimits {
+
+    private double permanentLimit;
+
+    private final TreeMap<Integer, TemporaryLimit> temporaryLimits;
+
+    protected final OperationalLimitsOwner owner;
```
mathbagu, line 25: Reorder these attributes: owner, permanentLimit, temporaryLimit

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractLoadingLimits.java
```diff
@@ -0,0 +1,98 @@
...
+            return acceptableDuration;
+        }
+
+        @Override
+        public boolean isFictitious() {
+            return fictitious;
+        }
+    }
+
+    AbstractLoadingLimits(double permanentLimit, TreeMap<Integer, TemporaryLimit> temporaryLimits, OperationalLimitsOwner owner) {
```
mathbagu, line 65: Move the owner at the first place.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ApparentPowerLimitsImpl.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+class ApparentPowerLimitsImpl extends AbstractLoadingLimits<ApparentPowerLimitsImpl> implements ApparentPowerLimits {
+
+    ApparentPowerLimitsImpl(double permanentLimit, TreeMap<Integer, TemporaryLimit> temporaryLimits, OperationalLimitsOwner owner) {
+        super(permanentLimit, temporaryLimits, owner);
+    }
+
+    @Override
+    public void remove() {
+        owner.setOperationalLimits(LimitType.APPARENT_POWER, null);
```
mathbagu, line 25: I think the remove method could me mutualized:
```
owner.setOperationLimit(getLimitType(), null);
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/OperationalLimitsHolderImpl.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+class OperationalLimitsHolderImpl implements OperationalLimitsOwner {
+
+    private final EnumMap<LimitType, OperationalLimits> operationalLimits = new EnumMap<>(LimitType.class);
+    private final AbstractIdentifiable<?> identifiable;
+    private final String attributeName;
+
+    OperationalLimitsHolderImpl(String attributeName, AbstractIdentifiable<?> identifiable) {
+        this.identifiable = identifiable;
+        this.attributeName = attributeName;
```
mathbagu, line 26: Are null allowed?
You should switch the parameter order

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ThreeWindingsTransformerImpl.java
```diff
@@ -37,7 +34,7 @@
 
         private double ratedS;
 
-        private CurrentLimits limits;
+        private OperationalLimitsHolderImpl operationalLimitsHolder;
```
mathbagu, line 37: Should be declared as final?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VoltageLimitsImpl.java
```diff
@@ -0,0 +1,57 @@
...
+    public double getHighVoltage() {
+        return highVoltage;
+    }
+
+    @Override
+    public VoltageLimits setHighVoltage(double highVoltage) {
+        if (Double.isNaN(lowVoltage) && Double.isNaN(highVoltage)) {
+            throw new ValidationException(owner, "At least the low or the high voltage limit must be defined.");
+        }
+        ValidationUtil.checkVoltageLimits(owner, lowVoltage, highVoltage);
```
mathbagu, line 38: Do we need to introduce a new method to avoid this duplication (3 times)

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/AbstractBranchAdapter.java
```diff
@@ -53,30 +52,83 @@ public Side getSide(final Terminal terminal) {
...
+    @Deprecated
     public CurrentLimits getCurrentLimits(final Side side) {
         return getDelegate().getCurrentLimits(side);
     }
 
     @Override
+    @Deprecated
     public CurrentLimits getCurrentLimits1() {
         return getDelegate().getCurrentLimits1();
     }
```
mathbagu, line 64: Remove these methods?

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/DanglingLineAdapter.java
```diff
@@ -92,12 +92,8 @@ public String getUcteXnodeCode() {
     }
 
     @Override
+    @Deprecated
     public CurrentLimits getCurrentLimits() {
         return getDelegate().getCurrentLimits();
     }
```
mathbagu, line 98: Remove this method

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/DanglingLineAdapter.java
```diff
@@ -92,12 +92,8 @@ public String getUcteXnodeCode() {
...
-
-    @Override
-    public CurrentLimitsAdder newCurrentLimits() {
-        return getDelegate().newCurrentLimits();
-    }
```
mathbagu, line 102: Why this method is removed?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Branch.java
MioRtia: Good remark... Not sure but I think for now, it will only be for current limits, I will keep it at that.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/LoadingLimits.java
MioRtia: Okay, I agree, is `hasOverloadingProtection()` an okay name?
Also, I am not sure limits other than current limits can have protection automaton, should this be specific to CurrentLimits.TL?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DanglingLine.java
MioRtia: Yes, this is it.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Switch.java
MioRtia: We don't have "side" object per see, but we discriminate between node1 and node2 (or bus1 and bus2) of a switch

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Injection.java
MioRtia: It is only to prevent breaking change (error in implementation - for example in persistent IIDM)

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Branch.java
MioRtia: apparentPower is voltage*current so it should be fine (ratedS is manufacturer information)

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Injection.java
MioRtia: Same answer as above

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/OperationalLimits.java
MioRtia: It is mostly to prevent breaking changes. Is it better in your opinion to throw an exception?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractBranch.java
MioRtia: Yes, for topological merge in `NetworkImpl`

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ThreeWindingsTransformerImpl.java
MioRtia: It is only set in `setTransformer()` i.e. when the transformer is created which is only **after** the legs are created, which explains why it isn't final.

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/DanglingLineAdapter.java
MioRtia: I deplaced it in `AbstractInjectionAdapter`

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Branch.java
MioRtia: The issue is: what do we do if a user passes a parent class of two different kinds of limits? (for example, LoadingLimits.class)

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Branch.java
MioRtia: Add snipets of code in javadoc to specify it

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Branch.java
MioRtia: Same

MioRtia:
Duplication percentage bothers me but I'm not sure how to reduce it since for example `Leg` and `Injection` do not have a common parent...

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -385,8 +392,9 @@ public static void checkConnected(Validable validable, Boolean connected) {
     }
 
     public static void checkPermanentLimit(Validable validable, double permanentLimit) {
+        // TODO: if (Double.isNaN(permanentLimit) || permanentLimit <= 0) {
```
mathbagu, line 395: What about this TODO?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ApparentPowerLimitsImpl.java
mathbagu: It seems you liked my comment, but you didn't take it into account

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
```diff
@@ -995,8 +994,8 @@ private void replaceDanglingLineByLine(List<MergedLine> lines, Multimap<Boundary
                 la.setNode2(mergedLine.node2);
             }
             TieLineImpl l = la.add();
-            l.setCurrentLimits(Side.ONE, (CurrentLimitsImpl) mergedLine.limits1);
-            l.setCurrentLimits(Side.TWO, (CurrentLimitsImpl) mergedLine.limits2);
+            l.getLimitsHolder1().setOperationalLimits(LimitType.CURRENT, mergedLine.limits1);
+            l.getLimitsHolder2().setOperationalLimits(LimitType.CURRENT, mergedLine.limits2);
```
mathbagu, line 998: I would hide the getLimitsHolder from the API, because it's not really user friendly

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VoltageLimitsImpl.java
```diff
@@ -0,0 +1,49 @@
...
+        return lowVoltage;
+    }
+
+    @Override
+    public VoltageLimits setLowVoltage(double lowVoltage) {
+        ValidationUtil.checkDefinedVoltageLimits(owner, lowVoltage, highVoltage);
+        this.lowVoltage = lowVoltage;
+        return this;
+    }
+}
```
mathbagu, line 49: Should we notify the listeners when setters are called. Same remark everywhere. I think it was done in the previous version
The `add` and `remove` method should also notify

mathbagu (commented):
I have exactly the same doubt than few months ago when I review this PR the first time. I recently see a CGMES training in video, in which the trainer said that none all OperatingLimits are relevant for any equipments. From my point of view, the design you propose is too generic. I don't know where to find the information, maybe in the OCL?

Another remark is about the API: maybe we should keep user-friendly method to get the different operating limits by their name (getCurrentLimits1() looks better than getOperationalLimits1(LimitType.CURRENT_LIMIT, CurrentLimits.class)). The implementation can however rely on a generic method that takes a type and a class.

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractVoltageLimitsTest.java
```diff
@@ -0,0 +1,46 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+
+/**
+ * @author Miora Ralambotiana <miora.ralambotiana at rte-france.com>
+ */
+public abstract class AbstractVoltageLimitsTest {
+
+    private static Network createNetwork() {
```
annetill, line 24: Finally we allow voltage limits on equipments ? I thought we decided to limit voltage limits to buses and bus bar sections. I am not sure to be able to simulate that level of details in the open load flow for example.

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/ThreeWindingsTransformerAdapter.java
```diff
@@ -95,6 +100,36 @@ public CurrentLimitsAdder newCurrentLimits() {
...
+            return getDelegate().getActivePowerLimits();
+        }
+
+        @Override
+        public ActivePowerLimitsAdder newActivePowerLimits() {
+            return getDelegate().newActivePowerLimits();
+        }
+
+        @Override
+        public VoltageLimits getVoltageLimits() {
```
annetill, line 124: I am not sure to understand how voltage limits work indeed. You associate it to the terminal connected at network side ? I imagine that it is similar for the other operational limits.

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/SwitchAdapter.java
```diff
@@ -50,4 +51,49 @@ public boolean isRetained() {
...
+        return getDelegate().getApparentPowerLimits();
+    }
+
+    @Override
+    public ApparentPowerLimitsAdder newApparentPowerLimits() {
+        return getDelegate().newApparentPowerLimits();
+    }
+
+    @Override
+    public VoltageLimits getVoltageLimits() {
```
annetill, line 91: Same remark about voltage limits.

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/SwitchAdapter.java
```diff
@@ -50,4 +51,49 @@ public boolean isRetained() {
     public void setRetained(final boolean retained) {
         getDelegate().setRetained(retained);
     }
+
+    @Override
+    public List<OperationalLimits> getOperationalLimits() {
```
annetill, line 56: I thought that we decide to ignore operational limits for switches ?

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/AbstractInjectionAdapter.java
```diff
@@ -22,4 +23,49 @@ protected AbstractInjectionAdapter(I delegate, MergingViewIndex index) {
...
+        return getDelegate().getActivePowerLimits();
+    }
+
+    @Override
+    public ActivePowerLimitsAdder newActivePowerLimits() {
+        return getDelegate().newActivePowerLimits();
+    }
+
+    @Override
+    public VoltageLimits getVoltageLimits() {
```
annetill, line 63: Same remark, to complex to deal with that for me.

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/AbstractBranchAdapter.java
```diff
@@ -77,6 +111,36 @@ public CurrentLimitsAdder newCurrentLimits2() {
...
+        return getDelegate().newActivePowerLimits2();
+    }
+
+    @Override
+    public VoltageLimits getVoltageLimits2() {
+        return getDelegate().getVoltageLimits2();
+    }
+
+    @Override
+    public VoltageLimitsAdder newVoltageLimits2() {
```
annetill, line 140: Same remark.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/SwitchImpl.java
```diff
@@ -137,4 +139,48 @@ protected String getTypeDescription() {
         return "Switch";
     }
 
+    @Override
```
annetill, line 142: For me, a switch does not have operational limits. It is going to increase considerably our network for nothing because secrutiy analysis are run on bus/breaker view.

annetill (changes requested):
I have made a partial review because I think that we have 2 issues:
- Switches should not be operational limits holder because it is quite impossible to deal with that inside a simulator. Most simulators work from bus/breaker or bus/view topology. So many limits will be ignore and store in IIDM for nothing ;
- High/low voltage limits could not be hold by equipements. The power flow engines can check voltage on buses (from bus/breaker topology or bus/view topology). It means that from node/breaker topology, we have to find a place that spreads the information to buses in the other levels of topology. Maybe node is to detailed, voltage level not enough detailed, maybe bus bar section.
What do you think about that ?

annetill (commented):
I have tried on open loadflow and it compiles without any changes. I understand for branch limits but I amn not sure to understand why from a voltage level, we still have high and low voltage limits. We should at least add a `deprecated` label ?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Branch.java
```diff
@@ -140,16 +143,81 @@
...
+
+    default ApparentPowerLimits getApparentPowerLimits1() {
+        return null;
+    }
+
     CurrentLimitsAdder newCurrentLimits1();
 
+    default ActivePowerLimitsAdder newActivePowerLimits1() {
+        throw new UnsupportedOperationException();
```
mathbagu, line 163: Still the same discussion: I suppose you will use this new method (and the others) in the converters. Suppose you use a different implementation of IIDM, this will crash the import.

I don't know if we should care about it, or if we have to change our strategy for defaulted methods, like providing a mock that does nothing but ensure that the importers work with not updated implementations

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Bus.java
```diff
@@ -17,7 +17,7 @@
  *
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
  */
-public interface Bus extends Identifiable<Bus> {
+public interface Bus extends Identifiable<Bus>, VoltageLimitsHolder {
```
mathbagu, line 20: I saw that the branch doesn't inherit from `FlowsLimitsHolder`. Do I miss something?

To be checked: how calculated bus manage limits.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/CalculatedBusImpl.java
```diff
@@ -188,6 +188,23 @@ public Component getSynchronousComponent() {
         return terminalRef == null ? null : scm.getComponent(terminalRef.getSynchronousComponentNumber());
     }
 
+    @Override
+    public VoltageLimits getVoltageLimits() {
+        return getConnectableStream(BusbarSection.class)
+                .map(VoltageLimitsHolder::getVoltageLimits)
+                .filter(Objects::nonNull)
+                .findFirst() // we suppose that all voltage limits in a same calculated bus are equals
```
mathbagu, line 196: I'm pretty sure it won't be the case: I propose to sort the voltageLimits and return the lowest to avoid missing violations

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/CalculatedBusImpl.java
```diff
@@ -188,6 +188,23 @@ public Component getSynchronousComponent() {
...
+                .orElse(null);
+    }
+
+    @Override
+    public VoltageLimitsAdder newVoltageLimits() {
+        return getConnectableStream(BusbarSection.class)
+                .findFirst()
+                .map(VoltageLimitsHolder::newVoltageLimits)
+                .orElseThrow(() -> new ValidationException(this, "No busbarsection is associated to this calculated bus: no voltage limit can be created."));
+    }
```
mathbagu, line 206: I would throw an exception in all case: we could probably accept that limits have to be created from "configured" objects (BusbarSections and ConfiguredBus). It's arguable.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/MergedBus.java
```diff
@@ -198,6 +199,23 @@ public Component getSynchronousComponent() {
...
+                .orElse(null);
+    }
+
+    @Override
+    public VoltageLimitsAdder newVoltageLimits() {
+        return buses.stream()
+                .findFirst()
+                .map(VoltageLimitsHolder::newVoltageLimits)
+                .orElseThrow(() -> new AssertionError("Should not happen")); // a merged bus should at least contain one configured bus
+    }
```
mathbagu, line 217: See CalculatedBusImpl remarks

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/OperationalLimitsHolderImpl.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @Override
+    public void setOperationalLimits(LimitType limitType, OperationalLimits operationalLimits) {
+        OperationalLimits oldValue = this.operationalLimits.put(limitType, operationalLimits);
+        identifiable.getNetwork().getListeners().notifyUpdate(identifiable, attributeName + "_" + limitType, oldValue, operationalLimits);
+    }
+
+    List<OperationalLimits> getOperationalLimits() {
+        return Collections.unmodifiableList(new ArrayList<>(operationalLimits.values()));
+    }
```
mathbagu, line 35: Do you really need to return a List, or a Collection could be OK?

action/action-dsl/src/main/java/com/powsybl/action/dsl/ast/ActionExpressionEvaluator.java
```diff
@@ -93,7 +93,7 @@ private Branch getBranch() {
         /**
          * TODO: to move to IIDM
          */
-        private static double getPermanentLimit(Branch branch, Branch.Side side) {
+        private static double getPermanentLimit(Branch<?> branch, Branch.Side side) {
```
mathbagu, line 96: To be reverted?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VoltageLimitsAdderImpl.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    @Override
+    public VoltageLimitsAdder setLowVoltage(double lowVoltage) {
+        this.lowVoltage = lowVoltage;
+        return this;
+    }
+
+    @Override
+    public VoltageLimits add() {
```
mathbagu, line 37: To be discuss: is it better to return the VoltageLimits or to return the owner to be able to create additional limits?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VoltageLimitsImpl.java
mathbagu: @MioRtia: Could you please considerer this remark

mathbagu (changes requested):
There are files that contain changed unrelated to this PR (import order or wildcard). Maybe you have to revert them.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Branch.java
MioRtia: To be honest, not sure what it is best for whole new objects... I worry that adding mock objects will only make the API more confusing

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/OperationalLimitsHolderImpl.java
MioRtia: A collection would be okay, I will modify this

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VoltageLimitsImpl.java
MioRtia: Ah, yes we can. Will be modified.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Bus.java
MioRtia: What I made was: when setting a calculated bus (NB to BB), it is actually the busbarsections underneath which are set (an error is thrown if there is none). For a merged bus (BusBreaker to Bus), it is actually the BB buses underneath which are set.

MioRtia: Changed a bit the mecanism in the new commits

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/OperationalLimitsHolderImpl.java
MioRtia: Just to be sure: do you mean in the API or in the implementation?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VoltageLimitsAdderImpl.java
MioRtia: Both approaches are good for me. In favor of returning the limits: this is what we are doing with reactive limits so it would be consistent regarding the API. Furthermore, not sure it makes sense for VoltageLimits for example: objects supporting voltage limits do not support other kinds of limits.

MioRtia:
> I have tried on open loadflow and it compiles without any changes. I understand for branch limits but I amn not sure to understand why from a voltage level, we still have high and low voltage limits. We should at least add a `deprecated` label ?

@annetill Are you sure ? For NB networks with voltage levels without busbar sections, we will lose this information.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Branch.java
mathbagu: The mock is not the good solution: what I mean is implementing the new methods with a throw will hide the issue. You don't break the compilation but it will fail at runtime that it's worst.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Bus.java
MioRtia: I've thought a bit about your proposition and I don't think it is relevant to make `Branch` inherit from `FlowsLimitHolder`: indeed, in any cases, it hasn't any common methods with flow limit holder without sides so if we needed to iterate on `FlowsLimitHolder`, we would at the end cast it as `Branch` to use it so it is a bit pointless...

MioRtia:
@mathbagu @annetill Some unit tests are missing but can you look at the rest of the code to see if it is okay for you ?

annetill (commented):
Just two comments: 
- I have run the open loadflow on that branch without any change and two unit tests failed. It was because high and low voltage limits on the voltage level were not really added even if all was okay during the compilation. It was not evident for me the fixes I had to make (using voltage limits adder).
- Before I was doing simply: 
 `bus.getVoltageLevel().getLowVoltageLimit()`
and now I have to do:
`bus.getVoltageLimits() != null ? bus.getVoltageLimits().getLowVoltage() : Double.NaN`

annetill:
@MioRtia : we need indeed also to be able to modify the current limits (at least). For the moment, the need is to introduce a multipliying factor for permanent limit and others for temporary limits (one for all or one by temporary limits, I don't know for the moment). These facors could depend from the variant, it is a question. What is the best way to handle that for you : through the model or through an extension ? If it is trhough the model, maybe we have to handle that in that PR. What do you think ?

MioRtia:
> Just two comments:
>    I have run the open loadflow on that branch without any change and two unit tests failed. It was because high and low voltage limits on the voltage level were not really added even if all was okay during the compilation. It was not evident for me the fixes I had to make (using voltage limits adder).
    Before I was doing simply:
    bus.getVoltageLevel().getLowVoltageLimit()
    and now I have to do:
    bus.getVoltageLimits() != null ? bus.getVoltageLimits().getLowVoltage() : Double.NaN

Yes, two options for me:
- completely delete `getLow/HighVoltageLimit()` methods
- update the wiki once this is merged

I was going to do the second one, is it sufficient for you?

> @MioRtia : we need indeed also to be able to modify the current limits (at least). For the moment, the need is to introduce a multipliying factor for permanent limit and others for temporary limits (one for all or one by temporary limits, I don't know for the moment). These facors could depend from the variant, it is a question. What is the best way to handle that for you : through the model or through an extension ? If it is trhough the model, maybe we have to handle that in that PR. What do you think ?

The permanent limit can already be modified so what you are saying is that we need a multiplying factor for all temporary limits ? (I don't really understand the "for one temporary limit": wouldn't it be easier to simply make temporary limit modifyable if this is the use case and let the user decide of the value?) In your use case, do we need the original limit value(s) ? If yes, it should be an extension in my opinion. If the original limit value(s) can be lost, we can put it there and simply make the limit values modifyable as I said above.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/CalculatedBus.java
```diff
@@ -6,12 +6,13 @@
...
 package com.powsybl.iidm.network.impl;
 
+import com.powsybl.iidm.network.Validable;
+
 /**
  *
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
  */
-interface CalculatedBus extends BusExt {
+interface CalculatedBus extends BusExt, Validable {
```
annetill, line 15: Do you still need this modification?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ThreeWindingsTransformer.java
```diff
@@ -210,7 +210,7 @@
      *     </tbody>
      * </table>
      */
-    public interface Leg extends RatioTapChangerHolder, PhaseTapChangerHolder {
+    public interface Leg extends RatioTapChangerHolder, PhaseTapChangerHolder, FlowsLimitsHolder {
```
annetill, line 213: I am not sure to get what you mean here with that `FlowLimitsHolder`: for you, only current limits can be associated to a three winding transformers ?

annetill: It is just the default behaviour?

annetill (approved):

geofjamg (commented):
What is the purpose of setOverloadingProtection?

MioRtia:
> What is the purpose of setOverloadingProtection?

It is just a renaming of `setFictitious` but with @mathbagu we thought it was more meaningful (if it is `false`, no overloading protection is triggered).

geofjamg:
> > What is the purpose of setOverloadingProtection?
> 
> It is just a renaming of `setFictitious` but with @mathbagu we thought it was more meaningful (if it is `false`, no overloading protection is triggered).

I really don't feel it as more meaningful...
"overloadingProtection" doesn't mean anything, something is missing like 'overloadingProtectionActive'

annetill:
> 
> 
> > > What is the purpose of setOverloadingProtection?
> > 
> > 
> > It is just a renaming of `setFictitious` but with @mathbagu we thought it was more meaningful (if it is `false`, no overloading protection is triggered).
> 
> I really don't feel it as more meaningful...
> "overloadingProtection" doesn't mean anything, something is missing like 'overloadingProtectionActive' or 'overloadingProtectionEnabled'

Is it really useful to break the API for that ? I understand that for current limits already implemented, this evolution will be transparent.

geofjamg:
> Is it really useful to break the API for that ? I understand that for current limits already implemented, this evolution will be transparent.

Agree, it does not worth it

MioRtia:
> > Is it really useful to break the API for that ? I understand that for current limits already implemented, this evolution will be transparent.
> 
> Agree, it does not worth it, I would keep fictitious.

@annetill @geofjamg Done

PR #1360, opened 2020-06-19
Current limits with missing permanent limit are not ignored anymore in CGMES import but fixed with permanent limit equal to lowest TATL

MioRtia:

Between bug and feature...

When a current limits set in CGMES does not contain a permanent limit, the whole set is ignored.

When a current limits set in CGMES does not contain a permanent limit, the permanent limit of the set is considered infinite.

No


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CurrentLimitsMapping.java
```diff
@@ -32,10 +32,12 @@ public CurrentLimitsAdder getCurrentLimitsAdder(String id, Supplier<CurrentLimit
     void addAll() {
         for (Map.Entry<String, CurrentLimitsAdder> entry : adders.entrySet()) {
             if (Double.isNaN(entry.getValue().getPermanentLimit())) {
-                context.ignored("Operational Limit Set of " + entry.getKey(), "An operational limit set must at least contain one value for permanent limit.");
-            } else {
-                entry.getValue().add();
+                context.fixed("Operational Limit Set of " + entry.getKey(),
```
annetill, line 35: The PATL seems to be mandatory in CGMES for all transformers and at both ends for a AC line. So I am not sure that it is safe just to replace them by infinite values. What do you think ?

MioRtia: In theory, this is true, but it seems that in French networks, current TATL are assigned without PATL...

MioRtia: We can take the highest value of TATL as PATL though? It may be more correct that putting an infinite PATL

annetill: No, I was more thinking about a warning.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CurrentLimitsMapping.java
```diff
@@ -32,10 +32,12 @@ public CurrentLimitsAdder getCurrentLimitsAdder(String id, Supplier<CurrentLimit
...
+                        "An operational limit set without permanent limit is considered with infinite limit.",
+                        Double.NaN, Double.MAX_VALUE);
+                entry.getValue().setPermanentLimit(Double.MAX_VALUE);
```
mathbagu, line 38: Are you sure this is a good workaround, specially if the CurrentLimits has TATL: The PATL has to be lesser or equal to the minimal TATL. Putting `MAX_VALUE` whould lead to never have violations on that equipment. From my point of view, if there are no TATL, we should not add a CurrentLimit for this object, and if there are TATL, we should pick the lowest one.

MioRtia: > The PATL has to be lesser or equal to the minimal TATL

Yes, I was wondering that as well... For most current limits in the French networks, the PATL seems to be equals to the minimal TATL. Maybe we should do that? But I'm not sure about putting an arbitrary value, I would prefer to put an undefined/infinite value (even if it has no electrical sense).

mathbagu: This is the definition of PATL and TATL, but maybe it's not strict. Honnestly, instead of setting an infinite value, don't create a currentLimits: the behaviour would be the same, and you don't create non sense data :)

MioRtia: I modified to take the lowest available TATL when no PATL is defined (when there is no TATL and no PATL, no current limit is created). Is it okay for you?

PR #1362, opened 2020-06-22
[WIP] Issue 83 id terminal ref optional

### LINKED ISSUE #83: The id attribute of a terminalRef is optional
The id attribute of a terminalRef will be optional. If not set, the transformer id will be used.
**IIDM v1.0**
```
<iidm:ratioTapChanger lowTapPosition="0" tapPosition="1" regulating="true" loadTapChangingCapabilities="true" targetV="158.0">
     <iidm:terminalRef id="NHV2_NLOAD" side="TWO"/>
```

**IIDM v1.1**
```
<iidm:ratioTapChanger lowTapPosition="0" tapPosition="1" regulating="true" loadTapChangingCapabilities="true" targetV="158.0">
     <iidm:terminalRef side="TWO"/>
```

jlabous:

fix #83

### DIFF
--- a/commons/src/main/java/com/powsybl/commons/xml/XmlUtil.java
+++ b/commons/src/main/java/com/powsybl/commons/xml/XmlUtil.java
@@ -6,6 +6,8 @@
  */
 package com.powsybl.commons.xml;
 
+import com.powsybl.commons.exceptions.UncheckedXmlStreamException;
+
 import javax.xml.stream.XMLStreamConstants;
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamReader;
@@ -122,6 +124,14 @@ public final class XmlUtil {
     public static void writeOptionalString(String name, String value, XMLStreamWriter writer) throws XMLStreamException {
         if (value != null) {
             writer.writeAttribute(name, value);
+        }
+    }
+
+    public static void writeStringAttribute(String attributeName, String value, XMLStreamWriter writer) {
+        try {
+            writer.writeAttribute(attributeName, value);
+        } catch (XMLStreamException e) {
+            throw new UncheckedXmlStreamException(e);
         }
     }
 
--- a/commons/src/main/java/com/powsybl/commons/xml/XmlUtil.java
+++ b/commons/src/main/java/com/powsybl/commons/xml/XmlUtil.java
@@ -127,7 +127,7 @@ public final class XmlUtil {
         }
     }
 
-    public static void writeStringAttribute(String attributeName, String value, XMLStreamWriter writer) {
+    public static void writeString(String attributeName, String value, XMLStreamWriter writer) {
         try {
             writer.writeAttribute(attributeName, value);
         } catch (XMLStreamException e) {
--- a/iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractTransformerXml.java
+++ b/iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractTransformerXml.java
@@ -14,6 +14,7 @@ import com.powsybl.iidm.xml.util.IidmXmlUtil;
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamWriter;
 
+import java.util.Optional;
 import java.util.function.BiConsumer;
 import java.util.function.DoubleConsumer;
 
@@ -105,7 +106,7 @@ abstract class AbstractTransformerXml<T extends Connectable, A extends Identifia
         context.getWriter().writeEndElement();
     }
 
-    protected static void readRatioTapChanger(String elementName, RatioTapChangerAdder adder, Terminal terminal, String transformerId, NetworkXmlReaderContext context) throws XMLStreamException {
+    protected static void readRatioTapChanger(Connectable owner, String elementName, RatioTapChangerAdder adder, Terminal terminal, NetworkXmlReaderContext context) throws XMLStreamException {
         int lowTapPosition = XmlUtil.readIntAttribute(context.getReader(), ATTR_LOW_TAP_POSITION);
         int tapPosition = XmlUtil.readIntAttribute(context.getReader(), ATTR_TAP_POSITION);
         boolean regulating = XmlUtil.readOptionalBoolAttribute(context.getReader(), ATTR_REGULATING, false);
@@ -125,12 +126,8 @@ abstract class AbstractTransformerXml<T extends Connectable, A extends Identifia
             switch (context.getReader().getLocalName()) {
                 case ELEM_TERMINAL_REF:
                     readTerminalRef(context, hasTerminalRef, (id, side) -> {
-                        if (id != null) {
-                            adder.setRegulationTerminal(TerminalRefXml.readTerminalRef(terminal.getVoltageLevel().getSubstation().getNetwork(), id, side));
-                        } else {
-                            adder.setRegulationTerminal(TerminalRefXml.readTerminalRef(terminal.getVoltageLevel().getSubstation().getNetwork(), transformerId, side));
-                        }
-                        adder.add();
+                        adder.setRegulationTerminal(TerminalRefXml
+                                .readTerminalRef(terminal.getVoltageLevel().getSubstation().getNetwork(), Optional.ofNullable(id).orElse(owner.getId()), side)).add();
                     });
                     break;
 
@@ -154,11 +151,11 @@ abstract class AbstractTransformerXml<T extends Connectable, A extends Identifia
     }
 
     protected static void readRatioTapChanger(TwoWindingsTransformer twt, NetworkXmlReaderContext context) throws XMLStreamException {
-        readRatioTapChanger(RATIO_TAP_CHANGER, twt.newRatioTapChanger(), twt.getTerminal1(), twt.getId(), context);
+        readRatioTapChanger(twt, RATIO_TAP_CHANGER, twt.newRatioTapChanger(), twt.getTerminal1(), context);
     }
 
-    protected static void readRatioTapChanger(int leg, ThreeWindingsTransformer.Leg twl, String transformerId, NetworkXmlReaderContext context) throws XMLStreamException {
-        readRatioTapChanger(RATIO_TAP_CHANGER + leg, twl.newRatioTapChanger(), twl.getTerminal(), transformerId, context);
+    protected static void readRatioTapChanger(Connectable owner, int leg, ThreeWindingsTransformer.Leg twl, NetworkXmlReaderContext context) throws XMLStreamException {
+        readRatioTapChanger(owner, RATIO_TAP_CHANGER + leg, twl.newRatioTapChanger(), twl.getTerminal(), context);
     }
 
     protected static void writePhaseTapChanger(Connectable owner, String name, PhaseTapChanger ptc, NetworkXmlWriterContext context) throws XMLStreamException {
@@ -183,7 +180,7 @@ abstract class AbstractTransformerXml<T extends Connectable, A extends Identifia
         context.getWriter().writeEndElement();
     }
 
-    protected static void readPhaseTapChanger(String name, PhaseTapChangerAdder adder, Terminal terminal, String transformerId, NetworkXmlReaderContext context) throws XMLStreamException {
+    protected static void readPhaseTapChanger(Connectable owner, String name, PhaseTapChangerAdder adder, Terminal terminal, NetworkXmlReaderContext context) throws XMLStreamException {
         int lowTapPosition = XmlUtil.readIntAttribute(context.getReader(), ATTR_LOW_TAP_POSITION);
         int tapPosition = XmlUtil.readIntAttribute(context.getReader(), ATTR_TAP_POSITION);
         boolean regulating = XmlUtil.readOptionalBoolAttribute(context.getReader(), ATTR_REGULATING, false);
@@ -202,12 +199,8 @@ abstract class AbstractTransformerXml<T extends Connectable, A extends Identifia
             switch (context.getReader().getLocalName()) {
                 case ELEM_TERMINAL_REF:
                     readTerminalRef(context, hasTerminalRef, (id, side) -> {
-                        if (id != null) {
-                            adder.setRegulationTerminal(TerminalRefXml.readTerminalRef(terminal.getVoltageLevel().getSubstation().getNetwork(), id, side));
-                        } else {
-                            adder.setRegulationTerminal(TerminalRefXml.readTerminalRef(terminal.getVoltageLevel().getSubstation().getNetwork(), transformerId, side));
-                        }
-                        adder.add();
+                        adder.setRegulationTerminal(TerminalRefXml
+                                .readTerminalRef(terminal.getVoltageLevel().getSubstation().getNetwork(), Optional.ofNullable(id).orElse(owner.getId()), side)).add();
                     });
                     break;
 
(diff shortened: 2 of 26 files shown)


mathbagu:
@MioRtia would confirm, but I think you should rebase your branch onto `evolution_xiidm/1.3`. I tried to change the base branch, but as you merged the master branch in it, there are a lot of unexpected commits that make the review difficult.

jlabous:
> 
> 
> @MioRtia would confirm, but I think you should rebase your branch onto `evolution_xiidm/1.3`. I tried to change the base branch, but as you merged the master branch in it, there are a lot of unexpected commits that make the review difficult.

My branch was already based on 'evolution_xiidm/1.3'
I revert the merge

iidm/iidm-xml-converter/src/test/java/com/powsybl/iidm/xml/TerminalMockXmlSerializer.java
```diff
@@ -66,7 +66,8 @@ public void write(TerminalMockExt extension, XmlWriterContext context) throws XM
         NetworkXmlWriterContext networkContext = (NetworkXmlWriterContext) context;
         String extensionVersion = networkContext.getOptions().getExtensionVersion(getExtensionName())
                 .orElseGet(() -> getVersion(networkContext.getVersion()));
-        TerminalRefXml.writeTerminalRef(extension.getTerminal(), networkContext, getNamespaceUri(extensionVersion), "terminal", context.getExtensionsWriter());
+        TerminalRefXml.writeTerminalRef(extension.getTerminal(), networkContext, getNamespaceUri(extensionVersion),
+                "terminal", "", context.getExtensionsWriter());
```
MioRtia, line 70: I think it would be clearer to use a method without an `id` parameter (`TerminalRefXml.writeTerminalRef(extension.getTerminal(), networkContext, getNamespaceUri(extensionVersion),
                "terminal", context.getExtensionsWriter());`) which calls `TerminalRefXml.writeTerminalRef(extension.getTerminal(), networkContext, getNamespaceUri(extensionVersion),
                "terminal", null, context.getExtensionsWriter());` and then putting the condition in the method:
```java
            if (id != null && !id.equals(c.getId())) {
                writeId[0] = true;
            }
```

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/TerminalRefXml.java
```diff
@@ -18,21 +19,34 @@
...
+        final boolean[] writeId = {false};
+        IidmXmlUtil.runUntilMaximumVersion(IidmXmlVersion.V_1_2, context, () -> writeId[0] = true);
+        IidmXmlUtil.runFromMinimumVersion(IidmXmlVersion.V_1_3, context, () -> {
+            if (id == null || !id.equals(c.getId())) {
+                writeId[0] = true;
+            }
+        });
+        if (writeId[0]) {
+            writer.writeAttribute("id", context.getAnonymizer().anonymizeString(c.getId()));
+        }
```
mathbagu, line 49: This code looks complicated for nothing:
```
IidmXmlUtil.runUntilMaximumVersion(IidmXmlVersion.V_1_2, context, () -> {
    writer.writeAttribute("id", context.getAnonymizer().anonymizeString(c.getId()));
}
IidmXmlUtil.runFromMinimumVersion(IidmXmlVersion.V_1_3, context, () -> {
    if (id == null || !id.equals(c.getId())) {
        writer.writeAttribute("id", context.getAnonymizer().anonymizeString(c.getId()));
    }
});
```

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/TerminalRefXml.java
```diff
@@ -18,21 +19,34 @@
...
-        writeTerminalRef(t, context, namespace, elementName, context.getWriter());
+    public static void writeTerminalRef(Terminal t, NetworkXmlWriterContext context, String namespace, String elementName, String id) throws XMLStreamException {
+        writeTerminalRef(t, context, namespace, elementName, id, context.getWriter());
     }
 
     public static void writeTerminalRef(Terminal t, NetworkXmlWriterContext context, String namespace, String elementName, XMLStreamWriter writer) throws XMLStreamException {
+        writeTerminalRef(t, context, namespace, elementName, null, writer);
+    }
+
+    public static void writeTerminalRef(Terminal t, NetworkXmlWriterContext context, String namespace, String elementName, String id, XMLStreamWriter writer) throws XMLStreamException {
```
mathbagu, line 34: From my point of view, the read/write method should take the Identifiable object instead of an idea:
```
 public static void writeTerminalRef(Connectable owner, Terminal t, NetworkXmlWriterContext context, String namespace, String elementName, XMLStreamWriter writer) throws XMLStreamException;
```

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/TerminalRefXml.java
jlabous: I made this to avoid 2 try-catch

mathbagu: @MioRtia What do you think to create utility functions to write attributes that throws only unchecked exception. Or to introduce a new utility function like the one you created for integer  in IidmXmlUtil (writeIntAttributeUntilMaximumVersion)

mathbagu: @jlabous We propose to add a writeStringAttribute in XmlUtil class that catch XmlStreamException and wrap them into UncheckedXmlStreamException, and to use this method in the lambdas.
Is it ok for you?

jlabous: Ok I'll go on this

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/TerminalRefXml.java
```diff
@@ -19,34 +20,43 @@
  */
 public final class TerminalRefXml {
 
-    public static void writeTerminalRef(Terminal t, NetworkXmlWriterContext context, String elementName, String id) throws XMLStreamException {
-        writeTerminalRef(t, context, context.getVersion().getNamespaceURI(), elementName, id);
+    @Deprecated
```
MioRtia, line 23: Add a comment :
```java
/**
 * @deprecated Use {@link #writeTerminalRef(Connectable, Terminal, NetworXmlWriterContext, String)} instead
 */
```

Same for other deprecated methods

MioRtia:
@mathbagu Is this PR okay for you?

commons/src/main/java/com/powsybl/commons/xml/XmlUtil.java
```diff
@@ -125,6 +127,14 @@ public static void writeOptionalString(String name, String value, XMLStreamWrite
         }
     }
 
+    public static void writeStringAttribute(String attributeName, String value, XMLStreamWriter writer) {
```
mathbagu, line 130: To be consistent, this method should be rename to writeString (see above)

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractTransformerXml.java
```diff
@@ -105,7 +105,7 @@ protected static void writeRatioTapChanger(String name, RatioTapChanger rtc, Net
         context.getWriter().writeEndElement();
     }
 
-    protected static void readRatioTapChanger(String elementName, RatioTapChangerAdder adder, Terminal terminal, NetworkXmlReaderContext context) throws XMLStreamException {
+    protected static void readRatioTapChanger(String elementName, RatioTapChangerAdder adder, Terminal terminal, String transformerId, NetworkXmlReaderContext context) throws XMLStreamException {
```
mathbagu, line 108: Why can't we use the owner their too? Same question for the whole file

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/GeneratorXml.java
```diff
@@ -92,7 +92,11 @@ protected void readSubElements(Generator g, NetworkXmlReaderContext context) thr
                 case "regulatingTerminal":
                     String id = context.getAnonymizer().deanonymizeString(context.getReader().getAttributeValue(null, "id"));
                     String side = context.getReader().getAttributeValue(null, "side");
-                    context.getEndTasks().add(() -> g.setRegulatingTerminal(TerminalRefXml.readTerminalRef(g.getTerminal().getVoltageLevel().getSubstation().getNetwork(), id, side)));
+                    if (id != null) {
+                        context.getEndTasks().add(() -> g.setRegulatingTerminal(TerminalRefXml.readTerminalRef(g.getTerminal().getVoltageLevel().getSubstation().getNetwork(), id, side)));
+                    } else {
+                        context.getEndTasks().add(() -> g.setRegulatingTerminal(TerminalRefXml.readTerminalRef(g.getTerminal().getVoltageLevel().getSubstation().getNetwork(), g.getId(), side)));
+                    }
```
mathbagu, line 99: I think we can write this code differently: why don't we pass a default value to the `getAttributeValue(null, "id", g.getId())` and remove this `if/else`?

I propose to make this change everywhere

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/TerminalRefXml.java
```diff
@@ -18,21 +20,52 @@
...
         if (!context.getFilter().test(c)) {
             throw new PowsyblException("Oups, terminal ref point to a filtered equipment " + c.getId());
         }
         writer.writeEmptyElement(namespace, elementName);
-        writer.writeAttribute("id", context.getAnonymizer().anonymizeString(c.getId()));
+        IidmXmlUtil.runUntilMaximumVersion(IidmXmlVersion.V_1_2, context, () ->
+                XmlUtil.writeStringAttribute("id", context.getAnonymizer().anonymizeString(c.getId()), writer)
+        );
+        IidmXmlUtil.runFromMinimumVersion(IidmXmlVersion.V_1_3, context, () -> {
+            if (!c.equals(owner)) {
```
mathbagu, line 65: Two connectable could be equals, but different: for example if there are in two different networks

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/TerminalRefXml.java
```diff
@@ -18,21 +20,52 @@
  */
 public final class TerminalRefXml {
 
+    /**
+     * @deprecated Use {@link #writeTerminalRef(Connectable, Terminal, NetworkXmlWriterContext, String)} instead
+     */
+    @Deprecated
     public static void writeTerminalRef(Terminal t, NetworkXmlWriterContext context, String elementName) throws XMLStreamException {
-        writeTerminalRef(t, context, context.getVersion().getNamespaceURI(), elementName);
+        writeTerminalRef(null, t, context, context.getVersion().getNamespaceURI(), elementName);
```
mathbagu, line 28: I don't know if passing null is OK, that could throw an exception? Is there a unit test to ensure that is OK?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/TerminalRefXml.java
jlabous: In this case, does it really matter ?
If they are equals, ids are same even if they are in two differents networks.

MioRtia: Yes, this is a bit reaching but I think you should do `c != owner` to prevent mistakes in any case.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/TerminalRefXml.java
MioRtia: Yes, re-reading the code I think this produces a nullpointerexception in the comparison with the connectable linked to the terminal. It should be solved by the modification suggested below. Add a unit test in any case.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/GeneratorXml.java
jlabous: Because of the deanonymize we will have an error if we define a new getAttributeValue.

Is it convenient like this:
String id = StringUtils.defaultIfEmpty(context.getAnonymizer().deanonymizeString(context.getReader().getAttributeValue(null, "id")), g.getId());

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/TerminalRefXml.java
jlabous: In IdentifiableExtensionXmlSerializerTest, some tests use TerminalmockXmlSerializer witch use this deprecated method

MioRtia:
@jlabous The branch `evolution_xiidm/1.3` has been rebased on master, could you please fix the conflicts? Thank you!

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/GeneratorXml.java
mathbagu:
```
String id = Optional.ofNullable(context.getAnonymizer().deanonymizeString(context.getReader().getAttributeValue(null, "id"))).orElse(g.id());
```

mathbagu:
Trying to remove the call to the deprecated method in TerminalRefXml, I found an unexpected behavior where it's possible to serialize the `terminalRef` element without any attributes. This is reproductible for local regulation for instance.

Also, this deprecated method is used in some extension and changing the behavior of these methods is a regression. We should have a method that continue to export as in XIIDM prior than v1.3, for extensions, and update only the code called by XIIDM exported?

MioRtia:
> Trying to remove the call to the deprecated method in TerminalRefXml, I found an unexpected behavior where it's possible to serialize the `terminalRef` element without any attributes. This is reproductible for local regulation for instance.
> 
> Also, this deprecated method is used in some extension and changing the behavior of these methods is a regression. We should have a method that continue to export as in XIIDM prior than v1.3, for extensions, and update only the code called by XIIDM exported?

jlabous:
> 
> 
> Trying to remove the call to the deprecated method in TerminalRefXml, I found an unexpected behavior where it's possible to serialize the `terminalRef` element without any attributes. This is reproductible for local regulation for instance.
> 
> Also, this deprecated method is used in some extension and changing the behavior of these methods is a regression. We should have a method that continue to export as in XIIDM prior than v1.3, for extensions, and update only the code called by XIIDM exported?

Do i have to export with an id if we don't have to export the attribute side ? Or don't write terminalRef if we don't have attributes to export ?


Deprecated methods always export like in XIIDM prior than v1.3, because we don't put an owner.
Do I have to add a mathod that only write TerminalRef prior than v1.3 ?

MioRtia:
This should be rebased on `evolution_xiidm/1.4` and the modifications will be on IIDM 1.4

PR #1366, opened 2020-06-23
Fix guava vulnerability

mathbagu:

N/A

Update

Current version of guava is 20.0, but this version contains a [vulnerability](https://github.com/powsybl/powsybl-core/network/alert/pom.xml/com.google.guava:guava/open)

We use the latest version of guava.

**Other information**:


mathbagu:
> Ok. To monitor in the next month with current 3.5.0-SNAPSHOT if no serious incompatibility issue come up. I know that Guava devs use to break API and because all projects depend on Guava, we can have unsolvable compatibility issue.

This is the message that DependencyBot reports to explain why they don't open a pull request. I saw that in powsybl-network-store, guava v28.2 is used. So I think you already validate that it works for our use case.

PR #1375, opened 2020-06-26
Compute star bus voltages of threeWindingsTransformers and boundary bus voltages of dangling lines

marqueslanauja:

Feature

Voltages were not calculated

Star bus voltage of `threeWindingsTransformers` has been calculated depending on connected legs
Boundary bus voltage of `danglingLines` has been calculated depending on  injection

**Other information**:


annetill:
@marqueslanauja Can you please open in WIP the PR on open loadflow side that uses this feature ? Because for the moment, the open loadflow uses two properties to get the angle and the voltage at star bus.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/DanglingLineData.java
```diff
@@ -0,0 +1,118 @@
...
+    double boundaryBusTheta;
+
+    public DanglingLineData(DanglingLine danglingLine) {
+
+        id = danglingLine.getId();
+        r = danglingLine.getR();
+        x = danglingLine.getX();
+        g1 = danglingLine.getG();
+        b1 = danglingLine.getB();
+        g2 = 0.0;
```
annetill, line 42: Do we miss the parameter ````twtSplitShuntAdmittance```` here ? It is a parameter for the loadflow but it could be used here.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/DanglingLineData.java
```diff
@@ -0,0 +1,118 @@
...
+        Complex v1 = new Complex(u1 * Math.cos(theta1), u1 * Math.sin(theta1));
+
+        Complex vBoundaryBus = new Complex(Double.NaN, Double.NaN);
+        if (p0 == 0.0 && q0 == 0.0) {
+            LinkData.BranchAdmittanceMatrix adm = LinkData.calculateBranchAdmittance(r, x, 1.0, 0.0, 1.0, 0.0, new Complex(g1, b1), new Complex(g2, b2));
+            vBoundaryBus = adm.y21.multiply(v1).negate().divide(adm.y22);
+        } else {
+
+            // Two buses Loadflow
+            double pBoundaryBus = -p0;
```
annetill, line 62: It is here an approximation ? As ````r```` and ````x```` are not equal to zero ?

marqueslanauja: No, the exact voltage at the boundary bus it is calculated. It is a two bus Loadflow and can be solved as it is a second degree equation.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/DanglingLineData.java
marqueslanauja: The `twtSplitShuntAdmittance` parameter is only for transformers. The shunt admittance is always split in lines, but we are not sure about dangling lines. We are checking if we should split or not.

annetill: Why do you say that the shunt admittance is always split for lines and dangling lines ? In IIDM, you can have what you want as B1, G1, B2 and G2 and it does not mean for me that it is split in 2 equal parts. So where do you have seen this in the code ? If you are right, I am not sure to understand why we leave the possibility to set B1, G1, B2 and G2. In a dangling line, as you have only B and G, you have the right to split it or not when running the powerflow. In the open loadflow, we split it for example.

marqueslanauja: When I said that the shunt admittance is always split for lines I was referring in the cgmes conversion.  In CGMES model the total shunt admittance is received and is split in the conversion process. I prefer the line model with G1, B1, G2, B2 attributes as it is more generic and allows a "clean convert process" from standard tools, like `Psse`, that use in line model the same four attributes. I should prefer also to have the four attributes in the `danglingLine` model, to be coherent with the `line` model and to eliminate the possibility of splitting when running the load flow.  There is only one `real model` so i prefer to have an` iidm model` that allows me to eliminate all these ambiguities in the convert process.

annetill: Thanks for your answer. I think that for dangling line, we only have a B and a G because a dangling line is a part of a merged line, but I agree if we have a computation on a not merged network, it is wierd. 
As you PR is over IIDM (and not CGMES model), we have to take B1, G1, B2 and G2 for a line and for dangling line, I agree that it could be better to split it as you have done as we don't have the 4 attributes...

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/DanglingLineData.java
annetill: I am going to trust you for this second degree equation resolution, I have tried but I am a bit lost with ````sigmar```` and ````sigmai```` part. But I understand well the equation system, so I think I am going to trust you.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/DanglingLineData.java
```diff
@@ -0,0 +1,118 @@
...
+        p0 = danglingLine.getP0();
+        q0 = danglingLine.getQ0();
+
+        u1 = getV(danglingLine);
+        theta1 = getTheta(danglingLine);
+
+        boundaryBusU = Double.NaN;
+        boundaryBusTheta = Double.NaN;
+
+        Complex v1 = new Complex(u1 * Math.cos(theta1), u1 * Math.sin(theta1));
```
zamarrenolm, line 53: We could use 
```
angle=getAngle(danglingLine)
Complex v1 = ComplexUtils.polar2Complex(v, Math.toRadians(angle));
```
So the conversion degrees -> radians is in a single point, and the creation of complex from polar coordinates is explicit.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/DanglingLineData.java
```diff
@@ -0,0 +1,118 @@
...
+            double pBoundaryBus = -p0;
+            double qBoundaryBus = -q0;
+            Complex ytr = new Complex(r, x).reciprocal();
+            Complex ysh2 = new Complex(g2, b2);
+            Complex zt = ytr.add(ysh2).reciprocal();
+            Complex v0 = ytr.multiply(v1).divide(ytr.add(ysh2));
+            double v02 = v0.abs() * v0.abs();
+
+            double sigmar = (zt.getImaginary() * qBoundaryBus + zt.getReal() * pBoundaryBus) / v02;
+            double sigmai = (zt.getImaginary() * pBoundaryBus - zt.getReal() * qBoundaryBus) / v02;
```
zamarrenolm, line 71: Maybe we could make `s0 = new Complex(pBoundaryBus, qBoundaryBus)` and then let sigma be also a `Complex` number calculated from `zt` and `s0`, so we end up with more compact code (less potential errors)

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/DanglingLineData.java
```diff
@@ -0,0 +1,118 @@
...
+                vBoundaryBus = new Complex(0.5 + Math.sqrt(d), sigmai).multiply(v0);
+            }
+        }
+
+        boundaryBusU = vBoundaryBus.abs();
+        boundaryBusTheta = vBoundaryBus.getArgument();
+    }
+
+    private static double getV(DanglingLine danglingLine) {
+        if (danglingLine.getTerminal().getBusBreakerView() != null) {
```
zamarrenolm, line 84: I think we could get always the voltage and angle from the `BusView` `Bus`. Why trying to obtain it from `BusBreakerView` first ?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/TwtData.java
```diff
@@ -310,8 +345,20 @@ private void calculateThreeConnectedLegsFlow(double u1, double theta1, double u2
...
     }
 
-    private Complex calculateOneConnectedLegFlow(double u, double theta,
-        LinkData.BranchAdmittanceMatrix admittanceMatrixLeg, LinkData.BranchAdmittanceMatrix admittanceMatrixFirstOpenLeg,
+    private Complex calculateTwoConnectedLegsStarBusVoltage(double u1, double theta1, double u2, double theta2,
+        LinkData.BranchAdmittanceMatrix admittanceMatrixLeg1, LinkData.BranchAdmittanceMatrix admittanceMatrixLeg2,
+        LinkData.BranchAdmittanceMatrix admittanceMatrixOpenLeg) {
+
+        Complex v1 = new Complex(u1 * Math.cos(theta1), u1 * Math.sin(theta1));
+        Complex v2 = new Complex(u2 * Math.cos(theta2), u2 * Math.sin(theta2));
```
zamarrenolm, line 353: I suggest we use `v1 = ComplexUtils.polar2Complex(u1, theta1)`

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/DanglingLineData.java
marqueslanauja: Try this link [Two-buses](http://elequant.com/wp-content/uploads/Trias-2002-Two-bus-loadflow-exact-developments.pdf) Toni Trias the father of HELM explains how to solve the two buses problem exactly (second degree equation) and how to solve it applying the HELM methodology to this particular case.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/TwtData.java
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/DanglingLineData.java
annetill: Thanks a lot and I have succeed with that help. I wonder about the support of this resolution in Powsybl. Mabye, could you add the link you gave me in the comment of the code or more comments ? We don't have our background but we will need to support that code in the future.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/DanglingLineData.java
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/DanglingLineData.java
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/DanglingLineData.java
marqueslanauja: Solved

PR #1380, opened 2020-07-06
Add double property

agnesLeroy:
Signed-off-by: Agnès Leroy <agnes.leroy@rte-france.com>


commons/src/main/java/com/powsybl/commons/config/ModuleConfigUtil.java
```diff
@@ -32,6 +29,13 @@ private ModuleConfigUtil() {
         return getOptionalProperty(moduleConfig, names, moduleConfig::getOptionalStringListProperty);
     }
 
+    public static Optional<Double> getOptionalDoubleProperty(ModuleConfig moduleConfig, List<String> names) {
```
mathbagu, line 32: It's better to return an OptionalDouble instead of Optional<Double>

agnesLeroy: I don't think I can here, because of the way getOptionalProperty is defined: it returns <T> Optional<T>

iidm/iidm-converter-api/src/test/java/com/powsybl/iidm/ConversionParametersTest.java
```diff
@@ -0,0 +1,57 @@
...
+        assertEquals("TestParam", ConversionParameters.readStringParameter("TEST", properties, paramString2, ParameterDefaultValueConfig.INSTANCE));
+
+        List<String> stringList = new ArrayList<>();
+        stringList.add("Test1Property");
+        stringList.add("Test2Property");
+        List<String> stringList2 = new ArrayList<>();
+        stringList2.add("Test1Param");
+        stringList2.add("Test2Param");
+        properties.put("test-param-string-list", stringList);
+        Parameter paramStringList = new Parameter("test-param-string-list", ParameterType.STRING_LIST, "", stringList);
```
MioRtia, line 46: Maybe the default value for this parameter should not be `stringList` to prevent false positive in the test?

iidm/iidm-converter-api/src/test/java/com/powsybl/iidm/ConversionParametersTest.java
```diff
@@ -0,0 +1,57 @@
...
+        Parameter paramStringList2 = new Parameter("test-param-string-list2", ParameterType.STRING_LIST, "", stringList2);
+        assertEquals("Test2Property", ConversionParameters.readStringListParameter("TEST", properties, paramStringList, ParameterDefaultValueConfig.INSTANCE).get(1));
+        assertEquals("Test1Param", ConversionParameters.readStringListParameter("TEST", properties, paramStringList2, ParameterDefaultValueConfig.INSTANCE).get(0));
+
+        properties.put("test-param-double", 0.06);
+        Parameter paramDouble = new Parameter("test-param-double", ParameterType.DOUBLE, "", 0.06);
```
MioRtia, line 52: Same remark

Issue #1307, opened 2020-05-10
Slack bus extension


geofjamg:
* **Do you want to request a *feature* or report a *bug*?**
feature

* **What is the current behavior?**
 - there is no way to configure the slack bus used in the loadflow.
 - there is no way in case of automatic slack bus selection in the loadflow to get the selected slack bus and use it for instance in the UCTE export (which requires to specify the slack bus).

* **What is the expected behavior?**

Design proposal:

**IIDM extensions:**
Slack bus will be marked in the IIDM network using an extension. As slack bus should be supported in node/brealer and bus/breaker voltage levels. 

So we will have a `SlackBus` extension (abstract) attached to a `VoltageLevel` and 2 subclasses:
  - one for node/breaker voltage level `NodeBreakerSlackBus` with a node (int) attribute
  - one for bus/breaker voltage level `BusBreakerSlackBus` with a busId (String) attribute

`SlackBus` extension base class has a  `SlackBus.getTerminal()` abstract method, implemented in both subclasses depending of the topology kind. This method allows to get bus/branch bus in a very convenient way.

```java
VoltageLevel vl = ...
SlackBus slackBus = vl.getExtension(SlackBus.class);
Bus bus = slackBus.getTerminal().getBusView().getBus(); 
```

!!! A check has to be added in both `SlackBus` subclasses to ensure it is added to a voltage level with the right topology kind.  This could be enforced by having private constructors in subclasses and a adder like this:

```java
vl.newExtension(SlackBus.class)
   .setNode(4)
  .add();
```

or

```java
vl.newExtension(SlackBus.class)
   .setBusId("b2")
  .add();
```

And 'add' method to check if node or busId has been correctly set depending of the topology kind. That will also be consistent with has already been done in the IIDM base model.

**Merging:**
This extension design allows nothing special to do in case of network merging. There is no need to merge extensions. The only consequence is that it is valid to have several slack bus extensions in a network resulting of the merge of several networks having already a slack bus set.

**Loadflow:**
It will be specified in the `LoadFlow` interface that when running a loadflow:
  - if the network contains no slack bus extensions, default selection strategy is used (like most meshed bus)
  - if only one slack bus extensions is found, it is used to locate slack bus.
  - if several slack bus extensions are found, default selection strategy is applied on all buses located from slack bus extensions.

At the end of a loadflow run:
 - if loadflow converged, all slack bus extensions are removed and a new slack bus extension is added at the selected slack bus.
[...]

PR #1381, opened 2020-07-06
Slackbus extension: interface and implementation

floriand-e2r:

Issue #1307 

No


floriand-e2r:
The chosen design is different from the one suggested in the issue #1307:
- no terminal is linked to the slackBus; a get() method is directly returning the bus (of the BusView) instead 
- slackBus is an interface with 3 inner interfaces, one for each view

 Indeed, 
- using the terminal to get the bus associated to the slackBus might confuse the user and would only be used to get the bus
- using an abstract class and two subclasses is a design different from classes with similar design issues (Network, Terminal, VoltageLevel); besides it brings casting prolems: for instance in the serializer class it forces us to cast the slackbus into the appropriate subclass to get its attribute

geofjamg:
By returning the bus instead of the terminal you can only get bus of bus view and not bus of bus breaker view so this is a bad design

floriand-e2r:
As the user is mainly interested by the bus of the bus view, a method get() directly returning the bus of the busView was proposed, but in fact the proposed design allows to get both buses:
- slackBus.getBusBreakerView().getBus() to get the bus from the bus breaker view
- slackBus.getBusView().getBus() to get the bus from the bus view

It is also possible to get the node if slackBus was defined at node breaker level: 
- slackBus.getNodeBreakerView().getNode()

geofjamg:
You are just cloning in the extension an already existing abstraction which is the terminal. This extension purpose is to allow a terminal lookup so why not just returning a terminal?

floriand-e2r:
Two identified issues in the current PR:
- When a user creates a slackbus, the corresponding buses are computed; if the topology changes the bus returned is not consistent anymore. That might be an acceptable behaviour, but it would be interesting to compute the corresponding buses only when requested 
- If the slackbus was defined at node breaker level, @MioRtia outlined that the node might have no associated terminal if it is between two switches but that this node still belongs to a bus. @MioRtia suggested to use a traverser to find the terminal, as it was done for CGMES import, where the same problem also occurs.

MioRtia:
> You are just cloning in the extension an already existing abstraction which is the terminal. This extension purpose is to allow a terminal lookup so why not just returning a terminal?

Hi @geofjamg, I am finding a bit weird to return a `Terminal` in an extension named `SlackedBus`... Maybe the issue is just the name?

MioRtia:
> > You are just cloning in the extension an already existing abstraction which is the terminal. This extension purpose is to allow a terminal lookup so why not just returning a terminal?
> 
> Hi @geofjamg, I am finding a bit weird to return a `Terminal` in an extension named `SlackedBus`... Maybe the issue is just the name?

We can also return a `Terminal` and with `get()` return the BusView bus?

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/SlackBusImpl.java
```diff
@@ -0,0 +1,102 @@
...
+        super(vl);
+        if (node == SlackBusAdderImpl.NODE_UNINITIALIZED) {
+            throw new PowsyblException("node needs to be set for a SlackBus in VoltageLevel.NodeBreakerView");
+        }
+        Terminal terminal = vl.getNodeBreakerView().getOptionalTerminal(node).get();
+        Bus busBusBreakerView = terminal.getBusBreakerView().getBus();
+        Bus busBusView = terminal.getBusView().getBus();
+        this.busBreakerView = new BusBreakerViewImpl(busBusBreakerView);
+        this.nodeBreakerView = new NodeBreakerViewImpl(node);
+        this.busView = new BusViewImpl(busBusView);
```
mathbagu, line 79: This is a really bad idea to store the bus in the view: in a Node/Breaker Voltage level, these buses are calculated automatically, so new buses will be created when the topology changes. For Bus/Breaker Voltage Level, the bus of the bus/breaker view is configured, so it could be OK to keep it, but it's not for the bus view.

geofjamg:
> 
> 
> > > You are just cloning in the extension an already existing abstraction which is the terminal. This extension purpose is to allow a terminal lookup so why not just returning a terminal?
> > 
> > 
> > Hi @geofjamg, I am finding a bit weird to return a `Terminal` in an extension named `SlackedBus`... Maybe the issue is just the name?
> 
> We can also return a `Terminal` and with `get()` return the BusView bus?

Why would "get" return bus from bus view and not one from bus breaker view, it is an arbitrary choice. Probably we will more often use bus/breaker bus (h2 is connected to bus/breaker view, ucte import too, olf will be too when security analysis will be implemented)
Maybe the problem is that we need another abstraction in IIDM `TopologicalPoint` (or whatever other naming) that would model a point of the graph in a node/breaker topo (int node) or in a bus/breaker topo (String busId). And a `TopologicalPoint` could be or not associated to terminals.
`SlackBus` is also probably not the right name for the extension as it does not really model a bus but a `TopologicalPoint` which could be used to get a bus.
What I don't like with current code proposal is that slack bus extension is not anymore just a "link" to something in the base IIDM model (terminal in my initial proposition) but it embeds custom views which have nothing to do here for me.
I agree that pointing to a terminal is not ideal too as it would need in some rare cases a closest terminal search.

I am also wondering if this issue is not link to past discussion about regulating terminal (and missing TopologicalPoint concept).

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/SlackBusAdderImpl.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.util.OptionalInt;
+
+/**
+ * @author Florian Dupuy <florian.dupuy at rte-france.com>
+ */
+public class SlackBusAdderImpl extends AbstractExtensionAdder<VoltageLevel, SlackBus> implements SlackBusAdder {
+
+    private String busId;
+    private OptionalInt node;
```
geofjamg, line 22: Optional as a class attribute is often not recommended (https://stackoverflow.com/questions/29033518/is-it-a-good-practice-to-use-optional-as-an-attribute-in-a-class)

mathbagu:
Maybe we can introduce this new `TopologicalNode` concept to associate:
- A voltage level and a node, if the VL is configured in Node/Breaker topology
- A voltage level and a busId, if the VL is configured in Bus/Breaker topology

Its interface could look like a classic terminal:
```
interface TopologicalNode {
    VoltageLevel vl;
    int node;
    String busId;

    // Get the expected terminal or an equivalent terminal found using the topology graph
    Terminal getTerminal();
}
```
The main difference with a classic terminal, is the way to retrieve the underlying terminal when the node **is not associated to** a terminal. We can look for an *equivalent* terminal, using the topology graph. This terminal should not be cached in the slack bus to be invalid when the topology changes.

The `SlackBus` extension would have a single attribute of type `TopologicalNode`. I think it's interesting to offer shortcuts to the API user to retrieve the bus he is interested in. So I would like to keep the view interfaces, that are more simple than the views of a terminal. This is the idea when we propose the `get()` method. We thought that most user would use the `busView` in power flow computation, as they do not need to change the topology. In H2, the `withCoupler` attributes is set to `false` by default, to use the `BusView`. It can be overridden using the configuration. Do I miss something? Anyway, we can probably remove these methods. Our idea was to propose a convenient way to get a `Bus` from this extension. The `getNode` and `getBusId` method would be use only for serialization.

```
class SlackBusImpl {
   TopologicalNode getTopologicalNode();

    // This interface provide a shortcut to get the associated bus in BusBreakerView
    class BusBreakerView() {
        public Bus getBus() {
            return tn.getTerminal().getBusBreakerView().getBus();
        }
    }

    // This interface provide a shortcut to get the associated bus in BusView
    class BusView() {
        public Bus getBus() {
            return tn.getTerminal().getBusView().getBus();
        }
    }
}
```
For serialization purpose, we will get the TopologicalNode and export the relevant attributes (node or busId).

@geofjamg @MioRtia @floriand-e2r: what do you think?

floriand-e2r:
I'm ok with this new `TopologicalNode` concept, especially if we think of other uses for it. Wouldn't it be used only for Slackbus extension? I'm also wondering if the view interfaces should not be in that `TopologicalNode`, hence getting the bus from the slackbus in that tedious way `slackBus.getTopologicalNode().getXxxView().getBus()`. The slackbus will then only be a reference to a topological node: we'd be back to the first design, replacing the `getTerminal()` abstract method by a `getTopologicalNode()` one.

I too like the current view interfaces as they seem a convenient and clear way for the user to retrieve the wanted bus. It's more a wrapping up than a clone, meant to "hide" the terminal to the user. I think I'm missing the reason why they would have nothing to do here... I'm ok to replace them by direct methods (`getBusBusBreakerView()` / `getBusBusView()` / `getNode()`) as we agree that the user doesn't expect to manipulate a Terminal (do we?), but it seemed to me that it isn't really consistent with other similar problems.

A slight issue if we think of the terminal as part of the solution: I think that the implementation of SlackBus for a BusBreakerView VoltageLevel currently needs to use the package protected BusExt interface to get the corresponding Terminal.

Note that the nodes which are not connected to a terminal are not really wise slackbuses. In the current code, creating such a slackbus thus throws an exception. Using the TopologicalNode concept will avoid this probably rare issue.

geofjamg:
TopologicalNode is a bad name as it is already something different in cgmes

annetill:
I am not sure to get all what you expected with the concept of ````TopologicalPoint````` but for me it could not be used to define a regulating point. I don't know if it was the purpose of @geofjamg 's comment. In some cases, as active power control or reactive power control, the regulating terminal gives us the flow direction. So, if I understand well in ````TopologicalPoint```` you can retrieved 2 terminals that will lead to an opposite target. 
Sorry if it is not the purpose of the PR :-)

MioRtia:
After some thinking:
- not sure it is really useful to add another concept: I'm not sure where it would be used, we discussed before that it would not be enough for active/reactive power regulation for example, plus it might generate some confusion
- I really think `SlackBus` is a bad name for the extension, we were discussing with @annetill about SlackPoint or SlackDefinition?
- I'm not sure to follow what the issue is with intermediary view interfaces @geofjamg... On the other hand, we can dictate that the chosen node/bus for the slack point has to have a Connectable associated with it if the issue is we want to use a Terminal. We want the most connected bus anyway, so it should be the case more often than not. The serialization can then be with the Connectable ID + side just as regulation...

floriand-e2r:
We were discussing the issue with @annetill and she proposed a solution if we want to stick with the terminal: simply to rename the class `SlackTerminal`. Then getting and manipulating a terminal from that class would be expected by the user. It seems an interesting solution to me - just a naming issue after all?

@geofjamg @MioRtia @mathbagu: what do you think?

MioRtia:
> We were discussing the issue with @annetill and she proposed a solution if we want to stick with the terminal: simply to rename the class `SlackTerminal`. Then getting and manipulating a terminal from that class would be expected by the user. It seems an interesting solution to me - just a naming issue after all?
> 
> @geofjamg @MioRtia @mathbagu: what do you think?

Indeed, that would be the simplest  and cleanest solution!

annetill:
@floriand-e2r @geofjamg @mathbagu I was thinking about something really more simple with just a terminal as attribute and just one implementation and just one adder. So, maybe you have some arguments to make it more complex and it would be nice to explain me as I understand more the problem. Thanks a lot !

floriand-e2r:
The reason why I didn't choose that simplest way of implementing the slackTerminal is because we thought at first to define the slackBus based on the node / the bus id, and then the corresponding terminal can change, see previous comment:
> This terminal should not be cached in the slack bus to be invalid when the topology changes.

But in fact
- with the new name (SlackTerminal instead of SlackBus), it's ok to create a slackTerminal from a terminal, and not from a busId / node id
- it doesn't really make sense to manipulate an old slackTerminal after the topology has changed: if the topology changes, one should expect a new corresponding slackTerminal

@mathbagu @geofjamg Do you agree on that?

geofjamg:
With a terminal based impl, it does not depend on topology and has not to be updated when topology changed.

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/SlackTerminalNodeBreakerImpl.java
```diff
@@ -0,0 +1,34 @@
...
+
+import com.powsybl.commons.PowsyblException;
+import com.powsybl.commons.extensions.AbstractExtension;
+import com.powsybl.iidm.network.Terminal;
+import com.powsybl.iidm.network.VoltageLevel;
+
+/**
+ * @author Florian Dupuy <florian.dupuy at rte-france.com>
+ */
+public class SlackTerminalNodeBreakerImpl extends AbstractExtension<VoltageLevel> implements SlackTerminal {
```
mathbagu, line 17: This extension should inherit from `AbstractMultiVariantConnectableExtension` because the slack bus could depend from the variant.

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/SlackTerminalXmlSerializer.java
```diff
@@ -0,0 +1,83 @@
...
+            voltageLevel.newExtension(SlackTerminalAdder.class)
+                .setTerminal(terminal)
+                .add();
+            return voltageLevel.getExtension(SlackTerminalAdder.class);
+        } else {
+            return null;
+        }
+    }
+
+    public static void writeTerminalRef(Terminal t, XMLStreamWriter writer) throws XMLStreamException {
```
geofjamg, line 48: I think this utility function already exists somewhere in iidm xml converter module.

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/SlackTerminalXmlSerializer.java
```diff
@@ -0,0 +1,83 @@
...
+            } else if (c instanceof ThreeWindingsTransformer) {
+                ThreeWindingsTransformer twt = (ThreeWindingsTransformer) c;
+                writer.writeAttribute("side", twt.getSide(t).name());
+            } else {
+                throw new AssertionError("Unexpected Connectable instance: " + c.getClass());
+            }
+        }
+    }
+
+    public static Terminal readTerminalRef(Network network, XmlReaderContext context) {
```
geofjamg, line 66: Same

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/SlackTerminal.java
```diff
@@ -0,0 +1,38 @@
...
+     * Get the terminal pointed by the current SlackTerminal
+     * @return the corresponding terminal
+     */
+    Terminal getTerminal();
+
+    /**
+     * Shortcut to getTerminal().getBusView().getBus()
+     * @return the corresponding bus in the bus view
+     */
+    default Bus getBus() {
```
geofjamg, line 34: This is an arbitrary choice to return the bus view bus and doc does not really help.
I would just remove it and force the user to explictly choose the view.

floriand-e2r: Removed.

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/SlackTerminalXmlSerializer.java
floriand-e2r: Indeed, in TerminalRefXml. A minor issue blocked the refactoring in the commit so far, but it is in progress!

iidm/iidm-util/src/main/java/com/powsybl/iidm/network/util/TerminalChooser.java
```diff
@@ -0,0 +1,66 @@
...
+        for (Predicate<Terminal> predicate : rules) {
+            Optional<? extends Terminal> terminal = StreamSupport.stream(terminals.spliterator(), false)
+                .filter(predicate)
+                .findFirst();
+            if (terminal.isPresent()) {
+                return terminal.get();
+            }
+        }
+        Iterator<? extends Terminal> it = terminals.iterator();
+        return it.hasNext() ? it.next() : null;
```
mathbagu, line 55: I'm not sure it's really efficient: we iterate many times over the list of terminals.

iidm/iidm-util/src/main/java/com/powsybl/iidm/network/util/TerminalChooser.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    /**
+     * @param terminals the terminals among which a terminal has to be chosen
+     * @return the first terminal satisfying a rule (rules are checked in ascending order)
+     */
+    public Terminal choose(Stream<? extends Terminal> terminals) {
+        return choose(terminals.collect(Collectors.toList()));
+    }
```
mathbagu, line 64: Not sure it's really efficient, specially because you iterate over the stream to create the list, and then we iterate several times over the list in the other `choose` method

iidm/iidm-xml-converter/src/test/resources/slackTerminal.xml
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mathbagu, line 1: Do we need to add copyright header in this file too?

mathbagu:
@floriand-e2r You also have 1 smell to fix about the missing `@deprecated` in the javadoc

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractMultiVariantConnectableExtension.java
```diff
@@ -6,31 +6,21 @@
...
  * @author Ghiles Abdellah <ghiles.abdellah at rte-france.com>
  */
-public abstract class AbstractMultiVariantConnectableExtension<T extends Connectable> extends AbstractExtension<T> implements MultiVariantObject {
+@Deprecated
+public abstract class AbstractMultiVariantConnectableExtension<T extends Connectable> extends AbstractMultiVariantIdentifiableExtension<T> {
 
+    /**
+     * Deprecated, use {@link AbstractMultiVariantIdentifiableExtension} instead.
+     */
+    @Deprecated
```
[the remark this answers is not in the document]
floriand-e2r: Yes, it's already done in fact: PR #1404

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/SlackTerminalImpl.java
```diff
@@ -0,0 +1,65 @@
...
+import com.powsybl.iidm.network.impl.AbstractMultiVariantIdentifiableExtension;
+
+import java.util.ArrayList;
+
+/**
+ * @author Florian Dupuy <florian.dupuy at rte-france.com>
+ */
+public class SlackTerminalImpl extends AbstractMultiVariantIdentifiableExtension<VoltageLevel> implements SlackTerminal {
+
+    private final ArrayList<Terminal> terminals;
```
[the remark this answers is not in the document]
floriand-e2r: I kept `ArrayList` here to be able to call `ArrayList.ensureCapacity(int)` method. I can remove that call to use List instead, but the capacity will be ensured at each add then.

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/extensions/SlackTerminalTest.java
```diff
@@ -0,0 +1,179 @@
...
+            .setBus2("B1")
+            .add();
+
+        return network;
+    }
+
+    private static Terminal getBestTerminal(Network network, String busBusBreaker) {
+        // TODO: use the utility rules function which decides which terminal to choose from a given bus: see TerminalChooser in iidm-util module
+        Iterator<? extends Terminal> connectedTerminals =
+            network.getBusBreakerView().getBus(busBusBreaker).getConnectedTerminals().iterator();
```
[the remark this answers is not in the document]
floriand-e2r: Unfortunately the iidm-util module depends on iidm-impl, hence I would introduce a cyclic dependency if using TerminalChooser. I discussed it with @MioRtia, I should prepare another PR to move the tests of iidm-util (which are the reason of the iidm-impl dependence) in iidm-impl tests.

iidm/iidm-xml-converter/src/test/resources/slackTerminal.xml
floriand-e2r: Maybe not, there is no copyright in the network xml files for the other extensions.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractMultiVariantConnectableExtension.java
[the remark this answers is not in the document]
floriand-e2r: Ok

iidm/iidm-util/src/main/java/com/powsybl/iidm/network/util/TerminalChooser.java
floriand-e2r: Corrected. I thought the number of terminals was too small to bother about efficiency... :pensive:

iidm/iidm-util/src/main/java/com/powsybl/iidm/network/util/TerminalChooser.java
floriand-e2r: Same...

iidm/iidm-util/src/main/java/com/powsybl/iidm/network/util/TerminalChooser.java
```diff
@@ -0,0 +1,66 @@
...
+ * @author Florian Dupuy <florian.dupuy at rte-france.com>
+ */
+public final class TerminalChooser {
+
+    private final List<Predicate<Terminal>> rules;
+
+    public static TerminalChooser getDefaultSlackTerminalChooser() {
+        List<Predicate<Terminal>> rules = new ArrayList<>();
+        rules.add(t -> t.getConnectable().getType() == ConnectableType.BUSBAR_SECTION);
+        rules.add(t -> t.getConnectable() instanceof Injection);
```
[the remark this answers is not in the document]
floriand-e2r: They could also be added if you think it makes sense, but @geofjamg would only leave the Busbar predicate... so that's a compromise? Anyway that's only the default chooser, other choosers can be defined and used.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/extensions/SlackTerminal.java
```diff
@@ -0,0 +1,84 @@
...
+    default SlackTerminal setTerminal(Terminal terminal, boolean cleanIfCleanable) {
+        if (cleanIfCleanable) {
+            if (setTerminal(terminal).isCleanable()) {
+                getExtendable().removeExtension(SlackTerminal.class);
+            }
+        } else {
+            setTerminal(terminal);
+        }
+        return this;
+    }
```
mathbagu, line 77: I would write the method like this:
```
    default SlackTerminal setTerminal(Terminal terminal, boolean cleanIfCleanable) {
        boolean cleanable = setTerminal(terminal);
        if (cleanIfCleanable && cleanable) {
            getExtendable().removeExtension(SlackTerminal.class);
        }
        return this;
    }
```

iidm/iidm-util/src/main/java/com/powsybl/iidm/network/util/TerminalChooser.java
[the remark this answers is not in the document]
mathbagu: There is a risk that there is no busbar section (Bus/Breaker VoltageLevel) and no injection neither. For instance, if we consider a star-like network, the star bus is a good slack bus, even there is no injection connected to it.

I would add the 3 other predicates to be sure you can always attach the extension to a terminal.

Can you please add a unit test using the Eurostag Tutorial 1, and ensure your are able to set the slack terminal

iidm/iidm-util/src/main/java/com/powsybl/iidm/network/util/TerminalChooser.java
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * A class which allows to choose a terminal, according to given filters, among several terminals
+ * @author Florian Dupuy <florian.dupuy at rte-france.com>
+ */
+public final class TerminalChooser {
+
+    private final List<Predicate<Terminal>> rules;
+
+    public static TerminalChooser getDefaultSlackTerminalChooser() {
```
mathbagu, line 26: I would rename this method `getDefault` to avoid redundancy between the name of the class, and the name of this method

iidm/iidm-util/src/main/java/com/powsybl/iidm/network/util/TerminalChooser.java
```diff
@@ -0,0 +1,66 @@
...
+        List<Predicate<Terminal>> rules = new ArrayList<>();
+        rules.add(t -> t.getConnectable().getType() == ConnectableType.BUSBAR_SECTION);
+        rules.add(t -> t.getConnectable() instanceof Injection);
+        return new TerminalChooser(rules);
+    }
+
+    public TerminalChooser(List<Predicate<Terminal>> rules) {
+        this.rules = rules;
```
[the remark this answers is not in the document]
mathbagu: I reopen this conversation because it's not fixed

iidm/iidm-util/src/main/java/com/powsybl/iidm/network/util/TerminalChooser.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.*;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+/**
+ * A class which allows to choose a terminal, according to given filters, among several terminals
+ * @author Florian Dupuy <florian.dupuy at rte-france.com>
+ */
+public final class TerminalChooser {
```
mathbagu, line 22: I would rename this class `TerminalFinder` and the rename the `choose` method `find`

iidm/iidm-util/src/main/java/com/powsybl/iidm/network/util/TerminalChooser.java
mathbagu: Same what?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/extensions/SlackTerminalXmlSerializer.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public SlackTerminalXmlSerializer() {
+        super("slackTerminal", "network", SlackTerminal.class, true, "slackTerminal.xsd",
+            "http://www.powsybl.org/schema/iidm/ext/slack_terminal/1_0", "slt");
+    }
+
+    @Override
+    public void write(SlackTerminal slackTerminal, XmlWriterContext context) throws XMLStreamException {
+        TerminalRefXml.writeTerminalRef(slackTerminal.getTerminal(), (NetworkXmlWriterContext) context, getNamespaceUri(), ELEM_TERMINAL_REF);
+    }
```
mathbagu, line 41: I would check if the terminal is not null. I would also override the `hasSubElements`: this method should return true if the terminal is set, false otherwise

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/extensions/SlackTerminalXmlSerializer.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Override
+    public SlackTerminal read(VoltageLevel voltageLevel, XmlReaderContext context) throws XMLStreamException {
+        NetworkXmlReaderContext networkContext = (NetworkXmlReaderContext) context;
+        XmlUtil.readUntilEndElement(getExtensionName(), networkContext.getReader(), () -> {
+            if (networkContext.getReader().getLocalName().equals(ELEM_TERMINAL_REF)) {
```
mathbagu, line 47: I think if you change a little bit the export, the terminalRef is required, so you can remove this test

iidm/iidm-xml-converter/src/main/resources/xsd/slackTerminal.xsd
```diff
@@ -0,0 +1,23 @@
...
+<xs:schema version="1.0"
+           xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           xmlns:iidm="http://www.powsybl.org/schema/iidm/1_3"
+           targetNamespace="http://www.powsybl.org/schema/iidm/ext/slack_terminal/1_0"
+           elementFormDefault="qualified">
+    <xs:import namespace="http://www.powsybl.org/schema/iidm/1_3" schemaLocation="iidm_V1_3.xsd"/>
+    <xs:element name="slackTerminal">
+        <xs:complexType>
+            <xs:sequence>
+                <xs:element name="terminalRef" type="iidm:TerminalRef"/>
```
mathbagu, line 19: Set the min/max occurs: this element is required

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/extensions/SlackTerminal.java
floriand-e2r: Yes it's more compact! but it would be better to check if the slackTerminal is cleanable only if required, so:

    default SlackTerminal setTerminal(Terminal terminal, boolean cleanIfCleanable) {
        setTerminal(terminal);
        if (cleanIfCleanable && isCleanable()) {
            getExtendable().removeExtension(SlackTerminal.class);
        }
        return this;
    }

Otherwise with the multivariant implementation we iterate over the array for nothing

floriand-e2r: If we really want to avoid calling isCleanable when not required, couldn't we also add `terminal!=null` in the condition?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/extensions/SlackTerminalXmlSerializer.java
floriand-e2r: Corrected. And hasSubElements is now always false, as the attributes are now directly in the slackTerminal (see comment below)

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/extensions/SlackTerminalXmlSerializer.java
floriand-e2r: Corrected, no test anymore (attributes directly in the slackTerminal object).

iidm/iidm-xml-converter/src/main/resources/xsd/slackTerminal.xsd
floriand-e2r: Attributes are now directly to the slackTerminal object.

iidm/iidm-util/src/main/java/com/powsybl/iidm/network/util/TerminalChooser.java
floriand-e2r: Same as previous conversation: I corrected it.

iidm/iidm-util/src/main/java/com/powsybl/iidm/network/util/TerminalChooser.java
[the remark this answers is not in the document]
floriand-e2r: Correctly corrected this time!

PR #1384, opened 2020-07-07
Compute and validate equivalent impedance of merged lines

zamarrenolm:

Calculation of equivalent PI model of merged lines in `MergingView`.
Fix calculation of flows for Lines with different nominal voltages at end buses (structural ratio does not have to be taken into account).
Validation of calculations against ENTSO-E conformity test case (MicroGrid).

**What is the current behavior?** *(You can also link to an open issue here)**
The impedance of a merged line in the `MergingView` is calculated adding the impedances of each merged lines. This is incorrect. Total difference in flows at boundary lines with reference test case (MicroGrid base case) is approx 20 MVAr.

Equivalent PI model of merged lines in `MergingView` is computed using cascade of quadripoles. The same approach was used in CGMES import when the input files contains data from two TSOs that share a boundary. Code for Quadripole computation has been moved to IIDM API util package.

The equivalent line computation is validated against reference data from ENTSO-E conformity test case MicroGrid. Base case data has been used (it has been obtained from Excel file present in the documentation). The unit test checks that all flows of transmission lines match the expected flows in three different situations:
1. Merging View obtained from imported Individual Grid Models of BE, NL.
2. CGMES import of assembled model, keeping the boundary nodes as explicit buses in IIDM.
3. CGMES import of assembled model, removing the boundary nodes and creating merged lines during import.

The merged line in the `MergingView` forbids the modification of the `r, x, g1, b1, g2, b2` attributes.


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/Quadripole.java
```diff
@@ -0,0 +1,92 @@
...
+
+    public static Quadripole fromSeriesImpedance(double r, double x) {
+        return new Quadripole(new Complex(1), new Complex(r, x), new Complex(0), new Complex(1));
+    }
+
+    public static Quadripole fromShuntAdmittance(double g, double b) {
+        return new Quadripole(new Complex(1), new Complex(0), new Complex(g, b), new Complex(1));
+    }
+
+    public Quadripole cascade(Quadripole q2) {
```
annetill, line 71: can you add a formula in the comment to ease the understanding ?

annetill:
The merged line in the MergedView should forbid the modification of the r, x, g1, b1, g2, b2 attributes and we will have to same issue for a classical merged line. Are you going to desactivate the setter in this PR or I should manage this on my side ?

iidm/iidm-mergingview/src/test/java/com/powsybl/iidm/mergingview/DanglingLineAdapterTest.java
```diff
@@ -161,18 +161,20 @@ public void mergedDanglingLine() {
...
-        assertEquals(dl1.getX() + dl2.getX(), mergedLine.getX(), 0.0d);
-        assertSame(mergedLine, mergedLine.setG1(3.0d));
-        assertEquals(dl1.getG(), mergedLine.getG1(), 0.0d);
-        assertSame(mergedLine, mergedLine.setG2(4.0d));
-        assertEquals(dl2.getG(), mergedLine.getG2(), 0.0d);
-        assertSame(mergedLine, mergedLine.setB1(5.0d));
-        assertEquals(dl1.getB(), mergedLine.getB1(), 0.0d);
-        assertSame(mergedLine, mergedLine.setB2(6.0d));
-        assertEquals(dl2.getB(), mergedLine.getB2(), 0.0d);
+//        XXX LUMA set r, x, g, b should not be allowed on merged lines
```
annetill, line 164: All these tests should be removed and depending on the test coverage, to be replaced.

zamarrenolm:
> The merged line in the MergedView should forbid the modification of the r, x, g1, b1, g2, b2 attributes and we will have to same issue for a classical merged line. Are you going to desactivate the setter in this PR or I should manage this on my side ?

The `MergingView` now throws an exception when the user tries to set the attributes. For destructive merge, the `TieLineImpl` throws an exception for the modification of attributes at TieLine level, but allows modification at each line _half_, the equivalent pi model is recomputed on the fly in the getters.

iidm/iidm-mergingview/src/test/java/com/powsybl/iidm/mergingview/DanglingLineAdapterTest.java
zamarrenolm: The commented tests have been removed (merged line do not allow modification of `r, x, g1, b1, g2, b2` attributes)

annetill:
See #1755.

PR #1386, opened 2020-07-08
CGMES updates: reduce size of changelog not recording state variables

zamarrenolm:

The changelog that records changes in IIDM attributes is used to perform updates in the CGMES triplestore when the user requests a CGMES export.
State variables (voltages, angles, flows at terminals) are not recorded explicitly in the changelog, as its values will be systematically obtained during the export to build the CGMES SV profile from scratch. No selective update of these attributes is required.

These attributes were previously stored in the changelog, and later ignored during the update.

Changelog size is reduced, and less operations are perform during update.

**Other information**:
The tests for CGMES update have been modified to remove size limitations on the number of changes performed.

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/elements16/ShuntCompensatorToShuntCompensator.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/elements16/ShuntCompensatorToShuntCompensator.java
@@ -15,7 +15,6 @@ import com.powsybl.cgmes.model.CgmesSubset;
 public class ShuntCompensatorToShuntCompensator extends IidmToCgmes {
 
     ShuntCompensatorToShuntCompensator() {
-        ignore("q");
 
         simpleUpdate("bPerSection", "cim:LinearShuntCompensator.bPerSection", CgmesSubset.EQUIPMENT);
         simpleUpdate("maximumSectionCount", "cim:ShuntCompensator.maximumSections", CgmesSubset.EQUIPMENT);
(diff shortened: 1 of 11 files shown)


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/elements16/ShuntCompensatorToShuntCompensator.java
```diff
@@ -15,7 +15,6 @@
 public class ShuntCompensatorToShuntCompensator extends IidmToCgmes {
 
     ShuntCompensatorToShuntCompensator() {
-        ignore("q");
 
         simpleUpdate("bPerSection", "cim:LinearShuntCompensator.bPerSection", CgmesSubset.EQUIPMENT);
```
annetill, line 19: Now, we have non linear shunt compensator, so it means that we also have to support that for the CGMES update. I wonder if the code is working for that kind of equipment.

zamarrenolm: I will have to review. Current export does not support `NonLinearShuntCompensators`.

zamarrenolm: I propose to add export support for that equipment in a separate, specific PR.

annetill: Yes, great. I am going to open an issue to keep that in mind and I am going to merge this PR as it is.

PR #1390, opened 2020-07-09
Fix sonar bug: no fail() while catching an assertionerror

MioRtia:

No

Quality enhancement/bug fix

 Use *fail()* inside a try-catch catching an AssertionError

Don't use *fail()* inside a try-catch catching an AssertionError

No


computation/src/test/java/com/powsybl/computation/CompletableFutureTaskTest.java
```diff
@@ -67,10 +67,8 @@ public void whenTaskThrowsThenThrowExecutionException() {
         try {
             task.get();
             fail();
-        } catch (ExecutionException exc) {
-            assertTrue(exc.getCause() instanceof MyException);
         } catch (Throwable exc) {
-            fail();
+            assertTrue(exc.getCause() instanceof MyException);
         }
```
mathbagu, line 72: I think you can simply remove the `catch (Throwable exc)`

PR #1395, opened 2020-07-13
add method to get all tapChanger steps

### LINKED ISSUE #663: Add convenient interface for iteration on TapChanger steps
As a user of IIDM API, I need for some process to iterate over a PhaseTapChanger's steps to find a specific tap position index based on PhaseTapChangerStep angle.

It is possible to do that kind of treatment using a for loop betwen low tap position and high tap position, but it is not very convenient for chaining the treatments.

I think it could be interesting to be able to iterate over maps or stream, by adding to the TapChanger interface a new method:
```java
Map<Integer, S> getAllSteps();
```
This way, you can turn (not actual use case):
```java
for (int tapPosition = phaseTapChanger.getLowTapPosition(); tapPosition <= phaseTapChanger.getHighTapPosition(); tapPosition++) {
    PhaseTapChangerStep step = phaseTapChanger.getStep(tapPosition);
    System.out.println(String.format("alpha[%d]=%f", tapPosition, step.getAlpha()));
}
```
into something like:

```java
phaseTapChanger.getAllSteps().forEach((tapPosition, step) -> System.out.println(String.format("alpha[%d]=%f", tapPosition, step.getAlpha())));
```

jlabous:

fix #663

### DIFF
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/TapChanger.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/TapChanger.java
@@ -6,6 +6,8 @@
  */
 package com.powsybl.iidm.network;
 
+import java.util.Map;
+
 /**
  *
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
@@ -121,4 +123,10 @@ public interface TapChanger<C extends TapChanger<C, S>, S extends TapChangerStep
      */
     void remove();
 
+    /**
+     * Get all Tap changer steps
+     */
+    default Map<Integer, S> getAllSteps() {
+        throw new UnsupportedOperationException("Not implemented");
+    }
 }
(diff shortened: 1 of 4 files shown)


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/TapChanger.java
```diff
@@ -121,4 +123,10 @@ default C setTargetDeadband(double targetDeadband) {
      */
     void remove();
 
+    /**
+     * Get all Tap changer steps
+     */
+    default Map<Integer, S> getAllSteps() {
+        throw new UnsupportedOperationException("Not implemented");
```
MioRtia, line 130: You can have a more working default implementation:
```java
Map<Integer, S> steps = new HashMap();
for (int i = getLowTapPosition(); i <= getHighTapPosition; i++) {
  steps.put(i, getStep(i));
}
return steps;
```

PR #1396, opened 2020-07-13
Terminals getters added to Bus interface

floriand-e2r:

No

Feature

Connected terminals getters exist but in the BusExt interface (impl package)

The two getters are added in the Bus interface. 

**Does this PR introduce a breaking change or deprecate an API?** No
*If yes, check the following:*

**Other information**:
These getters are needed for using the SlackTerminal introduced by issue #1307


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Bus.java
```diff
@@ -240,4 +240,14 @@
      */
     void visitConnectedOrConnectableEquipments(TopologyVisitor visitor);
 
+    /**
+     * Get an iterable of the terminals connected to the bus.
+     */
+    Iterable<? extends Terminal> getConnectedTerminals();
```
mathbagu, line 246: I think you should to provide a default behavior to not break existing code (powsybl-network-store)

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Bus.java
```diff
@@ -240,4 +240,14 @@
...
+
+    /**
+     * Get a stream of the terminals connected to the bus.
+     */
+    Stream<? extends Terminal> getConnectedTerminalStream();
```
mathbagu, line 251: Same remark

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Bus.java
```diff
@@ -240,4 +241,18 @@
      */
     void visitConnectedOrConnectableEquipments(TopologyVisitor visitor);
 
+    /**
+     * Get an iterable of the terminals connected to the bus.
+     */
+    default Iterable<? extends Terminal> getConnectedTerminals() {
+        return new LinkedList<>();
```
geofjamg, line 248: Why a LinkedList and not an ArrayList?

floriand-e2r: An old pre-Java 8 habit, as it is a constant empty list. Thus it does not require the whole ArrayList mechanism.

floriand-e2r:
@geofjamg what do you think about the code smell "Generic wildcard types should not be used in return types"? Here that allowed me to have a minimal influence on the Bus-inheriting classes, which can then have a list of TerminalExt and use TerminalExt methods. They don't though: all the calls to getConnectedTerminals() or getConnectedTerminalStream() can easily deal with Terminal instead of TerminalExt.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Bus.java
floriand-e2r: replaced by `Collections.emptyList()` (and not the raw type `Collections.EMPTY_LIST`)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusExt.java
```diff
@@ -16,8 +16,10 @@
  */
 interface BusExt extends Bus {
 
+    @Override
```
geofjamg, line 19: To remove as it is already in Bus?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractBus.java
```diff
@@ -52,8 +52,10 @@ public VoltageLevel getVoltageLevel() {
         return voltageLevel;
     }
 
+    @Override
```
geofjamg, line 55: Not necessary as it is already added to ConfiguredBusImpl and CalculaedBusImpl

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusExt.java
floriand-e2r: Leaving it forced the classes inheriting BusExt to use TerminalExt, and not another class extending Terminal.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractBus.java
floriand-e2r: It is already added in ConfiguredBusImpl and CalculatedBusImpl but the methods using it in AbstractBus assume it returns a TerminalExt and not a Terminal. As mentionned earlier, these methods can deal with a Terminal though.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusExt.java
floriand-e2r: For instance, with latest commit removing it, MergedBus.getConnectedTerminals() is now returning Iterable<Terminal> and not Iterable<TerminalExt>, and dealing with ConfiguredBus.getConnectedTerminals() hence with terminals whose type is `? extends Terminal`.

PR #1397, opened 2020-07-13
Update the javadoc with the Tables copied from the current website

agnesLeroy:
This PR is linked to the website refactoring: the Tables present in the current grid model documentation are copied to the javadoc.


MioRtia:
The displaying does not really work in javadoc (see image below). I think you have to use `<table>`.

Capture d’écran de 2020-07-15 11-33-48

agnesLeroy:
@MioRtia it should be fixed now

PR #1399, opened 2020-07-15
Runnables in XIIDM util methods can now throw XmlStreamException

MioRtia:

No

Update

When a runnable is used with XIIDM util that throws an XmlStreamException, it must be catch in the `run()` method

We now use an IidmXmlRunnable which throws an XmlStreamException 

No

**Other information**:
Modifications were done in order not to make any breaking changes

### DIFF
--- a/iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/util/IidmXmlUtil.java
+++ b/iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/util/IidmXmlUtil.java
@@ -21,6 +21,20 @@ import java.util.Objects;
  */
 public final class IidmXmlUtil {
 
+    public interface IidmXmlRunnable extends Runnable {
+
+        @Override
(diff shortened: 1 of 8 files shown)


iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/util/IidmXmlUtil.java
```diff
@@ -21,6 +21,20 @@
  */
 public final class IidmXmlUtil {
 
+    public interface IidmXmlRunnable extends Runnable {
+
+        @Override
+        default void run() {
```
mathbagu, line 27: You should mark this method final to avoid someone to override it?

MioRtia: It seems it is not possible to make methods in interfaces final (even with default implementation)

PR #1402, opened 2020-07-17
Add the possibility to associate a type to an alias (indicating what it corresponds to)

MioRtia:

No

Feature

When adding an alias, there is no indication of what the alias' origin is.

A `type` string can be associated to the alias, providing more information of where the alias comes from or what it indicates.

No

**Other information**:
Some fixes on the XIIDM serialization of aliases have been done too. If there is some discussion about this PR, I can move them to another one.
@geofjamg Is this behavior easily reproducible with the persistent implementation?

### DIFF
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Identifiable.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Identifiable.java
@@ -48,8 +48,8 @@ public interface Identifiable<I extends Identifiable<I>> extends Extendable<I> {
     /**
      * Get the aliases of the object with a given alias type.
      */
-    default Set<String> getAliases(String aliasType) {
-        return Collections.emptySet();
+    default Optional<String> getAliasFromType(String aliasType) {
+        return Optional.empty();
     }
 
     /**
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Identifiable.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Identifiable.java
@@ -46,10 +46,31 @@ public interface Identifiable<I extends Identifiable<I>> extends Extendable<I> {
     }
 
     /**
+     * Get the aliases of the object with a given alias type.
+     */
+    default Set<String> getAliases(String aliasType) {
+        return Collections.emptySet();
+    }
+
+    /**
+     * Get the alias type of an alias if it exists, an empty optional otherwise.
+     */
+    default Optional<String> getAliasType(String alias) {
+        return Optional.empty();
+    }
+
+    /**
      * Add an alias to the object. Aliases must be unique in associated Network, and different
-     * from any identifiable ID.
+     * from any identifiable ID. No alias type is associated to this alias.
      */
     default void addAlias(String alias) {
+    }
+
+    /**
+     * Add an alias to the object. Aliases must be unique in associated Network, and different
+     * from any identifiable ID. This alias is associated to a given alias type.
+     */
+    default void addAlias(String alias, String aliasType) {
     }
 
     /**
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Identifiable.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Identifiable.java
@@ -46,7 +46,7 @@ public interface Identifiable<I extends Identifiable<I>> extends Extendable<I> {
     }
 
     /**
-     * Get the aliases of the object with a given alias type.
+     * Get the alias of the object with a given alias type if it exists. Else return an empty optional.
      */
     default Optional<String> getAliasFromType(String aliasType) {
         return Optional.empty();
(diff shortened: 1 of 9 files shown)


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiable.java
```diff
@@ -24,10 +25,11 @@
 
     protected boolean fictitious = false;
 
-    protected Set<String> aliases = new TreeSet<>();
-
     protected final Properties properties = new Properties();
 
+    private final Set<String> aliases = new TreeSet<>();
+    private final Map<String, String> aliasesTypeByAlias = new HashMap<>();
```
geofjamg, line 31: We could just have one map for that ?  `Set<String> aliases` is now useless?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiable.java
```diff
@@ -58,17 +60,43 @@ public String getNameOrId() {
         return Collections.unmodifiableSet(aliases);
     }
 
+    @Override
+    public Set<String> getAliases(String aliasType) {
```
geofjamg, line 64: This is a very inefficient implementation. It depends how it is used but we have to be aware of potential performance issue using this method. Is it really needed?

MioRtia: I made a draft PR #1447 to show aliases' types can be used in CGMES update and you're right: we don't really need this method (or at least not in this form). Types are actually unique for one identifiable if they are not null in this use case (in the example, a switch can only have one alias associated with a "terminal1" type). I need to rethink how to do this.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Identifiable.java
```diff
@@ -45,13 +45,34 @@ default String getName() {
         return Collections.emptySet();
     }
 
+    /**
+     * Get the aliases of the object with a given alias type.
```
geofjamg, line 49: -> alias

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Identifiable.java
```diff
@@ -45,13 +45,34 @@ default String getName() {
...
     /**
      * Add an alias to the object. Aliases must be unique in associated Network, and different
-     * from any identifiable ID.
+     * from any identifiable ID. No alias type is associated to this alias.
      */
     default void addAlias(String alias) {
     }
 
+    /**
+     * Add an alias to the object. Aliases must be unique in associated Network, and different
```
geofjamg, line 70: Maybe just specify what should be the impl behaviour if an alias with the same type already exists or another alias with same value already exists? Exception?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Identifiable.java
```diff
@@ -45,13 +45,34 @@ default String getName() {
...
+     * Get the alias type of an alias if it exists, an empty optional otherwise.
+     */
+    default Optional<String> getAliasType(String alias) {
+        return Optional.empty();
+    }
+
     /**
      * Add an alias to the object. Aliases must be unique in associated Network, and different
-     * from any identifiable ID.
+     * from any identifiable ID. No alias type is associated to this alias.
```
geofjamg, line 64: Maybe just specify what should be the impl behaviour if another alias with same value already exists or an IIDM is? Exception?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Identifiable.java
geofjamg: Also you can specify that aliasType with a null value is supported and means no alias

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
```diff
@@ -149,8 +147,9 @@ protected void readElement(String id, ShuntCompensatorAdder adder, NetworkXmlRea
                     break;
                 case AliasesXml.ALIAS:
                     IidmXmlUtil.assertMinimumVersion(ROOT_ELEMENT_NAME, AliasesXml.ALIAS, IidmXmlUtil.ErrorMessage.NOT_SUPPORTED, IidmXmlVersion.V_1_3, context);
+                    String aliasType = context.getReader().getAttributeValue(null, "type");
                     String alias = context.getAnonymizer().deanonymizeString(context.getReader().getElementText());
-                    aliases.add(alias);
+                    aliases.put(alias, aliasType);
```
geofjamg, line 152: Just a thought but it is very dangerous to have only for shunt kind of duplication of alias management. It is because of model sub element?

geofjamg (approved):
Just a minor comment: If you can add more additional infos in API doc about error handling so that it is clear for IIDM store impl how it has to be implemented.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
MioRtia: Yes because shunts are the only equipment who needs its sub element to be created (other equipment can be created without their sub elements).

PR #1408, opened 2020-07-21
CGMES switch at boundary to dangling line z0

zamarrenolm:

Switches with one end at boundary node in CGMES are mapped to dangling lines with zero impedance in IIDM.

Boundaries that had switches were ignored. Their corresponding equivalent injections were not mapped to IIDM, resulting in potentially great imbalances.

Work in progress, still have to add unit tests and handle the case where a Line and a Switch are found at the same boundary node. In this case a Merged Line should be created, with impedance coming only from the "Line" side.


MioRtia (approved):
It is okay but just to be clear: is the parameter `MERGE_BOUNDARIES_USING_TIE_LINES_PARAMETER` to be deleted in the long run? If not, what is its use? In which cases should it be false?

annetill (approved):

PR #1420, opened 2020-08-03
CGMES Full model parser

geofjamg:

No

feature

A fast and lightweight parser is implemented to get FullModel section of a CGMES profile. 

**Other information**:

### DIFF
--- a/cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/FullModel.java
+++ b/cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/FullModel.java
@@ -16,10 +16,7 @@ import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamReader;
 import java.io.Reader;
 import java.time.ZonedDateTime;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Objects;
-import java.util.Optional;
+import java.util.*;
 
 /**
  *
@@ -82,15 +79,15 @@ public class FullModel {
     }
 
     public List<String> getProfiles() {
-        return profiles;
+        return Collections.unmodifiableList(profiles);
     }
 
     public List<String> getDependentOn() {
-        return dependentOn;
+        return Collections.unmodifiableList(dependentOn);
     }
 
     public List<String> getSupersedes() {
-        return supersedes;
+        return Collections.unmodifiableList(supersedes);
     }
 
     public String getModelingAuthoritySet() {
--- a/cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/FullModel.java
+++ b/cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/FullModel.java
@@ -183,7 +183,7 @@ public class FullModel {
         } catch (XMLStreamException e) {
             throw new UncheckedXmlStreamException(e);
         }
-        // the other attributes validity are be validated in the constructor
+        // the other attributes are checked in the constructor
         if (context.version == null) {
             throw new PowsyblException("Version is missing");
         }
--- a/cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/FullModel.java
+++ b/cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/FullModel.java
@@ -47,15 +47,18 @@ public class FullModel {
 
     public FullModel(String id, ZonedDateTime scenarioTime, ZonedDateTime created, String description, int version, List<String> profiles,
                      List<String> dependentOn, List<String> supersedes, String modelingAuthoritySet) {
-        this.id = Objects.requireNonNull(id);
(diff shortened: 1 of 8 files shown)


cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/FullModel.java
```diff
@@ -0,0 +1,208 @@
...
+        return profiles;
+    }
+
+    public List<String> getDependentOn() {
+        return dependentOn;
+    }
+
+    public List<String> getSupersedes() {
+        return supersedes;
+    }
```
mathbagu, line 91: You should return immutable version of these lists

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/FullModel.java
```diff
@@ -0,0 +1,208 @@
...
+            factory.setProperty(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "");
+
+            XMLStreamReader xmlReader = factory.createXMLStreamReader(reader);
+            try {
+                XmlUtil.readUntilStartElement(new String[] {"/", CgmesNames.RDF, CgmesNames.FULL_MODEL}, xmlReader, () -> {
+                    context.id = xmlReader.getAttributeValue(CgmesNamespace.RDF_NAMESPACE, CgmesNames.ABOUT);
+                    XmlUtil.readUntilEndElement(CgmesNames.FULL_MODEL, xmlReader, () -> {
+                        switch (xmlReader.getLocalName()) {
+                            case CgmesNames.SCENARIO_TIME:
+                                context.scenarioTime = ZonedDateTime.parse(XmlUtil.readUntilEndElement(CgmesNames.SCENARIO_TIME, xmlReader, () -> { }));
```
mathbagu, line 169: You should introduce an overload for readUntilElement, that takes only the two first parameters. This would make this code more readable:
```
XmlUtil.readUntilEndElement(CgmesNames.SCENARIO_TIME, xmlReader);
```

This overload would call the existing one, with the empty lambda

Another idea would be to name this method `XmlUtil.readText()`.

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/FullModel.java
```diff
@@ -0,0 +1,208 @@
...
+                        }
+                    });
+                });
+            } finally {
+                xmlReader.close();
+            }
+        } catch (XMLStreamException e) {
+            throw new UncheckedXmlStreamException(e);
+        }
+        context.validate();
```
mathbagu, line 204: I would merge the validate method and the constructor. As the constructor is public, one can call it with empty profiles/dependentOn lists, that is forbidden in the parser.

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/FullModel.java
geofjamg: As it is also passed to contrusctor we should maybe make a defensive copy in the constructor

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/FullModel.java
geofjamg: Ok for XmlUtil.readText()

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/FullModel.java
geofjamg: The only problem is for version as in the constructor it is an int and Integer in the parsing context to detect missing version

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/FullModel.java
geofjamg: Done, defensive copy in the contructor

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/FullModel.java
geofjamg: Done and sync with the other PR

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/FullModel.java
geofjamg: Done except for version

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/FullModel.java
mathbagu: It's not equivalent... How do you prevent a user to clear your own copy?
```
fullModel.getDependentOn().clear();
```
I think the copy in the constructor is not necessary because a new context is created each time the parsing is called, but you shoud protect your own data.

PR #1424, opened 2020-08-06
CGMES rdf4j update

mathbagu:

N/A

Update RDF4J dependency.

We use an old version of RDF4J that is not ready for Java 11 and is not compatible with last version of springs.

We use the latest release, but it completely changed the optimization plan of the queries. Some queries have to be fixed to have the same performance than before.

Query for `operationalLimits` was rewritten to avoid introducing new scopes (unions). Instead, we have used optionals. The performance with unions was very bad for some real-world cases that have a big number of limits defined at Equipment level. The good performance of the revised query has been verified also against RDF4J v2.4. 

**Other information**:
The use of the cache for all query results is not recommended a will be removed before final publish of this PR.

A related PR #1500 has been made to avoid running some selected queries multiple times (`terminals`, `connectivityNodes`, `topologicalNodes`).


cgmes/cgmes-conformity/src/test/java/com/powsybl/cgmes/conformity/test/CgmesConformity1NetworkCatalog.java
```diff
@@ -912,12 +912,12 @@ public static Network microType4BE() {
...
-                .setVoltageLevel1(vlAnvers225.getId())
-                .setConnectableBus2(busAnvers220.getId())
-                .setBus2(busAnvers220.getId())
-                .setVoltageLevel2(vlAnvers220.getId())
+                .setConnectableBus1(busAnvers220.getId())
+                .setBus1(busAnvers220.getId())
+                .setVoltageLevel1(vlAnvers220.getId())
+                .setConnectableBus2(busAnvers225.getId())
+                .setBus2(busAnvers225.getId())
+                .setVoltageLevel2(vlAnvers225.getId())
```
MioRtia, line 920: Not sure why this change occurred. I will look into it.

zamarrenolm (approved):

annetill (approved):
I have performed a functional validation on IOP dataset but it does not test operational limits. No regression have been found on classical import + powerflow calculation.

MioRtia (approved):

PR #1446, opened 2020-08-19
Empty extension should not be serialized

floriand-e2r:
**Please check if the PR fulfills these requirements** *(please use `'[x]'` to check the checkboxes, or submit the PR and then click the checkboxes)*
- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**Does this PR already have an issue describing the problem ?** *If so, link to this issue using `'#XXX'` and skip the rest*
No

**What kind of change does this PR introduce?** *(Bug fix, feature, docs update, ...)*
If extension is empty then it is not serialized.

**What is the current behavior?** *(You can also link to an open issue here)*
An extension could be empty for multivariant extensions and thus serialized with an empty body.

**What is the new behavior (if this is a feature change)?**
An empty extension is not serialized

**Does this PR introduce a breaking change or deprecate an API?** *If yes, check the following:*
- [ ] The *Breaking Change* or *Deprecated* label has been added
- [ ] The migration guide has been updated in the github wiki *(What changes might users need to make in their application due to this PR?)*

**Other information**:

(if any of the questions/checkboxes don't apply, please delete them entirely)

PR #1448, opened 2020-08-20
Make LoadDetail extension multi variant

MioRtia:

No

Between bug fix and feature

`LoadDetail` is not multivariant

`LoadDetail` is multivariant


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/LoadDetailImpl.java
```diff
@@ -0,0 +1,133 @@
...
+ */
+public class LoadDetailImpl extends AbstractMultiVariantIdentifiableExtension<Load> implements LoadDetail {
+
+    private final TDoubleArrayList fixedActivePower;
+
+    private final TDoubleArrayList fixedReactivePower;
+
+    private final TDoubleArrayList variableActivePower;
+
+    private final TDoubleArrayList variableReactivePower;
```
floriand-e2r, line 25: You could use `TFloatArrayList` instead. Or shouldn't we rather use a ArrayList of POJOs containing the four power values? That would allow a simple solution for variants with undefined LoadDetail, by using a null POJO.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/LoadDetailImpl.java
```diff
@@ -0,0 +1,133 @@
...
+    public void reduceVariantArraySize(int number) {
+        fixedActivePower.remove(fixedActivePower.size() - number, number);
+        fixedReactivePower.remove(fixedReactivePower.size() - number, number);
+        variableActivePower.remove(variableActivePower.size() - number, number);
+        variableReactivePower.remove(variableReactivePower.size() - number, number);
+    }
+
+    @Override
+    public void deleteVariantArrayElement(int index) {
+        // Nothing to do
```
floriand-e2r, line 121: Should it be a TODO? as something needs to be done here one day, doesn't it?

floriand-e2r (changes requested):

iidm/iidm-extensions/src/test/java/com/powsybl/iidm/network/extensions/LoadDetailTest.java
```diff
@@ -6,12 +6,18 @@
  */
 package com.powsybl.iidm.network.extensions;
```
floriand-e2r, line 7: Shouldn't the test be in package com.powsybl.iidm.network.impl.extensions of iidm-impl?

iidm/iidm-extensions/src/test/java/com/powsybl/iidm/network/extensions/LoadDetailTest.java
```diff
@@ -73,4 +79,64 @@ public void test() {
...
+        assertEquals(20f, ld.getFixedReactivePower(), 0f);
+        assertEquals(60f, ld.getVariableActivePower(), 0f);
+        assertEquals(30f, ld.getVariableReactivePower(), 0f);
+
+        // Removes a variant then adds another variant to test variant recycling (hence calling allocateVariantArrayElement)
+        variantManager.removeVariant(variant1);
+        List<String> targetVariantIds = new ArrayList<>();
+        targetVariantIds.add(variant1);
+        targetVariantIds.add(variant3);
+        variantManager.cloneVariant(INITIAL_VARIANT_ID, targetVariantIds);
```
floriand-e2r, line 121: Could be replaced by Arrays.asList(variant1, variant2)

iidm/iidm-extensions/src/test/java/com/powsybl/iidm/network/extensions/LoadDetailTest.java
MioRtia: Not sure about it, because it isn't a test specific to the in-memory implementation (it should also work with a persistent for example) however I was not sure if its place was in iidm-tck so I left it here.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/LoadDetailImpl.java
MioRtia: I used `TFloatArrayList` to be consistent with other similar equipments/implementations

iidm/iidm-extensions/src/test/java/com/powsybl/iidm/network/extensions/LoadDetailTest.java
floriand-e2r: Ah you're right, but then SlackTerminalTest should not be in iidm-impl either :/

floriand-e2r:
I'm wondering, the `Extension.isEmpty()` introduced by PR #1446 will require undefined values in LoadDetail to detect if the whole extension is unused. Should we already prepare it in current PR @MioRtia? Or is in fact a LoadDetail never unused?

MioRtia:
> I'm wondering, the `Extension.isEmpty()` introduced by PR #1446 will require undefined values in LoadDetail to detect if the whole extension is unused. Should we already prepare it in current PR @MioRtia, or does it belong to that other PR? Or is in fact a LoadDetail never unused?

For now, this extension can never be unused (there is no way to put undefined values) so for me, there is no issue about it (it might change if we change this behavior).

floriand-e2r (approved):

PR #1453, opened 2020-09-01
[WIP] Example of aliases and properties for CGMES import/export (start to delete use of mapping in export)

MioRtia:

**Other information**:


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -278,6 +285,17 @@ private void assignNetworkProperties(Context context) {
         LOG.info("network forecastDistance : {}", context.network().getForecastDistance());
     }
 
+    private void addCgmesSvDescription(Network network) {
+        PropertyBags svDescription = cgmes.fullModel(CgmesSubset.STATE_VARIABLES.getProfile());
+        if (svDescription != null && !svDescription.isEmpty()) {
+            network.setProperty(CgmesNames.SCENARIO_TIME, svDescription.get(0).getId("scenarioTime"));
```
zamarrenolm, line 291: if we add all these properties directly to the Network I think we should prefix all of them with "cgmes." or something like that, to avoid potential key collisions. 
All metadata related to CGMES should have that prefix.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/AbstractConductingEquipmentConversion.java
```diff
@@ -422,6 +418,17 @@ public void connect(LegAdder adder, int terminal) {
         }
     }
 
+    protected void addAliases(Identifiable<?> identifiable) {
+        int i = 1;
+        for (TerminalData td : terminals) {
+            if (td == null) {
+                return;
+            }
+            identifiable.addAlias(td.t.id(), CgmesNames.TERMINAL + i);
```
zamarrenolm, line 427: prefix the key for the alias name with "cgmes." ?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractTransformerConversion.java
```diff
@@ -95,4 +97,17 @@ protected CgmesRegulatingControlPhase setContextRegulatingDataPhase(TapChanger t
         }
         return rcPtc;
     }
+
+    @Override
+    protected void addAliases(Identifiable<?> identifiable) {
+        super.addAliases(identifiable);
+        String ptc = context.cgmes().phaseTapChangerForPowerTransformer(identifiable.getId());
+        if (ptc != null) {
+            identifiable.addAlias(ptc, CgmesNames.PHASE_TAP_CHANGER);
```
zamarrenolm, line 106: use "cgmes." prefix ?

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesTerminal.java
```diff
@@ -43,6 +44,7 @@ public CgmesTerminal(PropertyBag t) {
         // (TODO(Luma): missing reference here)
 
         this.connected = t.asBoolean("connected", false);
+        this.inService = t.containsKey("inService") ? Boolean.parseBoolean(t.getId("inService")) : null;
```
zamarrenolm, line 47: In CGMES the `inService` attribute comes from the `SvStatus` object, that is linked to the `ConductingEquipment`, not the `Terminal`. I know the SPARQL for terminals returns the `inService` property of its `ConductingEquipment`, and probably this is the best way of handling it, save it at each terminal. We just have to keep in mind that this inService refers to the status of the equipment, and should be updated when topology is calculated. For branches, if one terminal is disconnected, it does not mean that the equipment is out of service.

From CGMES documentation: _A valid case is that conducting equipment can be connected in one end and open in the other. In particular for an AC line segment, where the reactive line charging can be significant, this is a relevant case._

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesAdder.java
```diff
@@ -101,41 +102,42 @@ public void addStateVariablesToCgmes() {
...
     }
 
     private void addVoltagesForBoundaryNodes() {
         PropertyBags voltages = new PropertyBags();
         // add voltages for TpNodes existing in the Model's boundaries
-        boundaryNodesFromDanglingLines.forEach((line, bnode) -> {
+        for (DanglingLine dl : network.getDanglingLines()) {
             PropertyBag p = new PropertyBag(SV_VOLTAGE_PROPERTIES);
-            DanglingLine dl = network.getDanglingLine(line);
             Bus b = dl.getTerminal().getBusBreakerView().getBus();
```
zamarrenolm: Dangling lines should already have voltage of xnode as properties (v, angle). In PR#1455 I added some code to check if we have these properties, then use them; if not, we could compute voltage from dangling line impedance and known network voltage.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -138,6 +140,11 @@ public Network convert() {
             throw new CgmesModelException("Data source does not contain EquipmentCore data");
         }
         Network network = createNetwork();
+        network.setProperty("CGMES_topology", cgmes.isNodeBreaker() ? "NODE_BREAKER" : "BUS_BRANCH");
```
annetill, line 143: As I said before, I think that properties for a network is not the good solution. We will have issues through the merging view. A slight extension could be better (and documented).

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
annetill: I think that it will not work to export an SV for the CGM model with the merging view. It will select the first property which key is "SCENARIO_TIME".

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ACLineSegmentConversion.java
```diff
@@ -165,7 +166,11 @@ private void convertDanglingLine(int boundarySide) {
                         .setVoltageRegulationOn(false)
                     .add()
                     .add();
+            addAliases(dl);
         }
+        dl.setProperty("boundarySide", String.valueOf(boundarySide));
```
annetill, line 171: If you have a lot of properties, I think that an extension is better for dangling line.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
MioRtia: I agree with Anne, I think this should be in an extension that will be built either during the merging (merge post-processor?) or by the user. I made the PR #1459 to create this extension.

PR #1466, opened 2020-09-10
Add aliases during CGMES import for missing IDs

MioRtia:

No

Feature

Terminals' and tap changer's IDs are lost during CGMES import

These IDs are kept as aliases

No

**Other information**:


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ACLineSegmentConversion.java
```diff
@@ -110,6 +111,7 @@ public void convertMergedLinesAtNode(PropertyBag other, String boundaryNode) {
         } else {
             mline = createQuadripole(boundaryLine1, boundaryLine2);
         }
+        addAliases(mline);
```
zamarrenolm, line 114: Missing addAliases for merged line when line-switch configuration is seen at boundary ?

zamarrenolm (approved):

PR #1467, opened 2020-09-11
Import of SlackTerminal extension during CGMES import (SynchronousMachine.referencePriority)

MioRtia:

No

Feature

The information about slack buses is lost during CGMES import

The information about slack buses is stored in `SlackTerminal` extension during CGMES import

**Other information**:


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/SynchronousMachineConversion.java
```diff
@@ -55,6 +56,13 @@ public void convert() {
         Generator g = adder.add();
         convertedTerminals(g.getTerminal());
         convertReactiveLimits(g);
+        if (p.asInt("referencePriority", 0) > 0) {
+            if (g.getTerminal().getBusBreakerView().getBus() == null) {
+                context.ignored("referencePriority", () -> String.format("%s is disconnected from the network", g.getId()));
+            } else {
+                SlackTerminal.attach(g.getTerminal().getBusBreakerView().getBus());
```
annetill, line 63: I am not sure to understand: why do you attach a bus instead of a bus ?

annetill:
Why it is a breaking change?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/SynchronousMachineConversion.java
```diff
@@ -57,7 +57,11 @@ public void convert() {
         convertedTerminals(g.getTerminal());
         convertReactiveLimits(g);
         if (p.asInt("referencePriority", 0) > 0) {
-            SlackTerminal.attach(g.getTerminal().getBusBreakerView().getBus());
+            if (g.getTerminal().getBusBreakerView().getBus() == null) {
+                context.ignored("referencePriority", () -> String.format("%s is disconnected from the network", g.getId()));
```
zamarrenolm, line 61: Even if the generator is disconnected from the network at this point, it may become eventually connected (if a switch is closed). 
Also, we could keep the information to be able to reproduce in the CGMES SSH export the same Steady State Hypothesis we have seen in the original files.

cgmes/cgmes-conformity/src/test/resources/conformity-modified/cas-1.1.3-data-4.0.3/MiniGrid/BusBranch/BaseCase_v3_disconnected_slack/MiniGridTestConfiguration_BC_SSH_v3.0.0.xml
```diff
@@ -0,0 +1,232 @@
...
+    <cim:ControlArea.netInterchange>0</cim:ControlArea.netInterchange>
+  </cim:ControlArea>
+  <cim:RegulatingControl rdf:about="#_10d33824-507a-4999-bd51-f73373e7509a">
+    <cim:RegulatingControl.discrete>false</cim:RegulatingControl.discrete>
+    <cim:RegulatingControl.enabled>true</cim:RegulatingControl.enabled>
+    <cim:RegulatingControl.targetDeadband>0</cim:RegulatingControl.targetDeadband>
+    <cim:RegulatingControl.targetValue>10.0</cim:RegulatingControl.targetValue>
+    <cim:RegulatingControl.targetValueUnitMultiplier rdf:resource="http://iec.ch/TC57/2013/CIM-schema-cim16#UnitMultiplier.k" />
+  </cim:RegulatingControl>
+</rdf:RDF>
```
zamarrenolm, line 232: missing line end

zamarrenolm (approved):

zamarrenolm:
I think we do not have an extension serializer for the `slackTerminal` extension

zamarrenolm:
I see a problem of current implemented strategy for attaching `slackTerminal` extensions to IIDM Network.

Currently, `slackTerminal` extensions:

- Are attached to voltage levels.
- Only one `slackTerminal` extension is allowed at each voltage level. 
- When an extension is going to be created from an equipment (generator), the terminal stored in the extension is not the terminal of the equipment, but the first connected terminal in the same bus of the original terminal.

This creates a problem when we want to store slack information for disconnected generators (that may be later connected during a simulation). The disconnected generator does not have a valid bus in the original state.

Also, the process for selecting the terminal may change the point of the network that was originally defined for the slack position (from original generator to first connected terminal of the same bus).

zamarrenolm:
I think that we could merge this PR as it is (avoid setting extension for disconnected generators) and then create an issue for the SlackTerminal extension problem described. Do you agree @MioRtia, @annetill ?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/SynchronousMachineConversion.java
annetill, line 63: @MioRtia can you here not use the ````attach```` method but the ````set```` one in provinding directly the connected terminal of the generator ? With that, you can always set the extension, no ?

MioRtia: @zamarrenolm This will solve your third point (the terminal defined will be the terminal of the equipment). For the other point, since I think the case where a voltage level contains two slack terminals of different synchronous components is rare, couldn't we ignore this case? Maybe we should look at how often it happens in real life.

Issue #1435, opened 2020-08-07
[MergingView] Extension management


mathbagu:
* **Do you want to request a *feature* or report a *bug*?**
Bug

* **What is the current behavior?**
For most of the identifiables objects, we currently delegate the call to getExtension. But as the extension can return the identifiable object, they should be wrapped to return the adapted object, to be able to write:
```
assertSame(identifiable, identifiable.getExtension(Extension.class).getIdentifiable());
```
With the current version of the code, the getIdentifiable() will return the adaptee and not the adapter object.

For some objects, extension management is not working at all:
- DanglingLine extensions are not compatible with Line extensions: how to manage the extension for a MergedLine? If we allow the creation of extensions, we will lose them in the export because the MergedLine is transient. 
- The MergingView could delegate access to the underlying network extensions, but it's possible to have the same extension in several networks. In that case, what is the expected behaviour. Same question about extension creation.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**

* **What is the motivation / use case for changing the behavior?**

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

annetill:
Solved by #2442

PR #1469, opened 2020-09-14
[WIP] Proposition of a mechanism to merge extensions

MioRtia:

Answers some of issues addressed in #1435 

Feature

When underlying networks in a `MergingView` have an extension, it is not taken into account.

When underlying networks in a `MergingView` have an extension, it is merged in the merging view.

No

**Other information**:

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/extensions/CimCharacteristicsMergingView.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/extensions/CimCharacteristicsMergingView.java
@@ -8,6 +8,7 @@ package com.powsybl.cgmes.conversion.extensions;
 
 import com.google.auto.service.AutoService;
 import com.powsybl.cgmes.conversion.elements.CgmesTopologyKind;
+import com.powsybl.commons.PowsyblException;
 import com.powsybl.iidm.mergingview.extensions.ExtensionMergingView;
 import com.powsybl.iidm.network.Network;
 
@@ -36,10 +37,17 @@ public class CimCharacteristicsMergingView implements ExtensionMergingView<Netwo
                     .setTopologyKind(extension.getTopologyKind())
                     .add();
         } else {
+            checkCimVersions(original.getCimVersion(), extension.getCimVersion());
             extendable.newExtension(CimCharacteristicsAdder.class)
-                    .setCimVersion(Math.min(original.getCimVersion(), extension.getCimVersion()))
+                    .setCimVersion(extension.getCimVersion())
                     .setTopologyKind(getMergedTopologyKind(original.getTopologyKind(), extension.getTopologyKind()))
                     .add();
+        }
+    }
+
+    private static void checkCimVersions(int cimVersion1, int cimVersion2) {
+        if (cimVersion1 != cimVersion2) {
+            throw new PowsyblException("Networks are from different CIM versions (" + cimVersion1 + " and " + cimVersion2 + ")");
         }
     }
 
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/extensions/CimCharacteristicsMergingView.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/extensions/CimCharacteristicsMergingView.java
@@ -6,6 +6,7 @@
  */
 package com.powsybl.cgmes.conversion.extensions;
 
+import com.google.auto.service.AutoService;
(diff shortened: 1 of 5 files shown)


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/extensions/CimCharacteristicsMergingView.java
```diff
@@ -0,0 +1,52 @@
...
+    public void merge(Network extendable, CimCharacteristics extension) {
+        CimCharacteristics original = extendable.getExtension(CimCharacteristics.class);
+        if (original == null) {
+            extendable.newExtension(CimCharacteristicsAdder.class)
+                    .setCimVersion(extension.getCimVersion())
+                    .setTopologyKind(extension.getTopologyKind())
+                    .add();
+        } else {
+            extendable.newExtension(CimCharacteristicsAdder.class)
+                    .setCimVersion(Math.min(original.getCimVersion(), extension.getCimVersion()))
```
annetill, line 40: I think that the CIM version should be identical. How does it work if not?

MioRtia: 14 is less extensive than 16, so if one of the file is 14, I make the merging view 14.

annetill: But do we really want to support that ? Have we perform some tests with this kind of hybrid association ?

MioRtia: Good question, maybe not. Do you think it should throw an exception in this case?

annetill: Yes, it is better.

annetill (approved):
For me it is okay!

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/extensions/CimCharacteristicsMergingView.java
```diff
@@ -0,0 +1,60 @@
...
+            extendable.newExtension(CimCharacteristicsAdder.class)
+                    .setCimVersion(extension.getCimVersion())
+                    .setTopologyKind(extension.getTopologyKind())
+                    .add();
+        } else {
+            checkCimVersions(original.getCimVersion(), extension.getCimVersion());
+            extendable.newExtension(CimCharacteristicsAdder.class)
+                    .setCimVersion(extension.getCimVersion())
+                    .setTopologyKind(getMergedTopologyKind(original.getTopologyKind(), extension.getTopologyKind()))
+                    .add();
```
mathbagu, line 44: Why don't you update the topology kind of the existing extension, instead of creating a new one?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/extensions/CimCharacteristicsMergingView.java
```diff
@@ -0,0 +1,60 @@
...
+            throw new PowsyblException("Networks are from different CIM versions (" + cimVersion1 + " and " + cimVersion2 + ")");
+        }
+    }
+
+    private static CgmesTopologyKind getMergedTopologyKind(CgmesTopologyKind topologyKind1, CgmesTopologyKind topologyKind2) {
+        if (topologyKind1 == CgmesTopologyKind.NODE_BREAKER || topologyKind2 == CgmesTopologyKind.NODE_BREAKER) {
+            return CgmesTopologyKind.NODE_BREAKER;
+        }
+        return CgmesTopologyKind.BUS_BRANCH;
+    }
```
mathbagu, line 59: Are you sure about that? It's not possible in CGMES to have hybrid network? For CGM, why should not be possible to have some EQ in N/B and some EQ in B/B?
How this topologyKind is used?

mathbagu (changes requested):
I think the design you propose is not OK. Here a list of possible issue:
- what happen if an IGM extension is updated outside the merging view
- what happen if an IGM is added to the merging view (existing MergedExtension are not updated)
- Should the MergeableExtension concept be shared with the destructive merge?
- There is a new dependency between cgmes and the merging view: I'm not sure it's correct.

PR #1473, opened 2020-09-17
Modify limit violation builder copy

EtienneLt:

update

there is no method to read the LimitViolations
for voltage limitViolations, builder can not be used because limitName parameter is needed 

added a method toString to print the limitviolation
added limitName in build method of voltage limitViolation 

**Other information**:

### DIFF
--- a/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/LimitViolation.java
+++ b/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/LimitViolation.java
@@ -14,7 +14,6 @@ import javax.annotation.Nullable;
 import java.util.Objects;
 
 /**
- *
  * A generic representation of a violation of a network equipment security limit.
  * For example, it may represent a current overload on a line.
  *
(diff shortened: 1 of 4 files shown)


security-analysis/security-analysis-api/src/main/java/com/powsybl/security/LimitViolation.java
```diff
@@ -211,4 +213,16 @@ public double getValue() {
             return null;
         }
     }
+
+    public String toString() {
+        String base = "Subject id: " + this.subjectId + ", Subject name: " + this.subjectName + ", limitType: " +
+                this.limitType.toString() + ", limit: " + this.limit + ", limitName: " + this.limitName +
```
mathbagu, line 219: There is no need to call toString. If limitType is null, you would have a NullPointerException whereas it works perfectly without it

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/LimitViolation.java
```diff
@@ -211,4 +213,16 @@ public double getValue() {
...
+                ", acceptableDuration: " + this.acceptableDuration + ", limitReduction: " + this.limitReduction +
+                ", value: " + this.value;
+        if (side != null) {
+            return base + ", side: " + this.side.toString();
+        } else {
+            return base + ", side: null";
+        }
```
mathbagu, line 226: Great: you put an exemple of what I said just before!

Simply write something like this : ` + ", side: " + this.side`

mathbagu (changes requested):
This PR is OK except the two small remarks

mathbagu (approved):

PR #1476, opened 2020-09-21
Fix missing call to MultiVariantObject implementations

mathbagu:

N/A

Bug fix

In several IIDM class, that implement MultiVariantObject, the super method is not called. For most of them, their first parent implementation is AbstractIdentifiable that manage the extensions.
That means, if we have an extension that supports variant for these objects, it wouldn't work properly. A similar fix has been done in #1404 but not everywhere

Fix the implementation of all multivariantobjects

**Other information**:


MioRtia (approved):

mathbagu:
@MioRtia Is it ok for you if I force-merge?
- I can easily fix 2 smells (Remove this method to simply inherit it), but it's done like this everywhere
- I cannot fix the duplication (there is no link between Network and BusBreakerVoltageLevel), so it's a false positve

PR #1479, opened 2020-09-22
improve getCountry method to avoid nullPointerException when geographicalCode is not null and country is null

amirakahya:
Signed-off-by: KAHYA Amira <amira.kahya@rte-france.com>

NO

Bug fix

NullPointerException when the geographicalCode is not null but the Country is null. This is the case when the geographicalCode is UC or UX.

Return null when the geographicalCode is not null and the Country is Null.

### DIFF
--- a/entsoe-util/src/main/java/com/powsybl/entsoe/util/EntsoeFileName.java
+++ b/entsoe-util/src/main/java/com/powsybl/entsoe/util/EntsoeFileName.java
@@ -77,7 +77,11 @@ public class EntsoeFileName {
     }
 
     public String getCountry() {
-        return geographicalCode != null ? geographicalCode.getCountry().toString() : null;
+        if (geographicalCode != null) {
+            return this.geographicalCode.getCountry() != null ? this.geographicalCode.getCountry().toString() : null;
+        } else {
+            return null;
+        }
     }
 
 }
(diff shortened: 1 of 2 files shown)


murgeyseb (changes requested):
Could you please add a test case to validate the bug correction.

entsoe-util/src/main/java/com/powsybl/entsoe/util/EntsoeFileName.java
```diff
@@ -77,7 +77,11 @@ public EntsoeGeographicalCode getGeographicalCode() {
     }
 
     public String getCountry() {
-        return geographicalCode != null ? geographicalCode.getCountry().toString() : null;
```
murgeyseb, line 80: Perhaps a more dense implementation could be better ?
```java
return geographicalCode != null && geographicalCode.getCountry() != null ? geographicalCode.getCountry().toString() : null;
```
@mathbagu what do you think about it ?

mathbagu: Agree: @amirakahya could you change the code as @murgeyseb suggest.

For you information, it's not necessary to prefix geographicalCode with `this` like you did. Use this syntax only if there is a clash between a variable and an attribute of the class. For instance, it can happen in a setter:
```
public void setX(int x) {
    this.x = x;
}
```

mathbagu (changes requested):

mathbagu (approved):

PR #1480, opened 2020-09-22
Integration of the new way to export SV profile in CGMES export

MioRtia:

No

Feature

To do a CGMES export, the CGMES model (access to the original triplestore) is needed.

To do the CGMES export of the SV file, only the IIDM network is needed.

No

**Other information**:


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesExport.java
```diff
@@ -0,0 +1,408 @@
...
+            writer.writeAttribute(RDF_NAMESPACE, CgmesNames.RESOURCE, "#" + angleRefs.get(island.getKey()));
+            for (String tn : island.getValue()) {
+                writer.writeEmptyElement(CIM_NAMESPACE, "TopologicalIsland.TopologicalNodes");
+                writer.writeAttribute(RDF_NAMESPACE, CgmesNames.RESOURCE, "#" + tn);
+            }
+            writer.writeEndElement();
+        }
+    }
+
+    private static Map<String, String> buildAngleRefs(Network network) {
```
annetill, line 147: Maybe put as a comment that it works only for bus branch topology?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesExport.java
```diff
@@ -0,0 +1,408 @@
...
+                } else {
+                    Supplier<String> message = () -> String.format("Slack terminal at equipment %s is not connected and is not exported as slack terminal", slackTerminal.getTerminal().getConnectable().getId());
+                    LOG.info(message.get());
+                }
+            }
+        }
+        return angleRefs;
+    }
+
+    private static Map<String, List<String>> buildIslands(Network network) {
```
annetill, line 173: Maybe put a comment to say that it works only for bus branch topology.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesExport.java
```diff
@@ -0,0 +1,408 @@
...
+        }
+    }
+
+    private static void writeVoltagesForBoundaryNodes(Network network, XMLStreamWriter writer) throws XMLStreamException {
+        for (DanglingLine dl : network.getDanglingLines()) {
+            Bus b = dl.getTerminal().getBusBreakerView().getBus();
+            Optional<String> topologicalNode = dl.getAliasFromType(CGMES_PREFIX_ALIAS + CgmesNames.TOPOLOGICAL_NODE);
+            if (topologicalNode.isPresent()) {
+                if (dl.hasProperty("v") && dl.hasProperty("angle")) {
+                    writeVoltage(topologicalNode.get(), Double.valueOf(dl.getProperty("v", "NaN")), Double.valueOf(dl.getProperty("angle", "NaN")), writer);
```
annetill, line 204: I am not sure to get: what do you do if you have a NaN as v and angle? If you have that, I think that it could be nice to try to re-compute them as you do 2 lines below.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesExport.java
```diff
@@ -0,0 +1,408 @@
...
+
+    private static void writePowerFlows(Network network, XMLStreamWriter writer) {
+        writeInjectionPowerFlowToCgmes(network, writer, Network::getLoadStream);
+        writeInjectionPowerFlowToCgmes(network, writer, Network::getGeneratorStream);
+        writeInjectionPowerFlowToCgmes(network, writer, Network::getShuntCompensatorStream);
+        writeInjectionPowerFlowToCgmes(network, writer, Network::getStaticVarCompensatorStream);
+        writeInjectionPowerFlowToCgmes(network, writer, Network::getBatteryStream);
+
+        network.getDanglingLineStream().forEach(dl -> dl.getAliasFromType(CGMES_PREFIX_ALIAS + "Terminal_Boundary")
+                .ifPresent(terminal -> writePowerFlow(terminal, dl.getP0(), dl.getQ0(), writer)));
```
annetill, line 248: If you want to write here the updated values of P and Q at boundary point, I think that you cannot make it like that. In IIDM, after a power flow calculation, the only think that changes is P and Q of the terminal connected to the network. It means that we need to introduce new properties to be set by the power flow engine or re-computed them. This value will be use to update EquivalentInjection too. @zamarrenolm do you confirm?

annetill: And it will solve the question below: and what about danglignq line with generation part?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesExport.java
```diff
@@ -0,0 +1,408 @@
...
+        // TODO what about flows of dl's generations?
+    }
+
+    private static <I extends Injection<I>> void writeInjectionPowerFlowToCgmes(Network network, XMLStreamWriter writer, Function<Network, Stream<I>> getInjectionStream) {
+        getInjectionStream.apply(network).forEach(i -> writePowerFlow(i.getTerminal(), writer));
+    }
+
+    private static void writePowerFlow(Terminal terminal, XMLStreamWriter writer) {
+        String cgmesTerminal = ((Connectable<?>) terminal.getConnectable()).getAliasFromType(CGMES_PREFIX_ALIAS + CgmesNames.TERMINAL1).orElse(null);
```
annetill, line 257: And the TERMINAL2 ?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesExport.java
```diff
@@ -0,0 +1,408 @@
...
+    }
+
+    // Avoid trailing zeros
+    private static final DecimalFormat DOUBLE_FORMAT = new DecimalFormat("0.##############", new DecimalFormatSymbols(Locale.US));
+
+    private static String fs(double value) {
+        return Double.isNaN(value) ? DOUBLE_FORMAT.format(0.0) : DOUBLE_FORMAT.format(value);
+    }
+
+    private static Complex complexVoltage(double r, double x, double g, double b,
```
annetill, line 393: I am not sure to get why we introduce a new function here ? We already have a function to do that ? Or it is not the good place maybe.

annetill (commented):
I am going to make a functional validation but it could be great that you see my questions from now on. @zamarrenolm some questions are for you too!

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesExport.java
```diff
@@ -0,0 +1,408 @@
...
+        writer.writeEndElement();
+        writer.writeStartElement(MD_NAMESPACE, CgmesNames.VERSION);
+        writer.writeCharacters(is(context.getSvVersion()));
+        writer.writeEndElement();
+        for (String dependency : context.getDependencies()) {
+            writer.writeEmptyElement(MD_NAMESPACE, CgmesNames.DEPENDENT_ON);
+            writer.writeAttribute(RDF_NAMESPACE, CgmesNames.RESOURCE, dependency);
+        }
+        writer.writeStartElement(MD_NAMESPACE, CgmesNames.PROFILE);
+        writer.writeCharacters("http://entsoe.eu/CIM/StateVariables/4/1");
```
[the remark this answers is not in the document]
MioRtia: Maybe it has to be confirmed with @zamarrenolm but it seems that this bit is not something that can be changed... We could make it a parameter though for possible future versions I guess

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesExport.java
MioRtia: Same remark

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesExport.java
MioRtia: I'm not sure if this is the place to recompute. We should ignore NaN values though, I'm modifying the code to do this.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesExport.java
MioRtia: This method is only used for injections which only have one terminal, which is why terminal2 is never called. I'm going to check what is done for branches/transformers (do you know @zamarrenolm @marqueslanauja ?)

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesExport.java
MioRtia: Edit: when the value is NaN, we put 0.0 as the CGMES convention requires.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesExport.java
MioRtia: You're right, I think this is only valid if generation is zero... @zamarrenolm Can you confirm?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/AbstractConnectorConversion.java
```diff
@@ -99,7 +99,8 @@ public void convertToDanglingLine(int boundarySide, double r, double x, double g
                 .add()
                 .add();
         }
-        addAliases(dl);
+        dl.addAlias(terminalId(boundarySide), "CGMES.Terminal_Boundary");
+        dl.addAlias(terminalId(boundarySide == 1 ? 2 : 1), "CGMES.Terminal_Network");
```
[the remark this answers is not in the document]
zamarrenolm: For the dangling line, the terminals of the _network side_ (the modelling authority side) and the _boundary side_ (the side that lies in the boundary) have to be stored separately.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/AbstractConnectorConversion.java
```diff
@@ -99,7 +99,8 @@ public void convertToDanglingLine(int boundarySide, double r, double x, double g
...
         dl.addAlias(boundaryNode, "CGMES." + CgmesNames.TOPOLOGICAL_NODE);
```
zamarrenolm, line 104: prefix `CGMES.` should be defined in `Conversion` module and used here and during export, to query the aliases

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesExport.java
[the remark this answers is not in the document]
zamarrenolm: could be defined as a constant in the same module, for now

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesExport.java
zamarrenolm: The implementation is valid also if the IIDM model is node-breaker

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesExport.java
zamarrenolm: The flow at the boundary terminal is equal to the generation of the dangling line, but with opposite sign (`p0`, `q0` are the values of the injection, with load sign convention, and the flow in the dangling line from the boundary to the network side is the same quantity, also with load sign convention)

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesExport.java
```diff
@@ -0,0 +1,408 @@
...
+            writer.writeAttribute(RDF_NAMESPACE, CgmesNames.RESOURCE, "#" + terminal);
+            writer.writeEndElement();
+        } catch (XMLStreamException e) {
+            throw new UncheckedXmlStreamException(e);
+        }
+    }
+
+    private static void writeSvInjection(Load svInjection, XMLStreamWriter writer) {
+        try {
+            writer.writeStartElement(CIM_NAMESPACE, "SvInjection");
```
zamarrenolm, line 288: in CGMES `SvInjection` is reserved for reporting mismatches in buses, not for reporting values of `EquivalentInjections` (of boundary nodes). The value of the equivalent injection at the boundary must be entered as a `SvPowerFlow` value at the corresponding `Terminal` of the `EquivalentInjection`. Change will be proposed in a later commit

zamarrenolm: Writing `SvInjections` is ok, as we have converted to fictitious loads only the values that we have seen in the original CGMES input as `SvInjections` also. Sorry for the noise.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesExport.java
zamarrenolm: usually the flows for branches are not put in the SV output (they can always be easily recalculated, given the voltages and the impedances). We could export also these values, could be configurable.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesExport.java
annetill: Ok !

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesExport.java
annetill: Update: after a powerflow calculation, P0 and Q0 are considered as targets and are not updated in IIDM. But, after a balance ajustement, it should be updated and so, after topological merge + balance ajustement, these lines of codes will provide updated values.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesExport.java
MioRtia: I'm adding a `TODO` about this but I think we can not export them at first

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesExport.java
MioRtia: After discussing it,  it seems a mistake to export `p0` and `q0` as calculated flows. We should introduce an object called `middleBus` or something which will contain the calculated flows, voltage and angle of the boundary fictitious terminal of the dangling line. For now, I think we should delete this line and put a `TODO`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesExport.java
MioRtia: I put it in `CgmesExportUtil`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/update/StateVariablesExport.java
MioRtia: (I did not delete the line but put a `FIXME`)

zamarrenolm (approved):
passing away the CIM namespace through all the methods is a bit hard.

In the future maybe it could be stored in the context, or all these methods should not be static, and we may have some small internal state in the exporter, where we could save it.

Depending on the cim version we are exporting, some elements/attribute names should be reviewed (for example the position in the SvShuntCompensator).

PR #1485, opened 2020-09-25
Prepare for integration: CGMES SSH export using only Network data

zamarrenolm:

No

Feature

Current CGMES export requires the CGMES model in memory (access to the original triplestore).

CGMES export of the SSH (Steady State Hypothesis) can be made only using IIDM Network.

No


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -27,22 +38,50 @@
...
-
-        // Right now the network must contain the original CgmesModel
-        // In the future it should be possible to export to CGMES
-        // directly from an IIDM Network,
-        // without the need for the original CgmesModel
+        Objects.requireNonNull(network);
         CgmesModelExtension ext = network.getExtension(CgmesModelExtension.class);
-        if (ext == null) {
-            throw new CgmesModelException("No extension for CGMES model found in Network");
+        if (params != null && Boolean.valueOf(params.getProperty("cgmes.export.usingOnlyNetwork"))) {
```
MioRtia, line 43: `params` is never `null`. Furthermore, you should use  `ConversionParameters.readBooleanParameter` here.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -27,22 +38,50 @@
...
+            if (ext != null) {
+                CgmesModel cgmesSource = ext.getCgmesModel();
+                if (cgmesSource != null) {
+                    throw new CgmesModelException("CGMES model should not be available as Network extension");
```
MioRtia, line 47: Not sure why an exception is thrown? We could just ignore it.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -27,22 +38,50 @@
...
+            }
+            exportUsingOriginalCgmesModel(network, ds, ext);
         }
-        CgmesUpdate cgmesUpdate = ext.getCgmesUpdate();
+    }
 
+    private void exportUsingOnlyNetwork(Network network, DataSource ds) {
+        // At this point only SSH, SV can be exported when relying only in Network data
+        // (minimum amount of CGMES references are expected as aliases/properties/extensions)
+        String baseName = network.hasProperty("baseName") ? network.getProperty("baseName") : network.getNameOrId();
```
MioRtia, line 62: Don't really think it should be a property? I think it should use a parameter `cgmes.export.base-name` (if not filled, use `network.getNameOrId()`). Ideally, it should also be a field in `CgmesExportContext` but I will probably do it in a later PR.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -27,22 +38,50 @@
...
+        String filenameSv = baseName + "_SV.xml";
+        String filenameSsh = baseName + "_SSH.xml";
+        CgmesExportContext context = new CgmesExportContext(network);
+        try (OutputStream os = ds.newOutputStream(filenameSv, false)) {
+            XMLStreamWriter writer = XmlUtil.initializeWriter(true, "    ", os);
+            StateVariablesExport.write(network, writer, context);
+        } catch (IOException | XMLStreamException x) {
+            throw new PowsyblException("Exporting to CGMES using only Network");
+        }
```
MioRtia, line 71: I don't really agree with throwing a `PowsyblException` in any case. You should do:
```java
catch (IOException e) {
} catch (XMLStreamException e) {
    throw new UncheckedXmlStreamException(e);
}
```

You can write this only once at the end of the method (not twice)

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -306,6 +311,34 @@ private void addCgmesSvMetadata(Network network) {
...
+            sshControlAreas.forEach(sshControlArea -> {
+                String id = sshControlArea.getId("ControlArea");
+                double netInterchange = sshControlArea.asDouble("netInterchange");
+                double pTolerance = sshControlArea.asDouble("pTolerance");
+                ControlArea controlArea = new ControlArea(id, netInterchange, pTolerance);
+                adder.addControlArea(controlArea);
+            });
+            adder.add();
+        }
+    }
```
MioRtia, line 340: To confirm, but maybe it might be a good idea to use the extension in `powsybl-balance-adjustements `. I will look into it.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForGenerators.java
```diff
@@ -117,6 +117,9 @@ private boolean setRegulatingControlVoltage(String controlId,
                     .add();
         }
 
+        gen.setProperty("RegulatingControl", controlId);
+        gen.setProperty("targetDeadBand", String.valueOf(control.getTargetDeadBand()));
```
MioRtia, line 121: Is it useful? If the `RegulatingControl` only regulates generators, we can put the targetDeadband as 0.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMapping.java
```diff
@@ -58,7 +58,7 @@ public RegulatingControlMappingForStaticVarCompensators forStaticVarCompensators
         return regulatingControlMappingForStaticVarCompensators;
     }
 
-    static class RegulatingControl {
+    public static class RegulatingControl {
```
MioRtia, line 61: Just for me, the modifications in this class regards the updating of the original SSH export? If yes, it may be easier to review in another PR...

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForShuntCompensators.java
```diff
@@ -57,10 +57,15 @@ private void apply(ShuntCompensator shuntCompensator, CgmesRegulatingControlForS
         if (cgmesRc == null) {
             return;
         }
+        String rcId = cgmesRc.regulatingControlId;
+
+        // We are saving the relationship with the
+        // original regulating control even if it is not enabled
+        // or if we are not able to set it properly in IIDM
         if (!cgmesRc.controlEnabled) {
+            setAdditionalProperties(shuntCompensator, rcId);
```
MioRtia, line 66: Not sure it is warranted. Is it useful to keep information that are not consistent? I understand it is useful for debug purposes but is there another usecase where this information is meaningful?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForShuntCompensators.java
```diff
@@ -95,6 +100,21 @@ private void setRegulatingControl(ShuntCompensator shuntCompensator, RegulatingC
...
         // Take default terminal if it has not been defined in CGMES files (it is never null)
         shuntCompensator.setRegulatingTerminal(parent.getRegulatingTerminal(shuntCompensator, rc.cgmesTerminal));
+
+        shuntCompensator.setProperty("RegulatingControl", rcId);
+    }
+
+    private void setAdditionalProperties(ShuntCompensator shuntCompensator, String rcId) {
+        if (rcId == null) {
+            return;
+        }
```
MioRtia, line 110: Does it happen when this method is called?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForShuntCompensators.java
```diff
@@ -95,6 +100,21 @@ private void setRegulatingControl(ShuntCompensator shuntCompensator, RegulatingC
...
+        RegulatingControl rc = parent.cachedRegulatingControls().get(rcId);
+        if (rc == null) {
+            return;
```
MioRtia, line 113: Does it happen when this method is called?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForShuntCompensators.java
```diff
@@ -95,6 +100,21 @@ private void setRegulatingControl(ShuntCompensator shuntCompensator, RegulatingC
...
+        }
+        shuntCompensator.setProperty("RegulatingControl", rcId);
+        shuntCompensator.setProperty("targetValue", String.valueOf(rc.targetValue));
+        shuntCompensator.setProperty("targetDeadBand", String.valueOf(rc.targetDeadband));
```
MioRtia, line 117: See questions above

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForStaticVarCompensators.java
```diff
@@ -123,6 +123,8 @@ private boolean setRegulatingControl(CgmesRegulatingControlForStaticVarCompensat
         }
         svc.setRegulatingTerminal(terminal);
 
+        svc.setProperty("RegulatingControl", rc.regulatingControlId);
+        svc.setProperty("targetDeadBand", String.valueOf(control.getTargetDeadBand()));
```
MioRtia, line 127: Same remark as for generators: if a regulating control only regulates svc, maybe we can considerate the deadband is 0 and not keep this information from the CGMES file.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/SynchronousMachineConversion.java
```diff
@@ -60,6 +60,14 @@ public void convert() {
         if (p.asInt("referencePriority", 0) > 0) {
             SlackTerminal.reset(g.getTerminal().getVoltageLevel(), g.getTerminal());
         }
+        if (p.containsKey("normalPF")) {
+            // Convert to double then to string to have canonical representation of number
+            g.setProperty("GeneratingUnit.normalPF", Double.valueOf(p.get("normalPF")).toString());
+        }
+        String generatingUnit = p.getId("GeneratingUnit");
+        if (generatingUnit != null) {
+            g.setProperty("GeneratingUnit", generatingUnit);
```
MioRtia, line 69: Just for me, can a generating unit be linked to several synchronous machines?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractTransformerConversion.java
```diff
@@ -120,4 +122,81 @@ protected void addAliases(Identifiable<?> identifiable) {
             }
         }
     }
+
+    private static RegulatingControl getRegulatingControl(Context context, String regulatingControlId) {
+        Objects.requireNonNull(regulatingControlId);
```
MioRtia, line 127: This method is private. If we are sure the regulatingControlId is not null when the method is called, it is not useful for me to check it.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractTransformerConversion.java
```diff
@@ -120,4 +122,81 @@ protected void addAliases(Identifiable<?> identifiable) {
...
+        }
+
+        if (rtc.getRegulatingControlId() != null) {
+            String key = String.format("RatioTapChanger.%s.TapChangerControl", rtc.getId());
+            transformer.setProperty(key, rtc.getRegulatingControlId());
+        }
+
+        if (rtc.getHiddenCombinedTapChanger() != null) {
+            defineHiddenTapChangerProperties(transformer, rtc, rtc.getHiddenCombinedTapChanger(), "RatioTapChanger", context);
+        }
```
MioRtia, line 148: Do you need this because it is done after combination? Can't it be done the same way for hidden and non hidden tap changers?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
```diff
@@ -0,0 +1,177 @@
...
+
+    private boolean exportBoundaryPowerFlows = false;
+
+    public static final class ModelDescription {
+
+        private String description = "Model";
+        private int version = 1;
+        private final List<String> dependencies = new ArrayList<>();
+        private String modelingAuthoritySet = "powsybl.org";
+        // TODO Each model may have a list of profiles, not only one
```
MioRtia, line 43: Can you explain this comment to me?

commons/src/test/java/com/powsybl/commons/AbstractConverterTest.java
```diff
@@ -9,6 +9,7 @@
 import com.google.common.io.ByteStreams;
 import com.google.common.jimfs.Configuration;
 import com.google.common.jimfs.Jimfs;
+
```
MioRtia, line 12: Delete this empty line

MioRtia (commented):
Thank you for your work! I made a first technical review (I have not had the time to do functional tests yet). I have a lot of questions so don't hesitate to explain to me if there are things I have not understood.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
zamarrenolm: `params` may be null when called from specific unit tests. A Conversion Parameter has been added to the `CgmesExport` class. By the way, the generic `Exporter` interface does not define `getParameters` like its `Importer` counterpart.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
zamarrenolm: Check has been removed.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
zamarrenolm: Defined a new `BASE_NAME` parameter in `CgmesExport`.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
zamarrenolm: Done. Defined a single try-with-resources block for export SV and SSH export

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
zamarrenolm: We will analyse the `powsybl-balance-adjustements` extension for a later refactoring.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForGenerators.java
zamarrenolm: We tried to keep the original value read from input data, even it is not applicable in IIDM for generators. Some regulating controls in conformity test cases have `targetDeadband != 0`.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMapping.java
zamarrenolm: Yes the modification is related to SSH export. The change was very small (only visibility of inner class), so we decided to include it here.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForShuntCompensators.java
zamarrenolm: We want to keep original regulating control information without changing the current logic about its use in IIDM.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForShuntCompensators.java
zamarrenolm: yes it could happen

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForShuntCompensators.java
zamarrenolm: yes, it could happen; we keep the same logic that when regulating control is processed

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForShuntCompensators.java
zamarrenolm: trying to keep original data about regulating control

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForStaticVarCompensators.java
zamarrenolm: trying to keep original data about regulating control (the conformity tests contain examples of deadband != 0)

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/SynchronousMachineConversion.java
zamarrenolm: Yes, a `GeneratingUnit` can contain one or more synchronous machines. From CGMES doc, a `GeneratingUnit` is defined as _A single or set of synchronous machines for converting mechanical power into alternating-current power_.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractTransformerConversion.java
zamarrenolm: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractTransformerConversion.java
zamarrenolm: they must be done in a different way for hidden and not-hidden. Hidden after combination do not exist in IIDM Network, and non-hidden properties are taken from Network object.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
zamarrenolm: This is a comment relevant if we try to build a generic `ModelDescription` class. In general a single file CGMES XML can contain references to multiple profiles. For example an `EQ` file may contain references to `EquipmentBoundary`, `EquipmentCore`, `EquipmentOperation` (if it is node-breaker), `ShortCircuit`, ...

commons/src/test/java/com/powsybl/commons/AbstractConverterTest.java
zamarrenolm: Removed

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/SynchronousMachineConversion.java
```diff
@@ -60,6 +60,14 @@ public void convert() {
         if (p.asInt("referencePriority", 0) > 0) {
             SlackTerminal.reset(g.getTerminal().getVoltageLevel(), g.getTerminal());
         }
+        if (p.containsKey("normalPF")) {
+            // Convert to double then to string to have canonical representation of number
+            g.setProperty("GeneratingUnit.normalPF", Double.valueOf(p.get("normalPF")).toString());
```
annetill, line 65: We don't have an extension for that ? What is the difference with ````ActivePowerControl```` ?

annetill: Do we really need a property ? Increase an extension ?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractTransformerConversion.java
```diff
@@ -120,4 +121,80 @@ protected void addAliases(Identifiable<?> identifiable) {
...
+
+        if (ptc.getHiddenCombinedTapChanger() != null) {
+            defineHiddenTapChangerProperties(transformer, ptc, ptc.getHiddenCombinedTapChanger(), "PhaseTapChanger", context);
+
+            String key = String.format("PhaseTapChanger.%s.type", ptc.getHiddenCombinedTapChanger().getId());
+            transformer.setProperty(key, ptc.getHiddenCombinedTapChanger().getType());
+        }
+    }
+
+    private static void defineHiddenTapChangerProperties(Identifiable<?> transformer, TapChanger tc, TapChanger hiddenTc, String propertyTag, Context context) {
```
annetill, line 171: Why do we need all these properties ? We have most of these information in IIDM tap changer, no ?

annetill: We have to use IIDM model in priority.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/TapChanger.java
```diff
@@ -173,6 +180,11 @@ public TapChanger setTapChangerControlEnabled(boolean tapChangerControlEnabled)
         return this;
     }
 
+    public TapChanger setHiddenCombinedTapChanger(TapChanger hiddenCombinedTapChanger) {
```
annetill, line 183: For hidden combined tap changer: see my comment about the two modes of the conversion. I am not sure we have to support that in the mode ````USING_ONLY_NETWORK_PARAMETER````.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/extensions/CgmesSshControlAreas.java
```diff
@@ -0,0 +1,27 @@
...
+import com.powsybl.commons.extensions.Extension;
+import com.powsybl.iidm.network.Network;
+
+import java.util.List;
+
+/**
+ * @author Luma Zamarreño <zamarrenolm at aia.es>
+ * @author José Antonio Marqués <marquesja at aia.es>
+ */
+public interface CgmesSshControlAreas extends Extension<Network> {
```
annetill, line 19: I am really not sure about this extension as we don't have the use case.Maybe we have to accept to not support these fields in the beginning.

annetill (changes requested):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractTransformerConversion.java
MioRtia: @zamarrenolm How much does it happen?
Another thing: for a hidden tap changer, I would rather have an extension than a lot of properties...

MioRtia: @zamarrenolm If it doesn't happen a lot **and** if there is no physical sense, we can afford to lose them (maybe with a warning?)

MioRtia (commented):
Just to sum up, I think for this PR:
- hidden tap changers attributes should be ignored (maybe with a warning?). Furthermore, it seems to me you put the same fields as in input, what happens when the step has been changed? It doesn't change on the hidden tap changer? It is too error-prone for me (and "hidden" errors so to speak, which would be difficult to debug because it *seems* correct). The hidden tap changer ID should be an alias of the transformer I think. Either ignore it during the export or put the same attributes as the non-hidden tap changer in my opinion.
- regulatingControl properties should just be the ID of the regulatingControl (other attributes fetched by going throught the IIDM equipment)
- control areas should be ignored (maybe with an "unsupported" warning?)
- `GeneratingUnit.normalPF` might be an `ActivePowerControl` attribute

In parallel, it might be useful to do the SSH update with the triplestore if we really want an output SSH file with identical fields as the input SSH file but this is not the philosophy for this export: I really don't like that a whole equipment is "hidden" in IIDM property...
We must be in best effort to generate a correct file, not to be identical to the input SSH.

zamarrenolm:
> Just to sum up, I think for this PR:
> 
> * hidden tap changers attributes should be ignored (maybe with a warning?). Furthermore, it seems to me you put the same fields as in input, what happens when the step has been changed? It doesn't change on the hidden tap changer? It is too error-prone for me (and "hidden" errors so to speak, which would be difficult to debug because it _seems_ correct). The hidden tap changer ID should be an alias of the transformer I think. Either ignore it during the export or put the same attributes as the non-hidden tap changer in my opinion.
> * regulatingControl properties should just be the ID of the regulatingControl (other attributes fetched by going throught the IIDM equipment)
> * control areas should be ignored for now (maybe with an "unsupported" warning?): we have to think about how to model them in IIDM
> * `GeneratingUnit.normalPF` might be an `ActivePowerControl` attribute
> 
> In parallel, it might be useful to do the SSH update with the triplestore if we really want an output SSH file with identical fields as the input SSH file but this is not the philosophy for this export: I really don't like that a whole equipment is "hidden" in IIDM property...
> We must be in best effort to generate a correct file, not to be identical to the input SSH.

Requested changes have been applied.

- For combined (hidden) tap changers only original tap position and type are stored as properties. Combined IIDM tap changer result from building equivalent model of two tap changers in the same two-windings transformer. One of them keeps all tap position and for the other one (the hidden one) its current tap position is fixed. To be able to properly export the IIDM equivalent model, this is the minimum amount of information that must be kept.
- For regulating controls, only the identifier of the regulating control is stored as an attribute of the equipment. Rest of data for CGMES export (`discrete`, `targetDeadband`, `targetVoltage`, ...) is obtained from Network objects. When a regulating control is shared between multiple equipment, data taken from every IIDM equipment regulation is obtained and combined before being exported. As an example, a unit test has been created where a voltage regulating control is shared between shunt compensator and generator. 
- Control areas have been removed. They are not considered for export.
- `GeneratingUnit.normalPF` (economic participation factor) has been linked to the attribute `droop` of `ActivePowerControl` extension.

Unit tests have been updated to ignore some differences when comparing original SSH with result of SSH export. As an example, if a voltage regulating control with target deadband != 0 in the original SSH is used only in a generator, the value for target deadband is not stored in the IIDM Network (generators have continuous control and do not need a deadband), so it can not be recovered in the SSH export. So differences in these elements in the compared XML files must be ignored.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
```diff
@@ -0,0 +1,577 @@
...
+            writer.writeEndElement();
+            writer.writeStartElement(cimNamespace, "RotatingMachine.p");
+            writer.writeCharacters(CgmesExportUtil.format(g.getTerminal().getP()));
+            writer.writeEndElement();
+            writer.writeStartElement(cimNamespace, "RotatingMachine.q");
+            writer.writeCharacters(CgmesExportUtil.format(g.getTerminal().getQ()));
+            writer.writeEndElement();
+            writer.writeStartElement(cimNamespace, "SynchronousMachine.referencePriority");
+            // reference priority is used for angle reference selection (slack)
+            writer.writeCharacters(isInSlackBus(g) ? "1" : "0");
```
annetill, line 181: You cannot use the `slackTerminal` extension here ?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractTransformerConversion.java
```diff
@@ -109,15 +109,41 @@ protected void addAliases(Identifiable<?> identifiable) {
...
+        }
+        TapChanger tch = tc.getHiddenCombinedTapChanger();
+        if (tch != null) {
+            // All the tap changers have already been added as aliases,
+            // Through properties we only label which one has been combined and kept hidden
+            transformer.setProperty(cgmesReferenceKey(tc.getId(), "hiddenTapChangerId"), tch.getId());
+            transformer.setProperty(cgmesReferenceKey(tch.getId(), "step"), String.valueOf(tch.getTapPosition()));
+            if (tch.getType() != null) {
+                transformer.setProperty(cgmesReferenceKey(tch.getId(), "type"), tch.getType());
+            }
```
MioRtia, line 142: Why is it necessary?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForTransformers.java
```diff
@@ -64,9 +64,10 @@ public CgmesRegulatingControlRatio buildRegulatingControlRatio(String id, String
         return rtc;
     }
 
-    public CgmesRegulatingControlPhase buildRegulatingControlPhase(String regulatingControlId,
+    public CgmesRegulatingControlPhase buildRegulatingControlPhase(String id, String regulatingControlId,
```
MioRtia, line 67: I didn't find out where you added the regulatingControlId of transformers?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
```diff
@@ -0,0 +1,577 @@
...
+            switch (s.getModelType()) {
+                case LINEAR:
+                    linearNonlinear = "Linear";
+                    break;
+                case NON_LINEAR:
+                    linearNonlinear = "Nonlinear";
+                    break;
+                default:
+                    linearNonlinear = "";
+                    break;
```
MioRtia, line 138: Is this case supposed to happen? If not, throw an AssertionError

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
```diff
@@ -0,0 +1,577 @@
...
+    private static void writeTapChanger(Identifiable<?> eq, String tcId, TapChanger<?, ?> tc, String defaultType, Map<String, List<RegulatingControlView>> regulatingControlViews, String cimNamespace, XMLStreamWriter writer) throws XMLStreamException {
+        String type = eq.getProperty(cgmesTapChangerReferenceKey(tcId, "type"), defaultType);
+        writeTapChanger(type, tcId, tc, cimNamespace, writer);
+        addRegulatingControlView(tc, tcId, eq, regulatingControlViews);
+        writeHiddenTapChanger(eq, tcId, defaultType, cimNamespace, writer);
+    }
+
+    private static void writeShuntCompensators(Network network, String cimNamespace, Map<String, List<RegulatingControlView>> regulatingControlViews, XMLStreamWriter writer) throws XMLStreamException {
+        for (ShuntCompensator s : network.getShuntCompensators()) {
+            String linearNonlinear;
```
MioRtia, line 128: Maybe call it `shuntType`?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
```diff
@@ -0,0 +1,577 @@
...
+                if (c instanceof Injection) {
+                    // An injection should have only one terminal
+                } else if (c instanceof Branch) {
+                    switch (((Branch<?>) c).getSide(t)) {
+                        case ONE:
+                            numt = 1;
+                            break;
+                        case TWO:
+                            numt = 2;
+                            break;
```
MioRtia, line 388: Add default case and throw an exception

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
```diff
@@ -0,0 +1,577 @@
...
+                    switch (((ThreeWindingsTransformer) c).getSide(t)) {
+                        case ONE:
+                            numt = 1;
+                            break;
+                        case TWO:
+                            numt = 2;
+                            break;
+                        case THREE:
+                            numt = 3;
+                            break;
```
MioRtia, line 400: Same remark

MioRtia (commented):
Hi! Some minor technical remarks, functional testing remain to be done on our part. Thank you for all this work!

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -7,42 +7,91 @@
...
 public class CgmesExport implements Exporter {
 
+    public List<Parameter> getParameters() {
+        return Collections.unmodifiableList(STATIC_PARAMETERS);
+    }
+
     @Override
     public void export(Network network, Properties params, DataSource ds) {
+        Objects.requireNonNull(network);
+        if (params != null && ConversionParameters.readBooleanParameter(getFormat(), params, USING_ONLY_NETWORK_PARAMETER)) {
```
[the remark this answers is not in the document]
zamarrenolm: Sorry, I always thought parameters in `Exporters.export` API could be `null`. In fact `TestExporter` in package `com.powsybl.iidm.export` and a few related tests use `parameters == null`.

Also, I followed the `XmlExporter`, where the parameters are used in the `createExportOptions` method, that finally ends in the `ConversionParameters.readParameter` method, where parameters is explicitly checked for `null`, so `null` is also allowed in `iidm-xml-converter`.

Am I right ? Am I missing something ?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -7,42 +7,91 @@
...
+            SteadyStateHypothesisExport.write(network, writer, context);
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        } catch (XMLStreamException e) {
+            throw new UncheckedXmlStreamException(e);
         }
-        CgmesUpdate cgmesUpdate = ext.getCgmesUpdate();
+    }
 
+    private void exportUsingOriginalCgmesModel(Network network, DataSource ds, CgmesModelExtension ext) {
```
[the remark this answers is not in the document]
zamarrenolm: fixed

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -306,6 +309,25 @@ private void addCgmesSvMetadata(Network network) {
...
+                    .setModelingAuthoritySet(sshDescription.get(0).getId("modelingAuthoritySet"));
+            sshDescription.pluckLocals("DependentOn").forEach(adder::addDependency);
+            adder.add();
+        }
+    }
+
+    private void addCgmesSshControlAreas(Network network, Context context) {
+        PropertyBags sshControlAreas = cgmes.controlAreas();
+        if (sshControlAreas != null && !sshControlAreas.isEmpty()) {
+            context.ignored("ControlAreas", "Unsupported in current version");
```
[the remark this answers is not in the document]
zamarrenolm: added TODO comment

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractTransformerConversion.java
zamarrenolm: For phase tap changers, the subtype of tap changer (Linear, non-Linear Asymmetrical, non-Linear Symmetrical) must be stored as a property because the subtype class name is required for proper SSH export. The XML element must be named with the subtype. An example for the MicroGrid test case:

```xml
  <cim:PhaseTapChangerAsymmetrical rdf:about="#_6ebbef67-3061-4236-a6fd-6ccc4595f6c3">
    <cim:TapChanger.step>10</cim:TapChanger.step>
    <cim:TapChanger.controlEnabled>true</cim:TapChanger.controlEnabled>
  </cim:PhaseTapChangerAsymmetrical>
```

In PowSyBl all phase tap changers are tabular, and there is no other way to recover this information from the IIDM `PhaseTapChanger` object, so we decided to save it as a property.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForTransformers.java
zamarrenolm: In the `AbstractTransformerConversion.addCgmesReferences` method:

```java
if (tc.getRegulatingControlId() != null) {
    transformer.setProperty(cgmesReferenceKey(tc.getId(), "TapChangerControl"), tc.getRegulatingControlId());
}
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
zamarrenolm: `AssertionError` thrown

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
zamarrenolm: renamed

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
zamarrenolm: done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
zamarrenolm: done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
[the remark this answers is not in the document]
MioRtia: You are right `parameters` can be `null`, I stand corrected!
However, an issue remains: it means network can't be exported using only network when `params` is `null`, why is that? The test `params != null` should be removed for me

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
```diff
@@ -0,0 +1,580 @@
...
+                } else if (leg.hasRatioTapChanger()) {
+                    String rtcId = twt.getAliasFromType(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.RATIO_TAP_CHANGER + i).orElseThrow(PowsyblException::new);
+                    writeTapChanger(twt, rtcId, leg.getRatioTapChanger(), CgmesNames.RATIO_TAP_CHANGER, regulatingControlViews, cimNamespace, writer);
+                }
+                i++;
+            }
+        }
+    }
+
+    private static void writeTapChanger(Identifiable<?> eq, String tcId, TapChanger<?, ?> tc, String defaultType, Map<String, List<RegulatingControlView>> regulatingControlViews, String cimNamespace, XMLStreamWriter writer) throws XMLStreamException {
```
MioRtia, line 119: I think it is not the case when this method is called but the defaultType should be tabular

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractTransformerConversion.java
MioRtia: Okay! In this case, I wrote it in `SSHExport` file, but in export, the default type should be tabular

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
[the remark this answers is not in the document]
zamarrenolm: Totally right, there is no need to check for `params != null`. The `ConversionParameters.readBooleanParameter` will take care of parameters received as `Properties`, ignoring them if they are `null` or overriding configuration if they are provided.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
zamarrenolm: You are right, I was using `PhaseTapChanger` as default type for phase tap changers, but that is an abstract class in CGMES and can not . The right default value por tap changers, if no type information is provided in properties, should be `PhaseTapChangerTabular`.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractTransformerConversion.java
zamarrenolm: You are right, I fixed the default value (`PhaseTapChanger` is an abstract class in CGMES, a subtype must be used, and default should be `PhaseTapChangerTabular`, according to IIDM modelling).

MioRtia (approved):

MioRtia:
It is okay for me. If it is okay for you too @annetill, we can merge this.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
```diff
@@ -0,0 +1,629 @@
...
+            for (Terminal t : c.getTerminals()) {
+                writeTerminal(t, c, cimNamespace, writer);
+            }
+        }
+        for (Switch sw : network.getSwitches()) {
+            // Terminals for switches are exported as always connected
+            // The status of the switch is "open" if any of the original terminals were not connected
+            // An original "closed" switch with any terminal disconnected
+            // will be exported as "open" with terminals connected
+            sw.getAliasFromType(ALIAS_TYPE_TERMINAL_1)
```
annetill, line 90: I am not sure to get how we should model an open switch. Do we have to modify only the status and keep the terminals connected?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
```diff
@@ -0,0 +1,629 @@
...
+            }
+        }
+    }
+
+    private static boolean isInSlackBus(Generator g) {
+        VoltageLevel vl = g.getTerminal().getVoltageLevel();
+        SlackTerminal slackTerminal = vl.getExtension(SlackTerminal.class);
+        if (slackTerminal != null) {
+            Bus slackBus = slackTerminal.getTerminal().getBusBreakerView().getBus();
+            if (slackBus == g.getTerminal().getBusBreakerView().getBus()) {
```
annetill, line 268: Is it working if the generator is not connected?

annetill (commented):
I have seen no errors, except maybe if a generator was determied as slack bus in the import but not connected. @MioRtia has met this kind of configuration. I think that:
- I have to peform a functional validation for svc regulating status management ;
- Droop for active power control introduction ;
- An issue has to be written to say what we don't support (SVC with target deadband, topological island, control areas, etc.) to continue this work next year.
The whole PR is really clear and I am really impressed by this work. Thanks a lot!

annetill (approved):
I have tested this PR on IOP dataset to be sure that we don't have any strong regression during the import. I have tried to resolve the conflict without anys success. After resolving it, we can merge this PR. It is a great new feature, thanks a lot !

Issue #1482, opened 2020-09-23
The "ExpressionPrinter.toString(ExpressionNode)" method is platform-dependent


olperr1:
* **Do you want to request a *feature* or report a *bug*?**
A bug

* **What is the current behavior?**
The `com.powsybl.dsl.ast.ExpressionPrinter.toString(ExpressionNode)` method is platform-dependent and thus doesn't always produce an UTF-8 string.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
Create a `ContingencyOccurredNode` object with a `contingencyId` containing an accented character.
Use the `ExpressionPrinter.toString(ExpressionNode)` method to convert it in a iAL instruction.

  - Test it with the `-Dfile.encoding=utf8` JVM option: the accented character is correctly rendered.
  - Use the `-Dfile.encoding=ASCII` JVM option: the accented character is replaced by a '?'.

* **What is the expected behavior?**
Modifying the file encoding of the platform should not have an impact on this string generation method.

* **Please tell us about your environment:**
  - PowSyBl Version: 3.5.0

PR #1486, opened 2020-09-25
[iAL] fix encoding

yichen88:

#1482 

**Other information**:


dsl/src/main/java/com/powsybl/dsl/ast/ExpressionPrinter.java
```diff
@@ -6,133 +6,136 @@
...
+import java.io.UncheckedIOException;
 import java.util.Objects;
 
 /**
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
  */
 public class ExpressionPrinter extends DefaultExpressionVisitor<Void, Void> {
 
-    protected final PrintStream out;
+    protected final StringWriter out;
```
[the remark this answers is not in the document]
sylvlecl: (Otherwise, strictly speaking, removing the constructor based on a PrintStream is a breaking change. No big issue, but we can avoid it :) )

sylvlecl (commented):
See proposition of more "flexible" solution in comment.

dsl/src/main/java/com/powsybl/dsl/ast/ExpressionPrinter.java
```diff
@@ -37,102 +33,110 @@ public static void print(ExpressionNode node, PrintStream out) {
...
+    }
+
     public ExpressionPrinter(PrintStream out) {
-        this.out = Objects.requireNonNull(out);
+        this.out = new PrintWriter(Objects.requireNonNull(out));
+    }
+
+    public ExpressionPrinter(Writer writer) {
+        out = new PrintWriter(Objects.requireNonNull(writer));
     }
```
mathbagu, line 46: Looking at the javadoc it's not clear to me why we could not have a constructor that takes an OutputStream instead of a PrintStream. That will make `ExpressionPrinter` and `ActionExpressionPrinter` more convenient to use as we just wrap the objects in a `PrintWriter` that seems to work well with a simple OutputStream.

action/action-dsl/src/main/java/com/powsybl/action/dsl/ast/ActionExpressionPrinter.java
```diff
@@ -21,20 +21,28 @@
...
     }
 
     public static void print(ExpressionNode node, PrintStream out) {
-        node.accept(new ActionExpressionPrinter(out), null);
+        out.print(toString(node));
+    }
+
+    public ActionExpressionPrinter(StringWriter out) {
+        super(out);
     }
```
mathbagu, line 46: Same question here: do we really need a `StringWriter`? It seems a Writer is sufficient here. You should change constructors of both classes to take the highest interface you can from the inheritance tree.

dsl/src/main/java/com/powsybl/dsl/ast/ExpressionPrinter.java
sylvlecl: Agreed. The only thing which is not completely clear now is that this constructor will use "default charset" to write to the output stream: I think we should mention it in javadoc of the constructor.

action/action-dsl/src/main/java/com/powsybl/action/dsl/ast/ActionExpressionPrinter.java
sylvlecl: Agreed

sylvlecl (approved):
Approved: could you just add a short javadoc to explain that the constructor taking an `OutputStream` will used default charset?

mathbagu:
@sylvlecl @yichen88 There is a regression in this PR. To increase the code coverage, I tried to use the ActionExpressionPrinter with a OutputStream but it doesn't write anything. So I tried to print the expression:
```
ActionExpressionPrinter.print(node);
```
This works well with the master branch but not at all with this branch.

PR #1491, opened 2020-09-28
Prevent x of step to be undefined (NaN) during CGMES import for non tabular ptc

MioRtia:

Bug fix

When all alpha are equal to 0 in a PTC, the calculated x for non tabular PTC is undefined. 

In this case, x is equal to 0.

No

**Other information**:

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/CgmesPhaseTapChangerBuilder.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/CgmesPhaseTapChangerBuilder.java
@@ -142,8 +142,12 @@ public class CgmesPhaseTapChangerBuilder extends AbstractCgmesTapChangerBuilder 
             return;
         }
         double alphaMax = tapChanger.getSteps().stream().map(TapChanger.Step::getAngle)
-                .mapToDouble(Double::doubleValue).max().orElse(Double.NaN);
+                .mapToDouble(Double::doubleValue).max().orElse(0);
         tapChanger.getSteps().forEach(step -> {
+            if (alphaMax == 0 || xtx == 0) {
+                step.setX(0);
+                return;
+            }
             double alpha = step.getAngle();
             double x = getStepXforAsymmetrical(xMin, xMax, alpha, alphaMax, windingConnectionAngle);
             step.setX((x - xtx) / xtx * 100);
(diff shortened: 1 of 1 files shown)


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/CgmesPhaseTapChangerBuilder.java
```diff
@@ -142,8 +142,12 @@ private void addStepsAsymmetrical() {
             return;
         }
         double alphaMax = tapChanger.getSteps().stream().map(TapChanger.Step::getAngle)
-                .mapToDouble(Double::doubleValue).max().orElse(Double.NaN);
+                .mapToDouble(Double::doubleValue).max().orElse(0);
         tapChanger.getSteps().forEach(step -> {
+            if (alphaMax == 0 || xtx == 0) {
+                step.setX(0);
+                return;
+            }
```
marqueslanauja, line 150: I think the problem is that the `X `of the transformer is zero. We should not allow transformers with zero reactance. I think is better to fix this problem in the class `CgmesT2xModel `(Line 35) for `twoWindingsTransformer `and in the class `CgmesT3xModel `(Line 43) for `threeWindingsTransformers`. We can write an invalid warning and fix the `X `to a residual value (0.0001 by example) or maybe we can throw an exception.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/CgmesPhaseTapChangerBuilder.java
```diff
@@ -190,8 +194,12 @@ private void stepXforLinearAndSymmetrical() {
             return;
         }
         double alphaMax = tapChanger.getSteps().stream().map(TapChanger.Step::getAngle)
-            .mapToDouble(Double::doubleValue).max().orElse(Double.NaN);
+            .mapToDouble(Double::doubleValue).max().orElse(0);
         tapChanger.getSteps().forEach(step -> {
+            if (alphaMax == 0.0 || xtx == 0.0) {
+                step.setX(0);
+                return;
+            }
```
marqueslanauja, line 202: Same comment

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/CgmesPhaseTapChangerBuilder.java
MioRtia: Is it okay for you to put `Double.MIN_VALUE`? I would rather correct the value than throw an exception (the validation can be modified in another PR)

MioRtia: For clarification, I ran into this exception with `alphaMax` equals to 0.

marqueslanauja: Sorry, then my comment does not make sense. Maybe is better to check only alphaMax
```
if (alphaMax == 0) {
                step.setX(0);
                return;
            }
```

marqueslanauja (approved):

zamarrenolm (approved):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/CgmesPhaseTapChangerBuilder.java
MioRtia: I did what you suggested

annetill (approved):

PR #1494, opened 2020-10-01
Deplace LoadDetailXmlSerializer in iidm-xml-converter

MioRtia:

Moving code for the organisation to be more logical

More or less (it should not break anything in most cases since it should only be used in runtime).


iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/LoadDetailXmlSerializer.java
```diff
@@ -6,61 +6,10 @@
...
-        }
-        load.newExtension(LoadDetailAdder.class)
-                .withFixedActivePower(fixedActivePower)
-                .withFixedReactivePower(fixedReactivePower)
-                .withVariableActivePower(variableActivePower)
-                .withVariableReactivePower(variableReactivePower)
-                .add();
-        return load.getExtension(LoadDetail.class);
-    }
+public class LoadDetailXmlSerializer extends com.powsybl.iidm.xml.extensions.LoadDetailXmlSerializer {
```
mathbagu, line 14: I'm not sure it's useful (or it's working...):
- People that need this serializer needs powsybl-iidm-xml-converter, so they automatically have the serializer
- You will have 2 serializer for the same extension: is it working?

I propose to simply move the class without managing the compatibility at compile time. At runtime, it will work perfectly.

MioRtia: You're right, it is only ever used in runtime so keeping it is useless.

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/extensions/LoadDetailTest.java
```diff
@@ -24,7 +26,7 @@
  */
 public class LoadDetailTest {
```
mathbagu, line 27: Maybe this test could be part of the TCK and have a subclass in each implementation?

mathbagu (approved):

PR #1495, opened 2020-10-02
IIDM validation in 3w leg adder: substation id in the error message

zamarrenolm:

small feature to provide substation id in the error message when a validation error is thrown in the leg adder of the three windings transformer.

During import from XIIDM, if a leg was missing the connectable bus, the message simply said that there was an error adding leg of a three windings transformer, without providing additional context that could help the user locate the equipment in the network.

The substation of the tree windings transformer is added to the error message to facilitate finding which equipment had the error.


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ThreeWindingsTransformerAdderImpl.java
```diff
@@ -161,7 +161,7 @@ public ThreeWindingsTransformerAdderImpl add() {
 
         @Override
         public String getMessageHeader() {
-            return String.format("3 windings transformer leg%d: ", legNumber);
+            return String.format("3 windings transformer leg%d in substation %s: ", legNumber, substation.getId());
```
MioRtia, line 164: Maybe also specify the voltage level in which the leg is?

zamarrenolm: I think that an error could also be thrown if there is no voltage level or it is invalid, so we should take into account that in the message header, complicating things in excess. I thought that adding the substation was enough to quickly find the transformer with the problem.

marqueslanauja (approved):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ThreeWindingsTransformerAdderImpl.java
mathbagu: You cannot have access to the voltage level from here because the leg is not yet attached to a VoltageLevel. From my point of view the message will be strange and confusing. What if we print the ID of the 3WT instead?

zamarrenolm: @mathbagu The id of the transformer may also have not been set yet. The user is allowed to first build and add the legs and then set the `id` and `name` of the transformer. A currently valid way of building a three winding transformer is:

```
tx = substation.newThreeWindingsTransformer()
    .newLeg1()
        .setRatedU(ratedU1)
        .setR(r1).setX(x1).setG(g1).setB(b1)
        .setConnectableBus(bus380.getId())
        .setBus(bus380.getId())
        .setVoltageLevel(vl380.getId())
    .add()
    .newLeg2()
        ...
    .add()
    .newLeg3()
        ...
    .add()
    .setId(txId)
    .setName(txName)
.add();
```

So the only data item that is present in all possible situations is the substation. When I built the message I found it was enough to locate a transformer where the error has been reported.

MioRtia: Okay, thank you for your reply, in this case, it is okay for me.

MioRtia (approved):

mathbagu (approved):

Issue #1489, opened 2020-09-25
Reuse IIDM adder more than once


mathbagu:
* **Do you want to request a *feature* or report a *bug*?**
Bug

* **What is the current behavior?**
For some equipments, reuse an adder will lead to share the same instance between several equipment objects. This is the case for:
- DanglingLines can share the Generation
- Shunts can share the model
- TieLines can share the HalfLines
- ThreeWindingsTransformers can share the legs

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
Before adding an equipment, store the adder in a temporary variable.
Add the equipment to the network.
Modify at least the ID and the connectivity of the adder
Add the new equipment to the network
Assert that both equipments are sharing the same internal object.

* **What is the expected behavior?**
Have two different copies of the internal object.

* **What is the motivation / use case for changing the behavior?**
Be able to reuse an adder. 

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

PR #1496, opened 2020-10-02
[IIDM] Allow to reuse DanglingLineAdder twice

tadam50:

'#1489'

**Other information**:

Only Dangling part of issue is done on this PR


iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractDanglingLineTest.java
```diff
@@ -274,6 +274,10 @@ public void withRegulatingCapabilityTests() {
                 .add();
         assertNotNull(generation.getReactiveLimits());
         assertTrue(generation.getReactiveLimits() instanceof MinMaxReactiveLimits);
+
+        // Test if new Generation is instantiate at each add
+        DanglingLine dl2 = adder.setId(id + "_2").add();
+        assertNotEquals(dl.getGeneration(), dl2.getGeneration());
```
mathbagu, line 280: assertNotSame? Both could be equal (you're lucky because of the default implementation, but I could consider two generations equals if the have the same characteristics, but not attached to the same danglingLine).

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineAdderImpl.java
```diff
@@ -64,10 +64,6 @@ public GenerationAdder setTargetV(double targetV) {
 
         @Override
         public DanglingLineAdder add() {
-            ValidationUtil.checkActivePowerLimits(DanglingLineAdderImpl.this, minP, maxP);
-            ValidationUtil.checkActivePowerSetpoint(DanglingLineAdderImpl.this, targetP);
-            ValidationUtil.checkVoltageControl(DanglingLineAdderImpl.this, voltageRegulationOn, targetV, targetQ);
```
mathbagu, line 69: This is arguable: I think we should keep the tests here: It's not possible to update the GenerationAdder after you have called the add(). So you should be noticed that it's not complete when we consider it should be fully completed.
@MioRtia @tadam50 What do you think?

mathbagu (changes requested):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineAdderImpl.java
tadam50: My mistake. I'm agree with you. I think validation checks must be done as soon as possible.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineAdderImpl.java
```diff
@@ -148,7 +147,8 @@ public DanglingLineAdder setUcteXnodeCode(String ucteXnodeCode) {
 
     @Override
     public GenerationAdder newGeneration() {
-        return new GenerationAdderImpl();
+        generationAdder = new GenerationAdderImpl();
+        return generationAdder;
```
mathbagu, line 151: not sure this is correct: I would write
```
return new GenerationAdderImpl(this);
```
And in the `GenerationAdder::add` method, I would set the generationAdder of the parent. I don't know if it changes something, but it's seems better to me that the DanglingLineAdder is modified only once the GenerationAdder has been filled.

mathbagu: Regarding what you did for three windings transformers, it's coherent

tadam50: In deed, set the generationAdder in `GenerationAdder::add`, after validity checks, seems a better solution.
DanglingLineAdder will be set only with a GenerationAdder well filled

mathbagu (approved):

PR #1497, opened 2020-10-02
LoadFlowResult 1.1

geofjamg:

No

feature

New results have been added to `LoadFlowResult`. Per component:
  - component num
  - status
  - iteration count
  - slack bus id
  - slack bus active power mismatch 

**Other information**:


mathbagu:
It would be nice to update the loadflow tool to print the new results?

geofjamg:
> It would be nice to update the loadflow tool to print the new results?

What would be the best to print results by component, the new component dimension make it not easy to display?
Here is what we have right now:
```bash
loadflow results:
+--------+------------------------------------------------------+
| Result | Metrics                                              |
+--------+------------------------------------------------------+
| true   | {network_0_iterations=8, network_0_status=CONVERGED} |
+--------+------------------------------------------------------+
```

geofjamg:
I did it with 2 tables:
```bash
Loadflow result:
+------+------------------------------------------------------+
| Ok   | Metrics                                              |
+------+------------------------------------------------------+
| true | {network_0_iterations=8, network_0_status=CONVERGED} |
+------+------------------------------------------------------+
Components result:
+------------------+-----------+-----------------+--------------+--------------------+
| Component number | Status    | Iteration count | Slack bus ID | Slack bus mismatch |
+------------------+-----------+-----------------+--------------+--------------------+
| 0                | CONVERGED | 8               | VLEVAP6_0    | -0,00954794        |
+------------------+-----------+-----------------+--------------+--------------------+
```

mathbagu:
I think we could have one line per component, and display above of the component results, the global status of the computation. Below an example of what we could have:

```
Loadflow results: OK
Metrics: {network_0_iterations=8, network_0_status=CONVERGED}

+---------------+---------+-------------------+-------------+---------------------+
| Component | Status | Iteration count | Slack Bus | Mismatch (MW) |
+---------------+---------+-------------------+-------------+---------------------+
|                 0 | TRUE  |                      8 | BUS_1     |                     127 |
+---------------+---------+-------------------+-------------+---------------------+
```

About the computation status, as you suggested, one day we'll replace the boolean by an enum to carry more information.

mathbagu:
I think we was editing this PR at the same time... But our two proposals look the same so it's fine ! Good job

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowResult.java
```diff
@@ -6,18 +6,76 @@
...
+         * Get component number.
+         * @return component number
+         */
+        int getComponentNum();
+
+        /**
+         * Get detailed status of the computation on this component.
+         * @return the detailed status of the computation on this component
+         */
+        String getStatus();
```
annetill, line 35: I wonder about the possibility to have an enum here with at least:
- CONVERGED
- MAX_ITERATION_REACHED
- SOLVER_FAILED: I am not sure about the meaning
- NO_CALCULATION
- Hades2 has a status that says if the result is realist or not. I think we don't need it because limit violations detection should be in the post processing.

geofjamg:
> I think we was editing this PR at the same time... But our two proposals look the same so it's fine ! Good job
> 
> Maybe a small improvement is to add `(MW)` to give more context about the mismatch. What do you think? And I guess it should be `Components results` or `Results per component`

Done

annetill (commented):
Just a question about the status.

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowResult.java
geofjamg: I would keep a string as status set could be different between loadflow implementations

geofjamg: Thinking again, maybe you are right... if we want to rely on status to build more high level workflows we have to define all possible statuses.

geofjamg: Result not realist means we have converged to a state with very low voltages so outside by default [0.8, 1.2]. I think we should keep it.

annetill (approved):
I have made just a test with the open loadflow implementation. Just a small comment: the print is only available for itools, I don't have access to this kind of print from my IDE but it could be useful.
We can merge this PR.

PR #1498, opened 2020-10-02
Rename PSSE option to fit powsybl standards

mathbagu:

N/A

Arguable

The option name doesn't fit powsybl standards.

The property name is `psse.import.ignore-base-voltage`. The old key is still supported but the default value has changed.

**Other information**:


psse/psse-converter/src/main/java/com/powsybl/psse/converter/PsseImporter.java
```diff
@@ -41,10 +41,10 @@
 
     private static final String[] EXTS = {"raw", "RAW"};
 
-    private static final Parameter IGNORE_BASE_VOLTAGE_PARAMETER = new Parameter("ignore-base-voltage",
+    private static final Parameter IGNORE_BASE_VOLTAGE_PARAMETER = new Parameter("psse.import.ignore-base-voltage",
             ParameterType.BOOLEAN,
             "Ignore base voltage specified in the file",
-            Boolean.TRUE);
+            Boolean.FALSE).addAdditionalNames("ignore-base-voltage");
```
geofjamg, line 47: As PSSE is really is beta (or even alpha) mode. I don't think we have to pollute the code with option backward compatibility.

geofjamg (approved):

psse/psse-converter/src/main/java/com/powsybl/psse/converter/PsseImporter.java
mathbagu: I fix the code, I will fix the wiki and the documentation.

PR #1500, opened 2020-10-02
CGMES conversion performance improvement: avoid duplicated queries

zamarrenolm:

A performance improvement in the CGMES conversion

Queries that obtain terminals and nodes from the triplestore are run twice during the conversion process. For terminals, the results were already available in the CGMES model. 

Connectivity and Topological nodes have been cached explicitly in separate collections. Cached (_computed_) terminals are used from the CGMES model instead of re-querying the triplestore. 

**Other information**:
Wall time measurement for CGMES conversion unit tests has improved by 6%. Wall time measurement for a set of real-world test cases (load and convert to IIDM) has improved by 14%.


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ACLineSegmentConversion.java
```diff
@@ -71,19 +69,46 @@ private void convertLine() {
...
-                .setG1(gch / 2)
-                .setG2(gch / 2)
-                .setB1(bch / 2)
-                .setB2(bch / 2);
-        identify(adder);
-        connect(adder);
-        final Line l = adder.add();
-        addAliases(l);
-        convertedTerminals(l.getTerminal1(), l.getTerminal2());
+        if (isZeroImpedanceInsideVoltageLevel(r, x, bch, gch)) {
```
annetill, line 72: Thanks @zamarrenolm, it works well with a conversion into a switch when wa are inside the same voltage level. But I wonder if we should prefer an internal connexion to a swicth. The next step could be to make our internal connexion identifiable. @mathbagu seems to prefer an internal connection to preserve graph analysis.

zamarrenolm: You are right, the fictitious breaker allows to  preserve the id of the original equipment (for an internal connection we can not have identifiers or aliases). 

But there is another reason: an internal connection is always _closed_, it can not be created and declared _opened_. The corresponding ACLineSegment from CGMES could be opened (at one end or both), it is easier to map it to IIDM as an _opened_ fictitious switch.

annetill: The problem we see is that this interconnector has already 2 switches connected at each side. So, with our solution, we will have 3 switches. With an internal connection, we can open it at each side by opening the switch. It works in a node/breaker topology but what about the bus/breaker topology ?

zamarrenolm: In CGMES, even if the switches are closed, the `ACLineSegment` could be seen as opened because its terminals could have the `connected` attribute set to `false`. If we use an internal connection we lose that possibility. 

Having three switches sequentially should not be a problem for topology calculations at bus-breaker level, and at bus-view all three will be hidden if they are not declared as _retained_.

annetill (approved):
Maybe a test is needed to increase test coverage. I keep in mind that we need to open an issue to allow (or not) explicitely a zero impedance line inside a voltage level. If we accept it, we have to fix the topology calculation.

annetill (approved):

mathbagu (approved):

PR #1502, opened 2020-10-05
[IIDM] Allow to reuse ShuntCompensatorAdder twice

tadam50:

'#1489'

**Other information**:

Only ShuntCompensator part of issue is done on this PR


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractShuntCompensatorModel.java
```diff
@@ -19,9 +17,6 @@
 
     @Override
     public void setShuntCompensator(ShuntCompensatorImpl shuntCompensator) {
```
mathbagu, line 19: @MioRtia why the public API of ShuntCompensatorModel is empty? I would move everything except this method in the public API (i.e. outside the wrapper)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
```diff
@@ -177,6 +180,17 @@ public ShuntCompensatorAdder setTargetDeadband(double targetDeadband) {
     public ShuntCompensatorImpl add() {
         String id = checkAndGetUniqueId();
         TerminalExt terminal = checkAndGetTerminal();
+        ShuntCompensatorModelWrapper model = null;
```
mathbagu, line 183: I think this class is redundant with AbstractShuntCompensatorModel
@MioRtia Is it ok if we remove it?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
```diff
@@ -177,6 +180,17 @@ public ShuntCompensatorAdder setTargetDeadband(double targetDeadband) {
...
+        if (shuntCompensatorLinearModelAdder != null) {
+            model = new ShuntCompensatorLinearModelImpl(shuntCompensatorLinearModelAdder.bPerSection,
+                                                        shuntCompensatorLinearModelAdder.gPerSection,
+                                                        shuntCompensatorLinearModelAdder.maximumSectionCount);
+        } else if (shuntCompensatorNonLinearModelAdder != null) {
+            model = new ShuntCompensatorNonLinearModelImpl(shuntCompensatorNonLinearModelAdder.sections.stream()
+                                                                                                       .map(ShuntCompensatorNonLinearModelImpl.SectionImpl::copy)
+                                                                                                       .collect(Collectors.toList()));
```
mathbagu, line 191: I would perform the deep copy in the constructor of ShuntCompensatorNonLinearModelImpl (and document it)

mathbagu (changes requested):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractShuntCompensatorModel.java
MioRtia: It shouldn't be in the public API, should it? A shunt compensator model should not be created independently from the shunt for me, this method was just because it was created in the adder (so before the shunt itself is created). The method can be deleted if it is not useful anymore.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
MioRtia: Indeed, you can only use `AbstractShuntCompensatorModel` and make the following methods abstract:
```java
ShuntCompensatorModelType getType();

    int getMaximumSectionCount();

    double getB(int sectionCount);

    double getG(int sectionCount);
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
```diff
@@ -18,7 +18,9 @@
 
     private final VoltageLevelExt voltageLevel;
 
-    private ShuntCompensatorModelWrapper model;
+    private ShuntCompensatorNonLinearModelAdderImpl shuntCompensatorNonLinearModelAdder;
+
+    private ShuntCompensatorLinearModelAdderImpl shuntCompensatorLinearModelAdder;
```
mathbagu, line 23: This is not sufficient to ensure someone don't set two different models. You should introduce a common interface or reset the corresponding attribute in the add methods.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
```diff
@@ -177,6 +179,15 @@ public ShuntCompensatorAdder setTargetDeadband(double targetDeadband) {
...
         String id = checkAndGetUniqueId();
         TerminalExt terminal = checkAndGetTerminal();
+        AbstractShuntCompensatorModel model = null;
+        if (shuntCompensatorLinearModelAdder != null) {
+            model = new ShuntCompensatorLinearModelImpl(shuntCompensatorLinearModelAdder.bPerSection,
+                                                        shuntCompensatorLinearModelAdder.gPerSection,
+                                                        shuntCompensatorLinearModelAdder.maximumSectionCount);
+        } else if (shuntCompensatorNonLinearModelAdder != null) {
+            model = new ShuntCompensatorNonLinearModelImpl(shuntCompensatorNonLinearModelAdder.sections);
+        }
```
mathbagu, line 189: See my previous comment.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorNonLinearModelImpl.java
```diff
@@ -60,6 +61,10 @@ public Section setG(double g) {
             return this;
         }
 
+        public SectionImpl copy() {
+            return new SectionImpl(index, b, g);
+        }
```
mathbagu, line 66: I think `clone` is more commonly used

mathbagu (changes requested):

mathbagu (approved):

PR #1507, opened 2020-10-06
[IIDM] Allow to reuse TieLineAdder twice

tadam50:

'#1489'

**Other information**:

Only TieLineAdder part of issue is done on this PR


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/TieLineAdder.java
```diff
@@ -11,27 +11,54 @@
...
+    TieLineAdder.HalfLineAdder newHalfLine1();
+
+    TieLineAdder.HalfLineAdder newHalfLine2();
+
+    /**
+     * @deprecated Use {@link #newHalfLine1()} and {@link TieLineAdder#add()} instead.
+     */
+    @Deprecated
+    default TieLineAdder line1() {
+        throw new UnsupportedOperationException();
```
mathbagu, line 52: This will break all existing code: I don't if it's possible but we should create a proxy that wrap a HalfLineAdder in a TieLineAdder?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/TieLineAdderImpl.java
```diff
@@ -6,169 +6,182 @@
...
 /**
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
  */
 class TieLineAdderImpl extends AbstractBranchAdder<TieLineAdderImpl> implements TieLineAdder {
 
-    private final NetworkImpl network;
+    class HalfLineAdderImpl implements Validable, TieLineAdder.HalfLineAdder {
 
-    private String ucteXnodeCode;
+        protected int num;
```
mathbagu, line 18: Should be final and set in the constructor?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/TieLineAdderImpl.java
```diff
@@ -6,169 +6,182 @@
...
         return this;
     }
 
     @Override
-    public TieLineAdderImpl setXnodeQ(double xnodeQ) {
-        getActiveHalf().setXnodeQ(xnodeQ);
-        return this;
+    public HalfLineAdderImpl newHalfLine1() {
+        halfLineAdder1 = new HalfLineAdderImpl();
+        halfLineAdder1.num = 1;
```
mathbagu, line 176:
```
halfLineAdder1 = new HalfLineAdderImpl(1);
return halfLineAdder1;
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/TieLineAdderImpl.java
```diff
@@ -182,8 +195,40 @@ public TieLineImpl add() {
         if (ucteXnodeCode == null) {
             throw new ValidationException(this, "ucteXnodeCode is not set");
         }
-        checkHalf(half1, 1);
-        checkHalf(half2, 2);
```
mathbagu, line 186: I would check if both half1 and half2 adders are set before creating half1 and half2

mathbagu (changes requested):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/TieLineAdderProxy.java
```diff
@@ -0,0 +1,227 @@
...
+import com.powsybl.iidm.network.*;
+import java.util.Objects;
+
+/**
+ * @author Thomas Adam <tadam at silicom.fr>
+ *
+ * @deprecated Use {@link TieLineAdder#newHalfLine1()} or {@link TieLineAdder#newHalfLine2()} instead.
+ */
+@Deprecated
+class TieLineAdderProxy implements TieLineAdder, Validable {
```
mathbagu, line 18: The proxy should be moved in the API and used in the old API methods. So that it will be really easy to remove existing code in the implementation without breaking existing code.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/TieLineAdderProxy.java
```diff
@@ -0,0 +1,227 @@
...
+    @Override
+    public TieLineAdder line2() {
+        this.halfLine2 = delegate.newHalfLine2();
+        this.activeHalf = halfLine2;
+        return this;
+    }
+
+    @Override
+    public TieLineAdderImpl.HalfLineAdderImpl newHalfLine1() {
+        throw new UnsupportedOperationException("Use line1() instead");
```
mathbagu, line 153: return `this.halfLineAdder1` instead

mathbagu (changes requested):
I think the proxy class should be used in the API and the old code remove from the TieLineImpl

PR #1508, opened 2020-10-06
Add possibility to ensure alias unicity

MioRtia:

No

Feature: Add the possibility to ensure alias unicity

When an alias has already been added, an exception is thrown

When an alias has already been added, if `ensureUnicity` is `true`, the alias is modified to be unique.

Yes

**Other information**:
Another PR will follow this one to ensure alias unicity in aliases creation in CGMES import


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Identifiable.java
```diff
@@ -69,6 +69,10 @@ default String getName() {
     default void addAlias(String alias) {
     }
 
+    default void addAlias(String alias, boolean ensureAliasUnicity) {
+        addAlias(alias);
+    }
+
```
mathbagu, line 75: I wonder if it should be a property of the NetworkIndex (Network.setEnsureAliasUnicity) that can be set globally.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiable.java
```diff
@@ -80,20 +81,34 @@ public String getNameOrId() {
 
     @Override
     public void addAlias(String alias) {
-        addAlias(alias, null);
+        addAlias(alias, false);
```
mathbagu, line 84: This could be considered as a breaking change because the default behaviour is not the same anymore

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiable.java
```diff
@@ -80,20 +81,34 @@ public String getNameOrId() {
...
 
     @Override
     public void addAlias(String alias, String aliasType) {
+        addAlias(alias, aliasType, false);
+    }
+
+    @Override
+    public void addAlias(String alias, String aliasType, boolean ensureAliasUnicity) {
         Objects.requireNonNull(alias);
+        String checkedAlias = alias;
```
mathbagu, line 100: `uniqueAlias` seems better than checkedAlias?

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/tck/AliasesTest.java
```diff
@@ -0,0 +1,15 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.iidm.network.impl.tck;
+
+import com.powsybl.iidm.network.tck.AbstractAliasesTest;
+
+/**
+ * @author Sebastien Murgey <sebastien.murgey at rte-france.com>
```
mathbagu, line 12: In my memories, you looked different :)

mathbagu (commented):
Do you have a test that checked a conflict between an alias and an ID, with and without ensuring unicity?

MioRtia:
> Do you have a test that checked a conflict between an alias and an ID, with and without ensuring unicity?

Good question, I don't think so but I will make one!

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiable.java
MioRtia: I don't think it changes the default behaviour, before my change, this method behaved as if unicity was not ensured: an exception was thrown when an already existing alias was passed

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/tck/AliasesTest.java
MioRtia: I only moved this file, I didn't create it, this is why I didn't change the author!

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Identifiable.java
MioRtia: I followed the same logic than `setEnsureIdUnicity` but that can be a good idea for both!

MioRtia:
> > Do you have a test that checked a conflict between an alias and an ID, with and without ensuring unicity?
> 
> Good question, I don't think so but I will make one!

I actually made one in `AbstractAliasesTest.ensureAliasUnicityFromId` (`AbstractAliasesTest.failWhenAliasEqualToAnId` already existed)

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Identifiable.java
MioRtia: Just a question, if I put:
``java
network.setEnsureUnicity(true);
network.getLoad("load").addAlias("alias", false);
```
What do you think the behavior should be? I'm not sure...

MioRtia: I just came back to your comment, you seem to say it should not be set alias by alias but globally. I agree with you but do you think it should be the same for unique IDs?

mathbagu (approved):
Could you check with @annetill if we should not introduce an option to ensure (or not) aliases unicity (false as default to have the same behaviour than in IIDM)

Don't forget the documentation of the new option

PR #1511, opened 2020-10-07
Finish ExecutionReport implementation to get std out/err stream

mathbagu:

N/A

Feature

The `getStdOut` and `getStdErr` are not implemented in the only `ExecutionReport` subclass we have in powsybl-core. The methods are implemented in powsybl-afs, for slurm computation.

The `DefaultExecutionReport` override the two methods. As this class need to access to the working directory, the `ok` method is removed. It was used only in unit test, so it should be acceptable.

**Other information**:


computation/src/main/java/com/powsybl/computation/DefaultExecutionReport.java
```diff
@@ -22,29 +27,22 @@
...
-        return OK;
-    }
-
     /**
      * Create an execution report with no execution error.
      */
-    public DefaultExecutionReport() {
-        this(Collections.emptyList());
+    public DefaultExecutionReport(Path workingDirectory) {
+        this(workingDirectory, Collections.emptyList());
```
MioRtia, line 38: Is it relevant to make a constructor without parameters with a default directory (maybe not)(don't forget the Breaking change label if it is not the case)?

computation/src/main/java/com/powsybl/computation/DefaultExecutionReport.java
```diff
@@ -65,4 +63,29 @@ public void log() {
...
+        Objects.requireNonNull(command);
+        if (index < 0) {
+            throw new IllegalArgumentException("Invalid index: " + index);
+        }
+
+        Path path = workingDirectory.resolve(command.getId() + "_" + index + extension);
+        try {
+            return Optional.of(Files.newInputStream(path));
+        } catch (IOException e) {
+            LOGGER.warn("Unable to read {}: {}", path.toString(), e.getMessage());
```
MioRtia, line 86: Sonarclound said it is not useful to add the `toString()` method as the logger already formats paths

computation/src/main/java/com/powsybl/computation/DefaultExecutionReport.java
```diff
@@ -22,29 +27,22 @@
...
-    private static final ExecutionReport OK = new DefaultExecutionReport();
+    private final Path workingDirectory;
 
     private final List<ExecutionError> errors;
 
-    /**
-     * An execution report with no execution error.
-     * @return an execution report with no execution error.
-     */
-    public static ExecutionReport ok() {
```
MioRtia, line 33: Just for me, when did we use this method? (Don't forget the Breaking change label if we delete it)

computation/src/main/java/com/powsybl/computation/DefaultExecutionReport.java
mathbagu: Fixed in the last commit

computation/src/main/java/com/powsybl/computation/DefaultExecutionReport.java
mathbagu: This was used only in the unit test, but it's really useful to return always the same instance.

computation/src/main/java/com/powsybl/computation/DefaultExecutionReport.java
mathbagu: Not sure it's relevant to have a default constructor. The working directory is the temporary directory used during the execution. From my point of view, computations always run in a temporary directory, even if no data are stored in it, it's created by the computation manager (part of the API).

MioRtia: Okay!

MioRtia (approved):

PR #1516, opened 2020-10-13
Add a requireNonNull to ensure the extension XML serializer is not null when writeExtension() is called

MioRtia:

No

Bug fix: ensure that extension XML serializer is not null when `writeExtension` is called
Bug fix: filter extensions which are not in options during XML export

No

**Other information**:
If you agree with these fixes, they should be also merged for release 3.7.0

### DIFF
--- a/iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/NetworkXml.java
+++ b/iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/NetworkXml.java
@@ -150,7 +150,7 @@ public final class NetworkXml {
 
     private static void writeExtension(Extension<? extends Identifiable<?>> extension, NetworkXmlWriterContext context) throws XMLStreamException {
         XMLStreamWriter writer = context.getExtensionsWriter();
-        ExtensionXmlSerializer extensionXmlSerializer = getExtensionXmlSerializer(context.getOptions(), extension);
+        ExtensionXmlSerializer extensionXmlSerializer = Objects.requireNonNull(getExtensionXmlSerializer(context.getOptions(), extension));
 
         String namespaceUri = getNamespaceUri(extensionXmlSerializer, context.getOptions(), context.getVersion());
 
@@ -216,6 +216,7 @@ public final class NetworkXml {
 
             Collection<? extends Extension<? extends Identifiable<?>>> extensions = identifiable.getExtensions().stream()
                     .filter(e -> getExtensionXmlSerializer(options, e) != null)
+                    .filter(e -> context.getOptions().withExtension(e.getName()))
                     .collect(Collectors.toList());
             if (!extensions.isEmpty()) {
                 context.getExtensionsWriter().writeStartElement(context.getVersion().getNamespaceURI(), EXTENSION_ELEMENT_NAME);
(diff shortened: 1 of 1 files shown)


iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/NetworkXml.java
```diff
@@ -216,6 +216,7 @@ private static void writeExtensions(Network n, NetworkXmlWriterContext context,
 
             Collection<? extends Extension<? extends Identifiable<?>>> extensions = identifiable.getExtensions().stream()
                     .filter(e -> getExtensionXmlSerializer(options, e) != null)
+                    .filter(e -> context.getOptions().withExtension(e.getName()))
```
mathbagu, line 219: This test is already performed in getExtensionXmlSerializer
```
    private static ExtensionXmlSerializer getExtensionXmlSerializer(ExportOptions options, Extension<? extends Identifiable<?>> extension) {
        if (options.withExtension(extension.getName())) {
        ....
        }
     }
```

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/NetworkXml.java
```diff
@@ -150,7 +150,7 @@ private static void writeExtensionNamespaces(Network n, ExportOptions options, X
 
     private static void writeExtension(Extension<? extends Identifiable<?>> extension, NetworkXmlWriterContext context) throws XMLStreamException {
         XMLStreamWriter writer = context.getExtensionsWriter();
-        ExtensionXmlSerializer extensionXmlSerializer = getExtensionXmlSerializer(context.getOptions(), extension);
+        ExtensionXmlSerializer extensionXmlSerializer = Objects.requireNonNull(getExtensionXmlSerializer(context.getOptions(), extension));
```
mathbagu, line 153: `Objects.requireNonNull` throws a NullPointerException: do we need a more specific exception like it's thrown in getExtensionXmlSerializer?

I think we never have a null extensionXmlSerializer there. The `getExtensionXmlSerializer` return null in two cases:
- the extension should not be serialized (no data for the current variant). It never happens because these extensions are filtered in `writeExtensions`
- the extension is skipped (options). It never happens because these extensions are filtered in `writeExtensions`

From my point of view, we should throw an AssertionError here with an explicit message if the serializer is null because something really unexpected happened

mathbagu (changes requested):

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/NetworkXml.java
MioRtia: You're right!

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/NetworkXml.java
MioRtia: As said in the comment of the PR, extensionXmlSerializer is **never** null because when `writeExtension` is called, it has already been checked that `extension` has an XML serializer. I added a `requireNonNull` to reinforce that when we go here, we are not supposed to have a null XML serializer. What bothers me is that `getExtensionXmlSerializer` twice though.

mathbagu (approved):

PR #1519, opened 2020-10-15
[WIP] Fix bug topology in merging view

miovd:

No

It fixes a topology bug: calculated buses in bus breaker view between two retained switches didn't have a terminal reference even if they belonged to an electrical bus.

No

**Other information**:
The test I added is already present in the TCK: it should be deleted when the TCK tests the merging view

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConfiguredBusImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConfiguredBusImpl.java
@@ -70,6 +70,11 @@ class ConfiguredBusImpl extends AbstractBus implements ConfiguredBus {
     }
 
     @Override
+    public TerminalExt getTerminalReference() {
+        return getConnectedTerminalStream().findFirst().orElseGet(() -> getTerminals().isEmpty() ? null : getTerminals().get(0));
+    }
+
+    @Override
     public int getTerminalCount() {
         return terminals.get(network.get().getVariantIndex()).size();
     }
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConfiguredBusImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConfiguredBusImpl.java
@@ -71,12 +71,17 @@ class ConfiguredBusImpl extends AbstractBus implements ConfiguredBus {
 
     @Override
     public TerminalExt getTerminalReference() {
+        return getTerminalReference(new ArrayList<>());
+    }
+
+    private TerminalExt getTerminalReference(List<Bus> visited) {
         return getConnectedTerminalStream().findFirst().orElseGet(() -> {
             if (voltageLevel.getTopologyKind() == TopologyKind.BUS_BREAKER) {
                 // in bus-breaker voltage level, the terminal reference can be a connected terminal in another bus linked to this one by one or several closed switches
                 VoltageLevel.BusBreakerView view = voltageLevel.getBusBreakerView();
+                visited.add(this);
                 return view.getBusStream()
-                        .map(b -> getTerminalReference(b, this, view))
+                        .map(b -> getTerminalReference(b, this, view, visited))
                         .filter(Objects::nonNull)
                         .findFirst()
                         .orElse(null);
@@ -85,18 +90,14 @@ class ConfiguredBusImpl extends AbstractBus implements ConfiguredBus {
         });
     }
 
-    private static TerminalExt getTerminalReference(Bus b1, Bus b2, VoltageLevel.BusBreakerView view) {
-        for (Switch ss : view.getSwitches()) {
-            if (areConnectedBySwitch(b1, b2, ss, view)) {
-                TerminalExt t = (TerminalExt) b1.getTerminalReference();
-                if (t != null) {
-                    return t;
-                }
-            }
-            if (areConnectedBySwitch(b2, b1, ss, view)) {
-                TerminalExt t = (TerminalExt) b1.getTerminalReference();
-                if (t != null) {
-                    return t;
+    private static TerminalExt getTerminalReference(Bus b1, Bus b2, VoltageLevel.BusBreakerView view, List<Bus> visited) {
+        if (!visited.contains(b1)) {
+            for (Switch ss : view.getSwitches()) {
+                if (areConnectedBySwitch(b1, b2, ss, view) || areConnectedBySwitch(b2, b1, ss, view)) {
+                    TerminalExt t = ((ConfiguredBusImpl) b1).getTerminalReference(visited);
+                    if (t != null) {
+                        return t;
+                    }
                 }
             }
         }
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConfiguredBusImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConfiguredBusImpl.java
@@ -6,14 +6,14 @@
  */
 package com.powsybl.iidm.network.impl;
 
-import com.powsybl.iidm.network.Component;
-import com.powsybl.iidm.network.Terminal;
-import com.powsybl.iidm.network.ValidationException;
+import com.powsybl.iidm.network.*;
 import com.powsybl.iidm.network.impl.util.Ref;
 import gnu.trove.list.array.TDoubleArrayList;
 import gnu.trove.list.array.TIntArrayList;
+
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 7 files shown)


iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/BusAdapter.java
```diff
@@ -19,28 +23,48 @@
...
 
-        delegate.visitConnectedEquipments(new AbstractTerminalTopologyVisitor() {
-            @Override
-            public void visitTerminal(Terminal t) {
-                terminals.add(getIndex().getTerminal(t));
-            }
-        });
+        delegate.getConnectedTerminalStream()
+                .map(t -> getIndex().getTerminal(t))
+                .forEach(terminals::add);
```
mathbagu, line 33: As the `terminals` list doesn't change, I would change its initialization:
```
terminals = delegate.getConnectedTerminalStream()
                .map(t -> getIndex().getTerminal(t))
                .collect(Collectors.toList());
```

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/BusAdapter.java
```diff
@@ -19,28 +23,48 @@
...
+                                                             BiFunction<Switch, VoltageLevel, Bus> busGetter, BiFunction<Switch, VoltageLevel, Integer> nodeGetter) {
+        VoltageLevel vl = delegate.getVoltageLevel();
+        return vl.getNodeBreakerView().getSwitchStream()
+                .filter(Switch::isRetained)
+                .filter(s -> busGetter.apply(s, vl).equals(delegate))
+                .map(s -> nodeGetter.apply(s, vl))
+                .map(i -> Networks.getEquivalentTerminal(vl, i))
+                .filter(Objects::nonNull)
+                .map(index::getTerminal)
+                .findFirst();
```
mathbagu, line 55: Why don't we use the same implementation than in the `CalculatedBusImpl`?

Honestly, I don't understand whitch criterias you used to chose the reference terminal. I think you should comment a little bit the algorithm.

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/BusAdapter.java
```diff
@@ -214,7 +238,13 @@ public Component getConnectedComponent() {
...
         ccm.update();
 
-        return terminals.isEmpty() ? null : ccm.getComponent(terminals.get(0).getConnectedComponentNumber());
+        if (terminals.isEmpty()) {
+            if (terminalRef == null) {
+                return null;
+            }
+            return ccm.getComponent(terminalRef.getConnectedComponentNumber());
+        }
+        return ccm.getComponent(terminals.get(0).getConnectedComponentNumber());
```
mathbagu, line 247: Why don't we have the same implementation than in `CalculatedBusImpl`?

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/BusAdapter.java
```diff
@@ -242,7 +272,13 @@ public Component getSynchronousComponent() {
...
         ccm.update();
 
-        return terminals.isEmpty() ? null : ccm.getComponent(terminals.get(0).getSynchronousComponentNumber());
+        if (terminals.isEmpty()) {
+            if (terminalRef == null) {
+                return null;
+            }
+            return ccm.getComponent(terminalRef.getSynchronousComponentNumber());
+        }
+        return ccm.getComponent(terminals.get(0).getSynchronousComponentNumber());
```
mathbagu, line 281: Same question here

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/BusAdapter.java
MioRtia: It was because in merging view, we don't have the nodes included in the calculated bus (which are used in network-impl). However, I agree with you that implementing the algorithm a second time here to find the nodes is error-prone. I put the `getTerminalReference()` method in the code which simplify the code A LOT. The default implementation is not exactly correct but it is close enough and this way, we use the algorithm directly from `CalculatedBusImpl`.

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/BusAdapter.java
MioRtia: You're right, I think I made it complicated while trying out solutions... Changing it now.

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/BusAdapter.java
MioRtia: Same remark

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConfiguredBusImpl.java
```diff
@@ -69,6 +69,11 @@ public int getConnectedTerminalCount() {
         return getTerminals().stream().filter(Terminal::isConnected).map(Function.identity());
     }
 
+    @Override
+    public TerminalExt getTerminalReference() {
+        return getConnectedTerminalStream().findFirst().orElseGet(() -> getTerminals().isEmpty() ? null : getTerminals().get(0));
```
floriand-e2r, line 74: I'm wondering, if there's no connected terminal, shouldn't we return null instead of taking a disconnected one?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/MergedBus.java
```diff
@@ -81,6 +81,17 @@ public int getConnectedTerminalCount() {
...
 
+    @Override
+    public TerminalExt getTerminalReference() {
+        checkValidity();
+        return getConnectedTerminalStream()
+                .findFirst()
+                .orElseGet(() -> buses.stream()
+                        .map(BusExt::getTerminalReference)
+                        .findFirst()
+                        .orElse(null));
```
floriand-e2r, line 92: same question here... (if there's no connected terminal, then the terminal reference of the buses are disconnected terminals)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConfiguredBusImpl.java
MioRtia: You're right. I change the algorithm to do what I wanted to do (find a terminal reference for a configured bus in bus-breaker linked to a bus with connected terminals by closed switches)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/MergedBus.java
MioRtia: Yes, in this case you are right, `MergedBus` are buses in the `BusView` so either the terminal reference is a connected terminal either there is none since there is no switch in this view

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConfiguredBusImpl.java
```diff
@@ -71,7 +71,40 @@ public int getConnectedTerminalCount() {
...
+                        .map(b -> getTerminalReference(b, this, view))
+                        .filter(Objects::nonNull)
+                        .findFirst()
+                        .orElse(null);
+            }
+            return null;
+        });
+    }
+
+    private static TerminalExt getTerminalReference(Bus b1, Bus b2, VoltageLevel.BusBreakerView view) {
```
floriand-e2r, line 88: you wrote in comment above that there could be several closed switches between the two buses, but here I'm not so sure... it looks like you're only checking if they're directly connected?

floriand-e2r (changes requested):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConfiguredBusImpl.java
floriand-e2r: I'm closing, I was mistaken! it's perfect.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConfiguredBusImpl.java
```diff
@@ -71,7 +71,40 @@ public int getConnectedTerminalCount() {
...
+        for (Switch ss : view.getSwitches()) {
+            if (areConnectedBySwitch(b1, b2, ss, view)) {
```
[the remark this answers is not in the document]
MioRtia: I'm not sure, if we put the two we have the risk to loop between two switches depending on the order of the switches. For example, if we have :
b1[] -switchA- b2[] -switchB- b3[load]
and the list is {switchA, switchB}, if we go from b1, we have b1 -> switchA -> b2 -> switchA -> b1 -> switchA -> b2 etc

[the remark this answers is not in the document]
floriand-e2r: In fact, I think that merging in `if (areConnectedBySwitch(b1, b2, ss, view) || areConnectedBySwitch(b2, b1, ss, view))` is exactly the same as putting it in 2 if like it is now. But the risk you're pointing out can already occur! :fearful:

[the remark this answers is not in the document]
MioRtia: oops you're right! correcting that and using your use case to validate it

floriand-e2r (approved):
Just not sure I added the unit test at the right place!

PR #1520, opened 2020-10-15
Correct the message exception when adding an equipment with an existing ID in a merging view: the class of the existing object is returned, not the adder class

MioRtia:

No

Fix an exception message

When trying to create an equipment with an existing ID in a merging view, the message reads: `An [ADDER_CLASS] object already exists`

Now the message reads: `An [OBJECT_CLASS] object already exists`

No


iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/AbstractIdentifiableAdderAdapter.java
```diff
@@ -67,7 +69,7 @@ protected void checkAndSetUniqueId() {
             // Check Id is unique in all merging view
             if (getIndex().contains(id)) {
                 throw new PowsyblException("The network already contains an object '"
-                                           + getDelegate().getClass().getSimpleName()
+                                           + getIndex().get(n -> n.getIdentifiable(id), UnaryOperator.identity()).getClass().getSimpleName()
```
mathbagu, line 72: I think it's not working well if the identifiable has been wrap into a MergedLine

mathbagu: Use `getIndex().getView().getIdentifiable()` instead

mathbagu (approved):

PR #1535, opened 2020-11-03
Allows to log error rather than throwing exceptions when there is an IIDM-XML version incompatibility during export

MioRtia:

No

Feature

When a new equipment (non linear shunt or dangling line's generation) is present in a network, backward XIIDM export is not possible: an exception is thrown.

When a new equipment (non linear shunt or dangling line's generation) is present in a network, an export option can be set up in order not to throw an exception but only to log an error. This equipment is exported in best effort (dangling line's generation is reported on p0 and q0 and non linear shunt are exported as shunt with only one current section)

No

**Other information**:
Linked documentation PR: powsybl/powsybl.github.io#167


iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
```diff
@@ -51,8 +51,11 @@ protected void writeRootElementAttributes(ShuntCompensator sc, VoltageLevel vl,
...
         }
         IidmXmlUtil.runUntilMaximumVersion(IidmXmlVersion.V_1_2, context, () -> {
-            XmlUtil.writeDouble(B_PER_SECTION, sc.getModel(ShuntCompensatorLinearModel.class).getBPerSection(), context.getWriter());
-            context.getWriter().writeAttribute(MAXIMUM_SECTION_COUNT, Integer.toString(sc.getMaximumSectionCount()));
+            ShuntCompensatorModel model = sc.getModel();
+            double bPerSection = model instanceof ShuntCompensatorLinearModel ? ((ShuntCompensatorLinearModel) model).getBPerSection() : sc.getB();
+            XmlUtil.writeDouble(B_PER_SECTION, bPerSection, context.getWriter());
+            int maximumSectionCount = model instanceof ShuntCompensatorLinearModel ? sc.getMaximumSectionCount() : 1;
+            context.getWriter().writeAttribute(MAXIMUM_SECTION_COUNT, Integer.toString(maximumSectionCount));
             context.getWriter().writeAttribute("currentSectionCount", Integer.toString(sc.getSectionCount()));
```
mathbagu, line 59: I'm not totally sure this is OK. The new shunt API wasn't available before XIIDM v1.3. If the model is non-linear, should we throw an exception instead? From my point of view, it's not equivalent and should not be supported

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
```diff
@@ -81,13 +84,15 @@ private static void writeModel(ShuntCompensator sc, NetworkXmlWriterContext cont
...
-            context.getWriter().writeEndElement();
+            IidmXmlUtil.runFromMinimumVersion(IidmXmlVersion.V_1_3, context, () -> {
+                context.getWriter().writeStartElement(context.getVersion().getNamespaceURI(), SHUNT_NON_LINEAR_MODEL);
+                for (ShuntCompensatorNonLinearModel.Section s : sc.getModel(ShuntCompensatorNonLinearModel.class).getAllSections()) {
+                    context.getWriter().writeEmptyElement(context.getVersion().getNamespaceURI(), "section");
+                    XmlUtil.writeDouble("b", s.getB(), context.getWriter());
+                    XmlUtil.writeDouble("g", s.getG(), context.getWriter());
+                }
+                context.getWriter().writeEndElement();
+            });
```
mathbagu, line 95: If we encounter a non linear model, and should export it in a previous version, we silently ignore it. Should we throw an exception? The exported network is not equivalent then.

mathbagu:
I add the `doc needed` label, because this PR introduce a new property that has to be documented. I think we should also update the export section to explain the two conversions we are implementing here

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
MioRtia: No, we don't, a log is made (l.51) but it is considered a shunt with a unique section which has the current susceptance

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/util/IidmXmlUtil.java
```diff
@@ -57,92 +66,196 @@ default void run() {
...
+    private static void createExceptionOrLogError(String rootElementName, String elementName, ErrorMessage type, IidmXmlVersion refVersion, String reason, NetworkXmlWriterContext context) {
+        createExceptionOrLogError(rootElementName + "." + elementName, type, refVersion, reason, context);
+    }
+
+    private static void createExceptionOrLogError(String elementName, ErrorMessage type, IidmXmlVersion refVersion, String reason, NetworkXmlWriterContext context) {
+        if (context.getOptions().getIidmVersionIncompatibilityBehavior() == ExportOptions.IidmVersionIncompatibilityBehavior.THROW_EXCEPTION) {
+            throw createException(elementName, type, refVersion, context.getVersion(), reason);
+        } else if (context.getOptions().getIidmVersionIncompatibilityBehavior() == ExportOptions.IidmVersionIncompatibilityBehavior.LOG_ERROR) {
+            Supplier<String> message = () -> message(elementName, type, refVersion, context.getVersion(), reason);
+            LOGGER.error(message.get());
```
mathbagu, line 90: There is no good reason to create a supplier if you always call the `get()` method. I would replace it by a simple string and check if ERROR level is enabled

mathbagu (changes requested):

mathbagu (approved):

PR #1538, opened 2020-11-03
Add convenient methods on branch/injection for PQI and connection.

### LINKED ISSUE #1536: [IIDM] P, Q, I availability on Injection and Branch classes

Feature.

Some frequently used physical quantities like power flows and currents are only available on `Terminal` objects, whereas one could expect to have them available on the network equipments themselves, more specifically on `Branch` and on `Injection` classes.

For example to get the current on side 1 of a line, we need to write :
```java
double p1 = line.getTerminal1().getP();
```
or for a generator : 
```java
double p = generator.getTerminal().getP();
//surprising because on the other side we can retrieve targetP directly:
double targetP = generator.getTargetP();
```

The user should be able to retrieve those quantities more simply:
```java
double p1 = branch.getP1();
double p2 = branch.getP2();
double p = generator.getP();
```
Such shortcuts should be available on `Injection` and `Branch` for:
 * P (getter/setter)
 * Q (getter/setter)
 * I
 * connect/disconnect
 * voltage level

Making the API a little more user-friendly.

yichen88:
Currently we have to bypass equipment's terminal to modify or access those physical quantities.
In this implements, we add default methods on interface to access those variables.

fix #1536 

**Other information**:


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Branch.java
```diff
@@ -179,4 +179,162 @@
...
+        return getTerminal1().getP();
+    }
+
+    /**
+     * Set the active power in MW injected at side one of this equipment.
+     * <p>
+     * Depends on the working variant.
+     * @see VariantManager
+     */
+    default Branch<I> setP1(double p) {
```
MioRtia, line 199: For me, it should return directly `I` not `Branch<I>` (same remark for other setters)

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Injection.java
```diff
@@ -18,4 +18,83 @@
...
+        return getTerminal().getP();
+    }
+
+    /**
+     * Set the active power in MW injected at this equipment.
+     * <p>
+     * Depends on the working variant.
+     * @see VariantManager
+     */
+    default Injection<I> setP(double p) {
```
MioRtia, line 37: Same remark as above, should return `I`

MioRtia (commented):
Not sure if you should put the implementation in the implementation classes as well (even if it is the same as the default implementation)?

iidm/iidm-api/src/test/java/com/powsybl/iidm/network/util/BranchTest.java
```diff
@@ -0,0 +1,296 @@
...
+/**
+ * @author Yichen TANG <yichen.tang at rte-france.com>
+ */
+public class BranchTest {
+
+    @Test
+    public void testPQI() {
+        Terminal t1 = mock(Terminal.class);
+        Terminal t2 = mock(Terminal.class);
+        Branch branch = new Branch() {
```
sylvlecl, line 29: Might be shorter to use a Mock + `thenCallRealMethod()`. Not sure it really has some added value to be honest ...

sylvlecl (approved):

MioRtia (approved):
I agree with Sylvain's comment, otherwise this is great for me!

PR #1545, opened 2020-11-21
Fix : all VLs with the same nominal voltage in a substation are now created

SlimaneAmar:
Signed-off-by: Slimane AMAR <slimane.amar@rte-france.com>

- [ x] Tests for the changes have been added (for bug fixes / features)

No

Bug

When multiples VLs with the same nominal voltage are present in a substation, only one is created
which contains all the buses

All VLs with the same nominal voltage in a substation are now created

No

**Other information**:

### DIFF
--- a/ucte/ucte-network/src/main/java/com/powsybl/ucte/network/ext/UcteNetworkExt.java
+++ b/ucte/ucte-network/src/main/java/com/powsybl/ucte/network/ext/UcteNetworkExt.java
@@ -212,7 +212,8 @@ public class UcteNetworkExt implements UcteNetwork {
                             nodeCode2.getUcteCountryCode() == UcteCountryCode.XX) {
                         return -1;
                     } else {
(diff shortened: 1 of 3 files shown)


ucte/ucte-network/src/main/java/com/powsybl/ucte/network/ext/UcteNetworkExt.java
```diff
@@ -123,21 +123,35 @@ public UcteRegulation getRegulation(UcteElementId transfoId) {
         }
 
         // in the same substation...
-        addEdgeBetweenSameGeographicalSpotNodes(network, graph);
+        addEdgeBetweenSameGeographicalSpotNodes(network, Optional.empty(), graph);
```
geofjamg, line 126: It is really not recomanded to use an optiona as a parameter, only for return type.

PR #1547, opened 2020-11-24
Fix regulating equipment control enabled (generators)

zamarrenolm:

Bug fix.

CGMES equipment participating in a regulating control may specify that its control is not currently enabled. This info is given in the SSH profile. In addition, the regulating control itself may be considered enabled or disabled (also given in SSH).

For some generators (converted from CGMES synchronous machines and external network injections), the control enabled flag from the equipment was not read.

The problem has been fixed. An IIDM generator will have its voltage regulation set to `ON` if its regulating control is enabled AND its equipment control is also enabled.


annetill:
_The problem has been fixed. An IIDM generator will have its voltage regulation set to ON if its regulating control is enabled AND its equipment control is also enabled._ In the case of a generator with its regulating control enabled AND its equipment control disabled, I understand that the regulating status of the generator in the IIDM model is set to `false`. But it is a way to model in CGMES that a generator is in saturation ? If yes, how do you say to IIDM that the generator has voltage regulating capability ?

zamarrenolm:
> _The problem has been fixed. An IIDM generator will have its voltage regulation set to ON if its regulating control is enabled AND its equipment control is also enabled._ In the case of a generator with its regulating control enabled AND its equipment control disabled, I understand that the regulating status of the generator in the IIDM model is set to `false`. But it is a way to model in CGMES that a generator is in saturation ? If yes, how do you say to IIDM that the generator has voltage regulating capability ?

Checking the available CGMES documentation, it simply says that `RegulatingCondEq.controlEnabled` _Specifies the regulation status of the equipment.  True is regulating, false is not regulating._

In my opinion, it is a flag that specifies that the equipment should not be considered for regulation in the power flow calculations.

A saturated generator in a power flow solution would have its control enabled and its regulating control also enabled in the input. In the output, we should observe that the `Q` provided by the generator has moved to one of the `Qmin` / `Qmax` limits. Voltage setpoint may still be reached if multiple generators contribute to regulate voltage (some of them could be saturated, but remaining ones can keep the voltage). A second power flow execution on the same case could find an alternative solution where that particular generator is not saturated (if multiple alternatives exist). Also, after small changes in the case (reducing load or others) and a power flow re-run, the generator saturation may be lifted. 

In conclusion, if a power flow solution ends with a saturated generator, its regulation status should not be disabled in the outputs. The fact that is saturated would be observed in its reactive power output and eventually in the controlled voltage.

marqueslanauja (approved):

annetill (approved):
I approved this PR, but I would like to make some functional tests in order to evalute the number of CGMES files of our dataset that don't follow this rule. This impact in the voltage plan could be huge and has to be explained if needed.

MioRtia (approved):

PR #1555, opened 2020-12-02
[WIP] Cgmes fix regulating terminal

marqueslanauja:

Feature

In some models there are `phaseTapChangers` controlling flow where the `regulatingTerminal` is associated with a `busBarSection` terminal.

The `regulatingTerminal`  has been moved to the `transformerEnd` terminal if both are at the same `connectivityNode` (They are connected only by `internalConnections`)

**Other information**:

### DIFF
--- a/cgmes/cgmes-conformity/src/test/java/com/powsybl/cgmes/conformity/test/CgmesConformity1ModifiedCatalog.java
+++ b/cgmes/cgmes-conformity/src/test/java/com/powsybl/cgmes/conformity/test/CgmesConformity1ModifiedCatalog.java
@@ -776,7 +776,28 @@ public final class CgmesConformity1ModifiedCatalog {
         String baseBoundary = ENTSOE_CONFORMITY_1
                 + "/MiniGrid/NodeBreaker/CGMES_v2.4.15_MiniGridTestConfiguration_Boundary_v3/";
         return new TestGridModelResources(
-                "MiniGrid-NodeBreaker-PhaseTapChangerMoveRegulatingTerminalThroughAchain",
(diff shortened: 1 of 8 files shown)


MioRtia (commented):
The solution is okay for me but shouldn't this fix be also done for reactive power regulation of generators?

MioRtia (changes requested):
Second comment: maybe it is better (less confusing) to modify the method `findRegulatingTerminal` and filter bbs terminals if it is a flow regulation

zamarrenolm:
> Second comment: maybe it is better (less confusing) to modify the method `findRegulatingTerminal` and filter bbs terminals if it is a flow regulation

I would keep `findRegulatingTerminal` as it is.

The `findRegulatingTerminal` works ok here: the regulating terminal specified in the regulating control has been mapped to IIDM and is connected, so it is found directly.

The problem is that the original CGMES definition of the regulating terminal was incorrect. We are trying to fix a problem in the original data, not trying to recover an equivalent terminal that has been lost during the mapping from CGMES to IIDM.

zamarrenolm:
> The solution is okay for me but shouldn't this fix be also done for reactive power regulation of generators?

You are totally right, a similar reasoning applies to this kind of regulation. We have not found any real world case yet, but we could apply a similar fix for that situation.

annetill:
@zamarrenolm  Can we close this PR?

zamarrenolm:
> @zamarrenolm Can we close this PR?

I think so. We will analyze this after the changes we have made in #2042. If the case described here does not work after the changes in #2042 we will re-implement the solution using the new `RegulatingTerminalMapper` class.

PR #1562, opened 2020-12-09
Improve ShuntCompensatorAdder implementation

mathbagu:

N/A

Refactoring

The design of the `ShuntCompensatorAdderImpl` is a bit weird and difficult to implement in C++.

We change a little bit the implementation to set the model after the `ShuntCompensator` creation. We also ensure that the model is set only once.
I also fix the `setGeneration` method to add a similar check.

**Other information**:


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorImpl.java
```diff
@@ -59,9 +60,11 @@
         }
     }
 
-    private AbstractShuntCompensatorModel attach(AbstractShuntCompensatorModel model) {
-        model.setShuntCompensator(this);
-        return model;
+    void setModel(ShuntCompensatorModelExt model) {
```
tadam50, line 63: Make it fluent, can be usefull ?

MioRtia (commented):
It is okay for me, just be careful for the conflict with master in `DanglingLineImpl`

PR #1566, opened 2020-12-10
UCTE export: ignore fictitious YNodes introduced for transformers at boundary

zamarrenolm:

Bug fix for export of UCTE network that contains boundary configurations where a transformer is directly connected to an XNode in UCTE Network: `Transformer -- XNode`.
During import these configurations are detected and converted to `Transformer -- YNode -- DanglingLine -- XNode`. The `YNode` is a fictitious bus that allows to keep the whole transformer model inside the IIDM Network without having to create special elements or expand the `DanglingLine` to a more complex object. The `DanglingLine` created between the `YNode` and the `XNode` should be ideally a zero-impedance line, actually it has a very low impedance.

IIDM Network buses that correspond to YNodes are considered for export in UCTE format. The export fails because the identifier assigned to the YNode does not follow the UCTE conventions.

IIDM Network buses corresponding to `YNodes` and dangling lines between `YNodes` and `XNodes` are filtered for UCTE export.
The original transformer at boundary can be exported, because its identifier already refers to the original `XNode`, there is no need to modify the export of the transformer.

**Other information**:
We should consider making the check for `YNodes` more robust in the code. There is a related TODO annotation in `UcteExporter` class, in the method `isYNode(Bus)`.


annetill (approved):
I have added a test and a fix to retrieve the Xnode name that was missing at the export.

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
```diff
@@ -547,27 +548,18 @@ private static TwoWindingsTransformer createXnodeTransfo(UcteNetworkExt ucteNetw
 
         yVoltageLevel.getBusBreakerView().newBus()
                 .setId(yNodeName)
+                // TODO(UCTETransformerAtBoundary) consider setting fictitious to true ?
```
murgeyseb, line 551: I agree with this

PR #1573, opened 2020-12-14
Add alphabetical sort criterion to determine the main node in order to have name invariance of a substation

SlimaneAmar:

No

Feature

The name of a substation is  determined by the main node (highest voltage level and the lowest busbar number).
When there are several candidate nodes it is the first of the list and this list is not invariant.

When there are several candidate nodes for a main node in a substation, it is always the lexicographically greater that will be chosen.

No


mathbagu:
@SlimaneAmar Could you please try to solve the sonar issue by reducing the complexity of the algorithm?

ucte/ucte-converter/src/test/java/com/powsybl/ucte/converter/UcteImporterTest.java
```diff
@@ -176,5 +179,16 @@ public void testVoltageRegulatingXnode() {
         assertEquals(1.0, dl.getGeneration().getReactiveLimits().getMaxQ(dl.getGeneration().getTargetP()), 0.01);
         assertEquals(-1.0, dl.getGeneration().getReactiveLimits().getMinQ(dl.getGeneration().getTargetP()), 0.01);
     }
+
+    @Test
+    public void substationNameInvariance() {
+        ResourceDataSource dataSource = new ResourceDataSource("VLsCreationIssue", new ResourceSet("/", "VLsCreationIssue.uct"));
+
+        IntStream.iterate(0, i -> i + 1).limit(100).forEach(i -> {
```
mathbagu, line 187: Use `IntStream.range(0, 100)` instead of `IntStream.iterate(0, i -> i + 1).limit(100)`

ucte/ucte-network/src/main/java/com/powsybl/ucte/network/ext/UcteNetworkExt.java
```diff
@@ -201,7 +200,8 @@ private void updateSubstation() {
                                     nodeCode2.getUcteCountryCode() == UcteCountryCode.XX) {
                                 return -1;
                             } else {
-                                return compareVoltageLevelThenBusbar(nodeCode1, nodeCode2);
+                                int c = compareVoltageLevelThenBusbar(nodeCode1, nodeCode2);
+                                return (c != 0) ? c : nodeCode2.compareTo(nodeCode1); // Alphabetical order to always have the same main node (invariant)
                             }
                         })
                         .findFirst()
```
mathbagu, line 207: Looking to the code, this algorithm doesn't return the first node in the alphabetic order. I change your unit test like this:
```
##C 2007.05.01
Test case for UCTE import issue in PowSyBl core package
When multiples VLs with the same nominal voltage are present in a substation, only one is created
contact : slimane.amar@rte-france.com

##N
##ZFR
F1TEST11 1TEST11 1    0 0        0.15546 72.7296 0.00000 0.00000
F1TEST21 1TEST21 1    0 0        0.15546 72.7296 0.00000 0.00000
F1TEST2A 1TEST2A 1    0 0        0.15546 72.7296 0.00000 0.00000
FTESTA11 TESTAP7      0 0        0.00000 0.00000 0.00000 0.00000
FTESTU11 TESTUP7 1-2  0 0        0.00000 0.00000 0.00000 0.00000

##T
F1TEST11 FTESTU11 1 0 380.0 380.0 1186. 0.1280 44.280 0.000000 0.0000   1803 TESTUY771
FTESTA11 FTESTU11 1 0 380.0 380.0 1186. 0.1280 44.280 0.000000 0.0000   1803 TESTAY771
```

With your proposal, it returns `FTESTU11` but the first node in alphabetical order is `F1TEST11`. I think you have to fix the implementation to be simpler to understand or to explain what you expect.

This is my proposal:
- first, extract this lambda to a well named method and replace the sorted + findfirst by a min()
- then change the implementation of the lambda:
    - Prefer the real node rather than an XNODE (line 196-201 of the current implementation)
    - Prefer the highest voltage level (line 178)
    - Prefer the first busbar number: numbers before letters (line 181 has to be fixed)
    - Prefer the first nodeCode : use the alphabetical order

mathbagu (changes requested):

mathbagu:
@SlimaneAmar Could you check my proposal and approve / reject it?

ucte/ucte-converter/src/test/java/com/powsybl/ucte/converter/UcteImporterTest.java
SlimaneAmar: Done

ucte/ucte-network/src/main/java/com/powsybl/ucte/network/ext/UcteNetworkExt.java
SlimaneAmar: In the description above it is the lexicographically greater therefore the last and not the first.

SlimaneAmar: This choice is to give preference to a name that does not contain a number or with a right most number.
But it's not necessary.

SlimaneAmar: I let you decide.

SlimaneAmar: The only requirement is to generate a name of a susbstation in a deterministic way in order to be able to perform comparison tests of 2 imports.
So i think your fix is a sufficient solution.

MioRtia (approved):

PR #1585, opened 2020-12-22
Use ServiceLoader to create SecurityAnalysis

tadam50:

No

Refactoring of security-analysis-api

Factory mechanism is used to create implementation of SecurityAnalysis

Service Loader is used to create implementation of SecurityAnalysis

**Other information**:

In order to reduce changes on this PR to speed up reviews. A new class 'SecurityAnalysis2' was added instead of change 'SecurityAnalysis'. 
This will be done on next PR

### DIFF
--- a/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysis2.java
+++ b/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysis2.java
@@ -81,38 +81,6 @@ public final class SecurityAnalysis2 {
             return runAsync(network, LocalComputationManager.getDefault());
         }
 
-        public CompletableFuture<SecurityAnalysisResultWithLog> runAsyncWithLog(Network network,
-                                                                                String workingStateId,
-                                                                                LimitViolationDetector detector,
-                                                                                LimitViolationFilter filter,
-                                                                                ComputationManager computationManager,
-                                                                                SecurityAnalysisParameters parameters,
-                                                                                ContingenciesProvider contingenciesProvider,
-                                                                                List<SecurityAnalysisInterceptor> interceptors) {
-            Objects.requireNonNull(network, "Network should not be null");
-            Objects.requireNonNull(workingStateId, "WorkingVariantId should not be null");
-            Objects.requireNonNull(detector, "LimitViolation detector should not be null");
-            Objects.requireNonNull(filter, "LimitViolation filter should not be null");
-            Objects.requireNonNull(computationManager, "ComputationManager should not be null");
-            Objects.requireNonNull(contingenciesProvider, "Contingencies provider should not be null");
-            Objects.requireNonNull(parameters, "Security analysis parameters should not be null");
-            Objects.requireNonNull(interceptors, "Interceptor list should not be null");
-            return provider.runWithLog(network, workingStateId, detector, filter, computationManager, parameters, contingenciesProvider, interceptors);
-        }
-
-        public CompletableFuture<SecurityAnalysisResultWithLog> runAsyncWithLog(Network network, LimitViolationFilter filter,
-                                                                                ComputationManager computationManager) {
-            return runAsyncWithLog(network, network.getVariantManager().getWorkingVariantId(), new DefaultLimitViolationDetector(), filter, computationManager, SecurityAnalysisParameters.load(), new EmptyContingencyListProvider(), Collections.emptyList());
-        }
-
-        public CompletableFuture<SecurityAnalysisResultWithLog> runAsyncWithLog(Network network, ComputationManager computationManager) {
-            return runAsyncWithLog(network, LimitViolationFilter.load(), computationManager);
-        }
-
-        public CompletableFuture<SecurityAnalysisResultWithLog> runAsyncWithLog(Network network) {
-            return runAsyncWithLog(network, LocalComputationManager.getDefault());
-        }
-
         public SecurityAnalysisResult run(Network network,
                                           String workingStateId,
                                           LimitViolationDetector detector,
@@ -135,30 +103,6 @@ public final class SecurityAnalysis2 {
 
         public SecurityAnalysisResult run(Network network) {
             return run(network, LocalComputationManager.getDefault());
-        }
-
-        public SecurityAnalysisResultWithLog runWithLog(Network network,
-                                                        String workingStateId,
-                                                        LimitViolationDetector detector,
-                                                        LimitViolationFilter filter,
-                                                        ComputationManager computationManager,
-                                                        SecurityAnalysisParameters parameters,
-                                                        ContingenciesProvider contingenciesProvider,
-                                                        List<SecurityAnalysisInterceptor> interceptors) {
-            return runAsyncWithLog(network, workingStateId, detector, filter, computationManager, parameters, contingenciesProvider, interceptors).join();
-        }
-
-        public SecurityAnalysisResultWithLog runWithLog(Network network, LimitViolationFilter filter,
-                                                        ComputationManager computationManager) {
-            return runWithLog(network, network.getVariantManager().getWorkingVariantId(), new DefaultLimitViolationDetector(), filter, computationManager, SecurityAnalysisParameters.load(), new EmptyContingencyListProvider(), Collections.emptyList());
-        }
-
-        public SecurityAnalysisResultWithLog runWithLog(Network network, ComputationManager computationManager) {
-            return runWithLog(network, LimitViolationFilter.load(), computationManager);
-        }
-
-        public SecurityAnalysisResultWithLog runWithLog(Network network) {
-            return runWithLog(network, LocalComputationManager.getDefault());
         }
 
         @Override
@@ -219,30 +163,6 @@ public final class SecurityAnalysis2 {
         return find().runAsync(network);
     }
 
-    public static CompletableFuture<SecurityAnalysisResultWithLog> runAsyncWithLog(Network network,
-                                                                                   String workingStateId,
-                                                                                   LimitViolationDetector detector,
-                                                                                   LimitViolationFilter filter,
-                                                                                   ComputationManager computationManager,
-                                                                                   SecurityAnalysisParameters parameters,
-                                                                                   ContingenciesProvider contingenciesProvider,
-                                                                                   List<SecurityAnalysisInterceptor> interceptors) {
-        return find().runAsyncWithLog(network, workingStateId, detector, filter, computationManager, parameters, contingenciesProvider, interceptors);
-    }
-
-    public static CompletableFuture<SecurityAnalysisResultWithLog> runAsyncWithLog(Network network, LimitViolationFilter filter,
-                                                                            ComputationManager computationManager) {
-        return find().runAsyncWithLog(network, filter, computationManager);
-    }
-
-    public static CompletableFuture<SecurityAnalysisResultWithLog> runAsyncWithLog(Network network, ComputationManager computationManager) {
-        return find().runAsyncWithLog(network, computationManager);
-    }
-
-    public static CompletableFuture<SecurityAnalysisResultWithLog> runAsyncWithLog(Network network) {
-        return find().runAsyncWithLog(network);
-    }
-
     public static SecurityAnalysisResult run(Network network,
                                              String workingStateId,
                                              LimitViolationDetector detector,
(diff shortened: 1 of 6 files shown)


security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysis2.java
```diff
@@ -0,0 +1,292 @@
...
+import java.util.concurrent.CompletableFuture;
+
+/**
+ *
+ * Sensitivity analysis main API. It is a utility class (so with only static methods) used as an entry point for running
+ * a sensitivity analysis allowing to choose either a specific implementation or just to rely on the default one.
+ *
+ * @author Thomas Adam <tadam at silicom.fr>
+ */
+public final class SecurityAnalysis2 {
```
mathbagu, line 33: Add a TODO to not forget to rename this class at the end

security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysis2.java
```diff
@@ -0,0 +1,292 @@
...
+                                                                  ComputationManager computationManager) {
+            return runAsync(network, network.getVariantManager().getWorkingVariantId(), new DefaultLimitViolationDetector(), filter, computationManager, SecurityAnalysisParameters.load(), new EmptyContingencyListProvider(), Collections.emptyList());
+        }
+
+        public CompletableFuture<SecurityAnalysisResult> runAsync(Network network, ComputationManager computationManager) {
+            return runAsync(network, LimitViolationFilter.load(), computationManager);
+        }
+
+        public CompletableFuture<SecurityAnalysisResult> runAsync(Network network) {
+            return runAsync(network, new LocalComputationManagerFactory().create());
```
mathbagu, line 80: Use `LocalComputationManager.getDefault()` instead

security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysis2.java
```diff
@@ -0,0 +1,292 @@
...
+            Objects.requireNonNull(network, "Network should not be null");
+            Objects.requireNonNull(workingStateId, "WorkingVariantId should not be null");
+            Objects.requireNonNull(detector, "LimitViolation detector should not be null");
+            Objects.requireNonNull(filter, "LimitViolation filter should not be null");
+            Objects.requireNonNull(computationManager, "ComputationManager should not be null");
+            Objects.requireNonNull(contingenciesProvider, "Contingencies provider should not be null");
+            Objects.requireNonNull(parameters, "Sensitivity analysis parameters should not be null");
+            Objects.requireNonNull(interceptors, "Interceptor list should not be null");
+            return provider.runWithLog(network, workingStateId, detector, filter, computationManager, parameters, contingenciesProvider, interceptors);
+        }
```
mathbagu, line 100: I think you should remove all the `WithLog` method: as proposed last time, we would probably return an object that contain the result of the simulation and a report. To be discussed with the others

security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisProvider.java
```diff
@@ -0,0 +1,58 @@
...
+    default CompletableFuture<SecurityAnalysisResultWithLog> runWithLog(Network network,
+                                                                        String workingVariantId,
+                                                                        LimitViolationDetector detector,
+                                                                        LimitViolationFilter filter,
+                                                                        ComputationManager computationManager,
+                                                                        SecurityAnalysisParameters parameters,
+                                                                        ContingenciesProvider contingenciesProvider,
+                                                                        List<SecurityAnalysisInterceptor> interceptors) {
+        return run(network, workingVariantId, detector, filter, computationManager, parameters, contingenciesProvider, interceptors).thenApply(r -> new SecurityAnalysisResultWithLog(r, null));
+    }
```
mathbagu, line 57: Same remark: I think you can remove this method except if it breaks the compilation

security-analysis-api/src/test/java/com/powsybl/security/SecurityAnalysis2Test.java
```diff
@@ -0,0 +1,159 @@
...
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+
+import static org.junit.Assert.*;
+
+/**
+ * @author Thomas Adam<tadam at silicom.fr>
+ */
+public class SecurityAnalysis2Test {
```
mathbagu, line 31: Same remark: add a todo/fixme

geofjamg:
Why do we choose n this case to keep old classes? SecurityAnalysis2 is ugly. Why not just replace with new API, this is what we did in previous cases (loadflow and sensitivity analysis)?

geofjamg:
> Why do we choose n this case to keep old classes? SecurityAnalysis2 is ugly. Why not just replace with new API, this is what we did in previous cases (loadflow and sensitivity analysis)?

Ok, you explained why in the description, sorry.

security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysis2.java
geofjamg: Isn't it already the purpose of SecurityAnalysisResultWithLog?

security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisImplProvider.java
```diff
@@ -0,0 +1,50 @@
...
+    @Override
+    public CompletableFuture<SecurityAnalysisResult> run(Network network,
+                                                         String workingVariantId,
+                                                         LimitViolationDetector detector,
+                                                         LimitViolationFilter filter,
+                                                         ComputationManager computationManager,
+                                                         SecurityAnalysisParameters parameters,
+                                                         ContingenciesProvider contingenciesProvider,
+                                                         List<SecurityAnalysisInterceptor> interceptors) {
+        SecurityAnalysisImpl securityAnalysis = new SecurityAnalysisImpl(network, detector, filter, computationManager);
```
geofjamg, line 36: So in the next PR, the plan is to move the code of SecurityAnalysisImpl in this class?

security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisProvider.java
```diff
@@ -0,0 +1,58 @@
...
+     * @param filter
+     * @param computationManager
+     * @param parameters specific security analysis parameters
+     * @param contingenciesProvider provides list of contingencies
+     * @param interceptors
+     * @return a {@link CompletableFuture} on {@link SecurityAnalysisResult} that gathers sensitivity factor values
+     */
+    CompletableFuture<SecurityAnalysisResult> run(Network network,
+                                                  String workingVariantId,
+                                                  LimitViolationDetector detector,
```
geofjamg, line 41: For me LimitViolationDetector has nothing to do in the API, because it supposes that violation detection is done on IIDM model. It is not the case for instance in OLF, because violation detection is  done one a more internal data model to speed up performance. I guess it should be the case too, with hades2.
@sylvlecl i think this is you who add it to the API, do you really need it?

security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysis2.java
geofjamg: Maybe we should just remove SecurityAnalysisResultWithLog and move logBytes to SecurityAnalysisResult (and up to an impl to use it or not) so that the API will be cleaner?

geofjamg (changes requested):

security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisImplProvider.java
```diff
@@ -0,0 +1,50 @@
...
+import com.powsybl.security.interceptors.SecurityAnalysisInterceptor;
+
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * @author Thomas Adam <tadam at silicom.fr>
+ */
+@AutoService(SecurityAnalysisProvider.class)
+public class SecurityAnalysisImplProvider implements SecurityAnalysisProvider {
```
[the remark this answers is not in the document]
sylvlecl: Also, is it OK to leave it in the security-analysis-api module ?

The consequence is the following: you will always have that implementation in the classpath, which means that if you add another implementation you want to use, you will not be able to rely on the behaviour which allows to configure nothing to use the only implementation in the classpath.

security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysis2.java
sylvlecl: There was one question with this approach : how do you handle serialization of this ? Do you want to serialize it in the JSON representing the result or not ?

security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisProvider.java
sylvlecl: Indeed it's an important component for us, to allow custom violation detection, for example taking into account limit reduction factors based on the branch, the contingency, or the limit itself (acceptable duration ... ).

A goal of the detector was indeed to try to avoid differences in the way violations are created by different implementations, and to provide the same features to different implementations (like this "limit reduction factors" feature).

If we don't want to use it here anymore, this kind of processing will need to be done "externally" by a calling process, in this case :
* we should be very clear in the javadoc how detection violation should be performed by the implementation (for example, should it provide 1 violation for each exceeded limit, or only 1 for the "worse")
* a detail to be solved is the fact that for voltage violations, for the moment, we lose the "bus" information
* `itools security-analysis` will lose some functionality, except if we add the possibility to plug in some post-processing

security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisImplProvider.java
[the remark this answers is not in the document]
geofjamg: > `LoadFlowBasedSecurityAnalysis`

`DefaultSecurityAnalysis` would be fine. `LoadFlowBasedSecurityAnalysis` is maybe not appropriate as even an optimized security analysis is also based on a load flow.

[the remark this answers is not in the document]
geofjamg: > Also, is it OK to leave it in the security-analysis-api module ?
> 
> The consequence is the following: you will always have that implementation in the classpath, which means that if you add another implementation you want to use, you will not be able to rely on the behaviour which allows to configure nothing to use the only implementation in the classpath.

You are right, it would break the new api design, as it would require to always have the default impl in the config.yml or use .find("h2")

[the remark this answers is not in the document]
jonenst: ManyLoadflowSecurityAnalysis ?

security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisProvider.java
tadam50: Issue #1602 was created to keep these comments

security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysis2.java
tadam50: Issue #1602 was created to keep these comments

security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisImplProvider.java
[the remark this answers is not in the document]
tadam50: Issue #1604 was created to keep these comments

security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisProvider.java
tadam50: Removed waiting fix of Issue #1602

security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysis2.java
```diff
@@ -0,0 +1,292 @@
...
+        private final SecurityAnalysisProvider provider;
+
+        private Runner(SecurityAnalysisProvider provider) {
+            this.provider = Objects.requireNonNull(provider);
+        }
+
+        public CompletableFuture<SecurityAnalysisResult> runAsync(Network network,
+                                                                  String workingStateId,
+                                                                  LimitViolationDetector detector,
+                                                                  LimitViolationFilter filter,
```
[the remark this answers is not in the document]
tadam50: Issue #1603 was created to keep these comments

PR #1591, opened 2021-01-12
able to choose encoding for network xml export

PratheepDev:

Able to choose a Charset for NetworkXml export

Only UTF-8 charset available

add in export option a new parameter for charset


geofjamg:
Could you explain why you would need to use a different encoding? Isn't UTF-8 wide enough to fit every encoding needs?

pl-buiquang:
> Could you explain why you would need to use a different encoding? Isn't UTF-8 wide enough to fit every encoding needs?

We use network export in convergence which reads network in iso-8859-1 encoding

pl-buiquang:
We should add tests to check:
- if the current import correctly handle the xml encoding header (from 8859-1 using incompatible utf8 character)
- if an export with undefined character will throw an exception (using an undefined 8859-1 character)

PratheepDev:
I am not able to found to easy way to throw an exception when the XmlWriter found a undefined character because the writer (com.sun.xml.internal.stream.XMLOutputFactoryImpl) do not implement a solution for that.

I have a proposition, add to Network objet a new attribute encoding and during network writing use network encoding attribute (and we keep the correct encoding) this solution will imply lot of change

miovd (changes requested):

iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/export/ExportOptions.java
```diff
@@ -185,6 +189,14 @@ public ExportOptions setIidmVersionIncompatibilityBehavior(IidmVersionIncompatib
         return this;
     }
 
+    public Charset getCharset() {
+        return charset;
+    }
+
+    public void setCharset(Charset charset) {
+        this.charset = charset;
```
geofjamg, line 197: Add Objects.requireNonNull

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/NetworkXml.java
```diff
@@ -354,6 +354,7 @@ public static Anonymizer write(Network n, ExportOptions options, OutputStream os
             writeExtensions(n, context, options);
             context.getWriter().writeEndElement();
             context.getWriter().writeEndDocument();
+            context.getWriter().flush();
```
geofjamg, line 357: Are you sure you need this, I cannot see how it is related to the encoding config.

iidm/iidm-xml-converter/src/test/java/com/powsybl/iidm/xml/EncodingTest.java
```diff
@@ -0,0 +1,37 @@
+package com.powsybl.iidm.xml;
```
geofjamg, line 1: Licence is missing

iidm/iidm-xml-converter/src/test/java/com/powsybl/iidm/xml/EncodingTest.java
```diff
@@ -0,0 +1,37 @@
...
+import com.powsybl.iidm.network.Network;
+import org.junit.Test;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+
+public class EncodingTest {
```
geofjamg, line 14: Authorship is missing

commons/src/test/java/com/powsybl/commons/AbstractConverterTest.java
```diff
@@ -52,7 +52,7 @@ public void tearDown() throws IOException {
         fileSystem.close();
     }
 
-    protected static void compareXml(InputStream expected, InputStream actual) {
+    public static void compareXml(InputStream expected, InputStream actual) {
```
geofjamg, line 55: @miovd  for another PR: move all these static compare* methods to a dedicated utility class like `AssertUtil.compareXml` to avoid using a utility method from an abstract class.
`AbstractConverterTest.compareXml` is not a clean design.

commons/src/test/java/com/powsybl/commons/xml/XmlUtilTest.java
```diff
@@ -119,4 +122,19 @@ public void readTextTest() throws XMLStreamException {
             }
         }
     }
+
+    @Test
+    public void initializeWriterDefault() throws XMLStreamException {
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        XMLStreamWriter writer = XmlUtil.initializeWriter(true, " ", baos);
```
geofjamg, line 129: To close

commons/src/test/java/com/powsybl/commons/xml/XmlUtilTest.java
```diff
@@ -119,4 +122,19 @@ public void readTextTest() throws XMLStreamException {
...
+        assertEquals("<?xml version=\"1.0\" encoding=\"UTF-8\"?>", baos.toString());
+    }
+
+    @Test
+    public void initializeWriter() throws XMLStreamException {
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        XMLStreamWriter writer = XmlUtil.initializeWriter(false, " ", baos, StandardCharsets.ISO_8859_1);
+        writer.flush();
```
geofjamg, line 137: To close instead

geofjamg (changes requested):

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/NetworkXml.java
Lunaria: it seems that encoding config are cached data and we need to flush after the end in order to write those data.

removing this line, generate test failure when encoding is not set as default

Lunaria:
@geofjamg : I've resolve all comment, I leave you validate the last 2 comments.

commons/src/test/java/com/powsybl/commons/AbstractConverterTest.java
miovd: I created the issue #2208 related to this comment

miovd (approved):

PR #1600, opened 2021-01-22
[WIP] Import the CGMES Area interchanges information

marcosmc:

Feature

Import the CGMES information related to control areas used for interchange in a extension. This extension will be used in balance adjustment project to build a Network Area.


cgmes/cgmes-model/src/main/resources/CIM16.sparql
```diff
@@ -972,6 +972,26 @@ WHERE {
...
+{ GRAPH ?graph {
+    ?TieFlow
+        a cim:TieFlow ;
+        cim:TieFlow.Terminal ?Terminal ;
+        cim:TieFlow.ControlArea ?ControlArea .
+}}
+UNION 
+{ GRAPH ?graphCA {
+    ?ControlArea
+        cim:IdentifiedObject.name ?controlAreaName ;
```
MioRtia, line 987: Don't forget the energyIdentCodeEic

marcosmc: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/areainterchange/CgmesTieFlow.java
```diff
@@ -0,0 +1,30 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.cgmes.conversion.elements.areainterchange;
+
+/**
+ * @author Marcos de Miguel <demiguelm at aia.es>
+ */
+public class CgmesTieFlow {
```
MioRtia, line 12: I'm not sure this object is useful. Maybe using directly `Terminal` is enough

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/areainterchange/CgmesControlArea.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.List;
+
+/**
+ * @author Marcos de Miguel <demiguelm at aia.es>
+ */
+public class CgmesControlArea {
+    private final String id;
+    private final String name;
+    private final String energyIdentCodeEic;
+    private final List<CgmesTieFlow> tieFlows = new ArrayList<>();
```
MioRtia, line 19: Use a `Set` to be sure not to have twice the same object

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/areainterchange/CgmesControlArea.java
```diff
@@ -0,0 +1,53 @@
...
+    public CgmesControlArea(String id, String name, String energyIdentCodeEic, double netInterchange) {
+        this.id = id;
+        this.name = name;
+        this.energyIdentCodeEic = energyIdentCodeEic;
+        this.netInterchange = netInterchange;
+    }
+
+    public void addTieFLow(String tieFlowId, String tieFlowTerminal) {
+        tieFlows.add(new CgmesTieFlow(tieFlowId, tieFlowTerminal));
+    }
```
MioRtia, line 31: In my opinion, this can be replaced by `addTerminal`. We lose the ID but I don't think it really matters

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/areainterchange/CgmesControlArea.java
```diff
@@ -0,0 +1,53 @@
...
+        return name;
+    }
+
+    public String getEnergyIdentCodeEic() {
+        return energyIdentCodeEic;
+    }
+
+    public List<CgmesTieFlow> getTieFlows() {
+        return tieFlows;
+    }
```
MioRtia, line 47: Same remark as above

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/extensions/CgmesControlAreaMappingAdderImpl.java
```diff
@@ -0,0 +1,47 @@
...
+        String controlAreaId = tf.getId("ControlArea");
+        String controlAreaName = tf.getLocal("controlAreaName");
+        String energyIdentCodeEic = tf.getLocal("energyIdentCodeEic");
+        double netInterchange = tf.asDouble("netInterchange");
+
+        String tieFlowId = tf.getId("TieFlow");
+        String tieFlowTerminal = tf.getLocal("terminal");
+
+        cgmesControlAreas.computeIfAbsent(controlAreaId, s -> new CgmesControlArea(controlAreaId, controlAreaName, energyIdentCodeEic, netInterchange)).addTieFLow(tieFlowId, tieFlowTerminal);
+        return this;
```
MioRtia, line 35: Not sure we should use `PropertyBag` in the extension... This treatment might be moved in the conversion class and fields are passed here: `setControlAreaId() setControlAreaName() setEnergyIdentCodeEic()`. In my opinion, terminals should be added later directly in CgmesControlAreaMapping: `mapping.addTerminal(controlAreaId, terminal)`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/extensions/CgmesControlAreaMappingAdderImpl.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    public CgmesControlAreaMappingAdderImpl(Network extendable) {
+        super(extendable);
+    }
+
+    @Override
+    protected CgmesControlAreaMapping createExtension(Network extendable) {
```
MioRtia, line 43: I think you shouldn't wait to have all the cgmesControlArea to create the extension. When you have one, you can create the extension and then you can add others with a method `mapping.addCgmesControlArea()`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/extensions/CgmesControlAreaMappingImpl.java
```diff
@@ -0,0 +1,33 @@
...
+
+    private final List<CgmesControlArea> cgmesControlAreas = new ArrayList();
+
+    CgmesControlAreaMappingImpl(Map<String, CgmesControlArea> cgmesControlAreas) {
+        cgmesControlAreas.values().forEach(cca -> this.cgmesControlAreas.add(cca));
+    }
+
+    @Override
+    public List<CgmesControlArea> getCgmesControlAreas() {
+        return cgmesControlAreas;
```
MioRtia, line 28: I don't agree that it should return a modifiable list. Use `mapping.getCgmesContolArea(cgmesControlAreaId) and `mapping.addCgmesControlArea(cgmesControlAreaId....)` instead ?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/extensions/CgmesControlAreaMappingImpl.java
```diff
@@ -19,14 +17,14 @@
  */
 public class CgmesControlAreaMappingImpl extends AbstractExtension<Network> implements CgmesControlAreaMapping {
 
-    private final List<CgmesControlArea> cgmesControlAreas = new ArrayList();
+    private final Map<String, CgmesControlArea> cgmesControlAreas;
 
     CgmesControlAreaMappingImpl(Map<String, CgmesControlArea> cgmesControlAreas) {
-        cgmesControlAreas.values().forEach(cca -> this.cgmesControlAreas.add(cca));
+        this.cgmesControlAreas = cgmesControlAreas;
```
MioRtia, line 23: Add `Objects.requireNonNull()`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/extensions/CgmesControlAreaMappingAdderImpl.java
zamarrenolm: Agree on avoiding the dependence of `PropertyBag` in the extension

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/areainterchange/CgmesTieFlow.java
zamarrenolm: I think we should keep a TieFlow abstraction. The TieFlow identifier could be required for some later business process. Furthermore, there is an attribute that I think we should store for each TieFlow: `positiveFlowIn`, that states if the flow entering the terminal is also a flow entering the control area. From the CGMES documentation: 

_True if the flow into the terminal (load convention) is also flow into the control area. For example, this attribute should be true if using the tie line terminal further away from the control area. For example to represent a tie to a shunt component (like a load or generator) in another area, this is the near end of a branch and this attribute would be specified as false._

MioRtia: About `positiveFlowIn`: I've seen it but I'm not sure how it is used? In any case, in TSOs I have access to, they are all set to `true`.

Do you have a particular business process in mind? Because for now, it does not have any additive value except of an ID that can be passed as an alias if we really need it...

zamarrenolm: We prefer to keep the relevant information about the `TieFlow`: its identifier, the terminal identifier and the attribute for `positiveFlowIn`. This does not represent a big amount of information: there is a small number of control areas in each IGM (usually only one), each one having a relatively small number of tie flows (in the order of tens).

But if we really want to be strict on space we could drop and keep only the terminal identifier.

MioRtia: The issue I have is more about readibility. It's not really intuitive to me to understand that `CgmesTieFlow` wrap IIDM `Terminal`... Do you have an opinion @annetill ?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/extensions/CgmesControlAreaMappingAdderImpl.java
MioRtia: Another remark: just to be consistent with the rest of the IIDM API, we can use a `CgmesControlAreaAdder`? This way we will be able to use `CgmesControlAreaAdder newCgmesControlArea()` as a method of the mapping and of the mapping adder

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/areainterchange/CgmesTieFlow.java
MioRtia: After some talking with Anne, I think we can start with using only `Terminal` and later a `CgmesTieFlow` object if we notice it is necessary

marcosmc: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/areainterchange/CgmesControlArea.java
marcosmc: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/areainterchange/CgmesControlArea.java
marcosmc: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/areainterchange/CgmesControlArea.java
marcosmc: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/extensions/CgmesControlAreaMappingAdderImpl.java
marcosmc: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/extensions/CgmesControlAreaMappingImpl.java
marcosmc: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/extensions/CgmesControlAreaMappingImpl.java
marcosmc: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/areainterchange/CgmesControlArea.java
```diff
@@ -0,0 +1,82 @@
...
+import com.powsybl.iidm.network.Terminal;
+
+/**
+ * @author Marcos de Miguel <demiguelm at aia.es>
+ */
+public class CgmesControlArea {
+    private final String id;
+    private final String name;
+    private final String energyIdentCodeEic;
+    private final Set<EquipmentEnd> equipmentEns = new HashSet<>();
```
zamarrenolm, line 25: typo. Should be `equipmentEnds`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/areainterchange/CgmesControlArea.java
```diff
@@ -0,0 +1,82 @@
...
+    private final Set<Terminal> terminals = new HashSet<>();
```
zamarrenolm, line 26: To make explicit this is a "cache" of terminals, rename to `cachedTerminals`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/areainterchange/CgmesControlArea.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public CgmesControlArea(String id, String name, String energyIdentCodeEic, double netInterchange) {
+        this.id = id;
+        this.name = name;
+        this.energyIdentCodeEic = energyIdentCodeEic;
+        this.netInterchange = netInterchange;
+    }
+
+    public void addTerminal(String equipmentId, int end) {
+        equipmentEns.add(new EquipmentEnd(equipmentId, end));
```
zamarrenolm, line 37: When a new equipment end is added, the `cachedTerminals` should be refreshed (the corresponding Terminal added) or emptied so it would be rebuilt the next time getTerminals is called.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/areainterchange/CgmesControlArea.java
```diff
@@ -0,0 +1,82 @@
...
+    public String getName() {
+        return name;
+    }
+
+    public String getEnergyIdentCodeEic() {
+        return energyIdentCodeEic;
+    }
+
+    public Set<Terminal> getTerminals(Network network) {
+        if (terminals.isEmpty()) {
```
zamarrenolm, line 53: cached terminals are only computed once. This means if an equipment end is added later, it will no be present in subsequent calls to `getTerminals`.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/areainterchange/CgmesControlArea.java
```diff
@@ -0,0 +1,82 @@
...
+            calculate(network);
+        }
+        return terminals;
+    }
+
+    public double getNetInterchange() {
+        return netInterchange;
+    }
+
+    private void calculate(Network network) {
```
zamarrenolm, line 63: either update cached terminals each time an equipment end is added or mark as invalid (delete all elements) so next time getTerminals is invoked it is rebuilt.

zamarrenolm: If the method to rebuild the cachedTerminals is kept, consider renaming it to `updateCachedTerminals`.
Another option could be to completely remove the terminals collection from this class, and make `getTerminals` a static method that receives the Network and returns the Terminals based on its internal collection of Equipment ends.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -240,6 +248,22 @@ public Network convert() {
...
 
+    private void addControlArea(Context context, CgmesControlAreaAdder cgmesControlAreaAdder, PropertyBag tf) {
+        String controlAreaId = tf.getId("ControlArea");
+        String controlAreaName = tf.getLocal("controlAreaName");
+        String energyIdentCodeEic = tf.getLocal("energyIdentCodeEic");
+        double netInterchange = tf.asDouble("netInterchange");
+        CgmesControlArea cgmesControlArea = cgmesControlAreaAdder.newCgmesControlArea(controlAreaId, controlAreaName, energyIdentCodeEic, netInterchange);
+
+        String terminalId = tf.getId("terminal");
+        Terminal terminal = context.terminalMapping().find(terminalId);
```
annetill, line 259: I wonder if the terminal is linked to a switch (but I don't know if it is allowed indeed).

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/areainterchange/CgmesControlArea.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public Set<EquipmentEnd> getTerminals() {
+        return equipmentEnds;
+    }
+
+    public double getNetInterchange() {
+        return netInterchange;
+    }
+
+    public static class EquipmentEnd {
```
annetill, line 54: By end you mean side?

marcosmc: Yes, I use the same nomenclature used in PR#1540 by Miora

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/extensions/CgmesControlAreaAdderImpl.java
```diff
@@ -0,0 +1,36 @@
...
+public class CgmesControlAreaAdderImpl extends AbstractExtensionAdder<Network, CgmesControlAreaMapping> implements CgmesControlAreaAdder {
+
+    private Map<String, CgmesControlArea> cgmesControlAreas = new HashMap<>();
+
+    public CgmesControlAreaAdderImpl(Network extendable) {
+        super(extendable);
+    }
+
+    @Override
+    public CgmesControlArea newCgmesControlArea(String controlAreaId, String controlAreaName, String energyIdentCodeEic, double netInterchange) {
```
MioRtia, line 27: rather than `newCgmesControlArea`, maybe call it `addCgmesControlArea`, the `new` prefix in the API generally means it returns an adder. The other solution would be to create a full adder but I think it is too much...

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -240,6 +248,22 @@ public Network convert() {
         return network;
     }
 
+    private void addControlArea(Context context, CgmesControlAreaAdder cgmesControlAreaAdder, PropertyBag tf) {
+        String controlAreaId = tf.getId("ControlArea");
+        String controlAreaName = tf.getLocal("controlAreaName");
+        String energyIdentCodeEic = tf.getLocal("energyIdentCodeEic");
+        double netInterchange = tf.asDouble("netInterchange");
+        CgmesControlArea cgmesControlArea = cgmesControlAreaAdder.newCgmesControlArea(controlAreaId, controlAreaName, energyIdentCodeEic, netInterchange);
```
MioRtia, line 256: Okay, this is a bit confusing because reading the code, it seems like you may add the same control area several time (i know it is not true since it is a map in the implementation but the API does not transcribe it

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/extensions/CgmesControlAreaAdderImpl.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import com.powsybl.commons.extensions.AbstractExtensionAdder;
+import com.powsybl.iidm.network.Network;
+
+/**
+ * @author Marcos de Miguel <demiguelm at aia.es>
+ */
+public class CgmesControlAreaAdderImpl extends AbstractExtensionAdder<Network, CgmesControlAreaMapping> implements CgmesControlAreaAdder {
```
MioRtia, line 18: Did you rename it? Since it creates a `CgmesControlAreaMapping`, it should be called `CgmesControlAreaMappingAdder`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
MioRtia: It would only cause issue in one case for us: if the switch is inside a voltage level (so if a control area boundary is inside a voltage level) since a switch linking two voltage levels is a line in IIDM. So to investigate, but I don't think it will make problems.

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesModel.java
```diff
@@ -137,6 +137,8 @@ default PropertyBags fullModel(String cgmesProfile) {
 
     PropertyBags dcTerminalsTP();
 
+    PropertyBags tieFlows();
```
MioRtia, line 140: Add a default implementation (see for example `topologicalIslands()`)

marcosmc: Done

MioRtia (approved):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/extensions/CgmesControlArea.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+public interface CgmesControlArea {
+
+    String getId();
+
+    String getName();
+
+    String getEnergyIdentCodeEic();
+
+    double getNetInterchange();
```
annetill, line 24: It could be nice to explicit the convention chosen to model interchange here in the code.

cgmes/cgmes-extensions/src/main/java/com/powsybl/cgmes/extensions/CgmesControlArea.java
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * @author Miora Ralambotiana <miora.ralambotiana at rte-france.com>
+ */
+public interface CgmesControlArea {
+
+    String getId();
+
+    String getName();
+
+    String getEnergyIdentCodeEic();
```
annetill, line 22: What about `getEICCode` or `getEnergyIdentificationCode` ?

cgmes/cgmes-extensions/src/main/java/com/powsybl/cgmes/extensions/CgmesControlAreas.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.util.Collection;
+
+import com.powsybl.commons.extensions.Extension;
+import com.powsybl.iidm.network.Network;
+
+/**
+ * @author Marcos de Miguel <demiguelm at aia.es>
+ */
+public interface CgmesControlAreas extends Extension<Network> {
```
annetill, line 17: You need this second extension because you can find several `CgmesControlArea` in the same network?

PR #1611, opened 2021-01-28
Fix empty alias type management

mathbagu:

N/A

Feature

The IIDM implementation considers allows empty strings for alias type. Call `addAlias("alias", "")` should be equivalent to call `addAlias("alias", null)` or `addAlias("alias")`. To sum up, empty strings should be consider as null strings for alias types.

This PR change the behaviour of `addAlias` and add unit tests

**Other information**:


MioRtia (commented):
It is okay for me but could you please also update the javadoc? It only says `If the given alias type is null, no alias type is considered associated to the alias.`, you could add `is null or an empty string`.

MioRtia (approved):

Issue #1610, opened 2021-01-28
split time series loses data


jonenst:
* **Do you want to request a *feature* or report a *bug*?**
bug

* **What is the current behavior?**
 [  [ 1 2 3 ] [ 4 5 6 ] ] => split 2  => [ [ 1 2 ] [ 3 ] [ 4 ] ]
The 5,6 values are lost.

* **What is the expected behavior?**
The 5,6 values are not lost.

* **Please tell us about your environment:**
  - PowSyBl Version: all

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)
We should should what we want to return.
Possibilities:
* [[1 2][3][4][5 6]]
  but if you do that, and you split multiple timeseries at the same time with different chunks, you get a desynchronized result:
  [
  [ [ 1 2 3 ] [ 4 5 6 ] ]
  [ [ 1 2 ] [ 3 4 ] [ 5 6 ] ]
  ]
  =>
  [
  [ [ 1 2 ] [ 3 ] [ 4 ] [ 5 6 ] ]
  [ [ 1 2 ] [ 3 4 ] [ 5 6 ]
  ]

* [[1 2][3 4][5 6]]
This is the natural ideal result. But it needs to merge chunks together for this to work. the perceived complexity is probably the reason why this was not implemented.

bhorvilleur-artelys:
Hi,
I am trying to solve this problem (this being my very first contribution to PowSyBl, so please be gentle if my questions are naive).

I haven't been able to reproduce the problem. I created the following unit test:
<code java>
TimeSeriesIndex index = Mockito.mock(TimeSeriesIndex.class);
Mockito.when(index.getPointCount()).thenReturn(6);
TimeSeriesMetadata metadata = new TimeSeriesMetadata("ts1", TimeSeriesDataType.DOUBLE, Collections.emptyMap(), index);

UncompressedDoubleDataChunk chunk = new UncompressedDoubleDataChunk(0,  new double[]{1d, 2d, 3d, 4d, 5d, 6d});
DataChunk.Split<DoublePoint, DoubleDataChunk> splitChunk = chunk.splitAt(3);

StoredDoubleTimeSeries timeSeries = new StoredDoubleTimeSeries(metadata, splitChunk.getChunk1(),, splitChunk.getChunk2());
List<DoubleTimeSeries> split = timeSeries.split(2);
</code>
 
[...]

PR #1616, opened 2021-02-04
Timeseries split

bhorvilleur-artelys:

#1610


time-series/time-series-api/src/main/java/com/powsybl/timeseries/AbstractTimeSeries.java
```diff
@@ -117,36 +117,66 @@ public TimeSeriesMetadata getMetadata() {
...
+            DataChunk.Split<P, C> split = chunkToSplit.splitAt(newChunkLowIndex);
             if (LOGGER.isTraceEnabled()) {
                 LOGGER.trace("   Adding chunk [{}, {}]", split.getChunk1().getOffset(), split.getChunk1().getOffset() + split.getChunk1().getLength() - 1);
             }
-            splitList.add(createTimeSeries(split.getChunk1()));
-            split(split.getChunk2(), splitList, newChunkSize);
+
+            if (usePreviousChunk) {
+                C mergedChunk = previousChunk.merge(split.getChunk1());
+                splitChunks.remove(previousChunk);
```
jonenst, line 154: This is always the last chunk ? remove by index to avoid linearly scanning the arraylist ?

jonenst: Actually, it might be better to design a mutable API. a.merge(b) would modify a in place. This way, in the future, we may be able to implement merging by storing in a the reference to the array in b, and accessing it directly. This would allow to avoid array copies. What do you think ?

bhorvilleur-artelys: Yes, you are right, that would be more performant (less array copies, no creation of chunk, and no modification of the chunk list when usePreviouChunk is true).
I could rename the method extend() to be clear about what the API does. Do you think it is worth keeping the extent method merge, as it is now, for other purposes ?

time-series/time-series-api/src/main/java/com/powsybl/timeseries/AbstractTimeSeries.java
```diff
@@ -117,36 +117,66 @@ public TimeSeriesMetadata getMetadata() {
...
+                splitChunks.add(split.getChunk1());
+            }
+            split(split.getChunk2(), splitChunks, newChunkSize);
         } else {
             if (LOGGER.isTraceEnabled()) {
                 LOGGER.trace("   Too small...");
             }
-            splitList.add(createTimeSeries(chunk));
+            if (usePreviousChunk) {
+                C mergedChunk = previousChunk.append(chunkToSplit);
```
jonenst, line 165: In this case this can lead to quadratic performance for example:

```java
    @Test
    public void splitManyMultiChunkTimeSeriesTest() {
        TimeSeriesIndex index = Mockito.mock(TimeSeriesIndex.class);
        int tsSize = 200000;
        Mockito.when(index.getPointCount()).thenReturn(tsSize);
        TimeSeriesMetadata metadata = new TimeSeriesMetadata("ts1", TimeSeriesDataType.DOUBLE, Collections.emptyMap(),
                index);
        DoubleDataChunk[] chunks = new DoubleDataChunk[tsSize];
        for (int i = 0; i < chunks.length; i++) {
            chunks[i] = new UncompressedDoubleDataChunk(i, new double[] { i });
        }
        StoredDoubleTimeSeries timeSeries = new StoredDoubleTimeSeries(metadata, chunks);
        List<List<DoubleTimeSeries>> split = TimeSeries.split(Collections.singletonList(timeSeries), tsSize);

    }
```
with tsSize = 100000, this codes takes 4s to run.
with tsSize = 200000 (x2), this codes takes 15s(x4) to run.

jonenst (approved):

PR #1618, opened 2021-02-04
Replace SecurityAnalysis factory uses by loaded SecurityAnalysisProvider

tadam50:

No

Feature

security-analysis-api is using factory to create SecurityAnalysis class instances

security-analysis-api is using SecurityAnalysisProvider through service loading

**Other information**:

AbstractSecurityAnalysis & SecurityAnalysis was deleted
SecurityAnalysis2 was renamed to SecurityAnalysis


MioRtia (approved):

security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysis.java
```diff
@@ -1,74 +1,207 @@
...
+        }
+
+        @Override
+        public String getVersion() {
+            return provider.getVersion();
+        }
+    }
 
     /**
-     * To be consistent with {@link #run(String, SecurityAnalysisParameters, ContingenciesProvider)}, this method would also complete exceptionally
```
jonenst, line 40: this delete javadoc is no longer relevant ? Or should it be restored ? With the code example showing the typical exception handling

security-analysis-api/src/main/java/com/powsybl/security/tools/SecurityAnalysisTool.java
```diff
@@ -209,7 +209,7 @@ static SecurityAnalysisInput buildPreprocessedInput(SecurityAnalysisExecutionInp
 
     private static SecurityAnalysisExecutionBuilder createBuilder(PlatformConfig platformConfig) {
         return new SecurityAnalysisExecutionBuilder(() -> ExternalSecurityAnalysisConfig.load(platformConfig),
-            () -> SecurityAnalysisFactories.newDefaultFactory(platformConfig),
+            platformConfig.getModuleConfig(MODULE_CONFIG_NAME_PROPERTY).getStringProperty(DEFAULT_SERVICE_IMPL_NAME_PROPERTY),
```
jonenst, line 212: is it possible to use the standard mechanism PlatformConfigNamedProvider.Finder here instead ? it does a little more than just looking at the field, when there is only one implementation it uses it without any configuration.

security-analysis-api/src/main/java/com/powsybl/security/execution/SecurityAnalysisExecutionImpl.java
```diff
@@ -22,15 +23,25 @@
  */
 public class SecurityAnalysisExecutionImpl implements SecurityAnalysisExecution {
 
-    private final SecurityAnalysisFactory factory;
+    private final String providerName;
```
sylvlecl, line 26: A detail but I would be more in favor of having a `Runner` here :
 * it will allow to make the runner resolution only once and not for every execution
 * it is slightly more flexible since it can allow a third party user to provide its own runner (think for instance to wrap it in a runner which add some additional behaviour such as metrics/logging)

sylvlecl (approved):
Ok in general !
Suggestion to replace security analysis provider name by a "runner" object in "execution" object.

security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysis.java
tadam50: Comments was restored intto SecurityAnalysisProvider class

security-analysis-api/src/main/java/com/powsybl/security/tools/SecurityAnalysisTool.java
tadam50: Related to @sylvlecl comment

security-analysis-api/src/main/java/com/powsybl/security/execution/SecurityAnalysisExecutionImpl.java
tadam50: `String providerName` was replaced by `SecurityAnalysis.Runner runner`

PR #1626, opened 2021-02-12
Psse Support Version 32

marqueslanauja:

Feature

Only versions 33 and 35 were supported.

Versions 32, 33 and 35 are supported.

**Other information**:


marqueslanauja:
We prefer to explicitly see the blocks of each version even if code is duplicated.

zamarrenolm (approved):
Sonar warnings about duplicated code refer to the data blocks defined for every specific PSSE version supported. I think it is preferable to keep the current code, even it is marked as duplicated, because it reflects the structure of the file in each PSSE revision and is easy to check against published documentation.

psse/psse-model/src/main/java/com/powsybl/psse/model/pf/io/PowerFlowRawData32.java
```diff
@@ -0,0 +1,107 @@
+/**
+ * Copyright (c) 2020, RTE (http://www.rte-france.com)
```
MioRtia, line 2: *Copyright 2021

MioRtia (approved):

psse/psse-model/src/main/java/com/powsybl/psse/model/pf/io/PowerFlowRawData32.java
marqueslanauja: Done

geofjamg (approved):

PR #1627, opened 2021-02-12
[WIP] Boundary terminal

MioRtia:

No

Feature

There is no terminal on the boundary side of a dangling line.

There is a terminal on the boundary side of a dangling line. It is not linked to any bus or view, it only represents a connection point to another network. There are also boundary terminals in the middle of a tie line.

Yes

### DIFF
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Boundary.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Boundary.java
@@ -6,10 +6,55 @@
  */
 package com.powsybl.iidm.network;
 
+import com.powsybl.commons.PowsyblException;
+
 /**
  * @author Miora Ralambotiana <miora.ralambotiana at rte-france.com>
  */
 public interface Boundary {
+
+    interface BoundaryTerminal extends Terminal {
+
+        @Override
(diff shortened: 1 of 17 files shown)


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Boundary.java
```diff
@@ -6,16 +6,63 @@
...
+        default void traverse(VoltageLevel.TopologyTraverser traverser) {
+            throw new PowsyblException("Not supported for boundary terminals");
+        }
+    }
+
     double getV();
 
     double getAngle();
 
     double getP();
```
annetill, line 63: You prefer to leave these functions even if they are redondant with the ones accessible from the terminal? Then, I imagine that it is too complex to have a bus view from the boundary terminal, that is why you the access to V and angle remains from boundary object?

zamarrenolm:
To complete the implementation I will try to add methods to write/read references to these terminals in the XML for IIDM.

zamarrenolm:
> To complete the implementation I will try to add methods to write/read references to these terminals in the XML for IIDM.

ok, I see that you will drop the idea of having Terminals at boundary points and prefer to have explicit Boundaries in the control area extension (branch area-interchanges-boundary). So we don't need to extend XIIDM to write the special terminals.

PR #1640, opened 2021-02-26
BusVoltagePerTargetV sensitivity factor

annetill:

No.

A new feature. This PR introduce a new sensitivity factor. The sensitivity function is the voltage at a bus related to a dedicated terminal. As a bus is a very unstable concept, we have to refer to a terminal. The sensitivity variable is the target V at a bus, directly linked to the regulating terminal of one or several equipments responsible of targetV control.

Note that the JSON converter has to be fixed for terminals as we have to use the terminal ref concept.

**Other information**:


sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/factors/variables/TargetV.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Objects;
+
+/**
+ * Sensitivity variable on a voltage target at a regulating terminal
+ *
+ * @author Anne Tilloy {@literal <anne.tilloy at rte-france.com>}
+ */
+public class TargetV extends SensitivityVariable {
+
+    @JsonProperty("variableId")
```
sylvlecl, line 22: I think it should be regulatingTerminalId here ?  We should add those new classes in the JSON tests.

sensitivity-analysis-api/src/test/java/com/powsybl/sensitivity/factors/variables/LinearGlskTest.java
```diff
@@ -68,4 +70,12 @@ public void getGLSKs() {
         LinearGlsk linearGlsk = new LinearGlsk(VARIABLE_ID, VARIABLE_NAME, glskMap);
         assertEquals(2, linearGlsk.getGLSKs().size());
     }
+
+    @Test
+    public void test() throws JsonProcessingException {
```
sylvlecl, line 75: To be removed?

sensitivity-analysis-api/src/test/java/com/powsybl/sensitivity/factors/BusVoltagePerTargetVTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import org.mockito.Mockito;
+
+/**
+ * @author Anne Tilloy {@literal <anne.tilloy at rte-france.com>}
+ */
+public class BusVoltagePerTargetVTest {
```
sylvlecl, line 20: For consistency sake, we should maybe add a JSON test for that new factor in the `json` package.

sylvlecl (changes requested):
Looks good overall : minor changes requested / some minor topics to be decided

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/factors/variables/TargetV.java
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * Constructor
+     *
+     * @param id unique identifier of the variable
+     * @param name readable name of the variable
+     * @param regulatingTerminalId id of the network regulating terminal which targetV increase is used as sensitivity variable
+     * @throws NullPointerException if regulatingTerminalId is null
+     */
+    @JsonCreator
+    public TargetV(@JsonProperty("id") String id,
```
annetill, line 34: Indeed I think it will be really better to use the `BusRef` here to and remove the regulating terminal approach.

sensitivity-analysis-api/src/test/java/com/powsybl/sensitivity/factors/variables/TargetVTest.java
```diff
@@ -0,0 +1,51 @@
...
+import static org.junit.Assert.assertEquals;
+
+/**
+ * @author Anne Tilloy {@literal <anne.tilloy at rte-france.com>}
+ */
+public class TargetVTest {
+
+    private static final String VARIABLE_ID = "Variable ID";
+    private static final String VARIABLE_NAME = "Variable name";
+    private static final String REGULATING_TERMINAL_ID = "Regulating terminal ID";
```
annetill, line 23: To be changed for `BusRef`.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/IdBasedBusRef.java
```diff
@@ -0,0 +1,62 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * @author Yichen TANG <yichen.tang at rte-france.com>
+ */
+public class IdBasedBusRef extends AbstractVoltageLevelBasedBusRef {
```
annetill, line 18: So if I understand well: here the `BusRef` is based on the `id` of a bus from the bus view ? And I can use it to define a sensitivity factor? But, as we will perform a contingency analysis too in the sensitivity analysis, the buses from the bus view are not stable. In the OLF, the a `AC` sensitivity analysis is based on the bus/breaker view. So, it is really complex and dangerous to use this implementation of `BusRef` to model a sensitivity factor.

sylvlecl: I think the use of that implementation is valid for bus breaker voltage levels (which use the "stable" `ConfiguredBus`). 
we should throw for node breaker voltage levels.

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/factors/variables/TargetV.java
sylvlecl: It's not consistent with what I proposed above :) we need to agree : I think it's simpler to have the ID of the regulating equipment here (see mail from Vincent).

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/InjectionBasedBusRef.java
```diff
@@ -0,0 +1,59 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * @author Yichen TANG <yichen.tang at rte-france.com>
+ */
+public class InjectionBasedBusRef implements BusRef {
```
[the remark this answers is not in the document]
sylvlecl: An afterthought : we could also have a simpler approach with only one implementation for all identifiables.
* in case of a configured bus ID, return the bus
* in case of an injection / bus bar, return the bus it belongs to
* otherwise, throw

It would have the benefit of reducing the number of implementations from 3 to 1 without too much loss (it will be slightly less "type safe" ...).

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/factors/variables/TargetV.java
annetill: I think it becomes no simpler for other TSO networks as they rely a lot on shared and remote controls. And, furthermore, we will be obliged to look for the regulating bus inside the loadflow implementation. But, as they is no need for European process it is not a big deal on my side :-)

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/IdBasedBusRef.java
annetill: I think I am understanding that:
```    
@Override
    public Optional<Bus> resolve(Network network) {
        return safeGetVoltageLevel(network)
                .map(VoltageLevel::getBusView)
                .map(busView -> busView.getBus(busId));
    }
```

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/factors/variables/TargetV.java
sylvlecl: Ok, what I don't know is : in that use case, will it be simpler for them to define the regulated bus, or one of the equipments which regulate that bus ? I would think the second is more natural, but not sure.

Anyway, in that case, it will be better to have both possibilities, so I propose to start with the regulating equipment ID and add another implementation when needed, is it OK?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/BusRef.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.util.Optional;
+
+@JsonSubTypes({
+        @JsonSubTypes.Type(value = BranchBasedBusRef.class),
+        @JsonSubTypes.Type(value = IdBasedBusRef.class),
+        @JsonSubTypes.Type(value = InjectionBasedBusRef.class),
+        @JsonSubTypes.Type(value = NodeNumberBasedBusRef.class)
+})
+@JsonTypeInfo(use = JsonTypeInfo.Id.MINIMAL_CLASS)
```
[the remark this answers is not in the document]
yichen88: uniform with wrapper class

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/factors/variables/TargetV.java
annetill: Yes.

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/factors/variables/TargetV.java
annetill: After a short negotiation with @sylvlecl, it seems that in a first approach we can rely on the equipement ID here.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/IdBasedBusRef.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Override
+    public Optional<Bus> resolve(Network network) {
+        final Identifiable<?> identifiable = network.getIdentifiable(id);
+        if (identifiable == null) {
+            return Optional.empty();
+        }
+        if (identifiable instanceof Bus) {
+            String busId = identifiable.getId();
+            return Optional.of(network.getBusView().getBus(busId));
```
sylvlecl, line 39: In that case it is probably better to use the voltage level bus view, it will restrict the search.

Otherwise, for example In iidm-impl, the search will go over the list of all voltage levels.

sylvlecl (approved):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/IdBasedBusRef.java
sylvlecl: **Second thought:**
actually it's not what we want to do. We need to return the bus (of the bus view) which contains that identifiable bus !

I think there is no easy way for this in the API today (it's probably a missing feature). As a workaround we could use terminals : return the first bus which contains one of the terminals of the identifiable bus. In the weird case where there is no terminal connected to the identifiable bus, return empty.

sylvlecl (changes requested):
The behaviour of ID based bus ref on "identifiable" buses needs to be fixed, otherwise it's OK!

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/factors/variables/TargetV.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @JsonProperty("equipmentId")
+    private final String equipmentId;
+
+    /**
+     * Constructor
+     *
+     * @param id unique identifier of the variable
+     * @param name readable name of the variable
+     * @param equipmentId id of the network equipment (generator, static var compensator, ratio tap changer,
```
annetill, line 30: To be fixed here: it is not the id of the ratio tap changer but the id of the transformer.

MioRtia (approved):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/NodeNumberBasedBusRef.java
```diff
@@ -0,0 +1,79 @@
...
+
+        if (!getVoltageLevelId().equals(that.getVoltageLevelId())) {
+            return false;
+        }
+        return getNode() == that.getNode();
+    }
+
+    @Override
+    public int hashCode() {
+        int result = getVoltageLevelId().hashCode();
```
geofjamg, line 74: Use Objects.hash(....)

geofjamg (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/BusRef.java
```diff
@@ -0,0 +1,27 @@
...
+public interface BusRef {
+
+    /**
+     * @return an empty if not found
+     * @throws com.powsybl.commons.PowsyblException if underlying implementation not supported
+     */
+    Optional<Bus> resolve(Network network);
```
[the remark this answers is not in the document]
geofjamg: Thinking a little bit more, why not just get a terminal so user can choose if he gets a bus from bus view or bus/breaker view. And consequently why not call it TerminalRef?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/IdBasedBusRef.java
```diff
@@ -0,0 +1,107 @@
...
+            final Identifiable<?> identifiable = network.getIdentifiable(id);
+            if (identifiable == null) {
+                return Optional.empty();
+            }
+            if (identifiable instanceof Bus) {
+                Bus bus = (Bus) identifiable;
+                if (Objects.equals(TopologyLevel.BUS_BRANCH, level)) {
+                    return Optional.ofNullable(bus.getConnectedTerminalStream().map(t -> t.getBusView().getBus()).filter(Objects::nonNull).findFirst().orElse(null));
+                } else {
+                    return Optional.ofNullable(bus.getConnectedTerminalStream().map(t -> t.getBusBreakerView().getBus()).filter(Objects::nonNull).findFirst().orElse(null));
```
sylvlecl, line 53: Here I would like to return the bus without going through the terminal.

However the API and its doc do not fully guarantee that the configured bus and the bus from the bus breaker view will be the same, although it seems obvious ... so let's keep it this way for now :(

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/IdBasedBusRef.java
```diff
@@ -0,0 +1,107 @@
...
+    Optional<Bus> resolveByLevel(Network network, TopologyLevel level) {
+        if (side == null) {
+            final Identifiable<?> identifiable = network.getIdentifiable(id);
+            if (identifiable == null) {
+                return Optional.empty();
+            }
+            if (identifiable instanceof Bus) {
+                Bus bus = (Bus) identifiable;
+                if (Objects.equals(TopologyLevel.BUS_BRANCH, level)) {
+                    return Optional.ofNullable(bus.getConnectedTerminalStream().map(t -> t.getBusView().getBus()).filter(Objects::nonNull).findFirst().orElse(null));
```
sylvlecl, line 51: Note: you can return the result of `findFirst()`. Also we could reuse the `chooseBusByLevel` here since we rely on the terminal.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/NodeNumberBasedBusRef.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @Override
+    Optional<Bus> resolveByLevel(Network network, TopologyLevel level) {
+        final VoltageLevel voltageLevel = network.getVoltageLevel(voltageLevelId);
+        if (voltageLevel == null) {
+            return Optional.empty();
+        }
+        if (Objects.equals(TopologyKind.NODE_BREAKER, voltageLevel.getTopologyKind())) {
+            final Terminal terminal = voltageLevel.getNodeBreakerView().getTerminal(node);
```
sylvlecl, line 37: A node is not guaranteed to have an associated terminal, so this will not work as intended.

We would need a way to retrive the bus for this node, but the API does not allow it ... As it will not be a need for now I propose to simply delete that implementation (yes it leaves us with only one implementation for now).

sylvlecl (changes requested):
More minor fixes. Proposition to delete the node number implementation for now, the link node -> bus is not straight forward to implement, and not a real need now.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/AbstractBusRef.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * @author Yichen TANG <yichen.tang at rte-france.com>
+ */
+public abstract class AbstractBusRef implements BusRef {
+
+    @Override
+    public Optional<Bus> resolve(Network network, TopologyLevel level) {
+        Objects.requireNonNull(network);
+        Objects.requireNonNull(level);
+        if (Objects.equals(TopologyLevel.NODE_BREAKER, level)) {
```
geofjamg, line 21: why not just == ?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/IdBasedBusRef.java
sylvlecl: After reviewing again the API and javadoc of bus breaker view, I think I was wrong above:
it seems clear that the returned buses must be the same as the one added throught the `newBus` calls, so after all it seems better to me to return the `Bus` without going through the terminal.

sylvlecl (approved):

PR #1642, opened 2021-03-01
CGMES.normalPF is imported as a property

MioRtia:

No

Bug fix: `normalPF` is not equivalent to `ActivePowerControl` extension and will therefore be imported as a property

No


annetill (approved):
The idea here is to make this work-around and to wait for a more sustainable solution with a normalization tool box to deal with the participation factors of the CGM file. The droop, as it is filled, is not correct and our simulators have a wrong slack distribution.

zamarrenolm (approved):

PR #1645, opened 2021-03-02
Optimise getBus in Bus/Breaker view for default Network implementation

zamarrenolm:

Optimisation

Current lookup of bus by identifier in `BusBreakerView` is always done iterating through the voltage level stream of the network. As calculated buses are not included in the network index, bus/breaker buses created from node/breaker topologies have to be searched looking in every voltage level. But buses created directly at the bus/breaker topology level are added to the index, and could be located faster.

When looking for a bus with a given identifier in the `BusBreakerView`, first the index is checked. If the bus is not found, the previous search through all voltage levels of the network is performed.

It is not a breaking change.

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
@@ -76,6 +76,10 @@ class NetworkImpl extends AbstractIdentifiable<Network> implements Network, Vari
 
         @Override
         public Bus getBus(String id) {
+            Bus bus = index.get(id, Bus.class);
+            if (bus != null) {
+                return bus;
+            }
             return getVoltageLevelStream().map(vl -> vl.getBusBreakerView().getBus(id))
                     .filter(Objects::nonNull)
                     .findFirst()


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
```diff
@@ -76,6 +76,10 @@ public int getSwitchCount() {
 
         @Override
         public Bus getBus(String id) {
+            Bus bus = index.get(id, Bus.class);
```
geofjamg, line 79: Are you sure it works, and you don't have to use ConfiguredBusImpl.class ?

zamarrenolm: Yes it works, I have double checked it using a profiled run (performance gain factor around 2) and a debug session. The returned Bus instance is of class ConfiguredBusImpl, but I don't need to pass the subtype to the get method. Bus is enough to find it.

zamarrenolm: The implementation of `index.get` looks in the map of objects by id (without type) and later checks the type of the result:

```java
        Identifiable<?> obj = get(id);
        if (obj != null && clazz.isAssignableFrom(obj.getClass())) {
            return (T) obj;
        } else {
```
and a `Bus` is assignable from the subtype `ConfiguredBusImpl`.

geofjamg: Ok thanks

geofjamg (approved):

PR #1646, opened 2021-03-02
Psse: Sort the shunt blocks of the non-linear model.

marqueslanauja:

Feature

The shunt blocks of the non-linear model of the shunt compensator are not sorted.
The total admittance of the blocks is reset before adding the capacitors.

The shunt blocks of the non-linear model of the shunt compensator are sorted.
The total admittance of the blocks is only reset before adding the capacitors when the  adjustment method (ADJM) is 1.
When ADJM is zero steps and blocks are switched on in input order in the PSS/E and it does not make sense to reset the accumulated admittance before switching on the capacitor blocks.

**Other information**:


zamarrenolm (approved):
I agree with the approach to map PSSE switched shunts with adjustment type == 1 (all combinations of steps and blocks are allowed) to a single shunt compensator in IIDM where sections are built from capacitors and reactors considered separately and sorted by size. This strategy does not cover absolutely all original combinations, but provides enough alternatives to allow voltage control using a single controlling resource.

geofjamg:
> I agree with the approach to map PSSE switched shunts with adjustment type == 1 (all combinations of steps and blocks are allowed) to a single shunt compensator in IIDM where sections are built from capacitors and reactors considered separately and sorted by size. This strategy does not cover absolutely all original combinations, but provides enough alternatives to allow voltage control using a single controlling resource.

@annetill @zamarrenolm it makes sense as a short/middle term solution. Maybe to discuss later if it would worth it to allow a third shunt model in IIDM to better fit to this PSS/E feature.

MioRtia (approved):

psse/psse-converter/src/test/resources/SwitchedShunt.xiidm
```diff
@@ -0,0 +1,140 @@
...
+            <iidm:busBreakerTopology>
+                <iidm:bus id="B5" name="5           " v="486.39" angle="-42.3008"/>
+            </iidm:busBreakerTopology>
+            <iidm:load id="B5-L1 " loadType="UNDEFINED" p0="2000.0" q0="100.0" bus="B5" connectableBus="B5"/>
+            <iidm:load id="B5-L2 " loadType="UNDEFINED" p0="0.0" q0="50.0" connectableBus="B5"/>
+        </iidm:voltageLevel>
+        <iidm:twoWindingsTransformer id="T-3-5-1 " r="0.0" x="28.0" g="0.0" b="0.0" ratedU1="18.0" ratedU2="500.0" bus1="B3" connectableBus1="B3" voltageLevelId1="VL3" bus2="B5" connectableBus2="B5" voltageLevelId2="VL5">
+            <iidm:ratioTapChanger lowTapPosition="0" tapPosition="11" targetDeadband="2.700000000000003" loadTapChangingCapabilities="false" regulating="true" targetV="18.45">
+                <iidm:terminalRef id="B3-SwSH1"/>
+                <iidm:step r="0.0" x="0.0" g="10.803324099723" b="10.803324099723" rho="1.0526315789473684"/>
```
annetill, line 89: b and g are equal here? In general, g is close to zero...

psse/psse-converter/src/main/java/com/powsybl/psse/converter/SwitchedShuntCompensatorConverter.java
```diff
@@ -186,6 +187,11 @@ private static int defineSectionCount(double binit, List<ShuntBlock> shuntBlocks
             }
         }
 
+        // Add the zero block, shunt disconnected
+        shuntBlocks.add(new ShuntBlock(1, 1, 0.0));
+
+        shuntBlocks.sort(Comparator.comparing(ShuntBlock::getB));
```
annetill, line 193: Here, just to be sure: the sections are accumulated from the smaller b to the greater b in absolute value ?

marqueslanauja: Yes, the sections are accumulated from the smaller b to the greater b in absolute value.

psse/psse-converter/src/test/resources/SwitchedShunt.xiidm
marqueslanauja: Here, in the steps of the transformer r, x, g, and b are percentages, the percentages that should be applied to the real values of  r, x, g and b of the transformer for considering the correction after moving the ratio from one end to the other. The g and b percentages are identical as the ratio affects in the same way to both values of the shunt admittance.

annetill (approved):

Issue #1604, opened 2021-01-22
Refactoring security-analysis-api : SecurityAnalysisImpl


tadam50:
* **Do you want to request a *feature* or report a *bug*?**

Feature

* **What is the current behavior?**

SecurityAnalysis implementation is part of security-analysis-api.
The consequence is the following: you will always have that implementation in the classpath

* **What is the expected behavior?**

Rename implementation and move it outside of security-analysis-api.

* **What is the motivation / use case for changing the behavior?**

Refactoring of security-analysis-api

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

_Originally posted by @sylvlecl in https://github.com/powsybl/powsybl-core/pull/1585#discussion_r559442013_
Should we seize this opportunity to rename the `SecurityAnalysisImpl` to something more meaningful, like `DefaultSecurityAnalysis` or even more explicit `LoadFlowBasedSecurityAnalysis`  ?

The "impl" suffix does not seem appropriate if the target is to have multiple implementations (even more when this "impl" is not suitable for most users).

Also, is it OK to leave it in the security-analysis-api module ?

The consequence is the following: you will always have that implementation in the classpath, which means that if you add another implementation you want to use, you will not be able to rely on the behaviour which allows to configure nothing to use the only implementation in the classpath.

_Originally posted by @geofjamg in https://github.com/powsybl/powsybl-core/pull/1585#discussion_r559467254_
> `LoadFlowBasedSecurityAnalysis`

`DefaultSecurityAnalysis` would be fine. `LoadFlowBasedSecurityAnalysis` is maybe not appropriate as even an optimized security analysis is also based on a load flow.

_Originally posted by @geofjamg in https://github.com/powsybl/powsybl-core/pull/1585#discussion_r559468160_
> Also, is it OK to leave it in the security-analysis-api module ?
> 
> The consequence is the following: you will always have that implementation in the classpath, which means that if you add another implementation you want to use, you will not be able to rely on the behaviour which allows to configure nothing to use the only implementation in the classpath.

You are right, it would break the new api design, as it would require to always have the default impl in the config.yml or use .find("h2")

_Originally posted by @jonenst in https://github.com/powsybl/powsybl-core/pull/1585#discussion_r562646994_
ManyLoadflowSecurityAnalysis ?

PR #1648, opened 2021-03-04
Refactoring security-analysis-api : SecurityAnalysisImpl (issue 1604)

tadam50:

#1604 

Feature

security-analysis-api has an default implementation of SecurityAnalysisProvider

A new aggregate module is created : security-analysis
This new module contains previous security-analysis-api and the new one security-analysis-default
security-analysis-default module contains the default implementation of SecurityAnalysisProvider based on LoadFlow

**Other information**:

All interceptors tests have been move into new module security-analysis-default


security-analysis/security-analysis-default/pom.xml
```diff
@@ -0,0 +1,150 @@
...
+
+    <parent>
+        <groupId>com.powsybl</groupId>
+        <artifactId>powsybl-security-analysis</artifactId>
+        <version>3.9.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>powsybl-security-analysis-default</artifactId>
+    <name>Security Analysis Default Implementation</name>
+    <description>An default implementation to run security analysis</description>
```
sylvlecl, line 23: Typo : "A" instead of "An" :)

security-analysis/security-analysis-default/src/main/java/com/powsybl/security/impl/DefaultSecurityAnalysis.java
```diff
@@ -4,7 +4,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
-package com.powsybl.security;
+package com.powsybl.security.impl;
```
sylvlecl, line 7: Should we rename also the package to "default" ?

sylvlecl (approved):

MioRtia (approved):

security-analysis/security-analysis-default/src/main/java/com/powsybl/security/impl/DefaultSecurityAnalysis.java
tadam50: "default" is a keyword, we cannot use it as package name

sylvlecl: Ah ... that's an issue indeed ^^ go for impl then (workarounds like `default_` or `defaultimpl` will not be prettier)

geofjamg (approved):

PR #1659, opened 2021-03-19
CGMES EQ export from IIDM network

marcosmc:

**Other information**:

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
@@ -29,6 +29,7 @@ public class CgmesExportContext {
     private CgmesTopologyKind topologyKind = CgmesTopologyKind.BUS_BRANCH;
     private DateTime scenarioTime = DateTime.now();
 
+    private ModelDescription eqModelDescription = new ModelDescription("EQ Model", CgmesNamespace.EQ_PROFILE);
     private ModelDescription svModelDescription = new ModelDescription("SV Model", CgmesNamespace.SV_PROFILE);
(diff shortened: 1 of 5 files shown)


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/elements/EnergyConsumerEq.java
```diff
@@ -0,0 +1,51 @@
...
+        writer.writeAttribute(RDF_NAMESPACE, CgmesNames.ID, id);
+        writer.writeStartElement(cimNamespace, CgmesNames.NAME);
+        writer.writeCharacters(loadName);
+        writer.writeEndElement();
+        writer.writeEmptyElement(cimNamespace, "Equipment.EquipmentContainer");
+        writer.writeAttribute(RDF_NAMESPACE, CgmesNames.RESOURCE, "#" + equipmentContainer);
+        writer.writeEndElement();
+    }
+
+    private static String loadClassName(LoadDetail loadDetail) {
```
annetill, line 33: I think that conform load means that the variable part is not equal to zero. A non conform load means that the variable part is equal to zero.

marcosmc: Yes, the code follows your interpretation. To avoid duplication, this method will be moved to the module CgmesExportUtil, as it is required for EQ export and also for SSH export

marcosmc:
A modification is added to ensure that the export of the EQ and SSH files use the same Ids for the terminals and the Equivalent Injection of the danglinglines. Until now the SSH file only exported the information of the terminals and Equivalent Injection if it is a CGMES model, obtaining the CGMES Ids of the aliases of these elements

PR #1666, opened 2021-03-24
Functional logs POC

floriand-e2r:

No

Feature

All technical and functional information is logged on the fly

The functional information is piled up thanks to an API `Reporter`.
Each log information is pushed with a key, a default log string in english, values through a `Map<String, Object>`, and an object `Marker` which can be used to tag the type of information.
This API allows to produce a hierarchical report by creating a "child" `Reporter` for each new task and/or thread.
A default implementation `LoggerReporter` is provided. It uses logback to log on the fly and piles up the functional information with a `Map<String, Report>` for post-processing (those reports can be exported to a file for instance).


commons/src/main/java/com/powsybl/commons/reporter/LoggerReporter.java
```diff
@@ -0,0 +1,131 @@
...
+ */
+public class LoggerReporter implements Reporter, ReportSeeker {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(LoggerReporter.class);
+
+    private static final String DEFAULT_ROOT_TASK_KEY = "rootTaskKey";
+
+    private final String taskKey;
+    private final String defaultName;
+    private final LoggerReporter parentReporter;
```
floriand-e2r, line 30: This parameter is not used yet, but could (should?) be used to construct a map containing all descendant reporters. This would allow to access directly any reporter from the root one.

ucte/ucte-network/src/main/java/com/powsybl/ucte/network/UcteRecord.java
```diff
@@ -6,12 +6,14 @@
...
+import com.powsybl.commons.reporter.Reporter;
+
 /**
  *
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
  */
 public interface UcteRecord {
 
-    void fix();
+    void fix(Reporter reporter);
```
MioRtia, line 17: Deprecate and add default implementation?

ucte/ucte-network/src/main/java/com/powsybl/ucte/network/UcteNetwork.java
```diff
@@ -45,6 +47,6 @@
 
     UcteRegulation getRegulation(UcteElementId transfoId);
 
-    void fix();
+    void fix(Reporter reporter);
```
MioRtia, line 50: Deprecate and add default implementation?

ucte/ucte-network/src/test/java/com/powsybl/ucte/network/UcteNetworkImplTest.java
```diff
@@ -21,7 +21,7 @@
 
     @Test
     public void test() {
-        UcteNetwork network = UcteNetworkFactory.createNetwork(UcteNetworkImpl::new);
+        UcteNetwork network = UcteNetworkFactory.createNetwork(() -> new UcteNetworkImpl());
```
MioRtia, line 24: Why did you replace this by a lambda?

MioRtia (commented):
Thank you for your work! It is okay for me, I made some small remarks.

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
```diff
@@ -110,6 +112,7 @@ private static void createBuses(UcteNetworkExt ucteNetwork, UcteVoltageLevel uct
                 continue;
             }
 
+            reporter.report("createBus_" + ucteNodeCode, "Create bus ${bus}", "bus", ucteNodeCode, MarkerImpl.TRACE);
```
geofjamg, line 115: Here we have a design issue. To be translated to another language, keys have to be independent of data.
For instance, here you are going to generate reports with as many key as buses.
Like  createBus_AAAAA1
Like  createBus_AAAAA2
...
How are we going to manage translation dictionnary with this kind of keys?

geofjamg (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ReporterNetworkListener.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.powsybl.commons.reporter.MarkerImpl;
+import com.powsybl.commons.reporter.Reporter;
+
+import java.util.Map;
+
+/**
+ * @author Florian Dupuy <florian.dupuy at rte-france.com>
+ */
+public class ReporterNetworkListener implements NetworkListener {
```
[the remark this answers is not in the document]
floriand-e2r: Removed

ucte/ucte-network/src/test/java/com/powsybl/ucte/network/UcteNetworkImplTest.java
floriand-e2r: Strange! reverted

action/action-simulator/src/test/java/com/powsybl/action/simulator/LoadFlowProviderMock.java
```diff
@@ -38,7 +39,7 @@ public String getVersion() {
     }
 
     @Override
-    public CompletableFuture<LoadFlowResult> run(Network network, ComputationManager computationManager, String workingStateId, LoadFlowParameters parameters) {
+    public CompletableFuture<LoadFlowResult> run(Network network, ComputationManager computationManager, String workingStateId, LoadFlowParameters parameters, Reporter reporter) {
```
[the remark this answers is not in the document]
floriand-e2r: Eventually I kept the previous API and marked it as deprecated. Except UCTE import I didn't overload the new API in implementations.

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
geofjamg: With a builder style we would have:
```java
reporter.withKey("createBus").
            .withDefaultMessage("Create bus ${bus}")
            .withValue("bus", ucteNodeCode)
            .withSeverity("TRACE")
            .report();
```

geofjamg (changes requested):
Maybe also you could provide an in memory `Reporter` implementation that allows to store report in memory and serialize to JSON. I think most if the functional log user will use it so it makes sense to provide it in core.

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
floriand-e2r: Corrected. I got confused with the unicity of keys in one reporter, but that's not needed in fact.

commons/src/main/java/com/powsybl/commons/reporter/LoggerReporter.java
```diff
@@ -0,0 +1,131 @@
...
+        LoggerReporter childReporter = new LoggerReporter(taskKey, defaultName, values, this);
+        childReporters.add(childReporter);
+        return childReporter;
+    }
+
+    public void export(File file) {
+        try (PrintWriter writer = new PrintWriter(file)) {
+            printTaskReport(this, writer, "");
+        } catch (FileNotFoundException e) {
+            e.printStackTrace();
```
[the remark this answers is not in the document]
floriand-e2r: The interfaces have changed, so this is outdated, but in case of IOException in the new equivalent method I throw an UncheckedIOException (in fact I took that behaviour from powsybl-single-line-diagram).

commons/src/main/java/com/powsybl/commons/reporter/LoggerReporter.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    @Override
+    public LoggerReporter createChild(String taskKey, String defaultName, Map<String, Object> values) {
+        LoggerReporter childReporter = new LoggerReporter(taskKey, defaultName, values, this);
+        childReporters.add(childReporter);
+        return childReporter;
+    }
+
+    public void export(File file) {
```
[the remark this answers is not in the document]
floriand-e2r: Corrected. Path is used instead, and a method taking Writer has been added.

ucte/ucte-converter/src/test/java/com/powsybl/ucte/converter/UcteImporterTest.java
```diff
@@ -71,6 +74,15 @@ public void germanTsosImport() throws Exception {
 
     }
 
+    @Test
+    public void germanTsosImportReporterTest() throws Exception {
+        ReadOnlyDataSource dataSource = new ResourceDataSource("germanTsos", new ResourceSet("/", "germanTsos.uct"));
+
+        LoggerReporter loggerReporter = new LoggerReporter();
+        new UcteImporter().importData(dataSource, NetworkFactory.findDefault(), null, loggerReporter);
+        loggerReporter.export(new File("/tmp", "exportTest.txt")); // TODO: compare text file with reference
```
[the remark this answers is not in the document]
floriand-e2r: I finally used a StringWriter instead

commons/src/main/java/com/powsybl/commons/reporter/Marker.java
```diff
@@ -0,0 +1,22 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.commons.reporter;
+
+/**
+ * @author Florian Dupuy <florian.dupuy at rte-france.com>
+ */
+public interface Marker {
```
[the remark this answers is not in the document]
floriand-e2r: Marker removed, values are used instead for gravity/severity.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ReporterNetworkListener.java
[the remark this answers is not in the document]
floriand-e2r: removed.

commons/src/main/java/com/powsybl/commons/reporter/LoggerReporter.java
floriand-e2r: Parameter removed.

commons/src/main/java/com/powsybl/commons/reporter/LoggerReporter.java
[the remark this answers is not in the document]
floriand-e2r: code replaced by `throw new UncheckedIOExcpection(e)`

commons/src/main/java/com/powsybl/commons/reporter/LoggerReporter.java
```diff
@@ -0,0 +1,131 @@
...
+
+    public LoggerReporter() {
+        this(DEFAULT_ROOT_TASK_KEY, "Root task", Collections.emptyMap(), null);
+    }
+
+    public LoggerReporter(String rootTaskKey, String rootDefaultName, Map<String, Object> taskValues, LoggerReporter parent) {
+        this.taskKey = rootTaskKey;
+        this.defaultName = rootDefaultName;
+        this.taskValues = taskValues;
+        this.parentReporter = parent;
```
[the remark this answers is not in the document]
floriand-e2r: Done

commons/src/main/java/com/powsybl/commons/reporter/Report.java
```diff
@@ -0,0 +1,46 @@
...
+    public Object getValue(String valueKey) {
+        return values.get(valueKey);
+    }
+
+    public Marker getMarker() {
+        return marker;
+    }
+
+    public Map<String, Object> getValues() {
+        return values;
```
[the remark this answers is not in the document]
floriand-e2r: Done here and in similar getters.

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
```diff
@@ -147,6 +158,12 @@ private static void createBuses(UcteNetworkExt ucteNetwork, Network network) {
                 continue;
             }
 
+            ReportBuilder substationReport = new ReportBuilder()
```
geofjamg, line 161: You should write it inline:
```java
   substationReporter.report(new ReportBuilder()
                .withKey("createSubstation")
                .withDefaultMessage("Create substation ${substationName}")
                .withValue("substationName", ucteSubstation.getName())
                .withSeverity("TRACE")
                .build());
```

Would be better:
```java
   substationReporter.report(Report.builder()
                .withKey("createSubstation")
                .withDefaultMessage("Create substation ${substationName}")
                .withValue("substationName", ucteSubstation.getName())
                .withSeverity("TRACE")
                .build());
```

geofjamg (changes requested):
Maybe we should also support an optional type for values, like
```java
   substationReporter.report(new ReportBuilder()
                .withKey("createSubstation")
                .withDefaultMessage("Create substation ${substationName}")
                .withValue("substationName", ucteSubstation.getName(), "Substation")
                .withSeverity("TRACE")
                .build());
```

That would allow us to type some values and to known in this example, that this is a substation id and we could be able to have a hyperlink in the UI to navigate from a report UI to a single line diagram or center on it in a the map.

It would be also useful for numeric values to store unit for instance, like MW, MVar et that would allow use in the UI to adapt float precison formatting depending of the type.

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
floriand-e2r: Done, inlined with Report.builder()

ucte/ucte-network/src/main/java/com/powsybl/ucte/network/io/UcteReader.java
```diff
@@ -280,12 +273,16 @@ private void parseRecords(UcteRecordParser parser, UcteNetwork network) throws I
     }
 
     public UcteNetwork read(BufferedReader reader) throws IOException {
+
+        Reporter readReporter = reporter.createChild("UcteReading", "Reading UCTE network file");
```
jonenst, line 277: this kind of key is part of the public API for consummers of these logs to extract and manipulate those logs ? if so, maybe make it a constant and document it, instead of directly using a string in the code which is not very easy to find?

floriand-e2r: Indeed, we could define the value and taks keys as constants in class files (or in resource files?). But to keep the modularity of powsybl, the keys for one process should be in the corresponding artifact only. There won't be a unique file, unfortunately, which makes it a bit less interesting.

About documenting all the keys I think it would be a bit heavy, and redundant with the default message, don't you think? Maybe add the default message next to the keys constant? but then reading the report code would be unclear.

commons/src/main/java/com/powsybl/commons/reporter/LoggerReporter.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public LoggerReporter(String rootTaskKey, String rootDefaultName, Map<String, Object> taskValues) {
+        this.taskKey = Objects.requireNonNull(rootTaskKey);
+        this.defaultName = rootDefaultName;
+        this.taskValues = new HashMap<>(Objects.requireNonNull(taskValues));
+    }
+
+    @Override
+    public LoggerReporter createChild(String taskKey, String defaultName, Map<String, Object> values) {
+        getDefaultLogConsumer().accept(toString());
```
jonenst, line 51: is it ok to have a side effect here? What if you want to create a child, but not have the Logger.info log right now ? Also what if you want a different level than INFO for this log ?

floriand-e2r: Good catch, `getDefaultLogConsumer()` is not appropriate here. I'll add a method returning a `Consumer<String>` which could be overriden, if you're ok with that. Looking for severity in task values would be consistent with the reports but severity of a task doesn't really make sense.

floriand-e2r: Or just deleting that logging could be a solution in fact; not very useful information.

iidm/iidm-api/src/main/javadoc/com/powsybl/iidm/network/doc-files/currentLimits.svg
```diff
@@ -1,68 +1,16 @@
...
    xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
    xmlns:svg="http://www.w3.org/2000/svg"
    xmlns="http://www.w3.org/2000/svg"
-   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
-   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
-   width="400"
-   height="400"
-   id="svg2"
    version="1.1"
-   inkscape:version="0.91 r13725"
```
jonenst, line 25: why these changes ? Is the svg broken ? on githubpreview and eyeofgnome the svg is empty. On inkscape, you see that the text is outside of the svg
Capture d’écran de 2021-03-29 11-02-30

commons/src/main/java/com/powsybl/commons/reporter/ReportSeeker.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+package com.powsybl.commons.reporter;
+
+import java.util.Collection;
+import java.util.Map;
+
+/**
+ * @author Florian Dupuy <florian.dupuy at rte-france.com>
+ */
+public interface ReportSeeker {
```
jonenst, line 15: I don't really understand what this class is from its name "ReportSeeker".. Maybe find a better name ?

iidm/iidm-api/src/main/javadoc/com/powsybl/iidm/network/doc-files/currentLimits.svg
floriand-e2r: woops that's from my stash, nothing to do here

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
```diff
@@ -101,7 +103,7 @@ private static EntsoeGeographicalCode getRegionalGeographicalCode(Substation sub
         return res == EntsoeGeographicalCode.DE ? null : res;
     }
 
-    private static void createBuses(UcteNetworkExt ucteNetwork, UcteVoltageLevel ucteVoltageLevel, VoltageLevel voltageLevel) {
+    private static void createBuses(UcteNetworkExt ucteNetwork, UcteVoltageLevel ucteVoltageLevel, VoltageLevel voltageLevel, Reporter reporter) {
```
jonenst, line 106: Is an importer supposed to be thread-safe ? If not, would it be better to store the reporter in a field of the importer object so we don't have to pass it around explicitly everywhere ?

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
```diff
@@ -844,7 +896,7 @@ public boolean exists(ReadOnlyDataSource dataSource) {
             String ext = findExtension(dataSource, false);
             if (ext != null) {
                 try (BufferedReader reader = new BufferedReader(new InputStreamReader(dataSource.newInputStream(null, ext)))) {
-                    return new UcteReader().checkHeader(reader);
+                    return new UcteReader(Reporter.NO_OP).checkHeader(reader);
```
jonenst, line 899: Why always noop here ?

commons/src/main/java/com/powsybl/commons/reporter/LoggerTreeReporter.java
```diff
@@ -0,0 +1,97 @@
...
+    public void export(Writer writer) {
+        try {
+            printTaskReport(this, writer, "");
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
+
+    private void printTaskReport(ReportSeeker reportSeeker, Writer writer, String prefix) throws IOException {
+        writer.append(prefix).append("+ ").append(reportSeeker.toString()).append('\n');
```
jonenst, line 92: Why always '\n' ? printing this on windows will be ugly ? in AsciiTableFormatter.java we use System.lineSeparator()

floriand-e2r: corrected

commons/src/main/java/com/powsybl/commons/reporter/LoggerTreeReporterDeserializer.java
```diff
@@ -0,0 +1,148 @@
...
+
+                case "reports":
+                    parser.nextToken();
+                    reports = parser.readValueAs(new TypeReference<ArrayList<Report>>() {
+                    });
+                    break;
+
+                case "childReporters":
+                    parser.nextToken();
+                    childReporters = parser.readValueAs(new TypeReference<ArrayList<LoggerTreeReporter>>() {
```
jonenst: here you create children of type LoggerTreeReporter but they could have been any ReportSeeker before the export

commons/src/main/java/com/powsybl/commons/reporter/LoggerReporter.java
```diff
@@ -56,8 +56,8 @@ public LoggerReporter createChild(String taskKey, String defaultName, Map<String
     @Override
     public void addTaskValue(String key, Object value) {
         Objects.requireNonNull(value);
-        if (!value.getClass().isPrimitive()) {
-            throw new PowsyblException("Logger reporter expects only primitive values (value is an instance of " + value.getClass() + ")");
+        if (!(value instanceof Float || value instanceof Double || value instanceof Integer || value instanceof Long || value instanceof String)) {
+            throw new PowsyblException("Logger reporter expects only primitive or String values (value is an instance of " + value.getClass() + ")");
```
jonenst, line 60: what about byte, short, boolean, char ? Why this restriction instead of relying on Object.toString() ?

commons/src/main/java/com/powsybl/commons/reporter/Report.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * @author Florian Dupuy <florian.dupuy at rte-france.com>
+ */
+public class Report {
+
+    public static final String REPORT_SEVERITY_KEY = "reportSeverity";
```
jonenst, line 21: why is this in the values instead of a normal field ?

floriand-e2r: Severity might not be of interest for all consumers, hence using the values allows to keep the API generic.

ucte/ucte-network/src/main/java/com/powsybl/ucte/network/io/UcteReader.java
jonenst: I see what you mean, but when you are writing code to parse these reports, I think you want a definitive spec of what can be in the report, not just looking at one report file (which might not represent all possible reports) to guess things no ?

When I said document, I didn't mean to describe each and every key in the documentation, but at least list them. But you are right that the default message is a very good documentation, so having them together is a good idea. I think the report good would still be clear. Not sure what's the best way to group them together though, constants with similar names ? Is an enum better ? something else ?

I don't think we need a view of all possible report keys. We you want to parse an ucteimport report, you go and read the ucte documentation for its keys and you don't care about others ?
But if we wanted, we could do something with serviceloader to discover all the things that that write reports in the current jvm and ask them to give the list of keys, but not sure if it's useful.

commons/src/main/java/com/powsybl/commons/reporter/Reporter.java
```diff
@@ -0,0 +1,51 @@
...
+
+    Reporter createChild(String taskKey, String defaultName, Map<String, Object> values);
+
+    Reporter createChild(String taskKey, String defaultName);
+
+    Reporter createChild(String taskKey, String defaultName, String key, Object value);
+
+    void addTaskValue(String key, Object value);
+
+    void report(String reportKey, String defaultMessage, Map<String, Object> values);
```
jonenst, line 26: Do we want a system where the code doesn't create the report data if the severity is below a threshold in the context ? In slf4j they have info(data) and info(dataConsummer) for this

commons/src/main/java/com/powsybl/commons/reporter/LoggerReporter.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.function.Consumer;
+
+/**
+ * @author Florian Dupuy <florian.dupuy at rte-france.com>
+ */
+public class LoggerReporter extends AbstractReporter {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(LoggerReporter.class);
```
jonenst, line 25: do we want all these in the same category ?

jonenst:
Should we take this oportunity and add a more generic "Context" object, which would hold the Reporter, and by passed to all the code (either through a field in an not thread-safe object, or through the method parameters). For now the context would only have the reporter but it would make future modifications like this a lot easier.

commons/src/main/java/com/powsybl/commons/reporter/Report.java
jonenst: What about a default severity of "ALL" ? This way consummers who don't care about severity get everything and can ignore this part of the API, and others get what they want ?

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
```diff
@@ -110,6 +112,12 @@ private static void createBuses(UcteNetworkExt ucteNetwork, UcteVoltageLevel uct
                 continue;
             }
 
+            reporter.report(Report.builder()
+                .withKey("createBus")
+                .withDefaultMessage("Create bus ${bus}")
+                .withValue("bus", ucteNodeCode.toString())
+                .withSeverity(Report.SEVERITY_TRACE)
+                .build());
```
jonenst, line 120: I would have removed the equivalent Logger.trace and configured the LoggerReport (with allowing the dynamically specify the category) by default. Also allow to pass multiple reporter (or give an api to create a kind of DispatchingReporter..). Did you plan to do this in the future ?

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
geofjamg: It is supposed to be thread safe

geofjamg: And it is also need as UcteImporter is @Autoservice

commons/src/main/java/com/powsybl/commons/reporter/LoggerReporter.java
floriand-e2r: Yeah I liked relying on Object.toString(), but then I needed json serialization/deserialization.

commons/src/main/java/com/powsybl/commons/reporter/LoggerReporter.java
sylvlecl: Maybe the "task key" should be used as the logger name ?
Or LoggerReporter.class + task key, so that we can still filter out all log from those "functional" logs in logback config ?

sylvlecl: PS: this must be mentioned in the implementation's javadoc

commons/src/main/java/com/powsybl/commons/reporter/TreeReporter.java
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * @author Florian Dupuy <florian.dupuy at rte-france.com>
+ */
+public class TreeReporter extends AbstractReporter {
+
+    private final List<TreeReporter> childReporters = new ArrayList<>();
+    private final List<Report> reports = new ArrayList<>();
+
+    public TreeReporter() {
+        super();
```
geofjamg, line 30: Useless?

iidm/iidm-api/src/main/javadoc/com/powsybl/iidm/network/doc-files/currentLimits.svg
```diff
@@ -11,93 +11,93 @@ file, You can obtain one at http://mozilla.org/MPL/2.0/.
 <!-- Created with Inkscape (http://www.inkscape.org/) -->
```
geofjamg, line 11: You should revert this change

ucte/ucte-converter/src/test/resources/frVoltageRegulatingXnodeReport.json
```diff
@@ -0,0 +1,243 @@
...
+          "reports" : [ {
+            "reportKey" : "negativeLineResistance",
+            "values" : {
+              "reportSeverity" : {
+                "value" : "UCTE_ERROR",
+                "type" : "ERROR_LOGLEVEL"
+              },
+              "lineId" : {
+                "value" : "FFFFFF11 XXXXXX11 1",
+                "type" : "UNTYPED"
```
geofjamg, line 22: Maybe we should just not write the type in this case?

ucte/ucte-network/src/main/java/com/powsybl/ucte/network/ext/UcteReportConstants.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+package com.powsybl.ucte.network.ext;
+
+import com.powsybl.commons.reporter.TypedValue;
+
+/**
+ * @author Florian Dupuy <florian.dupuy at rte-france.com>
+ */
+public class UcteReportConstants {
+    public static final TypedValue TRACE_SEVERITY = new TypedValue("UCTE_TRACE", TypedValue.TRACE_LOGLEVEL);
```
geofjamg, line 15: private constructor + final?

iidm/iidm-api/src/main/javadoc/com/powsybl/iidm/network/doc-files/currentLimits.svg
floriand-e2r: Ouch, I already reverted it, but got mixed up apparently!

ucte/ucte-converter/src/test/resources/frVoltageRegulatingXnodeReport.json
floriand-e2r: Yes, indeed, I wanted to still add a Serializer/Deserializer for TypedValue to that end. I thought to either not write the type or directly write the value in fact.

commons/src/main/java/com/powsybl/commons/reporter/Reporter.java
```diff
@@ -0,0 +1,51 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.commons.reporter;
+
+import java.util.Map;
+
+/**
+ * @author Florian Dupuy <florian.dupuy at rte-france.com>
+ */
+public interface Reporter {
```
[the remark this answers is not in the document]
floriand-e2r: Yes, sure, it should be detailed! I'll try to include that in current PR, but, even though it would help to review, maybe that could be done in a second one?

commons/src/main/java/com/powsybl/commons/reporter/LoggerTreeReporter.java
```diff
@@ -0,0 +1,151 @@
...
+import java.io.Writer;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.*;
+
+/**
+ * @author Florian Dupuy <florian.dupuy at rte-france.com>
+ */
+public class LoggerTreeReporter extends LoggerReporter implements ReportSeeker {
```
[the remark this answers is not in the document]
floriand-e2r: Indeed. I separated these two behaviours in a first commit, and finally removed the logger implementation in a later commit, as for now there's no identified use of such an implementation. And we tried in this POC to keep only what is for sure useful (at the cost of adding features/implementations later).

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
floriand-e2r: I didn't plan it so far, as no usecase is identified, but passing multiple reporters could be done in the future, through a compound implementation as @sylvlecl suggested.

commons/src/main/java/com/powsybl/commons/reporter/LoggerReporter.java
floriand-e2r: I added boolean on changed the exception message to be more accurate.

commons/src/main/java/com/powsybl/commons/reporter/TreeReporterSerializer.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * @author Florian Dupuy <florian.dupuy at rte-france.com>
+ */
+public class TreeReporterSerializer extends StdSerializer<TreeReporter> {
+
+    private static final String VERSION = "1.0";
+    private final Map<String, String> dictionary;
```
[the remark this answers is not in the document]
floriand-e2r: Ok, thanks for explaining! Additional context is really needed here, so I'll see how to pass it through `DeserializationContext`.

commons/src/main/java/com/powsybl/commons/reporter/Reporter.java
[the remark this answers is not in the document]
MioRtia: Lack of an extensive javadoc for a beta feature is not a blocking issue for me as it might be changed later anyway but it should indeed be documented somewhere (@annetill Should we have a "beta feature" section on the website?)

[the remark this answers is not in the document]
sylvlecl: Yes, it was not clear to me what was the status of that feature. But let's not forget to document it when the main concepts are validated. As a user, I would not have been comfortable with using it in the current state !

commons/src/main/java/com/powsybl/commons/reporter/TreeReporter.java
floriand-e2r: Removed, forcing the user to provide at least key and name

commons/src/main/java/com/powsybl/commons/reporter/TreeReporterSerializer.java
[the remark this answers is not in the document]
floriand-e2r: The serializers/deserializers are now stateless. And in the end I managed to avoid that need of additional context, by directly calling methods recursively when serializing/deserializing instead of calling the serializer/deserializer itself.

iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/import_/Importers.java
```diff
@@ -440,6 +483,26 @@ public static Network loadNetwork(Path file, ComputationManager computationManag
...
+     *
+     * @param file               The file to be loaded.
+     * @param computationManager A computation manager which may be used by import post-processors
+     * @param config             The import config, in particular definition of post processors
+     * @param parameters         Import-specific parameters
+     * @param loader             Provides the list of available importers and post-processors
+     * @param reporter           The reporter used for functional logs
+     * @return                   The loaded network
+     */
+    public static Network loadNetwork(Path file, ComputationManager computationManager, ImportConfig config, Properties parameters, ImportersLoader loader, Reporter reporter) {
```
geofjamg, line 497: Is it possible to share code with the one without Reporter?

iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/import_/Importers.java
```diff
@@ -500,6 +563,28 @@ public static Network loadNetwork(String filename, InputStream data, Computation
...
+     * @param filename           The name of the file to be imported.
+     * @param data               The raw data from which the network should be loaded
+     * @param computationManager A computation manager which may be used by import post-processors
+     * @param config             The import config, in particular definition of post processors
+     * @param parameters         Import-specific parameters
+     * @param loader             Provides the list of available importers and post-processors
+     * @param reporter           The reporter used for functional logs
+     * @return                   The loaded network
+     */
+    public static Network loadNetwork(String filename, InputStream data, ComputationManager computationManager, ImportConfig config, Properties parameters, ImportersLoader loader, Reporter reporter) {
```
geofjamg, line 578: Same

commons/src/main/java/com/powsybl/commons/reporter/TreeReporterSerializer.java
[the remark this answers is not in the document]
geofjamg: Yes very good idea, we will need the jackson module too.

iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/import_/Importers.java
floriand-e2r: In fact, the old `Importer` API should be called when no reporter is given, as the new API is not yet implemented by all subclasses of `Importer.java`. I didn't put implementation in all the subclasses, there are in many other modules. Indeed this is still a POC so I didn't want to affect too many artifacts. Do you agree on that?

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
```diff
@@ -259,8 +285,15 @@ private static void createGenerator(UcteNode ucteNode, VoltageLevel voltageLevel
...
     private static void createDanglingLine(UcteLine ucteLine, boolean connected,
                                            UcteNode xnode, UcteNodeCode nodeCode, UcteVoltageLevel ucteVoltageLevel,
-                                           Network network) {
-
+                                           Network network, Reporter reporter) {
+
+        reporter.report(Report.builder()
+            .withKey("danglingLineCreation")
+            .withDefaultMessage("Create dangling line '${ucteLine}' (Xnode='${xnodeCode}')")
+            .withValue("ucteLine", ucteLine.getId().toString())
```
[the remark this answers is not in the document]
floriand-e2r: Yes, indeed, but I tried to keep some untyped value, to have both typed and untyped values in reports

commons/src/main/java/com/powsybl/commons/reporter/TreeReporterSerializer.java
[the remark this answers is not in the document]
floriand-e2r: Done!

geofjamg (approved):

sylvlecl (approved):
Some notes for further work, about in memory reporter :
 * would it be better to separate the resulting report model from the reporter implementation ? (it would make sense to obtain an "unmodifiable" report at the end)
 * should "reports" and "sub reporters" be ordered ?

PR #1668, opened 2021-03-25
PSS/E Add new dataBlocks

marqueslanauja:

Feature

The following datablock have been added:
- TwoTerminalDcTransmissionLineData
- VoltageSourceConverterDcTransmissionLineData
- MultiTerminalDcTransmissionLineData
- GneDeviceData
- InductionMachineData

**Other information**:


marqueslanauja:
Sonar warnings about duplicated code refer to fields  defined in more than one dataBlock. I think it is preferable to keep the current code, even it is marked as duplicated than create additional psse_records grouping non-related fields to avoid the duplicated code.

geofjamg (approved):

annetill:
@marqueslanauja Can you do something for code duplication or we leave it like this?

marqueslanauja:
> @marqueslanauja Can you do something for code duplication or we leave it like this?

We think that is better to leave the duplicated code. The duplicated code refer to setters and getters associated with fields defined in more than one psse dataBlock. We think it is preferable to keep the current code, even it is marked as duplicated than create additional external classes, grouping non-related fields of the psse records.

annetill (approved):

PR #1677, opened 2021-04-06
Add extension on Boundary Points specifying if they model DC Points and the EIC code of the line they are associated with if it exists

MioRtia:

No

Feature: import HVDC points and EIC codes of boundaries

Yes, as it breaks the CGMES import if there is no implementation of CgmesDanglingLineBoundaryNode and CgmesLineBoundaryNode

**Other information**:


zamarrenolm (changes requested):
I will propose a modification of the query for boundary nodes

zamarrenolm (approved):
I've simplified the query for boundary nodes and adjusted the rest of the code. Boundaries contain pairs of connectivity node, topological node. Node/breaker IGMs will use connectivity nodes, while bus/branch IGMs will use topological nodes.

Issue #1171, opened 2020-02-21
powsybl-cgmes-conversion test fails


redispatch:
powsybl-cgmes-conversion test fails.
See screenshot.

cloned 21.02.2020 20:00 GMT

![cgmes conversion test error](https://user-images.githubusercontent.com/61331406/75068407-6b293100-54ef-11ea-863e-8ad513a6480f.png)

* **Do you want to request a *feature* or report a *bug*?**
bug

mathbagu:
Could you give us more details about your environment (OS version, Java version, Maven version...).

Looking to your screenshot, the tests don’t fail, but the JVM crashes. Could you send us the dump file if it exists? Is this issue reproductible if you rerun the build?

redispatch:
It was on windows. I tried it on two different PCs. 
Maven was 3.6.3. 
JDK 13.0.2 on one PC and JDK 8 on the other PC. 
I started it directly from the command line with „mvn package“ without an IDE. 
When build with IntelliJ the error did not occur. 
The issue is reproducible with different Java versions on different PCs. It occurs again when the build is rerun.

redispatch:
OS was Win 10. 

Maybe file names get to long for windows?
I first also could not clone from git because some file pathes exceeded 260 chars. Then I changed to a shorter directory. 
I think the long path was some cgmes file. 
Just a blind guess.

mathbagu:
Hi @redispatch 
We are able to reproduce the bug. We will try to find a solution/workaround and will give you a feedback soon.

mathbagu:
The issue is linked to a timeout in maven-surefire-plugin that kills the forked JVM, due to a too long execution time. To solve this issue, we can:
- change the root level in [logback-test.xml](https://github.com/powsybl/powsybl-core/blob/master/cgmes/cgmes-conversion/src/test/resources/logback-test.xml)
- set the `forkedProcessExitTimeoutInSeconds` property to a higher value

@zamarrenolm: is it ok for you if I change the log level? Do we really need this amount of information.
@redispatch: as a workaround, you can also skip the tests

mathbagu:
The log level has been changed in #1266

PR #1680, opened 2021-04-08
CGMES import only using EQ file - First try

marcosmc:

Feature

Importing a CGMES model requires the EQ, SV, SSH and/or TP files

Import a CGMES model using only an EQ file


annetill:
I am not sure, it is all default values that we need to import an EQ file without SSH and SV files?

annetill:
@marcosmc is this PR updated with your local work because we would like to make some tests and start reviewing the functionality.

zamarrenolm:
> @marcosmc is this PR updated with your local work because we would like to make some tests and start reviewing the functionality. Thanks!

Yes, the PR is updated with all the latest changes and can be reviewed. 

The test fails because for testing this branch we depend on HVDC import working for n/b models (#1717). We planned to remove the "draft" status after #1171 was merged.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/hvdc/CgmesDcConversion.java
```diff
@@ -256,6 +256,8 @@ private static HvdcType decodeType(String stype) {
             } else if (cconverter1.asDouble(TARGET_PPCC) == 0 && cconverter2.asDouble(TARGET_PPCC) == 0) {
                 // Both ends are rectifier or inverter
                 return HvdcLine.ConvertersMode.SIDE_1_INVERTER_SIDE_2_RECTIFIER;
+            } else {
+                return HvdcLine.ConvertersMode.SIDE_1_INVERTER_SIDE_2_RECTIFIER;
```
MioRtia, line 260: I would log a message here indicating we don't know so we assumed

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesTerminal.java
```diff
@@ -42,7 +42,7 @@ public CgmesTerminal(PropertyBag t) {
         // two relationships may be different
         // (TODO(Luma): missing reference here)
 
-        this.connected = t.asBoolean("connected", false);
+        this.connected = t.asBoolean("connected", true);
```
MioRtia, line 45: I'm not sure about that... What do you think @annetill? Should all terminals be considered connected or disconnected? Or should it be configurable?

MioRtia (changes requested):

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesTerminal.java
marcosmc: This parameter is obtained from SSH, so if we leave the previous default value, when a model that only has EQ is imported, the entire model will be disconnected.

MioRtia: Yes, I understood that and indeed, you may be right, it might be more "logical" to consider all of them connected. Just not sure it is always what we want, depending on the use case and this is why I asked Anne.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -252,7 +252,7 @@ private static void createControlArea(CgmesControlAreas cgmesControlAreas, Prope
                 .setId(controlAreaId)
                 .setName(ca.getLocal("name"))
                 .setEnergyIdentificationCodeEic(ca.getLocal("energyIdentCodeEic"))
-                .setNetInterchange(ca.asDouble("netInterchange"))
```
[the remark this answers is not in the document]
marcosmc: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/EnergyConsumerConversion.java
```diff
@@ -39,14 +39,6 @@ public void convert() {
...
     }
 
-    private double p0() {
-        return powerFlow().defined() ? powerFlow().p() : p.asDouble("pFixed", Double.NaN);
-    }
-
-    private double q0() {
-        return powerFlow().defined() ? powerFlow().q() : p.asDouble("qFixed", Double.NaN);
-    }
-
```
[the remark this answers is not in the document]
marcosmc: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/AbstractConductingEquipmentConversion.java
```diff
@@ -662,6 +662,14 @@ protected BoundaryLine createBoundaryLine(String boundaryNode) {
...
     }
 
+    protected double p0() {
+        return powerFlow().defined() ? powerFlow().p() : 0.0;
+    }
+
+    protected double q0() {
+        return powerFlow().defined() ? powerFlow().q() : 0.0;
+    }
+
```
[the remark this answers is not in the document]
marcosmc: The protected methods are overridden in the Energy Consumer conversion to recover the use of pFixed and qFixed.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/hvdc/CgmesDcConversion.java
marcosmc: Done

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesTerminal.java
MioRtia: I discussed with Anne and for now, we think that as you coded, the default value should be to true. However (maybe in another PR), there should be an option to create a switch (instead of internal connections) in these case so users can disconnect manually some terminals afterwards.
I am still marking this conversation as resolved as it will not be coded in this PR.

MioRtia (approved):
It is okay for me, I will do some tests and then merge it.

annetill (commented):
Thanks for this work! I have some questions:
- This PR does not convert all the configurations we can met in the network: we have a way to find terminal from TN and it seems that we still have a dependency on the TP file ; without that, we will loose a lot of regulating terminals.
- What about importing regulating controls without target value and enabled status ? We have to find a rule for that, as the steady state hypothesis will arrived later in the state workflow.

MioRtia:
I made some testing.
You seem to have miss quite a lot about RegulatingControl: the information present in the EQ file is not imported if there is no SSH file.

Furthermore, we might want to find a solution to find equivalent terminals without looking at Topological Nodes (@zamarrenolm do you think it would be possible to do it from Connectivity Nodes in Node Breaker? It might be mandatory anyway for CGMES 3.0)

@annetill @zamarrenolm I don't know if you want to do these corrections in this PR or in another, those are supplementary features but do not make this PR wrong.

MioRtia:
> > I made some testing.
> > You seem to have miss quite a lot about RegulatingControl: the information present in the EQ file is not imported if there is no SSH file.
> > Furthermore, we might want to find a solution to find equivalent terminals without looking at Topological Nodes (@zamarrenolm do you think it would be possible to do it from Connectivity Nodes in Node Breaker? It might be mandatory anyway for CGMES 3.0)
> > @annetill @zamarrenolm I don't know if you want to do these corrections in this PR or in another, those are supplementary features but do not make this PR wrong.
> 
> First step has been to make optional all the information coming from SSH in the SPARQL queries.
> 
> We will review the conversion of the regulating controls to find why some are lost when providing default values.
> 
> Yes we agree in that we will have to define how to proceed when we do not have target values ... ideally, following the same set of rules to all kind of controls: providing default target values based on nominal if it makes sense (voltage), disabling the regulating control if there is no reasonable default (regulating active/reactive power) ...

I will open issues about what we discussed to keep track of what remains to be done and then I will merge this PR.

PR #1682, opened 2021-04-13
Consider transformers at boundaries during conversion

marqueslanauja:
Signed-off-by: José Antonio Marqués <marquesja@aia.es>

**Please check if the PR fulfills these requirements** *(please use `'[x]'` to check the checkboxes, or submit the PR and then click the checkboxes)*
- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**Does this PR already have an issue describing the problem ?** *If so, link to this issue using `'#XXX'` and skip the rest*

**What kind of change does this PR introduce?** *(Bug fix, feature, docs update, ...)*
Feature

**What is the current behavior?** *(You can also link to an open issue here)*
Transformers at boundaries are not considered

**What is the new behavior (if this is a feature change)?**

Transformers at boundaries are considered during conversion, but ratio and angle are ignored. Transformers are  mapped to a regular dangling line

**Does this PR introduce a breaking change or deprecate an API?** *If yes, check the following:*
- [ ] The *Breaking Change* or *Deprecated* label has been added
- [ ] The migration guide has been updated in the github wiki *(What changes might users need to make in their application due to this PR?)*

**Other information**:

(if any of the questions/checkboxes don't apply, please delete them entirely)

PR #1685, opened 2021-04-14
CGMES: add support for Equivalent Branches at boundaries

zamarrenolm:

Feature

Supported objects at boundaries are only ACLineSegment and Switches. 

Allow Equivalent Branches with one end at boundary. They will be mapped to an IIDM dangling line if an individual grid model is being loaded.

**Other information**:
Observed in some real-world models.

### DIFF
--- a/cgmes/cgmes-conformity/src/test/java/com/powsybl/cgmes/conformity/test/CgmesConformity1ModifiedCatalog.java
+++ b/cgmes/cgmes-conformity/src/test/java/com/powsybl/cgmes/conformity/test/CgmesConformity1ModifiedCatalog.java
@@ -410,6 +410,25 @@ public final class CgmesConformity1ModifiedCatalog {
                         "MicroGridTestConfiguration_TP_BD.xml"));
     }
 
+    public static TestGridModelResources microGridBaseCaseBEEquivalentBranchAtBoundary() {
+        String base = ENTSOE_CONFORMITY_1
+            + "/MicroGrid/BaseCase/CGMES_v2.4.15_MicroGridTestConfiguration_BC_BE_v2/";
+        String baseModified = ENTSOE_CONFORMITY_1_MODIFIED
+            + "/MicroGrid/BaseCase/BC_BE_v2_eqbranch_at_boundary/";
+        String baseBoundary = ENTSOE_CONFORMITY_1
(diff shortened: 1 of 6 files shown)


MioRtia (commented):
I agree that it works for EquivalentBranch at Boundary, but does it work when we import a CGM with EquivalentBranch-ACLineSegment junction? It seems that for that, the `delayedNodes` list should be used... Or will we handle this in a later PR? What do you think @annetill?

zamarrenolm:
> I agree that it works for EquivalentBranch at Boundary, but does it work when we import a CGM with EquivalentBranch-ACLineSegment junction? It seems that for that, the `delayedNodes` list should be used... Or will we handle this in a later PR? What do you think @annetill?

It will work if we create a CGM from multiple IGMs imported separately. Each IGM will create their dangling lines and the `MergingView` or the destructive merge will combine them in a `MergedLine` / `TieLine`.

The only scenario where this solution is not complete is when we import multiple EQ files (from many IGMs) in a single conversion (to build a single Network). As an example: the "Assembled" MicroGrid files of the ENTSO-E conformity tests, where BE and NL files are imported together. We will cover this configuration in the next PR about transformers at boundaries, where we have refactored a little bit how the different combinations of equipment at a boundary are handled for assembled imports.

marqueslanauja (approved):

zamarrenolm:
Handling of equipment that can be found at boundaries will be completed in #1682, for IGM, CGM and assembled conversions.

MioRtia:
> > I agree that it works for EquivalentBranch at Boundary, but does it work when we import a CGM with EquivalentBranch-ACLineSegment junction? It seems that for that, the `delayedNodes` list should be used... Or will we handle this in a later PR? What do you think @annetill?
> 
> It will work if we create a CGM from multiple IGMs imported separately. Each IGM will create their dangling lines and the `MergingView` or the destructive merge will combine them in a `MergedLine` / `TieLine`.
> 
> The only scenario where this solution is not complete is when we import multiple EQ files (from many IGMs) in a single conversion (to build a single Network). As an example: the "Assembled" MicroGrid files of the ENTSO-E conformity tests, where BE and NL files are imported together. We will cover this configuration in the next PR about transformers at boundaries, where we have refactored a little bit how the different combinations of equipment at a boundary are handled for assembled imports.

Okay, thank you for your answer! I agree with that and this PR is okay for me otherwise.

MioRtia (approved):

PR #1687, opened 2021-04-15
Add a tool to generate a bash autocompletion script for itools

sylvlecl:

Feature.

`itools` command proposes a subcommand to generate a bash autocompletion script.


sylvlecl:
Notes:
 * maybe we should use something else than velocity which introduces some dependencies and is very old: maybe freemarker ?
 * move it to a separate module or not ?

geofjamg:
> Notes:
> 
>     * maybe we should use something else than velocity which introduces some dependencies and is very old: maybe freemarker ?
> 
>     * move it to a separate module or not ?

You are right, last velocity release is 2010, it smells bad... Freemarker seems to be very active and a better choice.
I would keep code in "tools" module, it makes sens to me to have this code here.

tools/src/main/java/com/powsybl/tools/autocompletion/commonscli/CommonsCliAdapter.java
```diff
@@ -0,0 +1,51 @@
...
+ * Adapter for apache commons CLI options
+ *
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at gmail.com>
+ */
+public final class CommonsCliAdapter {
+
+    private CommonsCliAdapter() {
+    }
+
+    public static List<BashOption> adaptOptions(Options commonsCliOptions) {
```
geofjamg, line 28: As we do not intend to support multiples cli lib, maybe this class is not needed and we can directly move theses utilities Options to BashOption construction to BashOption it self?

geofjamg (approved):

PR #1691, opened 2021-04-17
Fix HalfLine creation and its Boundary object

mathbagu:

N/A

Refactoring

The implementation of the TieLineAdder and the HalfLineBoundary is not really good

This PR propose an alternative implementation that seems better to me.

**Other information**:


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/TieLineAdderImpl.java
```diff
@@ -124,6 +124,11 @@ public TieLineAdderImpl add() {
             return TieLineAdderImpl.this;
         }
 
+        public TieLineImpl.HalfLineImpl build() {
```
MioRtia, line 127: Can this be package-private?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/TieLineImpl.java
```diff
@@ -7,30 +7,47 @@
...
-        boolean fictitious = false;
-        double r = Double.NaN;
-        double x = Double.NaN;
-        double g1 = Double.NaN;
-        double g2 = Double.NaN;
-        double b1 = Double.NaN;
-        double b2 = Double.NaN;
-        HalfLineBoundaryImpl boundary;
+
+        public HalfLineImpl(String id, String name, boolean fictitious, double r, double x, double g1, double b1, double g2, double b2, Branch.Side side) {
```
MioRtia, line 35: Can this be package-private?

MioRtia (approved):
Two small remarks but otherwise okay for me!

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/TieLineAdderImpl.java
mathbagu: I change the visibility to private: this method should be called only by the enclosing TieLineAdder instance

PR #1708, opened 2021-04-29
Caching the network ref in AbstractConnectable

### LINKED ISSUE #1705: IIDM slow update
bug

I noticed that after a loadflow calculation, IIDM network update is very slow (line 100ms for a 6000 buses network). As a general observation every IIDM value update seem to be slow.

To highlight the issue I just profiled an OpenLoadFlow load flow calculation and to even better hightlight the issue I changed a little but the code to update 100 times.
```diff
Index: src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java b/src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java
--- a/src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java	(revision f74a0d848d5daf245e05d24cdf90e3db655cf670)
+++ b/src/main/java/com/powsybl/openloadflow/OpenLoadFlowProvider.java	(date 1619593965640)
@@ -198,10 +198,12 @@
         for (AcLoadFlowResult result : results) {
             // update network state
             if (result.getNewtonRaphsonStatus() == NewtonRaphsonStatus.CONVERGED) {
-                result.getNetwork().updateState(!parameters.isNoGeneratorReactiveLimits(),
-                                                parameters.isWriteSlackBus(),
-                                                parameters.isPhaseShifterRegulationOn(),
-                                                parameters.isTransformerVoltageControlOn());
+                for (int i = 0; i < 100; i++) {
+                    result.getNetwork().updateState(!parameters.isNoGeneratorReactiveLimits(),
+                            parameters.isWriteSlackBus(),
+                            parameters.isPhaseShifterRegulationOn(),
+                            parameters.isTransformerVoltageControlOn());
+                }
             }
 
             LoadFlowResult.ComponentResult.Status status;
```

Profiling (in my case using JProfiler) is very clear:
image

We can see than, when we update the voltage magnitude of a bus (setV), almost 100% of the time is in notifyUpdate, even if there is no listener attached to the network.
 
There is 2 reasons for that:
   * AbstractConnectable::getNetwork is very slow
```java
   @Override
    public NetworkImpl getNetwork() {
        if (terminals.isEmpty()) {
            throw new PowsyblException(id + " is not attached to a network");
        }

        // During the removal of a multi terminals component (Line, 2WT or 3WT), terminals are detached from the voltage level
        return terminals.stream()
                        .map(TerminalExt::getVoltageLevel)
                        .filter(Objects::nonNull)
                        .map(VoltageLevelExt::getNetwork)
                        .findFirst()
                        .orElse(null);
    }
```
   * More generaly to get listener list we need to go back to Network by following element parent links until the network, which is slow too.  

There is no reason this kind of code to be like 100 times faster. We should redesign the way we get listener list from every network element. 

In the case of a load flow calculation, network update almost as time consuming as calculation itsef.

  - PowSyBl Version: ...
  - OS Version: ...

flo-dup:

Yes, fixes #1705

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractBranch.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractBranch.java
@@ -8,6 +8,7 @@ package com.powsybl.iidm.network.impl;
 
 import com.powsybl.commons.PowsyblException;
 import com.powsybl.iidm.network.*;
+import com.powsybl.iidm.network.impl.util.Ref;
 import com.powsybl.iidm.network.util.LimitViolationUtils;
 
 import java.util.Collection;
@@ -22,8 +23,8 @@ abstract class AbstractBranch<I extends Branch<I>> extends AbstractConnectable<I
 
     private final OperationalLimitsHolderImpl operationalLimitsHolder2;
 
-    AbstractBranch(String id, String name, boolean fictitious) {
-        super(id, name, fictitious);
+    AbstractBranch(Ref<NetworkImpl> network, String id, String name, boolean fictitious) {
+        super(network, id, name, fictitious);
         operationalLimitsHolder1 = new OperationalLimitsHolderImpl(this, "limits1");
         operationalLimitsHolder2 = new OperationalLimitsHolderImpl(this, "limits2");
     }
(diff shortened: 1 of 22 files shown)


MioRtia (approved):

geofjamg (commented):
I think we should go further and not only cache by directly construct conectables with network ref. This solution will be nice for benchmarking where we update in a loop, but first update is still  slow.

geofjamg (changes requested):
See previous comment

geofjamg:
This is like 2 times faster than with master, so really better but this is still surprisingly relatively slow. Any idea why?

flo-dup:
It seems that it is due do the high number of times the `setV()`, `setAngle()`, `setP()` and `setQ()` are called. I tried to remove completely the notifyUpdate for testing and there's only a 15-20% win on the whole update.

geofjamg (approved):

PR #1658, opened 2021-03-19
Add RegulationMode to Generator

tadam50:
Signed-off-by: Thomas ADAM <tadam@silicom.fr>

**Please check if the PR fulfills these requirements** *(please use `'[x]'` to check the checkboxes, or submit the PR and then click the checkboxes)*
- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**Does this PR already have an issue describing the problem ?** *If so, link to this issue using `'#XXX'` and skip the rest*
No

**What kind of change does this PR introduce?** *(Bug fix, feature, docs update, ...)*
Feature

**What is the current behavior?** *(You can also link to an open issue here)*
Generator regulation is only on voltage level, manage by boolean with isVoltageRegulatorOn & setVoltageRegulatorOn

**What is the new behavior (if this is a feature change)?**
Generator regulation is only on voltage level, manage by RegulationMode enum
Generator will be able to be regulated as StaticVarCompensator

**Does this PR introduce a breaking change or deprecate an API?** *If yes, check the following:*
- [ ] The *Breaking Change* or *Deprecated* label has been added
- [ ] The migration guide has been updated in the github wiki *(What changes might users need to make in their application due to this PR?)*

**Other information**:

(if any of the questions/checkboxes don't apply, please delete them entirely)

PR #1709, opened 2021-04-30
[WIP] Add regulation mode to generator

MioRtia:

No

**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**
Feature

**What is the current behavior? (You can also link to an open issue here)**
Generator regulation is only on voltage level, manage by boolean with isVoltageRegulatorOn & setVoltageRegulatorOn

Generator regulation is only on voltage level, manage by RegulationMode enum
Generator will be able to be regulated as StaticVarCompensator

**Other information**:
DO NOT SQUASH AND MERGE, this PR needs to have commits sorted


MioRtia:
cf #1658 
@annetill Can you confirm if with previous method, when isVoltageRegulatorOn was false, the regulation mode was actually REACTIVE_POWER rather than OFF? I think so but I'm not sure.

annetill:
I think that we have to keep `OFF` for local `targetQ`. What is not clear for me is: do we miss something ? Because in a simulator, if this regulation is not implemented, you want to ignore it and so ignore the remote `targetQ`, but you still want to be able to perform a calculation. For that, you need a local `targetQ` but we don't store it...

annetill:
We have decided to model this through an extension https://github.com/powsybl/powsybl-core/pull/1732. Sorry for that...

PR #1711, opened 2021-05-03
Fix JSONUtil::skip, skip arrays and don't do an infinite busy loop when skipping starts precisely on next END_OBJECT

jonenst:

NO

Bug Fix

calling JsonUtil.skip() if the next token is going to be END_OBJECT does a busy 100%cpu infinite loop
calling JsonUtil.skip() if the next token is going to be START_ARRAY doesn't skip the array (it advances to its first element)

calling JsonUtil.skip() if the next token is going to be END_OBJECT throws
calling JsonUtil.skip() if the next token is going to be START_ARRAY skips the array.

NO, it should always have been called for START_OBJECT anyway I think (otherwise it was utterly broken)


commons/src/test/java/com/powsybl/commons/json/JsonUtilTest.java
```diff
@@ -0,0 +1,70 @@
+package com.powsybl.commons.json;
```
mathbagu, line 1: Don't forget to add the license header

commons/src/test/java/com/powsybl/commons/json/JsonUtilTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Test
+    public void testSkipStringValue() throws IOException {
+        String strJson = "\"foo\"";
+        JsonParser parser = factory.createParser(strJson);
+        JsonUtil.skip(parser);
+        assertEquals("Should have skipped to VALUE_STRING", JsonToken.VALUE_STRING, parser.currentToken());
+        assertNull("Should return null because the whole json is skipped when on next STRING_VALUE", parser.nextToken());
+    }
+}
```
mathbagu, line 70: I'm not sure that the tests you added are exhaustive, for instance when an object contains sub objects.

mathbagu (dismissed):

commons/src/test/java/com/powsybl/commons/json/JsonUtilTest.java
jonenst: fixed thx

sylvlecl (approved):
As discussed: we could deprecate this method in favor of jackson's one.
However, it works, so OK to leave it this way.

commons/src/test/java/com/powsybl/commons/json/JsonUtilTest.java
jonenst: Given the current implementation, I think they are exhaustive enough? We don't want to redo all the tests for jackson, it would be very long I thinkg

PR #1717, opened 2021-05-04
HVDC. At node-breaker level use connectivityNodes to determine the Hvdc configurations

marqueslanauja:

Feature

TopologicalNodes are always used.

TopologicalNodes are only used in bus-branch models. In node-breaker models the connectivityNodes are used.
These changes allow to import a model using only the EQ file.

**Other information**:


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/hvdc/TPnodeEquipments.java
```diff
@@ -86,70 +87,70 @@ private void computeThreeWindingsTransformer(CgmesModel cgmesModel, Adjacency ad
...
+        nodes.add(CgmesDcConversion.getAcNode(cgmesModel, t1));
+        nodes.add(CgmesDcConversion.getAcNode(cgmesModel, t2));
+        nodes.add(CgmesDcConversion.getAcNode(cgmesModel, t3));
+        if (isValidTransformer(adjacency, nodes)) {
+            addTransformer(adjacency, id, nodes, EquipmentType.TRANSFORMER);
         }
     }
 
-    private void addEquipment(Adjacency adjacency, String id, String topologicalNodeId1, String topologicalNodeId2,
+    private void addEquipment(Adjacency adjacency, String id, String nodeId1, String nodeId2,
```
zamarrenolm, line 99: Since we are not relying always on TP, I would suggest to rename this class as `NodeEquipment` instead of `TPnodeEquipments`. Also, consider removing the final "s" from "equipments" (word is uncountable and should not be used in plural).

zamarrenolm: The inner static class `TPnodeEquipment`, that stores an identifier and type of a piece of equipment could be renamed `EquipmentReference` to avoid naming collision with the main class.

zamarrenolm (changes requested):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/hvdc/TPnodeEquipments.java
marqueslanauja: Done

marqueslanauja:
After renaming `tpNodeEquipments` by `nodeEquipment ` coverage has been decreased from over 80% to 74% as in the renamed class `NodeEquipment` several debug methods have been modified and there are not tests associated with them.

zamarrenolm:
coverage is ok for me

zamarrenolm (approved):

MioRtia (approved):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/hvdc/CgmesDcConversion.java
```diff
@@ -58,41 +57,82 @@ public void convert() {
...
+        } else {
+            convert(converters.get(0).acDcConvertersEnd2, converters.get(0).acDcConvertersEnd1, dcLineSegments.get(0));
+        }
+    }
+
+    // (2)
+    //         CGMES Configuration                                                              IIDM configuration
     //
-    // (2)  AcDcConverter1End1 ---                 --- AcDcConverter1End2
+    //      AcDcConverter1End1 ---                 --- AcDcConverter1End2
```
annetill, line 115: Thanks a lot for the comments, I think we should put these clear explanations in our web site documentation for CGMES DC lines conversion.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/hvdc/IslandEndHvdc.java
```diff
@@ -32,7 +29,7 @@
     // T2: two transformers, C2: two acDcConverters, LS2: two dcLineSegments
     // TN: n transformers (usually 2), CN: n acDcConverters (usually 2), LSN: n dcLineSegments (usually 2)
     enum HvdcEndType {
-        HVDC_T0_C1_LS1, HVDC_T1_C1_LS1, HVDC_TN_CN_LSN, HVDC_T1_C1_LS2, HVDC_T2_C2_LS1,
+        HVDC_T0_C1_LS1, HVDC_T0_C1_LS2, HVDC_T0_C2_LS1, HVDC_T1_C1_LS1, HVDC_TN_CN_LSN, HVDC_T1_C1_LS2, HVDC_T2_C2_LS1,
```
annetill, line 32: Through that PR, you have found to new end type that were not supported before, why? Or it is just to be more generic?

annetill (commented):
Have you make some performances and validity tests on real test cases? If not, I can make it.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/hvdc/IslandEndHvdc.java
marqueslanauja: Before the DC configurations were only analyzed at bus branch level. Now we analyze them at both levels, node breaker level and bus branch level. All the real IOPs with DC components are defined at the node breaker level and at this level some of the configurations are a bit different. We support the same configurations, but at both levels.

marqueslanauja:
> > > Have you make some performances and validity tests on real test cases? If not, I can make it.
> > 
> > 
> > I validated on EIRGRIDSONI, DKE and DKW, all the IOPs of 2021-March where we can find DC components.
> 
> And for performances?

> > > Have you make some performances and validity tests on real test cases? If not, I can make it.
> > 
> > 
> > I validated on EIRGRIDSONI, DKE and DKW, all the IOPs of 2021-March where we can find DC components.
> 
> And for performances?

The performance is the same as before. 

> > > Have you make some performances and validity tests on real test cases? If not, I can make it.
> > 
> > 
> > I validated on EIRGRIDSONI, DKE and DKW, all the IOPs of 2021-March where we can find DC components.
> 
> And for performances?

The same as before. This PR does not have impact on performance.

PR #1719, opened 2021-05-05
Add synchronous component num to ComponentResult 

obrix:
- [X ] The commit message follows our guidelines

Add synchronous component num to ComponentResult. This is useful to filter result of a load flow on specific synchronous component. Also rename componentNum to connectedComponent num. 

Only a generic componentNum attribute in ComponentResult.

Two componentNum in ComponentResult : connectedComponentNum and synchronousComponentNum.


loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowResult.java
```diff
@@ -33,7 +33,13 @@
          * Get component number.
```
annetill, line 33: Can you detailed the comment here?

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowResultImpl.java
```diff
@@ -28,12 +30,13 @@
 
         private final double slackBusActivePowerMismatch;
 
-        public ComponentResultImpl(int componentNum, Status status, int iterationCount, String slackBusId, double slackBusActivePowerMismatch) {
-            this.componentNum = checkComponentNum(componentNum);
+        public ComponentResultImpl(int connectedComponentNum, int synchronousComponentNum, Status status, int iterationCount, String slackBusId, double slackBusActivePowerMismatch) {
```
annetill, line 33: @flo-dup I am not always familiar with `breaking change` but I can imagine that it is one. So it means that we need to deal with decrease version compatibility?

annetill (changes requested):
I need @flo-dup for a more technical review, especially for compatibility.

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowResult.java
obrix: Oups did not catch that. Thanks.

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowResultImpl.java
obrix: Yes probably. Don't know the process here, so waiting for @flo-dup or @geofjamg insight.

geofjamg: You have increase the json version (1.2) and adapt the deserializer to support all versions. We can consider than component number in the actual code is synchronous component number in the new one.

flo-dup: we have to deal with version compatibility, but no need no deprecate constructor.

geofjamg (approved):

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/json/LoadFlowResultDeserializer.java
```diff
@@ -34,17 +34,25 @@
...
         while (parser.nextToken() != JsonToken.END_OBJECT) {
             switch (parser.getCurrentName()) {
+                case "connectedComponentNum":
+                    parser.nextToken();
+                    connectedComponentNum = parser.getValueAsInt();
+                    break;
+
+                case "synchronousComponentNum":
+
                 case "componentNum":
```
flo-dup, line 53: Doing so means that we can read a malformed json file containing a result with componentNum and connectedComponentNum instead of throwing an error. I think you should test here which version you're reading.

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/json/LoadFlowResultDeserializer.java
```diff
@@ -72,7 +80,10 @@
             }
         }
 
-        if (componentNum == null) {
+        if (connectedComponentNum == null) {
+            connectedComponentNum = 0;
+        }
```
flo-dup, line 85: That should be only for 1.1 version; for above versions I think we should raise an exception?

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/json/LoadFlowResultDeserializer.java
```diff
@@ -72,7 +80,10 @@
...
+        if (synchronousComponentNum == null) {
             throw new IllegalStateException("Component number field not found");
```
flo-dup, line 87: The exception message needs to be updated (for 1.2 version)

flo-dup (changes requested):

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/json/LoadFlowResultDeserializer.java
obrix: Yes. I just added some exceptions throwing when loading unexpected fields with a specific version.

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/json/LoadFlowResultDeserializer.java
```diff
@@ -49,8 +51,17 @@
                     break;
 
                 case "synchronousComponentNum":
+                    if (cVersion.compareTo(ModuleDescriptor.Version.parse("1.2")) < 0) {
```
flo-dup, line 54: You can use `JsonUtil::assertGreaterThanReferenceVersion` instead

flo-dup (changes requested):

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/json/LoadFlowResultDeserializer.java
obrix: I just added mising assertGreaterOrEqualThanReferenceVersion and assertLessThanReferenceVersion functions.

PR #1722, opened 2021-05-06
Add a bus cache to network implementation

sylvlecl:

See original issue on powsybl-open-loadflow https://github.com/powsybl/powsybl-open-loadflow/issues/296 .

Perf improvement (in most cases)

When using method `Bus getBus(String busId)` of the network bus and bus-breaker views, a linear search is performed through voltages levels.
This is very, very, inefficient as soon as you start using the method multiple times. There is nothing to discourage such use (such as javadoc), and no real reason to do so.

When searching for a bus by its ID, a cache (map) of all buses of the network is built.
It's a very basic approach : the cache is invalidated as soon as one voltage level cache is invalidated.
However, it should behave fine in most use cases. 
It could be refined in the future if needed (for example, only invalidate part of the cache when only some voltage levels are invalidated).

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
@@ -767,9 +767,11 @@ class NetworkImpl extends AbstractIdentifiable<Network> implements Network, Vari
 
         private final BusCache busViewCache = new BusCache(() -> getVoltageLevelStream().flatMap(vl -> vl.getBusView().getBusStream()));
 
-        private final BusCache busBreakerViewCache = new BusCache(() ->
-            getVoltageLevelStream().filter(vl -> vl.getTopologyKind() != TopologyKind.BUS_BREAKER)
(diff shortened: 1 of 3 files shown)


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
```diff
@@ -727,6 +765,12 @@ protected SynchronousComponentImpl createComponent(int num, int size) {
         private final SynchronousComponentsManager synchronousComponentsManager
                 = new SynchronousComponentsManager(NetworkImpl.this);
 
+        private final BusCache busViewCache = new BusCache(() -> getVoltageLevelStream().flatMap(vl -> vl.getBusView().getBusStream()));
+
+        private final BusCache busBreakerViewCache = new BusCache(() ->
+            getVoltageLevelStream().filter(vl -> vl.getTopologyKind() != TopologyKind.BUS_BREAKER)
```
flo-dup, line 771: The filtering is a bit confusing at first! Could you add a short comment here to say that you filter out the bus breaker ones because of `NetworkIndex`?

flo-dup (approved):

PR #1728, opened 2021-05-10
SV utility class. Review and Refactoring

marqueslanauja:

Bug fix

PhaseTapChangers are not considered.
Calculation of voltage and flow at one side by using the other side values is considered symmetric.
For Lines, default otherSide does not take into account the right values for shunt admittances to ground.

PhaseTapChangers are considered.
Calculation of voltage and flow at one side by using the other side values is not symmetric.
For Lines, default otherSide takes into account the right values for shunt admittances to ground.

**Other information**:


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
```diff
@@ -6,67 +6,57 @@
...
+        this.u = u;
+        this.a = a;
+        this.end = end;
     }
 
-    public static double getB(TwoWindingsTransformer twt) {
-        return twt.getB() * twt.getOptionalRatioTapChanger().map(rtc -> 1 + rtc.getCurrentStep().getB() / 100).orElse(1d);
+    /**
+     * @deprecated Not used anymore. The end associated to the voltage and flow must be defined. Use
+     * {@link SV#SV(double, double, double, double, int)}
```
zamarrenolm, line 44: If we use `Branch.Side` type to specify the end instead of an `int` we would enforce the fact that this utility class is designed for calculating state variables at ends of **branches**.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
```diff
@@ -6,67 +6,57 @@
...
-        this.p = p;
-        this.q = q;
-        this.u = u;
-        this.a = a;
-        s1 = new Complex(p, q); // s=p+jq
-        u1 = ComplexUtils.polar2Complex(u, Math.toRadians(a));
-        v1 = u1.divide(Math.sqrt(3f)); // v1=u1/sqrt(3)
-        i1 = s1.divide(v1.multiply(3)).conjugate(); // i1=conj(s1/(3*v1))
-    }
+    private final int end;
```
zamarrenolm, line 59: Use `Branch.Side`

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
```diff
@@ -84,132 +74,341 @@ public double getA() {
...
-        Complex s2 = v2.multiply(3).multiply(i2.conjugate()); // s2=3*v2*conj(i2)
-        return v2.multiply(Math.sqrt(3f));
+    public int getEnd() {
+        return end;
     }
 
-    private Complex computeU2(Complex y, Complex z, double rho) {
-        return computeU2(y, new Complex(0, 0), z, rho);
+    /**
+     * @deprecated Not used anymore. Use
```
zamarrenolm, line 82: Indicate that we deprecate this version with a _simplified_ view of the parameters of a generic branch to avoid misuse. I would enforce the explicit use of version that includes `rho` AND `alpha`.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
```diff
@@ -84,132 +74,341 @@ public double getA() {
...
 
-    private Complex computeS2(Complex y1, Complex y2, Complex z, double rho) {
-        Complex v1p = v1.multiply(rho); // v1p=v1*rho
-        Complex i1p = i1.divide(rho); // i1p=i1/rho
-
-        Complex i2p = i1p.subtract(y1.multiply(v1p)); // i2p=i1p-y1*v1p
-        Complex v2 = v1p.subtract(z.multiply(i2p)); // v2p=v1p-z*i2
-        Complex i2 = i2p.subtract(y2.multiply(v2)); // i2=i2p-y2*v2
-        return v2.multiply(3).multiply(i2.conjugate()); // s2=3*v2*conj(i2)
+    public SV otherSide(double r, double x, double g1, double b1, double g2, double b2, double rho) {
```
zamarrenolm, line 90: I would also deprecate this version. I prefer to enforce the use of the version that includes ALL the parameters, including `rho` AND `alpha`.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
```diff
@@ -84,132 +74,341 @@ public double getA() {
...
+        return otherSide(adm);
     }
 
-    public SV otherSide(double r, double x, double g, double b, double rho) {
-        Complex z = new Complex(r, x); // z=r+jx
-        Complex y = new Complex(g, b); // y=g+jb
-        Complex u2 = computeU2(y, z, rho);
-        Complex s2 = computeS2(y, z, rho);
-        return new SV(-s2.getReal(), -s2.getImaginary(), u2.abs(), Math.toDegrees(u2.getArgument()));
+    // Has been modified
```
zamarrenolm, line 100: No need to comment on the previous implementation

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
```diff
@@ -84,132 +74,341 @@ public double getA() {
...
+        return otherSide(getR(twt), getX(twt), getG(twt), getB(twt), 0.0, 0.0, getRho(twt), getAlpha(twt));
     }
 
-    public SV otherSide(Line l) {
-        return otherSide(l.getR(), l.getX(), l.getG1() + l.getG2(), l.getB1() + l.getB2(), 1);
+    public SV otherSide(TwoWindingsTransformer twt, boolean splitShuntAdmittance) {
+        if (splitShuntAdmittance) {
+            return otherSide(getR(twt), getX(twt), getG(twt) * 0.5, getB(twt) * 0.5, getG(twt) * 0.5, getB(twt) * 0.5, getRho(twt), getAlpha(twt));
+        } else {
+            return otherSide(getR(twt), getX(twt), getG(twt), getB(twt), 0.0, 0.0, getRho(twt), getAlpha(twt));
```
zamarrenolm, line 123: Maybe rewrite else as:

`... else return otherSide(twt)`

to make clear that splitShuntAdmittance is not the "default" ?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
```diff
@@ -84,132 +74,341 @@ public double getA() {
...
+        return otherSide(dl.getR(), dl.getX(), dl.getG(), dl.getB(), 0.0, 0.0, 1.0, 0.0);
     }
 
-    public SV otherSide(DanglingLine dl) {
-        return otherSide(dl.getR(), dl.getX(), dl.getG() / 2.0, dl.getB() / 2.0, dl.getG() / 2.0, dl.getB() / 2.0, 1);
+    public SV otherSide(DanglingLine dl, boolean splitShuntAdmittance) {
+        if (splitShuntAdmittance) {
+            return otherSide(dl.getR(), dl.getX(), dl.getG() * 0.5, dl.getB() * 0.5, dl.getG() * 0.5, dl.getB() * 0.5, 1.0, 0.0);
+        } else {
+            return otherSide(dl.getR(), dl.getX(), dl.getG(), dl.getB(), 0.0, 0.0, 1.0, 0.0);
```
zamarrenolm, line 137: write else as `return otherSide(dl)` ?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
```diff
@@ -84,132 +74,341 @@ public double getA() {
...
     }
 
+    public double otherSideP(double r, double x, double g1, double b1, double g2, double b2, double rho, double alpha) {
+        LinkData.BranchAdmittanceMatrix adm = LinkData.calculateBranchAdmittance(r, x, 1 / rho, -alpha, 1.0, 0.0,
+            new Complex(g1, b1), new Complex(g2, b2));
+        return otherSideP(adm);
+    }
+
+    // Has been modified
+    // Before: otherSideP(dl.getR(), dl.getX(), dl.getG() / 2.0, dl.getB() / 2.0, dl.getG() / 2.0, dl.getB() / 2.0, 1.0, 0.0);
```
zamarrenolm, line 157: No need to comment on previous implementation

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
```diff
@@ -84,132 +74,341 @@ public double getA() {
...
-    public double otherSideQ(TieLine.HalfLine hl) {
-        return otherSideQ(hl.getR(), hl.getX(), hl.getG1(), hl.getB1(), hl.getG2(), hl.getB2(), 1.0);
+    public double otherSideQ(double r, double x, double g1, double b1, double g2, double b2, double rho, double alpha) {
+        LinkData.BranchAdmittanceMatrix adm = LinkData.calculateBranchAdmittance(r, x, 1 / rho, -alpha, 1.0, 0.0,
+            new Complex(g1, b1), new Complex(g2, b2));
+        return otherSideQ(adm);
     }
 
+    // Has been modified
+    // Before: otherSideQ(dl.getR(), dl.getX(), dl.getG() / 2.0, dl.getB() / 2.0, dl.getG() / 2.0, dl.getB() / 2.0, 1.0, 0.0);
```
zamarrenolm, line 190: No need to comment on previous implementation

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
```diff
@@ -84,132 +74,341 @@ public double getA() {
...
-    public double otherSideU(TieLine.HalfLine hl) {
-        return otherSideU(hl.getR(), hl.getX(), hl.getG1(), hl.getB1(), 1.0);
+    public double otherSideU(double r, double x, double g1, double b1, double g2, double b2, double rho, double alpha) {
+        LinkData.BranchAdmittanceMatrix adm = LinkData.calculateBranchAdmittance(r, x, 1 / rho, -alpha, 1.0, 0.0,
+            new Complex(g1, b1), new Complex(g2, b2));
+        return otherSideU(adm);
     }
 
+    // Has been modified
+    // Before: otherSideU(dl.getR(), dl.getX(), dl.getG() / 2.0, dl.getB() / 2.0, 0.0, 0.0, 1.0, 0.0);
```
zamarrenolm, line 223: no need to comment previous implementation

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
```diff
@@ -84,132 +74,341 @@ public double getA() {
...
     public double otherSideU(DanglingLine dl) {
-        return otherSideU(dl.getR(), dl.getX(), dl.getG() / 2.0, dl.getB() / 2.0, 1);
+        return otherSideU(dl.getR(), dl.getX(), dl.getG(), dl.getB(), 0.0, 0.0, 1.0, 0.0);
+    }
+
+    public double otherSideU(DanglingLine dl, boolean splitShuntAdmittance) {
+        if (splitShuntAdmittance) {
+            return otherSideU(dl.getR(), dl.getX(), dl.getG() * 0.5, dl.getB() * 0.5, dl.getG() * 0.5, dl.getB() * 0.5, 1.0, 0.0);
+        } else {
+            return otherSideU(dl.getR(), dl.getX(), dl.getG(), dl.getB(), 0.0, 0.0, 1.0, 0.0);
```
zamarrenolm, line 232: use `return otherSideU(dl)`

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
```diff
@@ -84,132 +74,341 @@ public double getA() {
...
+        }
+    }
+
+    // Has been modified
+    // Before: otherSideU(hl.getR(), hl.getX(), hl.getG1(), hl.getB1(), 0.0, 0.0, 1.0, 0.0);
```
zamarrenolm, line 237: remove comments about previous implementation

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
```diff
@@ -84,132 +74,341 @@ public double getA() {
...
+    public double otherSideA(DanglingLine dl, boolean splitShuntAdmittance) {
+        if (splitShuntAdmittance) {
+            return otherSideA(dl.getR(), dl.getX(), dl.getG() * 0.5, dl.getB() * 0.5, dl.getG() * 0.5, dl.getB() * 0.5, 1.0, 0.0);
+        } else {
+            return otherSideA(dl.getR(), dl.getX(), dl.getG(), dl.getB(), 0.0, 0.0, 1.0, 0.0);
+        }
+    }
+
+    // Has been modified
+    // Before: otherSideA(hl.getR(), hl.getX(), hl.getG1(), hl.getB1(), 0.0, 0.0, 1.0, 0.0);
```
zamarrenolm, line 272: remove comments about previous implementation

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
```diff
@@ -84,132 +74,341 @@ public double getA() {
...
+    public double otherSideA(TieLine.HalfLine hl) {
+        return otherSideA(hl.getR(), hl.getX(), hl.getG1(), hl.getB1(), hl.getG2(), hl.getB2(), 1.0, 0.0);
+    }
+
+    public static double getRho(TwoWindingsTransformer twt) {
```
zamarrenolm, line 277: Consider making these functions (`getRho`, `getAlpha`, `getR`, `getX`, ...) private instead of public. They are only helpers to obtain actual parameters of transformers, and are used only in this class.

zamarrenolm (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
marqueslanauja: Done

zamarrenolm (approved):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
```diff
@@ -18,55 +21,45 @@
...
 
-    public static double getB(TwoWindingsTransformer twt) {
-        return twt.getB() * twt.getOptionalRatioTapChanger().map(rtc -> 1 + rtc.getCurrentStep().getB() / 100).orElse(1d);
+    /**
+     * @deprecated Not used anymore. The end associated to the voltage and flow must be defined. Use
+     * {@link SV#SV(double, double, double, double, int)}
+     */
+    @Deprecated
+    public SV(double p, double q, double u, double a) {
+        throw new PowsyblException("Deprecated. Not used anymore");
```
MioRtia, line 51: If you deprecate it, don't throw exception. Log a warning explaining it always uses side ONE and return SV(p, q, u, a, ONE)

ampl-converter/src/main/java/com/powsybl/ampl/converter/AmplNetworkWriter.java
```diff
@@ -490,7 +490,7 @@ private void writeDanglingLineMiddleBuses(AmplExportContext context, TableFormat
                 context.busIdsToExport.add(middleBusId);
                 int middleBusNum = mapper.getInt(AmplSubset.BUS, middleBusId);
                 int middleVlNum = mapper.getInt(AmplSubset.VOLTAGE_LEVEL, middleVlId);
-                SV sv = new SV(t.getP(), t.getQ(), b != null ? b.getV() : Double.NaN, b != null ? b.getAngle() : Double.NaN).otherSide(dl);
+                SV sv = new SV(t.getP(), t.getQ(), b != null ? b.getV() : Double.NaN, b != null ? b.getAngle() : Double.NaN, Branch.Side.ONE).otherSide(dl, true);
```
MioRtia, line 493: Not sure I understand why you need to specify the side for dangling lines...

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
```diff
@@ -84,132 +77,336 @@ public double getA() {
...
-    private Complex computeU2(Complex y, Complex z, double rho) {
-        return computeU2(y, new Complex(0, 0), z, rho);
+    /**
+     * @deprecated Not used anymore. This version with a simplified view of the parameters
+     * of a generic branch has been deprecated to avoid misuse. Use the version that includes rho AND alpha.
+     * {@link SV#otherSide(double, double, double, double, double, double, double, double)}
+     */
+    @Deprecated
+    public SV otherSide(double r, double x, double g, double b, double rho) {
+        throw new PowsyblException("Deprecated. Not used anymore");
```
MioRtia, line 91: Same remark as above

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
```diff
@@ -84,132 +77,336 @@ public double getA() {
...
-        Complex i2 = i2p.subtract(y2.multiply(v2)); // i2=i2p-y2*v2
-        return v2.multiply(3).multiply(i2.conjugate()); // s2=3*v2*conj(i2)
+    /**
+     * @deprecated Not used anymore. This version with a simplified view of the parameters
+     * of a generic branch has been deprecated to avoid misuse. Use the version that includes rho AND alpha.
+     * {@link SV#otherSide(double, double, double, double, double, double, double, double)}
+     */
+    @Deprecated
+    public SV otherSide(double r, double x, double g1, double b1, double g2, double b2, double rho) {
+        throw new PowsyblException("Deprecated. Not used anymore");
```
MioRtia, line 101: Same remark as above

MioRtia (commented):
Just for me: I'm not sure what the side changes for lines/tie lines, can you explain why we have different results?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/HalfLineBoundaryImpl.java
```diff
@@ -29,28 +29,28 @@
     public double getV() {
         Terminal t = getConnectable().getTerminal(side);
         Bus b = t.getBusView().getBus();
-        return new SV(t.getP(), t.getQ(), getV(b), getAngle(b)).otherSideU(parent);
+        return new SV(t.getP(), t.getQ(), getV(b), getAngle(b), side).otherSideU(parent);
```
annetill, line 32: I don't understand why you have here `side` and not `Branch.Side.ONE`. Same remark for the three following methods. If I change in the code side by  `Branch.Side.ONE` or  `Branch.Side.TWO`, there is no impact in the unit tests. Same conclusion for dangling lines.

annetill (changes requested):
See comment.

marqueslanauja:
> Just for me: I'm not sure what the side changes for lines/tie lines, can you explain why we have different results? If there is no difference, side shouldn't be a parameter in the constructor but a parameter of otherSide(transformer, side)

The calculation of the flow and voltage on one side by using the values of the other side and the parameters of the branch is symmetrical when the impedance transmission FromTo and the impedance transmission ToFrom are equals and the shunt admittance at both ends are also equals. So in the following situations the calculation will not be symmetrical:

- Transformers with ratioTapChanger and ratio non-equal to 1.0. 
- Transformers with phaseTapChanger and angle non-zero or ratio non-equal to 1.0.
- Lines with different shunt admittance at both ends. (In the IIDM model, there is a specific shunt admittance for each end and in PSS/E you can specify additional shunt admittances at each end, so both ends will have different values in many cases)
- DanglingLines when the splitShuntAdmittance flag is not set to true (In the IIDM model, there is only shunt admittance at the end one) 

We think that the best implementation to avoid misuse is **to always specify** an additional field that defines the side associated with the given voltage and flow values.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
marqueslanauja: We prefer to throw an exception to force the user to review the code. In some cases the difference could be small and the best way to be sure that the code is ok is to review it.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/HalfLineBoundaryImpl.java
annetill: @marqueslanauja The `side` here refers to the side of the half line in the tie line. Are you sure that it is the good side and not `SIDE.ONE`?

ampl-converter/src/main/java/com/powsybl/ampl/converter/AmplNetworkWriter.java
marqueslanauja: It is not necessary if we specify the splitShuntAdmittance to true, but to avoid a future misuse we prefer to specify always the side. We think that the side must be always specified.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
marqueslanauja: To avoid misuse we prefer to have only one method with all the parameters.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
marqueslanauja: To avoid misuse we prefer to have only one method with all the parameters.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/HalfLineBoundaryImpl.java
marqueslanauja: In the halfLine of a TieLine the model end is not always at side ONE. There is a specific field `side` that defines the side associated with the model end. There will not be impact on the unit tests if the shunt admittance at both ends are equal, so that will be the situation of this test.

marqueslanauja: I think should be `side`. After this PR  I will make another one to fix the issues associated with the Tie-Lines (Branch `TieLine-review`). In this new PR I can add some tests to check that  should be `side`.

MioRtia:
I have to think, I find it confusing to use side for component which do not use explicitly side's concept (dangling lines and half lines)

marqueslanauja:
> I have to think, I find it confusing to use side for component which do not use explicitly side's concept (dangling lines and half lines)

I am going to prepare today some slides trying to clarify the side concept in branches, danglingLines and in TieLines.

ampl-converter/src/main/java/com/powsybl/ampl/converter/AmplNetworkWriter.java
zamarrenolm: Although dangling lines in IIDM only have one terminal (one _side_), they are modelling branches with two ends. So even for dangling lines, the SV utility class needs to know for which end are you giving the voltage and flow, when you want to calculate the _otherSide_ magnitudes. For dangling lines the _network_ end is always considered side `ONE` and the _boundary_ end is side `TWO`.

PR #1732, opened 2021-05-12
Remote reactive power control extension

obrix:

This pull request introduce an extension to handle remote reactive power control. This is work in progress.

Remote reactive power control is not supported.


MioRtia (approved):
It is okay for me. The only thing missing is a javadoc for `RemoteReactivePowerControl` to explain what it is.

obrix:
@MioRtia I'll add it, thanks for the review.

geofjamg (dismissed):
Be careful, licences header are missing everywhere

annetill (approved):

marqueslanauja (commented):
Maybe in the method  `setRegulatingControlReactivePower `of the class `RegulatingControlMappingForGenerators` we should consider the `qPercent` and `eqControlEnabled` attributes as in the voltage control (`setRegulatingControlVoltage `method) ?

annetill:
> Maybe in the method `setRegulatingControlReactivePower `of the class `RegulatingControlMappingForGenerators` we should consider the `qPercent` and `eqControlEnabled` attributes as in the voltage control (`setRegulatingControlVoltage `method) ?

It means that for you, for a generator with reactive remote control extension, we will have an other extension to say that that remote reactive control is shared with an other generator? What the `eqControlEnabled` will introduce as change in IIDM modelling?

marqueslanauja:
> > Maybe in the method `setRegulatingControlReactivePower `of the class `RegulatingControlMappingForGenerators` we should consider the `qPercent` and `eqControlEnabled` attributes as in the voltage control (`setRegulatingControlVoltage `method) ?
> 
> It means that for you, for a generator with reactive remote control extension, we will have an other extension to say that that remote reactive control is shared with an other generator? What the `eqControlEnabled` will introduced as change in IIDM modelling?

Since we record the reactive power control in an extension, we can use the same extension to record the  `qPercent` by adding a new field. 
The `eqControlEnabled` is used to determine if the IIDM control is enabled or not. The IIDM control is enabled if the enabled flag at the control and the enabled flag at the equipment are true.

MioRtia:
@marqueslanauja Are the corrections okay for you?

marqueslanauja (approved):
It is ok.

Issue #1699, opened 2021-04-26
[Security analysis] On demand results for specific equipments


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature

* **What is the current behavior?**

Security analysis API only reports limit violations, which means that lots of intermediate computation results which could be interesting to the user are lost.

* **What is the expected behavior?**

It would be useful to be able to request results for specific equipments, in a similar way as sensitivity analysis.
In particular, it would be useful to request : 
* current and active power on a set of branches
* voltage on a set of buses

* **What is the motivation / use case for changing the behavior?**

Efficiently computing the post-contingency currents on a list of contingencies.

sylvlecl:
So we need to add 2 things:
1. specification of the results we want, in the input of the security analysis
2. results corresponding to those specifications, in the security analysis results

**1. Requesting some results**

The specification of results should be POJOs provided to the `SecurityAnalysisProvider::run` method.

Why POJOs: it should be serializable, and easy to provide to an external tool. Any logic (like instantiating from the network) should be performed outside of the computation

They can extend a marker interface `ResultRequest`, which could have at first an implementation `BranchResultRequest`
```java
// Specification for 1 situation
class StateResultSpecification {
   String contingencyId;       //Optional, so that it's usable for N spec, to keep it simple ...
   List<String> branchIds;    //Ask detailed information for those branches
                                           // --> should we make it more detailed? (like P, Q, I...)
   List<String> busIds;         //Ask detailed information for those buses
                                           //  --> should we make it more detailed? is it OK to use bus IDs here ?
                                           //(in particular, what about buses that change after a contingency ...
}

class SecurityAnalysisResultSpecification {
    StateResultSpecification commonResultSpecifications;
    StateResultSpecification nSituation;
    List<StateResultSpecification> contingencyResultSpecifications;
}
```

The provider should:
 * provide values of currents, active power, and reactive power on both terminals of specified branches.
 * provide voltage for all buses specified in the list of buses 

For one particular state (N-situation or contingencies), the actual specification will be the concatenation of the "common" specification and of the specific one, if it exists.

**2. Results **

geofjamg:
- In the new sensitivity API (that will come soon on core side!), we have added the concept of contingency context (https://github.com/powsybl/powsybl-open-loadflow/blob/master/src/main/java/com/powsybl/openloadflow/sensi/ContingencyContext.java). Maybe it would make sens to share it with the security analysis?

[...]

PR #1737, opened 2021-05-19
Monitor security analysis

EtienneLt:

#1699 

**Other information**:


security-analysis/security-analysis-api/src/main/java/com/powsybl/security/PostContingencyResult.java
```diff
@@ -22,9 +26,19 @@
 
     private final LimitViolationsResult limitViolationsResult;
 
+    Map<String, BranchResult> branchResults;
+
+    Map<String, BusResults> busResults;
```
sylvlecl, line 31: We must provide getters for the new fields. More precisely, in order to not let the user modify anything in the map, we can provide accessors for the collection fo results, or for on result by its ID:
```java
public BusResult getBusResult(String id) { return busResults.get(id); }
public Collection<BusResult> getBusResults()  { return busResults.values(); }
```
Same for branch results

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/PostContingencyResult.java
```diff
@@ -22,9 +26,19 @@
 
     private final LimitViolationsResult limitViolationsResult;
 
+    Map<String, BranchResult> branchResults;
```
sylvlecl, line 29: We must add a test for JSON serialization / deserialization of the new results, too.
With the current code, I don't think the new result will be written --> to be tested :) 

Note that the result would be better serialized as a list, not a map, because otherwise the ID information will be repeated (once as a key, once as a field of the value). The `Map` is only useful to provide access to the results by their equipment ID.

security-analysis/security-analysis-default/src/main/java/com/powsybl/security/impl/DefaultSecurityAnalysis.java
```diff
@@ -212,6 +220,13 @@ private void setContingencyOkAndCheckViolations(String postContVariantId, Securi
         SecurityAnalysisResultBuilder.PostContingencyResultBuilder builder = resultBuilder.contingency(contingency).setComputationOk(lfResult.isOk());
         if (lfResult.isOk()) {
             violationDetector.checkAll(contingency, network, builder::addViolation);
+            monitors.forEach(stateMonitor -> {
```
sylvlecl, line 223: We must also do it for N situation result

sylvlecl, line 223: Here we will have a performance issue because of the linear search (if we have 10000 contingencies and 10000 "monitors", it will take forever).

In order to solve this, we must probably start by translating, at the start of the computaiton, the input list into an indexed object, so that here we just have to call a more efficient `StateMonitor contingencyMonitor = index.getMonitor(contingencyId);`

sylvlecl (changes requested):

MioRtia:
Could you please resolve the conflicts of this PR?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/execution/SecurityAnalysisExecutionInput.java
```diff
@@ -37,6 +39,7 @@
     private final List<String> resultExtensions = new ArrayList<>();
     private final Set<LimitViolationType> violationTypes = EnumSet.noneOf(LimitViolationType.class);
     private boolean withLogs = false;
+    private List<StateMonitor> monitors = new ArrayList<>();
```
MioRtia, line 42: This can be final

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/execution/SecurityAnalysisExecutionInput.java
```diff
@@ -58,6 +61,10 @@ public NetworkVariant getNetworkVariant() {
         return networkVariant;
     }
 
+    public List<StateMonitor> getMonitors() {
+        return monitors;
```
MioRtia, line 65: Return an immutable copy of it

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/monitor/ContingencyContext.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+package com.powsybl.security.monitor;
+
+import java.util.Objects;
+
+/**
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ * @author Etienne Lesot <etienne.lesot at rte-france.com>
+ */
```
MioRtia, line 15: Could you add some javadoc about this class?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/monitor/ContingencyContextType.java
```diff
@@ -0,0 +1,18 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package com.powsybl.security.monitor;
+
+/**
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ * @author Etienne Lesot <etienne.lesot at rte-france.com>
+ */
```
MioRtia, line 13: Could you add some javadoc about this class?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/monitor/StateMonitor.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+package com.powsybl.security.monitor;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * @author Etienne Lesot <etienne.lesot at rte-france.com>
+ */
```
MioRtia, line 15: Could you add some javadoc about this class?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/monitor/StateMonitor.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public List<String> getVoltageLevelIds() {
+        return voltageLevelIds;
+    }
+
+    public StateMonitor(ContingencyContext contingencyContext, List<String> branchIds, List<String> voltageLevelIds) {
+        this.contingencyContext = Objects.requireNonNull(contingencyContext);
+        this.branchIds = Objects.requireNonNull(branchIds);
+        this.voltageLevelIds = Objects.requireNonNull(voltageLevelIds);
```
MioRtia, line 39: To prevent abuse in lists' manipulation, create list and addAll the parameters

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/monitor/StateMonitor.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public StateMonitor() {
+        this.contingencyContext = null;
+        this.branchIds = Collections.emptyList();
+        this.voltageLevelIds = Collections.emptyList();
```
MioRtia, line 45: Call the complete constructor : `this(null, Collections.emptyList(), Collections.emptyList());`

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/results/BusResults.java
```diff
@@ -0,0 +1,51 @@
...
+package com.powsybl.security.results;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+
+/**
+ * @author Etienne Lesot <etienne.lesot at rte-france.com>
+ */
```
MioRtia, line 16: Could you add some javadoc?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/results/BranchResult.java
```diff
@@ -0,0 +1,73 @@
...
+package com.powsybl.security.results;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+
+/**
+ * @author Etienne Lesot <etienne.lesot at rte-france.com>
+ */
```
MioRtia, line 16: Could you add some javadoc?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/results/BusResults.java
```diff
@@ -0,0 +1,51 @@
+/**
+ * Copyright (c) 2016-2017, RTE (http://www.rte-france.com)
```
MioRtia, line 2: Could you correct the year of the copyright?

security-analysis/security-analysis-api/src/test/java/com/powsybl/security/json/JsonMonitorTest.java
```diff
@@ -0,0 +1,26 @@
+package com.powsybl.security.json;
```
MioRtia, line 1: Don't forget the copyright

security-analysis/security-analysis-api/src/test/java/com/powsybl/security/json/PostContingencyResultTest.java
```diff
@@ -0,0 +1,60 @@
+package com.powsybl.security.json;
```
MioRtia, line 1: Don't forget the copyright

MioRtia (changes requested):
Thank you for your work! I have some remarks:
- Please do check the copyrights of your new files (I don't think I checked them all)
- Could you add some javadoc on the new concepts you introduce?
- Do you intend to update the implementation on Hades2 for monitors? (all the other adaptation has already been done)

I also left some technical corrections to do on collections' getters.

EtienneLt:
> Thank you for your work! I have some remarks:
> 
>     * Please do check the copyrights of your new files (I don't think I checked them all)
> 
>     * Could you add some javadoc on the new concepts you introduce?
> 
>     * Do you intend to update the implementation on Hades2 for monitors? (all the other adaptation has already been done)
> 
> 
> I also left some technical corrections to do on collections' getters.

EtienneLt:
> Thank you for your work! I have some remarks:
> 
>     * Please do check the copyrights of your new files (I don't think I checked them all)
> 
>     * Could you add some javadoc on the new concepts you introduce?
> 
>     * Do you intend to update the implementation on Hades2 for monitors? (all the other adaptation has already been done)
> 
> 
> I also left some technical corrections to do on collections' getters.

On Hades2 I've been told I should not do it but I'm doing it on Open loadflow

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/monitor/ContingencyContext.java
```diff
@@ -0,0 +1,65 @@
...
+
+package com.powsybl.security.monitor;
+
+import java.util.Objects;
+
+/**
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ * @author Etienne Lesot <etienne.lesot at rte-france.com>
+ *
+ * provide the context to get informations of the network after a loadflow
```
MioRtia, line 16: Small typo, `information` is invariable

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/monitor/ContingencyContextType.java
```diff
@@ -0,0 +1,21 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package com.powsybl.security.monitor;
+
+/**
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ * @author Etienne Lesot <etienne.lesot at rte-france.com>
+ *
+ * Define if the context is N_Situation, N-k Situation (Specific) or both (All)
```
MioRtia, line 14: Could you define what each state of the network the enum's values describe?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/monitor/ContingencyContext.java
MioRtia: Could you add more detail? How is it used, what information does it contain?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/monitor/StateMonitor.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+package com.powsybl.security.monitor;
+
+import java.util.*;
+
+/**
+ * @author Etienne Lesot <etienne.lesot at rte-france.com>
+ *
+ *
+ * A stateMonitor allows to get informations on the network after a loadflow computation
```
MioRtia, line 15: Small typo, `information` is invariable.
Maybe specify the information is linked to the contingencies specified in the context

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/results/BranchResult.java
```diff
@@ -0,0 +1,74 @@
...
+package com.powsybl.security.results;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+
+/**
+ * provide electrical informations on a branch
```
MioRtia, line 15: Small typo, `information` is invariable.
Be specific about the information: of which state? when is this class used?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/results/BusResults.java
```diff
@@ -0,0 +1,54 @@
...
+package com.powsybl.security.results;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+
+/**
+ *
+ *  provide electrical informations on a bus
```
MioRtia, line 16: Small typo, `information` is invariable.
Same remark as other javadocs.

MioRtia (commented):
My comments are only about javadoc/documentation. Otherwise, it is good for me!

MioRtia:
> just a question : there is already a ContingencyContext in sensi api maybe I should place this class in a common repository like contingency ?

You're right! Indeed, `ContingencyContext` (and `ContingencyContextType`) should be in contingency-api. Please rename `N_SITUATION` as  `NONE` and explain in the javadoc that `NONE` refers to a pre-contingency situation, `SPECIFIC` focuses on specific contingencies in a post-contingency situation and `ALL` refers to pre- and post-contingency situations.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/monitor/StateMonitor.java
```diff
@@ -0,0 +1,63 @@
...
+package com.powsybl.security.monitor;
+
+import java.util.*;
+
+/**
+ * @author Etienne Lesot <etienne.lesot at rte-france.com>
+ *
+ *
+ * A stateMonitor allows to get information about branch and bus on the network after a security analysis computation.
+ * Contingency context allows to specify if the informations asked are about N situation N-k situation with a contingency id
```
MioRtia, line 16: Small typo: `information` does not take an `s`.

Moreover, could you give an example of information it could get? How getting information work? I understand that StateMonitor keeps IDs of branches and voltage levels in memory, how does it work from that?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/monitor/StateMonitor.java
```diff
@@ -0,0 +1,63 @@
...
+    private final ContingencyContext contingencyContext;
+
+    /**
+     * branchs ids on which information will be collected
+     */
+    private final Set<String> branchIds = new LinkedHashSet<>();
+    /**
+     * voltageLevels ids on which information will be collected
+     */
+    private final Set<String> voltageLevelIds = new LinkedHashSet<>();
```
MioRtia, line 34: Do we need order in these sets? If not, you can use `HashSet`

EtienneLt: Yes I need them

EtienneLt:
> > just a question : there is already a ContingencyContext in sensi api maybe I should place this class in a common repository like contingency ?
> 
> You're right! Indeed, `ContingencyContext` (and `ContingencyContextType`) should be in contingency-api. Please rename `N_SITUATION` as `NONE` and explain in the javadoc that `NONE` refers to a pre-contingency situation, `SPECIFIC` focuses on specific contingencies in a post-contingency situation and `ALL` refers to pre- and post-contingency situations.

I think N_SITUATION is more explicit and @sylvlecl and @geofjamg agreed that was better but I can change it

geofjamg:
> > > just a question : there is already a ContingencyContext in sensi api maybe I should place this class in a common repository like contingency ?
> > 
> > 
> > You're right! Indeed, `ContingencyContext` (and `ContingencyContextType`) should be in contingency-api. Please rename `N_SITUATION` as `NONE` and explain in the javadoc that `NONE` refers to a pre-contingency situation, `SPECIFIC` focuses on specific contingencies in a post-contingency situation and `ALL` refers to pre- and post-contingency situations.
> 
> I think N_SITUATION is more explicit and @sylvlecl and @geofjamg agreed that was better but I can change it

Probably None should be rename, but I am not convinced N-SITUATION is the best name. It looks like "franglais" and in my past European project it was often called "base case" or 'pre-contingency state"

geofjamg:
Thinking a little bit more NONE was not so bad associated to "contingency context". It is semantically clear, that it is no associated to any contingency so it is the base case.

sylvlecl:
> Thinking a little bit more NONE was not so bad associated to "contingency context". It is semantically clear, that it is no associated to any contingency so it is the base case.

But it's quite implicit, I would prefer to have an explicit reference. It sounds like you answer "none" to the question "on which contingency should I use it?", which just means "don't use it on contingencies", but not "Use it on N situation". However with some javadoc it would be more clear.

Then if we go for the explicit wording, the "N situation" comes from the european [system operations code](https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=uriserv:OJ.L_.2017.220.01.0001.01.ENG&toc=OJ:L:2017:220:TOC#d1e394-1-1), so it's written this way in the law :)  
However as discussed, it does not mean it is the most used word in practice and in all contexts : it would deserve a kind of survey.

zamarrenolm:
> > Thinking a little bit more NONE was not so bad associated to "contingency context". It is semantically clear, that it is no associated to any contingency so it is the base case.
> 
> But it's quite implicit, I would prefer to have an explicit reference. It sounds like you answer "none" to the question "on which contingency should I use it?", which just means "don't use it on contingencies", but not "Use it on N situation". However with some javadoc it would be more clear.
> 
> Then if we go for the explicit wording, the "N situation" comes from the european [system operations code](https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=uriserv:OJ.L_.2017.220.01.0001.01.ENG&toc=OJ:L:2017:220:TOC#d1e394-1-1), so it's written this way in the law :)
> However as discussed, it does not mean it is the most used word in practice and in all contexts : it would deserve a kind of survey.

Agree on the fact that "N-Situation" is the formal definition in Europe, but the expression "base case" is much more common in the literature about contingency analysis.

sylvlecl:
More thought about the wording : in the world of capacity calculation (flow based in particular), **base case** is used quite often.

But I think that notion means more than just "N situation", it also "contains" the set of inter-TSO exchanges corresponding to a "best estimate", as well as topology etc. 

So when talking specifically of contingencies, I think "precontingency" or "N situation" is more relevant.

security-analysis/security-analysis-default/src/test/java/com/powsybl/security/impl/mocks/LoadFlowProviderMock.java~HEAD
```diff
@@ -0,0 +1,45 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.Collections;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ */
+@AutoService(LoadFlowProvider.class)
+public class LoadFlowProviderMock implements LoadFlowProvider {
```
MioRtia, line 26: @EtienneLt Be careful, this file is called LoadFlowProviderMock.java-HEAD ! This is what is causing the fail in compilation

security-analysis/security-analysis-default/src/test/java/com/powsybl/security/impl/mocks/LoadFlowProviderMock.java~Refactoring security-analysis-api : SecurityAnalysisImpl (issue 1604) (#1648)
```diff
@@ -0,0 +1,45 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.Collections;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ */
+@AutoService(LoadFlowProvider.class)
+public class LoadFlowProviderMock implements LoadFlowProvider {
```
MioRtia, line 26: @EtienneLt Same issue about file naming

security-analysis/security-analysis-default/src/test/resources/com/powsybl/config/test/filelist.txt~HEAD
```diff
@@ -0,0 +1 @@
+config.yml
```
MioRtia, line 1: Same issue

security-analysis/security-analysis-default/src/test/resources/com/powsybl/config/test/filelist.txt~Refactoring security-analysis-api : SecurityAnalysisImpl (issue 1604) (#1648)
```diff
@@ -0,0 +1 @@
+config.yml
```
MioRtia, line 1: Same issue

security-analysis/security-analysis-default/src/main/java/com/powsybl/security/impl/DefaultSecurityAnalysis.java
```diff
@@ -78,15 +84,26 @@ private static ExecutorService createThreadPool(int poolSize) {
...
         this.violationDetector = Objects.requireNonNull(detector);
         this.violationFilter = Objects.requireNonNull(filter);
         this.interceptors = new ArrayList<>();
         this.computationManager = Objects.requireNonNull(computationManager);
-
+        Objects.requireNonNull(monitors);
+        this.monitors = new HashMap<>();
+        monitors.forEach(monitor -> {
+            String id = monitor.getContingencyContext().getContingencyId();
+            if (this.monitors.containsKey(id)) {
```
sylvlecl, line 101: Tip: this could be done with a 1 liner:
```
this.monitors.merge(id, monitor, StateMonitor::merge);
```

But as explained above we need to review it a little more deeply to handle `ALL` and `NONE` cases explicitly.
With the current code, all `ALL` and `NONE` contexts will be merged together, since they will have a `null` contingency ID. It will produce a bug, because at the end we will only have a `ALL` or `NONE` monitor, depending on the execution order.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/PostContingencyResult.java
```diff
@@ -35,11 +55,28 @@ public PostContingencyResult(Contingency contingency, boolean computationOk, Lis
...
     public Contingency getContingency() {
         return contingency;
     }
 
     public LimitViolationsResult getLimitViolationsResult() {
         return limitViolationsResult;
     }
+
+    public List<BranchResult> getBranchResultsAsList() {
+        return new ArrayList<>(branchResults.values());
```
sylvlecl, line 72: A detail but for all those getters I would advise to use `List.copyOf`, to return an immutable copy instead of a mutable one.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResult.java
```diff
@@ -49,4 +67,16 @@ public LimitViolationsResult getPreContingencyResult() {
     public List<PostContingencyResult> getPostContingencyResults() {
         return postContingencyResults;
     }
+
+    public List<BranchResult> getBranchResultsAsList() {
```
sylvlecl, line 71: In that class too:
* remove the `AsList` suffix
* use `List.copyOf` instead of array list
* provide getters by ID too

But we also need to rename those getters, because it's not clear that they are pre-contingency result.
But I must say it gets a little bit ugly : `getPreContingencyThreeWindingTransformerResults`.

Should we move all those results in a precontingency result objets ? 
If we do it and move the limit violations into it, we need to take care of compatibility (API and JSON format) ...  little bit annoying too.

Any opinion @MioRtia @geofjamg ?

contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyContext.java
```diff
@@ -0,0 +1,70 @@
...
+
+    private final String contingencyId;
+
+    /**
+     * Define if informations needed are in N_Situation, N-k Situation or both
+     * if it is N_situation contingencyid is null
+     */
+    private final ContingencyContextType contextType;
+
+    public ContingencyContext() {
```
sylvlecl, line 33: I would remove that constructor (not explicit enough: what  does it create ?). The `none()` method can be used instead by the user.

contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyContextType.java
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ * @author Etienne Lesot <etienne.lesot at rte-france.com>
+ *
+ * Define for a contingencyContext the type of information asked.
+ * It can be n situation (N_SITUATION), n-k situation on a specific contingency (SPECIFIC) or both (ALL)
+ */
+
+public enum ContingencyContextType {
+    ALL,
```
sylvlecl, line 19: Better place the javadoc on each value, like
```java
/**
 *  Corresponds to all contingencies and pre-contingency situation
 */
ALL,
...
```
This way we get the correct doc when we ask the doc for one value (in your IDE for example).

security-analysis/security-analysis-api/src/test/java/com/powsybl/security/json/PostContingencyResultTest.java
```diff
@@ -0,0 +1,70 @@
...
+        roundTripTest(postContingencyResult, this::write, this::read, "/PostContingencyResultTest.json");
+    }
+
+    public void write(PostContingencyResult postContingencyResult, Path jsonFile) {
+        try {
+            OutputStream out = Files.newOutputStream(jsonFile);
+            JsonUtil.createObjectMapper()
+                .registerModule(new SecurityAnalysisJsonModule())
+                .writer()
+                .writeValue(out, postContingencyResult);
```
sylvlecl, line 54: Note: we could use the "pretty print" option to have more readable JSON files in unit tests, but OK this way.

sylvlecl (changes requested):
A few additional change requests, in particular on the indexing logic.

contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyContext.java
EtienneLt: I need a default constructor for serialization

MioRtia: Could you elaborate? Is it for future JSON serialization?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResult.java
MioRtia: Are you thinking about a `PreContingencyResult` object in parallel of `PostContingencyResult`?

sylvlecl: Yes something like that.

contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyContext.java
sylvlecl: There is already some JON serialization used to read from input file in security analysis tool and to write down input files for distributed security analysis.

It's very strange : it seems that jackson is able to change `private final` fields after construction ...
I would propose to either make this constructor private, or say in javadoc that it builds a "NONE" context, or remove the constructor and annotate the other one with jackson annotations (I would prefer that more explicit solution).

MioRtia: I agree, I think it is better to remove the constructor and annotate the other one.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/PostContingencyResult.java
```diff
@@ -22,9 +24,27 @@
...
+    private final Map<String, ThreeWindingsTransformerResult> threeWindingsTransformerResults;
+
     public PostContingencyResult(Contingency contingency, LimitViolationsResult limitViolationsResult) {
+        this(contingency, limitViolationsResult, Collections.emptyMap(), Collections.emptyMap(), Collections.emptyMap());
+    }
+
+    public PostContingencyResult(Contingency contingency, LimitViolationsResult limitViolationsResult, Map<String, BranchResult> branchResults, Map<String, BusResults> busResults, Map<String, ThreeWindingsTransformerResult> threeWindingsTransformerResults) {
         this.contingency = Objects.requireNonNull(contingency);
         this.limitViolationsResult = Objects.requireNonNull(limitViolationsResult);
+        this.branchResults = ImmutableMap.copyOf(Objects.requireNonNull(branchResults));
```
annetill, line 40: Are we missing something for dangling lines?

MioRtia (approved):

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/monitor/StateMonitorJson.java
```diff
@@ -0,0 +1,44 @@
...
+import java.io.OutputStream;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+/**
+ * @author Etienne Lesot <etienne.lesot at rte-france.com>
+ */
+public final class StateMonitorJson {
```
geofjamg, line 22: You should move it as static methods in StateMonitor itself

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/execution/SecurityAnalysisExecutionInput.java
```diff
@@ -97,6 +104,11 @@ public SecurityAnalysisExecutionInput setNetworkVariant(Network network, String
         return this;
     }
 
+    public SecurityAnalysisExecutionInput setMonitors(List<StateMonitor> monitors) {
```
geofjamg, line 107: This method does not do what it should, as it does not "set" but "add"

contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyContext.java
```diff
@@ -0,0 +1,85 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+
+/**
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ * @author Etienne Lesot <etienne.lesot at rte-france.com>
+ * <p>
+ * provide the context to get information of the network after a security analysis
```
geofjamg, line 18: We should have a consistent vocabulary in this API. Evrywhere we use pre/post contingency so we should use sames in doc.

contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyContextType.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package com.powsybl.contingency;
+
+/**
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ * @author Etienne Lesot <etienne.lesot at rte-france.com>
+ *
+ * Define for a contingencyContext the type of information asked.
+ * It can be n situation (N_SITUATION), n-k situation on a specific contingency (SPECIFIC) or both (ALL)
```
geofjamg, line 15: Same

geofjamg (changes requested):

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/SecurityAnalysisResultDeserializer.java
```diff
@@ -73,6 +84,24 @@ public SecurityAnalysisResult deserialize(JsonParser parser, DeserializationCont
                     });
                     break;
 
+                case "preContingencyBranchResults":
```
sylvlecl, line 87: To be removed : it's embedded in the precontingency result object (same below)

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/monitor/StateMonitorIndex.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public StateMonitorIndex(List<StateMonitor> stateMonitors) {
+        allStateMonitor = new StateMonitor(new ContingencyContext(null, ContingencyContextType.ALL),
+                new HashSet<>(), new HashSet<>(), new HashSet<>());
+        noneStateMonitor = new StateMonitor(new ContingencyContext(null, ContingencyContextType.NONE),
+                new HashSet<>(), new HashSet<>(), new HashSet<>());
+        stateMonitors.forEach(monitor -> {
+            String id = monitor.getContingencyContext().getContingencyId();
+            if (id != null) {
+                if (this.specificStateMonitors.containsKey(id)) {
```
sylvlecl, line 32: could be done with 1 liner with Map.merge (should be slightly more efficient)

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/monitor/StateMonitorIndex.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public StateMonitor getNoneStateMonitor() {
+        return noneStateMonitor;
+    }
+
+    public StateMonitor getSpecificStateMonitor(String id) {
+        return specificStateMonitors.get(id);
+    }
+
+    public Map<String, StateMonitor> getSpecificStateMonitors() {
```
sylvlecl, line 57: I think we can remove this

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/results/PreContingencyResult.java
```diff
@@ -0,0 +1,66 @@
...
+        this.preContingencyBranchResults.putAll(Objects.requireNonNull(preContingencyBranchResults));
+        this.preContingencyBusResults.putAll(Objects.requireNonNull(preContingencyBusResults));
+        this.preContingencyThreeWindingsTransformerResults.putAll(Objects.requireNonNull(preContingencyThreeWindingsTransformerResults));
+    }
+
+    public void setLimitViolationsResult(LimitViolationsResult limitViolationsResult) {
+        this.limitViolationsResult = limitViolationsResult;
+    }
+
+    public void addPreContingencyBranchResults(Map<String, BranchResult> branchResults) {
```
sylvlecl, line 39: Would be better to add `Collection` instead, otherwise we must trust the user to provide a consistent map (key = ID in the mapped value)

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/results/PreContingencyResult.java
```diff
@@ -0,0 +1,66 @@
...
+    private final Map<String, BranchResult> preContingencyBranchResults = new HashMap<>();
+    private final Map<String, BusResults> preContingencyBusResults = new HashMap<>();
+    private final Map<String, ThreeWindingsTransformerResult> preContingencyThreeWindingsTransformerResults = new HashMap<>();
+
+    public PreContingencyResult() {
+        this(null, Collections.emptyMap(), Collections.emptyMap(), Collections.emptyMap());
+    }
+
+    public PreContingencyResult(LimitViolationsResult preContingencyResult, Map<String, BranchResult> preContingencyBranchResults,
+                                Map<String, BusResults> preContingencyBusResults,
```
sylvlecl, line 27: Same, would be better to use `Collection` instead

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/results/PreContingencyResult.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public void addPreContingencyBusResults(Map<String, BusResults> busResults) {
+        this.preContingencyBusResults.putAll(Objects.requireNonNull(busResults));
+    }
+
+    public void addPreContingencyThreeWindingsTransformerResults(Map<String, ThreeWindingsTransformerResult> threeWindingsTransformerResults) {
+        this.preContingencyThreeWindingsTransformerResults.putAll(threeWindingsTransformerResults);
+    }
+
+    public LimitViolationsResult getLimitViolationsResult() {
```
sylvlecl, line 51: We should add ID based getters.

security-analysis/security-analysis-api/src/test/resources/SecurityAnalysisResultV1.json
```diff
@@ -0,0 +1,115 @@
...
+        "side" : "ONE",
+        "extensions" : {
+          "ActivePower" : {
+            "value" : 220.0
+          }
+        }
+      } ],
+      "actionsTaken" : [ ]
+    },
+    "preContingencyBranchResults" : [ {
```
sylvlecl, line 26: Should not appear in that V1 JSON

security-analysis/security-analysis-default/src/main/java/com/powsybl/security/impl/DefaultSecurityAnalysis.java
```diff
@@ -222,4 +237,30 @@ private void applyContingency(String workingVariantId, String postContVariantId,
         contingency.toTask().modify(network, computationManager);
     }
 
+    private void addMonitorInfos(Network network, StateMonitor monitor, Consumer<BranchResult> branchResultConsumer,
+                                 Consumer<BusResults> busResultsConsumer, Consumer<ThreeWindingsTransformerResult> threeWindingsTransformerResultConsumer) {
+        monitor.getBranchIds().forEach(branchId -> branchResultConsumer.accept(createBranchResult(network.getBranch(branchId))));
```
sylvlecl, line 242: I think it can throw an NPE if the branch does not exist : we should just ignore it instead

security-analysis/security-analysis-default/src/main/java/com/powsybl/security/impl/DefaultSecurityAnalysis.java
```diff
@@ -222,4 +237,30 @@ private void applyContingency(String workingVariantId, String postContVariantId,
...
+        monitor.getVoltageLevelIds().forEach(voltageLevelId ->
+            network.getVoltageLevel(voltageLevelId).getBusView().getBuses().forEach(bus ->
```
sylvlecl, line 244: Same, can throw an NPE

security-analysis/security-analysis-default/src/test/java/com/powsybl/security/impl/SecurityAnalysisTest.java
```diff
@@ -172,4 +177,62 @@ private static ComputationManager createMockComputationManager() {
         Mockito.when(computationManager.getResourcesStatus()).thenReturn(computationResourcesStatus);
         return computationManager;
     }
+
+    @Test
+    public void testStateMonitors() {
```
sylvlecl, line 182: We should probably add some unit test for the case where IDs are invalid

sylvlecl (changes requested):

geofjamg (commented):
A more general thought about json serialization, now we have a mix of explicit serialization and automatic via jackson annotation. Is it on purpose? Is it ok for everybody? So when we will evolve one of the annotated object, we will need to switch to explicit de/serializer for handle backward compatibility? Is it clear for everybody?

sylvlecl:
Indeed. Looks OK to me, as long as we don't need more we can use annotations, and if the need appears we switch to the heavier approach.
Note that annotations are already used elsewhere in the framework in versioned APIs (see sensitivity).

flo-dup:
I don't really like this explicit/automatic mix as it's within SecurityAnalysisResult (maybe because I don't like that much the annotations!). But if we decide to make it more uniform that could wait another PR anyway.

contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyContext.java
MioRtia: I changed the javadoc to use pre/post contingency. Did I miss something or was it not what you were referring to?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResult.java
```diff
@@ -42,11 +60,27 @@ public SecurityAnalysisResult setNetworkMetadata(NetworkMetadata networkMetadata
...
 
     public List<PostContingencyResult> getPostContingencyResults() {
         return postContingencyResults;
     }
+
+    public List<BranchResult> getBranchResultsAsList() {
+        return preContingencyResult.getPreContingencyBranchResults();
+    }
+
+    public List<BusResults> getBusResultsAsList() {
```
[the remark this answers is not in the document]
MioRtia: I would rather delete them as it makes their names rather clunky and non-user-friendly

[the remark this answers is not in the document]
geofjamg: Agreed, all theses shorcuts are useless, explicit 2 methods chaining would be more clear.

contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyContext.java
geofjamg: Ok sorry I was reading old version of the javadoc.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/results/PreContingencyResult.java
MioRtia: Do you mean a getter that returns the LimitViolation associated with a given ID?

sylvlecl: Yes, to allow the user to retrieve result for a given equipment (already done in the `PostContingencyResult` I think). Otherwise there is not point in using an underlying map !

sylvlecl (approved):

geofjamg (approved):

PR #1718, opened 2021-05-05
Correct and add SV power flows for boundaries during CGMES SV export

MioRtia:
Signed-off-by: RALAMBOTIANA MIORA <miora.ralambotiana@rte-france.com>

**Please check if the PR fulfills these requirements** *(please use `'[x]'` to check the checkboxes, or submit the PR and then click the checkboxes)*
- [x] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**Does this PR already have an issue describing the problem ?** *If so, link to this issue using `'#XXX'` and skip the rest*
No

**What kind of change does this PR introduce?** *(Bug fix, feature, docs update, ...)*
Export during CGMES SV export all the correct power flows at boundaries

**Does this PR introduce a breaking change or deprecate an API?** *If yes, check the following:*
No

PR #1745, opened 2021-05-27
Add flows of merged tie lines in SV export

MioRtia:

No

Bug fix: export flows on merged tie lines (they were not exported before since aliases were not supported by merged lines)

No

**Other information**:
Replaces #1718


annetill (approved):

colinepiloquet (approved):

marqueslanauja (approved):
I reviewed and tested your code with two cases. As first case I used the  `microGridBaseCaseAssembled` case  and in the second case I merged the `microGridBaseCaseBE` and  `microGridBaseCaseNL` cases. 
Your code is perfect and works well but I found a side issue and two side improvements.

The issue is located in the method `convertBoundaryLines `of the class `ACLineSegmentConversion` where it is used the tag `_BOUNDARY` to define the `half` aliases and it must be changed to `_Boundary`

The first improvement is located in the method `mergeProperties` of the class `NetworkImpl` where all the `danglingLine `properties are included in the `TieLine `properties with and without subIndex. In my opinion, adding the subindex it is only necessary when a property included in both danglingLines has the same `propertyName `and different value. The subindex it is not necessary when the property name is different.

The second improvement is located in the class `StateVariablesExport`. When we export the PowerFlow of shunts, staticVarCompensators and batteries , the active power flow is exported as NaN. I review the 2021-March IOPs and always is exported as zero.

Maybe we can fix the issue in this PR  (it is a minor change) and prepare an additional PR including the improvements.

MioRtia:
> The issue is located in the method `convertBoundaryLines `of the class `ACLineSegmentConversion` where it is used the tag `_BOUNDARY` to define the `half` aliases and it must be changed to `_Boundary`

Yes, you're right! I will correct it in this PR.

> The first improvement is located in the method `mergeProperties` of the class `NetworkImpl` where all the `danglingLine `properties are included in the `TieLine `properties with and without subIndex. In my opinion, adding the subindex it is only necessary when a property included in both danglingLines has the same `propertyName `and different value. The subindex it is not necessary when the property name is different.

Having the same rule for all property names can make it more understandable but I do not have any strong opinion on it. @annetill do you have any?

> The second improvement is located in the class `StateVariablesExport`. When we export the PowerFlow of shunts, staticVarCompensators and batteries , the active power flow is exported as NaN. I review the 2021-March IOPs and always is exported as zero.

Isn't the condition `if (Double.isNaN(p) && Double.isNaN(q)) { return; }` in `writePowerFlow` supposed to prevent that? But yes, it should be exported as zero, it can be done in another PR as you suggest.

PR #1748, opened 2021-05-28
New limit violation types

obrix:
Signed-off-by: Bertrand Rix <bertrand.rix@artelys.com>

A feature

Intoduce two new limit violation types related to lines : ActivePower and ApparentPower.


security-analysis-api/src/main/java/com/powsybl/security/LimitViolationBuilder.java
```diff
@@ -105,6 +105,9 @@ public LimitViolation build() {
             case LOW_SHORT_CIRCUIT_CURRENT:
             case HIGH_SHORT_CIRCUIT_CURRENT:
                 return new LimitViolation(subjectId, subjectName, type, limitName, Integer.MAX_VALUE, limit, reduction, value, null);
+            case ACTIVE_POWER:
```
annetill, line 108: This should be the same treatment as `CURRENT`.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Branch.java
```diff
@@ -210,33 +210,46 @@ default ApparentPowerLimits getApparentPowerLimits(Branch.Side side) {
...
-    Overload checkTemporaryLimits1(float limitReduction);
+    Overload checkTemporaryLimits1(float limitReduction, LimitType type);
 
-    Overload checkTemporaryLimits1();
+    Overload checkTemporaryLimits1(LimitType type);
 
-    Overload checkTemporaryLimits2(float limitReduction);
+    Overload checkTemporaryLimits2(float limitReduction, LimitType type);
 
-    Overload checkTemporaryLimits2();
```
MioRtia, line 241: Do you want it to be a breaking change or can you keep these methods as deprecated? I would leave them as deprecated and returning `CURRENT` type limits in the default implementation (deprecated since 4.3.0)

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/LimitViolationUtils.java
```diff
@@ -22,11 +24,11 @@
     private LimitViolationUtils() {
     }
 
-    public static Branch.Overload checkTemporaryLimits(Branch branch, Branch.Side side, float limitReduction, double i) {
```
MioRtia, line 25: Same remark I would keep it as deprecated

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/LimitViolationUtils.java
```diff
@@ -42,8 +44,8 @@ private LimitViolationUtils() {
         return null;
     }
 
-    public static boolean checkPermanentLimit(Branch branch, Branch.Side side, float limitReduction, double i) {
```
MioRtia, line 45: Same remark, I would keep it as deprecated

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractBranch.java
```diff
@@ -196,95 +196,109 @@ public boolean isOverloaded() {
 
     @Override
     public boolean isOverloaded(float limitReduction) {
-        return checkPermanentLimit1(limitReduction) || checkPermanentLimit2(limitReduction);
+        return checkPermanentLimit1(limitReduction, LimitType.CURRENT) || checkPermanentLimit2(limitReduction, LimitType.CURRENT);
```
MioRtia, line 199: Maybe in the `Branch` API, specify in the javadoc that this method only looks at `CURRENT` type limits. Same remark for all `Overload` methods

MioRtia (commented):
Thank you for your contribution! I have some remarks, particularly on breaking changes you made.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Branch.java
obrix: Yes I'll add them back, make them check current, and make them deprecated.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractBranch.java
obrix: In fact I was wondering about adding a type parameter for those overloading checks, but don't know if it is relevant. I'll keep it like this and update the doc.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractBranch.java
```diff
@@ -196,95 +196,169 @@ public boolean isOverloaded() {
...
+
+    public double getValueForLimit(Terminal t, LimitType type) {
+        switch (type) {
+            case ACTIVE_POWER:
+                return t.getP();
+            case APPARENT_POWER:
+                return Math.sqrt(t.getP() * t.getP() + t.getQ() * t.getQ());
+            case CURRENT:
+                return t.getI();
+            case VOLTAGE:
```
annetill, line 359: You can try t.getBusView().getBus().getV() if the bus is not null maybe...

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergedLine.java
```diff
@@ -557,4 +617,18 @@ public HalfLine getHalf(Side side) {
...
+
+    public double getValueForLimit(Terminal t, LimitType type) {
+        switch (type) {
+            case ACTIVE_POWER:
+                return t.getP();
+            case APPARENT_POWER:
+                return Math.sqrt(t.getP() * t.getP() + t.getQ() * t.getQ());
+            case CURRENT:
+                return t.getI();
+            case VOLTAGE:
```
annetill, line 629: Same proposal?

MioRtia (commented):
- Could you please add default implementation for deprecated methods **and** new methods in `Branch` API? For deprecated methods, call new methods with `CURRENT` type and for new methods, call deprecated methods when type is `CURRENT`, throw an `UnsupportedOperationException` otherwise
- Could you extract `CurrentLimitType` and rename it `LoadingLimitType`?

obrix:
@MioRtia Just to be sure, in the new methods default implementations you want to call the deprecated code ? Why ?

MioRtia:
> @MioRtia Just to be sure, in the new methods default implementations you want to call the deprecated code ? Why ?

Yes, the point is not to break the API and let time for users to update their implementation. By calling deprecated code in new methods, you allow users to use code using new methods with their old implementation if they have one.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractBranch.java
obrix: As discussed voltage is not (yet?) supported as a LimitType.

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergedLine.java
obrix: As discussed voltage is not (yet?) supported as a LimitType.

obrix:
@MioRtia I made the requested changes. A few points :

> Yes, the point is not to break the API and let time for users to update their implementation.

Giving an implementation for deprecated method that calls the new one makes sense to me, from a "no brutal breaking changes point of view". Even if in this case, no default implementation existed in Branch.

> By calling deprecated code in new methods, you allow users to use code using new methods with their old implementation if they have one

But this make less sense to me here. Creating a new default implementation for the API based on code that just been deprecated and will disappear seems hazardous to me. I can live with it though.

MioRtia (approved):

MioRtia:
> @MioRtia I made the requested changes. A few points :
> 
> > Yes, the point is not to break the API and let time for users to update their implementation.
> 
> Giving an implementation for deprecated method that calls the new one makes sense to me, from a "no brutal breaking changes point of view".
> 
> > By calling deprecated code in new methods, you allow users to use code using new methods with their old implementation if they have one
> 
> But this make less sense to me here. Creating a new default implementation for the API based on code that just been deprecated and will disappear seems hazardous to me. I can live with it though.

Just to explain a bit: it happened before (on another part of code) that people used a module using the new API for an exterior treatment (for example, a conversion) using the new methods but with an implementation that only considered the old API. For example, here, if someone use the network-store implementation (which will not be immediately up to date with the new Branch API), but use the new LimitViolationDetector, they might be in this use case (using the new API via the loadflow module with an old IIDM implementation).

We were asked to at least keep the same behavior as before. It is a far reach for a use case but since then, we made it standard to call old methods when possible (e.g. when new methods extend the old methods).

Thank you for your work nonetheless!

PR #1751, opened 2021-06-01
Fix vulnerability: disable access to external entities in XML parsing [CWE-611][CWE-827]

MioRtia:

No

Fix vulnerabilities
[CWE-611](http://cwe.mitre.org/data/definitions/611.html)
[CWE-827](http://cwe.mitre.org/data/definitions/827.html)

### DIFF
--- a/cgmes/cgmes-conversion/pom.xml
+++ b/cgmes/cgmes-conversion/pom.xml
@@ -136,14 +136,6 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
-
-        <!-- Test with all known triple store engines -->
-        <dependency>
-            <groupId>${project.groupId}</groupId>
-            <artifactId>powsybl-triple-store-impl-jena</artifactId>
-            <version>${project.version}</version>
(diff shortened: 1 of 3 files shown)


cgmes/cgmes-conversion/pom.xml
```diff
@@ -136,14 +136,6 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
-
-        <!-- Test with all known triple store engines -->
-        <dependency>
-            <groupId>${project.groupId}</groupId>
-            <artifactId>powsybl-triple-store-impl-jena</artifactId>
```
flo-dup, line 143: Not sure you remove this on purpose in this PR?

flo-dup (approved):

cgmes/cgmes-conversion/pom.xml
MioRtia: Jena library seems to include an old version of xerces that made the fix impossible: the properties to disable access to external entities in XML parsing were introduced in a more recent version

PR #1753, opened 2021-06-03
support for CGMES 3 (CIM version 100)

zamarrenolm:

**Other information**:


annetill:
In a previous meeting, I have noted some key points for the main changes we will have with version 3.0:
```
- TP confirmed as output; no Boundary TP profile. 
- Model are always built with Connectivity Nodes.
- 1 TP instance for a merged model
- MAS declaration in header required for merged SV, DL and TP.
- Header dependency revised.
- Schedules moved to CoreEquipment; Operation: only measurement classes remain.
- Separate RDFS for CoreEquipment, Operation and ShortCircuit
```
etc. I can share the slides but it is not the final version. What I understand is that bus/branch topology is not supported in version 3.0, DL and TP profils becomes outputs. Do you have planned to do that in that PR for TP profile?

zamarrenolm:
> In a previous meeting, I have noted some key points for the main changes we will have with version 3.0:
> 
> ```
> - TP confirmed as output; no Boundary TP profile. 
> - Model are always built with Connectivity Nodes.
> - 1 TP instance for a merged model
> - MAS declaration in header required for merged SV, DL and TP.
> - Header dependency revised.
> - Schedules moved to CoreEquipment; Operation: only measurement classes remain.
> - Separate RDFS for CoreEquipment, Operation and ShortCircuit
> ```
> 
> etc. I can share the slides but it is not the final version. What I understand is that bus/branch topology is not supported in version 3.0, DL and TP profils becomes outputs. Do you have planned to do that in that PR for TP profile?

I would consider the TP export in a separate PR.

With the contents of this PR we are able to successfully import the available samples for the new format.

cgmes/cgmes-model/src/main/resources/CIM100.sparql
```diff
@@ -30,3 +30,63 @@ OPTIONAL { GRAPH ?grapBDTP {
...
+        cim:IdentifiedObject.name ?name ;
+        cim:RotatingMachine.GeneratingUnit ?GeneratingUnit .
+    OPTIONAL {
+        ?SynchronousMachine cim:RotatingMachine.ratedS ?ratedS
+       }
+    ?GeneratingUnit
+        a ?generatingUnitType ;
+        cim:GeneratingUnit.minOperatingP ?minP ;
+        cim:GeneratingUnit.maxOperatingP ?maxP ;
+        cim:GeneratingUnit.nominalP ?initialP ;
```
zamarrenolm, line 48: initialP attribute has been removed from EQ in CIM100: _This required attribute is deleted as there is a general agreement that it is not expected to be able to run power flow on the EQ alone. SSH is required to be able to execute power flow._ 

So previously a required attribute, now can be dropped entirely, as initial values should come from SSH. In the conversion, a default value of zero should be set if no data from SSH is read.

cgmes/cgmes-model/src/main/resources/CIM100.sparql
```diff
@@ -30,3 +30,63 @@ OPTIONAL { GRAPH ?grapBDTP {
...
+# query: operationalLimits
+SELECT *
+WHERE
+{ GRAPH ?graph {
+    ?OperationalLimit
+        a $OperationalLimitSubclass ;
+        cim:IdentifiedObject.name ?name ;
+        cim:OperationalLimit.OperationalLimitSet ?OperationalLimitSet ;
+        cim:OperationalLimit.OperationalLimitType ?OperationalLimitType .
+    OPTIONAL { ?OperationalLimit cim:CurrentLimit.normalValue ?value }
```
zamarrenolm, line 76: In CIM100 the attribute `normalValue` comes from EQ. There is also an attribute `value` defined now in the SSH profile. We could assume that values read from SSH should overwrite values obtained from EQ.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
```diff
@@ -192,7 +192,8 @@ public boolean valid() {
 
     @Override
     public void convert() {
-        double value = p.asDouble("value");
+        double normalValue = p.asDouble("normalValue");
+        double value = p.asDouble("value", normalValue);
```
annetill, line 196: I don't understand the conversion here. The limits could be in units or in percentages. In the second case, we need the normal value to convert the limit in unit in IIDM. Where have we that conversion?

cgmes/cgmes-conformity/src/test/java/com/powsybl/cgmes/conformity/test/Cgmes3Catalog.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    public static TestGridModelResources microGrid() {
+        String base = CGMES_3_TEST_MODELS + "/MicroGrid/";
+        String baseBoundary = base;
+        return new TestGridModelResources(
+                "MicroGrid",
+                null,
+                new ResourceSet(base, "" +
+                    "20210209T1930Z_1D_ASSEMBLED_DL_9.xml",
```
annetill, line 28: Are you sure that as inputs `20210209T1930Z_1D_ASSEMBLED_DL_9`, `20210209T1930Z_1D_ASSEMBLED_SV_9`, `20210209T2323Z_1D_ASSEMBLED_TP_9` should be imported as the other files? It should be outputs.

cgmes/cgmes-conformity/src/test/java/com/powsybl/cgmes/conformity/test/Cgmes3Catalog.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    public static TestGridModelResources miniGrid() {
+        String base = CGMES_3_TEST_MODELS + "/MiniGrid/";
+        String baseBoundary = base;
+        return new TestGridModelResources(
+                "MiniGrid",
+                null,
+                new ResourceSet(base, "" +
+                    "20210202T1930Z_1D_ASSEMBLED_DL_7.xml",
```
annetill, line 47: Same remark as before: `ASSEMBLED` files should be removed for me, and the `TP` file should be treated as an output too.

cgmes/cgmes-conformity/src/test/java/com/powsybl/cgmes/conformity/test/Cgmes3Catalog.java
```diff
@@ -0,0 +1,85 @@
...
+                new ResourceSet(baseBoundary, "MiniGridTestConfiguration_EQ_BD_v3.0.0.xml"));
+    }
+
+    public static TestGridModelResources smallGrid() {
+        String base = CGMES_3_TEST_MODELS + "/SmallGrid/";
+        String baseBoundary = base;
+        return new TestGridModelResources(
+                "SmallGrid",
+                null,
+                new ResourceSet(base, "" +
```
annetill, line 61: Same remark for `TP` file. It is what I have understand from the CGMES 3.0: it is becoming an output. Maybe to be done in another PR. If so, put please a FIXME and a comment to remember.

cgmes/cgmes-conformity/src/test/java/com/powsybl/cgmes/conformity/test/Cgmes3Catalog.java
```diff
@@ -0,0 +1,85 @@
...
+        String base = CGMES_3_TEST_MODELS + "/Svedala/";
+        return new TestGridModelResources(
+                "Svedala",
+                null,
+                new ResourceSet(base, "" +
+                    "20201202T1843Z_1D_Svedala Area_DL_001.xml",
+                    "20201202T1843Z_1D_Svedala Area_SV_001.xml",
+                    "20201202T1843Z_1D_Svedala Area_EQ_001.xml",
+                    "20201202T1843Z_1D_Svedala Area_SSH_001.xml",
+                    "20201202T1843Z_1D_Svedala Area_TP_001.xml"));
```
annetill, line 81: Same remark.

cgmes/cgmes-conformity/src/test/resources/cgmes3-test-models/MicroGrid/20171002T0930Z_ENTSO-E_EQ_BD_2.xml
```diff
@@ -0,0 +1,269 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<rdf:RDF xmlns:cim="http://iec.ch/TC57/CIM100#" xmlns:eu="http://iec.ch/TC57/CIM100-European#" xmlns:md="http://iec.ch/TC57/61970-552/ModelDescription/1#" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#">
```
annetill, line 2: This file is going to be deprecated. It means that all the infos should be in the `EQ` file. Is it true for the test files?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
zamarrenolm: In CGMES 3 values for limits have been moved from EQ to SSH. EQ now contains `normalValue`, that should be used when no explicit `value` is defined in SSH. That is, `normalValue` in EQ is the default value for the limit if a value is not explicitly given in the SSH.

For Operational Limits, all the limit values and normal values are given in the corresponding engineering units, according to the specification. I think is for Measurements that there is the possibility of providing values in defined units or as percentages of normal values.

zamarrenolm:
Net unit tests have been added that ensure CGMES 3 networks are properly imported even when the TP, SV files are not provided.

annetill (approved):
Thanks a lot for this relevant work.

PR #1755, opened 2021-06-06
TieLine. Fix issues without modifying the boundary interface

marqueslanauja:

Bug fix, feature.

`TieLine` parameters are calculated by adding `halfLine` parameters.

As the injection is zero at the boundary node, it can be removed by kron reduction getting an equivalent branch between the network model ends.
`Line `and `danglingLine `parameters are reoriented before mapping to the halfLines if it is necessary.

**Other information**:

See
[RTE-AIA 2022 - Tie Lines.V0.pptx](https://github.com/powsybl/powsybl-core/files/8322149/RTE-AIA.2022.-.Tie.Lines.V0.pptx)


miovd (changes requested):
I agree with your PR and your modifications, thank you!
I think you should also make these changes in the `MergedLine` implementation of the `TieLine` (in `iidm-mergingview`) which is also a merged view of two dangling lines.

Note: a correction in `network-store` should also be done

marqueslanauja:
> I agree with your PR and your modifications, thank you! I think you should also make these changes in the `MergedLine` implementation of the `TieLine` (in `iidm-mergingview`) which is also a merged view of two dangling lines.
> 
> Note: a correction in `network-store` should also be done

I made the changes in the `MergedLine `implementation. 
In this class I changed the behaviour of the following methods: `setR`, `setX`, `setG1`, `setB1`, `setG2` and `setB2` throwing an exception and not allowing to directly modify the characteristics of the mergedLine as before.

In the HalfLineAdapter class I kept the current implementation of the `setG1`, `setB1`, `setG2`, `setB2` methods although could cause unwanted effects as the danglingLine only has `G `and `B` attributes and does not allow to record `G1`, `B1`, `G2`, `B2 `separately.

I will take a look at the network-store repository.

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/HalfLineAdapter.java
```diff
@@ -32,14 +37,29 @@
...
+    }
+
+    public HalfLineAdapter(DanglingLine danglingLine, Branch.Side side, MergingViewIndex index, boolean mustBeReoriented) {
+
+        // TODO(mathbagu): is it correct? should we initialize only b1/g1 (resp. b2/g2) like it's done in the destructive merge?
+
+        // if mustBeReoriented is false then end1 and end2 of the halfLine correspond to end1 and end2 of the associated danglingLine
+        // if mustBeReoriented is true then end1 and end2 of the halfLine correspond to end2 and end1 of the associated danglingLine
+        // As in the danglingLine only G (g1 + g2) is recorded we do not need to know if the danglingLine has been reoriented
+        BranchReorientedParameters brp = new BranchReorientedParameters(danglingLine.getR(), danglingLine.getX(),
```
annetill, line 50: I am not able to understand why we need this object here because the attributes are symmetrical. Why it is different that what `main` is doing?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/HalfLineUtil.java
```diff
@@ -0,0 +1,50 @@
...
+import com.powsybl.iidm.network.Branch;
+import com.powsybl.iidm.network.TieLine;
+import com.powsybl.iidm.network.util.LinkData.BranchAdmittanceMatrix;
+
+/**
+ * @author Luma Zamarreño <zamarrenolm at aia.es>
+ * @author José Antonio Marqués <marquesja at aia.es>
+ */
+
+public final class HalfLineUtil {
```
annetill, line 20: I am not sure the name of this class is relevant. Is it not more related to tie line?

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/HalfLineAdapter.java
annetill: It is different that what you are doing in the `NetworkImpl`.

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/MergedLine.java
```diff
@@ -216,70 +222,70 @@ MergedLine setId(String id) {
 
     @Override
     public double getR() {
-        return half1.getR() + half2.getR();
+        LinkData.BranchAdmittanceMatrix adm = HalfLineUtil.equivalentBranchAdmittanceMatrix(half1, half2);
+        // Add 0.0 to avoid negative zero, tests where the R value is compared as text, fail
+        return adm.y12().negate().reciprocal().getReal() + 0.0;
```
annetill, line 227: @miovd what do you think about this `+0.0`.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/TieLineImpl.java
```diff
@@ -229,9 +234,12 @@ public HalfLineImpl getHalf(Side side) {
         }
     }
 
+    // Half1 and half2 are lines, so the transmission impedance of the equivalent branch is symmetric
     @Override
     public double getR() {
-        return half1.getR() + half2.getR();
+        LinkData.BranchAdmittanceMatrix adm = HalfLineUtil.equivalentBranchAdmittanceMatrix(half1, half2);
```
annetill, line 240: @miovd and @zamarrenolm these calculations are duplicated in each implementations and quite complex to maintain for non expert people. Is there a way to put that in the `TieLine` API?

annetill (commented):
Just some comments to start.

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/HalfLineAdapter.java
marqueslanauja: You are right, it is the same. I decided to use the `BranchReorientedParameters` method to record that at this point if the shunt admittances are asymmetrical it is necessary to reorient the parameters. That can happen, in the future, if we extend the dangling line model by adding shunt admittances at both ends or if we decide not to split the shunt admittance.

In the NetworkImpl the shunt admittance of the danglingLine is not split so in this case the parameters must be reoriented. I do not know why the shunt admittance is not managed in the same way in both cases. These inconsistencies will disappear if the dangling line model is extended to support shunt admittance at both ends.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ACLineSegmentConversion.java
```diff
@@ -151,31 +152,40 @@ private void convertLineAtBoundary(int boundarySide) {
         }
     }
 
-    // TODO support transformer + Line
+    // Before creating the TieLine the initial branches are reoriented if it is necessary,
```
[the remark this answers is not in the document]
marqueslanauja: The parameters of the branches (shunt admittances) should be in concordance with the ends of the branch (could be `Y1 != Y2`). In the TieLine   we always have the `NetworkSide end1 - boundary - NetworSide end2` configuration. So the `end1` of the first half line must be the network side and the `end2` the boundary side. In the second half line, just the opposite, `end1` must be the boundary side and `end2` must be the network side.

If the branch mapped to the first half line has in Cgmes the boundary side at the `end1` must be reoriented. Same happens if the branch mapped to the second half line has in Cgmes the network side at the `end1`.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/HalfLineUtil.java
marqueslanauja: Yes, better TieLineUtil as you implemented in the commit.

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/HalfLineAdapter.java
annetill: Why are we making things different in the merging view implementation and in the in-memory implementation? Maybe we don't have to split B and G of the dangling line here, in the merging view and keep the reorientation.

annetill: @miovd and @zamarrenolm too.
What I see is that we have two different behaviors depending on the implementation. In `NetworkImpl`, we have 
`BranchReorientedParameters brp2 = new BranchReorientedParameters(dl2.getR(), dl2.getX(), dl2.getG(), dl2.getB(), 0.0, 0.0);`
In the merging view, and more especially in the `HalfLineAdapter`, we have:
`BranchReorientedParameters brp = new BranchReorientedParameters(danglingLine.getR(), danglingLine.getX(),
 danglingLine.getG() / 2, danglingLine.getB() / 2, danglingLine.getG() / 2, danglingLine.getB() / 2, mustBeReoriented);`
We have two different behaviors, right? And the second one, in the merging view is a bi strange because in that case reorientation means nothing. I really prefer that we also fix that issue in that PR. Tie lines are a subject that is quite hard to maintain, I would like to make it as simple and clear as possible.

marqueslanauja: Done. I fixed the second as the first one.

zamarrenolm (approved):

annetill (approved):

miovd (approved):

PR #1757, opened 2021-06-07
Plugins series converter

yichen88:

**Other information**:

### DIFF
--- a/commons/src/main/java/com/powsybl/commons/extensions/ExtensionSeriesSerializer.java
+++ b/commons/src/main/java/com/powsybl/commons/extensions/ExtensionSeriesSerializer.java
@@ -1,0 +1,56 @@
+/**
+ * Copyright (c) 2021, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.commons.extensions;
+
+import java.util.Map;
+
+/**
+ * @author Yichen TANG <yichen.tang at rte-france.com>
+ */
+public interface ExtensionSeriesSerializer<T extends Extendable, E extends Extension<T>> extends ExtensionProvider<T, E> {
+
+    int STRING_SERIES_TYPE = 0;
+    int DOUBLE_SERIES_TYPE = 1;
+    int INT_SERIES_TYPE = 2;
+    int BOOLEAN_SERIES_TYPE = 3;
+
+    /**
+     * @param builder the extensions holder to write into
+     */
+    void serialize(ExtensionSeriesBuilder<?, T> builder);
+
+    /**
+     * Update extension double value
+     * @param element the element has the extension to update
+     * @param name field name of extension
+     * @param value
+     */
+    void deserialize(T element, String name, double value);
+
+    /**
+     * Update extension int value
+     * @param element the element has the extension to update
+     * @param name field name of extension
+     * @param value
+     */
+    void deserialize(T element, String name, int value);
+
+    /**
+     * Update extension string value
+     * @param element the element has the extension to update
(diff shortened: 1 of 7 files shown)


commons/src/main/java/com/powsybl/commons/extensions/ExtensionSeriesSerializer.java
```diff
@@ -0,0 +1,56 @@
...
+package com.powsybl.commons.extensions;
+
+import java.util.Map;
+
+/**
+ * @author Yichen TANG <yichen.tang at rte-france.com>
+ */
+public interface ExtensionSeriesSerializer<T extends Extendable, E extends Extension<T>> extends ExtensionProvider<T, E> {
+
+    int STRING_SERIES_TYPE = 0;
```
sylvlecl, line 16: I think here we should make this an enum:
we'll convert to an `int` in pypowsybl

commons/src/main/java/com/powsybl/commons/extensions/ExtensionSeriesSerializer.java
```diff
@@ -0,0 +1,56 @@
...
+    int DOUBLE_SERIES_TYPE = 1;
+    int INT_SERIES_TYPE = 2;
+    int BOOLEAN_SERIES_TYPE = 3;
+
+    /**
+     * @param builder the extensions holder to write into
+     */
+    void serialize(ExtensionSeriesBuilder<?, T> builder);
```
sylvlecl, line 24: maybe rename to `writeSeries` ?

commons/src/main/java/com/powsybl/commons/extensions/ExtensionSeriesSerializer.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Update extension double value
+     * @param element the element has the extension to update
+     * @param name field name of extension
+     * @param value
+     */
+    void deserialize(T element, String name, double value);
```
sylvlecl, line 32: maybe rename to `readSeries` ?

commons/src/test/java/com/powsybl/commons/extensions/ExtensionSeriesSerializerTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+        // add extension
+        extensionSeriesSerializer.deserialize(f2, "fooExt_value", 1);
+        assertEquals(new FooExt(true, ""), f2.getExtension(FooExt.class));
+        extensionSeriesSerializer.deserialize(f1, "fooExt_value", 1);
+        extensionSeriesSerializer.deserialize(f1, "fooExt_value2", "bar");
+        assertEquals(new FooExt(true, "bar"), f1.getExtension(FooExt.class));
+    }
+
+    class MockSeriesBuilder implements ExtensionSeriesBuilder<MockSeriesBuilder, Foo> {
```
sylvlecl, line 55: Maybe we can provide a real implementation that writes to lists: we could re-use it for unit test in pypowsybl ?

commons/src/main/java/com/powsybl/commons/extensions/ExtensionSeriesSerializer.java
```diff
@@ -0,0 +1,56 @@
...
+     * @param name field name of extension
+     * @param value
+     */
+    void deserialize(T element, String name, String value);
+
+    /**
+     * extension filed name as key, and value type as value
+     * @return
+     */
+    Map<String, Integer> getTypeMap();
```
sylvlecl, line 54: Maybe we would provide more metadata in the future about those series (like is it variant based, is it writable ...):
I suggest to return somethink like `List<SeriesMetadata>` instead where for now SeriesMetadata would just be name + type.
(plus, this way the "columns" are explicitly ordered : we should respect the same order here and in writeSeries)

sylvlecl (commented):
Looks very good, some renaming and small improvements suggestions.

PR #1759, opened 2021-06-08
First proposition of extensions for measurements

MioRtia:

No

Introduce extensions to model measurements (analog and discrete).
It is mainly based off CGMES. Maybe other standards attributes can be added.
Specific attributes can be added as properties of these extensions.

No

**Other information**:


geofjamg:
First very raw thought: I don't understand the name Analog and discrete. There should be somewhere in the name "Measure".

MioRtia:
> First very raw thought: I don't understand the name Analog and Discrete. There should be somewhere the name "Measure".

Yes, I think the complete names are `AnalogMeasurement(s)` and `DiscreteMeasurement(s)`, it's probably more understandable...

annetill:
First of all, thanks for your work. I have some points to discuss:
- I understand that you prefer two extensions and not just one. For `Discretes`, I understand that it is an extension of identifiables (needed for `Switch`, `TapChanger` and maybe `Section`) but `Analogs` extends connectables, that is why you prefer two extensions. My concern is about `Discretes`: maybe it is too permissive and should be more specific. Furthermore, `Analog` name choice is maybe not relevant for all measures.
- For analog measurements, you must introduce the concept of how we trust that value with standard deviation, or its inverse. It is what Agnès calls `weight` and it is an input of the state estimation. It should be explicit, not in a property because all measure come with its error. It is meaningful for discrete measurements? 
-  For all types of measures, both discrete and analog, you need also a enum that describe validity. It could be a boolean, but we will lost information. The validity could come from operators, but also directly from the network monitoring with a technical problem. We all have to think about this enum. This enum will lead at least to 2 categories: valid and invalid for the state estimation.
- For discrete measurements, I think that we have to think about a type that says if it is measured or estimated/calculated because we have some models that could do that. It means that outputs (tap positions for instance) will be put in the extension after the state estimation. It is a problem? 
- For analog types, I think that we have to start with common data (not on three phase for instance), with the vocabulary shared in powsybl-core.
It is just my personal understanding of our work and state estimation. I know that you want also these extensions to be used after Scada system. So maybe, we have to organize a dedicate meeting with @sylvlecl and other interested developpers.

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/Discrete.java
```diff
@@ -0,0 +1,31 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.iidm.network.extensions;
+
+/**
+ * @author Miora Ralambotiana <miora.ralambotiana at rte-france.com>
+ */
+public interface Discrete {
+
+    enum Type {
```
annetill, line 14: `SHUNT_COMPENSATOR_SECTION` should be added too.

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/Analog.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+package com.powsybl.iidm.network.extensions;
+
+/**
+ * @author Miora Ralambotiana <miora.ralambotiana at rte-france.com>
+ */
+public interface Analog {
+
+    enum Side {
+        ONE,
```
annetill, line 15: I think you need a `NONE` side for loads, generators, etc.

MioRtia:
@annetill Thank you for your feedback! A few questions/answers before I carry on with this.

> First of all, thanks for your work. I have some points to discuss:
> 
>     * I understand that you prefer two extensions and not just one. For `Discretes`, I understand that it is an extension of identifiables (needed for `Switch`, `TapChanger` and maybe `Section`) but `Analogs` extends connectables, that is why you prefer two extensions. My concern is about `Discretes`: maybe it is too permissive and should be more specific. Furthermore, `Analog` name choice is maybe not relevant for all measures.

About Discrete measurements: I agree but it might not be possible to be less permissive if we want a same object for the different discrete measurements. I'm not sure having a `SwitchDiscreteMeasurement`, `TapChangerDiscreteMeasurement` and so on is the good solution either, what do you think? I find it a bit too confusing but maybe this is just me.

About Analog measurements: I'm open to name ideas! I used CGMES terminology but there might be a more intuitive one.

>     * For analog measurements, you must introduce the concept of how we trust that value with standard deviation, or its inverse. It is what Agnès calls `weight` and it is an input of the state estimation. It should be explicit, not in a property because all measure come with its error. It is meaningful for discrete measurements?

Okay, does it have a default value? Indeed, it is not a value present in CGMES so we have to invent this value. Or should it be optional?

>     * For all types of measures, both discrete and analog, you need also a enum that describe validity. It could be a boolean, but we will lost information. The validity could come from operators, but also directly from the network monitoring with a technical problem. We all have to think about this enum. This enum will lead at least to 2 categories: valid and invalid for the state estimation.

Okay, so it will not be the same as what Agnès did for vadility, how is it mapped? Agnès had 4 values for validity (or this must be specified in properties?)


>     * For discrete measurements, I think that we have to think about a type that says if it is measured or estimated/calculated because we have some models that could do that. It means that outputs (tap positions for instance) will be put in the extension after the state estimation. It is a problem?

Agreed, I don't think it will be a problem.

>     * For analog types, I think that we have to start with common data (not on three phase for instance), with the vocabulary shared in powsybl-core.

Okay!

>       It is just my personal understanding of our work and state estimation. I know that you want also these extensions to be used after Scada system. So maybe, we have to organize a dedicate meeting with @sylvlecl and other interested developpers.

Yes, I am also interested in how to incorporate observability in this.

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/Analog.java
```diff
@@ -0,0 +1,42 @@
...
+        TWO,
+        THREE
+    }
+
+    enum Type {
+        THREE_PHASE_POWER,
```
annetill, line 21: I don't get these types. Can you start with `ACTIVE_POWER`, `REACTIVE_POWER`, `CURRENT`. For `VOLTAGE` and `PHASE`, it means that it will be attached to a connectable. It is what we have from CGMES conversion too? Are we obliged to extend a connectable? Just a suggestion: could it be possible to have just an extension of network or voltage level?

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/Analog.java
MioRtia: Is it not enough if it is `null` in this case?

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/Analog.java
MioRtia: I honestly have no idea, but those are the standard types in CGMES. We can not keep them but in the CGMES conversion, we will have to think about how to translate those types in powsybl's type.

annetill:
After our physical meeting:
- `Measurement` extending connectable and `DiscreteMeasurement` extending identifiable ;
- In a first time, we support the types required for a state estimation ;
- Validity: just a boolean ; the rule is to rely on properties for specificity ; `isValid` and a setter needed I imagine for interface user ;
- We add a field `StandardDeviation` to define the uncertainty around the measurement ;
- We have decided to create an other extension to define what should be estimated by the state estimation (`rho` for ratio tap changer for example). It will extend the two and three windings transformers in a first approach. This has to be done in an other PR ;
- Observability that will resume the quality of the results of the state estimation: we have decided to add extensions where we can have injections (P and Q). For V, a first good approach could be to extend bus bar sections but with the though that we could use the concept of bus ref. These extensions will contain a boolean that says if it is observable or not and standard deviations for estimated values. This has to be done in another PR.

sylvlecl (commented):
OK in general, some discussions about implementation details

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/DiscreteMeasurement.java
```diff
@@ -0,0 +1,63 @@
...
+
+    DiscreteMeasurement putProperty(String name, String property);
+
+    DiscreteMeasurement removeProperty(String name);
+
+    String getValueAsString();
+
+    int getValueAsInt();
+
+    DiscreteMeasurement setValue(String valueAsString, int valueAsInt);
```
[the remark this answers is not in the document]
MioRtia: You're right, it would probably be clearer.

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/DiscreteMeasurement.java
```diff
@@ -0,0 +1,63 @@
...
+        PHASE_TAP_CHANGER,
+        RATIO_TAP_CHANGER_1,
+        PHASE_TAP_CHANGER_1,
+        RATIO_TAP_CHANGER_2,
+        PHASE_TAP_CHANGER_2,
+        RATIO_TAP_CHANGER_3,
+        PHASE_TAP_CHANGER_3
+    }
+
+    String getId();
```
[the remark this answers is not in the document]
MioRtia: I don't think we should enforce a non null ID, I only ever saw identified measurement in CGMES, it is not mandatory in other formats and it would be weird to invent some for IIDM

geofjamg (approved):

PR #1765, opened 2021-06-13
Fix IEEE CDF nominal voltage import

geofjamg:

https://github.com/powsybl/pypowsybl/issues/120

**Other information**:


MioRtia (approved):

ieee-cdf/ieee-cdf-converter/src/main/java/com/powsybl/ieeecdf/converter/IeeeCdfNetworkFactory.java
```diff
@@ -32,7 +46,19 @@ public static Network create9() {
     }
 
     public static Network create14(NetworkFactory networkFactory) {
-        return create("ieee14cdf", networkFactory);
+        return create("ieee14cdf", networkFactory, null, ieeeCdfBus -> {
```
annetill, line 49: Maybe here we have to add comment saying that there is no convention values for this case so it is assumptions.

annetill (approved):

ieee-cdf/ieee-cdf-converter/src/main/java/com/powsybl/ieeecdf/converter/IeeeCdfNetworkFactory.java
MioRtia: I added some comments if they are okay with you.

PR #1776, opened 2021-06-24
BranchResult : Add double attribute to handle flowTransfer data

obrix:

Add a flow transfer data to the branch result to support line outage distributation factors.

**Other information**:


security-analysis/security-analysis-api/src/main/java/com/powsybl/security/results/BranchResult.java
```diff
@@ -34,17 +34,20 @@
 
     private final double i2;
 
+    private final double flowTransfer;
```
annetill, line 37: What is the definition of flow transfer? Javadoc is missing and just with the code, it is not explicit enough. First question: do you need one field or two ? In AC security analysis, there are two flows `p1` and `p2`. What is the definition of flow transfer for pre-contingency situations and for post-contingency situations when you lost more than one line? Other remark: I think that flow transfer must be optional in branch result, is it what you have done ?

obrix: This is work in progress (I just added the flag sorry).

To anwser your questions, from my comprehension from the dicussions I add with you and @geofjamg :

 > What is the definition of flow transfer? 
 
 The flow percentage from a branch contingency that is added to another branch.

> Javadoc is missing and just with the code

Yes once we know exactly what we are doing I will document it
 
 > do you need one field or two ?
 
 After discussion with @geofjamg it seems a single field is sufficient
 
 > In AC security analysis, there are two flows p1 and p2
 
 Is that relative to a side ? If yes then this mean we actually compute two percentage, one for each side ? This is not clear for me, could you clarify what you have in mind ? Maybe we should discuss it again.
 
> I think that flow transfer must be optional in branch result, is it what you have done ?

No I try to make something with the information I got, if this must be optional in the end it will be.

annetill: > This is work in progress (I just added the flag sorry).
> 
> To anwser your questions, from my comprehension from the dicussions I add with you and @geofjamg :
> 
> > What is the definition of flow transfer?
> 
> The flow percentage from a branch contingency that is added to another branch.
> 
> > Javadoc is missing and just with the code
> 
> Yes once we know exactly what we are doing I will document it
> 
> > do you need one field or two ?
> 
> After discussion with @geofjamg it seems a single field is sufficient
> 
> > In AC security analysis, there are two flows p1 and p2
> 
> Is that relative to a side ? If yes then this mean we actually compute two percentage, one for each side ? This is not clear for me, could you clarify what you have in mind ? Maybe we should discuss it again.
> 
> > I think that flow transfer must be optional in branch result, is it what you have done ?
> 
> No I try to make something with the information I got, if this must be optional in the end it will be.

So maybe a good start is:
- To focus on side 1, either for monitored branch than for contingency branch ; for me it could be better to call the field explicitly `flowTransfer1`.
- To try to make that field optional.
What do you think?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/results/BranchResult.java
```diff
@@ -34,17 +34,27 @@
 
     private final double i2;
 
+    /**
+     * Flow transfer from the branch in contingency to the branch with id branchId.
```
annetill, line 38: Can you define the flow transfer in a different way such as: 
Note p1i,N the active power flow on side 1 of the branch branchId, and p1i, N-1 the active power flow on side 1 of the same branch. Note p1j,N the active power flow on side 1 of a branch j. After the lost of this branch, we have:
p1,N-1 = P1, N + flow transfer(j->i) * p1j,N

obrix: For the last part did you rather mean p1i,N = P1i, N-1 + flow transfer(j->i) * p1j,N-1 ?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/results/BranchResult.java
```diff
@@ -34,17 +34,30 @@
...
+     * <p>
+     * It is a ratio computed as : <i>p1i,N-1 - p1i,N / p1j,N</i>, where :
+     * <b>p1i,N</b> the active power flow on side 1 of the branch branchId at pre contingency stage.
+     * <b>p1i,N-1</b> the active power flow on side 1 of the same branch at post contingency stage.
+     * <b>p1j,N</b> the active power flow on side 1 of lost branch j at pre contingency stage.
+     * <b>p1j,N-1</b> the active power flow on side 1 of lost branch j at post contingency stage.
```
annetill, line 44: You can remove p1j,N-1 as you don't use it indeed.

MioRtia (approved):

PR #1779, opened 2021-06-29
Extensions to specify if tap changers should be estimated during state estimation or not

MioRtia:

No

Feature

No way to tell if a tap changer position is an input or output of a state estimation

Two extensions are created to model this

No


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/ThreeWindingsTransformerToBeEstimatedAdderImpl.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * @author Miora Ralambotiana <miora.ralambotiana at rte-france.com>
+ */
+class ThreeWindingsTransformerToBeEstimatedAdderImpl extends AbstractExtensionAdder<ThreeWindingsTransformer, ThreeWindingsTransformerToBeEstimated>
+        implements ThreeWindingsTransformerToBeEstimatedAdder {
+
+    private final Set<ThreeWindingsTransformerToBeEstimated.TapChanger> tapChangers = new HashSet<>();
```
sylvlecl, line 24: I think we can use `EnumSet ` here and in other classes

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/TwoWindingsTransformerToBeEstimated.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Override
+    default String getName() {
+        return "twoWindingsTransformerToBeEstimated";
+    }
+
+    /**
+     * Get tap changers that are to be estimated during a state estimation.
+     */
+    Set<TapChanger> getTapChangers();
```
sylvlecl, line 38: No strong opinion here, but I wonder if it would not have been more natural with plain booleans (2 here and 6 for the 3 windings).

No problem to keep it this way if it's OK for everyone.

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/ThreeWindingsTransformerToBeEstimated.java
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Remove a given tap changer to the list of tap changers to be estimated during state estimation.
+     * If this tap changer is not considered to be estimated, do nothing.
+     */
+    ThreeWindingsTransformerToBeEstimated removeTapChanger(TapChanger tapChanger);
+
+    /**
+     * Remove the extension if no tap changer is specified to be estimated.
+     */
+    default void cleanIfEmpty() {
```
sylvlecl, line 65: maybe `removeIfEmpty` will be more exact

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/ThreeWindingsTransformerToBeEstimated.java
```diff
@@ -0,0 +1,70 @@
...
+ * @author Miora Ralambotiana <miora.ralambotiana at rte-france.com>
+ */
+public interface ThreeWindingsTransformerToBeEstimated extends Extension<ThreeWindingsTransformer> {
+
+    /**
+     * A three windings transformer has three legs which can each have a ratio tap changer and/or a phase tap changer
+     * that can be estimated during a state estimation.
+     */
+    enum TapChanger {
+        RATIO_TAP_CHANGER_1,
```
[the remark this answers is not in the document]
sylvlecl: - phase shifters : yes (maybe not all implementations support it, but it's theoretically possible)
- about 3 windings: I think in theory if you have the 3 voltages, you have 2 independent ratios, so you can estimate 2 of the 3 taps. So it seems it makes sense to have all possibilities

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/ThreeWindingsTransformerToBeEstimated.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.util.Set;
+
+/**
+ * Indicate for a state estimation which tap changers are to be estimated (i.e. their tap positions should be outputs).
+ * If a tap changer is not to be estimated, it should not be changed during a state estimation (i.e its tap position is only an input).
+ *
+ * @author Miora Ralambotiana <miora.ralambotiana at rte-france.com>
+ */
+public interface ThreeWindingsTransformerToBeEstimated extends Extension<ThreeWindingsTransformer> {
```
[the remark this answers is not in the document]
sylvlecl: We need to be careful to separate inputs and outputs: even if that parameter is `true`, it does not guarantee that we have actually estimated the tap : maybe the state estimation was run with a parameter to not estimated taps for example, or it was not able to estimate the tap ...

So the `ToBe` makes clear that it's an input for the state estimation computation which seems ok. If we remove it I think we should have 2 properties, one for input and one for output.
But maybe the output one should go into the broader "quality" extensions that would contain informations about observability, standard deviation and so on ?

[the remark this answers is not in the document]
annetill: So it means that for you, we need 2 extensions, one for the inputs to the state estimation and one for the outputs to the state estimation ? It is really want we want: multiply the number of extensions ?

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/TwoWindingsTransformerToBeEstimated.java
MioRtia: I actually agree with you, I think it would be more simple with booleans...

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/ThreeWindingsTransformerToBeEstimated.java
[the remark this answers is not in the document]
sylvlecl: No I agree we should not have too many, that's why if we want to have that information I would propose the 2 alternatives:
 * having 2 fields in that extension (1 for input 1 for output)
 * preferred: keep only input in that one, and put the output in the other extension which will carry more information about state estimation results (observability, standard dev of the estimation ...). Actually, I think that `observable` and `estimated` are the same thing here, so maybe we should not even worry about adding that information after all.

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/ThreeWindingsTransformerToBeEstimated.java
MioRtia: I deleted the method, not really useful with booleans

annetill:
For phase tap changers, it seems that we need, for state estimation, to mention the regulation mode (`CURRENT_LIMITER`, `ACTIVE_POWER` or `OFF` ?). I don't know if it is really used, but do you think that we need property for that?

MioRtia:
> For phase tap changers, it seems that we need, for state estimation, to mention the regulation mode (`CURRENT_LIMITER`, `ACTIVE_POWER` or `OFF` ?). I don't know if it is really used, but do you think that we need property for that?

We can discuss if it is standard, but if it is, we can add it on `DiscreteMeasurements`. In the meantime, properties can be used.

annetill:
> > For phase tap changers, it seems that we need, for state estimation, to mention the regulation mode (`CURRENT_LIMITER`, `ACTIVE_POWER` or `OFF` ?). I don't know if it is really used, but do you think that we need property for that?
> 
> We can discuss if it is standard, but if it is, we can add it on `DiscreteMeasurement.Type`. In the meantime, properties can be used.

Indeed what I don't understand is that we have the regulation mode in IIDM, so for me we don't store anything relating to that in these extensions for the moment.

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/ThreeWindingsTransformerToBeEstimated.java
```diff
@@ -0,0 +1,113 @@
...
+     * Else, return false.
+     */
+    boolean containsPhaseTapChanger3();
+
+
+    /**
+     * Return true if the phase tap changer of the leg of the given side of the extended three windings transformer is to be estimated during a state estimation.
+     * Else, return false.
+     */
+    boolean containsPhaseTapChanger(ThreeWindingsTransformer.Side side);
```
annetill, line 72: This method is enough for the need, why do you prefer to add `containsPhaseTapChanger1``, `containsPhaseTapChanger2`, etc...

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/ThreeWindingsTransformerToBeEstimated.java
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     * Specify if the ratio tap changer of the leg 1 of the extended three windings transformer is to be estimated during a state estimation.
+     */
+    ThreeWindingsTransformerToBeEstimated setRatioTapChanger1Status(boolean toBeEstimated);
+
+    /**
+     * Specify if the ratio tap changer of the leg 2 of the extended three windings transformer is to be estimated during a state estimation.
+     */
+    ThreeWindingsTransformerToBeEstimated setRatioTapChanger2Status(boolean toBeEstimated);
```
annetill, line 82: Same remark here: we cannot have just less methods that takes the side in argument?

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/TwoWindingsTransformerToBeEstimated.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @Override
+    default String getName() {
+        return "twoWindingsTransformerToBeEstimated";
+    }
+
+    /**
+     * Return true if the ratio tap changer of the extended two windings transformer is to be estimated during a state estimation. Else, return false.
+     */
+    boolean containsRatioTapChanger();
```
sylvlecl, line 28: maybe we could rephrase those methods with something more explicit like `shouldEstimateRatioTapChanger` ?

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/TwoWindingsTransformerToBeEstimated.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * Return true if the phase tap changer of the extended two windings transformer is to be estimated during a state estimation. Else, return false.
+     */
+    boolean containsPhaseTapChanger();
+
+    /**
+     * Specify if the ratio tap changer of the extended two windings transformer is to be estimated during a state estimation.
+     */
+    TwoWindingsTransformerToBeEstimated setRatioTapChangerStatus(boolean toBeEstimated);
```
sylvlecl, line 38: same kind of suggestion, I think "status" is not very well suited (it's not really a state of the tap changer, it's more a parameter). Maybe we could rename to something like `shouldEstimateRatioTapChanger` too ?

sylvlecl (approved):
Some naming propositions, otherwise everything's OK.

annetill (approved):

PR #1780, opened 2021-06-29
Allow Xnodes connected to more than 2 Nodes to be imported correctly.

murgeyseb:

No

Bug fix

Currently, Xnodes connected to more than 2 other nodes in a UCTE network lead to an exception in UCTE importer. But in some cases, only two connection are actually closed, the third one is open.

When Xnodes are connected to more than 2 other nodes, and if two of the associated branches are open, the file s considered as valid and the tie-line is created. If less than two are connected, all dangling lines are kept in final network.


MioRtia (commented):
Just a thought: is it possible to have a unique matching dangling line (```matchingDanglingLines.size() == 1```) that is **not** connected?

If yes, this solution is good to me.

murgeyseb:
> Just a thought: is it possible to have a unique matching dangling line (`matchingDanglingLines.size() == 1`) that is **not** connected?
> 
> If yes, this solution is good to me.

Yes it is possible if it is a standard tie line (connected to to nodes) but open on one side. I decided to model it as an open TieLine and not two DanglingLine.

MioRtia (approved):

PR #1783, opened 2021-07-05
Reporter POC: fix deserialization with no specified dictionary

flo-dup:

No

Bug fix

Dictionary name needs to be specified through injectable value, fallback to default name fails

    - If dictionary name missing falls back to "default" name
    - If dictionary name specified is not in the "dics" root entry, falls back to first entry of dics and warns the user
    - if "dics" root entry is missing or empty, warns the user

No

### DIFF
--- a/commons/src/main/java/com/powsybl/commons/reporter/ReporterModelDeserializer.java
+++ b/commons/src/main/java/com/powsybl/commons/reporter/ReporterModelDeserializer.java
@@ -29,6 +29,7 @@ import java.util.Objects;
 public class ReporterModelDeserializer extends StdDeserializer<ReporterModel> {
 
     public static final String DICTIONARY_VALUE_ID = "dictionary";
+    public static final String DICTIONARY_DEFAULT_NAME = "default";
 
     ReporterModelDeserializer() {
         super(ReporterModel.class);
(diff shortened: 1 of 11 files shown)


commons/src/main/java/com/powsybl/commons/reporter/ReporterModelDeserializer.java
```diff
@@ -41,29 +46,57 @@ public ReporterModel deserialize(JsonParser p, DeserializationContext ctx) throw
...
     }
 
-    private String getDicName(DeserializationContext ctx) throws JsonMappingException {
-        Object dicNameInjected = ctx.findInjectableValue(DICTIONARY_VALUE_ID, null, null);
-        return dicNameInjected instanceof String ? (String) dicNameInjected : "default";
+    private String getDictionaryName(DeserializationContext ctx) {
+        try {
+            Object dicNameInjected = ctx.findInjectableValue(DICTIONARY_VALUE_ID, null, null);
+            return dicNameInjected instanceof String ? (String) dicNameInjected : DICTIONARY_DEFAULT_NAME;
+        } catch (JsonMappingException e) {
```
MioRtia, line 72: Just a question: what could cause `JsonMappingException`? Is it something worthy to put a warning?

MioRtia (approved):
It is okay for me, I just have one question.

commons/src/main/java/com/powsybl/commons/reporter/ReporterModelDeserializer.java
flo-dup: The exception is raised if the injectable value does not exist. So, yes you're right, it's worth a warning! or at least a log with info level

PR #1784, opened 2021-07-06
PSSE. Convert to IIDM. Support non-transformer branches with different nominal voltage at ends.

marqueslanauja:

Feature

Only non-transformer branches with the same nominal voltage at the ends are well supported.

Conversion has been extended to also support non-transformer branches with different nominal voltage at the ends.

**Other information**:

`RemoteControl.raw` and `TwoTerminalDc.raw` cases have been improved. Both cases have a line and a transformer between the same buses. The line has been deleted as this configuration is not real.


zamarrenolm (approved):

zamarrenolm:
Functional test made by comparing load flow results in external tool (HELM-Flow) for IEEE 24 buses test case:

- using original PSSE file (available in psse-model test resources).
- using XIIDM produced from PSSE file imported in powsybl.

In the test case, the non-transformer branches between buses 8-7, 8-9, 8-10 have different nominal voltages at the two ends. Without the fix, the load flow results for the XIIDM case do not reproduce the voltages in the input case. With the fix, the same voltages are obtained.

PR #1789, opened 2021-07-13
[WIP] First proposition for custom validation (default already in API, custom in iidm-validation)

MioRtia:

No

Feature

Always strict validation, not appropriate for networks in transitional states. 

Are available:
- default validation (no change)
- final validation (only final check if called)
- report validation (only logs)

By default, default validation is used. Another can be specified in config.

Not really. If no validation is configured, default validation is used.


geofjamg (changes requested):
I don't think this is a right solution. I would break all pieces of code that rely on the fact that the network has been validated. Furthermore we don't need to a custom validation, we just need to switch off some validation in specific use cases.

**The key point for releasing part of the validation is to be able to know if the network is validated or not.**

So we should be able to check like this
```java
if (network.isValid()) {
  ...
}
```

So for instance in OpenLoadFlow we heavily rely on the fact that the network is validated and if it not the case it will just crash because we suppose some check have been done.
So to avoid regression we would start in OLF (and in many other place in powsybl and other apps where we supposed to network to be valid) with a check like this:
```java
if (!network.isValid()) {
   network.validate();
   if (!network.isValid()) {
       throw new PowsyblException("Inconsistent network");
   }
}
```

Or simpler with a more convenient method that would do exactly the same:
```java
network.ensureValidity();
``` 

So we need:
  - a validity status added to the network
  - a validation "mode" like
```java
network.setValidationActive(true)
network.isValidationActive()
```
Depending on this status, some validation are skipped (post-poned in reality)
  - a method to re-run validation (and validity status will be updated at the end) that return a boolean
  - an other one that re-run validation if needed and throw an exception in case of failure
  - an other one that re-run validation and generate a report so that we can have all error at the same time for debugging or to display the report in a UI 

At first I do not see the point to have a custom validation of different level of validation. Maybe one day we will have this kind of need (like operational validation or structural validation) but I don't think we should start with it.

A more general comment is about this kind of dev. We should not start coding for something so critical/important for the platform but first write a issue with all requirements and some basic specifications and then start the discussion so that everybody can contribute.

MioRtia:
> I don't think this is a right solution. I would break all pieces of code that rely on the fact that the network has been validated. Furthermore we don't need to a custom validation, we just need to switch off some validation in specific use cases.
> 
> **The key point for releasing part of the validation is to be able to know if the network is validated or not.**
> 
> So we should be able to check like this
> 
> ```java
> if (network.isValid()) {
>   ...
> }
> ```
> 
> So for instance in OpenLoadFlow we heavily rely on the fact that the network is validated and if it not the case it will just crash because we suppose some check have been done.
> So to avoid regression we would start in OLF (and in many other place in powsybl and other apps where we supposed to network to be valid) with a check like this:
> 
> ```java
> if (!network.isValid()) {
>    network.validate();
>    if (!network.isValid()) {
>        throw new PowsyblException("Inconsistent network");
>    }
> }
> ```
> 
> Or simpler with a more convenient method that would do exactly the same:
> 
> ```java
> network.ensureValidity();
> ```
> 
> So we need:
> 
>     * a validity status added to the network
> 
>     * a validation "mode" like
> 
> 
> ```java
> network.setValidationActive(true)
> network.isValidationActive()
> ```
> 
> Depending on this status, some validation are skipped (post-poned in reality)
> 
>     * a method to re-run validation (and validity status will be updated at the end) that return a boolean
> 
>     * an other one that re-run validation if needed and throw an exception in case of failure
> 
>     * an other one that re-run validation and generate a report so that we can have all error at the same time for debugging or to display the report in a UI
> 
> 
> At first I do not see the point to have a custom validation of different level of validation. Maybe one day we will have this kind of need (like operational validation or structural validation) but I don't think we should start with it.
> 
> A more general comment is about this kind of dev. We should not start coding for something so critical/important for the platform but first write a issue with all requirements and some basic specifications and then start the discussion so that everybody can contribute.

Thank you for your feedback! You make very valid points.

Just a precision: this dev was in no way meant to be integrated as such, it was more of a test to see the limitations and the possibilities of a custom validation feature. It was difficult on our side to discuss about the specifications without an example of implementation/API (even a wrong or incomplete one).

MioRtia:
Replaced by #1819

Issue #1787, opened 2021-07-09
Observability extensions


annetill:
We have nothing in IIDM to model observability after a state estimation.

annetill:
What we can do:
- One extension by equipments: we should extend `Injection`, `Line`, `TwoWindingsTransformer`, `BusBarSection` and maybe in a second time `ThreeWindingsTransformer` and `ConfiguredBus`.
- In all extensions, we will have a boolean that says if the equipment is observable or not. The definition of observable regroups active power, reactive power, voltage and angle (one for all).
- In all extensions, we will have some quality information. In the `Injection` extension, we will have a `standardDeviation` for P, another one for Q and another one for V. Maybe quality information about angle is not needed. In the other extensions, we need to add the side information and it multiplies the number of field for the standard deviations. For bus bar section and configured bus, only voltage standard deviation is required.
- In all extensions, we will also have an information about redondance. The number of redondance field is similar to what we need for standard deviations.

MioRtia:
For injection, I think we can have something like that:
```xml
<io:injectionObservability observable="true">
    <io:quality standardDeviationP="0.02" standardDeviationQ="0.5" standardDeviationV="0.0" redundant="false"/>
</io:injectionObservability>
```

annetill:
> For injection, I think we can have something like that:
> 
> ```
> <io:injectionObservability observable="true">
>     <io:quality standardDeviationP="0.02" standardDeviationQ="0.5" standardDeviationV="0.0" redundant="false"/>
> </io:injectionObservability>
> ```

The `redundant` field should be multiplied for P, Q and V I think.

MioRtia:
You're right! We should have `redundantP`, `redundantQ` and so on.

MioRtia:
Should we have several `observable` fields as well?

tadam50:
> > For injection, I think we can have something like that:
> > ```
> > <io:injectionObservability observable="true">
> >     <io:quality standardDeviationP="0.02" standardDeviationQ="0.5" standardDeviationV="0.0" redundant="false"/>
> > </io:injectionObservability>
> > ```
> 
> The `redundant` field should be multiplied for P, Q and V I think.

We should have something like that ?
```
<io:injectionObservability observable="true">
    <io:quality standardDeviationP="0.02" redundant="false"/>
    <io:quality standardDeviationQ="0.5" redundant="false"/>
    <io:quality standardDeviationV="0.0" redundant="false"/>
</io:injectionObservability>
[...]

PR #1790, opened 2021-07-13
Add new extension : BranchObservability

tadam50:

#1787

New extension for banches

No extension for banches

New extension for banches

**Other information**:


iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/BranchObservability.java
```diff
@@ -0,0 +1,48 @@
...
+    BranchObservability<B> setStandardDeviationQ(float standardDeviationQ, Branch.Side side);
+
+    boolean isRedundantQ(Branch.Side side);
+
+    BranchObservability<B> setRedundantQ(boolean redundant, Branch.Side side);
+
+    /** StandardDeviation for Voltage amplitude */
+    float getStandardDeviationV(Branch.Side side);
+
+    BranchObservability<B> setStandardDeviationV(float standardDeviationV, Branch.Side side);
```
annetill, line 43: @sylvlecl : it is what we decided together but are you sure that V quality has a place in injection and branch observability extensions ? For me it is more for bus bar section and maybe for voltage level (for jonction between several lines without bus bar section) and for configured bus, where we have a V. Furthermore, I have some doubts about observability definition. In operation, the equipments are observable if we have measurements at them (P, Q and V). But as a result of state estimation, what is the good definition of observability ? It seems that it is a definition relying to voltage only. Can you help us a little bit on that ?

sylvlecl: Sorry for the delay:
- no you're right, we should not have V here, V should only be on bus bars and configured buses, in order to not duplicate it. Note though, that we may drop some values, because some buses have no corresponding bus bar or configured bus.
- well the definition would be that the state estimation had enough information to estimate physical values for that element (which does not necessarily mean that there was some measurement right on that element)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/BranchObservabilityImpl.java
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * @author Thomas Adam <tadam at silicom.fr>
+ */
+public class BranchObservabilityImpl<T extends Branch<T>> extends AbstractExtension<T>
+        implements BranchObservability<T> {
+
+    private boolean observable;
+
+    private final Map<Branch.Side, Pair<Double, Boolean>> standardDeviationP = new EnumMap<>(Branch.Side.class);
```
sylvlecl, line 25: Same comment as in the injection part: having plain fields seems simpler / more memory efficient.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/BranchObservabilityImpl.java
```diff
@@ -0,0 +1,120 @@
...
+        return this;
+    }
+
+    @Override
+    public ObservabilityQuality<B> getQualityP1() {
+        return qualityP1;
+    }
+
+    @Override
+    public BranchObservability<B> setQualityP1(double standardDeviation, Boolean redundant) {
```
flo-dup, line 63: Like in InjectionObservability, I suppose we're trying to avoid creating new objects if quality changes everywhere? Then we could also add a `setQualityP1(double)`, to avoid boxing and creating Boolean on each branch at each quality change, even if the redundant value (often?) remains the same. Well I must admit that boxing boolean is not very significant!

flo-dup (approved):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/BranchObservabilityImpl.java
MioRtia: To reinforce what @flo-dup said, I think we should have two methods:
one `setQualityP1(double standardDeviation, boolean redundant)` and one `setQualityP1(double standardDeviation)`. Only the field in the class should be `Boolean` (same for all optional `redundant` fields).

annetill (approved):

PR #1793, opened 2021-07-16
First try at creating builders from existing object

miovd:

Feature: create components from existing components (e.g. a new generator initialized with all the values of an existing generator)

No

**Other information**:

Documentation not complete: adders for partial contained objects must be documented (if everyone agrees)


geofjamg:
Nice feature! 

Maybe we should be more clear about what is copied from the reference element and what is not.
For instance, with lines I think that we could also copy the voltageLevelId1 and voltageLevelId2 so the only thing that would need to be specified would be the node1 and node2, because this is the only builder parameter that have to be unique.

So
```java 
l2 = n.newLine(l1)
  .setId("l2")
  .setNode1(3)
  .setNode(4)
  .add();
```

Would create a l2 which is exactly like l2 (in parallel) but with a different connection point.


In case of bus/breaker this is different, it is allowed to have multiple elements connected to same bus so we would only need to specify a new id.
 ```java 
l2 = n.newLine(l1)
  .setId("l2")
  .add();
```

MioRtia:
Replaced by #1819

miovd:
> Nice feature!
> 
> Maybe we should be more clear about what is copied from the reference element and what is not. For instance, with lines I think that we could also copy the voltageLevelId1 and voltageLevelId2 so the only thing that would need to be specified would be the node1 and node2, because this is the only builder parameter that have to be unique.
> 
> So
> 
> ```java
> l2 = n.newLine(l1)
>   .setId("l2")
>   .setNode1(3)
>   .setNode(4)
>   .add();
> ```
> 
> Would create a l2 which is exactly like l2 (in parallel) but with a different connection point.
> 
> In case of bus/breaker this is different, it is allowed to have multiple elements connected to same bus so we would only need to specify a new id.
> 
> ```java
> l2 = n.newLine(l1)
>  .setId("l2")
>  .add();
> ```

I have modified the PR for the builder copying everything in any case from the existing object (which would be what a user expects). No field is left blank, the user has to rewrite the ones they need/want to define.

colineplqt:
This feature is now available (see https://github.com/powsybl/powsybl-core/pull/3208, https://github.com/powsybl/powsybl-core/pull/3260, https://github.com/powsybl/powsybl-core/pull/3130, https://github.com/powsybl/powsybl-core/pull/3116), so I close this PR.

PR #1796, opened 2021-07-20
Make substations optional

MioRtia:

No

Feature

Substation are mandatory to create voltage levels and transformers.

Substations are optional. You can create voltage levels directly from the network. You can also create transformers directly from the network **if and only if** at least one end of the transformer is in a voltage level without substation.

Yes (must be discussed if the API should be really broken or no).

### DIFF
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ThreeWindingsTransformer.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ThreeWindingsTransformer.java
@@ -8,6 +8,7 @@ package com.powsybl.iidm.network;
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.Optional;
 import java.util.stream.Stream;
 
 /**
@@ -301,6 +302,10 @@ public interface ThreeWindingsTransformer extends Connectable<ThreeWindingsTrans
      * Get the substation to which the transformer belongs.
      */
     Substation getSubstation();
+
+    default Optional<Substation> getOptionalSubstation() {
+        return Optional.ofNullable(getSubstation());
+    }
 
     /**
      * Get the leg at the primary side.
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ThreeWindingsTransformer.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ThreeWindingsTransformer.java
@@ -298,16 +298,7 @@ public interface ThreeWindingsTransformer extends Connectable<ThreeWindingsTrans
      */
     Side getSide(Terminal terminal);
 
-    /**
-     * Get the substation to which the transformer belongs.
-     */
-    default Substation getSubstation() {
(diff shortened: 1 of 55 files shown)


mathbagu (commented):
It seems when you create an equipment from the network, a new fictitious substation is created. In which case the getOptionalSubstation method could return an empty optional?

MioRtia:
> It seems when you create an equipment from the network, a new fictitious substation is created. In which case the getOptionalSubstation method could return an empty optional?

Do you mean returning an empty optional when substation is fictitious? It is a bit weird, isn't it? I only created fictitious substations in the default implementation in order not to break the API too much but maybe I should...

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractThreeWindingsTransformerTest.java
```diff
@@ -186,6 +186,94 @@ public void baseTests() {
         assertEquals(count - 1L, network.getThreeWindingsTransformerCount());
     }
 
+    @Test
+    public void invalidSubstationContainer() {
+        thrown.expect(ValidationException.class);
+        thrown.expectMessage("3 windings transformer 'twt': the 3 windings of the transformer shall belong to the substation 'sub'");
```
annetill, line 192: I am going to make a review just from the test review: so you forbid to create a two windings transformer between 2 voltage levels if they are not located in the same substation. Right ? But you allow to create a two windings transformer between 2 voltage levels that are not contained in a substation ? This last functionality could lead to non realistic network for sure.
- If you do that for network with no substation at all: ok but maybe we need a quality check?
- If you ahve another reason, why not going further and allowing two windings transformer between 1 voltage level in a substation and 1 without ?

annetill (approved):

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractThreeWindingsTransformerTest.java
annetill: Resolvedn I have miss understood the test.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ThreeWindingsTransformer.java
```diff
@@ -302,6 +303,10 @@ default Leg setRatedS(double ratedS) {
      */
     Substation getSubstation();
 
+    default Optional<Substation> getOptionalSubstation() {
```
geofjamg, line 306: The naming is really ugly and we are going ti use it EVERYWHERE. Even if it will create a breaking change, should we consider just replacing the getSubstation ?

MioRtia: Yes, I think you're right

annetill (approved):

PR #1797, opened 2021-07-26
Cgmes regulating controls: Refact. Uniform treatment.

marqueslanauja:

Feature

Some regulating controls are discarded if attributes are missing.
Regulating terminal is "localize" when it is not found.
Potential inconsistency between point of regulation and value of setpoint.

Do not lose regulating controls. Keep them even if not all required data is available.
Do not change the definition of the regulating control. Do not localize regulating terminals.
Do not provide default values. All attributes set to NaN or null if they are not explicitly defined.
Enable only valid regulating controls. All attributes must be defined and valid.

**Other information**:


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -66,6 +77,16 @@ public static void checkActivePowerLimits(Validable validable, double minP, doub
         }
     }
 
+    /**
+     * @deprecated
```
MioRtia, line 81: This remark is valid for all other deprecated methods in here from what I can see: why did you deprecate this method? It was implemented to do some check without checking everything when only one regulation parameter was changed (here targetDeadband). Not sure I understand why it shouldn't be used? I am asking since this change could cause issues in the future for other implementations using the "old" methods to keep track on checks.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/GeneratorXml.java
```diff
@@ -71,22 +87,23 @@ protected Generator readRootElementAttributes(GeneratorAdder adder, NetworkXmlRe
         double targetP = XmlUtil.readDoubleAttribute(context.getReader(), "targetP");
         double targetV = XmlUtil.readOptionalDoubleAttribute(context.getReader(), "targetV");
         double targetQ = XmlUtil.readOptionalDoubleAttribute(context.getReader(), "targetQ");
+        if (Double.isNaN(targetQ)) {
+            targetQ = 0.0;
+        }
```
MioRtia, line 92: I really don't agree with this: he serializer should not show different data from the Java object! Either it is relevant to put `targetQ = 0.0` in the implementation (and I'm not sure it should, it is really different to have `NaN` and `0.0`) or it should not write `targetQ` when it is `NaN` but changing attributes values on the fly during deserialization and serialization is not a good practice.

Furthermore, I don't understand why you changed the structure of reading generator? It seems to be for backwards compatibility but this doesn't seem necessary to me?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/GeneratorXml.java
```diff
@@ -63,6 +63,22 @@ protected GeneratorAdder createAdder(VoltageLevel vl) {
...
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    protected void readElement(String id, GeneratorAdder adder, NetworkXmlReaderContext context) throws XMLStreamException {
+        boolean voltageRegulatorOn = readRootElementAttributesGenerator(adder, context);
+        Generator g = adder.add();
+        readPQ(null, g.getTerminal(), context.getReader());
+        readSubElementsGenerator(g, context);
+        if (voltageRegulatorOn && g.getRegulatingTerminal() == null) {
```
MioRtia, line 75: This condition should be only if IIDM version < 1.6.
This PR is for me a breaking change in serialization and should justify a new version. Luckily, we bumped in 1.6 so this should be okay to only check that the version of the context is < 1.6

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/GeneratorXml.java
```diff
@@ -50,7 +50,7 @@ protected void writeRootElementAttributes(Generator g, VoltageLevel vl, NetworkX
 
     @Override
     protected void writeSubElements(Generator g, VoltageLevel vl, NetworkXmlWriterContext context) throws XMLStreamException {
-        if (!Objects.equals(g, g.getRegulatingTerminal().getConnectable())) {
+        if (g.getRegulatingTerminal() != null && !Objects.equals(g, g.getRegulatingTerminal().getConnectable())) {
```
MioRtia, line 53: The condition should be different depending on XIIDM version

iidm/iidm-xml-converter/src/test/resources/V1_0/staticVarCompensatorRoundTripRef.xml
```diff
@@ -5,7 +5,7 @@
             <iidm:busBreakerTopology>
                 <iidm:bus id="B1"/>
             </iidm:busBreakerTopology>
-            <iidm:generator id="G1" energySource="OTHER" minP="50.0" maxP="150.0" voltageRegulatorOn="true" targetP="100.0" targetV="400.0" bus="B1" connectableBus="B1">
+            <iidm:generator id="G1" energySource="OTHER" minP="50.0" maxP="150.0" voltageRegulatorOn="true" targetP="100.0" targetV="400.0" targetQ="0.0" bus="B1" connectableBus="B1">
```
MioRtia, line 8: As a general good practice: files of previous XIIDM versions should not change except in extremely rare cases

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/GeneratorAdder.java
```diff
@@ -35,10 +35,12 @@
     GeneratorAdder setVoltageRegulatorOn(boolean voltageRegulatorOn);
 
     /**
-     * Set the regulating terminal, If not set or set to null local terminal is used.
+     * Set the regulating terminal.
      */
     GeneratorAdder setRegulatingTerminal(Terminal regulatingTerminal);
 
+    GeneratorAdder useLocalRegulation(boolean use);
```
MioRtia, line 42: This should have a default implementation:
```java
default GeneratorAdder useLocalRegulation(boolean use) {
      return this; // does nothing
}
```
Maybe add javadoc too

MioRtia (changes requested):
Hi, thank you for your work!
The API is very good to me that way, it is lacking default implementations I think.
I have more reserves about the implementation and would like some explanation as why the checks changed so much.
I don't agree at all with what has been done for the XIIDM serialization as I don't think it is good practice to write something different that the values present in the Java object.

Don't hesitate to explain your choices if you feel they are warranted!

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/GeneratorAdder.java
zamarrenolm: Instead of leaving an empty implementation, I think we should throw an unsupported operation exception, keeping the same strategy we have used when extending interfaces for shunt compensator and static var compensator adders. Do you agree @MioRtia ?

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/network/compare/IssuesTest.java
```diff
@@ -69,17 +67,17 @@ private static Network createNetwork() {
...
             .setBus("B1")
             .setMinP(0)
             .setMaxP(1)
             .setTargetP(1)
             .setTargetQ(0)
+            .useLocalRegulation(true)
             .setVoltageRegulatorOn(true)
             .setTargetV(400)
             .add();
-        assertEquals(g1.getRegulatingTerminal(), g1.getTerminal());
```
MioRtia, line 82: You can keep this assert I think

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorAdder.java
```diff
@@ -36,6 +36,10 @@ default ShuntCompensatorAdder setRegulatingTerminal(Terminal regulatingTerminal)
         throw new UnsupportedOperationException();
     }
 
+    default ShuntCompensatorAdder useLocalRegulation(boolean use) {
+        throw new UnsupportedOperationException();
```
MioRtia: Like in `GeneratorAdder`, put a default implementation that does nothing

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/StaticVarCompensatorAdder.java
```diff
@@ -45,5 +45,9 @@ default StaticVarCompensatorAdder setRegulatingTerminal(Terminal regulatingTermi
         throw new UnsupportedOperationException("Not yet implemented");
     }
 
+    default StaticVarCompensatorAdder useLocalRegulation(boolean use) {
+        throw new UnsupportedOperationException("Not yet implemented");
```
MioRtia: Same remark

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
```diff
@@ -134,8 +149,11 @@ protected void readElement(String id, ShuntCompensatorAdder adder, NetworkXmlRea
             adder.setSectionCount(sectionCount);
         });
         readNodeOrBus(adder, context);
-        double p = XmlUtil.readOptionalDoubleAttribute(context.getReader(), "p");
-        double q = XmlUtil.readOptionalDoubleAttribute(context.getReader(), "q");
+
+        return voltageRegulatorOn[0];
+    }
+
+    private ShuntCompensator readSubElementsShuntCompensator(String id, ShuntCompensatorAdder adder, NetworkXmlReaderContext context) throws XMLStreamException {
```
MioRtia, line 156: I think you can call override the `readSubElements` method with this one and call that rather than creating another

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/StaticVarCompensatorXml.java
```diff
@@ -85,15 +103,12 @@ protected StaticVarCompensator readRootElementAttributes(StaticVarCompensatorAdd
...
         readNodeOrBus(adder, context);
-        StaticVarCompensator svc = adder.add();
-        readPQ(null, svc.getTerminal(), context.getReader());
-        return svc;
+        return regulationMode;
     }
 
-    @Override
-    protected void readSubElements(StaticVarCompensator svc, NetworkXmlReaderContext context) throws XMLStreamException {
+    private void readSubElementsStaticVarCompensator(StaticVarCompensator svc, NetworkXmlReaderContext context) throws XMLStreamException {
```
MioRtia, line 111: Same remark

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/GeneratorXml.java
```diff
@@ -63,30 +64,53 @@ protected GeneratorAdder createAdder(VoltageLevel vl) {
...
-                .add();
-        readPQ(null, g.getTerminal(), context.getReader());
-        return g;
+                .setVoltageRegulatorOn(false);
+        return rc;
     }
 
-    @Override
-    protected void readSubElements(Generator g, NetworkXmlReaderContext context) throws XMLStreamException {
+    private void readSubElementsGenerator(Generator g, NetworkXmlReaderContext context) throws XMLStreamException {
```
MioRtia, line 113: Same remark

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/GeneratorXml.java
MioRtia: Can you explain that to me again?

MioRtia (commented):
Thank you for your great work! I have a few remarks but it is pretty much okay for me.
Can you still explain to me why you set `targetQ` to 0.0 when it is `NaN` sometimes in `GeneratorXml`?
Also why in checks, you now always check all the regulation values even if only `targetDeadband` has been changed?
It is just for me to understand your choices!

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/GeneratorXml.java
MioRtia: This should change, shouldn't it? The terminal if it is local should be written in IIDM >= 1.6 in order to make the difference with a null (undefined) terminal. If the regulating terminal is null, it should throw an exception if IIDM < 1.6

marqueslanauja:
> Thank you for your great work! I have a few remarks but it is pretty much okay for me.
> Can you still explain to me why you set `targetQ` to 0.0 when it is `NaN` sometimes in `GeneratorXml`?
> Also why in checks, you now always check all the regulation values even if only `targetDeadband` has been changed?
> It is just for me to understand your choices!

I am improving the code of GneratorXml and i have to merge the last changes of the master branch. I will explain you the final version and i will try to document it.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
marqueslanauja: I decided to organize the validation methods in two sets: 

- The complete check methods that verify all the attributes associated with the regulating control.
- The individual check methods that only verify one attribute. 

Complete methods are used in the adder and when  the enabled attribute is changed after doing the adder. If the regulating control is enabled, all the attributes must be valid, complete verification.

Individual methods are used when the corresponding attribute is changed after the adder. The verification of one attribute depends always on the enabled attribute. It the control is enabled only valid attributes are allowed. 

If the control is not enabled the verification is not done. That allows to set the attributes of the controls after the adder. (Controls must be filled at the end of the conversion process so all remote regulating terminals have previously been defined). When only the EQ file is imported the control could be created disabled with not valid attributes.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/GeneratorXml.java
marqueslanauja: Finally, I preserved the structure of reading generator and I avoided ` targetQ = 0.0`.
In generators `targetQ `must be valid when the voltage regulating control is disabled. In this case the generator is considered as an injection and P and Q must be defined and valid.

When a generator is read from an xml file we have the following sequence:

- Read main attributes
- Create the generator and finally
- Set the `regulatingTerminal`.

Initially, the only option was to create the generator with the voltage regulating control disabled. Then targetQ must be set to 0.0 (a temporary valid targetQ) when it is not valid (NaN).

After supporting `useLocalRegulation` a new alternative appears. The regulating control can be created `enabled` by setting the regulatingTerminal to local (RegulatingTerminal is the only attribute read after creating the control). 
That allows to preserve the structure of reading generator and discard `targetQ = 0.0`

`UseLocalRegulation `is only used when the regulating control is enabled. At the third step of the sequence, the local regulatingTerminal is overwritten by the value received in the xml file.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/GeneratorXml.java
marqueslanauja: Done.

Until version 1.6 only remote `regulatingTerminals `were written. This generates an ambiguity in the reading process. In disabled voltage regulating controls the regulatingTerminal could be null or local. I decided to consider it null.

Since version 1.6 local and remote `regulatingTerminals `are written to eliminate the previous ambiguity. This modification of the code provokes adjustments in several xml and xiidm resource files.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/network/compare/IssuesTest.java
marqueslanauja: It is not necessary and can be deleted as useLocalRegulation has previously been set to true.

> Thank you for your great work! I have a few remarks but it is pretty much okay for me.
> Can you still explain to me why you set `targetQ` to 0.0 when it is `NaN` sometimes in `GeneratorXml`?
> Also why in checks, you now always check all the regulation values even if only `targetDeadband` has been changed?
> It is just for me to understand your choices!

I answered in conversations at the top

marqueslanauja: Done. I kept the assert

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensatorAdder.java
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/StaticVarCompensatorAdder.java
marqueslanauja: Done

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/ShuntXml.java
marqueslanauja: Finally,  the standard structure of reading has been preserved.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/StaticVarCompensatorXml.java
marqueslanauja: Finally, the standard structure of reading has been preserved.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/GeneratorXml.java
marqueslanauja: Finally, the standard structure of reading has been preserved.

MioRtia (commented):
A few remarks. VSC converter stations have also now regulating terminals, it would be good to refactor this for them also.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineImpl.java
```diff
@@ -122,7 +124,7 @@ public double getTargetQ() {
         @Override
         public GenerationImpl setTargetQ(double targetQ) {
             int variantIndex = danglingLine.network.get().getVariantIndex();
-            ValidationUtil.checkVoltageControl(danglingLine, voltageRegulationOn.get(variantIndex), targetV.get(variantIndex), targetQ);
+            ValidationUtil.checkReactivePowerTarget(this, VALIDABLE_TYPE_DESCRIPTION, targetQ, voltageRegulationOn.get(variantIndex));
```
MioRtia, line 127: Global remark but I'm still not sure why we want to change the checks in this PR?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/GeneratorXml.java
```diff
@@ -64,53 +63,35 @@ protected GeneratorAdder createAdder(VoltageLevel vl) {
...
-        if (Double.isNaN(targetQ)) {
-            targetQ = 0.0;
-        }
+        double targetQ = XmlUtil.readOptionalDoubleAttribute(context.getReader(), "targetQ");
         readNodeOrBus(adder, context);
-        adder.setEnergySource(energySource)
+        boolean useLocalRegulationOn = false;
+        if (voltageRegulatorOn) {
+            useLocalRegulationOn = true;
+        }
```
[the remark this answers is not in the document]
marqueslanauja: Done.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineImpl.java
marqueslanauja: I reviewed again the validation class trying to improve and simplify it. Additionally to the homogeneity changes, the following changes have been made. These changes are necessary.

Generator controls:
- Do not check `regulatingTerminal` if `useLocalRegulation` is used.
- Fix `checkRegulatingTerminal` as is not considering the enabled attribute.

StaticVarCompensator controls:
- Do not check `regulatingTerminal` if `useLocalRegulation` is used.
- Fix `checkRegulatingTerminal` as is not considering the enabled attribute.

ShuntCompensator controls:
- Do not check `regulatingTerminal` if `useLocalRegulation` is used.
- Fix `checkRegulatingTerminal` as is not considering the enabled attribute.
- Fix `setVoltageRegulatorOn` as the `regulatingTerminal` is not checked.

RatioTapChanger controls:
- Fix `setRegulating` as the `targetDeadband` is not checked.

PhaseTapChanger controls:
- Fix `setRegulating` as the `targetDeadband` is not checked.

Vsc converter controls:
- Do not check `regulatingTerminal` if `useLocalRegulation` is used.

If you want  I can make all the necessary changes trying to discard all the changes associated with homogeneity.

marqueslanauja:
> A few remarks. VSC converter stations have also now regulating terminals, it would be good to refactor this for them also.

Done

commons/src/main/java/com/powsybl/commons/extensions/AbstractExtensionAdder.java
```diff
@@ -19,7 +19,7 @@
 public abstract class AbstractExtensionAdder<T extends Extendable<T>, E extends Extension<T>>
         implements ExtensionAdder<T, E> {
 
-    private final T extendable;
+    protected final T extendable;
```
MioRtia, line 22: It seems you do that for `RemoteReactivePowerAdderImpl`. I'd suggest to keep this private and make `RemoteReactivePowerAdderImpl` implement `Validable` with
```java


    @Override
    public String getMessageHeader() {
        return "generatorRemoteReactivePowerControl of generator " + extendable.getId();
    }
```

and use this `ValidationUtil.checkReactivePowerControl(this, regulatingTerminal, targetQ, enabled, extendable.getNetwork());`

Also a note in `RemoteReactivePowerAdderImpl`: do the checks in `createExtension` method, not before

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -75,23 +186,13 @@ public static void checkTargetDeadband(Validable validable, String validableType
...
-
-    public static void checkVoltageControl(Validable validable, Boolean voltageRegulatorOn, double voltageSetpoint, double reactivePowerSetpoint) {
-        if (checkVoltageControl(validable, voltageRegulatorOn, voltageSetpoint) && Double.isNaN(reactivePowerSetpoint)) {
-            throw createInvalidValueException(validable, reactivePowerSetpoint, "reactive power setpoint", "voltage regulator is off");
+    /**
+     * For validating target deadband.
+     * If control is enabled target deadband must be defined and valid (>= 0.0).
+     */
+    public static void checkTargetDeadband(Validable validable, String validableType, double targetDeadband, boolean voltageRegulatorOn) {
+        if (voltageRegulatorOn && !validDeadband(targetDeadband)) {
```
MioRtia, line 194: Not sure if it is worth changing. Even when it is not regulating, having targetDeadband < 0 does not make sense... Before the check was: if regulating, targetDeadband is not NaN (no check if NaN if not regulating) but in every case we must have targetDeadband >=0 if not NaN

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -127,12 +228,41 @@ public static void checkHvdcMaxP(Validable validable, double maxP) {
...
         }
     }
 
+    /**
+     * For validating regulating terminal.
+     * If control is enabled regulating terminal must be defined and valid (Network associated with it should be ok).
+     */
+    public static void checkRegulatingTerminal(Validable validable, String validableType, Terminal regulatingTerminal,
+        boolean voltageRegulatorOn, Network network) {
+        if (voltageRegulatorOn && !validRegulatingTerminal(regulatingTerminal, network)) {
```
MioRtia, line 250: Same thing here, it does not make sense to have a regulating terminal in another network even if not regulating

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -127,12 +228,41 @@ public static void checkHvdcMaxP(Validable validable, double maxP) {
...
+        }
+    }
+
+    /**
+     * For validating target regulating terminal.
+     * If control is enabled regulating terminal must be defined and valid (Network associated with it should be ok).
+     */
+    public static void checkRegulatingTerminal(Validable validable, String validableType, Terminal regulatingTerminal,
+        RegulationMode regulationMode, Network network) {
+        if (regulationMode != StaticVarCompensator.RegulationMode.OFF && !validRegulatingTerminal(regulatingTerminal, network)) {
```
MioRtia, line 261: Same thing as above

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -127,12 +228,41 @@ public static void checkHvdcMaxP(Validable validable, double maxP) {
         }
     }
 
+    /**
+     * @deprecated
+     * Use
+     *     {@link #checkRegulatingTerminal(Validable, String, Terminal, boolean, Network)} or
+     *     {@link #checkRegulatingTerminal(Validable, String, Terminal, RegualtionMode, Network)} instead
```
MioRtia, line 235: Small typo in javadoc

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -357,13 +465,40 @@ public static void checkRatioTapChangerRegulation(Validable validable, boolean r
     }
 
     public static void checkRatioTapChangerRegulation(Validable validable, boolean regulating,
-                                                      Terminal regulationTerminal, double targetV, Network network) {
+        Terminal regulationTerminal, double targetV, double targetDeadband, Network network) {
+        checkRatioTapChangerRegulation(validable, regulating, true, regulationTerminal, targetV, targetDeadband, network);
+    }
```
MioRtia, line 470: You can delete this one

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -357,13 +465,40 @@ public static void checkRatioTapChangerRegulation(Validable validable, boolean r
...
     }
 
+    /**
+     * For validating ratio tapChanger controls.
+     * If the control is enabled voltage setpoint, target deadband and regulating terminal must be defined and valid.
+     */
+    public static void checkRatioTapChangerRegulation(Validable validable, boolean regulating, boolean loadTapChangingCapabilities,
+        Terminal regulationTerminal, double targetV, double targetDeadband, Network network) {
+        if (!loadTapChangingCapabilities) {
+            return;
```
MioRtia, line 489: Should not return if `!loadTapChangingCapabilities`, should log warnings (see old method)

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -383,13 +518,302 @@ public static void checkPhaseTapChangerRegulation(Validable validable, PhaseTapC
...
+            throw new ValidationException(validable, "phase regulation cannot be on if mode is FIXED");
+        }
+        checkCurrentOrActivePowerControl(validable, regulationTerminal, regulationValue, targetDeadband,
+            regulating, network);
+    }
+
+    private static void checkCurrentOrActivePowerControl(Validable validable, Terminal regulatingTerminal,
+        double valueSetpoint, double targetDeadband, boolean regulatorOn, Network network) {
+        if (!regulatorOn) {
+            return;
```
MioRtia, line 541: Same remarks as above, some checks make sense even without regulating

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -383,13 +518,302 @@ public static void checkPhaseTapChangerRegulation(Validable validable, PhaseTapC
...
+     */
+    public static void checkSvcRegulation(Validable validable, Terminal regulatingTerminal, double voltageSetpoint,
+        double reactivePowerSetpoint, RegulationMode regulationMode, Network network, boolean validateRegulatingTerminal) {
+        if (regulationMode == null) {
+            throw new ValidationException(validable, "Regulation mode is invalid");
+        }
+        if (regulationMode == StaticVarCompensator.RegulationMode.VOLTAGE) {
+            ValidationUtil.checkContinuousVoltageControl(validable, regulatingTerminal, voltageSetpoint, true, network, validateRegulatingTerminal);
+        } else if (regulationMode == StaticVarCompensator.RegulationMode.REACTIVE_POWER) {
+            ValidationUtil.checkReactivePowerControl(validable, regulatingTerminal, reactivePowerSetpoint, true, network);
```
MioRtia, line 628: You don't have to specify `ValidationUtil` in front of methods of same class

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -383,13 +518,302 @@ public static void checkPhaseTapChangerRegulation(Validable validable, PhaseTapC
...
+    /**
+    * For validating discrete voltage controls.
+    * User can select if regulating terminal should be validated.
+    * If the voltage regulation control is enabled voltage setpoint, target deadband and regulating terminal must be defined and valid.
+    */
+    public static void checkDiscreteVoltageControl(Validable validable, Terminal regulatingTerminal,
+        double voltageSetpoint, double targetDeadband, boolean voltageRegulatorOn, Network network,
+        boolean validateRegulatingTerminal) {
+        if (!voltageRegulatorOn) {
+            return;
```
MioRtia, line 709: Same remark as above

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -383,13 +518,302 @@ public static void checkPhaseTapChangerRegulation(Validable validable, PhaseTapC
...
+    }
+
+    /**
+    * For validating reactive power controls.
+    * If the control is enabled reactive power setpoint and regulating terminal must be defined and valid.
+    */
+    public static void checkReactivePowerControl(Validable validable, Terminal regulatingTerminal,
+        double reactivePowerSetpoint, boolean reactivePowerRegulatorOn, Network network) {
+        if (!reactivePowerRegulatorOn) {
+            return;
```
MioRtia, line 729: Same remark as above

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -383,13 +518,302 @@ public static void checkPhaseTapChangerRegulation(Validable validable, PhaseTapC
...
+        }
+        if (!validReactivePowerSetpoint(reactivePowerSetpoint)) {
+            throw createInvalidValueException(validable, reactivePowerSetpoint, REACTIVE_POWER_SETPOINT);
+        }
+    }
+
+    private static void checkContinuousVoltageControl(Validable validable, Terminal regulatingTerminal,
+        double voltageSetpoint, boolean voltageRegulatorOn, Network network, boolean validateRegulatingTerminal) {
+        if (!voltageRegulatorOn) {
+            return;
```
MioRtia, line 742: Same remark as above

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -383,13 +518,302 @@ public static void checkPhaseTapChangerRegulation(Validable validable, PhaseTapC
...
+    public static boolean validRegulatingReactivePowerControl(Terminal regulatingTerminal, double reactivePowerSetpoint, Network network) {
+        return validRegulatingTerminal(regulatingTerminal, network) && validReactivePowerSetpoint(reactivePowerSetpoint);
+    }
+
+    public static boolean validRegulatingCurrentOrActivePowerControl(Terminal regulatingTerminal, double valueSetpoint, double targetDeadband, Network network) {
+        return validRegulatingTerminal(regulatingTerminal, network) && validCurrentOrActivePowerSetpoint(valueSetpoint) && validDeadband(targetDeadband);
+    }
+
+    private static boolean validRegulatingTerminal(Terminal regulatingTerminal, Network network) {
+        return regulatingTerminal != null && regulatingTerminal.getVoltageLevel().getNetwork() == network;
```
MioRtia, line 798: It seems that it never checks that regulatingTerminal != null if regulating?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -383,13 +518,302 @@ public static void checkPhaseTapChangerRegulation(Validable validable, PhaseTapC
...
+    private static boolean validReactivePowerSetpoint(double reactivePowerSetpoint) {
+        return !Double.isNaN(reactivePowerSetpoint);
+    }
+
+    private static boolean validCurrentOrActivePowerSetpoint(double valueSetpoint) {
+        return !Double.isNaN(valueSetpoint);
+    }
+
+    private static boolean validDeadband(double deadband) {
+        return !Double.isNaN(deadband) && deadband >= 0;
```
MioRtia, line 814: It is a bit weird because you do not check if values are not NaN, which should be checked when regulating

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChanger.java
```diff
@@ -153,7 +151,6 @@ public double getTargetDeadband() {
 
     public C setTargetDeadband(double targetDeadband) {
         int variantIndex = network.get().getVariantIndex();
-        ValidationUtil.checkTargetDeadband(parent, type, this.regulating.get(variantIndex), targetDeadband);
```
MioRtia, line 156: You can keep this one I think

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineAdderImpl.java
```diff
@@ -66,7 +66,7 @@ public GenerationAdder setTargetV(double targetV) {
         public DanglingLineAdder add() {
             ValidationUtil.checkActivePowerLimits(DanglingLineAdderImpl.this, minP, maxP);
             ValidationUtil.checkActivePowerSetpoint(DanglingLineAdderImpl.this, targetP);
-            ValidationUtil.checkVoltageControl(DanglingLineAdderImpl.this, voltageRegulationOn, targetV, targetQ);
+            ValidationUtil.checkVoltageControlTargetQ(DanglingLineAdderImpl.this, targetV, targetQ, voltageRegulationOn);
```
MioRtia, line 69: Not sure why you changed the order of parameters (but it is not a blocking issue)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/GeneratorAdderImpl.java
```diff
@@ -109,19 +117,32 @@ public GeneratorAdder setRatedS(double ratedS) {
     public GeneratorImpl add() {
         String id = checkAndGetUniqueId();
         TerminalExt terminal = checkAndGetTerminal();
+        boolean validateRegulatingTerminal = true;
+        if (useLocalRegulation) {
+            regulatingTerminal = terminal;
+            validateRegulatingTerminal = false;
```
MioRtia, line 123: I don't see the use of this parameter in validation methods? You can check the terminal even if it is local

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/PhaseTapChangerImpl.java
```diff
@@ -83,17 +87,23 @@ public double getRegulationValue() {
...
+        ValidationUtil.checkCurrentOrActivePowerSetpoint(parent, VALIDABLE_TYPE_DESCRIPTION, regulationValue, isRegulating());
         double oldValue = this.regulationValue.set(variantIndex, regulationValue);
         String variantId = network.get().getVariantManager().getVariantId(variantIndex);
         notifyUpdate(() -> getTapChangerAttribute() + ".regulationValue", variantId, oldValue, regulationValue);
         return this;
     }
 
+    @Override
+    public PhaseTapChangerImpl setTargetDeadband(double targetDeadband) {
+        ValidationUtil.checkTargetDeadband(parent, VALIDABLE_TYPE_DESCRIPTION, targetDeadband, isRegulating());
```
MioRtia, line 100: If you don't delete the check in the parent method, this one is not useful

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/RatioTapChangerImpl.java
```diff
@@ -82,17 +86,23 @@ public double getTargetV() {
...
+        ValidationUtil.checkVoltageSetpoint(parent, VALIDABLE_TYPE_DESCRIPTION, targetV, isRegulating());
         double oldValue = this.targetV.set(variantIndex, targetV);
         String variantId = network.get().getVariantManager().getVariantId(variantIndex);
         notifyUpdate(() -> getTapChangerAttribute() + ".targetV", variantId, oldValue, targetV);
         return this;
     }
 
+    @Override
+    public RatioTapChangerImpl setTargetDeadband(double targetDeadband) {
+        ValidationUtil.checkTargetDeadband(parent, VALIDABLE_TYPE_DESCRIPTION, targetDeadband, isRegulating());
```
MioRtia, line 99: Same remark

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
```diff
@@ -211,9 +219,21 @@ public ShuntCompensatorImpl add() {
         String id = checkAndGetUniqueId();
         TerminalExt terminal = checkAndGetTerminal();
 
-        ValidationUtil.checkRegulatingTerminal(this, regulatingTerminal, getNetwork());
-        ValidationUtil.checkVoltageControl(this, voltageRegulatorOn, targetV);
-        ValidationUtil.checkTargetDeadband(this, "shunt compensator", voltageRegulatorOn, targetDeadband);
+        boolean validateRegulatingTerminal = true;
+        if (useLocalRegulation) {
+            regulatingTerminal = terminal;
+            validateRegulatingTerminal = false;
```
MioRtia, line 225: Same remark as above about `validateRegulatingTerminal`

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/StaticVarCompensatorAdderImpl.java
```diff
@@ -82,18 +84,37 @@ public StaticVarCompensatorAdderImpl setRegulatingTerminal(Terminal regulatingTe
...
+
     @Override
     public StaticVarCompensatorImpl add() {
         String id = checkAndGetUniqueId();
         String name = getName();
         TerminalExt terminal = checkAndGetTerminal();
+        boolean validateRegulatingTerminal = true;
+        if (useLocalRegulation) {
+            regulatingTerminal = terminal;
+            validateRegulatingTerminal = false;
```
MioRtia, line 101: Same remark

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VscConverterStationAdderImpl.java
```diff
@@ -62,22 +70,24 @@ public VscConverterStationImpl add() {
         String id = checkAndGetUniqueId();
         String name = getName();
         TerminalExt terminal = checkAndGetTerminal();
+
+        boolean validateRegulatingTerminal = true;
+        if (useLocalRegulation) {
+            regulatingTerminal = terminal;
+            validateRegulatingTerminal = false;
```
MioRtia, line 77: Same remark

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/RemoteReactivePowerAdderImpl.java
```diff
@@ -34,18 +38,22 @@ protected RemoteReactivePowerControl createExtension(final Generator extendable)
...
     public RemoteReactivePowerControlAdder withRegulatingTerminal(Terminal regulatingTerminal) {
+        ValidationUtil.checkRegulatingTerminal((Validable) super.extendable, VALIDABLE_TYPE_DESCRIPTION, regulatingTerminal, enabled, super.extendable.getNetwork());
         this.regulatingTerminal = regulatingTerminal;
         return this;
     }
 
     @Override
     public RemoteReactivePowerControlAdder withEnabled(boolean enabled) {
+        ValidationUtil.checkReactivePowerControl((Validable) super.extendable, regulatingTerminal,
+            targetQ, enabled, super.extendable.getNetwork());
```
MioRtia, line 56: Said it above but checks must be in `createExtension`, not in intermediary methods for adders

MioRtia (changes requested):
Hi, I have made quite a few remarks.
I also think you will need to base this PR on evolution_xiidm/1.7 and allow the new serialization in XIIDM 1.7 rather than in 1.6

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/GeneratorAdderImpl.java
marqueslanauja: Validation of regulating terminal checks that the terminal is not null and its network is the same as the object being added. 

`regulatingTerminal != null && regulatingTerminal.getVoltageLevel().getNetwork() == network;`

The network for the terminal is obtained from its voltage level and terminal voltage level is not set until the terminal is attached with the method (`voltageLevel.attach(terminal, false)`). The terminal is attached after the validation is performed and during the validation there is not voltageLevel associated to the terminal.

```
boolean validateRegulatingTerminal = true;
if (useLocalRegulation) {
            regulatingTerminal = terminal;
            validateRegulatingTerminal = false;
}

ValidationUtil.checkVoltageControlTargetQ(this, regulatingTerminal, targetV, targetQ, voltageRegulatorOn,
   getNetwork(), validateRegulatingTerminal);

GeneratorImpl generator
   = new GeneratorImpl(getNetwork().getRef(),
       id, getName(), isFictitious(), energySource,
       minP, maxP,
       voltageRegulatorOn, regulatingTerminal,
       targetP, targetQ, targetV,
       ratedS);
generator.addTerminal(terminal);
voltageLevel.attach(terminal, false);

```

One option is to do the attach before the validation but we do not want to move the order of validation and terminal attachment (if we attach first and then validation fails, we should unroll the changes).
We prefer do not check the regulating terminal if it is local. So for local regulation (useLocalRegulation) we assume the terminal will be ok since it will be the one of the equipment and it is set automatically by the adder method.

MioRtia: Ah okay, I did not get this! Indeed, if the terminal is not yet attached, it is not considered part of the network...

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
marqueslanauja: I do not quite understand what you mean. Could you please clarify this point?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
marqueslanauja: I do not quite understand what you mean. Could you please clarify this point?

MioRtia: After looking at the code, it seems it is a mistake on my part, please discard this comment.

commons/src/main/java/com/powsybl/commons/extensions/AbstractExtensionAdder.java
marqueslanauja: Done. But in the method `getMessagerHeader `i put a generic message as I do not have access to `extendable`.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
marqueslanauja: Changed. When the control is disabled only valid values and NaN (initial value, targetDeadband is not set) are allowed.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
marqueslanauja: Changed. When the control is disabled only valid values and Null (initial value, `regulatingTerminal` is not set) are allowed.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
marqueslanauja: Changed. When the control is disabled only valid values and Null (initial value, regulatingTerminal is not set) are allowed.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
marqueslanauja: Done. When `!loadTapChangingCapabilities` a warning is logged.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
marqueslanauja: Changed. When the control is disabled only valid values and the initial one (attribute is not set) are allowed.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
marqueslanauja: Changed. When the control is disabled only valid values and the initial one (attribute is not set) are allowed.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
marqueslanauja: Changed. When the control is disabled only valid values and the initial one (attribute is not set) are allowed.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
marqueslanauja: explained above.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/StaticVarCompensatorAdderImpl.java
marqueslanauja: explained above.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VscConverterStationAdderImpl.java
marqueslanauja: explained above.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/RemoteReactivePowerAdderImpl.java
marqueslanauja: Done

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/RatioTapChangerImpl.java
marqueslanauja: explained above

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/PhaseTapChangerImpl.java
marqueslanauja: explained above

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChanger.java
marqueslanauja: Now the method `checkTargetDeadband `is different for `ratioTapChanger `and `phaseTapChanger `as for `ratioTapChanger `the attribute `loadTapChangingCapabilities `has been added to be consistent with the rest of the `ratioTapChanger `methods.

marqueslanauja:
All the check methods when the control is enabled throw two exceptions, one if the attribute value has not set and the other if the attribute value is invalid.

PR #1800, opened 2021-07-30
Traverser from several vertices

flo-dup:

No

Feature

When starting a traverser for several nodes, one has to deal with an encountered array whereas there is already one in the graph traversal. + duplicate code for getting equivalent terminal

Add a method to traverse the graph from several nodes, given an array of vertex indices.

No


iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractNodeBreakerInternalConnectionsTest.java
```diff
@@ -220,43 +219,31 @@ private InternalConnections findInternalConnectionsTraverseStoppingAtTerminals(V
...
                 }
-                return topo.getTerminal(n2) == null ? TraverseResult.CONTINUE : TraverseResult.TERMINATE;
-            });
-        }
+                return TraverseResult.CONTINUE;
+            } else {
+                return TraverseResult.TERMINATE;
+            }
+        });
+
```
flo-dup, line 232: @zamarrenolm I changed a bit the stop criteria here, hence the test assert, but from what I understood you were only testing the traversal of internal connections with a random stop criteria? then my changes would be ok - otherwise I'd revert that part.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/NodeBreakerTopology.java
```diff
@@ -38,26 +35,14 @@ private NodeBreakerTopology() {
     public static void removeIsolatedSwitches(VoltageLevel.NodeBreakerView topo) {
         Objects.requireNonNull(topo, "Node breaker topology view is null.");
 
-        int[] nodes = topo.getNodes();
-        final TIntSet encountered = new TIntHashSet();
         final Set<Switch> encounteredSwitches = new HashSet<>();
+        int[] nodesWithTerminal = Arrays.stream(topo.getNodes()).filter(n -> topo.getTerminal(n) != null).toArray();
+        topo.traverse(nodesWithTerminal, (n1, sw, n2) -> {
+            encounteredSwitches.add(sw); // the traversing started from a terminal, thus the switch is not isolated
+            return TraverseResult.CONTINUE; // if n2 has a terminal, we could also choose to stop as it will be or has already been traversed
```
MioRtia, line 42: Shouldn't we stop when we encounter another terminal? If we don't, it seems to me that if there is a loop, the traverser never stops (but maybe I am mistaken)

MioRtia (commented):
Just left a comment but not sure at all it is relevant

flo-dup:
Unfortunately no, and the related commit goes way back to 3 years ago, but maybe @sylvlecl knows more?

sylvlecl:
> Unfortunately no, and the related commit goes way back to 3 years ago, but maybe @sylvlecl knows more?

I can't remember the exact use case, but I think it was meant to be used in the creation of inputs for Eurostag, in IPST code.
As far as I remember, one step consists in adding so called "step up" transformers (the transfromer between the generator and the network), which requires manipulation of the node breaker topology.

I guess in the process, some switches were "isolated", which must have causes issues down the road, hence this helper function.

Eventually, the function have never been actually used in ipst, so it would be a candidate for removal ...

PR #1801, opened 2021-07-30
Stopping traverser

flo-dup:

No

Feature

If the whole traversing process needs to be stopped, it needs to be done with a custom added state.

The whole traversing process can be stopped by returning `TraverseResult.BREAK` in `Traverser::traverse`

### DIFF
--- a/math/src/main/java/com/powsybl/math/graph/TraverseResult.java
+++ b/math/src/main/java/com/powsybl/math/graph/TraverseResult.java
@@ -11,6 +11,7 @@ package com.powsybl.math.graph;
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
  */
 public enum TraverseResult {
-    CONTINUE,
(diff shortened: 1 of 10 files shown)


math/src/main/java/com/powsybl/math/graph/TraverseResult.java
```diff
@@ -11,6 +11,7 @@
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
  */
 public enum TraverseResult {
-    CONTINUE,
-    TERMINATE
+    CONTINUE,   // Traversing should continue
+    TERMINATE,  // Traversing should terminate on current path
+    BREAK       // Traversing should break, i.e., terminate on all paths
```
MioRtia, line 16: Could you also put this explanation in the javadoc?

MioRtia (approved):

MioRtia (approved):

PR #1802, opened 2021-07-30
Get switches and internal connections from given node

### LINKED ISSUE #1523: Get all switches connected to a node
Feature

To get all switches connected to a node, one needs to iterate over all the switches of its voltage level.

Get all switches connected to a node directly with cached calculation.

Security analysis needs to find the switches corresponding to a node. Using such a method would avoid taking too much time on this search. Besides this list is almost ready in the graph through adjacencyList, which is cached.

  - PowSyBl Version: ...
  - OS Version: ...

flo-dup:

Yes, partly: fixes #1523 

Feature

To get all switches or internal connections from a node, one needs to loop over all of them

Methods are available to retrieve the switches connected to a given node and to retrieve the nodes connected with an internal connection to a given node.

No


math/src/main/java/com/powsybl/math/graph/UndirectedGraph.java
```diff
@@ -180,6 +181,42 @@ default int getVertexCapacity() {
...
     int getEdgeVertex1(int e);
 
+    /**
+     * Return the edge objects connected to the specified vertex.
+     * This method throws a {@link com.powsybl.commons.PowsyblException} if the vertex doesn't exist.
+     *
+     * @param v the vertex index.
+     * @return an iterable of the edge objects
+     */
+    Iterable<E> getEdgeObjects(int v);
```
geofjamg, line 191: For the 4 methods you added, the name is ambiguis, edges also have a "number" and also have an object associated like vertices. At first when just looking at the method name, without reading the javadoc, I misunderstand the purpose of the methods. 
Even if larger, maybe we could call it something like getEdgesConnectedToVertex, getEdgeObjectsConnectedToVertex

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/VoltageLevel.java
```diff
@@ -521,6 +522,34 @@ default void removeInternalConnections(int node1, int node2) {
...
+         * @throws com.powsybl.commons.PowsyblException if node is not found.
+         */
+        IntStream getNodeStreamInternalConnectedTo(int node);
+
+        /**
+         * Get the internal connections connected to the {@code node}.
+         *
+         * @throws com.powsybl.commons.PowsyblException if node is not found.
+         */
+        Iterable<Integer> getNodesInternalConnectedTo(int node);
```
geofjamg, line 551: I known that the API is everywhere made of Stream + Iterable to let impl optimize/lazy load data, but here the number of object returned is very small and maybe we can just return a List instead of the Iterable?

math/src/main/java/com/powsybl/math/graph/UndirectedGraph.java
flo-dup: Yes indeed, it's quite ambiguous, only the s of objects differentiates it from the `getEdgeObject(int e)`. Do you think the (old) existing `getEdgeObjects(int v1, int v2)` should also be renamed to something like `getEdgeObjectsBetweenVertices(int v1, int v2)` or `getEdgeObjectsConnectingVertices(int v1, int v2)`?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/VoltageLevel.java
flo-dup: Done.

geofjamg (approved):

PR #1806, opened 2021-08-19
Values for measurements and discrete measurements should be optional

MioRtia:

No

Bug fix

Values for measurements and discrete measurements are optional in the XSD and the checks but not during deserialization.

Deserializers do not try to read undefined values.

No

**Other information**:

### DIFF
--- a/iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/extensions/DiscreteMeasurementsXmlSerializer.java
+++ b/iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/extensions/DiscreteMeasurementsXmlSerializer.java
@@ -107,18 +107,20 @@ public class DiscreteMeasurementsXmlSerializer<I extends Identifiable<I>> extend
             adder.setTapChanger(DiscreteMeasurement.TapChanger.valueOf(tapChanger));
         }
         DiscreteMeasurement.ValueType valueType = DiscreteMeasurement.ValueType.valueOf(reader.getAttributeValue(null, "valueType"));
-        switch (valueType) {
-            case BOOLEAN:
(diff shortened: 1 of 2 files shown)


iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/extensions/DiscreteMeasurementsXmlSerializer.java
```diff
@@ -107,18 +107,20 @@ public void write(DiscreteMeasurements<I> extension, XmlWriterContext context) t
             adder.setTapChanger(DiscreteMeasurement.TapChanger.valueOf(tapChanger));
         }
         DiscreteMeasurement.ValueType valueType = DiscreteMeasurement.ValueType.valueOf(reader.getAttributeValue(null, "valueType"));
```
flo-dup, line 109: I think you should check before if `reader.getAttributeValue(null, "valueType")` is not null, otherwise `valueType` is required.

flo-dup (changes requested):

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/extensions/DiscreteMeasurementsXmlSerializer.java
MioRtia: `valueType` is  required, so I don't think we need to check if it is `null`?

flo-dup: Ah ok, got it! I thought that if value was optional, so should be valueType

flo-dup (approved):

PR #1808, opened 2021-08-31
Multiple ACLineSegments inside a Line container

marcosmc:

Feature

Not allowed multiple ACLineSegments inside a Line container.

In configurations with multiple ACLineSegment inside a Line container, a VoltageLevel without Substation is created for each ConnetivityNode whose container is the Line and all ACLineSegment are converted individually.


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -380,13 +381,108 @@ private void convertACLineSegmentsToLines(Context context, Set<String> delayedBo
             if (c.valid()) {
                 String node = c.boundaryNode();
                 if (node != null && !context.config().convertBoundary()) {
-                    context.boundary().addAcLineSegmentAtNode(line, node);
-                    delayedBoundaryNodes.add(node);
+                    String lineContainerId = line.getId("Line");
+                    if (lineContainerId != null) {
+                        delayedLines.computeIfAbsent(lineContainerId, l -> new PropertyBags()).add(line);
```
zamarrenolm, line 386: Here we are recording ONLY `ACLineSegments` that are contained inside a `cim:Line` AND have one end at the boundary. We could have `ACLineSegments` inside the network, none of the two ends placed at the boundary.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -380,13 +381,108 @@ private void convertACLineSegmentsToLines(Context context, Set<String> delayedBo
...
+                    ACLineSegmentConversion c = new ACLineSegmentConversion(line, context);
+                    context.boundary().addAcLineSegmentAtNode(line, c.boundaryNode());
+                    delayedBoundaryNodes.add(c.boundaryNode());
+                });
+                delayedLines.remove(lineContainerId, p);
+            }
+        });
+    }
+
+    private void convertLine(Context context, String lineId, PropertyBags lineSegments) {
```
zamarrenolm, line 411: I am not sure that we need the approach of delaying the conversion of `ACLineSegments` found in `cim:Line` containers. For me, they should be converted using the current existing considerations (could be located at boundaries, could be merged with others at boundary, ...). The only consideration that we must add is that one of the ends will be a fictitious voltage level, that we create as we need it. We should create a fictitious voltage level for every `ConnectivityNode` inside the cim:Line; all the `ACLineSegments` sharing the same connectivity node in the CGMES model will be mapped to IIDM Lines in the same fictitious voltage level.

zamarrenolm (changes requested):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -370,25 +369,116 @@ private void addCimCharacteristics(Network network) {
...
         while (k.hasNext()) {
             PropertyBag line = k.next();
             if (LOG.isDebugEnabled()) {
                 LOG.debug(line.tabulateLocals("ACLineSegment"));
             }
             ACLineSegmentConversion c = new ACLineSegmentConversion(line, context);
+            if (c.valid()) {
+                String lineContainerId = line.getId("Line");
+                if (lineContainerId != null) {
+                    lineContainers.computeIfAbsent(lineContainerId, l -> new PropertyBags()).add(line);
```
zamarrenolm, line 383: After the latest changes, I still see no reason to delay the conversion of ACLS contained in a cim:Line. We just have to create the fictitious voltage levels and/or substations associated to them.

zamarrenolm (changes requested):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -389,6 +393,66 @@ private void convertACLineSegmentsToLines(Context context, Set<String> delayedBo
...
+                .setTopologyKind(
+                        context.nodeBreaker()
+                                ? TopologyKind.NODE_BREAKER
+                                : TopologyKind.BUS_BREAKER)
+                .setLowVoltageLimit(lowVoltageLimit)
+                .setHighVoltageLimit(highVoltageLimit)
+                .setId(id)
+                .setName(name)
+                .setEnsureIdUnicity(context.config().isEnsureIdAliasUnicity())
+                .add();
```
MioRtia, line 453: I think you can put the Line container ID as property of the contained Voltage Level

MioRtia (approved):
One small remark but otherwise it is okay for me!

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
zamarrenolm: Done. We also rearranged a little bit the code.

zamarrenolm (approved):

PR #1811, opened 2021-09-03
Cgmes. Voltage and angle in 3wt star buses and boundary nodes

marqueslanauja:

Feature

Voltage and angle in 3wt star buses and in boundary nodes are not calculated .

Voltage and angle in 3wt star buses and in boundary nodes are calculated and set as properties `v` and `angle` at the end of the CGMES conversion process.

**Other information**:


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/DanglingLineBoundaryImpl.java
```diff
@@ -25,27 +25,42 @@ public DanglingLineBoundaryImpl(DanglingLine parent) {
...
+            return Math.toDegrees(danglingLineData.getBoundaryBusTheta());
+        }
         Terminal t = parent.getTerminal();
         Bus b = t.getBusView().getBus();
         return new SV(t.getP(), t.getQ(), getV(b), getAngle(b), Branch.Side.ONE).otherSideA(parent, true);
     }
 
     @Override
     public double getP() {
+        if (valid(parent.getP0(), parent.getQ0())) {
```
zamarrenolm, line 51: If we have explicit information about boundary injection, we use it. If it is not valid, we rely on flow and voltage at network side to calculate boundary values.

zamarrenolm (approved):

annetill (approved):

PR #1819, opened 2021-09-08
Allow to delay network validation

### LINKED ISSUE #1798: [Network Validation] Proposition for delayed validations (use cases and specifications)
Feature

Strict validation is enforced:
- at the creation of a network component (when the `add` method is called)
- when a network component is updated (when a network component's setter is called)

In some cases, we can have:
- networks in a transitory state: it does not contain all the needed information for all network components at the same time (can be the case real-time asynchronous information for example)
- slightly faulty networks (min>max, missing field that can be 0, etc)
- incomplete networks (can be the case for state estimation for example)

* **What do we need?**
- A mechanism for at least switch checks from `WARN` to `THROW_EXCEPTION` and vice versa*
- A method that returns a boolean to check if a network is valid or not 
```java

public class Network {
....
/**
 * Run all the checks and if one fails, returns false. If all succeed, return true.
 */
boolean isValid();
...
}
```
- A method that generate a report of all issues/inconsistencies of an invalid network
```java

public class Network {
....
/**
 * Generate logs about issues of inconsistencies of the network. If there is none, does nothing.
 */
void reportValidity();
...
}
```
- A method that run all the checks of the network at once
```java

public class Network {
....
/**
 * Throw an exception if the network is invalid
 */
void checkValidity();
...
}
```
- Optional: a mechanism to do simple fixes (? to be specified)

There is also the question of if we need to allow users to create a totally custom network validation but as we do not have use case for that, we might not consider it.

  - PowSyBl Version: 4.3.1

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)* 

@mathbagu @berthaultval @marifunf Please feel free to specify your use case if my description was not precise enough.

*PR #1789 has been opened for an idea of what can be done and/or a basis for discussion for this point in particular.

Any feedback is appreciated!

miovd:

Yes, fix #1798 

Feature

A network must be valid at all time (i.e. validate equipment checks at all time).

A network can be set to be invalid.
There are several available methods:
- `enableValidationChecks(boolean)` allows to disable or enable validation checks. By default, validation checks are enabled.
- `runValidationChecks()` runs all the validation checks on the network and throws a `ValidationException` if they fail
- `reportValidationChecks()` runs all the validation checks on the network and logs all the failed checks
- `getValidationStatus()` returns the validation status of the network. It can be `VALID`, `INVALID` or `UNCHECKED`. If validation checks have never been disabled, the validation status is `VALID`. When you disable the validation checks, each time you create or modify a "checkable" equipment, the validation status becomes `UNCHECKED`. When the network is `UNCHECKED`, you can run `updateValidationStatus()` to run all the validation checks. If they fail, the network is `INVALID`. If they succeed, the network is `VALID`

**Other information**:


geofjamg (changes requested):
2 remarks:
 - if only real identified use case is to delay validation until state estimation, why do we allow to disable all checks? Why not just the one identified important for a state estimation? Here there is some structural checks that could be delayed,why ? What is the use case for that? We should be conservative and step by step allow delaying some more checks when a real use case come up.
 - this is more a question, I just read quickly the code and I wonder what happen if I change the valid status in the XIIDM and import? Could I 'hack" the library by spoofing the validation status? Is the validation always re-run after XIIDM import to prevent hacking?

MioRtia:
> 2 remarks:
> 
>     * if only real identified use case is to delay validation until state estimation, why do we allow to disable all checks? Why not just the one identified important for a state estimation? Here there is some structural checks that could be delayed,why ? What is the use case for that? We should be conservative and step by step allow delaying some more checks when a real use case come up.
> 
>     * this is more a question, I just read quickly the code and I wonder what happen if I change the valid status in the XIIDM and import? Could I 'hack" the library by spoofing the validation status? Is the validation always re-run after XIIDM import to prevent hacking?

Thank you for your feedback.

Indeed, this is a good first point, maybe it is worth to review all check one by one and view which ones are worth disabling. We discussed it but we may have several validation profiles for specific processes.

For your second point, when the imported validation status is not consistent with the calculated validation status, an exception is thrown (as when the validation status is unchecked, the validation is calculated at the end of the import). Therefore, there is no possibility to "hack" the validation status.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/TapChanger.java
```diff
@@ -50,6 +51,8 @@
      */
     C setTapPosition(int tapPosition);
 
+    C removeTapPosition();
```
annetill, line 54: There is a linked between this change and the subject of this PR? I am not sure to understand.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationLevel.java
```diff
@@ -0,0 +1,21 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.iidm.network;
+
+/**
+ * @author Miora Ralambotiana <miora.ralambotiana at rte-france.com>
+ */
+public enum ValidationLevel {
+    STATE_ESTIMATION,
```
annetill, line 13: As explained to @geofjamg the distinction between levels `STATE_ESTIMATION` and `LOADFLOW` is not understandable and relevant for me. A network after state estimation is electronically coherent. I think that the level where we don't have all the steady state hypothesis is more a `SCADA` level. After SCADA, we collect from different sources the hypothesis needed to run a state estimation.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -31,18 +40,66 @@ private static ValidationException createInvalidValueException(Validable validab
...
+        return checkHvdcActivePowerSetpoint(validable, activePowerSetpoint, throwException, Reporter.NO_OP);
     }
 
-    public static void checkHvdcActivePowerSetpoint(Validable validable, double activePowerSetpoint) {
+    public static ValidationLevel checkHvdcActivePowerSetpoint(Validable validable, double activePowerSetpoint, boolean throwException, Reporter reporter) {
         if (Double.isNaN(activePowerSetpoint)) {
-            throw createInvalidValueException(validable, activePowerSetpoint, "active power setpoint");
+            throwExceptionOrLogErrorForInvalidValue(validable, activePowerSetpoint, ACTIVE_POWER_SETPOINT, throwException, reporter);
+            return ValidationLevel.STATE_ESTIMATION;
         } else if (activePowerSetpoint < 0) {
```
annetill, line 98: A negative value is forbidden for both validation levels, not just for loadflow level.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -66,33 +123,58 @@ public static void checkActivePowerLimits(Validable validable, double minP, doub
...
+        ValidationLevel validationLevel = ValidationLevel.LOADFLOW;
         if (regulating && Double.isNaN(targetDeadband)) {
-            throw new ValidationException(validable, "Undefined value for target deadband of regulating " + validableType);
+            throwExceptionOrLogError(validable, "Undefined value for target deadband of regulating " + validableType, throwException, reporter);
+            validationLevel = ValidationLevel.min(validationLevel, ValidationLevel.STATE_ESTIMATION);
         }
         if (targetDeadband < 0) {
-            throw new ValidationException(validable, "Unexpected value for target deadband of " + validableType + ": " + targetDeadband + " < 0");
+            throwExceptionOrLogError(validable, "Unexpected value for target deadband of " + validableType + ": " + targetDeadband + " < 0", throwException, reporter);
+            validationLevel = ValidationLevel.min(validationLevel, ValidationLevel.STATE_ESTIMATION);
```
annetill, line 138: Same comment as before, it should be forbidden to have a target deadband negative, for both level.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -66,33 +123,58 @@ public static void checkActivePowerLimits(Validable validable, double minP, doub
...
+            throwExceptionOrLogError(validable, "voltage regulator status is not set", throwException, reporter);
+            return ValidationLevel.STATE_ESTIMATION;
         }
-        boolean bVoltageRegulatorOn = voltageRegulatorOn; // make sonar happy java:S5411 Boxed "Boolean" should be avoided in boolean expressions
-        if (bVoltageRegulatorOn) {
-            if (Double.isNaN(voltageSetpoint) || voltageSetpoint <= 0) {
-                throw createInvalidValueException(validable, voltageSetpoint, "voltage setpoint", "voltage regulator is on");
-            }
-            return false;
+        if (voltageRegulatorOn && (Double.isNaN(voltageSetpoint) || voltageSetpoint <= 0)) {
```
annetill, line 152: Same here: a negative voltage set point should be forbidden for both level.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -66,33 +123,58 @@ public static void checkActivePowerLimits(Validable validable, double minP, doub
...
-    public static void checkVoltageControl(Validable validable, Boolean voltageRegulatorOn, double voltageSetpoint, double reactivePowerSetpoint) {
-        if (checkVoltageControl(validable, voltageRegulatorOn, voltageSetpoint) && Double.isNaN(reactivePowerSetpoint)) {
-            throw createInvalidValueException(validable, reactivePowerSetpoint, "reactive power setpoint", "voltage regulator is off");
+    public static ValidationLevel checkVoltageControl(Validable validable, Boolean voltageRegulatorOn, double voltageSetpoint, double reactivePowerSetpoint, boolean throwException, Reporter reporter) {
+        if (voltageRegulatorOn == null) {
+            throwExceptionOrLogError(validable, "voltage regulator status is not set", throwException, reporter);
+            return ValidationLevel.STATE_ESTIMATION;
         }
+        if (voltageRegulatorOn) {
+            if (Double.isNaN(voltageSetpoint) || voltageSetpoint <= 0) {
```
annetill, line 169: Same remark here for voltageSetpoint negative.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -256,25 +349,40 @@ public static void checkBPerSection(Validable validable, double sectionB) {
...
-            throw new ValidationException(validable,
-                    "the current number (" + currentSectionCount
-                            + ") of section should be lesser than the maximum number of section ("
-                            + maximumSectionCount + ")");
+        if (currentSectionCount == null) {
+            throwExceptionOrLogError(validable, "the current number of section is undefined", throwException, reporter);
+            validationLevel = ValidationLevel.min(validationLevel, ValidationLevel.STATE_ESTIMATION);
+        } else {
+            if (currentSectionCount < 0) {
+                throwExceptionOrLogError(validable, "the current number of section (" + currentSectionCount
```
annetill, line 373: Same remark: currentSectionCount < 0 should be forbidden at all.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -256,25 +349,40 @@ public static void checkBPerSection(Validable validable, double sectionB) {
...
+                        + ") should be greater than or equal to 0", throwException, reporter);
+                validationLevel = ValidationLevel.min(validationLevel, ValidationLevel.STATE_ESTIMATION);
+            }
+            if (currentSectionCount > maximumSectionCount) {
```
annetill, line 377: Same remark: should be forbidden at all. Only a transitory state is allowed where the current step is not defined.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -256,25 +349,40 @@ public static void checkBPerSection(Validable validable, double sectionB) {
...
-        }
+    public static ValidationLevel checkSections(Validable validable, Integer currentSectionCount, int maximumSectionCount) {
+        return checkSections(validable, currentSectionCount, maximumSectionCount, true);
+    }
+
+    public static ValidationLevel checkSections(Validable validable, Integer currentSectionCount, int maximumSectionCount, boolean throwException) {
+        return checkSections(validable, currentSectionCount, maximumSectionCount, throwException, Reporter.NO_OP);
+    }
+
+    public static ValidationLevel checkSections(Validable validable, Integer currentSectionCount, int maximumSectionCount, boolean throwException, Reporter reporter) {
```
annetill, line 365: Indeed it is wierd that this method depends on maximumSectionCount.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -330,64 +450,98 @@ public static void checkBmax(Validable validable, double bMax) {
...
+                                                                          boolean regulating, boolean throwException, Reporter reporter) {
         if (regulating && tapChangersNotIncludingTheModified.stream().anyMatch(TapChanger::isRegulating)) {
-            throw new ValidationException(validable, "Only one regulating control enabled is allowed");
+            throwExceptionOrLogError(validable, UNIQUE_REGULATING_TAP_CHANGER_MSG, throwException, reporter);
+            return ValidationLevel.STATE_ESTIMATION;
         }
+        return ValidationLevel.LOADFLOW;
     }
 
     public static void checkConvertersMode(Validable validable, HvdcLine.ConvertersMode converterMode) {
```
annetill, line 547: You can have a transitory mode where the converter mode is not already defined.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -330,64 +450,98 @@ public static void checkBmax(Validable validable, double bMax) {
...
+                                                                 Terminal regulationTerminal, double targetV, Network network, boolean throwException,
+                                                                 Reporter reporter) {
+        ValidationLevel validationLevel = ValidationLevel.LOADFLOW;
         if (regulating) {
             if (Double.isNaN(targetV)) {
-                throwExceptionOrWarningForRtc(validable, loadTapChangingCapabilities, "a target voltage has to be set for a regulating ratio tap changer");
+                validationLevel = ValidationLevel.min(validationLevel, errorOrWarningForRtc(validable, loadTapChangingCapabilities, "a target voltage has to be set for a regulating ratio tap changer", throwException, reporter));
             }
             if (targetV <= 0) {
-                throwExceptionOrWarningForRtc(validable, loadTapChangingCapabilities, "bad target voltage " + targetV);
```
annetill, line 348: Should be forbidden.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -424,4 +578,152 @@ public static void checkLossFactor(Validable validable, float lossFactor) {
             throw new ValidationException(validable, "loss factor must be >= 0 and <= 100");
         }
     }
+
+    private static ValidationLevel checkRtc(Validable validable, RatioTapChanger rtc, Network network, boolean throwException, Reporter reporter) {
+        ValidationLevel validationLevel = ValidationLevel.LOADFLOW;
+        if (rtc.getTapPosition().isEmpty()) {
+            throwExceptionOrLogError(validable, "tap position is not set", throwException, reporter);
+            validationLevel = ValidationLevel.min(validationLevel, ValidationLevel.STATE_ESTIMATION);
```
annetill, line 586: I don't getwhy we need a min...

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -424,4 +578,152 @@ public static void checkLossFactor(Validable validable, float lossFactor) {
...
+        }
+        if (rtc.getTapPosition().isPresent()) {
+            int tapPosition = rtc.getTapPosition().getAsInt();
+            int highTapPosition = rtc.getLowTapPosition() + rtc.getAllSteps().size() - 1;
+            if (tapPosition < rtc.getLowTapPosition() || tapPosition > highTapPosition) {
+                throwExceptionOrLogError(validable, "incorrect tap position "
+                        + tapPosition + " [" + rtc.getLowTapPosition() + ", "
+                        + highTapPosition + "]", throwException, reporter);
+                validationLevel = ValidationLevel.min(validationLevel, ValidationLevel.STATE_ESTIMATION);
```
annetill, line 595: This should be forbidden even for validation level state estimation.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -424,4 +578,152 @@ public static void checkLossFactor(Validable validable, float lossFactor) {
...
+            }
+        }
+        validationLevel = ValidationLevel.min(validationLevel, checkRatioTapChangerRegulation(validable, rtc.isRegulating(), rtc.hasLoadTapChangingCapabilities(), rtc.getRegulationTerminal(), rtc.getTargetV(), network, throwException, reporter));
+        validationLevel = ValidationLevel.min(validationLevel, checkTargetDeadband(validable, "ratio tap changer", rtc.isRegulating(), rtc.getTargetDeadband(), throwException, reporter));
+        return validationLevel;
+    }
+
+    private static ValidationLevel checkPtc(Validable validable, PhaseTapChanger ptc, Network network, boolean throwException, Reporter reporter) {
```
annetill, line 603: Same remarks for PTC.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -424,4 +578,152 @@ public static void checkLossFactor(Validable validable, float lossFactor) {
...
+                .map(ThreeWindingsTransformer.Leg::getRatioTapChanger)
+                .filter(Objects::nonNull)
+                .filter(TapChanger::isRegulating)
+                .count()
+                + twt.getLegStream()
+                .map(ThreeWindingsTransformer.Leg::getPhaseTapChanger)
+                .filter(Objects::nonNull)
+                .filter(TapChanger::isRegulating)
+                .count();
+        if (regulatingTc > 1) {
```
annetill, line 644: This should be forbidden.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -424,4 +578,152 @@ public static void checkLossFactor(Validable validable, float lossFactor) {
...
+        ValidationLevel validationLevel = ValidationLevel.LOADFLOW;
+        if (twt.hasRatioTapChanger()) {
+            validationLevel = ValidationLevel.min(validationLevel, checkRtc(validable, twt.getRatioTapChanger(), twt.getNetwork(), throwException, reporter));
+        }
+        if (twt.hasPhaseTapChanger()) {
+            validationLevel = ValidationLevel.min(validationLevel, checkPtc(validable, twt.getPhaseTapChanger(), twt.getNetwork(), throwException, reporter));
+        }
+        if (twt.getOptionalRatioTapChanger().map(RatioTapChanger::isRegulating).orElse(false)
+                && twt.getOptionalPhaseTapChanger().map(PhaseTapChanger::isRegulating).orElse(false)) {
+            throwExceptionOrLogError(validable, UNIQUE_REGULATING_TAP_CHANGER_MSG, throwException, reporter);
```
annetill, line 661: This should be fobbiden.

annetill (changes requested):
I have stopped my review at class `ValidationUtil`. For me we have indeed three levels : the loadflow one, the scada one and the one with an exception during import. There is a confusion here I think. Furthermore, have you done something for the switch status ? For sure we have to test this PR to see if we understand well the use case.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/TapChanger.java
miovd: Yes, the issue is that by making tap position optional, if we do not have this method, it means that once tap position is set, a user can never go back. Not sure if this is useful but that behavior seemed weird to me.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
miovd: It doesn't really matter today as we only have two validation level, but if we have others later, it allows us to only keep the lowest validation level between the current one and the one computed from the check

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
miovd: Not really as it checks if the current section count is feasible

action/action-dsl/src/test/java/com/powsybl/action/dsl/ActionDslLoaderTest.java
```diff
@@ -84,11 +84,11 @@ public void testFixTapDslExtension() {
         assertNotNull(fixedTapAction);
         addPhaseShifter(0);
         PhaseTapChanger phaseTapChanger = network.getTwoWindingsTransformer("NGEN_NHV1").getPhaseTapChanger();
-        assertEquals(0, phaseTapChanger.getTapPosition());
+        assertEquals(0, phaseTapChanger.getTapPosition().orElseThrow(AssertionError::new));
```
geofjamg, line 87: I known that there is many places in powsybl where we use AssertionError fir this kind of purpose but this is a bad practice. java.lang.Error are not easy (and very rarely) catched and could be mistaken with VM internal java.lang.Error so for error that could come up we shoud use a standard runtime exception. (for instance it is not the same a a switch case default that cannot be thrown because of cases have been covered)

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Network.java
```diff
@@ -859,4 +860,48 @@ default int getConnectableCount() {
...
+     * If network not valid and <code>throwsException</code> is <code>true</code>, check if each network component is valid. A {@link ValidationException} is thrown with an explicit message if one network component is not valid.<br>
+     * If all network components are valid, network validation status is updated to true.
+     * Return the network validation status.
+     */
+    default ValidationLevel runValidationChecks(boolean throwsException) {
+        return runValidationChecks(throwsException, Reporter.NO_OP);
+    }
+
+    /**
+     * If network is valid, do nothing.<br>
```
geofjamg, line 885: How do we get a report in that case even if we have reach the minimum validation level?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/TapChanger.java
geofjamg: Javadoc is missing

geofjamg: I would call it unsetTapPostion, I think "remove" is not correct

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -424,4 +586,149 @@ public static void checkLossFactor(Validable validable, float lossFactor) {
...
+            validationLevel = ValidationLevel.min(validationLevel, checkPtc(validable, twt.getPhaseTapChanger(), twt.getNetwork(), throwException, reporter));
+        }
+        if (twt.getOptionalRatioTapChanger().map(RatioTapChanger::isRegulating).orElse(false)
+                && twt.getOptionalPhaseTapChanger().map(PhaseTapChanger::isRegulating).orElse(false)) {
+            throw new ValidationException(validable, UNIQUE_REGULATING_TAP_CHANGER_MSG);
+        }
+        return validationLevel;
+    }
+
+    private static ValidationLevel checkIdentifiable(Identifiable<?> identifiable, ValidationLevel previous, boolean throwException, Reporter reporter) {
```
geofjamg, line 671: Code is really hard to read with all these min(). I have nothing to propose right now but I wish we could find a simpler way to deal with this.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChanger.java
```diff
@@ -7,23 +7,24 @@
...
+import java.util.OptionalInt;
 
 /**
  *
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
  */
 abstract class AbstractTapChanger<H extends TapChangerParent, C extends AbstractTapChanger<H, C, S>, S extends TapChangerStepImpl<S>> implements MultiVariantObject {
 
-    protected final Ref<? extends VariantManagerHolder> network;
+    protected final NetworkImpl network;
```
geofjamg, line 25: The purpose of Ref was to allow breaking revert link to network in case of merge. I think here you added a regression and this is redundant with networkRef which already allows you to get network in a safe manner.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChanger.java
```diff
@@ -43,21 +44,22 @@
...
         steps.forEach(s -> s.setParent(this));
         this.regulationTerminal = regulationTerminal;
-        int variantArraySize = network.get().getVariantManager().getVariantArraySize();
+        int variantArraySize = networkRef.get().getVariantManager().getVariantArraySize();
         this.tapPosition = new TIntArrayList(variantArraySize);
         this.regulating = new TBooleanArrayList(variantArraySize);
         this.targetDeadband = new TDoubleArrayList(variantArraySize);
         for (int i = 0; i < variantArraySize; i++) {
-            this.tapPosition.add(tapPosition);
+            this.tapPosition.add(tapPosition != null ? tapPosition : Integer.MIN_VALUE);
```
geofjamg, line 62: Maybe it would be more clear to just replace the TIntArray by a List<Integer> and performance loss will be neglictable?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChanger.java
```diff
@@ -78,32 +80,54 @@ public C setLowTapPosition(int lowTapPosition) {
...
-        String variantId = network.get().getVariantManager().getVariantId(variantIndex);
-        parent.getNetwork().getListeners().notifyUpdate(parent.getTransformer(), () -> getTapChangerAttribute() + ".tapPosition", variantId, oldValue, tapPosition);
+        int variantIndex = networkRef.get().getVariantIndex();
+        Integer oldValue = this.tapPosition.set(variantIndex, tapPosition);
+        if (oldValue == Integer.MIN_VALUE) {
+            oldValue = null;
+        }
+        String variantId = networkRef.get().getVariantManager().getVariantId(variantIndex);
+        network.getListeners().notifyUpdate(parent.getTransformer(), () -> getTapChangerAttribute() + ".tapPosition", variantId, oldValue, tapPosition);
+        network.invalidate();
```
geofjamg, line 117: Before the notification?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BatteryAdderImpl.java
```diff
@@ -89,20 +90,21 @@ public BatteryAdderImpl setMaxP(double maxP) {
      */
     @Override
     public BatteryImpl add() {
+        NetworkImpl network = getNetwork();
         String id = checkAndGetUniqueId();
         TerminalExt terminal = checkAndGetTerminal();
-        ValidationUtil.checkP0(this, p0);
-        ValidationUtil.checkQ0(this, q0);
+        network.setValidationLevelIfGreaterThan(ValidationUtil.checkP0(this, p0, network.getMinValidationLevel().compareTo(ValidationLevel.LOADFLOW) >= 0));
```
geofjamg, line 96: Maybe we could change the ValidationUtil API to avoid this pattern `network.getMinValidationLevel().compareTo(ValidationLevel.LOADFLOW) >= 0` everywhere in the code?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BatteryImpl.java
```diff
@@ -17,6 +17,8 @@
  */
 public class BatteryImpl extends AbstractConnectable<Battery> implements Battery, ReactiveLimitsOwner {
 
+    private final NetworkImpl network;
```
geofjamg, line 20: Use a Ref instead and get the network in the getter of setter, otherwise merging is broken.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BatteryImpl.java
```diff
@@ -76,19 +80,20 @@ public Battery setP0(double p0) {
...
     public Battery setQ0(double q0) {
-        ValidationUtil.checkQ0(this, q0);
-        int variantIndex = getNetwork().getVariantIndex();
+        ValidationUtil.checkQ0(this, q0, network.getMinValidationLevel().compareTo(ValidationLevel.LOADFLOW) >= 0);
+        int variantIndex = network.getVariantIndex();
         double oldValue = this.q0.set(variantIndex, q0);
-        String variantId = getNetwork().getVariantManager().getVariantId(variantIndex);
+        String variantId = network.getVariantManager().getVariantId(variantIndex);
         notifyUpdate("q0", variantId, oldValue, q0);
+        network.invalidate();
```
geofjamg, line 96: invalidate is confusing. First time I looked at the code it was not clear what kind of invalidation it was (connected components, a topo cache?)
`invalidateValidationLevel()` ?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractTransformerXml.java
```diff
@@ -223,11 +244,11 @@ private static void readTerminalRef(NetworkXmlReaderContext context, boolean[] h
     }
 
     private static void readSteps(NetworkXmlReaderContext context, StepConsumer consumer) {
-        double r = XmlUtil.readDoubleAttribute(context.getReader(), "r");
-        double x = XmlUtil.readDoubleAttribute(context.getReader(), "x");
-        double g = XmlUtil.readDoubleAttribute(context.getReader(), "g");
-        double b = XmlUtil.readDoubleAttribute(context.getReader(), "b");
-        double rho = XmlUtil.readDoubleAttribute(context.getReader(), "rho");
+        double r = XmlUtil.readOptionalDoubleAttribute(context.getReader(), "r");
```
geofjamg, line 247: Why are there read optionals?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractTransformerXml.java
```diff
@@ -193,7 +214,7 @@ protected static void readPhaseTapChanger(String name, PhaseTapChangerAdder adde
                             .setG(g)
                             .setB(b)
                             .setRho(rho));
-                    double alpha = XmlUtil.readDoubleAttribute(context.getReader(), "alpha");
+                    double alpha = XmlUtil.readOptionalDoubleAttribute(context.getReader(), "alpha");
```
geofjamg, line 217: Same?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/DanglingLineXml.java
```diff
@@ -111,12 +112,12 @@ protected DanglingLineAdder createAdder(VoltageLevel vl) {
...
     protected DanglingLine readRootElementAttributes(DanglingLineAdder adder, NetworkXmlReaderContext context) {
-        double p0 = XmlUtil.readDoubleAttribute(context.getReader(), "p0");
-        double q0 = XmlUtil.readDoubleAttribute(context.getReader(), "q0");
-        double r = XmlUtil.readDoubleAttribute(context.getReader(), "r");
-        double x = XmlUtil.readDoubleAttribute(context.getReader(), "x");
-        double g = XmlUtil.readDoubleAttribute(context.getReader(), "g");
-        double b = XmlUtil.readDoubleAttribute(context.getReader(), "b");
+        double p0 = XmlUtil.readOptionalDoubleAttribute(context.getReader(), "p0");
+        double q0 = XmlUtil.readOptionalDoubleAttribute(context.getReader(), "q0");
+        double r = XmlUtil.readOptionalDoubleAttribute(context.getReader(), "r");
```
geofjamg, line 117: ?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/LineXml.java
```diff
@@ -88,12 +90,12 @@ protected LineAdder createAdder(Network n) {
 
     @Override
     protected Line readRootElementAttributes(LineAdder adder, NetworkXmlReaderContext context) {
-        double r = XmlUtil.readDoubleAttribute(context.getReader(), "r");
-        double x = XmlUtil.readDoubleAttribute(context.getReader(), "x");
-        double g1 = XmlUtil.readDoubleAttribute(context.getReader(), "g1");
-        double b1 = XmlUtil.readDoubleAttribute(context.getReader(), "b1");
-        double g2 = XmlUtil.readDoubleAttribute(context.getReader(), "g2");
-        double b2 = XmlUtil.readDoubleAttribute(context.getReader(), "b2");
+        double r = XmlUtil.readOptionalDoubleAttribute(context.getReader(), "r");
```
geofjamg, line 93: ?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChanger.java
geofjamg: Same remark everywhere in the rest of the PR

geofjamg (changes requested):

iidm/iidm-scripting/src/main/groovy/com/powsybl/iidm/network/scripting/ShuntCompensatorExtension.groovy
```diff
@@ -58,8 +58,12 @@ class ShuntCompensatorExtension {
         }
     }
 
+    /*static int getSectionCount(ShuntCompensator self) {
+        self.getSectionCount().orElseThrow(AssertionError::new) // Throws a StackOverflow exception (cyclic method call)
+    }*/
```
miovd, line 63: @geofjamg Do you have any idea about how to prevent breaking changes in groovy for this API modification?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
miovd: Since there are only two validation levels at the moment, I think we can simply return SCADA when a check is not LOADFLOW

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
```diff
@@ -271,7 +271,7 @@ private static void writeShuntCompensators(Network network, String cimNamespace,
                 bPerSection = ((ShuntCompensatorLinearModel) s.getModel()).getBPerSection();
                 gPerSection = ((ShuntCompensatorLinearModel) s.getModel()).getGPerSection();
             }
-            ShuntCompensatorEq.write(s.getId(), s.getNameOrId(), s.getSectionCount(), s.getMaximumSectionCount(), s.getTerminal().getVoltageLevel().getNominalV(), s.getModelType(), bPerSection, gPerSection, cimNamespace, writer);
+            ShuntCompensatorEq.write(s.getId(), s.getNameOrId(), s.getSectionCount().orElseThrow(AssertionError::new), s.getMaximumSectionCount(), s.getTerminal().getVoltageLevel().getNominalV(), s.getModelType(), bPerSection, gPerSection, cimNamespace, writer);
```
annetill, line 274: It is a bit strange that for the export of EQ profile for shunt compensator, we depend on the section count. Same remark for tap changers and tap position. This PR indeed is done for an EQ export without steady state hypothesis.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
```diff
@@ -362,7 +362,7 @@ private static void writePhaseTapChanger(Identifiable<?> eq, PhaseTapChanger ptc
 
             int neutralStep = getPhaseTapChangerNeutralStep(ptc);
             String phaseTapChangerTableId = CgmesExportUtil.getUniqueId();
-            TapChangerEq.writePhase(tapChangerId, twtName + "_PTC", endId, ptc.getLowTapPosition(), ptc.getHighTapPosition(), neutralStep, ptc.getTapPosition(), neutralU, false, phaseTapChangerTableId, cimNamespace, writer);
+            TapChangerEq.writePhase(tapChangerId, twtName + "_PTC", endId, ptc.getLowTapPosition(), ptc.getHighTapPosition(), neutralStep, ptc.getTapPosition().orElseThrow(() -> new PowsyblException("SCADA network not supported")), neutralU, false, phaseTapChangerTableId, cimNamespace, writer);
```
annetill, line 365: For me, SCADA should be supported here.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/DebugPhaseTapChanger.java
```diff
@@ -59,7 +64,7 @@ private void explorePhaseTapPositions(
                     tx.getTerminal2().getBusView().getBus().getV(),
                     tx.getTerminal2().getBusView().getBus().getAngle()));
         }
-        int backup = tx.getPhaseTapChanger().getTapPosition();
+        int backup = tx.getPhaseTapChanger().getTapPosition().orElseThrow(() -> new PowsyblException("SCADA network not supported"));
```
annetill, line 67: Just an idea: can you provide the error message `SCADA validation level not supported` in a way or another through the API ?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Network.java
```diff
@@ -859,4 +860,48 @@ default int getConnectableCount() {
...
+     * If network not valid and <code>throwsException</code> is <code>true</code>, check if each network component is valid. A {@link ValidationException} is thrown with an explicit message if one network component is not valid.<br>
+     * If all network components are valid, network validation status is updated to true.
+     * Return the network validation status.
+     */
+    default ValidationLevel runValidationChecks(boolean throwsException, Reporter reporter) {
+        return ValidationLevel.LOADFLOW;
```
annetill, line 891: Why is it like this here: we don't run any check.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
```diff
@@ -117,7 +119,7 @@
      * Depends on the working variant.
      * @see VariantManager
      */
-    int getSectionCount();
+    OptionalInt getSectionCount();
```
annetill, line 122: This is a strong breaking change, do not forget to deal with the backward compatibility of groovy scripts at least. We have to ask the community about the potential impacts.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -330,70 +477,110 @@ public static void checkBmax(Validable validable, double bMax) {
...
+    public static ValidationLevel checkRatioTapChangerRegulation(Validable validable, boolean regulating, boolean loadTapChangingCapabilities,
+                                                                 Terminal regulationTerminal, double targetV, Network network, ValidationLevel validationLevel, Reporter reporter) {
+        return checkRatioTapChangerRegulation(validable, regulating, loadTapChangingCapabilities, regulationTerminal, targetV, network, validationLevel == ValidationLevel.LOADFLOW, reporter);
+    }
+
+    public static ValidationLevel checkRatioTapChangerRegulation(Validable validable, boolean regulating, boolean loadTapChangingCapabilities,
+                                                                 Terminal regulationTerminal, double targetV, Network network, boolean throwException,
+                                                                 Reporter reporter) {
+        ValidationLevel validationLevel = ValidationLevel.LOADFLOW;
         if (regulating) {
```
annetill, line 508: Where do you check this `regulating` ?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChanger.java
```diff
@@ -117,19 +131,25 @@ public S getStep(int tapPosition) {
...
-        return getStep(getTapPosition());
+        int position = tapPosition.get(network.get().getVariantIndex());
+        if (position == Integer.MIN_VALUE) {
+            return null;
+        }
+        return getStep(position);
     }
 
     public boolean isRegulating() {
         return regulating.get(network.get().getVariantIndex());
```
annetill, line 142: This boolean `regulating` is not part of the steady state hypotheses, as we have for generator voltage control for example?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevel.java
```diff
@@ -118,9 +118,6 @@ protected String getTypeDescription() {
 
         @Override
         public NodeBreakerView.SwitchAdder setKind(SwitchKind kind) {
-            if (kind == null) {
```
annetill, line 121: I am not sure to understand why these lines have been deleted.

loadflow/loadflow-validation/src/main/java/com/powsybl/loadflow/validation/ShuntCompensatorsValidation.java
```diff
@@ -90,7 +90,7 @@ public boolean checkShunts(ShuntCompensator shunt, ValidationConfig config, Vali
 
         double p = shunt.getTerminal().getP();
         double q = shunt.getTerminal().getQ();
-        int currentSectionCount = shunt.getSectionCount();
+        int currentSectionCount = shunt.getSectionCount().orElseThrow(AssertionError::new);
```
annetill, line 93: Why not `SCADA network not supported` here?

psse/psse-converter/src/main/java/com/powsybl/psse/converter/FixedShuntCompensatorConverter.java
```diff
@@ -88,7 +88,7 @@ private static int getStatus(ShuntCompensator fixedShunt) {
     }
 
     private static double getQ(ShuntCompensator fixedShunt) {
-        return shuntAdmittanceToPower(fixedShunt.getB(fixedShunt.getSectionCount()),
+        return shuntAdmittanceToPower(fixedShunt.getB(fixedShunt.getSectionCount().orElseThrow(AssertionError::new)),
```
annetill, line 91: Why not SCADA message here too ?

psse/psse-converter/src/main/java/com/powsybl/psse/converter/SwitchedShuntCompensatorConverter.java
```diff
@@ -296,7 +296,7 @@ private static int getStatus(ShuntCompensator switchedShunt) {
     }
 
     private static double getQ(ShuntCompensator switchedShunt) {
-        return shuntAdmittanceToPower(switchedShunt.getB(switchedShunt.getSectionCount()),
+        return shuntAdmittanceToPower(switchedShunt.getB(switchedShunt.getSectionCount().orElseThrow(AssertionError::new)),
```
annetill, line 299: Why not SCADA message here too ?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractConnectableXml.java
```diff
@@ -197,13 +197,13 @@ public static void readCurrentLimits(Integer index, CurrentLimitsAdder currentLi
     }
 
     static void writeActivePowerLimits(Integer index, ActivePowerLimits limits, XMLStreamWriter writer, IidmXmlVersion version,
-                                              ExportOptions exportOptions) throws XMLStreamException {
-        writeLoadingLimits(index, limits, writer, version.getNamespaceURI(), version, exportOptions, ACTIVE_POWER_LIMITS);
+                                              boolean valid, ExportOptions exportOptions) throws XMLStreamException {
```
annetill, line 200: Why do you need this `valide` boolean for operational limits ?

iidm/iidm-test/src/main/java/com/powsybl/iidm/network/test/ScadaNetworkFactory.java
```diff
@@ -0,0 +1,243 @@
...
+                .setMinP(100.0)
+                .setMaxP(200.0)
+                .add()
+                .add();
+        dl.getGeneration().newMinMaxReactiveLimits()
+                .setMinQ(100)
+                .setMaxQ(200)
+                .add();
+        Generator generator = vl.newGenerator()
+                .setVoltageRegulatorOn(true)
```
annetill, line 64: Where is the regulating terminal here? It is an equipment information. Same remark for shunt compensator, phase and ratio tap changers, VSC converter station. Without this information, the network is just invalid. It is what you want to show here?

annetill: Indeed this factory is not used, right?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
miovd: For now, I consider SCADA network not supported everywhere. We will make changes later

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
miovd: See above

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Network.java
miovd: For compatibility: in previous networks, all checks are void since the network is LOADFLOW

iidm/iidm-test/src/main/java/com/powsybl/iidm/network/test/ScadaNetworkFactory.java
miovd: It is used in `AbstractNetworkTest`

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractConnectableXml.java
miovd: For the namespace!

miovd:
Two big questions about this PR:
- should we change or keep initial values in adders?
- should we keep accessors (`getTapPosition`, `isRegulating`) with the same behavior and add `getOptional[...]` or should accessors return Optional?

sylvlecl:
**Some feedback on the first point:**
after more thought, I definitely think that we should keep the existing methods as is, and make them throw an exception.

Otherwise, I think 95% of user code will be cluttered with `orElseThrow` statements. Actually, I don't think there will be a lot of cases where we want to do something else than throwing ? In processes that don't require the values to be present, they will just not be used, so the getter will not be used. In processes that require the value, in general we will want to throw.

But indeed it means that we need to duplicated every getter which is a little painful ... 
About the naming, maybe it would sound better to name them `get***IfExist` or something similar, instead of `getOptional***`.

**About performances implications**
I wonder if it the implementation will have significant impact on memory usage and performances overall.
For example, using `List<Boolean>` instead of `TBooleanArrayList` will use a lot more memory. I don't know if it will be significant since most of the network memory is probably occupied by strings.

I also wonder what will be the performance impact on accessing those values (with unboxing), for example in open loadflow. I presume it would not be significant overall, but it deserves a check.

If the impact is actually significant, it might be a more efficient option to use masks, but i's gonna get ugly.

cgmes/cgmes-conversion/src/test/resources/cim14/7buses.xiidm
```diff
@@ -18,7 +18,7 @@
             <iidm:load id="_FP.ANC12_EC" name="FP.ANC12" loadType="UNDEFINED" p0="240.0" q0="2.4000000953674316" bus="_FP.AND11_TN" connectableBus="_FP.AND11_TN" p="240.0" q="2.4000000953674316"/>
         </iidm:voltageLevel>
         <iidm:twoWindingsTransformer id="_FP.AND11-FTDPRA11-1_PT" name="FP.AND11-FTDPRA11-1" r="0.009999940171837807" x="0.9999989867210388" g="0.0" b="0.0" ratedU1="380.0" ratedU2="380.0" bus1="_FP.AND11_TN" connectableBus1="_FP.AND11_TN" voltageLevelId1="_FP.AND11_VL" bus2="_FTDPRA11_TN" connectableBus2="_FTDPRA11_TN" voltageLevelId2="_FTDPRA11_VL">
-            <iidm:phaseTapChanger lowTapPosition="1" tapPosition="17" regulationMode="FIXED_TAP">
+            <iidm:phaseTapChanger lowTapPosition="1" tapPosition="17" regulationMode="FIXED_TAP" regulating="false">
```
annetill, line 21: Why do we have this redundant information regulating="false" as we already know that the regulationMode is fixed tap ?

cgmes/cgmes-conversion/src/test/resources/cim14/ieee14.xiidm
```diff
@@ -94,7 +94,7 @@
             <iidm:load id="_LOAD___4_EC" name="LOAD   4" loadType="UNDEFINED" p0="47.79999923706055" q0="-3.9000000953674316" bus="_BUS____4_TN" connectableBus="_BUS____4_TN" p="47.79999923706055" q="-3.9000000953674316"/>
         </iidm:voltageLevel>
         <iidm:twoWindingsTransformer id="_BUS____4-BUS____9-1_PT" name="BUS    4-BUS    9-1" r="0.0" x="1.0591881275177002" g="0.0" b="0.0" ratedU1="69.0" ratedU2="13.800000190734863" bus1="_BUS____4_TN" connectableBus1="_BUS____4_TN" voltageLevelId1="_BUS____4_VL" bus2="_BUS____9_TN" connectableBus2="_BUS____9_TN" voltageLevelId2="_BUS____9_VL">
-            <iidm:ratioTapChanger lowTapPosition="1" tapPosition="4" loadTapChangingCapabilities="false">
+            <iidm:ratioTapChanger lowTapPosition="1" tapPosition="4" loadTapChangingCapabilities="false" regulating="false">
```
annetill, line 97: The rule before was nothing means regulating at false ?

miovd: Yes.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -66,33 +132,80 @@ public static void checkActivePowerLimits(Validable validable, double minP, doub
...
 
-    public static void checkVoltageControl(Validable validable, Boolean voltageRegulatorOn, double voltageSetpoint, double reactivePowerSetpoint) {
-        if (checkVoltageControl(validable, voltageRegulatorOn, voltageSetpoint) && Double.isNaN(reactivePowerSetpoint)) {
-            throw createInvalidValueException(validable, reactivePowerSetpoint, "reactive power setpoint", "voltage regulator is off");
+    public static ValidationLevel checkVoltageControl(Validable validable, Boolean voltageRegulatorOn, double voltageSetpoint, double reactivePowerSetpoint, ValidationLevel validationLevel, Reporter reporter) {
+        return checkVoltageControl(validable, voltageRegulatorOn, voltageSetpoint, reactivePowerSetpoint, validationLevel == ValidationLevel.LOADFLOW, reporter);
+    }
+
+    public static ValidationLevel checkVoltageControl(Validable validable, Boolean voltageRegulatorOn, double voltageSetpoint, double reactivePowerSetpoint, boolean throwException, Reporter reporter) {
+        if (voltageRegulatorOn == null) {
```
annetill, line 192: Same comment. I am not sure to well understand the general rule. If `null` means validation status equals to SCADA or simply invalid.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -330,70 +486,114 @@ public static void checkBmax(Validable validable, double bMax) {
...
+    public static ValidationLevel checkOnlyOneTapChangerRegulatingEnabled(Validable validable,
+                                                                          Set<TapChanger<?, ?>> tapChangersNotIncludingTheModified, Boolean regulating, boolean throwException) {
+        return checkOnlyOneTapChangerRegulatingEnabled(validable, tapChangersNotIncludingTheModified, regulating, throwException, Reporter.NO_OP);
     }
 
-    public static void checkOnlyOneTapChangerRegulatingEnabled(Validable validable,
-                                                               Set<TapChanger> tapChangersNotIncludingTheModified, boolean regulating) {
+    public static ValidationLevel checkOnlyOneTapChangerRegulatingEnabled(Validable validable, Set<TapChanger<?, ?>> tapChangersNotIncludingTheModified,
+                                                                          Boolean regulating, boolean throwException, Reporter reporter) {
+        if (regulating == null) {
```
annetill, line 575: Same remark here.

annetill (commented):
Just some comments, but my review was not technical.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -291,20 +434,33 @@ public static void checkRatedU2(Validable validable, double ratedU2) {
...
+    public static ValidationLevel checkSvcRegulator(Validable validable, double voltageSetpoint, double reactivePowerSetpoint,
+                                                    StaticVarCompensator.RegulationMode regulationMode, ValidationLevel validationLevel, Reporter reporter) {
+        return checkSvcRegulator(validable, voltageSetpoint, reactivePowerSetpoint, regulationMode, validationLevel == ValidationLevel.LOADFLOW, reporter);
+    }
+
+    public static ValidationLevel checkSvcRegulator(Validable validable, double voltageSetpoint, double reactivePowerSetpoint,
+                                                    StaticVarCompensator.RegulationMode regulationMode, boolean throwException, Reporter reporter) {
         if (regulationMode == null) {
-            throw new ValidationException(validable, "Regulation mode is invalid");
+            throwExceptionOrLogError(validable, "Regulation mode is invalid", throwException, reporter);
```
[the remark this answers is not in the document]
miovd: Can't regulationMode be null?

psse/psse-converter/src/main/java/com/powsybl/psse/converter/TransformerConverter.java
```diff
@@ -450,7 +450,7 @@ private static void tapChangerToRatioTapChanger(TapChanger tapChanger, RatioTapC
                 .setB(step.getB1())
                 .setG(step.getG1())
                 .endStep());
-        rtc.add();
+        rtc.setRegulating(false).add();
```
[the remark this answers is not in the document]
miovd: @annetill I changed it back as it was to prevent regression.

psse/psse-converter/src/main/java/com/powsybl/psse/converter/SwitchedShuntCompensatorConverter.java
```diff
@@ -67,6 +61,7 @@ void create() {
         modelAdder.add();
 
         adder.setBus(psseSwitchedShunt.getStat() == 1 ? busId : null);
+        adder.setVoltageRegulatorOn(false);
```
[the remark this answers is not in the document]
miovd: Same remark.

iidm/iidm-xml-converter/src/test/resources/V1_7/danglingLineWithGeneration.xml
```diff
@@ -8,7 +8,7 @@
             <iidm:generator id="G" energySource="OTHER" minP="0.0" maxP="100.0" voltageRegulatorOn="true" targetP="50.0" targetV="100.0" targetQ="30.0" bus="BUS" connectableBus="BUS">
                 <iidm:minMaxReactiveLimits minQ="-1.7976931348623157E308" maxQ="1.7976931348623157E308"/>
             </iidm:generator>
-            <iidm:danglingLine id="DL" p0="50.0" q0="30.0" r="10.0" x="1.0" g="1.0E-4" b="1.0E-5" generationMinP="0.0" generationMaxP="900.0" generationVoltageRegulationOn="true" generationTargetP="440.0" generationTargetV="400.0" bus="BUS" connectableBus="BUS">
+            <iidm:danglingLine id="DL" p0="50.0" q0="30.0" r="10.0" x="1.0" g="1.0E-4" b="1.0E-5" withGeneration="true" generationMinP="0.0" generationMaxP="900.0" generationVoltageRegulationOn="true" generationTargetP="440.0" generationTargetV="400.0" bus="BUS" connectableBus="BUS">
```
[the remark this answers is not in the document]
miovd: Indeed, you're right it is not useful anymore since `voltageRegulationOn` might not be optional anymore for generation

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -424,4 +632,131 @@ public static void checkLossFactor(Validable validable, float lossFactor) {
...
+            Validable validable = (Validable) identifiable;
+            if (identifiable instanceof Battery) {
+                Battery battery = (Battery) identifiable;
+                validationLevel = ValidationLevel.min(validationLevel, checkP0(validable, battery.getP0(), throwException, reporter));
+                validationLevel = ValidationLevel.min(validationLevel, checkQ0(validable, battery.getQ0(), throwException, reporter));
+            } else if (identifiable instanceof DanglingLine) {
+                DanglingLine danglingLine = (DanglingLine) identifiable;
+                validationLevel = ValidationLevel.min(validationLevel, checkP0(validable, danglingLine.getP0(), throwException, reporter));
+                validationLevel = ValidationLevel.min(validationLevel, checkQ0(validable, danglingLine.getQ0(), throwException, reporter));
+                DanglingLine.Generation generation = danglingLine.getGeneration();
```
marcosmc, line 687: What happens if a dangling line has no generation?

marcosmc (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
miovd: Good catch!

marcosmc (approved):

zamarrenolm (commented):
I propose to change the validation level names:

- Instead of `SCADA`, use `EQUIPMENT_CHARACTERISTICS` or `EQUIPMENT`. `SCADA` is one potential source of data. Another could be a Power System Model database (for example). What is being validated at this level is the definition of the equipment, or the equipment characteristics, independent of its original source.
- Instead of `LOADFLOW`, use `STEADY_STATE_HYPOTHESIS`, `STEADY_STATE_INPUTS` or `LOADFLOW_INPUTS`. What is being validated is the information required to build a steady state of the network. They are hypothesis because we do not know if they are valid until we run a Loadflow. Calling it `LOADFLOW` could move to confusion: the user may think the Network contains a valid Loadflow solution.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractConnectableXml.java
```diff
@@ -216,7 +216,12 @@ public static void writeCurrentLimits(Integer index, CurrentLimits limits, XMLSt
 
     public static void writeCurrentLimits(Integer index, CurrentLimits limits, XMLStreamWriter writer, IidmXmlVersion version,
                                           ExportOptions exportOptions) throws XMLStreamException {
-        writeCurrentLimits(index, limits, writer, version.getNamespaceURI(), version, exportOptions);
+        writeCurrentLimits(index, limits, writer, version, true, exportOptions);
+    }
+
+    public static void writeCurrentLimits(Integer index, CurrentLimits limits, XMLStreamWriter writer, IidmXmlVersion version,
+                                          boolean valid, ExportOptions exportOptions) throws XMLStreamException {
+        writeCurrentLimits(index, limits, writer, version.getNamespaceURI(valid), version, exportOptions);
```
marcosmc, line 224: It is necessary to pass the "valid" flag in order to obtain the correct namespace. If it is not passed, and an attempt is made to export a network with SCADA level in ValidationLevel, it ends with an exception.

miovd: Thank you for yor feedback! I think the methods used in the export is using the valid boolean while writing current limits, the only methods not using it are used elsewhere in always valid networks. I will check however!

miovd: Fixed by d5fbedb849f2598af1d43e68df06515b50a76c06

geofjamg:
> I propose to change the validation level names:
> 
> * Instead of `SCADA`, use `EQUIPMENT_CHARACTERISTICS` or `EQUIPMENT`. `SCADA` is one potential source of data. Another could be a Power System Model database (for example). What is being validated at this level is the definition of the equipment, or the equipment characteristics, independent of its original source.
> * Instead of `LOADFLOW`, use `STEADY_STATE_HYPOTHESIS`, `STEADY_STATE_INPUTS` or `LOADFLOW_INPUTS`. What is being validated is the information required to build a steady state of the network. They are hypothesis because we do not know if they are valid until we run a Loadflow. Calling it `LOADFLOW` could move to confusion: the user may think the Network contains a valid Loadflow solution.

I agree that SCADA is badly named but EQUIPMENT_CHARACTERISTICS is wrong too as there is still many validations not linked to equipment structure. The real meaning of this level of validation is something like "ready for state estimation".
LOADFLOW_INPUTS is fine for me so in that case we should have STATE_ESTIMATOR_INPUTS and LOADFLOW_INPUTS.

zamarrenolm:
> > I propose to change the validation level names:
> > 
> > * Instead of `SCADA`, use `EQUIPMENT_CHARACTERISTICS` or `EQUIPMENT`. `SCADA` is one potential source of data. Another could be a Power System Model database (for example). What is being validated at this level is the definition of the equipment, or the equipment characteristics, independent of its original source.
> > * Instead of `LOADFLOW`, use `STEADY_STATE_HYPOTHESIS`, `STEADY_STATE_INPUTS` or `LOADFLOW_INPUTS`. What is being validated is the information required to build a steady state of the network. They are hypothesis because we do not know if they are valid until we run a Loadflow. Calling it `LOADFLOW` could move to confusion: the user may think the Network contains a valid Loadflow solution.
> 
> I agree that SCADA is badly named but EQUIPMENT_CHARACTERISTICS is wrong too as there is still many validations not linked to equipment structure. The real meaning of this level of validation is something like "ready for state estimation". LOADFLOW_INPUTS is fine for me so in that case we should have STATE_ESTIMATOR_INPUTS and LOADFLOW_INPUTS.

Sorry, but strongly disagree.

What is being checked now at "SCADA" level are characteristics of equipment. It is being checked that there is a valid "structural" definition of the Network, without any reference to its possible states. What attributes do you consider that are being validated currently at "SCADA" level and that are not part of the equipment characteristics?

To have a "network" that is valid to attempt a State Estimation we would need further information, relative to the measured, forecasted or scheduled values of some magnitudes ... State Estimation tries to build a coherent set of electrical magnitudes from "approximate" values, so you also need that initial "approximate" values to be able to start the state estimation process. For me it would be wrong to name the most basic validation level based on STATE_ESTIMATION_...

Also, naming the validation levels based on the analysis functions that would be run on them is less than ideal. I would prefer using more abstract names that describe the kind of data that is present in the network, not the analysis function that will use it.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VscConverterStationImpl.java
```diff
@@ -39,7 +36,9 @@
         this.voltageRegulatorOn = new TBooleanArrayList(variantArraySize);
         this.reactivePowerSetpoint = new TDoubleArrayList(variantArraySize);
         this.voltageSetpoint = new TDoubleArrayList(variantArraySize);
-        this.voltageRegulatorOn.fill(0, variantArraySize, voltageRegulatorOn);
+        for (int i = 0; i < variantArraySize; i++) {
```
geofjamg, line 39: Could be reverted to initial impl?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractTransformerXml.java
```diff
@@ -84,25 +86,29 @@ protected static void writeRatioTapChanger(String name, RatioTapChanger rtc, Net
...
         context.getWriter().writeEndElement();
     }
 
     protected static void readRatioTapChanger(String elementName, RatioTapChangerAdder adder, Terminal terminal, NetworkXmlReaderContext context) throws XMLStreamException {
         int lowTapPosition = XmlUtil.readIntAttribute(context.getReader(), ATTR_LOW_TAP_POSITION);
-        int tapPosition = XmlUtil.readIntAttribute(context.getReader(), ATTR_TAP_POSITION);
-        boolean regulating = XmlUtil.readOptionalBoolAttribute(context.getReader(), ATTR_REGULATING, false);
-        double targetDeadband = readTargetDeadband(regulating, context);
+        Integer tapPosition = XmlUtil.readOptionalIntegerAttribute(context.getReader(), ATTR_TAP_POSITION);
+        String regulatingStr = context.getReader().getAttributeValue(null, ATTR_REGULATING);
```
geofjamg, line 98: Add a utility function for that?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractTransformerXml.java
```diff
@@ -164,18 +172,23 @@ protected static void writePhaseTapChanger(String name, PhaseTapChanger ptc, Net
 
     protected static void readPhaseTapChanger(String name, PhaseTapChangerAdder adder, Terminal terminal, NetworkXmlReaderContext context) throws XMLStreamException {
         int lowTapPosition = XmlUtil.readIntAttribute(context.getReader(), ATTR_LOW_TAP_POSITION);
-        int tapPosition = XmlUtil.readIntAttribute(context.getReader(), ATTR_TAP_POSITION);
-        boolean regulating = XmlUtil.readOptionalBoolAttribute(context.getReader(), ATTR_REGULATING, false);
-        double targetDeadband = readTargetDeadband(regulating, context);
-        PhaseTapChanger.RegulationMode regulationMode = PhaseTapChanger.RegulationMode.valueOf(context.getReader().getAttributeValue(null, "regulationMode"));
+        Integer tapPosition = XmlUtil.readOptionalIntegerAttribute(context.getReader(), ATTR_TAP_POSITION);
+        String regulationModeStr = context.getReader().getAttributeValue(null, "regulationMode");
```
geofjamg, line 176: Also it would be nice to have a XmlUtil.readOptionalEnumAttribute(reader, PhaseTapChanger.RegulationMode.class)

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/HvdcLineXml.java
```diff
@@ -37,7 +37,9 @@ protected boolean hasSubElements(HvdcLine identifiable) {
     protected void writeRootElementAttributes(HvdcLine l, Network parent, NetworkXmlWriterContext context) throws XMLStreamException {
         XmlUtil.writeDouble("r", l.getR(), context.getWriter());
         XmlUtil.writeDouble("nominalV", l.getNominalV(), context.getWriter());
-        context.getWriter().writeAttribute("convertersMode", l.getConvertersMode().name());
+        if (l.getConvertersMode() != null) {
```
geofjamg, line 40: XmlUtil.writeEnum()

geofjamg (changes requested):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChanger.java
```diff
@@ -183,8 +211,10 @@ public void extendVariantArraySize(int initVariantArraySize, int number, int sou
 
     @Override
     public void reduceVariantArraySize(int number) {
+        List<Integer> tmpInt = new ArrayList<>(tapPosition.subList(0, number));
```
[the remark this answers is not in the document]
miovd: If we do not use a temporary copy we try to reset `tapPosition` while taking `tapPosition` as reference, which does not work. Or I am misunderstanding you?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
```diff
@@ -182,6 +183,12 @@ public ShuntCompensatorAdder setSectionCount(int sectionCount) {
         return this;
     }
 
+    @Override
+    public ShuntCompensatorAdder unsetSectionCount() {
```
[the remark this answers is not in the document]
miovd: Indeed, I did that with voltageRegulator in mind but as sectionCount and tapPosition are mandatory for LOADFLOW adders, unset methods in adders are not necessary

geofjamg:
> Sorry, but strongly disagree.

After thinking again, I think you are right and your naming proposal is the best option.

annetill (approved):
For me, the PR is finished and we can continue the work. In case of errors, it would be easy to return to this without any consequence steady state hypothesis on users.

PR #1820, opened 2021-09-12
Add matrix transpose

geofjamg:

No

Feature

There is no way to transpose a matrix.

We can transpose a matrix either for the dense or sparse implementation.

**Other information**:


math/src/main/java/com/powsybl/math/matrix/DenseMatrix.java
```diff
@@ -260,6 +260,18 @@ protected int getEstimatedNonZeroValueCount() {
...
+    @Override
+    public DenseMatrix transpose() {
+        double[] transposed = new double[columnCount * rowCount];
+        for (int i = 0; i < rowCount; i++) {
+            for (int j = 0; j < columnCount; j++) {
+                double value = buffer.getDouble(j * Double.BYTES * rowCount + i * Double.BYTES);
+                transposed[i * columnCount + j] = value;
+            }
+        }
+        return new DenseMatrix(columnCount, rowCount, transposed);
```
flo-dup, line 272: ouch, not an implementation meant to be very effective, but we have here a `columnCount * rowCount` temporary array. If we use the `set(int, int, double)`, we avoid that, at the cost of the `checkBounds` for each though.

math/src/main/java/com/powsybl/math/matrix/SparseMatrix.java
```diff
@@ -385,6 +385,15 @@ protected int getEstimatedNonZeroValueCount() {
         return values.size();
     }
 
+    private native SparseMatrix transpose(int m, int n, int[] ap, int[] ai, double[] ax);
+
+    @Override
+    public SparseMatrix transpose() {
+        SparseMatrix transposed = transpose(rowCount, columnCount, columnStart, rowIndices.getData(), values.getData());
+        transposed.setRgrowthThreshold(rgrowthThreshold);
```
flo-dup, line 393: Just noticed that we're setting the same r growth threshold to all matrix, which questions me about it being a parameter of SparseMatrix.

math/src/main/java/com/powsybl/math/matrix/DenseMatrix.java
geofjamg: Right, fixed by directly filling a bytebuffer, we can avoid unnecessary bounds check.

flo-dup (approved):

math/src/main/java/com/powsybl/math/matrix/SparseMatrix.java
geofjamg: We just consider that a derivated matrix (transposed, multiplied) has by default the same parameter value than the initial one which seems to be fine to me. The rgrowthThreshold can be defined in the SparseMatrixFactory, so all matrices created from the same factory have by default the same value.

PR #1828, opened 2021-09-16
[WIP] Allow to configure Country for Voltage levels

MioRtia:

No

Feature

Voltage level can only be considered in a country if they have a substation container

Voltage level can be considered in a country in every case

No

**Other information**:


geofjamg (changes requested):
I think this dev would only add confusion to how define country.
Could you explain more why do you need this? What is the exact use case?
Why when you need to set a country, you just not create a substation (this this probably the main pupose of the substation!)?

annetill:
This is not more needed.

PR #1855, opened 2021-10-16
Add missing call to invalidate the cache of the synchronous components manager

mathbagu:

N/A

When a `NodeBreakerVoltageLevel` is invalidated due to a structure change, all the caches should be invalidated. But it seems the synchronous connected component is not invalidated whereas it is called in `BusBreakerVoltageLevel`.

Synchronous components are not updated when a NodeBreakerVoltageLevel is changed.

Synchronous components cache is now cleared.

**Other information**:


annetill:
> integration

Most of my tests were on IGMs where we don't have any topology change. But I wonder about the slow security analysis, the merging and the merging view now... @mathbagu any opinion?

PR #1865, opened 2021-10-22
Fix transformers substation check

geofjamg:

**Other information**:


MioRtia (approved):

MioRtia:
> @MioRtia we need to add unit test to highligtht the bug, but do you agree with this fix?

Yes I agree with this correction! Thank you, the behavior was a bit weird.

MioRtia:
@geofjamg I reflected a bit on this and I think it means that transformers with one branch in a substation and one branch in no substation are forbidden, is it what we wanted? Indeed, it is forbidden to create a transformer in a substation when one branch is not in any substation but with your fix, it is also forbidden to create a transformer from the network when one branch is in a substation.

On the other hand, the previous implementation did not look great.

Do you think it is preferable to put a transformer with one branch in a substation at this substation's level or at network's level? I don't know @annetill if you have an opinion as a user. In any case, this will need to be documented.

PR #1874, opened 2021-10-29
Cgmes fix TieFlow mapping

marqueslanauja:

Feature

`TieFlows` with terminal associated to a `switch` or an `EquivalentInjection` are lost.

For `TieFlows` with terminal associated to a `switch`  the terminal will be the network side of the `DanglingLine` in the same `voltageLevel` connected to the original `switch` through a chain of switches.

For `TieFlows` with terminal associated to an `EquivalentInjection` the terminal will be the boundary side of the corresponding `DanglingLine`.

**Other information**:


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/SwitchesChain.java
```diff
@@ -0,0 +1,262 @@
...
+    private final NodeConnectables nodeConnectables;
+    private final Adjacency adjacency;
+    private final VoltageLevel vl;
+    private final Switch sw;
+    private final boolean isNodeBreakerModel;
+
+    SwitchesChain(VoltageLevel vl, Switch sw, boolean isNodeBreakerModel) {
+        this.vl = vl;
+        this.sw = sw;
+        this.isNodeBreakerModel = isNodeBreakerModel;
```
flo-dup, line 33: Trying to have the same code for the two kinds of view is a bit artificial, I'm not sure the benefit is real here.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/SwitchesChain.java
```diff
@@ -0,0 +1,262 @@
...
+        nodeConnectables = new NodeConnectables(vl, isNodeBreakerModel);
+        adjacency = new Adjacency(vl, sw.getId(), isNodeBreakerModel);
```
flo-dup, line 35: With the new classes `NodeConnectables` and `Adjacency` you're just reconstructing the graph don't you? I don't see which specific need forced you to do that instead of using the existing topology traversers, maybe I missed something. If the existing topology traversers miss something you need, we could also try to improve them, rather than adding this kind of code here.

flo-dup (changes requested):
As said in the comment, I find it a bit confusing not to use the existing topology traversers methods. Maybe they're not enough for your need here though - did you already give it a try?

And just a minor comment about the two huge EQ files, it might not be worth the effort, but we could (later?) remove them as they're almost exactly the same as the small grid EQ file. except for the one line of the corresponding TieFlow.Terminal.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/SwitchesChain.java
marqueslanauja: I refactored the code. I think now is more readable.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/SwitchesChain.java
marqueslanauja: I used the existing traverse for `nodeBreaker `and I implemented a new one for `busBreaker`. I added this new traverse method in the `BusBreakerView `interface of the `VoltageLevel `class. Should I label the PR as Breaking Change?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/SwitchesChain.java
```diff
@@ -31,232 +29,173 @@
...
-        private static void adjacencyBusBreaker(VoltageLevel vl, String switchId, Map<String, List<String>> adjacency) {
-            vl.getSwitches().forEach(sw -> {
-                if (sw.getId().equals(switchId)) {
+            c.getTerminals().forEach(terminal -> {
+                if (!isSameVoltageLevel((Terminal) terminal, vl)
+                    || !isTerminalInConfiguredBuses((Terminal) terminal, configuredBuses)) {
                     return;
                 }
-                addAdjacency(adjacency, vl.getBusBreakerView().getBus1(sw.getId()), vl.getBusBreakerView().getBus2(sw.getId()));
+                terminals.add((Terminal) terminal);
```
flo-dup, line 162: Rather than returning if (!a || !b), add the terminal only if (a && b). That could be done in a lambda then.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/SwitchesChain.java
```diff
@@ -31,232 +29,173 @@
...
+        return configuredBuses.contains(terminal.getBusBreakerView().getBus().getId());
+    }
+
+    private static boolean isSameVoltageLevel(Terminal terminal, VoltageLevel vl) {
+        return terminal.getVoltageLevel().equals(vl);
+    }
+
+    private static boolean isDiscarded(Connectable<?> connectable) {
+        return connectable.getType() == IdentifiableType.BUSBAR_SECTION;
+    }
```
flo-dup, line 186: I'd put that method directly in the code, it will be clearer I think

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/SwitchesChain.java
```diff
@@ -31,232 +29,173 @@
...
+
+    // The best terminal is the terminal associated to the line, branch at the border
```
flo-dup, line 188: Either put that comment inside the method or use the javadoc syntax
```
/**
 * xxx
 */
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/SwitchesChain.java
```diff
@@ -31,232 +29,173 @@
...
-                || connectable.getType() == IdentifiableType.DANGLING_LINE;
-        }
+        voltageLevel.getBusBreakerView().traverse(configuredBusId, traverser);
+        return configuredBuses;
     }
 
-    private static class Adjacency {
-
-        private final Map<String, List<String>> adjacencies;
+    private static Terminal uniqueTerminalAssociatedConductingEquipmentNodeBreaker(VoltageLevel vl, List<Integer> nodes) {
```
flo-dup, line 126: to avoid duplicating that method, you could use only one:
`private static <T> Terminal uniqueTerminalAssociatedConductingEquipment(VoltageLevel vl, List<T> vertices, Function<Terminal, T> terminalToVertex)`, where terminalToVertex will be 
- either `t -> t.getBusBreakerView().getBus()` 
- or `t -> t.getNodeBreakerView().getNode()`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/SwitchesChain.java
```diff
@@ -31,232 +29,173 @@
...
-                        }
-                        connectables.computeIfAbsent(((Terminal) t).getBusBreakerView().getBus().getId(), k -> new ArrayList<>()).add(c);
-                    });
-            });
-        }
+        VoltageLevel.NodeBreakerView.TopologyTraverser traverser = (node1, sw, node2) -> {
+            if (sw == swTerminal) {
+                return TraverseResult.TERMINATE_PATH;
+            }
+            nodes.add(node2);
```
flo-dup, line 102: In both expand methods, what if a switch is opened? shouldn't the switch chain be with only closed switches?

flo-dup (changes requested):
Great job, it's a lot clearer that way!

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/VoltageLevel.java
```diff
@@ -791,6 +791,19 @@ default boolean hasAttachedEquipment(int node) {
...
+        interface TopologyTraverser {
+            /**
+             * Called for each traversal step
+             * @param node1 the node the traversal comes from
+             * @param sw the {@link Switch} encountered, or null if it is an {@link InternalConnection}
+             * @param node2 the node the traversal will go to, if the returned TraverseResult is {@link TraverseResult#CONTINUE}
+             * @return {@link TraverseResult#CONTINUE} to continue traversal, {@link TraverseResult#TERMINATE_PATH}
+             * to stop the current traversal path, {@link TraverseResult#TERMINATE_TRAVERSER} to stop all the traversal paths
+             */
+            TraverseResult traverse(String busId1, Switch sw, String busId2);
```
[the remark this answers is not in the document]
flo-dup: After discussing with @geofjamg, I confirm, let's change this to `traverse(Bus bus1, Switch sw, Bus bus2)`

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/VoltageLevel.java
```diff
@@ -791,6 +791,19 @@ default boolean hasAttachedEquipment(int node) {
...
 
+        interface TopologyTraverser {
+            /**
+             * Called for each traversal step
+             * @param node1 the node the traversal comes from
+             * @param sw the {@link Switch} encountered, or null if it is an {@link InternalConnection}
+             * @param node2 the node the traversal will go to, if the returned TraverseResult is {@link TraverseResult#CONTINUE}
+             * @return {@link TraverseResult#CONTINUE} to continue traversal, {@link TraverseResult#TERMINATE_PATH}
+             * to stop the current traversal path, {@link TraverseResult#TERMINATE_TRAVERSER} to stop all the traversal paths
+             */
```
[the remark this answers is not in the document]
marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/SwitchesChain.java
```diff
@@ -31,232 +29,173 @@
...
+        if (terminals.size() == 1) {
+            return terminals.get(0);
         }
+        return null;
+    }
+
+    private static Terminal uniqueTerminalAssociatedConductingEquipmentBusBranch(VoltageLevel vl, List<String> configuredBuses) {
+        List<Terminal> terminals = new ArrayList<>();
+
+        vl.getConnectables().forEach(c -> {
```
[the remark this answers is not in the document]
marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/SwitchesChain.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/SwitchesChain.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/SwitchesChain.java
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/VoltageLevel.java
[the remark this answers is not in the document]
marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/SwitchesChain.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/SwitchesChain.java
```diff
@@ -31,232 +29,173 @@
...
+            if (isDiscarded(c)) {
+                return;
+            }
 
-        private static void adjacencyBusBreaker(VoltageLevel vl, String switchId, Map<String, List<String>> adjacency) {
-            vl.getSwitches().forEach(sw -> {
-                if (sw.getId().equals(switchId)) {
+            c.getTerminals().forEach(terminal -> {
```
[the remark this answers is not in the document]
marqueslanauja: I did, but slightly different. Here is the code I implemented:

```
            Connectable<?> c1 = c;
            c1.getTerminals().forEach(terminal -> {
                if (isSameVoltageLevel(terminal, vl) && vertices.contains(terminalToVertex.apply(terminal))) {
                    terminals.add(terminal);
                }
            });
```

If I do what you propose sonarlint generates the following code smell `Remove this unnecessary cast to "Connectable"`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/SwitchesChain.java
marqueslanauja: We have to consider both types of switches (opened and closed). The new terminal associated to the TieFlow should always have the same flow as the initial terminal defined in the switch, independently of the status of the switches in the chain. But this assertion is guaranteed in a chain, the active power flow will be the same at any node  independently of the status of the switches.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/SwitchesChain.java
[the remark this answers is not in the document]
flo-dup: Sonarlint is a bit confused here, so we could have ignored the code smell, but ok.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/SwitchesChain.java
flo-dup: Thanks for explaining! I saw you also added a comment clarifying that :+1:

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/SwitchesChain.java
flo-dup: I simplified what you changed to have a simpler lambda like I suggested

flo-dup (approved):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/SwitchesChain.java
marqueslanauja: ok

PR #1879, opened 2021-11-02
Use scientific notation for very small values in bPerSection

marcosmc:

Fix bug

When exporting to CGMES, shunts may appear with bPerSection equal to 0, even though its original value is not 0 but if a very small value, this error occurs when formatting the shunt values ​​in the export.

The option to export small values ​​with scientific format for the bPerSection attribute of the shunts has been added.

**Other information**:


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/elements/ShuntCompensatorEq.java
```diff
@@ -48,7 +48,11 @@ public static void write(String id, String shuntCompensatorName, int normalSecti
         writer.writeEndElement();
         if (modelType.equals(ShuntCompensatorModelType.LINEAR)) {
             writer.writeStartElement(cimNamespace, EQ_LINEARSHUNTCOMPENSATOR_BPERSECTION);
-            writer.writeCharacters(CgmesExportUtil.format(bPerSection));
+            if (Math.abs(bPerSection) < 1E-10) {
+                writer.writeCharacters(CgmesExportUtil.scientificFormat(bPerSection));
+            } else {
+                writer.writeCharacters(CgmesExportUtil.format(bPerSection));
+            }
```
MioRtia, line 55: Just a question, why do you not always use  scientific format? Is there an added value not to always use it? In any case: if you want to discriminate, please put `1E-10` as a private final static field at the start of the file in order to quickly identify hard-coded threshold values.

MioRtia (commented):
Just a question about the cases discrimination. Otherwise, it is good for me.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/elements/ShuntCompensatorEq.java
marcosmc: It is preferable not to use the scientific notation, unless it is necessary because precision is lost when writing the value in the default format, to facilitate the reading by the user.
A static variable (BPERSECTION_FORMAT_THRESHOLD) is declared to determine the threshold value used to discriminate which format to use in the export

miovd (approved):

PR #1894, opened 2021-11-16
CGMES import: Fix nominalVoltage if equal to zero

miovd:

No

Fix during CGMES import


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/VoltageLevelConversion.java
```diff
@@ -50,6 +50,10 @@ public void convert() {
             missing(bv);
             throw new CgmesModelException(String.format("nominalVoltage not found for %s", bv));
         }
+        if (nominalVoltage == 0) {
```
annetill, line 53: I think that in a first approach, we have to drop these voltage levels. It will lead to errors during calculations.

PR #1896, opened 2021-11-16
CGMES. Optimize verification of model is node/breaker

zamarrenolm:

A model is considered node/breaker based on the profiles defined in its files. Recompute if the model is node/breaker only when new files are added to the triplestore.

The model profiles are queried and reviewed everytime the model is checked for node/breaker.

The flag that stores if a model is node/breaker is cached and only recomputed when new data is added to the triplestore.

### DIFF
--- a/cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/triplestore/CgmesModelTripleStore.java
+++ b/cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/triplestore/CgmesModelTripleStore.java
@@ -45,7 +45,7 @@ public class CgmesModelTripleStore extends AbstractCgmesModel {
     @Override
     public void read(InputStream is, String baseName, String contextName) {
         // Reset cached isNodeBreaker everytime we read new data
-        isNodeBreakerComputed = false;
+        nodeBreakerComputed = false;
         tripleStore.read(is, baseName, contextName);
     }
 
@@ -147,11 +147,11 @@ public class CgmesModelTripleStore extends AbstractCgmesModel {
 
     @Override
     public boolean isNodeBreaker() {
-        if (!isNodeBreakerComputed) {
-            isNodeBreaker = computeIsNodeBreaker();
-            isNodeBreakerComputed = true;
+        if (!nodeBreakerComputed) {
+            nodeBreaker = computeIsNodeBreaker();
+            nodeBreakerComputed = true;
         }
-        return isNodeBreaker;
+        return nodeBreaker;
     }
 
     private boolean computeIsNodeBreaker() {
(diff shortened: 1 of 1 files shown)


cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/triplestore/CgmesModelTripleStore.java
```diff
@@ -688,6 +698,8 @@ private String injectParams(String queryText, String... params) {
     private final int cimVersion;
     private final TripleStore tripleStore;
     private final QueryCatalog queryCatalog;
+    private boolean isNodeBreaker;
```
geofjamg, line 701: To remove the Sonar warning (naming collision between attribute and method), we should rename attribute to nodeBreaker and nodeBreakerComputed?

zamarrenolm: Agree, renamed the attributes. Also reduced cognitive complexity of nodeBreaker computation by splitting in two separate methods.

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/triplestore/CgmesModelTripleStore.java
```diff
@@ -688,6 +703,8 @@ private String injectParams(String queryText, String... params) {
     private final int cimVersion;
     private final TripleStore tripleStore;
     private final QueryCatalog queryCatalog;
+    private boolean nodeBreaker;
+    private boolean nodeBreakerComputed = false;
```
miovd, line 707: Rather than adding an intermediary boolean, you can use `Boolean` and compute it when it is null

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/triplestore/CgmesModelTripleStore.java
```diff
@@ -157,36 +167,41 @@ public boolean isNodeBreaker() {
...
+                    "All FullModel objects have EquipmentOperation profile, so conversion will be considered node-breaker");
+        } else {
+            LOG.info(
+                    "Following FullModel objects do not have EquipmentOperation profile, so conversion will not be considered node-breaker:");
+            modelHasOperationProfile.entrySet().forEach(meqop -> {
+                if (!meqop.getValue()) {
+                    LOG.info("    {}", meqop.getKey());
+                }
+            });
+        }
```
miovd, line 183: Maybe you should check if `LOG.isInfoEnabled()` before entering this code, otherwise you might go through it for no reason...

miovd (changes requested):

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/triplestore/CgmesModelTripleStore.java
zamarrenolm: Done, and moved the log to a separate method.

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/triplestore/CgmesModelTripleStore.java
zamarrenolm: Done

miovd (approved):

PR #1898, opened 2021-11-18
Fix base voltage config loading

geofjamg:

No

Regression

We don't have anymore the jar resource fallback when config is not overloaded by user in its $HOME/.itools.

**Other information**:

### DIFF
--- a/commons/src/main/java/com/powsybl/commons/config/BaseVoltagesConfig.java
+++ b/commons/src/main/java/com/powsybl/commons/config/BaseVoltagesConfig.java
@@ -10,7 +10,10 @@ import com.powsybl.commons.PowsyblException;
 import org.yaml.snakeyaml.Yaml;
 import org.yaml.snakeyaml.constructor.Constructor;
 import org.yaml.snakeyaml.error.YAMLException;
-import org.yaml.snakeyaml.nodes.*;
+import org.yaml.snakeyaml.nodes.MappingNode;
+import org.yaml.snakeyaml.nodes.Node;
+import org.yaml.snakeyaml.nodes.NodeTuple;
+import org.yaml.snakeyaml.nodes.ScalarNode;
 
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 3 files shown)


commons/src/main/java/com/powsybl/commons/config/BaseVoltagesConfig.java
```diff
@@ -66,12 +69,23 @@ public static BaseVoltagesConfig fromInputStream(InputStream configInputStream)
...
+        Objects.requireNonNull(configFileName);
+        Path configFile = configDir.resolve(configFileName);
+        if (Files.exists(configFile)) {
+            try (InputStream configInputStream = Files.newInputStream(configFile)) {
+                return fromInputStream(configInputStream);
+            } catch (IOException e) {
+                throw new UncheckedIOException(e);
+            }
+        } else {
+            InputStream configInputStream = BaseVoltagesConfig.class.getResourceAsStream("/" + configFileName);
```
flo-dup, line 83: Could you add a log to tell the user the default config file has been loaded?

commons/src/main/java/com/powsybl/commons/config/BaseVoltagesConfig.java
```diff
@@ -66,12 +69,23 @@ public static BaseVoltagesConfig fromInputStream(InputStream configInputStream)
...
+            if (configInputStream != null) {
+                return fromInputStream(configInputStream);
+            } else {
+                throw new PowsyblException("No base voltages configuration found");
```
flo-dup, line 87: Maybe add default: `"No default base voltages configurations found"`

flo-dup (approved):

commons/src/main/java/com/powsybl/commons/config/BaseVoltagesConfig.java
geofjamg: Done

commons/src/main/java/com/powsybl/commons/config/BaseVoltagesConfig.java
geofjamg: Done

flo-dup (approved):

PR #1900, opened 2021-11-18
[WIP] During CGMES import, correct nominal voltage if possible while creating fictitious voltage levels for line containers

miovd:
Signed-off-by: VEDELAGO MIORA <miora.vede@gmail.com>

No

Bug fix

Depending on the order of creation of line containers, nominal voltage can be badly set.

From the configuration of line containers we have, this solution allows to only have correct nominal voltages.

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
@@ -385,9 +385,7 @@ public class Conversion {
     }
 
     private void convertACLineSegmentsToLines(Context context, Set<String> delayedBoundaryNodes) {
-        Iterator<PropertyBag> k = cgmes.acLineSegments().iterator();
-        while (k.hasNext()) {
-            PropertyBag line = k.next();
+        for (PropertyBag line : cgmes.acLineSegments()) {
             if (LOG.isDebugEnabled()) {
                 LOG.debug(line.tabulateLocals("ACLineSegment"));
             }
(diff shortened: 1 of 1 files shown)


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -460,8 +470,8 @@ private VoltageLevel newFictitiousVoltageLevel(Context context, LineContainerFic
                         context.nodeBreaker()
                                 ? TopologyKind.NODE_BREAKER
                                 : TopologyKind.BUS_BREAKER)
-                .setLowVoltageLimit(vlref != null ? vlref.getLowVoltageLimit() : 1.0)
```
annetill, line 463: I am not sure that it is correct to set the nominal voltage to 1 here too... how do you retrieve the reference voltage level for a client connection in a middle of a line?

miovd: We can't, this is why it is done a posteriori. I checked for every line configuration we have, they end up being set even if not done immediately.

marcosmc (commented):
We have reviewed the changes and they are correct, but it is somewhat confusing to understand their purpose. 
We would like to analyze and make a proposal that takes into account all potential configurations (where fictitious voltage levels do not have adjacent real voltage levels) that may arise without excessively penalizing the conversion performance.

miovd:
> We have reviewed the changes and they are correct, but it is somewhat confusing to understand their purpose. We would like to analyze and make a proposal that takes into account all potential configurations (where fictitious voltage levels do not have adjacent real voltage levels) that may arise without excessively penalizing the conversion performance.

The issue was when a AC line between two fictitious voltages was at the beginning of the list... However, we might change our proposition as well. I look forwards to your insights!

marcosmc:
> From the configuration of line containers we have, this solution allows to only have correct nominal voltages.

In draft # 1905 we have developed our proposal to solve the bug in the fictitious voltage levels, our proposal consists of locating for each line container a Voltage level that serves as a reference to create all the necessary fictitious ones. We have done a performance analysis and this is affected by a slight increase in conversion, about an extra 1%

miovd:
Replaced by #1905

PR #1901, opened 2021-11-18
CGMES. Add functional logs

### LINKED ISSUE #1887: Functional logs for CIM-CGMES converter

A feature.

We log with SLF4J. And there is a lot of logs that were useful before, during the development of the gateway. But now, it is finished and some logs should be cleaned.

As for UCTE importer, we should add functional logs (reporters) to have a mechanism of logging compatible with micro-services architecture. Furthermore, it could be nice to clean globally the logs.

  - PowSyBl Version: ...
  - OS Version: ...

zamarrenolm:

Closes #1887.


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
```diff
@@ -113,16 +114,21 @@ public String getFormat() {
...
+        Objects.requireNonNull(networkFactory);
+        Objects.requireNonNull(reporter);
         TripleStoreOptions options = new TripleStoreOptions();
         String sourceForIidmIds = Parameter.readString(getFormat(), p, SOURCE_FOR_IIDM_ID_PARAMETER, defaultValueConfig);
         if (sourceForIidmIds.equalsIgnoreCase(SOURCE_FOR_IIDM_ID_MRID)) {
             options.setRemoveInitialUnderscoreForIdentifiers(true);
         } else if (sourceForIidmIds.equalsIgnoreCase(SOURCE_FOR_IIDM_ID_RDFID)) {
             options.setRemoveInitialUnderscoreForIdentifiers(false);
         }
+        options.setReporter(reporter.createSubReporter("Read", "Read"));
```
miovd, line 128: Maybe a more comprehensive key? `CGMESTriplestore`?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
```diff
@@ -113,16 +114,21 @@ public String getFormat() {
...
         CgmesModel cgmes = CgmesModelFactory.create(ds, boundary(p), tripleStore(p), options);
-        return new Conversion(cgmes, config(ds, p), activatedPostProcessors(p), networkFactory).convert();
+        Reporter conversionReporter = reporter.createSubReporter("Conversion", "Conversion");
```
miovd, line 130: `CGMESConversion` as key?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Context.java
```diff
@@ -7,39 +7,41 @@
...
  */
 public class Context {
 
-    // Log messages
-    private static final String FIXED_REASON = "Fixed {}. Reason: {}";
-    private static final String INVALID_REASON = "Invalid {}. Reason: {}";
-    private static final String IGNORED_REASON = "Ignored {}. Reason: {}";
-
     public Context(CgmesModel cgmes, Config config, Network network) {
+        this(cgmes, config, network, null);
```
miovd, line 37: Not null, pass `Reporter.NO_OP`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Context.java
```diff
@@ -203,69 +205,132 @@ public PropertyBags phaseTapChangerTable(String tableId) {
...
+
+    private void handleIssue(ConversionIssueCategory category, String what, String reason) {
+        if (reporter != null) {
+            reportIssue(category, what, reason);
+        }
+        logIssue(category, what, reason);
+    }
+
+    private void handleIssue(ConversionIssueCategory category, String what, Supplier<String> reason) {
+        if (LOG.isWarnEnabled() || reporter != null) {
```
miovd, line 280: Report in any case. In default case, reporter should be `Reporter.NO_OP` meaning no operation is done

miovd, line 280: Report in any case. In default case, reporter should be `Reporter.NO_OP` meaning no operation is done, reporter should **not** be null

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Context.java
```diff
@@ -7,39 +7,41 @@
...
-
     public Context(CgmesModel cgmes, Config config, Network network) {
+        this(cgmes, config, network, null);
+    }
+
+    public Context(CgmesModel cgmes, Config config, Network network, Reporter reporter) {
         this.cgmes = Objects.requireNonNull(cgmes);
         this.config = Objects.requireNonNull(config);
         this.network = Objects.requireNonNull(network);
+        this.reporter = reporter;
```
miovd, line 44: Add `requireNonNull`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -116,12 +117,16 @@ public Conversion(CgmesModel cgmes, Conversion.Config config, List<CgmesImportPo
...
+        this(cgmes, config, postProcessors, networkFactory, null);
+    }
+
+    public Conversion(CgmesModel cgmes, Config config, List<CgmesImportPostProcessor> activatedPostProcessors, NetworkFactory networkFactory, Reporter reporter) {
         this.cgmes = Objects.requireNonNull(cgmes);
         this.config = Objects.requireNonNull(config);
-        this.postProcessors = Objects.requireNonNull(postProcessors);
+        this.postProcessors = Objects.requireNonNull(activatedPostProcessors);
         this.networkFactory = Objects.requireNonNull(networkFactory);
+        this.reporter = reporter;
```
miovd, line 129: Add `requireNonNull`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -116,12 +117,16 @@ public Conversion(CgmesModel cgmes, Conversion.Config config, List<CgmesImportPo
         this(cgmes, config, postProcessors, NetworkFactory.findDefault());
     }
 
-    public Conversion(CgmesModel cgmes, Conversion.Config config, List<CgmesImportPostProcessor> postProcessors,
-                      NetworkFactory networkFactory) {
+    public Conversion(CgmesModel cgmes, Conversion.Config config, List<CgmesImportPostProcessor> postProcessors, NetworkFactory networkFactory) {
+        this(cgmes, config, postProcessors, networkFactory, null);
```
miovd, line 121: By default, use `Reporter.NO_OP`

miovd (changes requested):
Some remarks.
Please consider that Reporter is never null. By default, it is `NO_OP`, not null.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Context.java
zamarrenolm: I've kept the check for the reporter being different of `Reporter.NO_OP`. The check is done to avoid calling the `reason` supplier if the resulting message will not be used (an _optimization_ introduced some time ago).

zamarrenolm: changed to check `reporter` against `Reporter.NO_OP`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
zamarrenolm: changed the key

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Context.java
zamarrenolm: avoided null reporter

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Context.java
zamarrenolm: added check

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
zamarrenolm: added

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
zamarrenolm: used no-operation reporter instead of null

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Context.java
```diff
@@ -203,69 +205,126 @@ public PropertyBags phaseTapChangerTable(String tableId) {
...
+        }
+    }
+
+    private static boolean isValidForGrouping(String what) {
+        return what.equals("SvVoltage") || what.equals("Regulating Terminal") || what.equals("EmptyVscRegulation");
+    }
+
+    private void reportIssue(ConversionIssueCategory category, String what, Supplier<String> reason) {
+        Reporter categoryReporter = issueReporters.computeIfAbsent(category, cat -> reporter.createSubReporter(cat.name(), cat.toString()));
+        // FIXME(Luma) A first approach to functional logs, report issues grouped by "what" or by "reason"
```
flo-dup, line 293: I'm not completely convinced we should do this grouping, or at least that way. This could be done by the functional logs viewer, but the strength of what you proposed is if there are many similar logs. Then you might want to show to the user a group of similar issues instead of the details of all issues, and the details can be looked into if he wants by unfolding this group.

Nonetheless, the title of that subReporter should then sum up the content, for instance the number of issues of that type, which is not the case here. There's some refactoring needed to do that.

flo-dup (changes requested):
There's been a misunderstanding here, the reporter goal is missed out. The functional logs should contain typed values, to format them according to user standards (for instance we could think of precision of displayed value depending on its type, color displayed, link to the voltage level diagram if vl id, ...). It's also meant to be used for multilingual support, and should thus be easily translated based on the user language.

Here so far the logs are preformatted using `String.format("xxx %d %f %s", x, y, z)` and the keys are equal to the log string: `reporter.report(item, item)`. Hence the front end cannot do anything more with the log than printing them, and multilingual support is impossible.

A good way to expose those problems is to export the test case in json and look at the dictionary.

miovd (changes requested):

zamarrenolm:
> There's been a misunderstanding here, the reporter goal is missed out. The functional logs should contain typed values, to format them according to user standards (for instance we could think of precision of displayed value depending on its type, color displayed, link to the voltage level diagram if vl id, ...). It's also meant to be used for multilingual support, and should thus be easily translated based on the user language.
> 
> Here so far the logs are preformatted using `String.format("xxx %d %f %s", x, y, z)` and the keys are equal to the log string: `reporter.report(item, item)`. Hence the front end cannot do anything more with the log than printing them, and multilingual support is impossible.
> 
> A good way to expose those problems is to export the test case in json and look at the dictionary.

Your are completely right in your observations. This PR was a first effort to adapt the information about issues found during the conversion to the reporter API without making almost any change to the current code.

miovd:
> > There's been a misunderstanding here, the reporter goal is missed out. The functional logs should contain typed values, to format them according to user standards (for instance we could think of precision of displayed value depending on its type, color displayed, link to the voltage level diagram if vl id, ...). It's also meant to be used for multilingual support, and should thus be easily translated based on the user language.
> > Here so far the logs are preformatted using `String.format("xxx %d %f %s", x, y, z)` and the keys are equal to the log string: `reporter.report(item, item)`. Hence the front end cannot do anything more with the log than printing them, and multilingual support is impossible.
> > A good way to expose those problems is to export the test case in json and look at the dictionary.
> 
> Your are completely right in your observations. This PR was a first effort to adapt the information about issues found during the conversion to the reporter API without making almost any change to the current code.

I would prefer if you start with few logs in the reporter but correctly built to be usable by an external application! Maybe you can start with the `invalid` messages for example.

zamarrenolm:
> > > There's been a misunderstanding here, the reporter goal is missed out. The functional logs should contain typed values, to format them according to user standards (for instance we could think of precision of displayed value depending on its type, color displayed, link to the voltage level diagram if vl id, ...). It's also meant to be used for multilingual support, and should thus be easily translated based on the user language.
> > > Here so far the logs are preformatted using `String.format("xxx %d %f %s", x, y, z)` and the keys are equal to the log string: `reporter.report(item, item)`. Hence the front end cannot do anything more with the log than printing them, and multilingual support is impossible.
> > > A good way to expose those problems is to export the test case in json and look at the dictionary.
> > 
> > 
> > Your are completely right in your observations. This PR was a first effort to adapt the information about issues found during the conversion to the reporter API without making almost any change to the current code.
> 
> I would prefer if you start with few logs in the reporter but correctly built to be usable by an external application! Maybe you can start with the `invalid` messages for example.

Sure, I will refactor the change to follow that.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
```diff
@@ -113,16 +114,21 @@ public String getFormat() {
...
+        Objects.requireNonNull(networkFactory);
+        Objects.requireNonNull(reporter);
         TripleStoreOptions options = new TripleStoreOptions();
         String sourceForIidmIds = Parameter.readString(getFormat(), p, SOURCE_FOR_IIDM_ID_PARAMETER, defaultValueConfig);
         if (sourceForIidmIds.equalsIgnoreCase(SOURCE_FOR_IIDM_ID_MRID)) {
             options.setRemoveInitialUnderscoreForIdentifiers(true);
         } else if (sourceForIidmIds.equalsIgnoreCase(SOURCE_FOR_IIDM_ID_RDFID)) {
             options.setRemoveInitialUnderscoreForIdentifiers(false);
         }
+        options.setReporter(reporter.createSubReporter("CGMESTriplestore", "Read"));
```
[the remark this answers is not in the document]
zamarrenolm: `Reporter` removed from `TripleStoreOptions`.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
```diff
@@ -113,16 +114,21 @@ public String getFormat() {
...
         CgmesModel cgmes = CgmesModelFactory.create(ds, boundary(p), tripleStore(p), options);
-        return new Conversion(cgmes, config(ds, p), activatedPostProcessors(p), networkFactory).convert();
+        Reporter conversionReporter = reporter.createSubReporter("CGMESConversion", "Conversion");
+        return new Conversion(cgmes, config(ds, p), activatedPostProcessors(p), networkFactory, conversionReporter).convert();
```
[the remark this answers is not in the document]
zamarrenolm: Reporter has been moved to a parameter of `Conversion::convert`.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/FunctionalLogsTest.java
```diff
@@ -0,0 +1,102 @@
...
+public class FunctionalLogsTest {
+
+    @Test
+    public void testImport() {
+        TestGridModel testCase = CgmesConformity1Catalog.microGridBaseCaseBE();
+        ReporterModel reporter = new ReporterModel("testFunctionalLogs",
+                "Test importing ${name}", Map.of("name", new TypedValue(testCase.name(), TypedValue.UNTYPED)));
+        Importers.importData("CGMES", testCase.dataSource(), null, reporter);
+
+        String expected = "+ Test importing MicroGrid-BaseCase-BE" + System.lineSeparator() +
```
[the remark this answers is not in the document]
zamarrenolm: Expected result has been moved to a resource file.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Context.java
```diff
@@ -203,69 +205,126 @@ public PropertyBags phaseTapChangerTable(String tableId) {
...
+        // FIXME(Luma) simple compound key
+        String categoryGroup = category.toString();
+        String description = "Identified issues not yet organized";
+        if (!group.isEmpty() && !group.isBlank()) {
+            categoryGroup += "_" + group;
+            description = group;
+        }
+        final String description1 = description;
+        reportersByCategoryAndGroup.computeIfAbsent(categoryGroup, cg -> categoryReporter.createSubReporter(cg, description1))
+            .report(item, item);
```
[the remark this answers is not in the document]
zamarrenolm: I've made a first implementation where I keep the previous `invalid` calls, adding a detailed report built at the source code point where the issue is first detected. 
The previous calls to `fixed, invalid, missing, ignored` simply emit a log message.
The idea will be to make reports only for relevant issues, following a similar approach to what is done in UCTE importer (minor issues are logged, relevant things are reported).

miovd (approved):

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/AbstractCgmesModel.java
```diff
@@ -23,8 +26,10 @@
  */
 public abstract class AbstractCgmesModel implements CgmesModel {
 
-    protected AbstractCgmesModel() {
+    protected AbstractCgmesModel(Reporter reporter) {
+        // FIXME(Luma) we must remove properties from here. They are not used!
         this.properties = new Properties();
+        this.reporter = reporter;
```
flo-dup, line 32: I'd rather not retain the reporter in an object, especially if there's only one call (or few calls) on it. You'd rather put the reporter as an argument of CgmesModel::read, I find it's easier to see what's going on with it that way... at the cost of polluting a bit the API.

A good compromise is what you did in another part of that PR, save it in an existing context object which is passed along. Unfortunately here there's no such context object.

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/AbstractCgmesModel.java
```diff
@@ -252,6 +257,14 @@ public void read(ReadOnlyDataSource ds) {
         CgmesOnDataSource cds = new CgmesOnDataSource(ds);
         for (String name : cds.names()) {
             LOG.info("Reading [{}]", name);
+            if (reporter != null) {
```
flo-dup, line 260: you can suppose the reporter is never null

flo-dup (changes requested):

flo-dup (approved):

PR #1902, opened 2021-11-19
Ignore tables with missing steps during CGMES import: tap changers considered linears

miovd:

No

Feature

Does not take into account step number in tables: can fail depending on tap position + can import faulty tables

Ignore faulty tables

No

**Other information**:

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/CgmesPhaseTapChangerBuilder.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/CgmesPhaseTapChangerBuilder.java
@@ -13,6 +13,7 @@ import java.util.function.Supplier;
 import com.powsybl.cgmes.conversion.Context;
 import com.powsybl.cgmes.conversion.RegulatingControlMappingForTransformers;
 import com.powsybl.cgmes.model.CgmesNames;
+import com.powsybl.commons.PowsyblException;
 import com.powsybl.triplestore.api.PropertyBag;
 import com.powsybl.triplestore.api.PropertyBags;
 
@@ -62,7 +63,21 @@ public class CgmesPhaseTapChangerBuilder extends AbstractCgmesTapChangerBuilder 
         if (isLinear()) {
             addStepsLinear();
         } else if (isTabular()) {
-            addStepsFromTable();
+            PropertyBags table = context.phaseTapChangerTable(tableId);
+            if (table == null) {
+                addStepsLinear();
+                return;
+            }
+            int min = table.stream().map(p -> p.asInt(CgmesNames.STEP)).min(Integer::compareTo).orElseThrow(() -> new PowsyblException("Should at least contain one step"));
+            for (int i = min; i < min + table.size(); i++) {
+                int index = i;
+                if (table.stream().noneMatch(p -> p.asInt(CgmesNames.STEP) == index)) {
+                    context.ignored("PhaseTapChanger table", () -> String.format("There is at least one missing step (%s) in table %s. Phase tap changer considered linear", index, tableId));
+                    addStepsLinear();
+                    return;
+                }
+            }
+            addStepsFromTable(table);
         } else if (isAsymmetrical()) {
             addStepsAsymmetrical();
         } else if (isSymmetrical()) {
@@ -91,11 +106,7 @@ public class CgmesPhaseTapChangerBuilder extends AbstractCgmesTapChangerBuilder 
         stepXforLinearAndSymmetrical();
     }
 
-    private void addStepsFromTable() {
-        PropertyBags table = context.phaseTapChangerTable(tableId);
-        if (table == null) {
-            return;
-        }
+    private void addStepsFromTable(PropertyBags table) {
         Comparator<PropertyBag> byStep = Comparator
                 .comparingInt((PropertyBag p) -> p.asInt(CgmesNames.STEP));
         table.sort(byStep);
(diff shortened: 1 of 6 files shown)


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/CgmesRatioTapChangerBuilder.java
```diff
@@ -39,14 +40,27 @@ protected void addRegulationData() {
     protected void addSteps() {
         String tableId = p.getId(CgmesNames.RATIO_TAP_CHANGER_TABLE);
         if (tableId != null) {
-            addStepsFromTable(tableId);
+            PropertyBags table = context.ratioTapChangerTable(tableId);
+            if (table == null) {
+                addStepsFromLowHighIncrement();
+                return;
+            }
+            int min = table.stream().map(p -> p.asInt(CgmesNames.STEP)).min(Integer::compareTo).orElseThrow(() -> new PowsyblException("Should at least contain one step"));
```
annetill, line 48: Just to be sure to understand well: if we have a table defined by `highStep`, `lowStep` and `voltageStepIncrement`, and if we also have a table, we check that the table is coherent with the increment definition before importing the data from the table. What is the criteria to define that coherence? The index? It should not be the size of the table?

miovd: We check that there is no "hole" in the table i.e. that the number of steps is continuous to chose if we keep or not the table. For example, if we have a table with a step 1, a step 3 but no step 2, we do not keep it. We keep however the table even if it has less steps than what the parameter `highStep` suggests.

marqueslanauja: Maybe, we can preserve the values of the table and only complete linearly the holes on the table. Try to use the table as much as possible.  If you want we can do it.

miovd: I don't think we need to go to that extent.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/CgmesPhaseTapChangerBuilder.java
```diff
@@ -62,7 +63,21 @@ protected void addSteps() {
...
+            int min = table.stream().map(p -> p.asInt(CgmesNames.STEP)).min(Integer::compareTo).orElseThrow(() -> new PowsyblException("Should at least contain one step"));
+            for (int i = min; i < min + table.size(); i++) {
+                int index = i;
+                if (table.stream().noneMatch(p -> p.asInt(CgmesNames.STEP) == index)) {
+                    context.ignored("PhaseTapChanger table", () -> String.format("There is at least one missing step (%s) in table %s. Phase tap changer considered linear", index, tableId));
+                    addStepsLinear();
+                    return;
+                }
+            }
+            addStepsFromTable(table);
```
marqueslanauja, line 80: When the tapChanger is built the lowTapPosition is set to the lowStep (AbstractCgmesTapChangerBuilder class)

   ```
protected TapChanger build() {
        int lowStep = p.asInt(CgmesNames.LOW_STEP);
        int highStep = p.asInt(CgmesNames.HIGH_STEP);
        int neutralStep = p.asInt(CgmesNames.NEUTRAL_STEP);
        int normalStep = p.asInt(CgmesNames.NORMAL_STEP, neutralStep);
        int position = initialTapPosition(normalStep);
        if (position > highStep || position < lowStep) {
            position = neutralStep;
        }
        tapChanger.setLowTapPosition(lowStep).setTapPosition(position);

```
so to be consistent we should take `min = lowStep` instead of obtaining it from the table.

Maybe we can use a common method, something  like `isTableValid`, defined in the AbstractCgmesTapChangerBuilder class for both, `RatioTapChanger `and `PhaseTapChanger`.

File: cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/CgmesRatioTapChangerBuilder.java [hunk 3]
```diff
@@ -39,14 +40,27 @@ protected void addRegulationData() {
...
+            for (int i = min; i < min + table.size(); i++) {
+                int index = i;
+                if (table.stream().noneMatch(p -> p.asInt(CgmesNames.STEP) == index)) {
+                    context.ignored("RatioTapChanger table", () -> String.format("There is at least one missing step (%s) in table %s", index, tableId));
+                    addStepsFromLowHighIncrement();
+                    return;
+                }
+            }
+            addStepsFromTable(table, tableId);
         } else {
```
marqueslanauja, line 58: Same remark as in PhaseTapChanger

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/CgmesPhaseTapChangerBuilder.java
miovd: I agree with you but as in quality requirements, the table has priority over the description by attributes, we should obtain it from the table in inconsistent files. However, I will refactor a bit as you said since the code seems alike indeed.

marqueslanauja (approved):
I reviewed and tested the implementation and works fine.

Only a minor thing. In the ignored message of  the`isTableValid `method we can change `PhaseTapChanger `and `Phase tap changer` to `TapChanger` and `tap changer` as the method is used by `RatioTapChanger` and `PhaseTapChanger`.

PR #1904, opened 2021-11-22
Add time series DSL module

geofjamg:

No

Design issue

`powsybl-time-series-api` depends on Groovy library and consequently `powsybl-dynamic-simulation-api` too.
Groovy dependency is not always necessary and sometimes could cause some issue like for GraalVM native image compilation. 

Part of `powsybl-time-series-api` code regarding Groovy DSL has be moved to new `powsybl-time-series-dsl` module.  

**Other information**:


time-series/time-series-api/src/main/java/com/powsybl/timeseries/CalculatedTimeSeriesDslLoader.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.ServiceLoader;
+
+/**
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ */
+public interface CalculatedTimeSeriesDslLoader {
+
+    Map<String, NodeCalc> load(String script, ReadOnlyTimeSeriesStore store);
+
+    static CalculatedTimeSeriesDslLoader find() {
```
miovd, line 22: Should we have have a "name" identifier for dsl loaders? Or we take one randomly? Please note that it may be done in another PR.

miovd (changes requested):
Don't forget to add time-series-dsl in distribution pom

time-series/time-series-api/src/main/java/com/powsybl/timeseries/CalculatedTimeSeriesDslLoader.java
geofjamg: As there is no plan for adding a second DSL language and worst no plan for dealing with 2 different DSL scripts engine at the same time in the classpath I would stay with a simple impl that suppose there is only one in the classpath.
So you agree with this proposition, my code is not correct as we should also check there is only one and I will add a check to ensure unicity.

miovd (approved):

PR #1914, opened 2021-11-29
Clarify SV import for terminal values in switches and dangling lines.

marqueslanauja:

Feature

In shunts and dangling lines we try  first to get values from SSH and only if there is not values we try from the SV. Shunts and dangling lines never have SSH values so terminal values (P and Q) are always taken from SV.

Take values directly from SV.

**Other information**:


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/AbstractConductingEquipmentConversion.java
```diff
@@ -254,7 +254,8 @@ public void convertToDanglingLine(int boundarySide, double r, double x, double g
         // In a Dangling Line the CGMES side and the IIDM side may not be the same
         // Dangling lines in IIDM only have one terminal, one side
         addMappingForTopologicalNode(dl, modelSide, 1);
-        context.convertedTerminal(terminalId(modelSide), dl.getTerminal(), 1, powerFlow(modelSide));
+        // We do not have SSH values at the model side, it is a line flow. We take directly SV values
+        context.convertedTerminal(terminalId(modelSide), dl.getTerminal(), 1, powerFlowSV(modelSide));
```
annetill, line 258: What's happen if we don't have any SV profile at the import? Do we really want to keep a dependency on SV profile ? I though we have decided to convert if only if a user parameter is active ? In the config .yml we have
`iidm.import.cgmes.profile-used-for-initial-state-values: SV`

annetill: Forgot my comment :-)

annetill (approved):

PR #1917, opened 2021-12-02
Replace CGMES tap changers properties by extensions

miovd:

No

Feature

Transformers have a lot of properties that are not really user-friendly

These properties are replaced by extensions

Yes

### DIFF
--- a/cgmes/cgmes-extensions/src/main/java/com/powsybl/cgmes/extensions/CgmesTapChangerAdder.java
+++ b/cgmes/cgmes-extensions/src/main/java/com/powsybl/cgmes/extensions/CgmesTapChangerAdder.java
@@ -1,0 +1,27 @@
+/**
+ * Copyright (c) 2021, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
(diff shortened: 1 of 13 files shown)


cgmes/cgmes-extensions/src/main/java/com/powsybl/cgmes/extensions/CgmesTapChangerAdder.java
```diff
@@ -0,0 +1,27 @@
...
+
+    CgmesTapChangerAdder setId(String id);
+
+    CgmesTapChangerAdder setCombinedTapChangerId(String combinedTapChangerId);
+
+    CgmesTapChangerAdder setType(String type);
+
+    CgmesTapChangerAdder setHiddenStatus(boolean hidden);
+
+    CgmesTapChangerAdder setStep(int step);
```
annetill, line 22: Why do we need to store the step here?

annetill (approved):

marqueslanauja (approved):
I reviewed the implementation and I checked against IOP 2021 March 10:30 and It is ok.

cgmes/cgmes-extensions/src/main/java/com/powsybl/cgmes/extensions/CgmesTapChangerAdder.java
marqueslanauja: It is the tapPosition of the hidden tapChanger.

PR #1923, opened 2021-12-09
Add generator entsoe category extension

colinepiloquet:

No

Extension


iidm/iidm-xml-converter/src/main/resources/xsd/generatorEntsoeCategory.xsd
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<xs:schema version="1.0"
+           xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           targetNamespace="http://www.itesla_project.eu/schema/iidm/ext/generator_entsoe_category/1_0"
+           elementFormDefault="qualified">
+    <xs:element name="entsoeCategory">
+        <xs:simpleType>
+            <xs:restriction base="xs:integer">
+                <xs:minInclusive value="1"/>
+                <xs:maxInclusive value="42"/>
```
geofjamg, line 10: I think we should remove this upper bound check, because they very often add new values and it would be painful and useless to upgrade the extension just for that.

iidm/iidm-xml-converter/src/test/resources/generatorEntsoeCategoryRef.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
geofjamg, line 1: Why do you need to have it here in addition to the 1.6 one?

iidm/iidm-xml-converter/src/test/resources/xsd/generatorEntsoeCategory.xsd
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
geofjamg, line 1: This is duplicated from runtime one.

geofjamg (changes requested):

geofjamg (approved):

geofjamg (changes requested):
Just the deprecated method that is used and it is ok.

geofjamg (approved):

PR #1935, opened 2022-01-07
add vertex validation to UndirectedGraphImpl

### LINKED ISSUE #1845: Restrict the node number in node breaker view
Feature

When creating a new node in a node breaker view, there is no restriction on the int number provided:
   - if very large, will create a unnecessary large array or even produce a OutOfMemoryError
   - if negative, the vertex created cannot be found after creation: it is created but at the index of the vertices size

Restrict the node number to [0, max], where max=?

Avoid misuse of node number

leblowl:

Fixes #1845

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractVoltageLevel.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractVoltageLevel.java
@@ -9,6 +9,7 @@ package com.powsybl.iidm.network.impl;
 import com.google.common.collect.FluentIterable;
 import com.google.common.collect.Lists;
 import com.google.common.primitives.Ints;
+import com.powsybl.commons.config.PlatformConfig;
 import com.powsybl.commons.PowsyblException;
 import com.powsybl.iidm.network.*;
 import com.powsybl.iidm.network.impl.util.Ref;
@@ -23,6 +24,10 @@ import java.util.stream.Stream;
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
  */
 abstract class AbstractVoltageLevel extends AbstractIdentifiable<VoltageLevel> implements VoltageLevelExt {
+
+    private static final int DEFAULT_NODE_INDEX_LIMIT = 1000;
+
+    public static final int NODE_INDEX_LIMIT = loadNodeIndexLimit(PlatformConfig.defaultConfig());
 
     private final Ref<NetworkImpl> networkRef;
 
@@ -42,6 +47,13 @@ abstract class AbstractVoltageLevel extends AbstractIdentifiable<VoltageLevel> i
         this.nominalV = nominalV;
         this.lowVoltageLimit = lowVoltageLimit;
         this.highVoltageLimit = highVoltageLimit;
+    }
+
+    protected static int loadNodeIndexLimit(PlatformConfig platformConfig) {
+        return platformConfig
+            .getOptionalModuleConfig("iidm")
+            .map(moduleConfig -> moduleConfig.getIntProperty("node-index-limit", DEFAULT_NODE_INDEX_LIMIT))
+            .orElse(DEFAULT_NODE_INDEX_LIMIT);
     }
 
     @Override
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusBreakerVoltageLevel.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusBreakerVoltageLevel.java
@@ -100,7 +100,7 @@ class BusBreakerVoltageLevel extends AbstractVoltageLevel {
 
     }
 
-    private final UndirectedGraphImpl<ConfiguredBus, SwitchImpl> graph = new UndirectedGraphImpl<>();
+    private final UndirectedGraphImpl<ConfiguredBus, SwitchImpl> graph = new UndirectedGraphImpl<>(NODE_INDEX_LIMIT);
 
     /* buses indexed by vertex number */
     private final Map<String, Integer> buses = new HashMap<>();
--- a/iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/TestAbstractVoltageLevel.java
+++ b/iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/TestAbstractVoltageLevel.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2021, RTE (http://www.rte-france.com)
+ * Copyright (c) 2022, The PowSyBl Authors
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
--- a/iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/TestAbstractVoltageLevel.java
+++ b/iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/TestAbstractVoltageLevel.java
@@ -1,0 +1,40 @@
+/**
+ * Copyright (c) 2021, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.iidm.network.impl;
+
+import com.google.common.jimfs.Configuration;
+import com.google.common.jimfs.Jimfs;
+import com.powsybl.commons.config.InMemoryPlatformConfig;
+import com.powsybl.commons.config.MapModuleConfig;
+import com.powsybl.commons.config.PlatformConfig;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.nio.file.FileSystem;
+
+import static org.junit.Assert.*;
+
+/**
+ * @author Lucas Leblow <lucasleblow@mailbox.org>
+ */
+public class TestAbstractVoltageLevel {
+
+    @Test
+    public void testLoadNodeIndexLimit() throws IOException {
+        assertEquals(1000, AbstractVoltageLevel.NODE_INDEX_LIMIT);
+        assertEquals(1000, AbstractVoltageLevel.loadNodeIndexLimit(PlatformConfig.defaultConfig()));
+
+        try (FileSystem fileSystem = Jimfs.newFileSystem(Configuration.unix())) {
+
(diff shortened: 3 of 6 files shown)


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusBreakerVoltageLevel.java
```diff
@@ -41,6 +42,13 @@
  */
 class BusBreakerVoltageLevel extends AbstractVoltageLevel {
 
+    private static final int DEFAULT_NODE_LIMIT = 1000;
+
+    public static final int NODE_LIMIT = PlatformConfig.defaultConfig()
+            .getOptionalModuleConfig("iidm")
+            .map(moduleConfig -> moduleConfig.getIntProperty("node-limit", DEFAULT_NODE_LIMIT))
+            .orElse(DEFAULT_NODE_LIMIT);
```
leblowl, line 50: @flo-dup I tried to copy the example you provided in #1845, but not totally sure if this is correct.

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/BusBreakerVoltageLevelTest.java
```diff
@@ -0,0 +1,85 @@
...
+import com.powsybl.math.graph.*;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * @author Florian Dupuy <florian.dupuy at rte-france.com>
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ */
+public class BusBreakerVoltageLevelTest {
```
leblowl, line 19: This test is copied from the powsybl-tutorials repo. I couldn't find existing BusBreakerVoltageLevel tests, so I added this file.

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/BusBreakerVoltageLevelTest.java
```diff
@@ -0,0 +1,85 @@
...
+    @Test
+    public void testBusBreakerVoltageLevel() {
+        VoltageLevel vl = createVoltageLevel();
+        assertEquals(1000, BusBreakerVoltageLevel.NODE_LIMIT);
+    }
+
+    @Test
+    public void testBusBreakerVoltageLevelConfig() {
+        VoltageLevel vl = createVoltageLevel();
+        assertEquals(5, BusBreakerVoltageLevel.NODE_LIMIT);
```
leblowl, line 83: @flo-dup I am not really sure how I can test the configuration here. I was thinking it would be good to configure the node-limit and verify that the BusBreakerVoltageLevel is using the correct value.

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevelTest.java
```diff
@@ -0,0 +1,146 @@
...
+import com.powsybl.math.graph.*;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * @author Florian Dupuy <florian.dupuy at rte-france.com>
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ */
+public class NodeBreakerVoltageLevelTest {
```
leblowl, line 19: This test is copied from the powsybl-tutorials repo. I couldn't find existing NodeBreakerVoltageLevel tests, so I added this file.

math/src/test/java/com/powsybl/math/graph/UndirectedGraphImplTest.java
```diff
@@ -76,6 +78,26 @@ public void testAddVertex() {
         assertEquals(1, graph.getVertexCount());
     }
 
+    @Test
+    public void testAddVertexIfNotPresent() {
+        graph.addVertexIfNotPresent(0);
+        assertEquals(1, graph.getVertexCount());
+        graph.addVertexIfNotPresent(0);
+        assertEquals(1, graph.getVertexCount());
```
flo-dup, line 86: Maybe also add `VERTEX_LIMIT - 1`

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusBreakerVoltageLevel.java
flo-dup: seems good to me!

flo-dup (changes requested):
Good job! Just a small refactor if you agree, and then you could add the config test you suggested

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusBreakerVoltageLevel.java
flo-dup: seems good to me!

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusBreakerVoltageLevel.java
```diff
@@ -41,6 +42,13 @@
  */
 class BusBreakerVoltageLevel extends AbstractVoltageLevel {
 
+    private static final int DEFAULT_NODE_LIMIT = 1000;
+
+    public static final int NODE_LIMIT = PlatformConfig.defaultConfig()
```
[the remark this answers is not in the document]
leblowl: Yes, I think that's a better place for it and makes it easier to test. Thanks!

leblowl:
@flo-dup I made some changes in response to your feedback, I hope the changes align with what you were thinking.

I moved the node index limit properties and configuration code to the AbstractVoltageLevel class and am setting the NODE_INDEX_LIMIT property statically with a new static `loadNodeIndexLimit` method: https://github.com/powsybl/powsybl-core/blob/9c415ecb4a9a9e1222aeb4d11bef750b614a5956/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractVoltageLevel.java#L30

I was looking at how some other classes are using the configuration and some are using this static initializer approach: https://github.com/powsybl/powsybl-core/blob/76d5a4b8e209837cb8b4d612b9b5edca2365a0b5/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/LimitViolationFilter.java#L43-L58
We could do that too, it just seemed like it would require a little more refactoring, for example here: https://github.com/powsybl/powsybl-core/blob/76d5a4b8e209837cb8b4d612b9b5edca2365a0b5/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VoltageLevelAdderImpl.java#L99-L108

I removed the Node/Bus VoltageLevel tests and replaced them with a single AbstractVoltageLevel test which tests the `loadNodeIndexLimit` method currently.

I made some small wording/test changes as suggested too.

flo-dup (changes requested):
Great! could you just correct your signoff, so that it passes the DCO test? The first commit is signed off with @mailbox.com but expects @gmail.com, and the second one is not signed off. Then we could have a look at the sonar scan results.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractVoltageLevel.java
```diff
@@ -24,6 +25,10 @@
  */
 abstract class AbstractVoltageLevel extends AbstractIdentifiable<VoltageLevel> implements VoltageLevelExt {
 
+    private static final int DEFAULT_NODE_INDEX_LIMIT = 1000;
+
+    public static final int NODE_INDEX_LIMIT = loadNodeIndexLimit(PlatformConfig.defaultConfig());
```
miovd, line 30: Isn't there some risks doing that? It means that the `NODE_INDEX_LIMIT` is fixed when creating the voltage level but what if the user changes it? However, not sure if this is really relevant as it is not a parameter that should change a lot... What do you think @flo-dup @geofjamg ?

miovd (commented):
Thank you for your work! I have a question about the definition of `NODE_INDEX_LIMIT` from the configuration but otherwise, great execution.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractVoltageLevel.java
flo-dup: What I had in mind was a high number which would fit in most if not all cases - not sure 1000 is a good enough candidate for that. And then it should be hardly ever changed, only for very specific network with huge voltage levels. Or with voltage levels of any size but poorly numbered, that is, with high indices.

leblowl:
@flo-dup It looks like I got the DCO passing, but I am not seeing the Sonar scan running. Is there something I can do?

flo-dup:
> @flo-dup It looks like I got the DCO passing, but I am not seeing the Sonar scan running. Is there something I can do?

I haven't experienced that so far, strange... I'm wondering if it's not because you forked powsybl-core. Maybe if I dig into sonar settings, or otherwise a solution would be for you to create a copycat pull request on a branch of powsybl-core itself? You should have the rights to do it.

leblowl:
I created a new branch in powsybl-core and opened a copy PR: https://github.com/powsybl/powsybl-core/pull/1940 which seemed to start the other validation steps, which also got reflected in this PR. Since the checks seem to be tied to the commit, we can just keep using this PR and I can remove the copy or vice versa. For the future, I can create a branch in powsybl-core

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractVoltageLevel.java
miovd: Even in this case, shouldn't then this parameter not be static?

miovd (approved):

flo-dup (changes requested):

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/TestAbstractVoltageLevel.java
```diff
@@ -0,0 +1,40 @@
+/**
+ * Copyright (c) 2021, RTE (http://www.rte-france.com)
```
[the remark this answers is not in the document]
flo-dup: Sorry for unresolving, but in fact after asking a colleague more experienced in that matter, it should rather be your name directly instead of "powsybl authors". Plus it doesn't really make sense to put "Powsybl authors" only in this file and not in the others...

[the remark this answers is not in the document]
leblowl: No worries, fixed!

flo-dup (approved):

PR #1938, opened 2022-01-12
Matpower: update resources to be consistent with the value of the phase shift angle

marqueslanauja:

Feature

Voltages correspond to case with phase shift 0 deg

Voltages consistent with the value of the phase shift angle (-5 deg). They correspond to the solved case on Matpower.

**Other information**:


zamarrenolm (approved):

zamarrenolm:
Could we call load flow results validation to check that the case is solved? like it is done in the IEEE CDF importer test in the method `testSolved`.

marqueslanauja:
> Could we call load flow results validation to check that the case is solved? like it is done in the IEEE CDF importer test in the method `testSolved`.

Done

zamarrenolm (approved):

matpower/matpower-converter/src/test/java/com/powsybl/matpower/converter/MatpowerImporterTest.java
```diff
@@ -136,4 +149,52 @@ private void testNetwork(Network network, String id) throws IOException {
...
+        config.setVerbose(true);
+        config.setThreshold(threshold);
+        config.setOkMissingValues(false);
+        LoadFlowParameters lf = new LoadFlowParameters();
+        lf.setTwtSplitShuntAdmittance(true);
+        config.setLoadFlowParameters(lf);
+        return config;
+    }
+
+    private static void computeMissingFlows(Network network, LoadFlowParameters lfparams) {
```
annetill, line 182: Just to understand well: here we are performing a loadflow validation to check that the solved case is well filled. In the parameter of the loadflow validation, I understand that the slack is distributed (or not ?) on generator. So why do we fill missing value with `targetP` ? And maybe some generators have gone from PV to PQ, so why do we fill the Q of the terminal with targetQ. How does it work ?

marqueslanauja: I modified the code to be more clear. We only set the `SSH` values in loads and generators  if the `P` and `Q` terminal values  are `NaN` (not initialized). We import a solved case and we perform a loadflow validation that fills the flows of the branches using the received voltages and angles and calculates a balance in all the buses.

Maybe we can include both setting loops (load and generators) inside the `LoadFlowResultsCompletion` method. We can do it in a new PR as we have to review all the `LoadFlowResultsCompletion` references.

annetill: When you say "a solved case", what are the parameters of the load flow computation ? How was the slack distributed ? It was solved by open load flow ? It was the LoadFlow default parameter except the split shunt admittance parameter to true ?

marqueslanauja: I solved on the `matpower` loadflow and I have not specified any parameter so the case was solved with the default matpower parameters.

PR #1939, opened 2022-01-12
Replace simulShunt in LoadFlowParameters by shuntCompensatorVoltageControlOn

### LINKED ISSUE #1929: LoadFlowParameters simulShunt

A improvement indeed.

In `LoadFlowParameters`, when we want to activate shunt's voltage control, the parameter to use is `simulShunt`. But this term is not really explicit compared to `transformerVoltageControlOn` or `phaseShifterRegulationOn`.

I propose to change `simulShunt` for `shuntVoltageControlOn`.

  - PowSyBl Version: ...
  - OS Version: ...

miovd:

Yes. Fix #1929 

Yes


loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowParameters.java
```diff
@@ -354,7 +372,7 @@ public BalanceType getBalanceType() {
                 .put("noGeneratorReactiveLimits", noGeneratorReactiveLimits)
                 .put("phaseShifterRegulationOn", phaseShifterRegulationOn)
                 .put("twtSplitShuntAdmittance", twtSplitShuntAdmittance)
-                .put("simulShunt", simulShunt)
+                .put("simulShunt", shuntCompensatorVoltageControlOn)
```
flo-dup, line 375: you could also replace the key here

flo-dup (approved):

annetill (approved):

PR #1940, opened 2022-01-14
(copy) add vertex validation to UndirectedGraphImpl and node (vertex) limit …

### LINKED ISSUE #1845: Restrict the node number in node breaker view

leblowl:
…configuration in NodeBreakerVoltageLevel and BusBreakerVoltageLevel

Signed-off-by: Lucas Leblow <lucasleblow@mailbox.org>

Fixes #1845


flo-dup:
Copy of #1935 used for sonarcloud analysis

PR #1943, opened 2022-01-17
fix getBaseName in DataSourceUtil so we can read files with multiple dots in their name

### LINKED ISSUE #1934: CGMES import crash when the file name has the dot character

A bug

When importing a cgmes network from a zip file, if the zip file is formatted with a version number like "3.0" before the extension ".zip" then the import will crash with an error similar to : 

```
com.powsybl.cgmes.model.CgmesModelException: Listing CGMES names in data source com.powsybl.commons.datasource.ZipFileDataSource@3f6f6701
	at com.powsybl.cgmes.model.CgmesOnDataSource.names(CgmesOnDataSource.java:73)
	at com.powsybl.cgmes.model.CgmesOnDataSource.namespaces(CgmesOnDataSource.java:96)
	at com.powsybl.cgmes.model.CgmesOnDataSource.exists(CgmesOnDataSource.java:38)
	at com.powsybl.cgmes.conversion.CgmesImport.exists(CgmesImport.java:84)
	at com.powsybl.iidm.import_.Importers.findImporter(Importers.java:437)
	at com.powsybl.iidm.import_.Importers.loadNetwork(Importers.java:461)
	at com.powsybl.iidm.import_.Importers.loadNetwork(Importers.java:483)
	at com.powsybl.iidm.import_.Importers.loadNetwork(Importers.java:497)
	at com.powsybl.iidm.import_.Importers.loadNetwork(Importers.java:512)
	at com.powsybl.iidm.import_.Importers.loadNetwork(Importers.java:522)
	at com.powsybl.cgmes.conversion.CgmesImportPostProcessorTest.testImportCornet(CgmesImportPostProcessorTest.java:107)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:64)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:564)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:69)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:33)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:220)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:53)
Caused by: java.nio.file.NoSuchFileException: D:\tmp\ENTSO-E_v3.0\RealGrid\RealGrid-Merged_v3.zip
	at java.base/sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:85)
	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103)
	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:108)
	at java.base/sun.nio.fs.WindowsFileSystemProvider.newByteChannel(WindowsFileSystemProvider.java:235)
	at java.base/java.nio.file.Files.newByteChannel(Files.java:375)
	at java.base/java.nio.file.Files.newByteChannel(Files.java:426)
	at com.powsybl.commons.datasource.ZipFileDataSource.listNames(ZipFileDataSource.java:209)
	at com.powsybl.cgmes.model.CgmesOnDataSource.names(CgmesOnDataSource.java:69)
	... 37 more
```

Import a CGMES network like this : 

`Network importedNetwork = Importers.loadNetwork("D:\\tmp\\ENTSO-E_v3.0\\RealGrid\\RealGrid-Merged_v3.0.zip");`

Then the import crash with the error above. If the zip file is renamed RealGrid-Merged.zip, then the import succeed. See the attached cgmes zip file.

The success of the import should not depend on the name of the file.

After some debugging it seems the error may come from the class DataSourceUtil in function getBaseName at line 37.

`int pos = fileName.indexOf('.'); // find first dot in case of double extension (.xml.gz)
`

  - PowSyBl Version: ...
  - OS Version: ...

[RealGrid-Merged_v3.0.zip](https://github.com/powsybl/powsybl-core/files/7808346/RealGrid-Merged_v3.0.zip)

leblowl:
Signed-off-by: Lucas Leblow <lucasleblow@mailbox.org>

Fixes #1934 

It looks like there is an issue with DataSourceUtil.getBaseName where if you
pass it a filename with multiple dots, it returns a truncated base name. For
example, if passing `test_1.0.xml` as a file name, it would return `test_1` as a
basename. This leads to some errors in the CGMES importer and the iTools
network-convert command, for example:

```
$ distribution-core/target/powsybl/bin/itools convert-network --input-file cgmes/cgmes-conversion/src/test/resources/RealGrid-Merged_v3.0.zip --output-file test.zip --output-format AMPL
com.powsybl.cgmes.model.CgmesModelException: Listing CGMES names in data source com.powsybl.commons.datasource.ZipFileDataSource@23348b5d
        at com.powsybl.cgmes.model.CgmesOnDataSource.names(CgmesOnDataSource.java:73)
        at com.powsybl.cgmes.model.CgmesOnDataSource.namespaces(CgmesOnDataSource.java:96)
        at com.powsybl.cgmes.model.CgmesOnDataSource.exists(CgmesOnDataSource.java:38)
        at com.powsybl.cgmes.conversion.CgmesImport.exists(CgmesImport.java:84)
        at com.powsybl.iidm.import_.Importers.findImporter(Importers.java:437)
        at com.powsybl.iidm.import_.Importers.loadNetwork(Importers.java:461)
        at com.powsybl.iidm.import_.Importers.loadNetwork(Importers.java:483)
        at com.powsybl.iidm.import_.Importers.loadNetwork(Importers.java:497)
        at com.powsybl.iidm.tools.ConversionTool.run(ConversionTool.java:111)
        at com.powsybl.tools.CommandLineTools.run(CommandLineTools.java:164)
        at com.powsybl.tools.Main.main(Main.java:29)
Caused by: java.nio.file.NoSuchFileException: /home/lucas/src/powsybl/powsybl-core/cgmes/cgmes-conversion/src/test/resources/RealGrid-Merged_v3.zip
        at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
        at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
        at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:116)
        at java.base/sun.nio.fs.UnixFileSystemProvider.newByteChannel(UnixFileSystemProvider.java:219)
        at java.base/java.nio.file.Files.newByteChannel(Files.java:373)
        at java.base/java.nio.file.Files.newByteChannel(Files.java:424)
        at com.powsybl.commons.datasource.ZipFileDataSource.listNames(ZipFileDataSource.java:209)
        at com.powsybl.cgmes.model.CgmesOnDataSource.names(CgmesOnDataSource.java:69)
        ... 10 more
```

This PR fixes the issue by using the whole filename up to the last dot as the
base name. In the case of files ending with `.xml.gz`, `.xml` is part of the
basename and `.gz` is the extension.

Here is the same command above with the fix:
```
distribution-core/target/powsybl/bin/itools convert-network --input-file cgmes/cgmes-conversion/src/test/resources/RealGrid-Merged_v3.0.zip --output-file test.zip --output-format AMPL
Generating file /home/lucas/src/powsybl/powsybl-core/test.zip:test_network_buses.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.zip:test_network_tct.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.zip:test_network_rtc.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.zip:test_network_ptc.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.zip:test_network_branches.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.zip:test_network_limits.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.zip:test_network_generators.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.zip:test_network_batteries.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.zip:test_network_loads.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.zip:test_network_shunts.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.zip:test_network_static_var_compensators.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.zip:test_network_substations.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.zip:test_network_vsc_converter_stations.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.zip:test_network_lcc_converter_stations.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.zip:test_network_hvdc.txt...
```

also tested multiple dots in the `ouput-file`:
```
distribution-core/target/powsybl/bin/itools convert-network --input-file cgmes/cgmes-conversion/src/test/resources/RealGrid-Merged_v3.0.zip --output-file test.3.0.zip --output-format AMPL
Generating file /home/lucas/src/powsybl/powsybl-core/test.3.0.zip:test.3.0_network_buses.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.3.0.zip:test.3.0_network_tct.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.3.0.zip:test.3.0_network_rtc.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.3.0.zip:test.3.0_network_ptc.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.3.0.zip:test.3.0_network_branches.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.3.0.zip:test.3.0_network_limits.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.3.0.zip:test.3.0_network_generators.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.3.0.zip:test.3.0_network_batteries.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.3.0.zip:test.3.0_network_loads.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.3.0.zip:test.3.0_network_shunts.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.3.0.zip:test.3.0_network_static_var_compensators.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.3.0.zip:test.3.0_network_substations.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.3.0.zip:test.3.0_network_vsc_converter_stations.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.3.0.zip:test.3.0_network_lcc_converter_stations.txt...
Generating file /home/lucas/src/powsybl/powsybl-core/test.3.0.zip:test.3.0_network_hvdc.txt...
```

getBaseName is used in another place (https://github.com/powsybl/powsybl-core/blob/c6d7506d190c38dac3d2ce14aae3e87bd30f8959/action/action-simulator/src/main/java/com/powsybl/action/simulator/tools/ActionSimulatorTool.java#L251), which is also causing an error:
```
distribution-core/target/powsybl/bin/itools action-simulator --case-file test.3.0.xiidm --dsl-file dsl.groovy
Loading network 'test.3.0.xiidm'
com.powsybl.commons.PowsyblException: Unsupported file format or invalid file.
        at com.powsybl.iidm.import_.Importers.loadNetwork(Importers.java:472)
        at com.powsybl.iidm.import_.Importers.loadNetwork(Importers.java:486)
        at com.powsybl.iidm.import_.Importers.loadNetwork(Importers.java:500)
        at com.powsybl.action.simulator.tools.ActionSimulatorTool.run(ActionSimulatorTool.java:259)
        at com.powsybl.tools.CommandLineTools.run(CommandLineTools.java:164)
        at com.powsybl.tools.Main.main(Main.java:29)
```

and is fixed with this change:
```
distribution-core/target/powsybl/bin/itools action-simulator --case-file test.3.0.xiidm --dsl-file dsl.groovy
Loading network 'test.3.0.xiidm'
Loading DSL 'file:/home/lucas/src/powsybl/powsybl-core/dsl.groovy'
Using 'loadflow' rules engine
Starting pre-contingency analysis
    Round 0
2022-01-17_16:22:54.408 [ForkJoinPool.commonPool-worker-3] ERROR c.p.o.n.impl.LfNetworkLoaderImpl - Discard network {CC0 SC0} because there is no equipment to control voltage
    Load flow diverged
Final result
Pre-contingency violations:
+--------+---------------+-----+---------+--------------+----------------+----------------+-------+-------+------------------+----------------+
| Action | Equipment (0) | End | Country | Base voltage | Violation type | Violation name | Value | Limit | abs(value-limit) | Loading rate % |
+--------+---------------+-----+---------+--------------+----------------+----------------+-------+-------+------------------+----------------+
```

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
@@ -15,7 +15,6 @@ import com.powsybl.cgmes.model.CgmesModelFactory;
 import com.powsybl.cgmes.model.CgmesOnDataSource;
 import com.powsybl.commons.config.PlatformConfig;
 import com.powsybl.commons.datasource.DataSource;
-import com.powsybl.commons.datasource.DataSourceUtil;
 import com.powsybl.commons.datasource.GenericReadOnlyDataSource;
 import com.powsybl.commons.datasource.ReadOnlyDataSource;
 import com.powsybl.commons.util.ServiceLoaderCache;
@@ -129,7 +128,7 @@ public class CgmesImport implements Importer {
                         boundaryLocationParameter,
                         defaultValueConfig));
         // Check that the Data Source has valid CGMES names
-        ReadOnlyDataSource ds = new GenericReadOnlyDataSource(loc, DataSourceUtil.getBaseName(loc));
+        ReadOnlyDataSource ds = new GenericReadOnlyDataSource(loc, loc.getFileName().toString());
         if ((new CgmesOnDataSource(ds)).names().isEmpty()) {
             return null;
         }
--- a/commons/src/main/java/com/powsybl/commons/datasource/Bzip2FileDataSource.java
+++ b/commons/src/main/java/com/powsybl/commons/datasource/Bzip2FileDataSource.java
@@ -17,12 +17,12 @@ import java.nio.file.Path;
  */
 public class Bzip2FileDataSource extends FileDataSource {
 
-    public Bzip2FileDataSource(Path directory, String baseName, DataSourceObserver observer) {
-        super(directory, baseName, observer);
+    public Bzip2FileDataSource(Path directory, String fileName, DataSourceObserver observer) {
+        super(directory, fileName, observer);
     }
 
-    public Bzip2FileDataSource(Path directory, String baseName) {
-        super(directory, baseName);
+    public Bzip2FileDataSource(Path directory, String fileName) {
+        super(directory, fileName);
     }
 
     @Override
--- a/commons/src/main/java/com/powsybl/commons/datasource/DataSourceUtil.java
+++ b/commons/src/main/java/com/powsybl/commons/datasource/DataSourceUtil.java
@@ -34,7 +34,7 @@ public interface DataSourceUtil {
 
     static String getBaseName(String fileName) {
         Objects.requireNonNull(fileName);
-        int pos = fileName.indexOf('.'); // find first dot in case of double extension (.xml.gz)
+        int pos = fileName.lastIndexOf('.'); // find last dot
         return pos == -1 ? fileName : fileName.substring(0, pos);
     }
 
@@ -63,11 +63,11 @@ public interface DataSourceUtil {
         Objects.requireNonNull(fileNameOrBaseName);
 
         if (fileNameOrBaseName.endsWith(".zip")) {
-            return new ZipFileDataSource(directory, getBaseName(fileNameOrBaseName.substring(0, fileNameOrBaseName.length() - 4)), observer);
+            return new ZipFileDataSource(directory, getBaseName(fileNameOrBaseName), observer);
         } else if (fileNameOrBaseName.endsWith(".gz")) {
-            return new GzFileDataSource(directory, getBaseName(fileNameOrBaseName.substring(0, fileNameOrBaseName.length() - 3)), observer);
+            return new GzFileDataSource(directory, getBaseName(fileNameOrBaseName), observer);
         } else if (fileNameOrBaseName.endsWith(".bz2")) {
-            return new Bzip2FileDataSource(directory, getBaseName(fileNameOrBaseName.substring(0, fileNameOrBaseName.length() - 4)), observer);
+            return new Bzip2FileDataSource(directory, getBaseName(fileNameOrBaseName), observer);
         } else {
             return new FileDataSource(directory, getBaseName(fileNameOrBaseName), observer);
         }
--- a/commons/src/main/java/com/powsybl/commons/datasource/DataSourceUtil.java
+++ b/commons/src/main/java/com/powsybl/commons/datasource/DataSourceUtil.java
@@ -47,29 +47,29 @@ public interface DataSourceUtil {
         } else {
             switch (compressionExtension) {
                 case GZIP:
-                    return new GzFileDataSource(directory, basename, observer);
+                    return new GzFileDataSource(directory, basename + ".gz", observer);
                 case BZIP2:
-                    return new Bzip2FileDataSource(directory, basename, observer);
+                    return new Bzip2FileDataSource(directory, basename + ".bz2", observer);
                 case ZIP:
-                    return new ZipFileDataSource(directory, basename, observer);
+                    return new ZipFileDataSource(directory, basename + ".zip", observer);
                 default:
                     throw new AssertionError("Unexpected CompressionFormat value: " + compressionExtension);
             }
         }
     }
 
-    static DataSource createDataSource(Path directory, String fileNameOrBaseName, DataSourceObserver observer) {
+    static DataSource createDataSource(Path directory, String fileName, DataSourceObserver observer) {
         Objects.requireNonNull(directory);
-        Objects.requireNonNull(fileNameOrBaseName);
+        Objects.requireNonNull(fileName);
 
-        if (fileNameOrBaseName.endsWith(".zip")) {
-            return new ZipFileDataSource(directory, getBaseName(fileNameOrBaseName), observer);
-        } else if (fileNameOrBaseName.endsWith(".gz")) {
-            return new GzFileDataSource(directory, getBaseName(fileNameOrBaseName), observer);
-        } else if (fileNameOrBaseName.endsWith(".bz2")) {
-            return new Bzip2FileDataSource(directory, getBaseName(fileNameOrBaseName), observer);
+        if (fileName.endsWith(".zip")) {
+            return new ZipFileDataSource(directory, fileName, observer);
+        } else if (fileName.endsWith(".gz")) {
+            return new GzFileDataSource(directory, fileName, observer);
+        } else if (fileName.endsWith(".bz2")) {
+            return new Bzip2FileDataSource(directory, fileName, observer);
         } else {
-            return new FileDataSource(directory, getBaseName(fileNameOrBaseName), observer);
+            return new FileDataSource(directory, fileName, observer);
         }
     }
 }
--- a/commons/src/main/java/com/powsybl/commons/datasource/FileDataSource.java
+++ b/commons/src/main/java/com/powsybl/commons/datasource/FileDataSource.java
@@ -26,23 +26,23 @@ public class FileDataSource implements DataSource {
 
     private final Path directory;
 
-    private final String baseName;
+    private final String fileName;
 
     private final DataSourceObserver observer;
 
-    public FileDataSource(Path directory, String baseName) {
-        this(directory, baseName, null);
+    public FileDataSource(Path directory, String fileName) {
+        this(directory, fileName, null);
     }
 
-    public FileDataSource(Path directory, String baseName, DataSourceObserver observer) {
+    public FileDataSource(Path directory, String fileName, DataSourceObserver observer) {
         this.directory = Objects.requireNonNull(directory);
-        this.baseName = Objects.requireNonNull(baseName);
+        this.fileName = Objects.requireNonNull(fileName);
         this.observer = observer;
     }
 
     @Override
     public String getBaseName() {
-        return baseName;
+        return DataSourceUtil.getBaseName(fileName);
     }
 
     protected String getCompressionExt() {
@@ -59,12 +59,12 @@ public class FileDataSource implements DataSource {
 
     private Path getPath(String fileName) {
         Objects.requireNonNull(fileName);
-        return directory.resolve(fileName + getCompressionExt());
+        return directory.resolve(fileName);
     }
 
     @Override
     public OutputStream newOutputStream(String suffix, String ext, boolean append) throws IOException {
-        return newOutputStream(DataSourceUtil.getFileName(baseName, suffix, ext), append);
+        return newOutputStream(DataSourceUtil.getFileName(DataSourceUtil.getBaseName(fileName), suffix, ext), append);
     }
 
     @Override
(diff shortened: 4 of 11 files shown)


flo-dup (changes requested):
Not exactly: we still want to get "dummy" in the unit test you changed. The double extension should still be removed in case of gzipped files. I think that a gzipped file is always compressing a single unique file, so we should look for the second to last dot for them. Maybe we should also double check that the string between the two last dots it is a 3 (or 2) letter extension?

leblowl:
@flo-dup thanks, I see now that I didn't test thoroughly enough; I think my initial attempt might break every importer except for CGMES!

To @obrix's point, I think in general I agree the success of the import shouldn't depend on the name of the file… if it doesn't need to. In PowSyBl, we use the file extension to determine the type of file so that we can import it correctly, but in many cases it looks like the file extension is only used as an optimization (to short-circuit the file validation), since (except in MatpowerImporter) we still parse the file contents to verify the validity of the file:

- https://github.com/powsybl/powsybl-core/blob/c6d7506d190c38dac3d2ce14aae3e87bd30f8959/iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/XMLImporter.java#L120
- https://github.com/powsybl/powsybl-core/blob/c6d7506d190c38dac3d2ce14aae3e87bd30f8959/ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java#L880
- https://github.com/powsybl/powsybl-core/blob/c6d7506d190c38dac3d2ce14aae3e87bd30f8959/psse/psse-converter/src/main/java/com/powsybl/psse/converter/PsseImporter.java#L95
- https://github.com/powsybl/powsybl-core/blob/c6d7506d190c38dac3d2ce14aae3e87bd30f8959/ieee-cdf/ieee-cdf-converter/src/main/java/com/powsybl/ieeecdf/converter/IeeeCdfImporter.java#L86-L89

I think whether to depend on the file extension or not is perhaps another discussion. But I think we should try to accept as many file names as possible to reduce friction for users. Having versions with dots in a file name and multiple dots in general is very common. And I think where things can get complicated is when we make assumptions about what the base name is. In most libraries I have seen the base name is simply defined as everything in the file name up to the last dot (e.g. https://commons.apache.org/proper/commons-io/javadocs/api-2.7/) which I think might be the only reliable way to do it unless we do an exhaustive check for "double" extensions, e.g. after removing the first extension (`file.xml.gz` -> `file.xml`), then only removing the second extension if it is one of: `[".xml", ".iidm.xml", ".txt", …]`. I haven't test it yet, but I think that should work with the existing code and it might be the least invasive change that would allow users to import/export files like "file.v1.0.xml.zip". However, if someone adds a new importer with a new extension (e.g. ".new"), they would have to update that list of extensions. What do you think of this approach?

Alternatively, I don't really see the need to have an API that depends so much on the base name of a file, so I made some bigger changes in a new commit that depend more on the file name rather than the base name, which to me seems like a simpler foundation to build on. It's just a WIP commit in order to communicate what I've been thinking, and I haven't changed any tests, but it appears to be working with most importers and data sources that I've tested. What do you think of going in this direction? (it still needs work, but the general idea should be there).

flo-dup:
This has (eventually!) been fixed in latest release [v6.5.0](https://github.com/powsybl/powsybl-core/releases/tag/v6.5.0) through a big work of @rolnico around the DataSources. Solving this had indirect implications on various uses, which made the fix quite tedious. I think this is showing that the design should be improved... still some work to come on that topic I hope!

PR #1947, opened 2022-01-21
Improve three phase shortcircuits API to support elementary computation

colinepiloquet:

No

Adds API to run three phase selective shortcircuits

For now, only systematic three phase short-circuits are supported


shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitBusResult.java
```diff
@@ -0,0 +1,160 @@
...
+/**
+ * @author Coline Piloquet <coline.piloquet at rte-france.com>
+ */
+public class ShortCircuitBusResult {
+    private double threePhaseFaultCurrent;
+    private double threePhaseFaultPower;
+    private double timeConstant;
+    private double faultResistance;
+    private double faultReactance;
+    private double currentValue;
```
colinepiloquet, line 21: Add direct in the name

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitBusResult.java
```diff
@@ -0,0 +1,160 @@
...
+    private double currentAngle;
+    private double voltageValue;
+    private double voltageAngle;
+    private ValuesOnThreePhases currentValuePerPhase;
```
colinepiloquet, line 25: Value => magnitude

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitVoltageMapResult.java
```diff
@@ -0,0 +1,86 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.shortcircuit;
+
+/**
+ * @author Coline Piloquet <coline.piloquet at rte-france.com>
+ */
+public class ShortCircuitVoltageMapResult {
```
annetill, line 12: I think that we don't need this class. In my mind, a selective short circuit calculation will provide a list, a map (or a monitoring as we have in security analysis) of `ShortCircuitBusResult`. In each bus result, we will find voltage magnitudes and angles by phase, and the values for the symmetrical calculations. This bus result will be increased with phase to phase results too. This has to be specified with @JB-H.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitAnalysisResult.java
```diff
@@ -42,15 +38,6 @@ public ShortCircuitAnalysisResult(List<FaultResult> faultResults, List<LimitViol
         return Collections.unmodifiableList(faultResults);
     }
 
-    /**
-     * The list of limit violations: for instance when the computed short-circuit current on a given equipment is higher
-     * than the maximum admissible value for that equipment. In general, the equipment ID can be completed by the side where
-     * the violation occurs. In a first simple approach, the equipment is a voltage level, and no side is needed.
-     */
-    public List<LimitViolation> getLimitViolations() {
```
sylvlecl, line 50: If we move them to inner classes, it would be nice to keep a convenience method here which concatenates them, so that we can easily get all limit violations at once.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitParameters.java
```diff
@@ -48,9 +58,22 @@ public static ShortCircuitParameters load(PlatformConfig platformConfig) {
 
         ModuleConfig config = platformConfig.getOptionalModuleConfig("short-circuit-parameters").orElse(null);
         if (config != null) {
+            //TODO: add parameter for a list of equipments on which the analysis should be done (sytematic study but with specified equipments)
             parameters.setSubTransStudy(config.getBooleanProperty("subTransStudy", ShortCircuitConstants.SUBTRANS_STUDY));
             parameters.setWithFeederResult(config.getBooleanProperty("withFeederResult", ShortCircuitConstants.WITH_FEEDER_RESULT));
-
+            parameters.setStudyType(config.getEnumProperty("study-type", ShortCircuitConstants.StudyType.class, DEFAULT_STUDY_TYPE));
+            if (parameters.studyType == ShortCircuitConstants.StudyType.SELECTIVE_STUDY) {
```
sylvlecl, line 65: I have a very high level interrogation : 

selective and systematic studies are quite different things, from the feature point of view (comparable to loadflow and security analysis).

Maybe it would be a better idea to separate them in 2 different APIs, with possibly different inputs and result structures ?

sylvlecl: Another, also high level consideration:

I think all those parameters which define the fault would find a better place as an argument of the computation (like contingencies), than in parameters.

Parameters should only be pretty generic parameters of the computation tool, not something that will often change from one computation to another.

sylvlecl (commented):
High level considerations:
 - I think it may be better to separate systematic and selective studies in 2 APIs, typically with different inputs, and maybe different result structures
 - Also I think inputs of the computation such as the definition of the fault (its location, resistance, etc), should go into an argument to the "run" method, and not in parameters.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitConstants.java
```diff
@@ -14,7 +14,23 @@ public final class ShortCircuitConstants {
...
+        BRANCH_STUDY,
+    }
 
+    public enum ImpedanceConnection {
+        SERIES,
+        PARALLEL,
+    }
+
+    public static final boolean SUBTRANS_STUDY = false;
     public static final boolean WITH_FEEDER_RESULT = true;
```
JB-H, line 33: Shouldn't simplify this at a first implementation:
- if SYSTEMATIC_STUDY, we don't display non-mandatory data (all bus voltages, feeder)
- if SELECTIVE_STUDY, then everything is provided (maybe check user practices...)

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitParameters.java
JB-H: @sylvlecl : this is slightly different from a loadflow because in a loadflow, results of systematic analysis are obtained by re-running a new LF computation for each fault. Here, one single computation (one matrix inversion) is valid for all faults of the same type (ex: triphased). Additional info results from a post processing of mandatory information. 
But from a user point of view, I don't know what is the best...

I agree with @sylvlecl , we need to seperate parameters that define the tool general behaviour and input data that describes each contingency.

Parameters:
- subTransStudy
- Type of Norm used (IEC, IEEE, RTE =COURCIRC)
- shunt/no shunt
- type of voltage profile for short circuit computation (provided in input or use of nominal values)

General Input data for shortcircuit analysis
- SELECTIVE versus SYSTEMATIC
- equipment if SELECTIVE
- "mandatory result data only" or "detailed data" if selective (could be set to "detailed data" by default if selective)
- fault impedance (resistance + reactance)
- fault location
- Type of shortcircuit: triphased, monophased, biphased, biphased ground. Later, additional information will be needed to fully describe unbalanced shortcircuits.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/DetailedShortCircuitValue.java
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * @author Coline Piloquet <coline.piloquet at rte-france.com>
+ */
+public class DetailedShortCircuitValue {
+    /**
+     * The aim of this class is to store the magnitude and angle of current or voltage on the three phases, and the
+     * the magnitude and angle of current or voltage for the Fortescue direct, zero and indirect components.
+     */
+    // Results on three phases.
+    private double magnitude1;
```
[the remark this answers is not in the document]
JB-H: I suggest to name it fortescue vector: it could be a current, a voltage, an impedance (ex: thevenin impedance)

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FaultResult.java
```diff
@@ -50,6 +78,15 @@ public double getThreePhaseFaultCurrent() {
         return threePhaseFaultCurrent;
     }
 
+    /**
+     * The list of limit violations: for instance when the computed short-circuit current on a given equipment is higher
+     * than the maximum admissible value for that equipment. In general, the equipment ID can be completed by the side where
+     * the violation occurs. In a first simple approach, the equipment is a voltage level, and no side is needed.
+     */
+    public List<LimitViolation> getLimitViolations() {
+        return Collections.unmodifiableList(limitViolations);
```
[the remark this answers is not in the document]
JB-H: Thevenin impedance is missing in the result.
It may be modeled as a fortescue vector [Zthdx, Zthdy, Zthox, Zthoy, Zthix, Zthiy]

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FeederResult.java
```diff
@@ -17,7 +17,7 @@
 public class FeederResult {
 
     /**
-     * The ID of the connectable contributing to the three phase short circuit current
+     * The ID of the connectable contributing to the three phase short circuit current.
      */
     private final String connectableId;
```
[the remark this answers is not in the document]
annetill: Indeed, a list of feeder results is associated to a fault result. In a fault result, you have, in principle the id of a bus (or in a first approach the id of a voltage level), to be checked with @colinepiloquet. It means that you have always only one side of a feeder connected to that bus (or voltage level). Maybe in some specific cases, we can have both...

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FeederResult.java
```diff
@@ -26,18 +26,31 @@ public class FeederResult {
      */
     private final double feederThreePhaseCurrent;
```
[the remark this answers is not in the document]
annetill: I am thinking about making `current`, that contains really detailed values, optional in order to light the API in case of symmetrical computation. The other option could be to have 2 data structures, one `ThreePhaseValue` for assymmetrical calculation and another one `OnePhaseValue` (or any better proposal) for symmetrical computations.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitAnalysisResult.java
annetill: Solved, the symmetrical computation API is conserved, but renamed `ShortCircuitAnalysisMultiResult`.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitConstants.java
```diff
@@ -14,7 +14,23 @@ public final class ShortCircuitConstants {
...
+        BUS_STUDY,
+        BRANCH_STUDY,
+    }
 
+    public enum ImpedanceConnection {
+        SERIES,
+        PARALLEL,
+    }
+
+    public static final boolean SUBTRANS_STUDY = false;
```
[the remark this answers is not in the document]
annetill: Okay, so @colinepiloquet please provide an enum for this parameter.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitConstants.java
annetill: As we will have two different `run`, one for systematic study, one for selective study, this parameter could be removed.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitParameters.java
annetill: In case of a selective short circuit analysis, we will have a dedicated `run` method, that we have to design. I agree, we can dedicate a new object to fault description, that will be in the argument of the run: fault location, fault impedance. Do you think we have to add the type of short-circuit calculation here too? Triphased, monophased, biphased, biphased ground. Or in parameters ?

JB-H: In the same run, you could ask to simulate a fault for triphased, biphased etc. and take the most conservative result. It is maybe better to link it to a fault rather then a parameter of the model. In the long term, it would be nice to have systematic shortcircuit analysis mixing different types of balanced and unbalanced faults

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FeederResult.java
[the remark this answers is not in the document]
JB-H: See my comment above, having a unique structure but more dynamic to be able to compare balanced/unbalanced results and avoid to store useless information in case of balanced

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/Fault.java
```diff
@@ -0,0 +1,57 @@
...
+
+    // How the fault impedance is connected to the network.
+    enum ConnectionType {
+        SERIES,
+        PARALLEL,
+    }
+
+    // What kind of fault is simulated
+    enum FaultType {
+        THREEPHASE,
```
annetill, line 25: `THREE_PHASE`, `TWO_PHASE` and `SINGLE_PHASE`?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/Fault.java
```diff
@@ -0,0 +1,57 @@
...
+
+    // The equipment or bus id where the fault is simulated.
+    String getId();
+
+    default double getProportionalLocation() {
+        return Double.NaN;
+    }
+
+    // Characteristics of the short-circuit.
+    double getR();
```
annetill, line 40: `R` and `X` should have default values. I think that we have to introduce parameters that provide these default values, no? In the `ShortCircuitParameters`.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/Fault.java
```diff
@@ -0,0 +1,57 @@
...
+        TWOPHASE,
+        SINGLEPHASE,
+    }
+
+    //TODO : add the numbers of the phase for two and single phase
+
+    // The equipment or bus id where the fault is simulated.
+    String getId();
+
+    default double getProportionalLocation() {
```
annetill, line 35: Optional?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/Fault.java
```diff
@@ -0,0 +1,57 @@
...
+
+    default FaultType getFaultType() {
+        return FaultType.THREEPHASE;
+    }
+
+    // Whether the result should indicate a limit violation
+    boolean isWithLimitViolation();
+
+    // Whether the results should include the voltage map on the whole network
+    boolean isWithDetailedResults();
```
annetill, line 56: Can you find a more explicit name linked to the network voltage plane?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FaultResult.java
```diff
@@ -16,31 +17,58 @@
...
  *
  * @author Boubakeur Brahimi
+ * @author Coline Piloquet <coline.piloquet at rte-france.com>
  */
 public final class FaultResult extends AbstractExtendable<FaultResult> {
 
-    private final String id;
+    private final Fault fault;
 
     private final double threePhaseFaultCurrent;
```
annetill, line 26: After our talk, it seems that there is a kind of redondance between `threePhaseFaultCurrent` and `current`. I continue to think that it is useful to expose only the `threePhaseFaultCurrent`. @sylvlecl what do you think about the `FaultResult` design?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/AbstractFault.java
```diff
@@ -0,0 +1,69 @@
...
+        return this.connection;
+    }
+
+    @Override
+    public FaultType getFaultType() {
+        return this.faultType;
+    }
+
+    @Override
+    public boolean isWithLimitViolation() {
```
annetill, line 61: Can you change the name of this method to have just `withLimitViolations`?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/AbstractFault.java
```diff
@@ -0,0 +1,69 @@
...
+        return this.withLimitViolation;
+    }
+
+    @Override
+    public boolean isWithDetailedResults() {
```
annetill, line 66: Can you change the name of this method to has just `withVoltageMap`?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/Fault.java
annetill: Can you take this into account?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FeederResult.java
```diff
@@ -16,28 +16,38 @@
...
-    /**
-     * The ID of the connectable contributing to the three phase short circuit current
-     */
     private final String connectableId;
 
-    /**
-     * The value of the current of the connectable contributing to the total fault current in kA
-     */
-    private final double feederThreePhaseCurrent;
+    private ThreePhaseValue current = null;
```
annetill, line 21: Can you make it final to be coherent with the rest?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitAnalysisResult.java
```diff
@@ -60,4 +48,7 @@ public ShortCircuitAnalysisResult setNetworkMetadata(NetworkMetadata networkMeta
         return this;
     }
 
+    public List<LimitViolation> getLimitViolations() {
```
annetill, line 51: I think that here you have to index by fault id.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/AbstractFault.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Override
+    public boolean withLimitViolations() {
+        return this.withLimitViolation;
+    }
+
+    @Override
+    public boolean withVoltageMap() {
+        return this.withDetailedResults;
```
annetill, line 67: Can you change the attributes too?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/BranchFault.java
```diff
@@ -0,0 +1,31 @@
...
+ * Used for elementary short-circuit calculation only.
+ *
+ * @author Anne Tilloy <anne.tilloy at rte-france.com>
+ */
+public class BranchFault extends AbstractFault {
+
+    // Location of the fault in % of the branch length (with side ONE as reference).
+    private final double proportionalLocation;
+
+    public BranchFault(String id, double r, double x, Fault.ConnectionType connection, Fault.FaultType faultType, boolean withLimitViolation, boolean withDetailedResults,
```
annetill, line 20: Here too.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/BusFault.java
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Abstract class to describe the characteristics of the fault to be simulated.
+ * Used for elementary short-circuit calculation only.
+ *
+ * @author Anne Tilloy <anne.tilloy at rte-france.com>
+ */
+public class BusFault extends AbstractFault {
+
+    public BusFault(String id, double r, double x, Fault.ConnectionType connection, Fault.FaultType faultType, boolean withLimitViolation, boolean withDetailedResults) {
```
annetill, line 17: Here too.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FeederResult.java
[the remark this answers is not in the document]
annetill: Now I wonder about a fault that occurs in the middle of a line. What means a List<FeederResult> in that case? The `connectableId` will be each side of the line itself ?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitConstants.java
```diff
@@ -14,7 +14,13 @@ public final class ShortCircuitConstants {
...
 
-    public static final boolean SUBTRANS_STUDY = false;
+    public enum StudyType {
+        SUB_TRANSIENT,
+        TRANSIENT,
+        STEADY_STATE
+    }
 
+    public static final StudyType DEFAULT_STUDY_TYPE = StudyType.TRANSIENT;
     public static final boolean WITH_FEEDER_RESULT = true;
```
annetill, line 24: This boolean can be removed, except if we choose to have a global parameter that will be a priority over the boolean we have inside each fault description.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/Fault.java
```diff
@@ -0,0 +1,60 @@
...
+    default ConnectionType getConnectionType() {
+        return ConnectionType.SERIES;
+    }
+
+    default FaultType getFaultType() {
+        return FaultType.THREE_PHASE;
+    }
+
+    // Whether the result should indicate a limit violation
+    boolean withLimitViolations();
```
sylvlecl, line 56: **About the question of having those options in the fault or not:**

As discussed, I think it's not very good to mix computation options within an object which has a business signification outside the context of this computation, the same way we don't include computation options in `Contingency`.

Another option would be to include it in parameters, for example as a list of identifiers.
But this does not look good either, because parameters should not depend too much on the network of one particular computation.

Maybe a third and better alternative, and more consistent with security analysis API, is to take inspiration from the `StateMonitor` mechanism : we could provide an additional argument to the computation to specify for which faults we want detailed results

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/Fault.java
```diff
@@ -0,0 +1,60 @@
...
+        TWO_PHASE,
+        SINGLE_PHASE,
+    }
+
+    //TODO : add the numbers of the phase for two and single phase
+
+    // The equipment or bus id where the fault is simulated.
+    String getId();
+
+    default double getProportionalLocation() {
```
sylvlecl, line 38: This should not be part of the `Fault` class but only present in branch faults

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/Fault.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+package com.powsybl.shortcircuit;
+
+/**
+ * Interface to describe the characteristics of the fault to be simulated.
+ * Used for elementary short-circuit calculation only.
+ *
+ * @author Anne Tilloy <anne.tilloy at rte-france.com>
+ */
+public interface Fault {
```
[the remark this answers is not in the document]
sylvlecl: We need to add an enum for the type of fault (bus or branch):
this will be used to downcast, and to serialize/deserialize.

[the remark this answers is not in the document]
sylvlecl: > It would be better to have classes for faults: we will never want to have alternative implementations --> I thin we can promote the `AbstractFault` to `Fault` :)

After thinking again: not necessary after all.
Interface is fine as long as:
- we have an enum telling the type of the implementation
- we remove info specific to an implementation from the interface (proportional location)

If we go for a class, it will be a little strange since it should not be instantiable, and we will need to provide a dummy implementation for `getType` (or make it abstract, bu then we have a naming issue with checkstyle rules).

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/converter/FaultDeserializer.java
```diff
@@ -0,0 +1,89 @@
...
+        Fault fault;
+        switch (type) {
+            case BUS:
+                fault = new BusFault(id, r, x, connection, faultType);
+                break;
+            case BRANCH:
+                fault = new BranchFault(id, r, x, connection, faultType, proportionalLocation);
+                break;
+            default:
+                throw new AssertionError("Unexpected datatype: " + type.name());
```
sylvlecl, line 85: datatype -> type

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitParameters.java
```diff
@@ -25,17 +26,23 @@
  */
 public class ShortCircuitParameters extends AbstractExtendable<ShortCircuitParameters> {
```
sylvlecl, line 27: Note that we should add some javadoc for all parameters: but can wait for API to be validated ?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FaultResult.java
```diff
@@ -16,45 +18,87 @@
...
  * @author Boubakeur Brahimi
+ * @author Coline Piloquet <coline.piloquet at rte-france.com>
  */
 public final class FaultResult extends AbstractExtendable<FaultResult> {
 
-    private final String id;
+    private final Fault fault;
 
-    private final double threePhaseFaultCurrent;
+    private final double threePhaseFaultActivePower;
```
sylvlecl, line 27: Naming:
maybe @JB-H has a better vision of what we want to put in here, but I think the right name would be "short circuit power". Actually it seems it has nothing to do with active power, it's computed by convention using nominal voltage.

From what I understand, it does not make sense either for non-symmetrical faults, so "three phase" is redundant.

sylvlecl (commented):
Several kinds of comments:
 - minor impl details
 - naming questions, to be agreed upon
 - some usage-related questions, that we will only be able to answer by implementing / getting more knowledge of the use cases

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitAnalysisResult.java
```diff
@@ -10,45 +10,30 @@
...
-     * The list of limit violations: for instance when the computed short-circuit current on a given equipment is higher
-     * than the maximum admissible value for that equipment. In general, the equipment ID can be completed by the side where
-     * the violation occurs. In a first simple approach, the equipment is a voltage level, and no side is needed.
-     */
-    public List<LimitViolation> getLimitViolations() {
-        return Collections.unmodifiableList(limitViolations);
+        return faultResults;
     }
 
     public NetworkMetadata getNetworkMetadata() {
```
geofjamg, line 39: Why do we need this?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitAnalysisResult.java
```diff
@@ -60,4 +45,9 @@ public ShortCircuitAnalysisResult setNetworkMetadata(NetworkMetadata networkMeta
         return this;
     }
 
+    public Map<String, List<LimitViolation>> getLimitViolations() {
```
geofjamg, line 48: To remove, instead we could just get a FaultResult by its ID:

getFaultResult(String id)

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitAnalysisResult.java
```diff
@@ -10,45 +10,30 @@
...
+ * Will contain a list of short-circuit bus result for each buses of the network.
  *
  * @author Boubakeur Brahimi
  */
 public class ShortCircuitAnalysisResult extends AbstractExtendable<ShortCircuitAnalysisResult> {
 
     private NetworkMetadata networkMetadata;
 
-    private final List<FaultResult> faultResults = new ArrayList<>();
+    private final List<FaultResult> faultResults;
```
geofjamg, line 26: To replace by Map<String, FaultResult>

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FaultResult.java
```diff
@@ -16,45 +18,79 @@
...
 
-    private final List<FeederResult> feederResults; // optional
+    private final Duration timeConstant;
```
geofjamg, line 29: What is the meaning/purpose of this attribute?
Is it something that could really be filled by a static simulator?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FaultResult.java
```diff
@@ -16,45 +18,79 @@
...
+    private final FortescueValue voltage;
+
+    private final List<ShortCircuitBusResults> shortCircuitBusResults;
+
+    public FaultResult(Fault fault, double threePhaseFaultActivePower, List<FeederResult> feederResults,
+                       List<LimitViolation> limitViolations, FortescueValue current, FortescueValue voltage, List<ShortCircuitBusResults> shortCircuitBusResults,
+                       Duration timeConstant) {
+        this.fault = Objects.requireNonNull(fault);
+        this.threePhaseFaultActivePower = threePhaseFaultActivePower;
         this.feederResults = List.copyOf(feederResults);
```
geofjamg, line 46: All copies are useless as only simulators will build these objects

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/option/FaultOptions.java
```diff
@@ -0,0 +1,102 @@
...
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * @author Thomas Adam <tadam at silicom.fr>
+ */
+public class FaultOptions {
```
geofjamg, line 24: I don't understand this design, FaultOptions could only be linked to one Fault through a FaultContext:
  - FaultContext is a useless object only made of one ID field
  - As it is not embeded in Fault class, we need to resolve link between Fault and FaultOption after using a utility class FaultOptionsIndex which is painful.
 
For me this is a very poor design API that will be painful to use:

Why not just embed all these parameters in the fault?
I would understand to have it separated in case of multiple same options associated to a set of fault, but here this is not the case. We have just a one to one relation between Fault and FaultOptions.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitConstants.java
```diff
@@ -14,7 +14,13 @@ public final class ShortCircuitConstants {
     private ShortCircuitConstants() {
     }
 
-    public static final boolean SUBTRANS_STUDY = false;
+    public enum StudyType {
```
geofjamg, line 17: Should be moved outside of ShortCircuitConstants

geofjamg (changes requested):

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitParameters.java
```diff
@@ -25,17 +26,23 @@
...
 
+    // VERSION = 1.0
+    public static final String VERSION = "1.0";
+
     private static final Supplier<ExtensionProviders<ConfigLoader>> SUPPLIER = Suppliers
             .memoize(() -> ExtensionProviders.createProvider(ConfigLoader.class, "short-circuit-parameters"));
 
+    private boolean withFeederResult = WITH_FEEDER_RESULT;
+    private StudyType studyType = DEFAULT_STUDY_TYPE;
+    private double minVoltageDropProportionalThreshold = DEFAULT_MIN_VOLTAGE_DROP_PROPORTIONAL_THRESHOLD;
```
colinepiloquet, line 41: This field should also be in the FaultParameters class.

sylvlecl (approved):
Nice work :)
I think we should decide about the following remaining details pretty soon to avoid verion updates:
- `threePhaseFaultActivePower` result attribute: I am pretty sure it's not the right naming, and maybe not something that we systematically want (does it make sense for unbalanced short circuits?)
- `feederResults`: would they include results for feeders that are not connected to the fault bus ? If not, don't we need a place for that ?

annetill (approved):
I think it is a good first version. As I told you before, this will be confirmed or reviewed by the models integration. So, still a beta version. Thanks everybody!

geofjamg (approved):

PR #1951, opened 2022-01-24
User functional logs reporter in import data variants and importer wrapper

zamarrenolm:

No

Fix importer wrapper not passing the functional logs reporter to the inner importer.

Importer wrapper passes the reporter to the post-processors, but not to its inner importer.

The importer wrapper passes the reporter also to the inner reporter.

**Other information**:
In addition, `Importer.importData` has been extended with variants that allow passing a functional logs `Reporter`.


flo-dup (approved):

iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/import_/Importers.java
```diff
@@ -280,18 +280,36 @@ public static Importer removePostProcessors(Importer importer) {
      * @param dataSource data source
      * @param parameters some properties to configure the import
      * @param computationManager computation manager to use for default post processors
+     * @param config the import configuration
+     * @param reporter the reporter used for functional logs
      * @return the model
      */
-    public static Network importData(ImportersLoader loader, String format, ReadOnlyDataSource dataSource, Properties parameters, ComputationManager computationManager, ImportConfig config) {
```
miovd, line 285: You should still keep this signature method calling Reporter.NO_OP

iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/import_/Importers.java
```diff
@@ -280,18 +280,36 @@ public static Importer removePostProcessors(Importer importer) {
...
         Importer importer = getImporter(loader, format, computationManager, config);
         if (importer == null) {
             throw new PowsyblException("Import format " + format + " not supported");
         }
-        return importer.importData(dataSource, NetworkFactory.findDefault(), parameters);
+        if (reporter == Reporter.NO_OP) {
+            return importer.importData(dataSource, NetworkFactory.findDefault(), parameters);
+        } else {
+            return importer.importData(dataSource, NetworkFactory.findDefault(), parameters, reporter);
+        }
```
miovd, line 296: We shouldn't need to do this distinction

miovd (changes requested):

iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/import_/Importers.java
zamarrenolm: Yes, the signature with all previous parameters is being kept, and calls this one with Reporter.NO_OP. It is the variant just below this one in the new code. I choose to update this lines to include the parameter instead of adding new lines in another part of the code because this is the only variant that is completely documented

iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/import_/Importers.java
zamarrenolm: you are right. fixed.

miovd (approved):

miovd:
Thank you for the corrections! Can you please add some coverage test?

PR #1955, opened 2022-01-26
Fix bug: switches were not removed from index when all edges were removed

miovd:

No

Bug fix

Switches are not removed from index and still appear (for example in `getSwitchesCount()`) after all edges removed. 

Yes

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusBreakerVoltageLevel.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusBreakerVoltageLevel.java
@@ -353,7 +353,7 @@ class BusBreakerVoltageLevel extends AbstractVoltageLevel {
             }
 
             @Override
-            public void allEdgesRemoved() {
+            public void allEdgesRemoved(Collection<SwitchImpl> obj) {
                 invalidateCache();
             }
         });
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevel.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevel.java
@@ -528,18 +528,14 @@ class NodeBreakerVoltageLevel extends AbstractVoltageLevel {
             }
 
             @Override
-            public void allEdgesRemoved() {
-                List<String> removedSwitchesIds = new ArrayList<>(graph.getEdgeCount());
+            public void allEdgesRemoved(Collection<SwitchImpl> aSwitches) {
                 NetworkImpl network = getNetwork();
-                for (SwitchImpl s : graph.getEdgesObject()) {
-                    if (s != null) {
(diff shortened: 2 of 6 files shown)


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusBreakerVoltageLevel.java
```diff
@@ -353,7 +353,7 @@ public void edgeRemoved(int e, SwitchImpl obj) {
             }
 
             @Override
-            public void allEdgesRemoved() {
+            public void allEdgesRemoved(Collection<SwitchImpl> obj) {
                 invalidateCache();
```
flo-dup, line 357: Sorry I'm missing something here, why didn't we remove the switches from the network index in bus breaker view?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevel.java
```diff
@@ -528,18 +528,14 @@ public void edgeRemoved(int e, SwitchImpl aSwitch) {
...
+            public void allEdgesRemoved(Collection<SwitchImpl> aSwitches) {
                 NetworkImpl network = getNetwork();
-                for (SwitchImpl s : graph.getEdgesObject()) {
-                    if (s != null) {
-                        network.getListeners().notifyBeforeRemoval(s);
-                        removedSwitchesIds.add(s.getId());
-                        network.getIndex().remove(s);
-                    }
-                }
+                aSwitches.forEach(ss -> {
```
flo-dup, line 533: Is there any reason to have a difference between the `aSwitches` passed on and the `switches` map?

flo-dup (commented):
It's worth adding an assert of `Network::getSwitchCount` in `AbstractNodeBreakerInternalConnectionsTest::testRemoveVoltageLevelWithInternalConnectionsIssue`, don't you think? or a specific test

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusBreakerVoltageLevel.java
miovd: We do remove the switches from the network index in bus breaker view but it is not done in the graph listener (I don't actually remember why)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevel.java
miovd: `aSwitches` contains the list of switches while `switches` contains ids and the associated edge's number

miovd:
> `testRemoveVoltageLevelWithInternalConnectionsIssue`

This test is not really related as the issue appeared only with `removeAllSwitches` (when removing a whole voltage level for example). However, I added a test in `DefaultNetworkReducerTest`

flo-dup (approved):

Issue #1954, opened 2022-01-26
Increasing `ShortCircuitAnalysisResult`


annetill:
* **Do you want to request a *feature* or report a *bug*?**

A feature. For the moment, the results of a symmetrical short circuit calculation are the limit violations and the fault result. The fault result is a simple class with:
```
public final class FaultResult extends AbstractExtendable<FaultResult> {

    private final String id;

    private final float threePhaseFaultCurrent;
```
For the moment, and as it is written in the java doc, the `id` could be anything but for the private implementation the `id` is the voltage level `id`. But after calculation we get the three phase current by bus (from the bus view indeed) and the part of each equipment connected to the bus of this three phase current. We want to have access to these results. Is it a generic result ?

* **What is the current behavior?**

See above.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**

For security analysis, we have a class called `BusResults`. We can imagine having increasing `FaultResult` to store the id of the bus and a list or a map to given for each connectable connected to the bus the contribution to the three phase current of the bus. Any idea ? 

* **What is the motivation / use case for changing the behavior?**

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

sylvlecl:
I agree it would be better to have the ID of the bus available.

The reason behind using voltage level ID, like in voltage violations, was that calculated buses are not "stable" and therefore the ID of the bus might be invalid shortly after the computation. However I think it's a far bigger problem to not be able at all to differentiate the result of 2 buses in the same voltage level.

annetill:
Solved !

PR #1956, opened 2022-01-28
Add contributions to the shortcircuits results

colinepiloquet:

#1954

### DIFF
--- a/shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ContributionResult.java
+++ b/shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ContributionResult.java
@@ -10,19 +10,13 @@ package com.powsybl.shortcircuit;
  * @author Coline Piloquet <coline.piloquet at rte-france.com>
  */
 public class ContributionResult {
-    private final String subjectId;
     private final String id;
 
     private final float contribution; //in kA
 
-    public ContributionResult(String subjectId, String id, float contribution) {
-        this.subjectId = subjectId;
+    public ContributionResult(String id, float contribution) {
         this.id = id;
         this.contribution = contribution;
-    }
-
-    public String getSubjectId() {
-        return subjectId;
     }
 
     public String getId() {
--- a/shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ContributionResult.java
+++ b/shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ContributionResult.java
@@ -10,23 +10,23 @@ package com.powsybl.shortcircuit;
  * @author Coline Piloquet <coline.piloquet at rte-france.com>
  */
 public class ContributionResult {
-    private final String bus;
-    private final String name;
+    private final String subjectId;
+    private final String id;
 
     private final float contribution; //in kA
 
-    public ContributionResult(String bus, String name, float contribution) {
-        this.bus = bus;
-        this.name = name;
+    public ContributionResult(String subjectId, String id, float contribution) {
+        this.subjectId = subjectId;
+        this.id = id;
         this.contribution = contribution;
     }
 
-    public String getBus() {
-        return bus;
+    public String getSubjectId() {
+        return subjectId;
     }
 
-    public String getName() {
-        return name;
+    public String getId() {
+        return id;
     }
 
     public float getContribution() {
--- a/shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ContributionResult.java
+++ b/shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ContributionResult.java
@@ -1,0 +1,35 @@
+/**
+ * Copyright (c) 2022, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.shortcircuit;
+
+/**
+ * @author Coline Piloquet <coline.piloquet at rte-france.com>
+ */
+public class ContributionResult {
+    private final String bus;
+    private final String name;
+
+    private final float contribution; //in kA
+
+    public ContributionResult(String bus, String name, float contribution) {
+        this.bus = bus;
+        this.name = name;
+        this.contribution = contribution;
+    }
+
+    public String getBus() {
+        return bus;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public float getContribution() {
+        return contribution;
+    }
+}
--- a/shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FaultResult.java
+++ b/shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FaultResult.java
@@ -8,6 +8,7 @@ package com.powsybl.shortcircuit;
 
 import com.powsybl.commons.extensions.AbstractExtendable;
 
+import java.util.List;
 import java.util.Objects;
 
 /**
@@ -21,9 +22,16 @@ public final class FaultResult extends AbstractExtendable<FaultResult> {
 
     private final float threePhaseFaultCurrent;
 
-    public FaultResult(String id, float threePhaseFaultCurrent) {
+    private List<ContributionResult> contributionResults; //in case of systemactic study, optional
+
+    public FaultResult(String id, float threePhaseFaultCurrent, List<ContributionResult> contributionResults) {
         this.id = Objects.requireNonNull(id);
         this.threePhaseFaultCurrent = threePhaseFaultCurrent;
+        this.contributionResults = contributionResults;
+    }
+
+    public FaultResult(String id, float threePhaseFaultCurrent) {
+        this(id, threePhaseFaultCurrent, null);
     }
 
     /**
@@ -41,4 +49,10 @@ public final class FaultResult extends AbstractExtendable<FaultResult> {
         return threePhaseFaultCurrent;
     }
 
+    /**
+     * List of contributions of each connectable connected to the equipment to the three phase fault current
+     */
+    public List<ContributionResult> getContributionResults() {
+        return contributionResults;
+    }
 }
--- a/shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitAnalysisResult.java
+++ b/shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitAnalysisResult.java
@@ -30,12 +30,12 @@ public class ShortCircuitAnalysisResult extends AbstractExtendable<ShortCircuitA
 
     private final List<LimitViolation> limitViolations = new ArrayList<>();
 
-    private final List<ContributionResult> contributionResults = new ArrayList<>();
+    private List<ContributionResult> contributionResults;
 
     public ShortCircuitAnalysisResult(List<FaultResult> faultResults, List<LimitViolation> limitViolations, List<ContributionResult> contributionResults) {
         this.faultResults.addAll(Objects.requireNonNull(faultResults));
         this.limitViolations.addAll(Objects.requireNonNull(limitViolations));
-        this.contributionResults.addAll(Objects.requireNonNull(contributionResults));
+        this.contributionResults = contributionResults;
     }
 
     public ShortCircuitAnalysisResult(List<FaultResult> faultResults, List<LimitViolation> limitViolations) {
--- a/shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitAnalysisResult.java
+++ b/shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitAnalysisResult.java
@@ -30,9 +30,16 @@ public class ShortCircuitAnalysisResult extends AbstractExtendable<ShortCircuitA
 
     private final List<LimitViolation> limitViolations = new ArrayList<>();
 
-    public ShortCircuitAnalysisResult(List<FaultResult> faultResults, List<LimitViolation> limitViolations) {
+    private final List<ContributionResult> contributionResults = new ArrayList<>();
+
+    public ShortCircuitAnalysisResult(List<FaultResult> faultResults, List<LimitViolation> limitViolations, List<ContributionResult> contributionResults) {
         this.faultResults.addAll(Objects.requireNonNull(faultResults));
         this.limitViolations.addAll(Objects.requireNonNull(limitViolations));
+        this.contributionResults.addAll(Objects.requireNonNull(contributionResults));
+    }
+
+    public ShortCircuitAnalysisResult(List<FaultResult> faultResults, List<LimitViolation> limitViolations) {
+        this(faultResults, limitViolations, null);
     }
 
     /**
@@ -53,6 +60,10 @@ public class ShortCircuitAnalysisResult extends AbstractExtendable<ShortCircuitA
 
     public NetworkMetadata getNetworkMetadata() {
         return networkMetadata;
+    }
+
+    public List<ContributionResult> getContributionResults() {
+        return Collections.unmodifiableList(contributionResults);
     }
 
     public ShortCircuitAnalysisResult setNetworkMetadata(NetworkMetadata networkMetadata) {
--- a/shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitConstants.java
+++ b/shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitConstants.java
@@ -15,4 +15,6 @@ public final class ShortCircuitConstants {
     }
 
     public static final boolean SUBTRANS_STUDY = false;
+
+    public static final boolean WITH_CONTRIBUTIONS = true;
 }
--- a/shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitParameters.java
+++ b/shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitParameters.java
@@ -27,6 +27,8 @@ public class ShortCircuitParameters extends AbstractExtendable<ShortCircuitParam
 
     private boolean subTransStudy = ShortCircuitConstants.SUBTRANS_STUDY;
 
+    private boolean withContributions = ShortCircuitConstants.WITH_CONTRIBUTIONS; //In case of systematic study only
+
     public interface ConfigLoader<E extends Extension<ShortCircuitParameters>>
             extends ExtensionConfigLoader<ShortCircuitParameters, E> {
     }
@@ -47,8 +49,9 @@ public class ShortCircuitParameters extends AbstractExtendable<ShortCircuitParam
         ModuleConfig config = platformConfig.getOptionalModuleConfig("short-circuit-parameters").orElse(null);
         if (config != null) {
             parameters.setSubTransStudy(config.getBooleanProperty("subTransStudy", ShortCircuitConstants.SUBTRANS_STUDY));
-        }
+            parameters.setWithContributions(config.getBooleanProperty("withContributions", ShortCircuitConstants.WITH_CONTRIBUTIONS));
 
+        }
         return parameters;
     }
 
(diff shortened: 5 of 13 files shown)


shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ContributionResult.java
```diff
@@ -0,0 +1,29 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.shortcircuit;
+
+/**
+ * @author Coline Piloquet <coline.piloquet at rte-france.com>
+ */
+public class ContributionResult {
+    private final String id;
```
colinepiloquet, line 13: Maybe this class could also have the id of the bus as an attribute

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ContributionResult.java
```diff
@@ -0,0 +1,29 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.shortcircuit;
+
+/**
+ * @author Coline Piloquet <coline.piloquet at rte-france.com>
+ */
+public class ContributionResult {
```
sylvlecl, line 12: Should we rename to it "feeder result" or "feeder contribution" to make explicit that it's associated to a feeder ?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ContributionResult.java
sylvlecl: I think it will be a duplicate of the one in fault result, so I would not be in favor of this.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FaultResult.java
```diff
@@ -21,9 +22,16 @@
 
     private final float threePhaseFaultCurrent;
 
-    public FaultResult(String id, float threePhaseFaultCurrent) {
+    private List<ContributionResult> contributionResults; //in case of systemactic study, optional
```
sylvlecl, line 25: can be final :)

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FaultResult.java
```diff
@@ -21,9 +22,16 @@
...
+
+    public FaultResult(String id, float threePhaseFaultCurrent, List<ContributionResult> contributionResults) {
         this.id = Objects.requireNonNull(id);
         this.threePhaseFaultCurrent = threePhaseFaultCurrent;
+        this.contributionResults = contributionResults;
```
sylvlecl, line 30: probably better o make an immutable copy (`List.copyOf`)

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FaultResult.java
```diff
@@ -21,9 +22,16 @@
...
+    }
+
+    public FaultResult(String id, float threePhaseFaultCurrent) {
+        this(id, threePhaseFaultCurrent, null);
```
sylvlecl, line 34: Probably better to use an empty list here to be consistent with powsybl habits

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitAnalysisResult.java
```diff
@@ -30,9 +30,16 @@
...
+    private List<ContributionResult> contributionResults;
+
+    public ShortCircuitAnalysisResult(List<FaultResult> faultResults, List<LimitViolation> limitViolations, List<ContributionResult> contributionResults) {
         this.faultResults.addAll(Objects.requireNonNull(faultResults));
         this.limitViolations.addAll(Objects.requireNonNull(limitViolations));
+        this.contributionResults = contributionResults;
+    }
+
+    public ShortCircuitAnalysisResult(List<FaultResult> faultResults, List<LimitViolation> limitViolations) {
+        this(faultResults, limitViolations, null);
```
sylvlecl, line 42: Same: probably better to use an empty list

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitAnalysisResult.java
```diff
@@ -30,9 +30,16 @@
 
     private final List<LimitViolation> limitViolations = new ArrayList<>();
 
-    public ShortCircuitAnalysisResult(List<FaultResult> faultResults, List<LimitViolation> limitViolations) {
+    private List<ContributionResult> contributionResults;
+
+    public ShortCircuitAnalysisResult(List<FaultResult> faultResults, List<LimitViolation> limitViolations, List<ContributionResult> contributionResults) {
         this.faultResults.addAll(Objects.requireNonNull(faultResults));
         this.limitViolations.addAll(Objects.requireNonNull(limitViolations));
+        this.contributionResults = contributionResults;
```
sylvlecl, line 38: Same here: better to make an immutable copy, + we should check not null if we go for an empty list instead of null.

Actually we should do it also for the other existing fields.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitAnalysisResult.java
```diff
@@ -30,9 +30,16 @@
 
     private final List<LimitViolation> limitViolations = new ArrayList<>();
 
-    public ShortCircuitAnalysisResult(List<FaultResult> faultResults, List<LimitViolation> limitViolations) {
+    private List<ContributionResult> contributionResults;
```
sylvlecl, line 33: Can be final!

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/converter/FaultResultDeserializer.java
```diff
@@ -52,6 +54,14 @@ public FaultResult deserialize(JsonParser parser, DeserializationContext deseria
                     threePhaseFaultCurrent = parser.readValueAs(Float.class);
                     break;
 
+                case "contributionsResult":
+                    while (parser.nextToken() != JsonToken.END_OBJECT) {
```
sylvlecl, line 58: we'll need to add a unit test for testing the serialization/deserialization.
We can deserialize as a List instead of iterating explicitly.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitAnalysisResult.java
```diff
@@ -55,6 +62,10 @@ public NetworkMetadata getNetworkMetadata() {
         return networkMetadata;
     }
 
+    public List<ContributionResult> getContributionResults() {
+        return Collections.unmodifiableList(contributionResults);
```
sylvlecl, line 66: If we go for an immutable list we will not need the wrapping anymore

sylvlecl (commented):
Naming question + implementation details

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ContributionResult.java
```diff
@@ -0,0 +1,29 @@
...
+    private final String id;
+
+    private final float contribution; //in kA
+
+    public ContributionResult(String id, float contribution) {
```
annetill, line 17: Use double here.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ContributionResult.java
annetill: You have to be more explicit or to increase Javadoc. Here, you don't need the id of the bus but the id of the connectable (load, generator, line, etc.) that has its part in the total amount of three phase current.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FaultResult.java
```diff
@@ -41,4 +49,10 @@ public float getThreePhaseFaultCurrent() {
         return threePhaseFaultCurrent;
     }
 
+    /**
+     * List of contributions of each connectable connected to the equipment to the three phase fault current
+     */
+    public List<ContributionResult> getContributionResults() {
+        return contributionResults;
+    }
```
annetill, line 57: Do you think it could be useful to have a method that gives the contribution given the id of the feeder ?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitConstants.java
```diff
@@ -15,4 +15,6 @@ private ShortCircuitConstants() {
     }
 
     public static final boolean SUBTRANS_STUDY = false;
+
+    public static final boolean WITH_CONTRIBUTIONS = true;
```
annetill, line 19: Maybe better WITH_FEEDER_CONTRIBUTIONS.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitParameters.java
```diff
@@ -67,4 +70,13 @@ public ShortCircuitParameters setSubTransStudy(boolean subTransStudy) {
         return this;
     }
 
+    public boolean isWithContributions() {
```
annetill, line 73: Here too, it would be better to add feeder in the method name.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitConstants.java
geofjamg: are we sure contribution is the right word for that? Looks like badly french translated :-)

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FaultResult.java
colinepiloquet: Yes! It's a good idea, I will add it

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ContributionResult.java
annetill: Seen together: `FeederResult` and instead of contribution, just `FeederThreePhaseFaultCurrent` for the part linked to a feeder.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FaultResult.java
```diff
@@ -17,13 +21,22 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(FaultResult.class);
+
     private final String id;
 
     private final float threePhaseFaultCurrent;
 
-    public FaultResult(String id, float threePhaseFaultCurrent) {
+    private final List<FeederResult> feederResults; //in case of systematic study, optional
+
+    public FaultResult(String id, float threePhaseFaultCurrent, List<FeederResult> feederResults) {
```
annetill, line 32: In general, use double instead of float.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FaultResult.java
```diff
@@ -41,4 +54,26 @@ public float getThreePhaseFaultCurrent() {
...
+        try {
+            double current = 0;
+            for (FeederResult feederResult : feederResults) {
+                if (feederResult.getConnectableId().equals(feederId)) {
+                    current = feederResult.getFeederThreePhaseCurrent();
+                }
+            }
+            return current;
+        } catch (IllegalArgumentException e) {
+            LOGGER.warn("Feeder id does not exist in FeederResults");
```
annetill, line 74: Can you be more explicit in our message with printing the id in question ?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitConstants.java
```diff
@@ -15,4 +15,6 @@ private ShortCircuitConstants() {
     }
 
     public static final boolean SUBTRANS_STUDY = false;
+
+    public static final boolean WITH_FEEDER_CONTRIBUTIONS = true;
```
annetill, line 19: contributions -> results ?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitParameters.java
```diff
@@ -47,8 +49,10 @@ public static ShortCircuitParameters load(PlatformConfig platformConfig) {
         ModuleConfig config = platformConfig.getOptionalModuleConfig("short-circuit-parameters").orElse(null);
         if (config != null) {
             parameters.setSubTransStudy(config.getBooleanProperty("subTransStudy", ShortCircuitConstants.SUBTRANS_STUDY));
-        }
+            // TODO: add a condition on systematic study
+            parameters.setWithFeederContributions(config.getBooleanProperty("withContributions", ShortCircuitConstants.WITH_FEEDER_CONTRIBUTIONS));
```
annetill, line 53: Can you find a more explicit name property, linked to feeder currents ?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitParameters.java
```diff
@@ -67,4 +71,13 @@ public ShortCircuitParameters setSubTransStudy(boolean subTransStudy) {
         return this;
     }
 
+    public boolean isWithFeederContributions() {
```
annetill, line 74: I though we decided to remove the term "contributions" from the API ?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FaultResult.java
geofjamg: An exception is "exceptionnal" and should not be used to handle normal case where a data does not exist

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitParameters.java
geofjamg: feederLevelResults ????

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FaultResult.java
```diff
@@ -37,8 +50,25 @@ public String getId() {
...
 
+    /**
+     * List of contributions to the three phase fault current of each connectable connected to the equipment
+     */
+    public List<FeederResult> getFeederResults() {
+        return feederResults;
+    }
+
+    public double getFeederCurrent(String feederId) {
+        double current = 0;
```
miovd, line 65: Here, if the `id` does not exist, the current is 0. Is it what we want? Wouldn't we rather have `Double.NaN`?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FaultResult.java
```diff
@@ -37,8 +50,25 @@ public String getId() {
...
+        for (FeederResult feederResult : feederResults) {
+            if (feederResult.getConnectableId().equals(feederId)) {
+                current = feederResult.getFeederThreePhaseCurrent();
```
miovd, line 68: For performance purposes, I think you should return the current as soon as you find it (avoiding going through all feederResult). You should have something like this:
```java
for (FeederResult feederResult : feederResults) {
      if (feederResult.getConnectableId().equals(feederId)) {
            return feederResult.getFeederThreePhaseCurrent();
      }
}
return Double.NaN;
```

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FeederResult.java
```diff
@@ -0,0 +1,40 @@
...
+     * The value of the current of the connectable contributing to the total fault current in kA
+     */
+    private final double feederThreePhaseCurrent;
+
+    public FeederResult() {
+        this("", 0);
+    }
+
+    public FeederResult(String connectableId, double feederThreePhaseCurrent) {
+        this.connectableId = connectableId;
```
miovd, line 29: Add `Objects.requireNonNull`?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FeederResult.java
```diff
@@ -0,0 +1,40 @@
...
+        this.feederThreePhaseCurrent = feederThreePhaseCurrent;
```
miovd, line 30: Do we need a check on the current? (e.g. forbidding it to be NaN). Maybe not.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FeederResult.java
```diff
@@ -0,0 +1,40 @@
...
+     */
+    private final String connectableId;
+
+    /**
+     * The value of the current of the connectable contributing to the total fault current in kA
+     */
+    private final double feederThreePhaseCurrent;
+
+    public FeederResult() {
+        this("", 0);
```
miovd, line 25: Just by curiosity, why is it useful? For serialization purposes?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitAnalysisResult.java
```diff
@@ -30,9 +30,18 @@
...
 
-    public ShortCircuitAnalysisResult(List<FaultResult> faultResults, List<LimitViolation> limitViolations) {
+    private List<FeederResult> feederResults;
+
+    public ShortCircuitAnalysisResult(List<FaultResult> faultResults, List<LimitViolation> limitViolations, List<FeederResult> feederResults) {
         this.faultResults.addAll(Objects.requireNonNull(faultResults));
         this.limitViolations.addAll(Objects.requireNonNull(limitViolations));
+        if (feederResults != null) {
+            this.feederResults = List.copyOf(feederResults);
+        }
```
miovd, line 40: The constructor is called when `feederResulrts` is null, so this code is never used. What did you want to do?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitAnalysisResult.java
```diff
@@ -60,4 +69,8 @@ public ShortCircuitAnalysisResult setNetworkMetadata(NetworkMetadata networkMeta
         return this;
     }
 
+    public List<FeederResult> getFeederResults() {
+        return feederResults;
```
miovd, line 73: Good practice for getters of lists: never directly pass the lists to prevent them to be modified without security. Always pass a unmodifiable copy: ```Collections.unmodifiableList(feederResults)```

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FaultResult.java
```diff
@@ -37,8 +50,25 @@ public String getId() {
...
+    public double getThreePhaseFaultCurrent() {
         return threePhaseFaultCurrent;
     }
 
+    /**
+     * List of contributions to the three phase fault current of each connectable connected to the equipment
+     */
+    public List<FeederResult> getFeederResults() {
+        return feederResults;
+    }
```
miovd, line 62: Good practice for getters of lists: never directly pass the lists to prevent them to be modified without security. Always pass a unmodifiable copy: ```Collections.unmodifiableList(feederResults)```

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/converter/FaultResultDeserializer.java
```diff
@@ -37,8 +40,9 @@
     @Override
     public FaultResult deserialize(JsonParser parser, DeserializationContext deserializationContext) throws IOException {
         String id = "";
-        float threePhaseFaultCurrent = 0;
+        double threePhaseFaultCurrent = 0;
```
miovd, line 43: I understand 0 was the previous default value but is there a reason for that?

miovd (changes requested):

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FeederResult.java
colinepiloquet: Yes! The serialization doesn't work without this constructor.

miovd (approved):

PR #1957, opened 2022-01-28
topological nodes set can not be null

marcosmc:

Feature

When obtaining the list of topological nodes of a bus, it is verified that it is not null.

Currently this list cannot be null, so the check is unnecessary.

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
@@ -126,20 +126,12 @@ public final class StateVariablesExport {
             LOG.info(log.get());
             return;
         }
-        Set<CgmesIidmMapping.CgmesTopologicalNode> topologicalNodes = context.getTopologicalNodesByBusViewBus(busId);
-        if (topologicalNodes == null) {
-            return;
-        }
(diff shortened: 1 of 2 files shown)


zamarrenolm (approved):
I think we can merge this PR, but after reviewing it I think we need to improve the export at least in two ways:

* When doing a full export from a pure IIDM Network, the extension CgmesIidmMapping should also be created and filled. It contains mapping data between CGMES and IIDM that should be reused if the same Network is exported again (base voltages, for example). Other mapping information is already been stored as aliases and properties, so it is also reused if re-exporting.
* The mapping between Network buses and TNs should be "updated". Currently, it is either read from the extension or created from scratch. If a user imports a Network from CGMES, then creates new buses on the Network, the mapping would became invalid (would miss mappings for the new buses).

miovd:
> I think we can merge this PR, but after reviewing it I think we need to improve the export at least in two ways:
> 
>     * When doing a full export from a pure IIDM Network, the extension CgmesIidmMapping should also be created and filled. It contains mapping data between CGMES and IIDM that should be reused if the same Network is exported again (base voltages, for example). Other mapping information is already been stored as aliases and properties, so it is also reused if re-exporting.
> 
>     * The mapping between Network buses and TNs should be "updated". Currently, it is either read from the extension or created from scratch. If a user imports a Network from CGMES, then creates new buses on the Network, the mapping would became invalid (would miss mappings for the new buses).

I agree with you. I was already working on some improvements including making CgmesIidmMapping topology invalid when the network topology is modified.

PR #1974, opened 2022-02-14
Bug fix: when removed, network components cannot access the network anymore

### LINKED ISSUE #1960: Network still referenced from a VoltageLevel after removal
Bug

After removing a VoltageLevel from a network, calling `VoltageLevel::getNetwork()` on it still returns the network.

```
    @Test
    public void vlRemove() {
        Network network = Network.create("test", "test");
        network.newVoltageLevel()
                .setTopologyKind(TopologyKind.BUS_BREAKER)
                .setId("bbVL")
                .setName("bbVL_name")
                .setNominalV(200.0)
                .setLowVoltageLimit(100.0)
                .setHighVoltageLimit(200.0)
                .add();
        VoltageLevel voltageLevel = network.getVoltageLevel("bbVL");
        voltageLevel.remove();
        assertNull(voltageLevel.getNetwork());
    }
```

`VoltageLevel::getNetwork()` should return null.

  - PowSyBl Version: ...
  - OS Version: ...

miovd:

Fix #1960 

Bug fix

**Other information**:


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractConnectable.java
```diff
@@ -49,7 +50,7 @@ void addTerminal(TerminalExt terminal) {
 
     @Override
     public NetworkImpl getNetwork() {
-        return networkRef.get();
+        return Optional.ofNullable(networkRef).map(Ref::get).orElse(null);
```
flo-dup, line 53: One detail, isn't it less clear to create an `Optional` instead of doing the classical ternary `return x!=null ? y(x) : null`, or  of writing `if x equals null return null`?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTerminal.java
```diff
@@ -70,11 +73,14 @@ public void setNum(int num) {
...
     @Override
     public double getP() {
-        return p.get(network.get().getVariantIndex());
+        return Optional.ofNullable(network).map(n -> p.get(n.get().getVariantIndex())).orElse(Double.NaN);
     }
 
     @Override
     public Terminal setP(double p) {
+        if (network == null) {
+            throw new PowsyblException("Can not modify removed equipment");
```
flo-dup, line 82: An even smaller detail, we write "cannot" in the other exceptions, but both are ok anyway

flo-dup (approved):

PR #1975, opened 2022-02-15
Add parameters to limit post contingency limit violations result

annetill:

No.

The PR introduces three new parameters in security analysis parameters:
- A threshold to filter the flow violations after contingencies if they are increasing less than a given factor (without unit).
- A delta to filter the low voltage violations after contingencies if the decrease is less than a given value (in kV).
- A delta to filter the high voltage violations after contingencies if the increase is less than a given value (in kV).

**Other information**:


security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisParameters.java
```diff
@@ -61,12 +76,132 @@ public static SecurityAnalysisParameters load(PlatformConfig platformConfig) {
...
+                                      double increasedLowVoltageViolationsDelta, double increasedHighVoltageViolationsDelta) {
+        this.increasedFlowViolationsThreshold = increasedFlowViolationsThreshold;
+        this.increasedLowVoltageViolationsThreshold = increasedLowVoltageViolationsThreshold;
+        this.increasedHighVoltageViolationsThreshold = increasedHighVoltageViolationsThreshold;
+        this.increasedLowVoltageViolationsDelta = increasedLowVoltageViolationsDelta;
+        this.increasedHighVoltageViolationsDelta = increasedHighVoltageViolationsDelta;
+    }
+
+    public SecurityAnalysisParameters() {
+        this(DEFAULT_INCREASED_FLOW_VIOLATIONS_THRESHOLD, DEFAULT_INCREASED_LOW_VOLTAGE_VIOLATIONS_THRESHOLD, DEFAULT_INCREASED_HIGH_VOLTAGE_VIOLATIONS_THRESHOLD,
```
sylvlecl, line 107: No need to define default values here AND in the field definitions.

security-analysis/security-analysis-api/src/test/resources/SecurityAnalysisParameters.json
```diff
@@ -1,5 +1,10 @@
 {
-  "version" : "1.0",
+  "version" : "1.1",
+  "increasedFlowViolationsThreshold" : "0.1",
```
sylvlecl, line 3: We should use plain floats here, not strings

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/SecurityAnalysisParametersDeserializer.java
```diff
@@ -41,12 +41,25 @@ public SecurityAnalysisParameters deserialize(JsonParser parser, Deserialization
                 case "version":
                     parser.nextToken();
                     break;
-
+                case "increasedFlowViolationsThreshold":
+                    parameters.setIncreasedFlowViolationsThreshold(Double.parseDouble(parser.nextTextValue()));
```
sylvlecl, line 45: See comment in JSON file: should be parsed straight as double, not as text and then parsed.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/SecurityAnalysisParametersSerializer.java
```diff
@@ -30,12 +30,15 @@ public void serialize(SecurityAnalysisParameters parameters, JsonGenerator jsonG
         jsonGenerator.writeStartObject();
 
         jsonGenerator.writeStringField("version", SecurityAnalysisParameters.VERSION);
-
+        jsonGenerator.writeStringField("increasedFlowViolationsThreshold", String.valueOf(parameters.getIncreasedFlowViolationsThreshold()));
```
sylvlecl, line 33: Since "loadflow-parameters" is in snake case, we should probably use snake case for all parameters (although different from fields inside the loadflow parameters).

security-analysis/security-analysis-api/src/test/java/com/powsybl/security/SecurityAnalysisParametersTest.java
```diff
@@ -38,6 +46,24 @@ public void testNoExtensions() {
         assertFalse(parameters.getExtensions().contains(new DummyExtension()));
         assertFalse(parameters.getExtensionByName("dummyExtension") instanceof DummyExtension);
         assertFalse(parameters.getExtension(DummyExtension.class) instanceof DummyExtension);
+
+        assertEquals(0.1, parameters.getIncreasedFlowViolationsThreshold(), EPS);
```
sylvlecl, line 50: Better to move it to another unit test or rename this one

security-analysis/security-analysis-api/src/test/resources/SecurityAnalysisParameters.json
```diff
@@ -1,5 +1,10 @@
...
+  "increasedLowVoltageViolationsThreshold" : "0.0",
```
sylvlecl, line 4: Better to use other values than 0 for all params, to ensure it's not the default values

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/SecurityAnalysisParametersSerializer.java
geofjamg: We have not been consistent in existing parameters naming, I don't know which one is better to follow.

security-analysis/security-analysis-api/src/test/resources/SecurityAnalysisParameters.json
geofjamg: Done

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/SecurityAnalysisParametersDeserializer.java
geofjamg: Done

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisParameters.java
geofjamg: Done

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisParameters.java
```diff
@@ -30,16 +30,103 @@
...
+    // VERSION = 1.1 IncreasedViolationsParameters adding.
+    public static final String VERSION = "1.1";
 
     private static final Supplier<ExtensionProviders<ConfigLoader>> SUPPLIER =
-        Suppliers.memoize(() -> ExtensionProviders.createProvider(ConfigLoader.class, "security-analysis-parameters"));
+            Suppliers.memoize(() -> ExtensionProviders.createProvider(ConfigLoader.class, "security-analysis-parameters"));
 
     private LoadFlowParameters loadFlowParameters = new LoadFlowParameters();
 
+    public static class IncreasedViolationsParameters {
```
geofjamg, line 45: I am not sure it is a good idea to add this object as it forbid you to set parameters using method chaining which I personally use very often when setting calculation parameters
```java
var params = new SecurityAnalysisParameters()
  .setIncreaseViolationHighVoltageProportionalThreshold(10)
  .setIncreaseViolationLowVoltageProportionalThreshold(13);
```

Or maybe:
```java
var params = new SecurityAnalysisParameters()
  .getIncreaseViolationParameters().setHighVoltageProportionalThreshold(10)
  .getIncreaseViolationParameters().setLowVoltageProportionalThreshold(13);
```

Not sure it is better.

geofjamg (changes requested):

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisParameters.java
sylvlecl: It adds another instruction but maybe you still can write it this way?
```java
var params = new SecurityAnalysisParameters()
    ....
    .set...();
params.getIncreaseViolationParameters()
    .setHighVoltageProportionalThreshold(10)
    .setLowVoltageProportionalThreshold(13);
```

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/SecurityAnalysisParametersSerializer.java
sylvlecl: I think we should be consistent at least within a class : here loadflow parameters is already in snake case!
Or indeed we could change it, in order to be consistent with the fields of loadflow parameters?

geofjamg: Ok let's change these new params to camel case

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisParameters.java
geofjamg: Ok why not. But in this case we have to create a sub object in the json too

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/SecurityAnalysisParametersSerializer.java
EtienneLt: Done

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisParameters.java
```diff
@@ -28,18 +29,127 @@
...
+        static final double DEFAULT_LOW_VOLTAGE_ABSOLUTE_THRESHOLD = 0.0; // 0.0 kV
+        static final double DEFAULT_HIGH_VOLTAGE_ABSOLUTE_THRESHOLD = 0.0; // 0.0 kV
+
+        private double lowVoltageAbsoluteThreshold = DEFAULT_LOW_VOLTAGE_ABSOLUTE_THRESHOLD;
+        private double lowVoltageProportionalThreshold = DEFAULT_LOW_VOLTAGE_PROPORTIONAL_THRESHOLD;
+        private double highVoltageAbsoluteThreshold = DEFAULT_HIGH_VOLTAGE_ABSOLUTE_THRESHOLD;
+        private double highVoltageProportionalThreshold = DEFAULT_HIGH_VOLTAGE_PROPORTIONAL_THRESHOLD;
+        private double flowProportionalThreshold = DEFAULT_FLOW_PROPORTIONAL_THRESHOLD;
+
+        public IncreasedViolationsParameters(@JsonProperty("increased-low-voltage-violations-absolute-threshold")
```
sylvlecl, line 61: Now we can use shorter names in the JSON, no need anymore for the "increased***" prefix!
Like just "low-voltage-absolute-threshold", etc.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/SecurityAnalysisParametersSerializer.java
```diff
@@ -30,12 +30,21 @@ public void serialize(SecurityAnalysisParameters parameters, JsonGenerator jsonG
...
+        jsonGenerator.writeNumberField("increased-high-voltage-violations-absolute-threshold", parameters.getIncreasedViolationsParameters().getHighVoltageAbsoluteThreshold());
+        jsonGenerator.writeEndObject();
 
         jsonGenerator.writeFieldName("load-flow-parameters");
         JsonLoadFlowParameters.serialize(parameters.getLoadFlowParameters(), jsonGenerator, serializerProvider);
 
         JsonUtil.writeExtensions(parameters, jsonGenerator, serializerProvider, JsonSecurityAnalysisParameters.getExtensionSerializers());
 
         jsonGenerator.writeEndObject();
+        jsonGenerator.writeRaw('\n');
```
sylvlecl, line 48: we should not add this, the exact formatting should be left to more general options (pretty printing etc)

sylvlecl (changes requested):
Last minor changes

security-analysis/security-analysis-api/src/test/resources/SecurityAnalysisParametersV1.json
```diff
@@ -17,4 +17,4 @@
     "countriesToBalance" : [ ],
     "connectedComponentMode" : "MAIN"
   }
-}
\ No newline at end of file
+}
```
geofjamg, line 20: here too

sylvlecl (approved):

geofjamg (approved):

PR #1976, opened 2022-02-15
Remove feeder result from short circuit analysis result

colinepiloquet:

No

Remove the feederResult from ShortCircuitAnalysisResult as it is not useful there


shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitAnalysisResult.java
```diff
@@ -30,16 +30,9 @@
 
     private final List<LimitViolation> limitViolations = new ArrayList<>();
 
-    private List<FeederResult> feederResults = new ArrayList<>();
```
colinepiloquet, line 33: This is useless in this class, as it is already in the FaultResult class.

miovd (approved):

PR #1979, opened 2022-02-17
Fix removing the "last" variant of a network for tap changers

phiedw:
Signed-off-by: EDWARDS Philippe <edwardsphi@gm0winl197.bureau.si.interne>

No

Bug fix

When deleting the network variant with the highest index, the tap changers will lose the wrong number of values for the tap position attribute.

It now loses the expected number of values.

**Other information**:

### DIFF
--- a/action/action-dsl-spi/pom.xml
+++ b/action/action-dsl-spi/pom.xml
@@ -15,7 +15,7 @@
     <parent>
         <groupId>com.powsybl</groupId>
         <artifactId>powsybl-action</artifactId>
-        <version>4.7.0-RC1</version>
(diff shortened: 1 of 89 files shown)


geofjamg (commented):
I suspect same bug in `ShuntCompensatorImpl`

phiedw:
indeed, I'll look for that line of code in all the files to make sure there are no other bugs

geofjamg:
> indeed, I'll look for that line of code in all the files to make sure there are no other bugs

I would say it is ok now. I searched in the delayed validation PR for other `reduceVariantArraySize` changes and there is not.

geofjamg (approved):

PR #1980, opened 2022-02-18
Fix CGMES import with InputStream

colinepiloquet:

No

Bug fix

The import of CGMES files fails when using loadNetwork(String filename, InputStream data)


commons/src/main/java/com/powsybl/commons/datasource/ZipReadOnlyMemDataSource.java
```diff
@@ -0,0 +1,117 @@
...
+
+    @Override
+    public Set<String> listNames(String regex) throws IOException {
+        Pattern p = Pattern.compile(regex);
+        return data.keySet().stream()
+                .filter(name -> p.matcher(name).matches())
+                .collect(Collectors.toSet());
+    }
+
+    private static final class ZipEntryInputStream extends ForwardingInputStream<InputStream> {
```
flo-dup, line 100: you can delete this as you're not using it

commons/src/main/java/com/powsybl/commons/datasource/ZipReadOnlyMemDataSource.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+
+public class ZipReadOnlyMemDataSource implements ReadOnlyDataSource {
```
flo-dup, line 20: Should we add this new class which is almost the same as ReadOnlyMemDataSource? The only difference is `putData`. I think we could have reused ReadOnlyMemDataSource in fact, and put the code you wrote in `putData` in `DataSourceUtil`, like the already existing one for file. Or we could have extended ReadOnlyMemDataSource maybe? what do you think?

flo-dup (changes requested):

commons/src/main/java/com/powsybl/commons/datasource/ZipReadOnlyMemDataSource.java
colinepiloquet: I actually agree, it should be possible to use only ReadOnlyMemDataSource. I will try to add the code in DataSourceUtil.

iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/import_/Importers.java
```diff
@@ -549,8 +550,16 @@ public static Network loadNetwork(String file) {
      * @return                   The loaded network
      */
     public static Network loadNetwork(String filename, InputStream data, ComputationManager computationManager, ImportConfig config, Properties parameters, ImportersLoader loader, Reporter reporter) {
-        ReadOnlyMemDataSource dataSource = new ReadOnlyMemDataSource(DataSourceUtil.getBaseName(filename));
-        dataSource.putData(filename, data);
+        ReadOnlyDataSource dataSource;
+        if (filename.endsWith(".zip")) {
```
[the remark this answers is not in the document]
colinepiloquet: I didn't think about the other kinds of compressed files! But I will add a condition

geofjamg (changes requested):
I really doubt this is the right way to solve the issue. The initial need if I understand well is to read a zip file from an `InputStream`, but not to buffer all the zip data. CGMES files which are the main purpose of this feature could be like 1Go deflatted. Is there another way to solve this issue, without buffering everything?

flo-dup:
I don't really see how we can improve that. Indeed, we start with an InputStream and we need to go from one zip entry to the other, not only onward but also backward. Therefore we need to copy the stream, hence buffering it in a way or another.

Ah, but maybe the solution would be writing a temporary zip file and use `ZipFileDataSource`?

geofjamg:
> I don't really see how we can improve that. Indeed, we start with an InputStream and we need to go from one zip entry to the other, not only onward but also backward. Therefore we need to copy the stream, hence buffering it in a way or another.
> 
> Ah, but maybe the solution would be writing a temporary zip file and use `ZipFileDataSource`?

But we can reset position of an inputstream?

flo-dup:
Yeah it's probably the best way to do, not supporting a zipped inputStream and throwing a clear exception.

flo-dup:
Well, we can reset an InputStream only if there's some buffering done behind. I don't know exactly, maybe it's better than inflating all the files.

If I remember correctly there are quite many InputStreams created when reading a CGMES file, for each file more than once. Like it's done once for all for checking the namespaces for instance.

PR #1982, opened 2022-02-18
Make a set of extensions multi-variant

### LINKED ISSUE #1936: Make more IIDM extensions "variants enabled"

Feature.

Some extensions carry data which "look" variant-bound, but which are actually common to all variants, see for instance `ActivePowerControlImpl`.

The following extensions should be "variants enabled", i.e carry one value for each network variant:
 - `ActivePowerControlImpl`
 - `HvdcAngleDroopActivePowerControlImpl`
 - `HvdcOperatorActivePowerRangeImpl`
 - `CoordinatedReactiveControlImpl`
 - `RemoteReactivePowerControlImpl`
 - `VoltagePerReactivePowerControlImpl` (to be confirmed?)

Being able to use those data in multi-variant processes.

In our actual use case, we need to generate many variant with different HVDC droop values.

CBiasuzzi:
Signed-off-by: Christian Biasuzzi <christian.biasuzzi@soft.it>

Fix #1936

**Other information**:

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/RemoteReactivePowerControlImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/RemoteReactivePowerControlImpl.java
@@ -50,7 +50,7 @@ public class RemoteReactivePowerControlImpl extends AbstractMultiVariantIdentifi
     @Override
     public RemoteReactivePowerControl setEnabled(boolean enabled) {
         this.enabled.set(getVariantIndex(), enabled);
-        return null;
+        return this;
     }
 
     @Override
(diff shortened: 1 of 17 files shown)


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/RemoteReactivePowerControlImpl.java
```diff
@@ -6,40 +6,104 @@
...
  * @author Bertrand Rix <bertrand.rix at artelys.com>
  */
-public class RemoteReactivePowerControlImpl extends AbstractExtension<Generator> implements RemoteReactivePowerControl {
+public class RemoteReactivePowerControlImpl extends AbstractMultiVariantIdentifiableExtension<Generator> implements RemoteReactivePowerControl {
 
-    double targetQ;
+    private TDoubleArrayList targetQ;
 
-    Terminal regulatingTerminal;
+    private final ArrayList<Terminal> regulatingTerminal;
```
miovd, line 25: Might not be linked directly to this PR but should regulatingTerminal be multi-variant in regulating equipment? Because they are not for now @annetill @sylvlecl

miovd (approved):
It is okay for me! Just one more general question

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/RemoteReactivePowerControlImpl.java
sylvlecl: Ah you're right we should remain consistent with the base model, I thought it was multi-variant.
Actually now I think that Anne mentioned yesterday that in CGMES, that info is in the EQ file, so it makes sense to keep it this way.

@CBiasuzzi : could you revert just this one please?

sylvlecl (approved):

PR #1983, opened 2022-02-21
Geographical and ControlArea persistence

marcosmc:

Feature

The original Id of the Geographical and Subgeographical Regions are not preserved in the export to CGMES.

Keep the Id of the Geographical and Subgeographical Regions if it exists, or create a new Id to use for all network exports.


zamarrenolm (approved):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
```diff
@@ -435,6 +450,24 @@ private static void addIidmMappingsEquivalentInjection(Network network) {
...
+                    .setName("Network")
+                    .setEnergyIdentificationCodeEic("Network--1")
+                    .add();
+            CgmesControlArea cgmesControlArea = cgmesControlAreas.getCgmesControlArea(cgmesControlAreaId);
+            for (DanglingLine danglingLine : network.getDanglingLines()) {
+                cgmesControlArea.add(danglingLine.getTerminal());
+            }
+        }
+    }
+
```
miovd, line 470: Could you explain why you're adding Control areas here? I don't think it is needed...

zamarrenolm: The CGMES export was creating a "default" control area if none was defined in the input files, or the Network came from IIDM.
Instead of only writing it in the export module, we also keep it in the CGMES-IIDM mappings, as we are doing for the rest of information produced. The export module will write what is found in the CGMES-IIDM mapping data.

miovd: Ok, thank you for your answer

miovd (approved):

PR #1986, opened 2022-02-22
Add new overload for loadNetwork and improve code in Importers

miovd:

No

Feature : Add some overload of `Importers.loadNetwork` which allows `ReadOnlyDataSource` as parameter
+ some code improvements

No

### DIFF
--- a/iidm/iidm-converter-api/src/test/java/com/powsybl/iidm/import_/ImportersTest.java
+++ b/iidm/iidm-converter-api/src/test/java/com/powsybl/iidm/import_/ImportersTest.java
@@ -9,6 +9,7 @@ package com.powsybl.iidm.import_;
 import com.powsybl.commons.PowsyblException;
 import com.powsybl.commons.TestUtil;
 import com.powsybl.commons.datasource.DataSource;
+import com.powsybl.commons.datasource.ReadOnlyDataSource;
 import com.powsybl.commons.reporter.ReporterModel;
 import com.powsybl.computation.ComputationManager;
 import com.powsybl.iidm.AbstractConvertersTest;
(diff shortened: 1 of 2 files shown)


iidm/iidm-converter-api/src/test/java/com/powsybl/iidm/import_/ImportersTest.java
```diff
@@ -145,8 +146,8 @@ public void setPostProcessor() {
     }
 
     @Test
-    public void importData() {
-        Network network = Importers.importData(loader, TEST_FORMAT, null, null, computationManager, importConfigMock);
+    public void importData() throws IOException {
+        Network network = Importers.loadNetwork((ReadOnlyDataSource) null);
```
geofjamg, line 150: Probably has nothing to do with this PR, but what is the purpose of this test?????
Why are we allowing a null data source and then check that a mock importer, we have succeed to import a network?

geofjamg (approved):

iidm/iidm-converter-api/src/test/java/com/powsybl/iidm/import_/ImportersTest.java
geofjamg: Maybe it would worth a fix but could be done in another PR, as you wish

colinepiloquet (approved):

iidm/iidm-converter-api/src/test/java/com/powsybl/iidm/import_/ImportersTest.java
miovd: I think it was to test if the importer was found in the LOADER and not the importer itself (it tests the mechanisms to find an importer).

geofjamg: > I think it was to test if the importer was found in the LOADER and not the importer itself (it tests the mechanisms to find an importer).

In any case null datasource should throws an exception even in such a test

miovd: I'm writing an issue about it

Gb2real: That means  the test finds problem and fixing the issues right???

Issue #1984, opened 2022-02-21
LF parameters extensions: loading from map + link with LF provider


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature

* **What is the current behavior?**

- We have no link, in the API, between a LF provider and its specific parameters
- parameters cannot be loaded from a plain list of properties (key/value pairs)

It makes impossible the creation of LF-specific parameters in a way which would be independent from the implementation.
For example, in `pypowsybl`, we cannot create LF specific parameters from a python dictionary.

* **What is the expected behavior?**

1. we should provide a link between LF provider and their parameters
2. we should provide a method, in LF parameters extensions (and LF parameters?), to load parameters from any list of key/value pair:

* **What is the motivation / use case for changing the behavior?**

See above, being able to write something like this in pypowsybl:
```python
pp.loadflow.run_ac(network, provider='OpenLoadFlow', provider_params={ 'param1': 'value1', ... }
```

sylvlecl:
**Design notes:**

In the current state, the features of a LF param extension are dispatched among at least 2 services:

1. The one which allows to load it from config (`ConfigLoader`)
2. The one which allows to serialize/deserialize to/from JSON

I think it would make more sense to centralize more those features into the LF provider.
We could then have something like :
```java
interface LoadFlowProvider {
...
    // Defines the JSON serializer/deserializer to use
    ExtensionSerializer<E extends Extension<LoadFlowParameters>> getParametersExtensionSerializer();
    
    // Load specific parameters from any map
    Extension<LoadFlowParameters> loadSpecificParameters(Map<String, String> properties);

    // Load from platform config : will probably share code with the one above
    Extension<LoadFlowParameters> loadSpecificParameters(PlatformConfig config);
}
```

sylvlecl:
**After discussion:**

1. Let's put the JSON serializer in the provider, but with the explicit JSON interface ExtensionJsonSerializer
2. Let's treat that in 2 evolutions:
    - First refactor the provider to include loading from platform config and JSON serializer.
      We need to do it for all computations: loadflow, security analysis, sensitivity analysis
    - Then add to it the loading from map

geofjamg:
Maybe also add Optional ?

```
interface LoadFlowProvider {
...
    // Defines the JSON serializer/deserializer to use
    Optional<ExtensionJsonSerializer<E extends Extension<LoadFlowParameters>>> getParametersExtensionSerializer();
    
[...]

PR #1987, opened 2022-02-23
extension loadflow parameters added with Loadflow provider

EtienneLt:

#1984


loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowProvider.java
```diff
@@ -43,15 +47,22 @@
...
+     * @param parameters         load flow execution parameters
+     * @param reporter           the reporter used for functional logs
      * @return a {@link CompletableFuture} on {@link LoadFlowResult]
      */
     default CompletableFuture<LoadFlowResult> run(Network network, ComputationManager computationManager, String workingVariantId, LoadFlowParameters parameters, Reporter reporter) {
         return run(network, computationManager, workingVariantId, parameters);
     }
 
+    // Defines the JSON serializer/deserializer to use
+    Optional<JsonLoadFlowParameters.ExtensionSerializer<? extends Extension<LoadFlowParameters>>> getParametersExtensionSerializer();
```
geofjamg, line 62: You should use ExtensionJsonSerializer directly. JsonLoadFlowParameters.ExtensionSerializer<? extends Extension<LoadFlowParameters>> should be removed as it is a marker interface now (bad practice) and was here before only as a base interface for service loader (now removed).

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowProvider.java
```diff
@@ -43,15 +47,22 @@
...
+
+    // Load from platform config : will probably share code with the one above
+    <E extends Extension<LoadFlowParameters>> Optional<E> loadSpecificParameters(PlatformConfig config);
+
+    <E extends Extension<LoadFlowParameters>> Optional<Class<E>> getExtensionClass();
```
geofjamg, line 67: The only purpose of this method is to better control the class of the extension. Would it be enough to just rely on extension.getClass() ? Is there a use case for have adding an extension with a different class (base one?) than the extension.getClass() one?

geofjamg (changes requested):

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowProvider.java
sylvlecl: It's not only a marker interface, it's also a specialization (T = LoadFlowParameters) --> I don't think we should remove it!

geofjamg: I never seen in Java, "typedef" like design pattern considered as a good practice.
When you have a complex type like Map<String, List<Foo>> you never create a class to simplify the typing (and Map is probably final to avoid this)

sylvlecl: Ok!

geofjamg (approved):

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowProvider.java
```diff
@@ -43,15 +47,20 @@
...
      */
     default CompletableFuture<LoadFlowResult> run(Network network, ComputationManager computationManager, String workingVariantId, LoadFlowParameters parameters, Reporter reporter) {
         return run(network, computationManager, workingVariantId, parameters);
     }
 
+    // Defines the JSON serializer/deserializer to use
+    Optional<ExtensionJsonSerializer> getParametersExtensionSerializer();
+
+    // Load from platform config : will probably share code with the one above
+    <E extends Extension<LoadFlowParameters>> Optional<E> loadSpecificParameters(PlatformConfig config);
```
sylvlecl, line 65: I think we should be more consistent in the naming (sorry for the spec):
above we talk about "parameters extension", here about "specific parameters".

Maybe we can use the "specific parameters" naming in both places (the fact that it's an extension is more a technical detail):
`getSpecificParametersSerializer`
`loadSpecificParameters`

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/json/JsonLoadFlowParameters.java
```diff
@@ -43,20 +43,15 @@
...
-    public static ExtensionProviders<ExtensionSerializer> getExtensionSerializers() {
-        return SUPPLIER.get();
+    private JsonLoadFlowParameters() {
     }
 
-    private JsonLoadFlowParameters() {
+    public static Map<String, ExtensionJsonSerializer> getExtensionSerializers() {
+        List<LoadFlowProvider> providers = new ServiceLoaderCache<>(LoadFlowProvider.class).getServices();
+        return providers.stream()
+            .filter(loadFlowProvider -> loadFlowProvider.getParametersExtensionSerializer().isPresent())
```
sylvlecl, line 52: I think you can use a `flatMap(loadFlowProvider -> loadFlowProvider.getParametersExtensionSerializer().stream())` for shorter syntax and avoid to use a raw `get()` on the optional later

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowParameters.java
```diff
@@ -79,9 +69,7 @@
     public static final boolean DEFAULT_DC_USE_TRANSFORMER_RATIO_DEFAULT = true;
     public static final Set<Country> DEFAULT_COUNTRIES_TO_BALANCE = EnumSet.noneOf(Country.class);
     public static final ConnectedComponentMode DEFAULT_CONNECTED_COMPONENT_MODE = ConnectedComponentMode.MAIN;
-
-    private static final Supplier<ExtensionProviders<ConfigLoader>> SUPPLIER =
-            Suppliers.memoize(() -> ExtensionProviders.createProvider(ConfigLoader.class, "loadflow-parameters"));
+    List<LoadFlowProvider> providers = new ServiceLoaderCache<>(LoadFlowProvider.class).getServices();
```
sylvlecl, line 72: To be removed (could be static final, but it's better to call it when really needed, in the `loadExtensions`)

sylvlecl (changes requested):
Final naming + implementation details

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowProvider.java
```diff
@@ -43,15 +47,20 @@
...
+     * @param parameters         load flow execution parameters
+     * @param reporter           the reporter used for functional logs
      * @return a {@link CompletableFuture} on {@link LoadFlowResult]
      */
     default CompletableFuture<LoadFlowResult> run(Network network, ComputationManager computationManager, String workingVariantId, LoadFlowParameters parameters, Reporter reporter) {
         return run(network, computationManager, workingVariantId, parameters);
     }
 
+    // Defines the JSON serializer/deserializer to use
+    Optional<ExtensionJsonSerializer> getParametersExtensionSerializer();
```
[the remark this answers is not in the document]
sylvlecl: Added some javadoc in new commits

PR #1991, opened 2022-03-01
Do not allow switches with the same node or bus at both ends.

marqueslanauja:

Feature

Switches with the same node or bus at both ends are allowed.

An exception is thrown when a switch with the same node or bus at both ends is added.

**Other information**:


zamarrenolm (approved):

marqueslanauja:
> compatibility with old XIIDM files

We decided to handle compatibility with old XIIDM files by discarding switches with the same node or bus. This verification is always done independently of the XIIDM version.
Another alternative is to add a new attribute in the SwitchAdder to configure the check, throw an exception (by default) or discard the switch (when importing XIIDM files). We selected the first alternative to avoid adding new attributes in the adder.

miovd:
Hello @marqueslanauja @zamarrenolm I made some modifications to add to your work:
- check that the identifiable is not null before reading sub-elements to prevent NPE
- discard switches with same ends during writing (for example while writing a calculated view) to allow round trip

Do you agree? If yes, it is good for me!

miovd (approved):

marqueslanauja:
> Hello @marqueslanauja @zamarrenolm I made some modifications to add to your work:
> 
> * check that the identifiable is not null before reading sub-elements to prevent NPE
> * discard switches with same ends during writing (for example while writing a calculated view) to allow round trip
> 
> Do you agree? If yes, it is good for me!

It is ok for me. Thanks.

PR #1997, opened 2022-03-02
CGMES export to specific CIM version

zamarrenolm:

The ability to export Network to a specific CIM version when exporting to CGMES format.

Code in `CgmesExport` wrote in the output the CIM headers corresponding to the CIM version stored in the extension `CimCharacteristics`. If the extension was not available (if the Network was not imported from CGMES data), CIM version 16 was used by default.

The user can specify the output CIM version to use using a conversion parameter. A test unit demonstrating its use has been added: `ExportToCimVersionTest:testExportIEEE14Cim14ToCim16`.

**Other information**:
In addition:

- Unique identifiers created during export have been fixed (they can not start with a digit).
- Basename for export is read: first from parameter, then from data source, last from network name or id.


miovd:
Oh, good catch! But is this the fix we want? Don't we want the most recent and stable version (CIM16) each time?

miovd (approved):

colinepiloquet (approved):

zamarrenolm:
> Oh, good catch! But is this the fix we want? Don't we want the most recent and stable version (CIM16) each time?

Yes, this is still the default if you do not come from CGMES data. If you create a Network from scratch or import from a different source, when you export to CGMES CIM 16 will be used by default.

If you have imported from CGMES and you do not configure nothing, by default the same version will be put in the exported files.

Finally, no matter the origin, the user can set the param to the desired CIM version he wants to produce.

PR #2001, opened 2022-03-07
Add fictitious injections

### LINKED ISSUE #1937: Modeling fictitious injections
Feature to model fictitious injections correctly on the network (after a state estimation for example).

Fictitious injections can only be modelled as loads. However, the creation of a network component during calculation or a simulation does not allow reproducibility.

A new component should be implemented to model fictitious injections. It should be created at bus-branch level on electrical buses.
```java
bus.setFictitiousP0(1.0).setFictitiousQ0(1.0);
```
Hence, to create it from a terminal:
```java
terminal.getBusView().getBus().setFictitiousP0(1.0).setFictitiousQ0(1.0);
```
For XML serialization, they should be written
- in node breaker as an attribute of calculated buses:
```xml
            <iidm:nodeBreakerTopology>
                <iidm:busbarSection id="D" name="E" node="0"/>
                <iidm:switch id="F" name="G" kind="DISCONNECTOR" retained="false" open="false" fictitious="true" node1="0" node2="1"/>
                <iidm:switch id="H" name="I" kind="DISCONNECTOR" retained="false" open="false" fictitious="true" node1="0" node2="3"/>
                <iidm:switch id="J" name="K" kind="BREAKER" retained="true" open="false" fictitious="true" node1="1" node2="2"/>
                <iidm:switch id="L" name="M" kind="BREAKER" retained="true" open="false" fictitious="true" node1="3" node2="4"/>
                <iidm:bus v="234.40912" angle="0.0" nodes="0,1,2,3,4" fictitiousP0="1.0" fictitiousQ0="1.0"/>
            </iidm:nodeBreakerTopology>
```
- in bus-breaker and bus-branch as an attribute of buses:
```xml
            <iidm:busBreakerTopology>
                <iidm:bus id="NLOAD" ficititousP0="1.0" fictitiousQ0="1.0"/>
            </iidm:busBreakerTopology>
```
Please note that at bus-breaker level, it is written as attributes on all buses of the same electrical bus.

  - PowSyBl Version: 4.7.0-SNAPSHOT

miovd:

Fix #1937 

### DIFF
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/Networks.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/Networks.java
@@ -20,6 +20,7 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.*;
+import java.util.function.Function;
 import java.util.function.Supplier;
 
 /**
@@ -375,6 +376,33 @@ public final class Networks {
         }
 
         return nodesByBus;
+    }
+
+    public static Set<Integer> getNodes(String busId, VoltageLevel voltageLevel, Function<Terminal, Bus> getBusFromTerminal) {
+        if (voltageLevel.getTopologyKind() != TopologyKind.NODE_BREAKER) {
+            throw new IllegalArgumentException("The voltage level " + voltageLevel.getId() + " is not described in Node/Breaker topology");
+        }
+        Set<Integer> nodes = new TreeSet<>();
+        for (int i : voltageLevel.getNodeBreakerView().getNodes()) {
+            Terminal terminal = voltageLevel.getNodeBreakerView().getTerminal(i);
+            if (terminal != null) {
+                Bus bus = getBusFromTerminal.apply(terminal);
+                if (bus != null && bus.getId().equals(busId)) {
+                    nodes.add(i);
+                }
+            } else {
+                // If there is no terminal for the current node, we try to find one traversing the topology
+                Terminal equivalentTerminal = Networks.getEquivalentTerminal(voltageLevel, i);
+
+                if (equivalentTerminal != null) {
+                    Bus bus = getBusFromTerminal.apply(equivalentTerminal);
+                    if (bus != null && bus.getId().equals(busId)) {
+                        nodes.add(i);
+                    }
+                }
+            }
+        }
+        return nodes;
     }
 
     /**
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/Networks.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/Networks.java
@@ -22,6 +22,7 @@ import java.nio.file.Path;
 import java.util.*;
 import java.util.function.Function;
 import java.util.function.Supplier;
+import java.util.stream.IntStream;
 
 /**
  *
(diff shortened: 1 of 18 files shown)


So-Fras (approved):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevel.java
```diff
@@ -587,6 +590,80 @@ CalculatedBusTopology getCalculatedBusTopology() {
...
+            }
+            return 0.0;
+        }
+
+        @Override
+        public NodeBreakerView setFictitiousP0(int node, double p0) {
+            if (Double.isNaN(p0)) {
+                throw new ValidationException(NodeBreakerVoltageLevel.this, "undefined value cannot be set as fictitious p0");
+            }
+            TDoubleArrayList p0ByVariant = fictitiousP0ByNode.computeIfAbsent(node, n -> {
```
sylvlecl, line 610: Note that it introduces a "race condition" between variants, if called in multithread:
if 2 threads try to set p0, we might (although unlikely) create the list in the 2 threads, and end up with only once of them, so one of the values will be ignored.

Note sure of the best solution:
- should we simply ignore it ?
- should we synchronize the map ?
- should we invert the implementation and use one list of doubles by variant ?
- should we follow implementation for V/angle which puts it in node terminal?
- ...

sylvlecl (commented):
Comments for discussion about implementation details

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/Networks.java
```diff
@@ -377,6 +376,31 @@ public static void printGeneratorsSetpointDiff(Network network, Logger logger) {
         return nodesByBus;
     }
 
+    public static Set<Integer> getNodes(String busId, VoltageLevel voltageLevel, Function<Terminal, Bus> getBusFromTerminal) {
```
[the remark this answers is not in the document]
miovd: I followed what was done in `Map<String, Set<Integer>> getNodesByBus(VoltageLevel)` but that can be changed, indeed!

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevel.java
miovd: Following your review I made the maps synchronized. What do you think?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/Networks.java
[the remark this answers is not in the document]
sylvlecl: Ah yes I did not see this one.

Still it's a little different because in the `getNodesByBus`, it can avoid to go over the list of nodes multiple times if we want to go over the buses (we "cache" the nodes in sets), so it makes more sense to return such an object.

For this new method, it's really just more memory efficient to return an `IntStream`, and if the user of the method needs it, he can still collect it in a set.

But indeed honestly it will not make much difference for the framework as a whole :)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevel.java
sylvlecl: Ok!

So-Fras (approved):

sylvlecl (approved):

So-Fras (approved):

PR #2002, opened 2022-03-07
 Refactor cgmes regulating controls after delayed validation

miovd:

Feature

Regulating controls are discarded if attributes are missing.
Default values are provided for some attributes (regulating terminal, voltage setpoint)
Potential inconsistency between point of regulation and value of setpoint
Serialization: only remote regulating terminals are written.

Always preserve the regulating terminal, do not “localize” if regulating terminal is not found (except for `VscConverterStation` where the regulating terminal is set to the local terminal if `pccTerminal` is not defined).
Do not provide default values. All attributes set to NaN or null if they are not explicitly defined.
Keep regulating controls even if not all required data is available (but disabled).
Enable only valid regulating controls. All attributes must be defined and valid.
Serialization: local and remote regulating terminals are written.

**Other information**:


miovd:
Hi @marqueslanauja 
I closed and reopened your PR because of a minor renaming of branch to follow branch syntax!

annetill:
After a second thought: I don't see the benefit of `useLocalRegulation(true)`, compared to the actual behavior that is lighter for user and for serialization/deserialization.
I understand that it is for the CGMES import:
- For close to real time imports, there is for me no utility at keeping a `null` regulating terminal, especially for VSCs. Localize the regulating terminal and keeping the regulation on is a pragmatic choice, and if the targetV is inconscient, it is the responsibility of the power flow to discard it. You can have such an inconsistency in many equipments in IIDM (T2W, T3W, shunts, etc.).
- For long-term studies, of EQ only import, we have the possibility to import the file with a minimum validation level `EQUIPMENT`. In case of a `null` regulating terminal, it could mean that 1) the regulation is local or 2) the regulating terminal has been *explicitly* set with `setRegulatingTerminal` but the terminal as argument is null. I think that before doing such a breaking change, we just have to think about a way to pass `EQUIPMENT` and warn (or throw) the user when we *explicitly* give the regulating terminal. And then the user will have all the information to fix the file.

zamarrenolm:
> After a second thought: I don't see the benefit of `useLocalRegulation(true)`, compared to the actual behavior that is lighter for user and for serialization/deserialization. I understand that it is for the CGMES import:
> 
> * For close to real time imports, there is for me no utility at keeping a `null` regulating terminal, especially for VSCs. Localize the regulating terminal and keeping the regulation on is a pragmatic choice, and if the targetV is inconscient, it is the responsibility of the power flow to discard it. You can have such an inconsistency in many equipments in IIDM (T2W, T3W, shunts, etc.).
> * For long-term studies, or for EQ only import, we have the possibility to import the file with a minimum validation level `EQUIPMENT`. In case of a `null` regulating terminal, it could mean that 1) the regulation is local or 2) the regulating terminal has been _explicitly_ set with `setRegulatingTerminal` but the terminal as argument is null. I think that before doing such a breaking change, we just have to think about a way to pass the network to `EQUIPMENT` validation level and warn (or throw) the user when the import _explicitly_ gives the regulating terminal. And then the user will have all the information to fix the file.

In fact our main concern when proposing this change was to remove potential inconsistencies when defining regulating controls. Not only when coming from CGMES, but also from any other source. We believe that IIDM should contain a definition of the network as "clean" and "valid" as possible.

As an example of current situation, one of the main sources of "noise" on this aspect comes from the default assignment of the regulating terminal that is made in the IIDM adder for generators:
```
        GeneratorImpl generator
                = new GeneratorImpl(network.getRef(),
                                    id, getName(), isFictitious(), energySource,
                                    minP, maxP,
                                    voltageRegulatorOn, regulatingTerminal != null ? regulatingTerminal : terminal,
                                    targetP, targetQ, targetV,
                                    ratedS);
```
The regulating terminal is set explicitly to the the generator's terminal if it is not given explicitly at this point. Here one major problem, as you point out, is that the `targetV` could be inconsistent with the new, changed, terminal. Relying on the analysis engines to accept/discard the `targetV` does not seem the best option: if voltage levels of the original terminals are very different the `targetV` could be easily identified as "invalid", but if they are similar it is not so clear.

Another argument against this default assignment is that some gateways (not only CGMES) could decide to defer the setting of regulating terminals, waiting until all equipment has been mapped to IIDM and the maximum amount of network "points" (IIDM terminals) is known, to be sure that the regulating points in the source model have been mapped to equivalent points in the IIDM model. The gateways could leave intentionally the generator definition "incomplete" at first, and fill it when all required information is available.

We agree that the IIDM model and the different import gateways should be kept as simple as possible for the user. I think we may need further discussion on the approach.

annetill:
I close this PR because the breaking changes are too important compared to the need, we have to find something lighter.

PR #2003, opened 2022-03-08
[WIP] IIDM adders with default values

annetill:

No.

- In this PR, we change the default values of some attributes such as `B`, `G` for line, dangling lines, tie lines. It also changes the default values for steps of tap changers. For shunt compensators, as `G` is optional by default, the default value of zero is set only if the boolean `addersWithDefaultValues` is set to true. 
- If the boolean `addersWithDefaultValues` is set to true: 
       1) For transformers, we also provide some default values for `ratedU1` and `ratedU2` for two windings transformers, and for each leg `ratedU` for three windings transformers. 
       2) For phase tap changers, `rho` has a default value of `1`.
       3) For loads and batteries, we propose to has default values of `zero` for reactive power. It should be discuss but it could be useful to build network for DC calculations.
       4) For dangling lines, we only provide a default value for `Q0` to zero and in case of a generation, we can have a default value of zero for `targetP`. 
       5) In general, min P and max P or min B and max B are set respectively to - max value and + max value.
       6) In case of an HVDC, and it should be discuss, this PR provides zero as default value for the resistance, and the nominal voltage on rectifier side if the nominal voltage of the line is not set. It could be seen as a huge approximation but it is used only to compute losses due to the cable.
       7) For conversion stations, the default value for loss factor is set to zero. It should be discuss.
       8) Target deadbands will receive a default value of zero. 

I wonder about that the line of code `regulatingTerminal  = regulatingTerminal != null ? regulatingTerminal : terminal` that we put everywhere we have a regulating control could be done only if `addersWithDefaultValues` is set to true.

**Other information**:


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BatteryAdderImpl.java
```diff
@@ -92,6 +92,11 @@ public BatteryImpl add() {
         NetworkImpl network = getNetwork();
         String id = checkAndGetUniqueId();
         TerminalExt terminal = checkAndGetTerminal();
+        if (network.getAddersWithDefaultValues()) {
```
annetill, line 95: Min P or Q and max P or Q could be transferred as a default behavior.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DanglingLineAdderImpl.java
```diff
@@ -65,6 +65,11 @@ public GenerationAdder setTargetV(double targetV) {
         @Override
         public DanglingLineAdder add() {
             NetworkImpl network = getNetwork();
+            if (network.getAddersWithDefaultValues()) {
+                minP = Double.isNaN(minP) ? -Double.MAX_VALUE : minP;
+                maxP = Double.isNaN(maxP) ? Double.MAX_VALUE : maxP;
+                targetP = Double.isNaN(targetP) ? 0.0 : targetP;
```
annetill, line 71: The solution is maybe to make `targetP` optional.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/LoadAdderImpl.java
```diff
@@ -61,6 +61,10 @@ public LoadImpl add() {
         NetworkImpl network = getNetwork();
         String id = checkAndGetUniqueId();
         TerminalExt terminal = checkAndGetTerminal();
+        if (network.getAddersWithDefaultValues()) {
+            q0 = Double.isNaN(q0) ? 0.0 : q0;
+            loadType = loadType == null ? LoadType.UNDEFINED : loadType;
```
annetill, line 66: LoadType useless.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/PhaseTapChangerAdderImpl.java
```diff
@@ -94,6 +94,14 @@ public StepAdder setB(double b) {
 
         @Override
         public PhaseTapChangerAdder endStep() {
+            NetworkImpl network = getNetwork();
+            if (network.getAddersWithDefaultValues()) {
+                r = Double.isNaN(r) ? 0.0 : r;
+                x = Double.isNaN(x) ? 0.0 : x;
+                g = Double.isNaN(g) ? 0.0 : g;
```
annetill, line 101: FIXME: could we transfer all this in the default behavior.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
```diff
@@ -87,6 +87,10 @@ public ShuntCompensatorLinearModelAdder setMaximumSectionCount(int maximumSectio
 
         @Override
         public ShuntCompensatorAdder add() {
+            NetworkImpl network = getNetwork();
+            if (network.getAddersWithDefaultValues()) {
+                gPerSection = Double.isNaN(gPerSection) ? 0.0 : gPerSection;
```
annetill, line 92: Not needed.

geofjamg:
@annetill the problem with this PR is that some of the change are straightforward and could be merged ASAP (like r, x, g, b to 0) and some other should be debated. I would split in 2 parts and merge the easy part.

PR #2013, opened 2022-03-14
Reduce CGMES and Triplestore tests logs by default to ERROR level

miovd:

No

Reduce compilation time by reducing log verbosity in CGMES tests and Triplestore tests


annetill (changes requested):
Can we, in the same PR, try to remove the print of querys that we have in TRACE level?

miovd:
> Can we, in the same PR, try to remove the print of querys that we have in TRACE level?

It is not in trace, it is debug which might be okay by me, do you still want to remove them?

annetill (approved):

PR #2040, opened 2022-03-25
add affected terminals to traverse in TrippingTopologyTraverser to trace which terminals have been traversed

JacquesBorsenberger:
…ace which terminals have been traversed.

**Other information**:


FranckLecuyer (changes requested):

iidm/iidm-modification/src/test/java/com/powsybl/iidm/modification/tripping/BranchTrippingTest.java
```diff
@@ -126,8 +126,10 @@ public void fictitiousSwitchTest() {
 
         Set<Switch> switchesToOpen = new HashSet<>();
         Set<Terminal> terminalsToDisconnect = new HashSet<>();
-        tripping.traverse(network, switchesToOpen, terminalsToDisconnect);
-        assertEquals(switchIds, switchesToOpen.stream().map(Switch::getId).collect(Collectors.toSet()));
+        Set<Terminal> affectedTerminals = new HashSet<>();
+        tripping.traverse(network, switchesToOpen, terminalsToDisconnect, affectedTerminals);
+        assertEquals(affectedTerminals.stream().map(Terminal::getConnectable).map(Connectable::getId).collect(Collectors.toSet()),
```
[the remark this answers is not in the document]
JacquesBorsenberger: should not have been removed

FranckLecuyer (approved):

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/tripping/AbstractInjectionTripping.java
```diff
@@ -26,10 +26,10 @@ protected AbstractInjectionTripping(String id) {
     }
 
     @Override
-    public void traverse(Network network, Set<Switch> switchesToOpen, Set<Terminal> terminalsToDisconnect) {
+    public void traverse(Network network, Set<Switch> switchesToOpen, Set<Terminal> terminalsToDisconnect, Set<Terminal> affectedTerminals) {
```
geofjamg, line 29: "affected" is not the correct word: traversedTerminal?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/tripping/TrippingTopologyTraverser.java
```diff
@@ -41,6 +41,8 @@ public TraverseResult traverse(Terminal terminal, boolean connected) {
                         terminalsToDisconnect.add(terminal);
                     }
                     return TraverseResult.TERMINATE_PATH;
+                } else if (affectedTerminals != null) {
+                    affectedTerminals.add(terminal);
```
geofjamg, line 45: Maybe we should only fill traversed terminal in case of bus/breaker topo. Even if in that case we will have the same list of terminal than in terminalsToDisconnect, but that would more consistent.

geofjamg (changes requested):

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/tripping/AbstractInjectionTripping.java
JacquesBorsenberger: done

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/tripping/TrippingTopologyTraverser.java
JacquesBorsenberger: done

geofjamg (approved):

flo-dup (changes requested):
I don't see why we couldn't have reused `terminalsToDisconnect` instead of adding a new set `traversedTerminals`. And then simply adding it in the `TopologyKind.NODE_BREAKER` case. It seems to lead to the same result as what's done here. You probably already tried that, could you tell me what I'm missing?

geofjamg:
> `terminalsToDisconnect`



> I don't see why we couldn't have reused `terminalsToDisconnect` instead of adding a new set `traversedTerminals`. And then simply adding it in the `TopologyKind.NODE_BREAKER` case. It seems to lead to the same result as what's done here. You probably already tried that, could you tell me what I'm missing?

I think the initial use of terminalsToDisconnect only for bus/breaker topo?

geofjamg:
> > `terminalsToDisconnect`
> 
> > I don't see why we couldn't have reused `terminalsToDisconnect` instead of adding a new set `traversedTerminals`. And then simply adding it in the `TopologyKind.NODE_BREAKER` case. It seems to lead to the same result as what's done here. You probably already tried that, could you tell me what I'm missing?
> 
> I think the initial use of terminalsToDisconnect only for bus/breaker topo?

But maybe you are right we can change its meaning to simplify the API but in that case, it is a serious breaking change.

geofjamg:
But maybe it is not a so serious breaking change because people that we using this utillity class for a bus/breaker topo were not using terminalsToDisconnect, but just switches

JacquesBorsenberger:
actually, after calling traverse, operation are called on each elements of terminalsToDisconnect 
AbstractTripping.java: disconnect
Hades2ADNConverter.java: disconnect then connect 

this may change some behavior, so I'll  need to add guard depending on network type,

flo-dup:
> this may change some behavior, so I'll need to add guard depending on network type,

Thanks for pointing that out... Indeed that would lead to some extra care afterwards which makes it heavier to use.

flo-dup (approved):

PR #2042, opened 2022-03-28
Fix regulating terminal mapping. Map cgmesTerminals associated to switches.

marqueslanauja:

Some regulating terminals in CGMES models may not have an equivalent terminal in the IIDM model. Specifically, terminals located in CGMES at the ends of switches can not be mapped directly to IIDM because switches in IIDM do not have a proper terminal. With this change we introduce a way to find, for these CGMES regulating terminals, an equivalent terminal in the IIDM model, so we can keep the regulating control defined.

Current behavior relies on finding a terminal in the same CGMES topological node that has been mapped to IIDM. This behavior is not right for regulating controls of a flow magnitude, as it may return a point in the network that is not equivalent to the original regulating point. Also, the current approach requires the definition of Topological Nodes to be available (CGMES TP instance file), and this is not always true (bus/branch models, building the Network from EQ only).  

The proposed approach finds equivalent terminals that respect the restrictions for regulation of flow magnitudes (active or reactive power flow) and  node magnitudes (voltages). It relies on connectivity received through EQ files. Only for bus/branch models where TP file is required, and only for voltage control, Topological Nodes are used as a last resort if no equivalent terminal can be found exploring through switch connectivity.

Some configurations may be impossible to map to IIDM terminals:

- If a flow control is specified at the end of a switch and there is more than one IIDM terminals found from that point there is no way to decide how to "split" the control between them. These regulating terminals can not be mapped to IIDM model.
- Some modellers may select a regulating point for voltage that lies at a boundary node. As boundary nodes do not have proper terminals in IIDM, we have decided to avoid mapping these regulation points instead of mapping them to the network side of a branch, because we would be introducing noise related to the impedance of the branch.

**Other information**:


miovd:
@zamarrenolm @marqueslanauja Can you please complete your PR message to explain which cases are fixed by this? Thank you!

miovd (approved):

PR #2044, opened 2022-03-28
Return a boolean for all traverse method (Terminal, VoltageLevel)

mathbagu:

N/A

Feature

There is an inconsistency in the API: some of the traverse methods don't return a boolean.

The inconsistency has been fixed.

I think it's could be a breaking change, specially in the NetworkStore implementation (@geofjamg). Be careful to update the implementation on your side regarding this change.

**Other information**:


geofjamg (changes requested):
I do not see where is the inconsistency, the boolean seems to be only on implementation side. This boolean is an impl detail and is useless for IIDM API users. We don't want to expose it in the API. 
Maybe I missed something but have you seen the boolean on API side?

PR #2050, opened 2022-03-29
Fix: does not crash during CGMES export when a terminal is disconnected

miovd:

No

Fix

An exception is thrown when there is no bus of the bus view at both ends of a switch.

A fictitious topological node is created.

No

**Other information**:


zamarrenolm:
I am not sure of this approach. I think instead of creating fictitious topological nodes we should just write that terminal of the switch is not connected. This information goes in the SSH: `Terminal.connected` should be `false` in this case. There is a note in the corresponding section of the CGMES standard stating that : _For an open retained breaker, both Terminals are not connected_.

miovd:
> I am not sure of this approach. I think instead of creating fictitious topological nodes we should just write that terminal of the switch is not connected. This information goes in the SSH: `Terminal.connected` should be `false` in this case. There is a note in the corresponding section of the CGMES standard stating that : _For an open retained breaker, both Terminals are not connected_.

I agree but I checked in the implementation guide and Topological Node is a mandatory attribute, does it mean we do not write Terminal in TP in this case?

colinepiloquet (approved):

Issue #667, opened 2019-03-04
Use Java9 newIncompleteFuture of CompletableFuture to ease LocalComputationManager canceling


jonenst:
Right now when using LocalComputationManager, the only CompletableFuture that kills the external process and interrupts the thread is the one that was returned by the call to execute. If you cancel new CompletableFuture created by chaining computations, it does not do it.

Using newIncompleteFuture from Java9 would allow the chained CompletableFuture to be of our subclass and have the interrupting behavior.

geofjamg:
Solved by #2051

PR #2051, opened 2022-03-30
Propagate Cancel to CompletableFutureTask (kill external process of localcomputationmanager)

jonenst:

NO

BugFix / Feature

calling cancel(true) on the CompletableFuture returned by the SecurityAnalysis API (probably other APIs too) do not kill the external process (when using an implementation based on LocalComputationManager)

kills the external process.
Calling cancel() multiple times now returns true only for the one time that cause a cancel. (it was already the case in some paths of the code, just not all of them)

No

### DIFF
--- a/computation/src/main/java/com/powsybl/computation/CompletableFutureTask.java
+++ b/computation/src/main/java/com/powsybl/computation/CompletableFutureTask.java
@@ -6,6 +6,7 @@
  */
 package com.powsybl.computation;
 
+import java.util.Objects;
 import java.util.concurrent.*;
 
 /**
@@ -96,10 +97,10 @@ public class CompletableFutureTask<R> extends CompletableFuture<R> implements Ru
      * @author Jon Harper <jon.harper at rte-france.com>
      */
     static class SourceCancelingCompletableFuture<T> extends CompletableFuture<T> {
-        CompletableFuture<?> source;
+        private final CompletableFuture<?> source;
 
         public SourceCancelingCompletableFuture(CompletableFuture<?> source) {
-            this.source = source;
+            this.source = Objects.requireNonNull(source);
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


computation/src/main/java/com/powsybl/computation/CompletableFutureTask.java
```diff
@@ -88,4 +88,33 @@ public boolean cancel(boolean mayInterruptIfRunning) {
...
 
+    /**
+     * A CompletableFuture that remembers the original CompletableFutureTask that created it
+     * and propagates the cancel to it. It will also return the return value from the source
+     * for cancel() if it has to cancel it. This means that a second cancel will return false,
+     * unlike for regular CompletableFutures for which all cancels return true (if actually canceled).
+     * @author Jon Harper <jon.harper at rte-france.com>
+     */
+    static class SourceCancelingCompletableFuture<T> extends CompletableFuture<T> {
+        CompletableFuture<?> source;
```
mathbagu, line 99: This attribute should probably be private and final.

computation/src/main/java/com/powsybl/computation/CompletableFutureTask.java
```diff
@@ -88,4 +88,33 @@ public boolean cancel(boolean mayInterruptIfRunning) {
...
+
+        public SourceCancelingCompletableFuture(CompletableFuture<?> source) {
+            this.source = source;
```
mathbagu, line 102: Ensure that `source` is not null?

mathbagu (approved):

sylvlecl (approved):
Nice finding, this `newIncompleteFuture()` method!

geofjamg:
@jonenst so this PR solved the 3 years old issue https://github.com/powsybl/powsybl-core/issues/667 ?

jonenst:
I completely forgot that I made this issue 3 years ago wow. Switching to java11 finally paid off hehe

PR #2053, opened 2022-03-31
Export without topological nodes mapping

marcosmc:

**Other information**:

### DIFF
--- a/cgmes/cgmes-conformity/src/test/java/com/powsybl/cgmes/conformity/test/CgmesConformity1Catalog.java
+++ b/cgmes/cgmes-conformity/src/test/java/com/powsybl/cgmes/conformity/test/CgmesConformity1Catalog.java
@@ -179,6 +179,20 @@ public final class CgmesConformity1Catalog {
                         "SmallGridTestConfiguration_TP_BD_v3.0.0.xml"));
     }
 
+    public static TestGridModelResources smallBusBranchEqTp() {
+        String base = ENTSOE_CONFORMITY_1
+                + "/SmallGrid/BusBranch/CGMES_v2.4.15_SmallGridTestConfiguration_BaseCase_Complete_v3.0.0/";
+        String baseBoundary = ENTSOE_CONFORMITY_1
+                + "/SmallGrid/BusBranch/CGMES_v2.4.15_SmallGridTestConfiguration_Boundary_v3.0.0/";
+        return new TestGridModelResources(
+                "SmallGrid-BusBranch-EQ-TP",
+                null,
+                new ResourceSet(base, "SmallGridTestConfiguration_BC_EQ_v3.0.0.xml",
+                        "SmallGridTestConfiguration_BC_TP_v3.0.0.xml"),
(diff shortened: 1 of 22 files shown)


zamarrenolm (approved):

miovd:
@zamarrenolm @marcosmc In this case, do we lose topological mapping? (IDs, etc) I thought we agreed to keep the behavior available for some processes that only export SSH and SV files for example

zamarrenolm:
> @zamarrenolm @marcosmc In this case, do we lose topological mapping? (IDs, etc) I thought we agreed to keep the behavior available for some processes that only export SSH and SV files for example

Moved the PR to draft again while we review these use cases.

zamarrenolm:
> > @zamarrenolm @marcosmc In this case, do we lose topological mapping? (IDs, etc) I thought we agreed to keep the behavior available for some processes that only export SSH and SV files for example
> 
> Moved the PR to draft again while we review these use cases.

With @miovd, @annetill we have agreed on completely removing persistence of Topological Node identifiers for node/breaker models. For use cases like the one described (user is interested only in SV export), a new TP must be exported to obtain proper references of SV values. 

At this point, the user must be aware that the TP export must be requested together with the SV export. Maybe we should consider emitting a warning or trying to enforce TP export in these situations ?

miovd:
> > > @zamarrenolm @marcosmc In this case, do we lose topological mapping? (IDs, etc) I thought we agreed to keep the behavior available for some processes that only export SSH and SV files for example
> > 
> > 
> > Moved the PR to draft again while we review these use cases.
> 
> With @miovd, @annetill we have agreed on completely removing persistence of Topological Node identifiers for node/breaker models. For use cases like the one described (user is interested only in SV export), a new TP must be exported to obtain proper references of SV values.
> 
> At this point, the user must be aware that the TP export must be requested together with the SV export. Maybe we should consider emitting a warning or trying to enforce TP export in these situations ?

Yes, a warning would be nice!
I have another question: if we make two successive CGMES exports, are the TN IDs persistent for node-breaker or are they always re-generated?

miovd (approved):

miovd:
Hi @zamarrenolm @marcosmc ! I was just wondering if you have seen my last comment?

> Yes, a warning would be nice!
I have another question: if we make two successive CGMES exports, are the TN IDs persistent for node-breaker or are they always re-generated?

zamarrenolm:
> Hi @zamarrenolm @marcosmc ! I was just wondering if you have seen my last comment?
> 
> > Yes, a warning would be nice!
> > I have another question: if we make two successive CGMES exports, are the TN IDs persistent for node-breaker or are they always re-generated?

Sure. We saw your last comment. I am working on adding the warning.

About successive CGMES exports and TN IDs persistence: we use the IIDM bus-breaker (_calculated_) identifiers as TN identifiers, so they will be the same in successive exports. Also, if you use a naming strategy and you keep the mapping between exports, the final identifiers should be the same, even if they are fixed because they are not valid CGMES.

miovd (approved):

PR #2055, opened 2022-03-31
Powerfactory lines: shunt admittance calculation.

marqueslanauja:

Feature

An exception is thrown if `TypLne` parameters are not defined.
`tline` parameter of TypLne is managed as conductance.
Only the `bline` parameter is accepted as susceptance.
`0.00001` factor is used to convert micro siemens to siemens.

`TypLne` parameters are managed as optional.
Different alternatives are accepted to define conductance: `gline` parameter, conductance defined in terms of the insulation factor `g = tline * b`
Different alternatives are accepted to define susceptance: `bline` parameter, `cline` parameter.
`0.000001` factor is used to convert micro siemens to siemens.

**Other information**:


powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/AbstractConverter.java
```diff
@@ -0,0 +1,71 @@
...
+    TwoNodeRefs checkAndGetTwoNodeRefs(DataObject obj) {
+        List<NodeRef> nodeRefs = importContext.objIdToNode.get(obj.getId());
+        if (nodeRefs == null || nodeRefs.size() != 2) {
+            throw new PowsyblException("Inconsistent number (" + (nodeRefs != null ? nodeRefs.size() : 0)
+                    + ") of connection for '" + obj + "'");
+        }
+        return new TwoNodeRefs(nodeRefs.get(0), nodeRefs.get(1));
+    }
+
+    static class TwoNodeRefs {
```
zamarrenolm, line 54: I think we do not need to create this class. We could keep the method `checkNodes(obj, int connections)` returning directly a `List<NodeRef>`. For 3-winding transformers we would need to extend it anyway

powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/LineConverter.java
```diff
@@ -0,0 +1,115 @@
...
+            this.x = x;
+            this.g1 = g1;
+            this.b1 = b1;
+            this.g2 = g2;
+            this.b2 = b2;
+        }
+
+        static Optional<LineModel> createFromTypLne(DataObject elmLne) {
+            Optional<DataObject> typLne = elmLne.getObjectAttributeValue(TYP_ID).resolve();
+            if (typLne.isPresent()) {
```
zamarrenolm, line 70: use of `isPresent`, `Optional.of` and `empty` can be simplified using `map`

geofjamg (approved):

PR #2062, opened 2022-04-06
Dummy platform config when no PlatformConfigProvider found

flo-dup:

No

Feature

A PlatformConfigProvider is needed in many cases, even when a default behaviour is defined if the module is missing in the config file.

A dummy platform config is created if no PlatformConfigProvider is found. This platform config returns empty for all calls to `getModuleConfig(s)`, and return Optional<Path> to `getConfigDir` to better handle the nullity of the config dir.


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
```diff
@@ -48,14 +49,17 @@ public CgmesImport(PlatformConfig platformConfig, List<CgmesImportPostProcessor>
         this.defaultValueConfig = new ParameterDefaultValueConfig(platformConfig);
         this.postProcessors = Objects.requireNonNull(postProcessors).stream()
                 .collect(Collectors.toMap(CgmesImportPostProcessor::getName, e -> e));
+        Path boundary = platformConfig.getOptionalConfigDir()
+                .map(dir -> dir.resolve(FORMAT).resolve("boundary"))
+                .orElseThrow(() -> new PowsyblException("Cannot resolve boundary files loaction as configuration directory is not defined in platform config"));
```
geofjamg, line 54: Do we want to always throw an exception? In many case we can just embed boundary files with the case and so we don't care about this directory.

geofjamg (commented):
Looks good! Just a question about how to deal with optional config dir and CGMES.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
miovd: I agree, you can simply put `null` here and in `boundary(Properties p)` method:
```java
     String loc = ConversionParameters.readStringParameter(
                        getFormat(),
                        p,
                        boundaryLocationParameter,
                        defaultValueConfig);
        if (loc == null) {
             return null;
        }
        // Check that the Data Source has valid CGMES names
        ReadOnlyDataSource ds = new GenericReadOnlyDataSource(Paths.get(loc));
        if ((new CgmesOnDataSource(ds)).names().isEmpty()) {
            return null;
        }
        return ds;
```

loadflow/loadflow-validation/src/main/java/com/powsybl/loadflow/validation/ValidationConfig.java
```diff
@@ -52,34 +52,21 @@ public static ValidationConfig load() {
...
-            checkMainComponentOnly = config.getBooleanProperty("check-main-component-only", CHECK_MAIN_COMPONENT_ONLY_DEFAULT);
-            noRequirementIfSetpointOutsidePowerBounds = config.getBooleanProperty("no-requirement-if-setpoint-outside-power-bounds", NO_REQUIREMENT_IF_SETPOINT_OUTSIDE_POWERS_BOUNDS);
-        }
+        Optional<ModuleConfig> config = platformConfig.getOptionalModuleConfig("loadflow-validation");
+        double threshold = config.map(c -> c.getOptionalDoubleProperty("threshold").orElse(THRESHOLD_DEFAULT)).orElse(THRESHOLD_DEFAULT);
+        boolean verbose = config.flatMap(c -> c.getOptionalBooleanProperty("verbose")).orElse(VERBOSE_DEFAULT);
+        String loadFlowName = config.flatMap(c -> c.getOptionalStringProperty("load-flow-name")).orElse(null);
+        Class<? extends TableFormatterFactory> tableFormatterFactory = config.flatMap(c -> c.getOptionalClassProperty("table-formatter-factory", TableFormatterFactory.class)).orElse(TABLE_FORMATTER_FACTORY_DEFAULT);
+        double epsilonX = config.map(c -> c.getOptionalDoubleProperty("epsilon-x").orElse(EPSILON_X_DEFAULT)).orElse(EPSILON_X_DEFAULT);
```
flo-dup, line 61: Two `orElse` as unfortunately there's no `Optional::flatMap(OptionalDouble)` :disappointed:

computation-local/src/main/java/com/powsybl/computation/local/LocalComputationConfig.java
```diff
@@ -63,17 +63,13 @@ private static Optional<Path> getTmpDir(ModuleConfig config, String name) {
...
-                           .orElseGet(() -> getTmpDir(config, "tmpDir")
-                                                .orElseGet(() -> getDefaultLocalDir(fileSystem)));
-            availableCore = config.getOptionalIntProperty("available-core")
-                                  .orElseGet(() -> config.getOptionalIntProperty("availableCore")
-                                                         .orElse(DEFAULT_AVAILABLE_CORE));
-        }
+        Optional<ModuleConfig> config = platformConfig.getOptionalModuleConfig(CONFIG_MODULE_NAME);
+        Path localDir = config.flatMap(c -> getTmpDir(c, "tmp-dir").or(() -> getTmpDir(c, "tmpDir")))
+                .orElse(getDefaultLocalDir(fileSystem));
+        int availableCore = config.map(c -> c.getOptionalIntProperty("available-core")
```
flo-dup, line 69: and, worse, three `orElse`... as unfortunately there's no `OptionalInt::or(OptionalInt)` :disappointed:

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/tools/SecurityAnalysisTool.java
```diff
@@ -217,9 +217,11 @@ static SecurityAnalysisInput buildPreprocessedInput(SecurityAnalysisExecutionInp
     }
 
     private static SecurityAnalysisExecutionBuilder createBuilder(PlatformConfig platformConfig) {
+        String providerName = platformConfig.getOptionalModuleConfig(MODULE_CONFIG_NAME_PROPERTY)
+                .flatMap(c -> c.getOptionalStringProperty(DEFAULT_SERVICE_IMPL_NAME_PROPERTY))
+                .orElse(null);
         return new SecurityAnalysisExecutionBuilder(() -> ExternalSecurityAnalysisConfig.load(platformConfig),
-            platformConfig.getModuleConfig(MODULE_CONFIG_NAME_PROPERTY).getStringProperty(DEFAULT_SERVICE_IMPL_NAME_PROPERTY),
```
flo-dup, line 221: I changed the behaviour here, as it was throwing an exception before if the provider name was not specified.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/execution/SecurityAnalysisExecutionBuilder.java
```diff
@@ -44,7 +44,7 @@ public SecurityAnalysisExecutionBuilder(Supplier<ExternalSecurityAnalysisConfig>
                                             String providerName,
                                             SecurityAnalysisInputBuildStrategy inputBuildStrategy) {
         this.externalConfig = Objects.requireNonNull(externalConfig);
-        this.providerName = Objects.requireNonNull(providerName);
```
flo-dup, line 47: And then I needed to change that to avoid crashing here, and to make the constructor consistent with its javadoc (which says that null is allowed)

flo-dup (commented):
As a result of the last commits cleanup, `PlatformConfig::moduleExists` and `PlatformConfig::getModuleConfig` are not used anymore, except in the two `GroovyDslContingenciesProviderFactory` (to get the path of the dsl file). I think we should delete/deprecate them to force the use of `getOptionalModulConfig` instead.

entsoe-util/src/test/java/com/powsybl/entsoe/util/BoundaryPointXlsParserTest.java
```diff
@@ -63,4 +70,21 @@ public void test() throws IOException {
...
     }
+
+    @Test
+    public void testMissingBoundaryPointFile() {
+        BoundaryPointXlsParser parser = new BoundaryPointXlsParser();
+
+        PlatformConfig mockedPc = Mockito.mock(PlatformConfig.class);
+        Mockito.when(mockedPc.getConfigDir()).thenReturn(Optional.empty());
+
+        try (MockedStatic<PlatformConfig> staticMockedPc = Mockito.mockStatic(PlatformConfig.class)) {
```
geofjamg, line 81: Why do you need this static mock ? There is already a way to mock platform config for test.
Is unit tests are still thread safe using static mocks?

flo-dup: It was only for coverage reason. The `Mockito.mockStatic` is only for current thread:

> Creates a thread-local mock controller for all static methods of the given class or interface.


I'll remove it in this test, as the corresponding coverage is not worth it. It was maybe more interesting in `EmptyPlatformConfigTest`, but the test is a bit twisted after all: it is mocking `ServiceProvider::load` and `ServiceProvider::iterator` to test `PlatformConfig::defaultConfig`.

geofjamg (approved):

PR #2065, opened 2022-04-07
Ensure IDs of calculated buses are unique in IIDM networks

miovd:

No

Bug fix: IDs of calculated buses can be identical to IDs of existing equipment, creating issues on some cases.


geofjamg:
@miovd in which case we cannot have the unicity with actual code?

miovd:
> @miovd in which case we cannot have the unicity with actual code?

In DIE import, when busbar sections have the same IDs as bus of the bus view... It can create issues during CGMES export.

flo-dup (approved):

colinepiloquet (approved):

PR #2067, opened 2022-04-08
All CGMES models after version 3 can be considered node/breaker

zamarrenolm:

CGMES 3 EQ profile enforces use of `ConnectivityNodes` even for bus/branch models, so it is safe to assume we can process any CGMES 3 or later using `ConnectivityNodes` instead of relying on `TopologicalNodes`. If model is bus/branch, internal connections between the equipment and the busbar sections associated to connectivity nodes will be created.

**Other information**:

### DIFF
--- a/cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesNamespace.java
+++ b/cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesNamespace.java
@@ -45,6 +45,7 @@ public final class CgmesNamespace {
     private static final String CIM_16_SV_PROFILE = "http://entsoe.eu/CIM/StateVariables/4/1";
     private static final String CIM_16_SSH_PROFILE = "http://entsoe.eu/CIM/SteadyStateHypothesis/1/1";
 
+    private static final String CGMES_EQ_3_OR_GREATER_PREFIX = "http://iec.ch/TC57/ns/CIM/CoreEquipment-EU/";
     private static final String CIM_100_EQ_PROFILE = "http://iec.ch/TC57/ns/CIM/CoreEquipment-EU/3.0";
     private static final String CIM_100_EQ_OPERATION_PROFILE = "http://iec.ch/TC57/ns/CIM/Operation-EU/3.0";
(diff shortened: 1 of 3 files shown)


cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesNamespace.java
```diff
@@ -45,6 +45,7 @@ private CgmesNamespace() {
     private static final String CIM_16_SV_PROFILE = "http://entsoe.eu/CIM/StateVariables/4/1";
     private static final String CIM_16_SSH_PROFILE = "http://entsoe.eu/CIM/SteadyStateHypothesis/1/1";
 
+    private static final String CGMES_EQ_3_OR_GREATER_PREFIX = "http://iec.ch/TC57/ns/CIM/CoreEquipment-EU/";
```
miovd, line 48: Is this used? Does it exist?

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesNamespace.java
```diff
@@ -79,4 +80,8 @@ public static String getProfile(int cimVersion, String profile) {
         }
         throw new AssertionError("Unsupported CIM version " + cimVersion);
     }
+
+    public static boolean isEqCgmes3OrGreater(String profile) {
+        return profile.startsWith(CGMES_EQ_3_OR_GREATER_PREFIX) && profile.compareTo(CIM_100_EQ_PROFILE) >= 0;
+    }
```
miovd, line 86: I really don't like this kind of really specific method in a final class... If this is only used once, you can move this method in CgmesModelTripleStore.

miovd (approved):

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesNamespace.java
zamarrenolm: yes, to check if an EQ file is version 3 or later (when checking if all loaded EQs are CGMES 3), `CgmesModelTripleStore::allEqCgmes3OrGreater`

PR #2091, opened 2022-04-25
Use mRID CGMES instead of rdf:ID for IIDM identifiers

### LINKED ISSUE #2064: Use mRID CGMES instead of rdf:ID for IIDM identifiers
Propose a change in behavior

When importing CGMES data, IIDM identifiers are built using `rdf:ID` attribute of CIMXML input files.

To use as object identifier, it seems correct to use the `mRID` of the object (Master Resource Identifier), instead of `rdf:ID`. CIM standard for CIMXML serialization states that `mRID` are mapped to `rdf:ID`, that are URNs. To bypass the restriction of URNs not being allowed to start with a digit, a "`_`" prefix is added to all `mRID` when mapping them to `rdf:ID`s.

This was already planned and considered in class `PropertyBag` of `triple-store` module. When obtaining the value of an attribute corresponding to an identifier, leading underscores could be removed. But this behavior is not active by default.

Just changing the default value of `PropertyBag.removeInitialUnderscoreForIdentifiers` to true would make the change ready, but many unit tests have to be reviewed and fixed, when they are looking for elements with hard-coded identifiers starting with "`_`".

The CIM-CGMES export is already dealing with the underscores in related #2059.

IIDM identifiers will not have the leading character "`_`".

Alignment with standars.

  - PowSyBl Version: 4.9.0-SNAPSHOT
  - OS Version: all

CIMXML serialization described in IEC 61970-552 (https://webstore.iec.ch/publication/25939).

zamarrenolm:

Closes #2064


zamarrenolm:
Rebased on top of branch 2047_cgmes3_improvements that provides proper export of RDF identifiers from mRID identifiers in memory.

miovd (approved):

PR #2098, opened 2022-05-03
PowerFactory : select nodeRef in concordance with the busIndexIn parameter

marqueslanauja:

Feature

The busIndexIn parameter is not used. NodeRefs are selected based on the line position of the case file.
In some cases the nodeRefs of the equipment are switched.

NodeRefs are selected in concordance with the busIndexIn parameter and associated properly to the equipment.

**Other information**:


powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/PowerFactoryImporter.java
```diff
@@ -376,6 +380,7 @@ private void createSwitch(ImportContext importContext, VoltageLevel vl, MutableI
             throw new PowsyblException("Multiple staSwitch not supported");
         }
         DataObject staSwitch = staSwitches.isEmpty() ? null : staSwitches.get(0);
+        int busIndexIn = staCubic.getIntAttributeValue("obj_bus");
```
geofjamg, line 383: Are we sure obj_bus is always defined?

marqueslanauja: No, I am not sure, but I have the same problem with almost all the parameters. I have decided to manage a parameter as optional only when  I have a case where this parameter is not defined. By the moment  the `obj_bus` parameter is defined in all the cases that I have.

geofjamg: Ok fine. Yes this is a major difficulty with such a loose model. But maybe worst it is very difficult to known from which version of power Factory an attribute has been added or replace another one and so on to state clearly which version we are supporting

Issue #2092, opened 2022-04-26
[Export] NPE when exporting new extensions to an older XIIDM version 


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Bug

* **What is the current behavior?**

The following code throws an NPE:
```java
Network network = IeeeCdfNetworkFactory.create9();
Properties params = new Properties();
params.setProperty(XMLExporter.VERSION, IidmXmlVersion.V_1_2.toString("."));
Exporters.export("XIIDM", network, params, Paths.get("ieee9.xiidm"));
```

This is because the `SlackTerminal` extension did not exist in XIIDM v1.2.

Here is the stack:
```
java.lang.NullPointerException
	at com.powsybl.iidm.xml.extensions.AbstractVersionableNetworkExtensionXmlSerializer.getVersion(AbstractVersionableNetworkExtensionXmlSerializer.java:82)
	at com.powsybl.iidm.xml.NetworkXml.lambda$getNamespaceUri$3(NetworkXml.java:208)
	at java.base/java.util.Optional.orElseGet(Optional.java:369)
	at com.powsybl.iidm.xml.NetworkXml.getNamespaceUri(NetworkXml.java:208)
	at com.powsybl.iidm.xml.NetworkXml.getNamespaceUri(NetworkXml.java:197)
	at com.powsybl.iidm.xml.NetworkXml.writeExtensionNamespaces(NetworkXml.java:137)
	at com.powsybl.iidm.xml.NetworkXml.initializeWriter(NetworkXml.java:253)
	at com.powsybl.iidm.xml.NetworkXml.write(NetworkXml.java:338)
```

* **What is the expected behavior?**

The export should either throw an explicit exception, so that the user can for example remove the extension, or skip the extension export.
The choice between the 2 behaviours could be an export option.

* **What is the motivation / use case for changing the behavior?**

The situation occurred when using powsybl-dynaflow, which tries to export to XIIDM v1.2.

PR #2099, opened 2022-05-03
fix the export of new extensions with an older version of iidm

EtienneLt:

#2092


iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/export/ExportOptions.java
```diff
@@ -42,6 +42,8 @@ public enum IidmVersionIncompatibilityBehavior {
 
     private boolean throwExceptionIfExtensionNotFound = false;
 
+    private boolean throwExceptionIfExtensionVersionNotFound = false;
```
miovd, line 45: I'm not sure it is necessary to add another parameter, you can only use `throwExceptionIfExtensionNotFound` (same remark in `XMLExporter`)

EtienneLt: I thought it is a different case and @sylvlecl agreed. But I can change it.

miovd: I think we can consider a unique case as I don't really see a use-case where we need to throw an exception for one and ignore the other (and vice versa).

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/NetworkXml.java
```diff
@@ -157,18 +173,31 @@ private static void writeExtension(Extension<? extends Identifiable<?>> extensio
...
         }
-
-        String namespaceUri = getNamespaceUri(extensionXmlSerializer, context.getOptions(), context.getVersion());
-
-        if (extensionXmlSerializer.hasSubElements()) {
-            writer.writeStartElement(namespaceUri, extension.getName());
-        } else {
-            writer.writeEmptyElement(namespaceUri, extension.getName());
+        boolean versionExist = true;
+        if (extensionXmlSerializer instanceof AbstractVersionableNetworkExtensionXmlSerializer) {
```
sylvlecl, line 177: It's a little beyond the scope of that issue, but I think instead of the abstract class we should have an sub-interface for XIIDM extensions serializers.

The `ExtensionXmlSerializer` interface is too generic for the XIIDM export, because it does not provide the mapping between XIIDM version and extension version, which is necessary.

It would simplify the code in here.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/extensions/AbstractVersionableNetworkExtensionXmlSerializer.java
```diff
@@ -75,10 +75,14 @@ public String getNamespaceUri(String extensionVersion) {
...
-        return getVersion(IidmXmlConstants.CURRENT_IIDM_XML_VERSION);
+        return getLastVersion(IidmXmlConstants.CURRENT_IIDM_XML_VERSION);
     }
 
-    public String getVersion(IidmXmlVersion networkVersion) {
+    public boolean versionExists(IidmXmlVersion networkVersion) {
+        return extensionVersions.containsKey(networkVersion);
+    }
+
+    public String getLastVersion(IidmXmlVersion networkVersion) {
```
sylvlecl, line 85: Not sure we should rename it because it's public, so strictly speaking it will be a breaking change.
Maybe it's better to just add some javadoc to better explain what it does?

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/NetworkXml.java
sylvlecl: Actually there is a quite complex logic in that `writeExtension` method, in particular for version handling, that we apply to each of the extension object we are serializing:
it would be better to execute it only once for each extension class we want to serialize!
For example by applying a preprocessing to each serializer to instantiate an object where version is well defined, and re-use it afterwards.

iidm/iidm-xml-converter/src/test/java/com/powsybl/iidm/xml/XMLExporterTest.java
```diff
@@ -48,4 +55,29 @@ public void paramsTest() {
...
+        String busId = "NHV2";
+        Bus bus = vl.getBusBreakerView().getBus(busId);
+        assertNotNull(bus);
+        Terminal t = bus.getConnectedTerminals().iterator().next();
+        assertNotNull(t);
+        vl.newExtension(SlackTerminalAdder.class).withTerminal(t).add();
+
+        Properties params = new Properties();
+        params.setProperty(XMLExporter.VERSION, IidmXmlVersion.V_1_2.toString("."));
+        params.setProperty(XMLExporter.THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND, "true");
```
sylvlecl, line 75: We should also test the case where throw = false

sylvlecl (changes requested):
- The behaviour is OK I think (except the new parameter to delete).
 - A unit test to add to check the behaviour when throwing is disabled
 - Some design thoughts outside the scope of this issue

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/NetworkXml.java
```diff
@@ -234,6 +231,23 @@ private static void writeExtensions(Network n, NetworkXmlWriterContext context,
         }
     }
 
+    private static boolean canTheExtensionBeWritten(ExtensionXmlSerializer extensionXmlSerializer, IidmXmlVersion version, ExportOptions options) {
```
sylvlecl, line 234: Is it a star wars reference ? :D

iidm/iidm-xml-converter/src/test/java/com/powsybl/iidm/xml/XMLExporterTest.java
```diff
@@ -48,4 +54,48 @@ public void paramsTest() {
...
+        Path fakePath = Paths.get("network.xiidm");
```
sylvlecl, line 75: In tests we should use jimfs paths, not the real filesystem.

iidm/iidm-xml-converter/src/test/java/com/powsybl/iidm/xml/XMLExporterTest.java
```diff
@@ -48,4 +54,48 @@ public void paramsTest() {
...
+        assertThrows(PowsyblException.class, () -> Exporters.export("XIIDM", network, params, fakePath));
+        try {
+            Exporters.export("XIIDM", network, params, Paths.get("ieee9.xiidm"));
```
sylvlecl, line 78: In tests we should use jimfs paths, not the real filesystem.

sylvlecl (changes requested):
Ok except few fixes in the tests

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/NetworkXml.java
EtienneLt: no ?

PR #2102, opened 2022-05-04
CGMES export : phase tap changers exported as tabular if EQ is exported

zamarrenolm:

Bug fix.

EQ and SSH export of phase tap changers is not consistent.
CGMES SSH export of phase tap changers is always writing data using the original class name of the tap changer (`PhaseTapChangerLinear`, `PhaseTapChangerSymmetrical` or `PhaseTapChangerAsymmetrical`), but the EQ export always writes the phase tap changer as `Tabular`. 
The SSH export is correct if only SSH instance file is produced as output (written SSH will be assembled with original EQ). But it is wrong if also EQ is produced in the export.

EQ and SSH exports are now consistent. 
When imported from CGMES, all phase tap changers are converted to tabular in IIDM. The EQ export sets the type of tap changer to `Tabular`. SSH is exported after EQ and will use always the correct type.


miovd (changes requested):
I don't agree with this change... What if only SSH and SV files are exported? PhaseTapChanger type in original EQ and new SSH will not be consistent...

zamarrenolm:
> I don't agree with this change... What if only SSH and SV files are exported? PhaseTapChanger type in original EQ and new SSH will not be consistent...

Yes, you are right. But if we also export the EQ then the SSH will be inconsistent. I have to consider both situations.

miovd:
> > I don't agree with this change... What if only SSH and SV files are exported? PhaseTapChanger type in original EQ and new SSH will not be consistent...
> 
> Yes, you are right. But if we also export the EQ then the SSH will be inconsistent. I have to consider both situations.

Yes, I think what we can do is keep the type during import and during export, reset the type while writing EQ: if EQ is not written, it will not be reset.

zamarrenolm:
> > > I don't agree with this change... What if only SSH and SV files are exported? PhaseTapChanger type in original EQ and new SSH will not be consistent...
> > 
> > 
> > Yes, you are right. But if we also export the EQ then the SSH will be inconsistent. I have to consider both situations.
> 
> Yes, I think what we can do is keep the type during import and during export, reset the type while writing EQ: if EQ is not written, it will not be reset.

I've implemented your proposed solution and added a specific unit test that verifies it.

miovd (approved):

PR #2105, opened 2022-05-05
CGMES export EQ : tap changer regulating controls

zamarrenolm:

Export definition of tap changer regulating controls in EQ instance file.

Regulating controls of tap changers were not exported to EQ file. SSH references to these objects were broken if EQ, SSH were exported together.

Regulating controls of tap changers are exported to EQ file. SSH references now point to proper EQ objects.


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportUtil.java
```diff
@@ -174,7 +177,8 @@ public static String loadClassName(LoadDetail loadDetail) {
         return "EnergyConsumer";
     }
 
-    public static int getTerminalSide(Terminal t, Connectable<?> c) {
```
miovd, line 177: Can you keep this method as deprecated?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
```diff
@@ -359,21 +358,51 @@ private static void writeThreeWindingsTransformerEnd(ThreeWindingsTransformer tw
...
+    }
+
+    private static String getPhaseTapChangerRegulationMode(PhaseTapChanger ptc) {
+        switch (ptc.getRegulationMode()) {
+            case CURRENT_LIMITER:
+                return "currentFlow";
+            case ACTIVE_POWER_CONTROL:
+                return "activePower";
+        }
+        return null;
```
miovd, line 403: I don't think this use case is taken into account later? (can result in NPE)

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
```diff
@@ -359,21 +358,51 @@ private static void writeThreeWindingsTransformerEnd(ThreeWindingsTransformer tw
...
+    private static <C extends Connectable<C>> String getTapChangerControlId(C eq, String tcId) {
+        CgmesTapChangers<C> cgmesTcs = eq.getExtension(CgmesTapChangers.class);
+        CgmesTapChanger cgmesTc = null;
+        if (cgmesTcs != null) {
+            cgmesTc = cgmesTcs.getTapChanger(tcId);
+            if (cgmesTc != null) {
+                return cgmesTc.getControlId();
+            }
+        }
+        return null;
```
miovd, line 393: I don't think this use case is taken into account later? (can result in NPE)

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportUtil.java
zamarrenolm: fixed. Method kept as deprecated and for removal

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
zamarrenolm: Yes, it is taken into account. `TapChangerEq.writeRatio` and `writePhase` only write the reference to the regulating control if the identifier is not null. And the regulating control definition itself is only written in `EquipmentExport.writeRatioTapChanger` and `writePhaseTapChanger` if the regulating control id is not null. 

But I will write the return value of this function as an `Optional` so it is explicit that it may or may not be present.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/elements/TapChangerEq.java
```diff
@@ -65,14 +72,25 @@ public static void writePhaseTablePoint(String id, String phaseTapChangerTableId
...
+        if (regulatingControlId.isPresent()) {
+            CgmesExportUtil.writeReference("TapChanger.TapChangerControl", regulatingControlId.get(), cimNamespace, writer);
+        }
+        writer.writeEndElement();
+    }
+
+    public static void writeControl(String id, String name, String mode, String terminalId, String cimNamespace, XMLStreamWriter writer) throws XMLStreamException {
+        CgmesExportUtil.writeStartIdName("TapChangerControl", id, name, cimNamespace, writer);
+        writer.writeEmptyElement(cimNamespace, "RegulatingControl.mode");
+        writer.writeAttribute(RDF_NAMESPACE, CgmesNames.RESOURCE, String.format("%s%s.%s", cimNamespace, "RegulatingControlModeKind", mode));
```
miovd, line 92: Here, `mode` is always written but it can be null

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
zamarrenolm: I've decided to ignore the regulating control in the output if it is `FIXED_TAP` in IIDM.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/elements/TapChangerEq.java
zamarrenolm: `only` valid regulating modes will be written. when a tap changer has IIDM regulation mode FIXED_TAP it won't be exported as a regulating control to CGMES.

zamarrenolm:
I will solve the conflict ...

miovd (approved):

PR #2107, opened 2022-05-06
CGMES EQ export : consider 2 and 3-winding transformer tap changers

marcosmc:

Bug fix

EQ file export fails when exporting a ThreeWindingTransformer that has a TapChanger on leg 3.
Additionally, the right tap changers for each end are exported.
Special treatment for 2-winding transformers is required, as tap changers at end 2 are always moved to IIDM end 1 during import.
Also fixed missing identifiers when both phase and ratio tap changers were present in the same end.


zamarrenolm:
Adjusted aliases for exported tap changers of two-winding transformers following the logic:

IIDM model always has tap changers (ratio and/or phase) at end 1, and only at end 1.
We have to adjust the aliases for (potential) original tap changers coming from end 1, end 2.
Potential tc2 is always converted to a tc at end 1.
If both tc1 and tc2 were present, tc2 was combined during import (fixed at current step) with tc1. Steps from tc1 were kept.
If we only had tc2, it mas moved to end 1. 

When we had only tc2, the alias for tc1 if we do EQ export should contain the identifier of original tc2.
In the rest of situations, we keep the same id under alias for tc1.

miovd (approved):

PR #2108, opened 2022-05-09
Tieflow may not be a dangling line

marcosmc:

Bug fix

Boundary tie flows of control areas are always expected at terminals of dangling lines.
A tie flow for a control area may be specified at the terminal of a half line of a merged line. Not all control area boundaries correspond to terminals of dangling lines.

Boundary tie flows of control areas that correspond to boundary terminals of dangling lines and terminals of half lines are exported properly

**Other information**:
Observed in real world cases where a boundary point is inside a MAS.


zamarrenolm (approved):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
```diff
@@ -750,23 +749,38 @@ private static void writeAcdcConverterDCTerminal(String id, String conductingEqu
...
+        if (c instanceof DanglingLine) {
+            terminalId = c.getAliasFromType(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + "Terminal_Boundary").orElseThrow(PowsyblException::new);
+        } else {
+            int sequenceNumber = 0;
+            if (boundary.getSide().equals(Branch.Side.ONE)) {
+                sequenceNumber = 1;
+            } else if (boundary.getSide().equals(Branch.Side.TWO)) {
+                sequenceNumber = 2;
             }
+            terminalId = c.getAliasFromType(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.TERMINAL + sequenceNumber).orElseThrow(PowsyblException::new);
```
miovd, line 781: In case of a tie line, are you sure, it should not be terminals at boundaries rather than on network side?

zamarrenolm: Yes, the terminal is located at the boundary node in CGMES, but because both halves of the tie line lie inside the IIDM network, none of them is created as a dangling line, and the identifier of the corresponding terminal can not be obtained accessing the alias `"Terminal_Boundary"`, that only is created for dangling lines. So in this case we have to select the proper alias type.

miovd: Yes, but we create alias of boundaries for tie lines in `ACLineSegmentConversion.convertBoundaryLines` so maybe we should retrieve that?
```java
mline.addAlias(boundaryLine1.getBoundaryTerminalId(), Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + "HALF1." + CgmesNames.TERMINAL + "_Boundary");
...
mline.addAlias(boundaryLine2.getBoundaryTerminalId(), Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + "HALF2." + CgmesNames.TERMINAL + "_Boundary");
```

zamarrenolm: Oh, I see ... when the Boundary object is created the side does not refer to one of the half line sides, but is the identification of the half inside the tie line:
```java
..
} else if (identifiable instanceof TieLine) {
    side = networkContext.getReader().getAttributeValue(null, "side");
    TieLine tl = (TieLine) identifiable;
    cgmesControlArea.add(tl.getHalf(Branch.Side.valueOf(side)).getBoundary());

```
the object added as a tie flow to the control area is always the "boundary" side of one of the halves.

So you are right, we should not use the "side" at this point, but always obtain the boundary terminal from the right half line in the tie line ...

zamarrenolm: Additionally ...

The object `Boundary` obtained for that half, that is what is stored in the `CgmesControlArea` extension, has a side. I have checked that it stores the half "number" with respect to its `TieLine` (not the location of the boundary in the half), so it is correct to build the `aliasType` using it as you propose. 

Thanks for the detailed review!

zamarrenolm: There is an issue here:

The half lines of the tie lines present in the Network are not exported to the EQ. It is the tie line itself that is exported as a regular AC Line Segment. So, in the exported EQ file we do not have the terminals for the boundary points, only the terminals for the half lines network sides.

I will try to export the half lines separately, and their terminals, instead of the tie lines.

zamarrenolm: Proposal: close this pull request ignoring tie flows at boundaries that are located at tie lines, and then open a new one where we export the two halves of a tie line as separate equipment. What do you think ?

zamarrenolm:
Proposed to throw an exception when a tie flow of a control area is defined at the boundary of a tie line.
The unit test has been updated to check the proper exception is thrown.
The complete fix requires to export the two halves of the tie line instead of the tie line as a single equipment, with proper reference to external boundary node. We think this should be handled in a separate PR.

PR #2109, opened 2022-05-09
CGMES export : create equivalent injection id and terminal if not present

zamarrenolm:

Bug fix

A CGMES network that is missing equivalent injections at boundaries is exported (no equivalent injections are present in the original CGMES files).
When the Network is exported, a new identifier is created for the exported equivalent injection, but not for its terminal.
The export fails because terminal identifier for the equivalent injection is not found.

When exporting to CGMES, if identifiers for equivalent injection equipment and terminals are not found, they are created.


miovd (commented):
Is it what we want or do we want to identify these as `SvInjection` during export? What do you think @annetill ?


EDIT: Never mind, I've had misunderstood the dev! I agree with it :)

miovd (approved):

PR #2122, opened 2022-05-17
Import/export parameter possible values

geofjamg:

No

Feature

In some case a parameter type could model an enum and it might be interesting to specify list of possible values.

**Other information**:


iidm/iidm-converter-api/src/main/java/com/powsybl/iidm/parameters/ConversionParameters.java
```diff
@@ -78,14 +76,22 @@ public static double readDoubleParameter(String format, Properties parameters, P
     }
 
     private static <T> T readParameter(String format, Properties parameters, Parameter configuredParameter, T defaultValue,
```
geofjamg, line 78: @miovd @flo-dup all static methods of ConversionParameters could go in Parameters  and ConversionParameters should better be removed?

geofjamg: Done!

PR #2126, opened 2022-05-18
Bus and branch result extension

geofjamg:

No

Feature

We cannot add extension to BusResult, BranchResult and ThreeWindingsTransformerResult.

We can

This is a breaking change because now we need to use custom jackson serializer (to handle the extension part) to have same json as before and because BusResults has been renamed to BusResult to be consistent with the others.

**Other information**:


geofjamg:
Sonar fail because of code duplication. I don't think we can/should fix it as it is not really possible/reasonable to share code for json parsing.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/BusResultSerializer.java
```diff
@@ -0,0 +1,36 @@
...
+public class BusResultSerializer extends StdSerializer<BusResult>  {
+
+    public BusResultSerializer() {
+        super(BusResult.class);
+    }
+
+    @Override
+    public void serialize(BusResult busResult, JsonGenerator jsonGenerator, SerializerProvider serializerProvider) throws IOException {
+        jsonGenerator.writeStartObject();
+        jsonGenerator.writeStringField("voltageLevelId", busResult.getVoltageLevelId());
```
miovd, line 29: Maybe not to be completely handled in this PR, but shouldn't these results be versionable? Maybe you could only add `jsonGenerator.writeStringField("version", "1.0");` for now and handle it properly if we add a change

miovd (commented):
I mostly agree with this PR, the note I made for `BusResult` is valid for the other serializers

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/BusResultSerializer.java
geofjamg: Bus/branch/3wt results are part of security analysis result which is already versioned.

miovd: Indeed, you're right... Okay for me then!

miovd (approved):

PR #2129, opened 2022-05-19
[WIP] Operational Limits Set

### LINKED ISSUE #2110: Operational limit sets

A feature.

For a line or a transformer, we have only one group (permanent limit + temporary limits for two sides) of each limit type (current, active power and apparent power) available in the core model, such as:
````
<currentLimits1 permanentLimit="580"/>
<currentLimits2 permanentLimit="580">
      <temporaryLimit name="N/A" value="734"/>
      <temporaryLimit name="IT105" acceptableDuration="6300" value="580"/>
      <temporaryLimit name="IT1" acceptableDuration="60"/>
</currentLimits2>
````
By definition, as we have only one group, it means that it is the active one. By active, we mean the one that has to be considered by the analysis tools (security analysis for example). 

In some studies (network development) or operational context (CGMES), we can have a set of operational limits. The season can be important to consider, the day vs the night, and many other factor that can lead to store a set of limits and not one, and let the user chooses the active set. 
- Important : I think that for reading and writing performance issue, we must add a parameter that forces import and export of only active limits (CGMES, internal importers). 
- We must have three `OperationalLimitSet`, one for current, one for active power and one for apparent power. The design looks nice in Miora's PR (see below), excepted that `activated` must be changed for `enabled` to define the active limit set. 
- Inside a set, each group (permanent limit + temporary limit for 2 sides) must have an id (String).
- In a set, the active group of limits is designed through its id (String).
- Most of the use cases, no breaking changes must appear. 
- Maybe, High or low for the modelling interpretation of the whole set. Note in that PR, to be checked with CGMES modelling.

This leads to a new version of IIDM (1.12).

Note that we must think about additional information in `CurrentLimits`, but not in that PR:
- A type can be added to a group. This type will described if the group is linked to an automaton/actions, or if it is only for supervision.
- An enum to distinguish protections from study limits.

  - PowSyBl Version: ...
  - OS Version: ...

miovd:

Fix #2110

**Other information**:


annetill:
See #2802

PR #2130, opened 2022-05-19
Matpower export

geofjamg:

No

Feature

We can import a Matpower file but not export. 

We can export an IIDM network to Matpower.

**Other information**:


matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerExporter.java
```diff
@@ -0,0 +1,419 @@
...
+                    .filter(Bus::isInMainConnectedComponent)
+                    .max(Comparator.comparingInt(MatpowerExporter::getBranchCount))
+                    .orElseThrow()
+                    .getId();
+            LOGGER.debug("Matpower reference bus automatically selected: {}", context.refBusId);
+        }
+        createBuses(network, model, context);
+        createBranches(network, model, context);
+        createGenerators(network, model, context);
+
```
sylvlecl, line 410: Should we check for "exotic" elements not supported by the format (SVCs, HVDCs...) and throw an exception in that case ? (or warn)

sylvlecl (approved):

PR #2137, opened 2022-05-23
CGMES export: ignore fictitious switches created for disconnected terminals

zamarrenolm:

Feature

When a CGMES model that contains disconnected terminals is imported, IIDM fictitious switches are created. If the IIDM network is exported back to CGMES, these switches are written as new equipment.

Fictitious switches corresponding to disconnected terminals are ignored for export.

**Other information**:
The CGMES `EquipmentExport` class has been refactored to improve readability.
The fictitious switches created for disconnected terminals are marked with the property `CGMES.isCreatedForDisconnectedTerminal` set to `true`.
In a similar way, the fictitious loads created for SvInjection objects are now marked with the property `CGMES.isCreatedForSvInjection` set to `true`.


marcosmc (approved):

miovd (commented):
Did you check on networks imported from a different format than CGMES that the information is not lost ? (for example with an IIDM network from scratch with fictitious switches)

I thought that this PR would have written more Terminal connectivity status, it is already done?

The PR is okay in principle for me, I just want to check that we are not missing something :)

miovd:
I'm thinking for example for IIDM network with fictitious switches, it means that we lose the switch ID? It's not ideal...

zamarrenolm:
> Did you check on networks imported from a different format than CGMES that the information is not lost ? (for example with an IIDM network from scratch with fictitious switches)
> 
> I thought that this PR would have written more Terminal connectivity status, it is already done?
> 
> The PR is okay in principle for me, I just want to check that we are not missing something :)

Yes, one of the unit tests for the export used the FictitiousSwitchFactory class to build an IIDM network full of fictitious switches. I will verify the results on it.

zamarrenolm:
> I'm thinking for example for IIDM network with fictitious switches, it means that we lose the switch ID? It's not ideal...

Yes, all fictitious switch are not present at all in the CGMES output.

If we do not want this behaviour we can restrict the ignored equipment in the export only to the switches we introduce when importing from CGMES to represent disconnected terminals. We could identify them based on the IIDM identifier that we build during the conversion.

zamarrenolm:
Updated after initial review:

- Only ignore for export fictitious switches corresponding to disconnected terminals.
- Mark these switches with a specific property.
- Also marked fictitious loads for SV injections with a specific property.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/SvInjectionConversion.java
```diff
@@ -49,6 +50,7 @@ public void convert() {
         identify(adder);
         connect(adder);
         Load load = adder.add();
+        load.setProperty(Conversion.PROPERTY_IS_CREATED_FOR_SV_INJECTION, "true");
```
miovd, line 53: I don't think you need to do that for fictitious loads (and in fact, I think we shouldn't as it makes the model heavier): it is not a problem in my opinion to export fictitious loads as SV injections in every case as we do not lose any IDs. The issue with switches was the possibility to lose IDs after export/import (for example).

miovd (changes requested):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/SvInjectionConversion.java
zamarrenolm: The aim of adding the property was to be sure that we do not ignore in the export other fictitious loads that may have been created in the model for other purposes.
Native IIDM cases could contain loads marked as "fictitious" for some reason by the users, but they want them to be exported to the EQ as regular loads. 
If we remove this property we won't we able to distinguish between these two situations, and we will ignore all fictitious loads in the export.
Is it ok for you?

miovd: For me, it is not an issue as no information loss is performed in this case. @annetill do you think it is of importance?

miovd: I've just discussed with Anne: we agree to export fictitious loads as SvInjection in any case, at least for now (we will see later if we have a new use case)

zamarrenolm: Ok. I will remove the property for fictitious loads. All fictitious loads will be exported as `SvInjection`, they will not show in the EQ file.

miovd (approved):

PR #2144, opened 2022-05-30
Move GeneratorStartup extension to powsybl-core

miovd:

No

Moving code

`GeneratorStartup` is in [powsybl/powsybl-network-hypothesis](https://github.com/powsybl/powsybl-network-hypothesis)

`GeneratorStartup` is in core

**Other information**:

### DIFF
--- a/iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/GeneratorStartup.java
+++ b/iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/GeneratorStartup.java
@@ -1,0 +1,41 @@
+/**
+ * Copyright (c) 2021, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.iidm.network.extensions;
+
+import com.powsybl.commons.extensions.Extension;
+import com.powsybl.iidm.network.Generator;
+
+/**
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ */
+public interface GeneratorStartup extends Extension<Generator> {
+
+    @Override
+    default String getName() {
+        return "startup";
+    }
+
+    float getPredefinedActivePowerSetpoint();
+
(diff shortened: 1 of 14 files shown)


geofjamg:
@miovd @annetill attribute naming seems not to bad?

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/GeneratorStartup.java
```diff
@@ -0,0 +1,56 @@
...
+import com.powsybl.iidm.network.Generator;
+
+/**
+ * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
+ */
+public interface GeneratorStartup extends Extension<Generator> {
+
+    @Override
+    default String getName() {
+        return "generatorStartup";
```
miovd, line 19: I would change it back to `startup` to be consistent with other extension name (see `LoadDetail` for example). The extension name should not contain the equipment.

geofjamg: Be careful to keep compat with existing data

miovd:
> > > I changed the PR to WIP, to let time to answer an important question: we are not able to load again xiidm from CVG because everything has been renamed. Is it really ok? Maybe we should keep compat in xml read?
> > 
> > 
> > You're right! There has been two renamed attributes, I will add the backward compatibility
> 
> This is also incompatible because name of the extension has changed. We should add a unit test with old xml to ensure it is correctly imported.

The extension's name has not changed, I reverted Anne's changes, it is still `startup`

miovd:
> > > > > I changed the PR to WIP, to let time to answer an important question: we are not able to load again xiidm from CVG because everything has been renamed. Is it really ok? Maybe we should keep compat in xml read?
> > > > 
> > > > 
> > > > You're right! There has been two renamed attributes, I will add the backward compatibility
> > > 
> > > 
> > > This is also incompatible because name of the extension has changed. We should add a unit test with old xml to ensure it is correctly imported.
> > 
> > 
> > The extension's name has not changed, I reverted Anne's changes, it is still `startup`
> 
> Ah ok... so it was an old mistake, generatorStartup would have been more meaningful. To fix this, we could support some extension name aliases in `ExtensionXmlSerializer` to allow extension renaming.

As I said above, is it what we want? All other extensions don't have the extendable type in their name. Or do we want to change everywhere?

geofjamg (approved):

PR #2147, opened 2022-05-30
Add possible values for extensions (from present extensions modules)

miovd:

No

Add possible values for extensions parameters in import/export
NB: not sure this should be done as it is a bit restrictive (you have to put something compatible in your config)... What do you think? @annetill @geofjamg @sylvlecl 


geofjamg:
@miovd why it would be too restrictive as it list all network extensions at runtime (so all other choice would lead to a extension not found) ?

geofjamg (approved):

sylvlecl (approved):

PR #2151, opened 2022-05-31
CGMES export: ensure all exported identifiers are valid

zamarrenolm:

Feature

Some real world CGMES cases contain invalid mRID identifiers that do not follow the pattern.
If these cases are imported in PowSyBl and exported back to CGMES, some identifiers where not "fixed" by the `CgmesAliasNamingStrategy`, resulting in CGMES output files keeping the "bad" IDs.

The naming strategy is used in the CGMES export to process identifiers for additional objects present in the CGMES model:

- Equivalent injections.
- Tap changers.
- Transformer ends.
- Regions and sub-regions.
- Terminals.
- Generating units.


zamarrenolm:
Review of read and write id mappings:

All mappings stored in the idByUuid bi-map are read and written to persistent storage. The new mapping additions are preserved.

`readIdMapping` creates alias `UUID` only for selected elements, it is called on request. There is no problem if the mapping contains more objects than the ones that are stored in UUID aliases.

miovd:
I'm not sure this is useful? The `CgmesNamingStrategy` is supposed to be used only if networks are created from scratch or imported from another format at the start. If IDs are not valid, but the origin format is CGMES, I don't think it is powsybl role to correct it.

zamarrenolm:
> I'm not sure this is useful? The `CgmesNamingStrategy` is supposed to be used only if networks are created from scratch or imported from another format at the start. If IDs are not valid, but the origin format is CGMES, I don't think it is powsybl role to correct it.

Agree, but we have found some real-world CGMES example were all used IDs were not following the rules and thought it was a good example to test that the naming strategy could be extended to solve it. This helps us ensure that we are generating "valid" CGMES files.

miovd:
> > I'm not sure this is useful? The `CgmesNamingStrategy` is supposed to be used only if networks are created from scratch or imported from another format at the start. If IDs are not valid, but the origin format is CGMES, I don't think it is powsybl role to correct it.
> 
> Agree, but we have found some real-world CGMES example were all used IDs were not following the rules and thought it was a good example to test that the naming strategy could be extended to solve it. This helps us ensure that we are generating "valid" CGMES files.

I see! I will check if this is something we want to support though. In the meantime, can you also check how much this PR affects performance, with and without IDs changes?

zamarrenolm:
> > > I'm not sure this is useful? The `CgmesNamingStrategy` is supposed to be used only if networks are created from scratch or imported from another format at the start. If IDs are not valid, but the origin format is CGMES, I don't think it is powsybl role to correct it.
> > 
> > 
> > Agree, but we have found some real-world CGMES example were all used IDs were not following the rules and thought it was a good example to test that the naming strategy could be extended to solve it. This helps us ensure that we are generating "valid" CGMES files.
> 
> I see! I will check if this is something we want to support though. In the meantime, can you also check how much this PR affects performance, with and without IDs changes?

Sure, I will measure impact on performance.
BTW, in the latest available dataset, there was ONE TSO that used identifiers not based on UUIDs for everything. And ALL of them that had "only" some invalid identifiers, usually built with an UUID and a suffix: `<UUID>_RATIO`, `<UUID>_GU`, `<UUID>-1`, `<UUID>_TERM1`, ...

zamarrenolm:
Impact on performance: the CGMES naming strategy increases export time by 10%, compared to the identity naming strategy (that does not review identifiers to ensure they are valid and does not change them).

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesAliasNamingStrategyFixInvalidIds.java
```diff
@@ -20,24 +20,16 @@
...
 import java.util.Arrays;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
 
 /**
  * @author Miora Vedelago <miora.ralambotiana at rte-france.com>
  */
-public class CgmesAliasNamingStrategy implements NamingStrategy {
+public class CgmesAliasNamingStrategyFixInvalidIds implements NamingStrategy {
```
miovd, line 28: Maybe use a more user-friendly name such as  `FixedCgmesAliasNamingStrategy` and `SimpleCgmesAliasNamingStrategy`?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesAliasNamingStrategyFixOnlyIidmIdentifiableInvalidIds.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+package com.powsybl.cgmes.conversion;
+
+import com.powsybl.commons.PowsyblException;
+import com.powsybl.iidm.network.Identifiable;
+
+/**
+ * @author Luma Zamarreño <zamarrenolm at rte-france.com>
+ */
+public class CgmesAliasNamingStrategyFixOnlyIidmIdentifiableInvalidIds extends CgmesAliasNamingStrategyFixInvalidIds {
```
miovd, line 15: Rather than extending the other class, I would advise to use a common abstract parent class. Otherwise, we might not control evolutions (if we forget to override something for example)...

miovd (changes requested):
Some remarks.

Note: not to be done in this PR, but it would be great if naming strategies could be retrieve from service loaders... to be improved later.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesAliasNamingStrategyFixInvalidIds.java
```diff
@@ -20,24 +20,16 @@
...
 
     private final BiMap<String, String> idByUuid = HashBiMap.create();
 
-    public CgmesAliasNamingStrategy() {
-    }
-
-    public CgmesAliasNamingStrategy(Map<String, String> idByUuid) {
-        this.idByUuid.putAll(Objects.requireNonNull(idByUuid));
-    }
```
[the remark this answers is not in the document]
zamarrenolm: I will keep the constructors. I removed the constructor receiving a map because it was not used.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesAliasNamingStrategyFixOnlyIidmIdentifiableInvalidIds.java
zamarrenolm: I've followed your advice and created an abstract class that is a common parent for the two implementations. I've added an abstract method `getName`, that also will be useful when we refactor these classes with service loaders.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesAliasNamingStrategyFixInvalidIds.java
zamarrenolm: Followed your naming proposal.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
```diff
@@ -328,6 +330,11 @@ private void copyStream(ReadOnlyDataSource from, DataSource to, String fromName,
             ParameterType.STRING,
             "Path of ID mapping file",
             null);
+    private static final Parameter ID_MAPPING_FILE_NAMING_STRATEGY_PARAMETER = new Parameter(
+            ID_MAPPING_FILE_NAMING_STRATEGY,
+            ParameterType.STRING,
+            "Configure what type of naming strategy you want to use for the provided ID mapping file",
+            "cgmes");
```
miovd, line 337: In order to prevent breaking change, I think the default parameter should be "identity"

miovd (commented):
A small comment and then, it is okay for me

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
zamarrenolm: Done. The unit test had to be adjusted to avoid using identity naming strategy when there is a mapping file.

miovd (approved):

PR #2152, opened 2022-05-31
CGMES export: refactor profile writers to use context

zamarrenolm:

code refactoring

The classes that write the profiles have static methods with too many parameters.

Refactor writers of CGMES instance files to remove Sonar code smells and issues.

**Other information**:

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
@@ -39,6 +39,7 @@ public class CgmesExportContext {
     private CgmesTopologyKind topologyKind = CgmesTopologyKind.BUS_BRANCH;
     private DateTime scenarioTime = DateTime.now();
 
+    private final Network network;
     private final ModelDescription eqModelDescription = new ModelDescription("EQ Model", cim.getProfile("EQ"));
     private final ModelDescription tpModelDescription = new ModelDescription("TP Model", cim.getProfile("TP"));
     private final ModelDescription svModelDescription = new ModelDescription("SV Model", cim.getProfile("SV"));
@@ -82,6 +83,10 @@ public class CgmesExportContext {
                 getSvModelDescription().addDependency(sshModelId);
             }
         }
+    }
+
+    public Network getNetwork() {
+        return this.network;
     }
 
     public static final class ModelDescription {
@@ -169,6 +174,8 @@ public class CgmesExportContext {
     }
 
     public CgmesExportContext() {
+        // FIXME(Luma) review how this is used, no default constructor should be allowed
+        this.network = null;
     }
 
     public CgmesExportContext(Network network) {
@@ -180,6 +187,7 @@ public class CgmesExportContext {
     }
 
     public CgmesExportContext(Network network, boolean withTopologicalMapping, NamingStrategy namingStrategy) {
+        this.network = network;
         this.namingStrategy = namingStrategy;
         CimCharacteristics cimCharacteristics = network.getExtension(CimCharacteristics.class);
         if (cimCharacteristics != null) {
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
@@ -24,7 +24,7 @@ import java.util.*;
 /**
  * @author Marcos de Miguel <demiguelm at aia.es>
  */
-public final class EquipmentExport {
+public final class EquipmentExport extends AbstractCgmesProfileWriter {
 
     private static final String ACDCCONVERTERDCTERMINAL = "ACDCConverterDCTerminal";
     private static final String CONNECTIVITY_NODE_SUFFIX = "CN";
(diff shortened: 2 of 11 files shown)


zamarrenolm:
A thorough review is required after the recent changes in export

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
```diff
@@ -28,93 +28,93 @@
...
+    private final Map<String, List<RegulatingControlView>> regulatingControlViews = new HashMap<>();
 
-    public static void write(Network network, XMLStreamWriter writer, CgmesExportContext context) {
-        final Map<String, List<RegulatingControlView>> regulatingControlViews = new HashMap<>();
-        String cimNamespace = context.getCim().getNamespace();
+    SteadyStateHypothesisExport(CgmesExportContext context, XMLStreamWriter xmlWriter) {
+        super(context, xmlWriter);
+    }
 
+    public void export() {
```
miovd, line 43: You can add an abstract `export()` method in `AbstractCgmesExporter` (to be implemented in inherited classes)

miovd: Didn't realize it was already the case! Only add `@Override` to indicate it.

zamarrenolm: Added the missing `@override` annotations.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
```diff
@@ -186,13 +214,16 @@ public String getId() {
     }
 
     public CgmesExportContext() {
+        // FIXME(Luma) review how this is used, no default constructor should be allowed
+        this.network = null;
```
miovd, line 218: This is used to export merging view: we don't want the export to be built from the merging view but from the extensions of the different networks. In this case you can't set `network` to `null`. I actually don't think network should be in the context... Why this design? We should pass the network to retrieve information/data when necessary, but not store it  in my opinion.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportUtil.java
```diff
@@ -53,6 +53,7 @@ private CgmesExportUtil() {
     private static final Pattern ENTSOE_BD_EXCEPTIONS_PATTERN1 = Pattern.compile("(?i)[a-f\\d]{8}-[a-f\\d]{4}-[a-f\\d]{4}-[a-f\\d]{4}-[a-f\\d]{7}");
     private static final Pattern ENTSOE_BD_EXCEPTIONS_PATTERN2 = Pattern.compile("(?i)[a-f\\d]{8}[a-f\\d]{4}[a-f\\d]{4}[a-f\\d]{4}[a-f\\d]{12}");
 
+    // FIXME(Luma) move these methods to cgmes writer
```
miovd, line 56: Can you explain these `FIXME`? Do you plan other PRs?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/AbstractCgmesExporter.java
```diff
@@ -0,0 +1,64 @@
...
+
+public abstract class AbstractCgmesExporter {
+    protected final CgmesExportContext context;
+    protected final XMLStreamWriter xmlWriter;
+    protected final String cimNamespace;
+
+    protected AbstractCgmesExporter(CgmesExportContext context, XMLStreamWriter xmlWriter) {
+        this.context = context;
+        this.xmlWriter = xmlWriter;
+        this.cimNamespace = context.getCim().getNamespace();
```
miovd, line 18: Add `Objects.requireNonNull`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
```diff
@@ -35,92 +35,95 @@ public final class EquipmentExport {
...
+    private final Set<String> regulatingControlsWritten = new HashSet<>();
+
+    EquipmentExport(CgmesExportContext context, XMLStreamWriter xmlWriter) {
+        super(context, xmlWriter);
+        euNamespace = context.getCim().getEuNamespace();
+        limitValueAttributeName = context.getCim().getLimitValueAttributeName();
+        limitTypeAttributeName = context.getCim().getLimitTypeAttributeName();
+        limitKindClassName = context.getCim().getLimitKindClassName();
+        writeInfiniteDuration = context.getCim().writeLimitInfiniteDuration();
+        writeInitialP = context.getCim().writeGeneratingUnitInitialP();
```
miovd, line 56: I don't know if you need to store all these if they are already in the context which is stored.

miovd (changes requested):
I'm still thinking about the design but some remarks to start off.

I think you shouldn't pass the network as a context parameter. Just pass it in `export(Network network)` methods.

miovd:
About this refactoring: not sure this is the best moment to work on it as the export will probably move a bit with the merging issues...

zamarrenolm:
> About this refactoring: not sure this is the best moment to work on it as the export will probably move a bit with the merging issues...

I will try to keep the branch updated with main.

zamarrenolm:
> Can I close this old PR as no agreement has been found on that subject?

Sure. We can reopen the task for cleaning-up export modules when the changes in this feature are less frequent.

PR #2161, opened 2022-06-08
CGMES export: fix hidden tap changer export

zamarrenolm:

Bug fix

Hidden tap changers are stored at two-winding transformers when CGMES original data contains two tap changers, one at each end (primary and secondary): they are found in configurations where one of them can be changed off-load. In IIDM, both are combined in a single one. The hidden tap changer is treated as fixed, its step can not be changed in PowSyBl simulations. 
The SSH data for these hidden tap changers is always written in the output.
The SV data (the original step) is never written in the output.

Hidden tap changers data is exported to SSH and SV files for results-only export, that is, when SSH or SV are being exported, but not EQ definitions.
If equipment definitions are exported, the hidden tap changers are not written in the EQ file (their model has been combined with the "main" tap changers left in the IIDM model), and they must not be present in the SSH, SV files.


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -74,6 +74,7 @@ public void export(Network network, Properties params, DataSource ds) {
         try {
             List<String> profiles = Parameter.readStringList(getFormat(), params, PROFILES_PARAMETER, defaultValueConfig);
             if (profiles.contains("EQ")) {
+                context.setExportEquipment(true);
```
miovd, line 77: Shouldn't be used in `CgmesExport` for me but called in `EquipmentExport` for those directly using the classes

miovd (changes requested):
Small changes required and then it is good for me!

miovd (approved):

PR #2186, opened 2022-06-29
Sensitivity analysis: add contingency status in the results

annetill:

**Other information**:


obrix:
@annetill Thanks for opening the PR. I added the CSV and JSON writers but they are still unused. Is it relevant to add the output of contingency status to the SensitivityAnalysisTool ? @geofjamg maybe you have an opinion.

obrix:
One issue arising from using the same writer for Contingency Status and Sensitivity Value is that the user is forced to have the same output format for both (json or csv).
The SensitivityAnalysisTool is impacted. I chose to keep the same interface and to simply output the contingency status in a file with the same name as for the sensitivity value (suffixed by "ContingencyStatus"). Another solution could be to add a new parameter to explicitily name the file for contingency status and maybe to add a format parameter (ie "json" or "csv") to make sure the output format is the same for both sensitivity value and contingency status files.
Please let me know if another implementation is more suited.

flo-dup:
As discussed, I think it could be of interest to let the user choose the two json/csv names, but avoid adding an extra parameter for the format. If they do not have the same extension we could deal with it, I thought at first with an exception but that's a bit harsh, a warning and replacing the second extension with the first could be also ok don't you think?

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysis.java
```diff
@@ -49,7 +49,7 @@ public Runner(SensitivityAnalysisProvider provider) {
         public CompletableFuture<Void> runAsync(Network network,
                                                 String workingVariantId,
                                                 SensitivityFactorReader factorReader,
-                                                SensitivityValueWriter valueWriter,
+                                                SensitivityResultWriter valueWriter,
```
geofjamg, line 52: valueWriter to rename to resultWriter everywhere

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisTool.java
```diff
@@ -176,22 +217,46 @@ public void run(CommandLine line, ToolRunningContext context) throws Exception {
...
                 }
+                List<SensitivityFactor> factors = new ArrayList<>();
+                factorsReader.read((functionType, functionId, variableType, variableId, variableSet, contingencyContext) -> {
+                    factors.add(new SensitivityFactor(functionType, functionId, variableType, variableId, variableSet, contingencyContext));
+                });
+                SensitivityAnalysisResult result = SensitivityAnalysis.run(network, network.getVariantManager().getWorkingVariantId(),
+                        factors, contingencies, variableSets, params,
+                        computationManager, Reporter.NO_OP);
+                ObjectMapper sensiObjectMapper = JsonUtil.createObjectMapper().registerModule(new SensitivityJsonModule());
+                JsonUtil.writeJson(outputFile, result, sensiObjectMapper);
```
geofjamg, line 233: So when single output is chosen, we have a larger memory consumption. Why not, but as contingency status are very light data, it would have been possible to bufferered all and write at the end (which is not possible with values as we can have millions of them)

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityResultJsonWriter.java
```diff
@@ -15,28 +15,37 @@
...
 
     @Override
-    public void write(int factorIndex, int contingencyIndex, double value, double functionReference) {
+    public void writeSensitivityValue(int factorIndex, int contingencyIndex, double value, double functionReference) {
         SensitivityValue.writeJson(jsonGenerator, factorIndex, contingencyIndex, value, functionReference);
     }
 
+    @Override
+    public void writeContingencyStatus(SensitivityAnalysisResult.SensitivityContingencyStatus status) {
+        SensitivityAnalysisResult.SensitivityContingencyStatus.writeJson(jsonStatusGenerator, status);
```
geofjamg, line 41: Maybe a single file strategy can be easily implemented here by just storing status in that method and write all in the close method.

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/json/SensitivityAnalysisResultDeserializer.java
```diff
@@ -0,0 +1,62 @@
+package com.powsybl.sensitivity.json;
```
geofjamg, line 1: Licence

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/json/SensitivityAnalysisResultDeserializer.java
```diff
@@ -0,0 +1,62 @@
...
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import com.powsybl.sensitivity.SensitivityAnalysisResult;
+import com.powsybl.sensitivity.SensitivityFactor;
+import com.powsybl.sensitivity.SensitivityValue;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class SensitivityAnalysisResultDeserializer extends StdDeserializer<SensitivityAnalysisResult> {
```
geofjamg, line 17: authorship

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/json/SensitivityAnalysisResultSerializer.java
```diff
@@ -0,0 +1,27 @@
+package com.powsybl.sensitivity.json;
```
geofjamg, line 1: Licence

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/json/SensitivityAnalysisResultSerializer.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+import com.powsybl.sensitivity.SensitivityAnalysisResult;
+
+import java.io.IOException;
+
+public class SensitivityAnalysisResultSerializer extends StdSerializer<SensitivityAnalysisResult> {
```
geofjamg, line 10: authorship

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/json/SensitivityContingencyStatusJsonDeserializer.java
```diff
@@ -0,0 +1,18 @@
+package com.powsybl.sensitivity.json;
```
geofjamg, line 1: Licence

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/json/SensitivityContingencyStatusJsonDeserializer.java
```diff
@@ -0,0 +1,18 @@
...
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import com.powsybl.sensitivity.SensitivityAnalysisResult;
+
+public class SensitivityContingencyStatusJsonDeserializer extends StdDeserializer<SensitivityAnalysisResult.SensitivityContingencyStatus> {
```
geofjamg, line 8: authorship

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/json/SensitivityContingencyStatusJsonSerializer.java
```diff
@@ -0,0 +1,18 @@
+package com.powsybl.sensitivity.json;
```
geofjamg, line 1: Licence

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/json/SensitivityContingencyStatusJsonSerializer.java
```diff
@@ -0,0 +1,18 @@
...
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+import com.powsybl.sensitivity.SensitivityAnalysisResult;
+
+public class SensitivityContingencyStatusJsonSerializer extends StdSerializer<SensitivityAnalysisResult.SensitivityContingencyStatus> {
```
geofjamg, line 8: authorship

sensitivity-analysis-api/src/test/resources/contingencyStatusRef.json
```diff
@@ -0,0 +1,4 @@
+{
+  "ContingencyId" : "C1",
```
geofjamg, line 2: Should start with a lower case

sensitivity-analysis-api/src/test/resources/contingencyStatusRef.json
```diff
@@ -0,0 +1,4 @@
...
+  "ContingencyStatus" : "CONVERGED"
```
geofjamg, line 3: Same

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisTool.java
```diff
@@ -55,6 +58,8 @@ public class SensitivityAnalysisTool implements Tool {
     private static final String CONTINGENCIES_FILE_OPTION = "contingencies-file";
     private static final String VARIABLE_SETS_FILE_OPTION = "variable-sets-file";
     private static final String PARAMETERS_FILE = "parameters-file";
+    private static final String OUTPUT_CONTINGENCY_STATUS_FILE_OPTION = "output-contingency-file";
+    private static final String OUTPUT_UNIFIED = "output-unified";
```
geofjamg, line 62: single-output would be better?

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisTool.java
```diff
@@ -133,11 +146,39 @@ private static boolean isCsv(Path outputFile) {
         }
     }
 
+    private static String buildContingencyStatusPath(boolean csv, String outputFile) {
+        if (csv) {
+            return outputFile.replace(".csv", "ContingencyStatus.csv");
```
geofjamg, line 151: Maybe something with a _ would be more clean like
toto.csv -> toto_contingency_status.csv ?
toto.json -> toto_contingency_status.json ?

geofjamg (changes requested):

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityResultJsonWriter.java
obrix: @geofjamg It can. But it will output an heterogeneous format that will need a specific deserializer to read back (see SensitivityAnalysisToolTest.runJsonOutput). We will then have three possible formats  : SensitivityValue / ContingencyStatus separated, an hybrid SensitivityValue / ContingencyStatus with a custom deserializer to read back the output from the writer, and SensitivityAnalysisResult which handle status / contingency / factors. Is that okay for you ?

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisTool.java
obrix: Yes for efficiency writer based API will remain the best.

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityResultJsonWriter.java
obrix: Note : In fact object mapper has some convenient function to read heterogeneous arrays of data directly. So no need for a new custom deserializer (see last commit).

geofjamg (approved):

PR #2214, opened 2022-07-13
Network modification: create feeder bay

colinepiloquet:

No

Feature

With this PR, it is now possible to add an injection to an existing voltageLevel on a given network. The topology is created as well : a disconnector between the injection and the busbar sections and a switch on every bus bar section of the slice where the injection is connected. This feature is only available in node/breaker mode.


iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoadOnVoltageLevel.java
```diff
@@ -0,0 +1,171 @@
...
+import java.util.List;
+
+import static com.powsybl.iidm.modification.topology.TopologyModificationUtils.*;
+
+/**
+ * This method adds a new load on an existing voltage level. The voltage level should be described
+ * in node/breaker.
+ * @author Coline Piloquet <coline.piloquet at rte-france.com>
+ */
+public class AttachLoadOnVoltageLevel implements NetworkModification {
```
geofjamg, line 24: Attach does not really mean anything, CreateFeeder ou CreateFeederCell ou maybe CreateBay ?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoadOnVoltageLevel.java
```diff
@@ -0,0 +1,171 @@
...
+    private final double q0;
+    private final LoadType loadType; //add constant
+
+    //VoltageLevel attributes
+    private final String voltageLevelId;
+    private String bbsId; //Id of the busBar where the switch will be open
+
+    private List<String> bbsIds;
+
+    public AttachLoadOnVoltageLevel(String loadId, double p0, double q0,
```
geofjamg, line 38: This is not the right design because:
 - the idea is not to duplicate the IIDM adder which already do most of the job
 - but just kind of extend adder capabilities to create a full feeder cell instead of just the injection
 - we don't want to duplicate all this for each of the network equipment, it should be generic to injections and branches.

It has to look like:
```java
Network n = ...
LoadAdder adder = vl.newLoadAdder
  .setId(...)
 // WE DON'T NEED TO SET THE NODE !!!!!!!
  ...
.add()
new InjectionFeederCallCreation(adder, "BBS", CellType::ONE_BREAKER) // busbar section could be optional, and if not defined automatically choosen
   .apply(n);
```

geofjamg (changes requested):

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoadOnVoltageLevel.java
geofjamg: Also one very important thinks when it comes to network modification is functional logging, otherwise there is no way to known what has been done especially in loose mode when it case of equipment not done where there is no exception.
So we need:
  - to create a report when modification is ok to explain what has been done
  - to create a report when modification has not been done to explain exactly what was wrong

@annetill also we need to to the same in AttachNewLineOnLine and AttachVoltageLevelOnLine. In gridsuite integration we added functional logs externally but this is not the right way to do, it should have been done inside powsybl. Also in case of error this is the only place to describe what was wrong.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoadOnVoltageLevel.java
```diff
@@ -0,0 +1,171 @@
...
+        int loadNode = voltageLevel.getNodeBreakerView().getMaximumNodeIndex() + 1;
+        int forkNode = loadNode + 1;
+        loadAdder.setNode(loadNode).add();
+
+        //Create a switch and a breaker linking the load to the busbar sections
+        if (bbsId != null) {
+            createTopologyAutomatically(network, voltageLevel, loadNode, forkNode);
+        }
+
+        if (bbsIds != null) {
```
miovd, line 163: If someone enters a list that contains only one element, what behavior should this have?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoadOnVoltageLevel.java
```diff
@@ -0,0 +1,171 @@
...
+
+    public void setBbsIds(List<String> bbsIds) {
+        this.bbsIds = bbsIds;
+    }
+
+    private void createTopologyAutomatically(Network network, VoltageLevel voltageLevel, int loadNode, int forkNode) {
+        BusbarSection bbs = network.getBusbarSection(bbsId);
+        int bbsNode = bbs.getTerminal().getNodeBreakerView().getNode();
+        createNodeBreakerSwitches(loadNode, forkNode, bbsNode, loadId, voltageLevel.getNodeBreakerView());
+        if (voltageLevel.getNodeBreakerView().getBusbarSectionCount() > 1) {
```
miovd, line 107: Is the other case normal? Maybe we should at least write a warning (to be integrated with Reporter later)

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoadOnVoltageLevel.java
```diff
@@ -0,0 +1,171 @@
...
+            BusbarSectionPosition position = bbs.getExtension(BusbarSectionPosition.class);
+            voltageLevel.getNodeBreakerView().getBusbarSectionStream().forEach(busbarSection -> {
+                if (busbarSection.getExtension(BusbarSectionPosition.class).getSectionIndex() == position.getSectionIndex() && !busbarSection.getId().equals(bbsId)) {
```
miovd, line 110: Same remark

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoadOnVoltageLevel.java
```diff
@@ -0,0 +1,171 @@
...
+                }
+            });
+        }
+    }
+
+    private void createTopologyFromBusbarList(Network network, VoltageLevel voltageLevel, int loadNode, int forkNode) {
+        createNodeBreakerBreaker(loadNode, forkNode, "", loadId, voltageLevel.getNodeBreakerView(), false);
+        bbsIds.stream().forEach(id -> {
+            BusbarSection bbs = network.getBusbarSection(id);
+            int bbsNode = bbs.getTerminal().getNodeBreakerView().getNode();
```
miovd, line 122: If bbs is null, log or throw

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoad.java
```diff
@@ -0,0 +1,265 @@
...
+
+import static com.powsybl.iidm.modification.ModificationUtils.throwExceptionOrLogError;
+import static com.powsybl.iidm.modification.topology.TopologyModificationUtils.*;
+
+/**
+ * This method adds a new load on an existing voltage level. The voltage level should be described
+ * in node/breaker topology.
+ * @author Coline Piloquet <coline.piloquet at rte-france.com>
+ */
+public class AttachLoad implements NetworkModification {
```
geofjamg, line 33: To rename: AttachLoad does not mean anything

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoad.java
```diff
@@ -0,0 +1,265 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AttachLoad.class);
+
+    private final LoadAdder loadAdder;
```
geofjamg, line 37: Should be a Function<VoltageLevel, LoadAdder> 
otherwise we could write dump and not working code:
```java
n1 =  ...
n2 = ...
adder = n1.getVoltageLevel('vl').newLoadAdder()
   .setId("x")
   ...
   .add();
new AttachLoad(adder, ...)
  .apply(n2);
```

geofjamg (changes requested):

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoad.java
flo-dup: Couldn't we check that the network of the `Load` is the same as the one passed in `apply`? Unfortunately this requires to create the load, as we don't have access to the adder parameters. We need to find a way to avoid the function, as if we think about serializing the `AttachLoad`, how do we deal with the `Function<VoltageLevel, LoadAdder>`?

iidm/iidm-modification/src/test/java/com/powsybl/iidm/modification/topology/AttachLoadTest.java
```diff
@@ -0,0 +1,161 @@
...
+                .withLoadPositionInsideSection(AttachLoad.PositionInsideSection.LAST)
+                .build();
+        assertEquals(AttachLoad.PositionInsideSection.LAST, modification1.getLoadPositionInsideSection());
+    }
+
+    @Test
+    public void testFeederOrders() {
+        Network network = Importers.loadNetwork("testNetworkNodeBreaker.xiidm", getClass().getResourceAsStream("/testNetworkNodeBreaker.xiidm"));
+        Map<String, Integer> feederOrders = TopologyModificationUtils.getFeederOrders(network.getVoltageLevel("vl1"));
+        assertEquals(90, feederOrders.get("trf8_terminal1"), 0);
```
flo-dup, line 149: I think we'd better compare ints here instead of floats with a delta

flo-dup (changes requested):

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoad.java
flo-dup: It was named like that mainly because of the existing `AttachVoltageLevelOnLine` and `AttachNewLineOnLine`, but you suggested in #2218 to rename them by replacing `Attach` into `Insert`. Shouldn't we use the same prefix here and for those two other modifications? 

What was objected against the `Create` you suggested previously is that the `InjectionAdder` is given here, so that the load is already half-way to its creation. Isn't it more a connect (or insert, or ... attach!) operation then?

flo-dup:
A more general comment, what makes the current `AttachLoad` specific to a `Load`? It seems to me that it still works if we rename it to `AttachInjection` with an `InjectionAdder` parameter instead of a `LoadAdder` one.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoad.java
sylvlecl: Why not completely add the adder from the user, and expose in this class only the methods that are still relevant?
I didn't look at the code in details, but for ex. I guess that methods such as setNode() or add() should not be called in those use cases:
hiding the adder would simply make it impossible for the user to do it, which would be more clear.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoad.java
geofjamg: No, until add has been called nothing has been creates, just some data have been filled

geofjamg:
> A more general comment, what makes the current `AttachLoad` specific to a `Load`? It seems to me that it still works if we rename it to `AttachInjection` with an `InjectionAdder` parameter instead of a `LoadAdder` one.

Yes, that was the goal

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoad.java
geofjamg: The idea was to not duplicate the adder code in network modification.
Load creation is not a good example as it is very simple, but for a transformer for instance where you have a complex logic with tap changer, limits it would be a pity a replicate all of this here.
Also basing everything on adder would allow to have only one network modification for all injections, one for all branches. So InjectionFeederCreation (is should be renamed with something like this), would be a very short and simple code that would allow to create and connect all possible injection.

setNode could be call outside, BUT it will be rewritten inside 'modify' method before calling 'add' so this is not really a problem.

I am not sure to understand what you mean by hiding the adder.

The only drawback relying on adder provider is that is does not allow to serialize the whole modification (which is probably unacceptable). So we have to think about a solution for that.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoad.java
geofjamg: A adder since add is not called is just a form, nothing else. So this this the modify method which call the add that create the injection and create also other topology stuffs.

geofjamg: Regarding insert or create. In the the AttachVoltageLevelOnLine and AttachNewLineOnLine it is a creation, but more precisely we insert it into an existing equipment.
In this PR, it is really a creation PLUS a cell creation (all switches, internal connections).

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoad.java
geofjamg: > Couldn't we check that the network of the `Load` is the same as the one passed in `apply`

That would be a design flaw

geofjamg: > Couldn't we check that the network of the `Load` is the same as the one passed in `apply`? Unfortunately this requires to create the load, as we don't have access to the adder parameters. We need to find a way to avoid the function, as if we think about serializing the `AttachLoad`, how do we deal with the `Function<VoltageLevel, LoadAdder>`?

Yes this is the worried part.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoad.java
```diff
@@ -0,0 +1,265 @@
...
+    private final String voltageLevelId;
+    private String bbsId;
+    private int loadPositionOrder;
+    private PositionInsideSection loadPositionInsideSection = PositionInsideSection.SPECIFIC;
```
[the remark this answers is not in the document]
geofjamg: what do you mean?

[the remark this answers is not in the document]
flo-dup: I mean that `AttachLoad` should have only an explicit `loadPositionOrder` I think, and not the current pair `loadPositionInsideSection`, `loadPositionOrder`. The enum `PositionInsideSection` is there to help the user which does not know the orders in his voltage level, I think it would make more sense to put that apart that class.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoad.java
sylvlecl: _Still about using the adder:_
actually it seems that at some point we will need to have `NetworkModification`s for adding elements anyway, in which we will need to store all the data of the element ?
So I think that at some point we will anyway need some POJO representation of all those attributes, separate from the adders, no ?

(by hinding the adder, I meant not exposing it in the API of the class, but maybe use it internally, for ex. having a `setId` which would call the `setId` of the adder, but finally I am not sure it's a good idea either, because of the serialization issue)

_About the re-use of one `InjectionFeederCreation`_, I understand the idea, but I am not sure it's very interesting since you need to specify their "connection point" individually anyway.

colinepiloquet:
The main class is renamed CreateBay as advised by IEC 60050 (https://www.electropedia.org/iev/iev.nsf/display?openform&ievref=605-02-09).
However, for now it is not possible to generalize to all injections because the add() is not in the InjectionAdder...

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoad.java
colinepiloquet: After checking for the right wording, we came to the conclusion that CreateBay is the best. The load is indeed created and connected with this modification. According to IEC60050, the correct word is Bay.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoad.java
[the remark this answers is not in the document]
colinepiloquet: I moved this into a method in TopologyModificationUtil. (Not sure if it’s the right place though). Now the loadOrderPosition must be provided.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/ModificationUtils.java
```diff
@@ -0,0 +1,36 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * @author Coline Piloquet <coline.piloquet at rte-france.com>
+ */
+public final class ModificationUtils {
+
+    private static final org.slf4j.Logger LOGGER = LoggerFactory.getLogger(ModificationUtils.class);
+
+    public static void throwExceptionOrLogError(String message, String key, boolean throwException, Reporter reporter) {
```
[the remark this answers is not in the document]
colinepiloquet: About the logs and reporter, it is what we’ve been doing everywhere…

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/TopologyModificationUtils.java
```diff
@@ -198,6 +211,145 @@ static void createBusBreakerSwitches(String busId1, String middleBusId, String b
...
+                        break;
+                    case LOAD:
+                    case GENERATOR:
+                    case SHUNT_COMPENSATOR:
+                    case STATIC_VAR_COMPENSATOR:
+                    case HVDC_CONVERTER_STATION:
+                    case BATTERY:
+                    case DANGLING_LINE:
+                        order = position.getFeeder().getOrder();
+                        if (order.isPresent()) {
```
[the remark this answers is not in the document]
colinepiloquet: The code was deleted

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/TopologyModificationUtils.java
```diff
@@ -198,6 +211,145 @@ static void createBusBreakerSwitches(String busId1, String middleBusId, String b
...
+                position.getFeeder2().getOrder().ifPresent(orders::add);
+                if (position.getFeeder3() != null) {
+                    position.getFeeder3().getOrder().ifPresent(orders::add);
+                }
+            }
+        }
+    }
+
+    /**
+    * Utility method that associates the connectable id (precised by the terminal if needed) and its order inside
```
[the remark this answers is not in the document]
colinepiloquet: I deleted it.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoad.java
```diff
@@ -0,0 +1,265 @@
...
+            if (bbs == null) {
+                throwExceptionOrLogError(String.format("Bus bar section %s not found.", bbsId), "notFoundBusbarSection", throwException, reporter);
+                return;
+            }
+            if (bbs.getTerminal().getVoltageLevel() != voltageLevel) {
+                throwExceptionOrLogError(String.format("Bus bar section %s is not in voltageLevel %s", bbsId, voltageLevelId), "busbarSectionNotInVoltageLevel", throwException, reporter);
+                return;
+            }
+        }
+        if (bbs == null) {
```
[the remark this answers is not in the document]
colinepiloquet: Is it?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoad.java
sylvlecl: Not sure, but maybe CreateFeederBay ? To be a little more specific. For example maybe we will want to have a CreateBusBarCouplerBay which will be different.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoad.java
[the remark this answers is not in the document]
flo-dup: Yes, `bbs` cannot be null if going through `if (bbsId != null)` above: then we're sure that `bbs != null` as it is checked within and in that case `return` is called.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoad.java
geofjamg: I'm afraid that for making it serializable we will need to somehow duplicate adders with simple pojos not dependent of iidm objects. Let's see that in next pr as it is an big work for having it for all equipments.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachLoad.java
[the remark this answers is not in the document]
flo-dup: In fact, if we want to be consistent with what was done with the enum `PositionInsideSection`, I'm wondering if we shouldn't also move that part in the `TopologyModificationUtils`. Meaning that the `voltageLevelId` would not be anymore in the parameters and that `bbsId` should therefore always be provided. What do you think?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/CreateFeederBay.java
```diff
@@ -0,0 +1,297 @@
...
+                    .add();
+        }
+
+        // create switches and a breaker linking the injection to the bus bar sections.
+        createTopology(network, voltageLevel, injectionNode, forkNode, injectionId, reporter);
+
+        LOGGER.info("New injection {} was added to voltage level {} on busbar section {}", injection.getId(), voltageLevel.getId(), bbs.getId());
+        reporter.report(Report.builder()
+                .withKey("newInjectionAdded")
+                .withDefaultMessage("New injection ${injectionId} was added to voltage level ${voltageLevelId} on busbar section ${bbsId}")
```
flo-dup, line 240: The message could be more explicit, that is describing whether it is connected to other busbar sections

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/TopologyModificationUtils.java
```diff
@@ -198,6 +211,90 @@ static void createBusBreakerSwitches(String busId1, String middleBusId, String b
...
+            }
+        }
+    }
+
+    public enum PositionInsideSection {
+        FIRST,
+        LAST
+    }
+
+    public static int getOrderPosition(PositionInsideSection positionInsideSection, VoltageLevel voltageLevel, BusbarSection bbs) {
```
flo-dup, line 288: Is the enum still that useful?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/TopologyModificationUtils.java
```diff
@@ -198,6 +211,90 @@ static void createBusBreakerSwitches(String busId1, String middleBusId, String b
...
+        Map<Integer, List<Integer>> allOrders = getSliceOrdersMap(voltageLevel);
+        BusbarSectionPosition busbarSectionPosition = bbs.getExtension(BusbarSectionPosition.class);
```
flo-dup, line 290: busbarSectionPosition can be null

flo-dup (changes requested):

flo-dup (approved):
This seems ok to me as a first version.

There are still some issues which need to be addressed,but I'd say let's do that in other PRs. The issues being:
- serialization of the hypothesis
- the design flaw that @geofjamg pointed out (currently we add then remove the created injection if we detect that it's from another network)
- to a lesser extent the missing add in InjectionAdder

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/CreateFeederBay.java
```diff
@@ -0,0 +1,183 @@
...
+        this.injectionPositionOrder = injectionPositionOrder;
+        this.injectionDirection = Objects.requireNonNull(injectionDirection);
+    }
+
+    public CreateFeederBay(InjectionAdder<?> injectionAdder, String bbsId, int injectionPositionOrder) {
+        this(injectionAdder, bbsId, injectionPositionOrder, ConnectablePosition.Direction.BOTTOM);
+    }
+
+    @Override
+    public void apply(Network network, ComputationManager computationManager) {
```
geofjamg, line 64: You have to move these 2 apply methods in the NetworkModification interface otherwise the added throwException is a breaking change

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/TopologyModificationUtils.java
```diff
@@ -198,6 +212,191 @@ static void createBusBreakerSwitches(String busId1, String middleBusId, String b
...
+        connectablesBySectionIndex.forEach((sectionIndex, connectables) -> {
+            List<Integer> orders = new ArrayList<>();
+            connectables.forEach(connectable -> addOrder(connectable, voltageLevel, orders));
+            ordersBySectionIndex.put(sectionIndex, orders);
+        });
+
+        return ordersBySectionIndex;
+    }
+
+    static void fillConnectablesMap(BusbarSection bbs, Map<BusbarSection, Set<Connectable<?>>> connectablesByBbs) {
```
annetill, line 248: Can you add a detailed javadoc for each utility method?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/CreateFeederBay.java
annetill: @colinepiloquet can you take this comment into account?

colinepiloquet: What do you mean? Aren't they already in the interface?

annetill: @geofjamg For all implementations, we have this override:
````
    @Override
    public void apply(Network network, ComputationManager computationManager) {
        apply(network);
    }
````
So if we want to put this default behavior in the interface, it should be done in another PR.
The other override is needed because this hypothesis does not implement the apply(network, computationManager).
````
    @Override
    public void apply(Network network) {
        apply(network, false, Reporter.NO_OP);
    }
````

File: iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/ModificationReports.java [hunk 20]
```diff
@@ -0,0 +1,93 @@
...
+import com.powsybl.commons.reporter.Reporter;
+import com.powsybl.commons.reporter.TypedValue;
+import com.powsybl.iidm.network.BusbarSection;
+import com.powsybl.iidm.network.IdentifiableType;
+import com.powsybl.iidm.network.Injection;
+
+/**
+ * @author Coline Piloquet <coline.piloquet at rte-france.com>
+ */
+public final class ModificationReports {
```
miovd, line 19: It seems to be really specific to topology modifications, maybe it should be in topology(/util?) and be package private rather than public

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/NetworkXml.java
```diff
@@ -135,7 +135,7 @@ private static void writeExtensionNamespaces(Network n, ExportOptions options, X
                         .stream()
                         .filter(e -> canTheExtensionBeWritten(getExtensionXmlSerializer(options, e), networkVersion, options))
                         .map(extension -> (ExtensionXmlSerializer<?, ?>) getExtensionXmlSerializer(options, extension)))
-                .collect(Collectors.toSet());
+                .collect(Collectors.toCollection(LinkedHashSet::new));
```
miovd, line 138: Is this modification linked to the PR? If it is for tests (order), you should condition this to the parameter "sorted" in XML export.

miovd (commented):
Mostly agree with the PR, just two questions :)

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/TopologyModificationUtils.java
```diff
@@ -198,6 +212,224 @@ static void createBusBreakerSwitches(String busId1, String middleBusId, String b
...
+                    Optional<Integer> order = position.getFeeder1().getOrder();
+                    order.ifPresent(feederPositionsOrders::add);
+                } else if (twt.getLeg2().getTerminal().getVoltageLevel() == voltageLevel) {
+                    Optional<Integer> order = position.getFeeder2().getOrder();
+                    order.ifPresent(feederPositionsOrders::add);
+                } else if (twt.getLeg3().getTerminal().getVoltageLevel() == voltageLevel) {
+                    Optional<Integer> order = position.getFeeder3().getOrder();
+                    order.ifPresent(feederPositionsOrders::add);
+                } else {
+                    throw new AssertionError();
```
miovd, line 402: Add log / report and take `throwException` into account

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/TopologyModificationUtils.java
```diff
@@ -198,6 +212,224 @@ static void createBusBreakerSwitches(String busId1, String middleBusId, String b
...
+            } else if (connectable instanceof Branch) {
+                Branch<?> branch = (Branch<?>) connectable;
+                if (branch.getTerminal1().getVoltageLevel() == voltageLevel) {
+                    Optional<Integer> order = position.getFeeder1().getOrder();
+                    order.ifPresent(feederPositionsOrders::add);
+                } else if (branch.getTerminal2().getVoltageLevel() == voltageLevel) {
+                    Optional<Integer> order = position.getFeeder2().getOrder();
+                    order.ifPresent(feederPositionsOrders::add);
+                } else {
+                    throw new AssertionError();
```
miovd, line 388: Same remark

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/NetworkModification.java
```diff
@@ -28,4 +28,8 @@ default void apply(Network network, ComputationManager computationManager, Repor
     default void apply(Network network, Reporter reporter) {
         apply(network);
     }
+
+    default void apply(Network network, boolean throwException, Reporter reporter) {
```
miovd, line 32: Add javadoc to explain what `throwException` does

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/ModificationReports.java
```diff
@@ -0,0 +1,101 @@
...
+import com.powsybl.commons.reporter.TypedValue;
+import com.powsybl.iidm.network.*;
+
+/**
+ * @author Coline Piloquet <coline.piloquet at rte-france.com>
+ */
+final class ModificationReports {
+    static String voltageLevelIdString = "voltageLevelId";
+
+    public static void notFoundBusbarSectionReport(Reporter reporter, String bbsId) {
```
miovd, line 20: Should these methods be public or is package-private enough?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/CreateFeederBay.java
```diff
@@ -0,0 +1,190 @@
...
+    @Override
+    public void apply(Network network) {
+        apply(network, false, Reporter.NO_OP);
+    }
+
+    /**
+     * Applies the modification to the given network. If throwException is set to true, then in case of error, an
+     * exception will be thrown. Otherwise, computation will continue but the injection will not be added to the network
+     * in case of error.
+     */
```
miovd, line 79: This javadoc is valid for all modifications, you should move it in the interface

miovd (commented):
Two small modifications and it is okay for me

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/ModificationReports.java
colinepiloquet: At least for now, I guess package-private is enough

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/NetworkXml.java
colinepiloquet: Actually it's something @flo-dup did. I think it must be needed.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/CreateFeederBay.java
```diff
@@ -0,0 +1,189 @@
...
+            notFoundBusbarSectionReport(reporter, bbsId);
+            if (throwException) {
+                throw new PowsyblException(String.format("Busbar section %s not found.", bbsId));
+            }
+            return;
+        }
+
+        VoltageLevel voltageLevel = bbs.getTerminal().getVoltageLevel();
+        TopologyKind topologyKind = voltageLevel.getTopologyKind();
+        if (topologyKind != TopologyKind.NODE_BREAKER) {
```
colinepiloquet, line 89: Am I wrong or we can remove this check? A busbar section makes no sense if the voltage level is not described at node/breaker level?

geofjamg (approved):

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/NetworkXml.java
flo-dup: This modification came from PR #2231, it appeared there only because the merge from main hadn't been done.

PR #2216, opened 2022-07-15
Allow boundaries to be assembled with other network data in the same instance file

### LINKED ISSUE #2215: CGMES assembled model import support
feature

We need to provide a zip file with all profiles, to import a CGMES.

We should also support a single file loading containing the assembled model.

  - PowSyBl Version: ...
  - OS Version: ...

zamarrenolm:

Closes #2215.

### DIFF
--- a/cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/triplestore/CgmesModelTripleStore.java
+++ b/cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/triplestore/CgmesModelTripleStore.java
@@ -221,9 +221,8 @@ public class CgmesModelTripleStore extends AbstractCgmesModel {
                 updateModelHasOperationProfile(modelHasOperationProfile, modelHasBoundaryOperationProfile, m, p);
             }
         }
-        Map<String, Boolean> modelHasOperationProfile1 = new HashMap<>(modelHasOperationProfile);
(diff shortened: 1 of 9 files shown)


cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/triplestore/CgmesModelTripleStore.java
```diff
@@ -209,24 +209,40 @@ private void logNodeBreaker(boolean consideredNodeBreaker, Map<String, Boolean>
...
-                    }
-                }
+                updateModelHasOperationProfile(modelHasOperationProfile, modelHasBoundaryOperationProfile, m, p);
             }
         }
-        return modelHasOperationProfile;
+        Map<String, Boolean> modelHasOperationProfile1 = new HashMap<>(modelHasOperationProfile);
+        modelHasBoundaryOperationProfile.forEach((m, v) -> modelHasOperationProfile1.merge(m, v, (vm, vbd) -> vm && vbd));
+        return modelHasOperationProfile1;
+    }
```
marqueslanauja, line 227: I think we do not need  `modelHasOperationProfile1`, we can merge over `modelHasOperationProfile`

zamarrenolm: you are right

marqueslanauja (approved):

marcosmc (approved):

miovd (approved):

PR #2218, opened 2022-07-21
Proposition for network modification builder

miovd:

No

Improvement

A constructor (sometimes not really user-friendly) is used to create a network modification.

We can add a builder for each modification as proposed. If the design is agreed upon, it can be implemented for other network modifications.

**Other information**:

### DIFF
--- a/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachNewLineOnLineBuilder.java
+++ b/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachNewLineOnLineBuilder.java
@@ -1,0 +1,124 @@
+/**
+ * Copyright (c) 2022, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.iidm.modification.topology;
+
+import com.powsybl.iidm.network.Line;
+import com.powsybl.iidm.network.LineAdder;
+
+/**
+ * @author Miora Vedelago <miora.ralambotiana at rte-france.com>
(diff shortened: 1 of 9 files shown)


geofjamg:
Good idea.
Also theses modifications are really badly named. I would suggest to rename AttachNewLineOnLine to InsertLineIntoLine and AttachVoltageLevelOnLine to InsertVoltageLevelIntoLine

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachNewLineOnLine.java
```diff
@@ -54,7 +54,9 @@ public class AttachNewLineOnLine implements NetworkModification {
      * line2Id is line.getId() + "_2". <br>
      *
      * @see #AttachNewLineOnLine(double, String, String, String, String, String, Line, LineAdder)
+     * @deprecated Use {@link AttachNewLineOnLineBuilder} instead.
      */
+    @Deprecated(since = "4.10.0")
```
flo-dup, line 59: Just wondering if deprecated is the appropriate term / tag, as we will just change the access to protected (and not really delete it one day).

flo-dup (approved):
Indeed, nice to use a builder instead

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AttachNewLineOnLine.java
miovd: Not sure if this is appropriate indeed! I assumed this javadoc was only useful from a user standpoint (not a developer one) but maybe I am wrong. What do you suggest?

miovd: I deprecated all constructors except for those used in the builders and just added in the javadoc that they will eventually be package private. Do you agree?

PR #2236, opened 2022-08-10
Improve dense matrix multiplication and addition performance

geofjamg:

No

Performance improvement

Dense matrix multiplication and addition rely on Jama and so on need an internal data model conversion. This is slow.

Dense matrix multiplication and addition have been reimplemented on internal representation.

**Other information**:


math/src/main/java/com/powsybl/math/matrix/DenseMatrix.java
```diff
@@ -233,12 +237,54 @@ public LUDecomposition decomposeLU() {
...
+                for (int k = 0; k < columnCount; k++) {
+                    s += getUnsafe(i, k) * otherColumnJ[k];
+                }
+                result.setUnsafe(i, j, s);
+            }
+        }
+
+        return result;
+    }
+
```
JB-H, line 267: In the same way as add(), wouldn't it be interesting to have more general time() function such as times(DenseMatrix other, Double scalar) to speed up operations like [A] * [B] * scalar ?
Would it be relevant to have a time(Double scalar) function as well?

geofjamg: [A] * [B] * scalar added

JB-H (approved):
Agree with evolutions

Issue #2172, opened 2022-06-16
Implementation of Redispatching of generation action


Hadrien-Godard:
* **Do you want to request a *feature* or report a *bug*?**
Feature.

* **What is the current behavior?**
The action is not implemented.

* **What is the expected behavior?**
The Redispatching of generation action is implemented as an implementation of the Action class (see issue #2015).
This action consists in increasing or decreasing the active power generation of a given generator.
A starting idea of possible implementation is:
```java
class RedispatchGenerationAction extends Action {
    String generatorId;
    boolean increase;
    double value;
}
```

* **What is the motivation / use case for changing the behavior?**
Allow security analysis to take this action type into account.

PR #2237, opened 2022-08-16
Generator action

Hadrien-Godard:

Yes issue #2172 

**Other information**:


security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/GenerationRedispatchAction.java
```diff
@@ -0,0 +1,51 @@
...
+package com.powsybl.security.action;
+
+import java.util.Objects;
+
+/**
+ * An action increasing or decreasing the active power generation of a generator.
+ *
+ * @author Hadrien Godard <hadrien.godard@artelys.com>
+ */
+public class GenerationRedispatchAction extends AbstractAction {
```
annetill, line 16: Why is it a redispatching? It will be a redispatching if it involves at least two generators, no? Maybe we have to be generic with something like `GeneratorAction`, and a redispatch action will be a list of `GeneratorAction`. What do you think @sylvlecl ?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/GenerationRedispatchAction.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public static final String NAME = "GENERATION_REDISPATCH";
+
+    private final String generatorId;
+    private final boolean increasing;
```
annetill, line 21: The boolean `increasing` is not needed as you have the `value`. What we should know is that if  `value` is the expected value for `targetP` or if `value` represents a relative variation. So maybe we have to add a boolean that explicits that ('delta' ?).

annetill (changes requested):

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/GeneratorAction.java
```diff
@@ -0,0 +1,91 @@
...
+
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * An action to:
+ * - increase or decrease the targetV of a generator.
+ * - move to voltage control or remove from voltage control.
+ * - change targetQ of a generator that is not a voltage controller.
+ * - change targetV of a generator that is a voltage controller.
```
jeandemanged, line 17: targetV and targetQ can always be changed, independently of voltage control on/off status I think.

voltage control is called voltage regulator in powsybl

```suggestion
 * An action to:
 * - change the targetP of a generator, either by specifying a new absolute value, or a relative change (MW)
 * - enable or disable voltage regulator
 * - change targetQ of a generator (MVAr)
 * - change targetV of a generator (kV)
```

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/GeneratorAction.java
```diff
@@ -0,0 +1,91 @@
...
+ * @author Hadrien Godard <hadrien.godard@artelys.com>
+ * @author Anne Tilloy <anne.tilloy@rte-france.com>
+ */
+public class GeneratorAction extends AbstractAction {
+
+    public static final String NAME = "GENERATOR";
+
+    private final String generatorId;
+    private Boolean activePowerRelativeValue; // true if it is a relative variation, false if it is a new targetP
+    private Double activePowerValue; // could be a new targetP if relativeVariation equals false or a relative variation of targetP.
```
jeandemanged, line 28: activePowerRelative**Value** is misleading (could be understood as the relative delta **MW value** if not carefully looking at the type)

```suggestion
    private Boolean activePowerRelative; // true if it is a relative variation, false if it is a new targetP
    private Double activePowerValue; // new targetP if activePowerRelative is false, otherwise a relative change of current targetP value.
```

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/GeneratorAction.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    public Optional<Double> getActivePowerValue() {
+        return Optional.ofNullable(activePowerValue);
+    }
+
+    public GeneratorAction setActivePowerValue(Double activePowerValue) {
+        this.activePowerValue = activePowerValue;
+        return this;
+    }
```
jeandemanged, line 63: I see a risk of API users setting activePowerValue but forgetting to set if it is relative or absolute.

Maybe we should provide a single api as follows to force users specifying both values, using primitive types (non nullable), and delete above setActivePowerRelativeValue method.

```suggestion
    public GeneratorAction setActivePowerValue(double activePowerValue, boolean relative) {
        this.activePowerValue = activePowerValue;
        this.activePowerRelative= relative;
        return this;
    }
    
    public GeneratorAction unsetActivePowerValue() {
        this.activePowerValue = null;
        this.activePowerRelative= null;
        return this;
    }
```

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/GeneratorAction.java
```diff
@@ -0,0 +1,91 @@
...
+ * An action to:
+ * - increase or decrease the targetV of a generator.
+ * - move to voltage control or remove from voltage control.
+ * - change targetQ of a generator that is not a voltage controller.
+ * - change targetV of a generator that is a voltage controller.
+ *
+ * @author Hadrien Godard <hadrien.godard@artelys.com>
+ * @author Anne Tilloy <anne.tilloy@rte-france.com>
+ */
+public class GeneratorAction extends AbstractAction {
```
jeandemanged, line 22: for the setter methods, I would avoid boxed types and use only primitive types in arguments.

If needed to unset some attributes I think API users would prefer to have unset methods instead of calling set with null.

jeandemanged (changes requested):
on functional point of view I approve, only area of improvements I see: javadoc could be made clearer, API could be enhanced to be more API-user friendly (see suggested changes)

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/ActionDeserializer.java
```diff
@@ -122,6 +151,24 @@ public Action deserialize(JsonParser parser, DeserializationContext deserializat
...
                     return new PhaseTapChangerTapPositionAction(context.id, context.transformerId, context.relativeValue, context.value);
                 }
+            case GeneratorAction.NAME:
+                GeneratorAction generatorAction = new GeneratorAction(context.id, context.generatorId);
+                if (context.activePowerRelativeValue != null) {
+                    generatorAction.setActivePowerRelativeValue(context.activePowerRelativeValue);
+                }
+                if (context.activePowerValue != null) {
+                    generatorAction.setActivePowerValue(context.activePowerValue);
+                }
```
jeandemanged, line 161: both or none should be provided
```suggestion
                if ((context.activePowerRelativeValue != null) && (context.activePowerValue != null)) {
                    generatorAction.set... //TODO see other suggested changes
                } else if ((context.activePowerRelativeValue != null) ^ (context.activePowerValue != null)) {
                throw JsonMappingException.from(parser, "for generator action, both or none of these two attributes must be provided: activePowerValue, activePowerRelative");
                }
```

jeandemanged (changes requested):

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/GeneratorAction.java
annetill: Yes of course, but with no impact on a power flow engine. So the issue in the documentation here? Voltage regulator: yes I know but I don't like this choice :-)

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/GeneratorAction.java
annetill: Ok, thanks. I have chosen `activePowerRelativeValue` to remain coherent but already available action (`PhaseTapChangerTapPositionAction`). So what do you think?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/GeneratorAction.java
annetill: Great idea.

sylvlecl: Optionals are generally used for return types (to allow method chaining), not input types: it's considered better to have primitive or boxed types as inputs/class fields.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/GeneratorAction.java
sylvlecl: Actually, another design concern:
I would advise to not have setters at all in `Action` implementations:
from the point of view of the simulator, they should be immutable objects, the simulator is not supposed to change them.

If constructors (or static methods) are not enough and we want a fluent API maybe we should have a builder class ? (in which the comment above about the setter method applies :) )

jeandemanged: hi, not so sure about making these objects immutable. In simulators e.g. openloadflow indeed the simulator should not modify the actions.

but the same action objects defined by powsybl-core could be used to e.g. build a GUI, and the user might want to modify an action instead of having to delete the existing and create a new one.

sylvlecl: I understand the use case but in that case I think the best design will be to have a separate class in an application layer (where there might be some additional app-specific logic if needed), here the objects are really meant to be security analysis inputs: it's very important that the API does not allow the simulator to change them, or even other external components to change them during the execution of the simulator.
Otherwise, it will open doors for tricky bugs :)

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/GeneratorAction.java
jeandemanged: yes issue here is only documentation

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/GeneratorAction.java
jeandemanged: ok we can leave this as it is

jeandemanged (approved):

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/GeneratorAction.java
```diff
@@ -0,0 +1,82 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.security.action;
+
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * An action to:
+ * - change the targetP of a generator, either by specifying a new absolute value (MW) or a relative change (MW).
```
sylvlecl, line 14: Note that the "markdown style" bullet points do not render well in javadoc (including IDE):
we should use HTML marks instead (\<ul> for the list and \<li> for list items)

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/GeneratorAction.java
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * @author Hadrien Godard <hadrien.godard@artelys.com>
+ * @author Anne Tilloy <anne.tilloy@rte-france.com>
+ */
+public class GeneratorAction extends AbstractAction {
+
+    public static final String NAME = "GENERATOR";
+
+    private final String generatorId;
+    private final Boolean activePowerRelativeValue; // true if it is a relative variation of targetP, false if it is a new targetP
```
sylvlecl, line 27: the comments would have more value as javadoc (note that it's even more important for those classe which do not implement the behaviour, but only specify them)

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/action/ActionSerializer.java
```diff
@@ -1,66 +0,0 @@
...
- * @author Etienne Lesot <etienne.lesot@rte-france.com>
- */
-public class ActionSerializer extends StdSerializer<Action> {
-
-    public ActionSerializer() {
-        super(Action.class);
-    }
-
-    @Override
-    public void serialize(Action action, JsonGenerator jsonGenerator, SerializerProvider serializerProvider) throws IOException {
```
sylvlecl, line 28: Thanks for the removal :)

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/action/GeneratorActionDeserializer.java
```diff
@@ -0,0 +1,99 @@
...
+                    return true;
+                case "targetQ":
+                    parser.nextToken();
+                    context.targetQ = parser.getValueAsDouble();
+                    return true;
+                default:
+                    return false;
+            }
+        });
+        GeneratorActionBuilder generatorActionBuilder = new GeneratorActionBuilder();
```
sylvlecl, line 78: not important but we could have use the builder straight during the parsing, saving the creation of the parsing context object

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/GeneratorActionBuilder.java
```diff
@@ -0,0 +1,60 @@
...
+    private String id;
+    private String generatorId;
+    private Boolean activePowerRelativeValue;
+    private Double activePowerValue;
+    private Boolean voltageRegulatorOn;
+    private Double targetV;
+    private Double targetQ;
+
+    public GeneratorAction build() {
+        if (activePowerRelativeValue != null ^ activePowerValue != null) {
```
sylvlecl, line 20: I didn't even know that operator :)

sylvlecl (approved):
Minor comments, nothing blocking the merge

annetill (approved):

PR #2246, opened 2022-08-19
Extension adder: change returns type of add method

geofjamg:

No

add method of extension adder returns the adder instead of the added object.

add method of extension adder now returns the added extension.

**Other information**:


miovd (approved):

miovd:
Should we handle backward compatibility for groovy? Not sure it is used in scripts though... To be discussed

geofjamg:
> Should we handle backward compatibility for groovy? Not sure it is used in scripts though... To be discussed

I was not used in Java, at least on powybl core there was nowhere such code:
```java
gen.newExtension(TotoAdder.class)
  .add()
  .newExtension(TutuAdder.class)
  .add()
```
So it is very unlikely that there is somewhere such code in Groovy scripts.

PR #2257, opened 2022-09-05
Create Hypothesis for creation of branch feeders

miovd:

Feature: same hypothesis as `CreateFeederBay` but for branches (not injection).

No

**Other information**:

### DIFF
--- a/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeeders.java
+++ b/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeeders.java
@@ -9,7 +9,7 @@ package com.powsybl.iidm.modification.topology;
 import com.powsybl.commons.PowsyblException;
 import com.powsybl.commons.reporter.Reporter;
 import com.powsybl.computation.ComputationManager;
-import com.powsybl.iidm.modification.NetworkModification;
+import com.powsybl.iidm.modification.AbstractNetworkModification;
 import com.powsybl.iidm.network.*;
 import com.powsybl.iidm.network.extensions.BusbarSectionPosition;
 import com.powsybl.iidm.network.extensions.ConnectablePosition;
@@ -30,7 +30,7 @@ import static com.powsybl.iidm.modification.topology.TopologyModificationUtils.c
 /**
  * @author Miora Vedelago <miora.ralambotiana at rte-france.com>
  */
-abstract class AbstractCreateConnectableFeeders implements NetworkModification {
+abstract class AbstractCreateConnectableFeeders extends AbstractNetworkModification {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AbstractCreateConnectableFeeders.class);
 
@@ -57,22 +57,7 @@ abstract class AbstractCreateConnectableFeeders implements NetworkModification {
     }
 
     @Override
-    public void apply(Network network, ComputationManager computationManager) {
-        apply(network, false, Reporter.NO_OP);
-    }
-
-    @Override
-    public void apply(Network network, ComputationManager computationManager, Reporter reporter) {
-        apply(network, false, reporter);
-    }
-
-    @Override
-    public void apply(Network network, Reporter reporter) {
-        apply(network, false, reporter);
-    }
-
-    @Override
-    public void apply(Network network, boolean throwException, Reporter reporter) {
+    public void apply(Network network, boolean throwException, ComputationManager computationManager, Reporter reporter) {
         for (int side : sides) {
             String bbsId = getBbsId(side);
             BusbarSection bbs = network.getBusbarSection(bbsId); //If the busbar exists, topology of the associated voltage level is node/breaker
--- a/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeeders.java
+++ b/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeeders.java
@@ -38,7 +38,7 @@ abstract class AbstractCreateConnectableFeeders implements NetworkModification {
 
     protected abstract String getBbsId(int side);
 
-    protected abstract void setNode(int side, int node);
+    protected abstract void setNode(int side, int node, String voltageLevelId);
 
     protected abstract Connectable<?> add();
 
@@ -49,6 +49,8 @@ abstract class AbstractCreateConnectableFeeders implements NetworkModification {
     protected abstract ConnectablePosition.Direction getDirection(int side);
 
     protected abstract int getNode(int side, Connectable<?> connectable);
+
+    protected abstract ConnectablePositionAdder.FeederAdder<?> getFeederAdder(int side, ConnectablePositionAdder<?> connectablePositionAdder);
 
     protected AbstractCreateConnectableFeeders(int... sides) {
         this.sides = Arrays.copyOf(sides, sides.length);
@@ -84,7 +86,7 @@ abstract class AbstractCreateConnectableFeeders implements NetworkModification {
             }
             VoltageLevel voltageLevel = bbs.getTerminal().getVoltageLevel();
             int injectionNode = voltageLevel.getNodeBreakerView().getMaximumNodeIndex() + 1;
-            setNode(side, injectionNode);
+            setNode(side, injectionNode, voltageLevel.getId());
         }
 
         Connectable<?> connectable = add();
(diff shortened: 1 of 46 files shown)


iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeeders.java
```diff
@@ -0,0 +1,153 @@
...
+            }
+            VoltageLevel voltageLevel = bbs.getTerminal().getVoltageLevel();
+            int injectionNode = voltageLevel.getNodeBreakerView().getMaximumNodeIndex() + 1;
+            setNode(side, injectionNode, voltageLevel.getId());
+        }
+
+        Connectable<?> connectable = add();
+        if (connectable.getNetwork() != network) {
+            connectable.remove();
+            LOGGER.error("Network given in parameters and in injectionAdder are different. Injection '{}' of type {} was added then removed",
```
colinepiloquet, line 80: The string is not adapted to branches here

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/CreateBranchFeedersBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+ * @author Miora Vedelago <miora.ralambotiana at rte-france.com>
+ */
+public class CreateBranchFeedersBuilder {
+
+    private BranchAdder<?> branchAdder = null;
+    private String bbsId1 = null;
+    private String bbsId2 = null;
+    private Integer positionOrder1 = null;
+    private Integer positionOrder2 = null;
+    private ConnectablePosition.Direction direction1 = ConnectablePosition.Direction.BOTTOM;
```
annetill, line 22: default value UP is better.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeeders.java
```diff
@@ -0,0 +1,153 @@
...
+                    return;
+                }
+                getFeederAdder(side, connectablePositionAdder)
+                        .withDirection(getDirection(side))
+                        .withOrder(positionOrder)
+                        .withName(connectableId)
+                        .add();
+                createConnectablePosition = true;
+            } else {
+                LOGGER.warn("No extensions found on voltageLevel {}. The extension is not created.", voltageLevel.getId());
```
flo-dup, line 112: In fact it means two things here:
- either there are feeders but no extensions on them, then not adding the extension might be ok (I would add it anyway, but that was already discussed)
- or there are no feeders hence no extensions; for instance you're creating a voltage level from scratch and you first add the busbar, then the line(s), then the other feeders. In that case we should add the extension anyway, don't you think so? To that purpose, we could just check that `voltageLevel.getConnectables()` is empty.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeeders.java
```diff
@@ -0,0 +1,153 @@
...
+                noConnectablePositionExtension(reporter, voltageLevel);
+            }
+            // create switches and a breaker linking the connectable to the busbar sections.
+            int node = getNode(side, connectable);
+            createTopology(side, network, voltageLevel, node, node + 1, connectable, reporter);
```
flo-dup, line 117: here we suppose that the node used for the connectable has a spot available just after. We could be more robust by just doing `int forkNode = voltageLevel.getNodeBreakerView().getMaximumNodeIndex() + 1` inside `createTopology`.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/CreateBranchFeeders.java
```diff
@@ -0,0 +1,138 @@
...
+    private final ConnectablePosition.Direction direction1;
+    private final ConnectablePosition.Direction direction2;
+
+    CreateBranchFeeders(BranchAdder<?> branchAdder, String bbsId1, String bbsId2, Integer positionOrder1, Integer positionOrder2,
+                        ConnectablePosition.Direction direction1, ConnectablePosition.Direction direction2) {
+        super(1, 2);
+        this.branchAdder = Objects.requireNonNull(branchAdder);
+        this.bbsId1 = Objects.requireNonNull(bbsId1);
+        this.bbsId2 = Objects.requireNonNull(bbsId2);
+        this.positionOrder1 = Objects.requireNonNull(positionOrder1);
```
flo-dup, line 39: The constructor takes an `Integer`, but it cannot be null, as in `CreateFeederBay` in fact. It's because of the builder I suppose, but shouldn't the `Objects.requireNonNull` be in the builder then?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/CreateBranchFeeders.java
```diff
@@ -0,0 +1,138 @@
...
+
+/**
+ * This class allows to add a new branch's feeders on existing busbar sections. The voltage level containing the
+ * busbar section should be described in node/breaker topology. The branch is created and connected to
+ * the busbar sections with a breaker and a closed disconnector each. The branch is also connected to all
+ * the parallel busbar sections, if any, with an open disconnector.
+ *
+ * @author Miora Vedelago <miora.ralambotiana at rte-france.com>
+ */
+public class CreateBranchFeeders extends AbstractCreateConnectableFeeders {
```
flo-dup, line 23: As we are creating bays also on each side, shouldn't the keyword `Bay` be in class name?

flo-dup (changes requested):
Nice refactor, now coding the 3wt will be a piece of cake! Just a couple of comments/questions.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeeders.java
miovd: The node used is already `int connectableNode = voltageLevel.getNodeBreakerView().getMaximumNodeIndex() + 1;` so the spot just after is always available, this is why we only added 1!

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/CreateBranchFeeders.java
miovd: Yes, it is because of the builder. I put all the checks for homogenization in the constructor (none of the builders have checks so far) but this can be changed, I don't really have any preference. I would prefer however have all the checks in one place, would you rather have them all in the builder (when calling `build()`) or in the constructor?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeeders.java
miovd: I followed your note, only checking if getConnectables() contains one and only one element (the connectable we've just created)

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeeders.java
flo-dup: Yes, now it's ok, but if one day you add another node between `setNode` and `getNode`, then it's not working anymore. That was what I meant by "more robust".

flo-dup: Which makes me think about the special case of a branch with both side in the same voltage level! The fork nodes would then be `getNode(side, connectable) + 2`.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/CreateBranchFeeders.java
flo-dup: In fact I was just thinking about a user which would use `CreateBranchFeeders` instead of the builder, then it's strange to ask for an `Integer`. But I missed the fact that such a user shouldn't exist as it's package-private.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeeders.java
miovd: Ah yes, you're right... Changing it

flo-dup: you probably got it, but it seems that for that special case, the current code in the setNode loop results in the same node for both sides.

miovd: Indeed... But I don't think this case ever happens? Theoretically it is allowed but I don't think it does.

A quick hack can be to use side even if it create holes:
`int connectableNode = voltageLevel.getNodeBreakerView().getMaximumNodeIndex() + (side == 0 ? 1 : side);`

What do you suggest?

miovd: (To be clear, I don't like the hack I proposed but I'm not sure about any solution that would not be too cost-heavy)

miovd: Done what you suggested! Feel free to approve if the PR is now okay with you

flo-dup (approved):

Issue #2232, opened 2022-08-08
New contingency list models


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature

* **What is the current behavior?**

Only one implementation of `ContingencyList` is proposed by powsybl-core, if we exclude the groovy implementation.

Therefore, we cannot represent as a POJO or as JSON such things as "automatic" list, for example the list of all 400kV lines.

* **What is the expected behavior?**

The following implementations would be helpful:

**1. Multi-list**
Simply a list of lists.

**2. Criterion-based implementation**
A list based on some filters:
- List of countries to be included
- List of element types to be included
- List of regex on the element IDs, to be included
- List of regex on the element IDs, to be excluded
- List of nominal voltages (or nominal voltage ranges) to be included

**3. Multi-identifiers explicit list**

In some use cases, we want to use the same list on networks coming from different sources/formats.
In order to achieve this, we need to be able to specify alternative identifiers for the same network element.
Plus, the identifier cannot always be a simple string: sometimes for a branch we only have the 2 voltage levels and the order.
Therefore we could use something like:

```json
{
    "id": "multi-identifiers-contingency",
    "elements" : [
         {
             "type": "LINE",
             "identifiers": [
                  { "type": "id", "id": "LINE1" },
                  { "type": "voltage_levels_and_order", "vl1": "UCTE_VL1",  "vl2": "UCTE_VL2", "order": "1" },
             ]
         }
    ]
}
```
This describes a contingency on line "LINE1" if it's present in the network, or on the line of order code 1 between "UCTE_VL1" and "UCTE_VL2" else.

* **What is the motivation / use case for changing the behavior?**

We need to use the same list definition for various networks, with elements being added or removed or renamed from time to time, and from different data sources.

geofjamg:
We already have this kind of feature in GridSuite, but it has not been shared with PowSyBl. We should refactor what we have done to support what you described in PowSyBl.

sylvlecl:
About criterion-based lists, I wonder about the possibility of having a more re-usable approach:
defining a set of elements or filters of elements is something that could be used for many features, for example scalables etc.

We could have a common model for those sets/filters of network elements, not only related to contingencies definition, that we would re-use here.

We could maybe define a kind of mini-language (not groovy or introspection based) for those filters, that could be supported by some front-end components (think JIRA filters for example).

PR #2308, opened 2022-09-29
New contingency list models

EtienneLt:

#2232


contingency/contingency-api/src/main/java/com/powsybl/contingency/json/ContingencyListDeserializer.java
```diff
@@ -0,0 +1,48 @@
...
+        super(ContingencyList.class);
+    }
+
+    @Override
+    public ContingencyList deserialize(JsonParser parser, DeserializationContext deserializationContext) throws IOException {
+        parser.nextToken();
+        while (parser.getCurrentName() != null) {
+            if ("type".equals(parser.getCurrentName())) {
+                switch (parser.nextTextValue()) {
+                    case "default":
```
sylvlecl, line 31: We should probably try to leverage jackson mechanisms to automate the association name -> deserializer.
Otherwise we change a little the behaviour of the deserializers used below, because it will expect to have already parsed the beginning of the object, so if someone wants to use them to deserialize a plain "DefaultContingencyList" for example, it will not work.

If we want to customize this, I think the only way is to use the `JsonTypeInfo` annotation, and it's probably slightly better on a "mixin" object to not define it directly into the `ContingencyList` interface.

contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyListsList.java
```diff
@@ -0,0 +1,53 @@
...
+public class ContingencyListsList implements ContingencyList {
+
+    // VERSION = 1.0 : first version
+    public static final String VERSION = "1.0";
+    private final String name;
+    private final List<ContingencyList> contingencyLists;
+
+    public ContingencyListsList(String name, List<ContingencyList> contingencyLists) {
+        this.name = name;
+        this.contingencyLists = contingencyLists;
```
sylvlecl, line 26: Better to make an immutable copy here

contingency/contingency-api/src/main/java/com/powsybl/contingency/contingency/list/criterion/Criterion.java
```diff
@@ -0,0 +1,28 @@
...
+import com.powsybl.iidm.network.Identifiable;
+import com.powsybl.iidm.network.IdentifiableType;
+
+/**
+ * @author Etienne Lesot <etienne.lesot@rte-france.com>
+ */
+public interface Criterion {
+
+    enum CriterionType {
+        COUNTRY,
```
sylvlecl, line 18: We also need a criterion on a pattern (regex) for identifiers

contingency/contingency-api/src/main/java/com/powsybl/contingency/IdentifierContingencyList.java
```diff
@@ -0,0 +1,69 @@
...
+    private final String name;
+    private final IdentifiableType identifiableType;
+    private final List<Identifier> identifiers;
+
+    public IdentifierContingencyList(String name, String identifiableType, List<Identifier> identifiers) {
+        this(name, IdentifiableType.valueOf(identifiableType), identifiers);
+    }
+
+    public IdentifierContingencyList(String name, IdentifiableType identifiableType, List<Identifier> identifiers) {
+        this.name = name;
```
sylvlecl, line 34: non null checks + immutable copy

contingency/contingency-api/src/main/java/com/powsybl/contingency/contingency/list/criterion/CountryCriterion.java
```diff
@@ -0,0 +1,113 @@
...
+package com.powsybl.contingency.contingency.list.criterion;
+
+import com.powsybl.iidm.network.*;
+
+/**
+ * @author Etienne Lesot <etienne.lesot@rte-france.com>
+ */
+public class CountryCriterion implements Criterion {
+
+    private final Country country;
```
sylvlecl, line 16: Do we want to allow multiple countries?

contingency/contingency-api/src/main/java/com/powsybl/contingency/contingency/list/criterion/PropertyCriterion.java
```diff
@@ -0,0 +1,43 @@
...
+import com.powsybl.iidm.network.IdentifiableType;
+
+import java.util.Objects;
+
+/**
+ * @author Etienne Lesot <etienne.lesot@rte-france.com>
+ */
+public class PropertyCriterion implements Criterion {
+    private final String propertyKey;
+    private final String propertyValue;
```
sylvlecl, line 19: Do we want to allow multiple values?

contingency/contingency-api/src/main/java/com/powsybl/contingency/CriterionContingencyList.java
```diff
@@ -0,0 +1,70 @@
...
+ * @author Etienne Lesot <etienne.lesot@rte-france.com>
+ */
+public class CriterionContingencyList implements ContingencyList {
+
+    // VERSION = 1.0 : first version
+    public static final String VERSION = "1.0";
+
+    private final String name;
+    private final IdentifiableType identifiableType;
+    private final List<Criterion> criteria;
```
sylvlecl, line 29: After some thinking, I tend to think it will be better to hard code in the fields the expected criterions (instead of having an arbitrary list), because for example we will never want 2 criterions on countries, or 2 criterions on voltage.

@geofjamg : any opinion about this design ?

Note that it does not prevent to use a List<Criterion> to actually perform the filtering.

Then if we go for this, the next question was "do we want to have different implementations for different types", to be able to have well defined fields for each type. Same thing, I tend to think that yes, now ... But it means one implementation for each type of element, not any more one for all.

In that case, we could keep this one as an abstract class.

contingency/contingency-api/src/test/resources/contingencies.json
```diff
@@ -1,4 +1,5 @@
 {
+  "type" : "default",
```
sylvlecl, line 2: Note that we should try to ensure retro-compatibility, so it should work without this

contingency/contingency-api/src/main/java/com/powsybl/contingency/contingency/list/identifiant/UcteIdentifier.java
```diff
@@ -0,0 +1,70 @@
...
+            throw new IllegalArgumentException("order must be an integer > 0");
+        }
+        this.order = order;
+    }
+
+    @Override
+    public Contingency filterIdentifiable(Network network) {
+        VoltageLevel voltageLevel1 = network.getVoltageLevel(voltageLevelId1);
+        VoltageLevel voltageLevel2 = network.getVoltageLevel(voltageLevelId2);
+        if (voltageLevel1 == null || voltageLevel2 == null) {
```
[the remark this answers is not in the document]
sylvlecl: My mistake, I thought it was a check on IDs, but voltage levels can indeed be null (they don't exist in the network).

contingency/contingency-api/src/main/java/com/powsybl/contingency/contingency/list/identifiant/Identifier.java
```diff
@@ -0,0 +1,25 @@
...
+import com.powsybl.iidm.network.Network;
+
+/**
+ * @author Etienne Lesot <etienne.lesot@rte-france.com>
+ */
+public interface Identifier {
+    Contingency filterIdentifiable(Network network);
+
+    enum IdentifierType {
+        SIMPLE,
```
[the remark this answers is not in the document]
EtienneLt: ok for ID_BASED

contingency/contingency-api/src/main/java/com/powsybl/contingency/contingency/list/AbstractEquipmentCriterionContingencyList.java
```diff
@@ -0,0 +1,50 @@
...
+
+import com.powsybl.contingency.contingency.list.criterion.*;
+import com.powsybl.iidm.network.IdentifiableType;
+
+import java.util.Objects;
+
+/**
+ * @author Etienne Lesot <etienne.lesot@rte-france.com>
+ */
+public abstract class AbstractEquipmentCriterionContingencyList implements ContingencyList {
```
miovd, line 17: Can be package-private

miovd (commented):
Is it also possible to refactor the deserializer by having a  `AbstractEquipmentCriterionContingencyListDeserializer`?
The refactoring would be great but if it is not possible, it is okay for me.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Network.java
```diff
@@ -917,4 +917,8 @@ default Network setMinimumAcceptableValidationLevel(ValidationLevel validationLe
         }
         return this;
     }
+
+    default <I extends Identifiable<I>> Stream<I> getIdentifiableTypeStream(IdentifiableType identifiableType) {
```
geofjamg, line 921: getIdentifiableStream?

geofjamg (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Network.java
EtienneLt: ok

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Network.java
```diff
@@ -917,4 +917,8 @@ default Network setMinimumAcceptableValidationLevel(ValidationLevel validationLe
         }
         return this;
     }
+
+    default <I extends Identifiable<I>> Stream<I> getIdentifiableStream(IdentifiableType identifiableType) {
```
geofjamg, line 921: Maybe we could move the impl into the default method to keep API compat?

contingency/contingency-api/src/main/java/com/powsybl/contingency/contingency/list/criterion/PropertyCriterion.java
```diff
@@ -0,0 +1,45 @@
...
+import com.powsybl.iidm.network.Identifiable;
+import com.powsybl.iidm.network.IdentifiableType;
+
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * @author Etienne Lesot <etienne.lesot@rte-france.com>
+ */
+public class PropertyCriterion implements Criterion {
```
sylvlecl, line 19: As identified by gridsuite project, it's too restrictive to check properties only on the identifiable itself:
for example, we would like to be able to check the property on the substation containing the identifiable.

In order to achieve this, we could add an enum to that criterion with values "SELF / VOLTAGE_LEVEL / SUBSTATION".
Then at filter time, we will get the property either from the identifiable itself, or from the voltage level, or from the substation.

It also means that for multi-sided elements, we will need to allow multiple property criterions ... (2 for branches, 3 for 3WT)

geofjamg (approved):

ucte/ucte-converter/src/test/resources/sameId.uct
```diff
@@ -1,6 +1,6 @@
 ##C 2007.05.01
 Test case to check the import of a file with a line and a transformer with identical
-node-node-order "identifier".
+node-node-order "networkElementIdentifier".
```
miovd, line 3: I think this is a mistake?

miovd (approved):
Okay with me! Just a small typo I think?

ucte/ucte-converter/src/test/resources/sameId.uct
EtienneLt: yes probably a bad refactor from intellij

Issue #2309, opened 2022-10-02
Topology Processor: Components with single bus are ignored


jeandemanged:
* **Do you want to request a *feature* or report a *bug*?**
that is debatable

* **What is the current behavior?**
Islands (components) with a single bus are ignored.
Precision: This happens only for BUS-BREAKER topology style. In NODE-BREAKER a single busbar e.g. with a load and a generator and no branch connected is correctly reported as a synchronous component.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
PR reproducing the issue: #2310 

* **What is the expected behavior?**
Islands (components) with a single bus to be reported just like any island.

* **What is the motivation / use case for changing the behavior?**
Current behavior is confusing. A single bus network should be valid and could even be computed in LoadFlow (single V-theta bus).

PR #2310, opened 2022-10-02
[WIP] Topology Processor: Components with single bus are ignored (#2309)

jeandemanged:
Signed-off-by: Damien Jeandemange <damien.jeandemange@artelys.com>

Reproduce #2309

### DIFF
--- a/iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/ComponentsTest.java
+++ b/iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/ComponentsTest.java
@@ -1,0 +1,140 @@
+/**
+ * Copyright (c) 2022, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.iidm.network.impl;
+
+import com.powsybl.iidm.network.HvdcLine;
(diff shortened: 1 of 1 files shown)


annetill:
https://github.com/powsybl/powsybl-core/pull/1316

We have tried to fix the bus definition from a node/breaker topology and from a bus/breaker topology in the past. The definition is not equivalent and I think that it is really a bug. Thanks for reopening this issue. @geofjamg and @sylvlecl any opinion about that?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/Networks.java
```diff
@@ -33,8 +33,8 @@ public final class Networks {
     private Networks() {
     }
 
-    public static boolean isBusValid(int branchCount) {
-        return branchCount >= 1;
+    public static boolean isBusValid(int feederCount) {
```
annetill, line 36: @zamarrenolm we have to choose between this fix, that seems to be close to the node/breaker voltage level condition to the reasonable one that is 
````
public static boolean isBusValid(int branchCount, int feederCount) {
    return branchCount >= 1 || branchCount == 0 && feederCount > 1;
}
````

zamarrenolm: For me this condition is ok. Only isolated nodes with no equipment connected will be considered as invalid buses. Components with only one bus but that have at least one equipment connected to it will be considered.

zamarrenolm (approved):

geofjamg (approved):

miovd (approved):

PR #2314, opened 2022-10-04
Add network FourSubstationsNodeBreaker with extensions.

colinepiloquet:

No

Enhance FourSubstationNodeBreaker network.

Adds a factory where the network has position extensions to match the network in the javadoc.


iidm/iidm-test/src/main/java/com/powsybl/iidm/network/test/FourSubstationsNodeBreakerWithExtensionsFactory.java
```diff
@@ -0,0 +1,224 @@
+/**
```
colinepiloquet, line 1: @miovd Do you think it makes sense to put it into this module? Or should I move it to iidm-extensions?

miovd: I would rather have it in iidm-extensions as it could cause issues for implementations that have not implemented all of IIDM extensions

miovd (approved):

PR #2316, opened 2022-10-07
Fix ShortCircuitParametersSerializer boolean default values

etiennehomer:
Signed-off-by: Etienne Homer <etiennehomer@gmail.com>

No

It's a bug fix: the ShortCircuitParametersSerializer uses some default values for boolean values that are different from the default values used in the empty constructor. 

The serialization/deserialization is false for some values.


shortcircuit-api/src/main/java/com/powsybl/shortcircuit/json/ShortCircuitParametersSerializer.java
```diff
@@ -27,9 +28,9 @@ public ShortCircuitParametersSerializer() {
     public void serialize(ShortCircuitParameters parameters, JsonGenerator jsonGenerator, SerializerProvider serializerProvider) throws IOException {
         jsonGenerator.writeStartObject();
         jsonGenerator.writeStringField("version", ShortCircuitParameters.VERSION);
-        JsonUtil.writeOptionalBooleanField(jsonGenerator, "withLimitViolations", parameters.isWithLimitViolations(), false);
-        JsonUtil.writeOptionalBooleanField(jsonGenerator, "withVoltageMap", parameters.isWithVoltageMap(), false);
-        JsonUtil.writeOptionalBooleanField(jsonGenerator, "withFeederResult", parameters.isWithFeederResult(), false);
+        JsonUtil.writeOptionalBooleanField(jsonGenerator, "withLimitViolations", parameters.isWithLimitViolations(), DEFAULT_WITH_LIMIT_VIOLATIONS);
+        JsonUtil.writeOptionalBooleanField(jsonGenerator, "withVoltageMap", parameters.isWithVoltageMap(), DEFAULT_WITH_VOLTAGE_MAP);
+        JsonUtil.writeOptionalBooleanField(jsonGenerator, "withFeederResult", parameters.isWithFeederResult(), DEFAULT_WITH_FEEDER_RESULT);
```
miovd, line 33: Not sure this is good practice. In JSON (serialization in general), we can consider that no field for a boolean means it is false. However I agree there is a bug but this should not be the fix. The cleanest way would be for me to **always** serialize these fields. The second option is to fix the deserializer so that no field means `false`.

miovd (changes requested):
See comment in code.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/json/ShortCircuitParametersSerializer.java
etiennehomer: It's ok for me to always serialize these fields. I made the change

miovd (approved):

PR #2319, opened 2022-10-10
2318: allow target deadband negative values, allow configure importer…

### LINKED ISSUE #2318: IIDM validation: do not throw exception if target deadband < 0 found in a regulating control
Feature

In network validation of regulating controls when a negative target deadband is found an exception is thrown, instead of treating the error as bad data that could be fixed later.

Load a network with target deadband < 0.

Instead of throwing an exception, log the error and reduce the validation level of the network to EQUIPMENT. This way, the user will be able to completely load the case and fix later the bad values.

  - PowSyBl Version: 4.10.1

It would be interesting to check with the provider of the case where the issue was found if a negative value in the target deadband of a regulating control has a special meaning. The CIM/CGMES standard does not specify any requirement about the sign of the deadband, but maybe some users apply an interpretation here? (like only values below the setpoint will be allowed?).

zamarrenolm:

Closes #2318 

**Other information**:
The proposal here is to have a way of letting any importer know that the network we are creating should not be validated at the "most strict" validation level.


zamarrenolm:
Replaced by solution merged in #2317

PR #2326, opened 2022-10-17
Handle UCTE mixed ratio and phase tap changers

JB-H:
Signed-off-by: JB-H <jbheyberger@gmail.com>

#209
Here we propose to handle at the import the mixed ratio and phase tap changers

**Other information**:


ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
```diff
@@ -549,6 +570,96 @@ private static void createPhaseTapChanger(UcteAngleRegulation ucteAngleRegulatio
...
+        // and build both equivalent phase tap changer and ratio tap changer such that it will be exact at ratio = nro
+
+        int lowerRatioTap = getLowTapPosition(uctePhaseRegulation, transformer);
+        RatioTapChangerAdder rtca = transformer.newRatioTapChanger()
+                .setLowTapPosition(lowerRatioTap)
+                .setTapPosition(uctePhaseRegulation.getNp())
+                .setLoadTapChangingCapabilities(!Double.isNaN(uctePhaseRegulation.getU()));
+        if (!Double.isNaN(uctePhaseRegulation.getU())) {
+            rtca.setLoadTapChangingCapabilities(true)
+                    .setRegulating(true)
```
annetill, line 592: I am trying to refactor (and understand) this PR. In case of combination, I understand that the approximation is right only for ratio tap changer current position. So for me we maybe have to set regulating boolean to false.

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
```diff
@@ -549,6 +549,96 @@ private static void createPhaseTapChanger(UcteAngleRegulation ucteAngleRegulatio
...
+            double dy = i * ucteAngleRegulation.getDu() / 100 * Math.sin(Math.toRadians(ucteAngleRegulation.getTheta()));
+            double dyEq = dy * (1. + nrduInit);
+            double dxEq = dx * (1. + nrduInit) + nrduInit;
+            switch (ucteAngleRegulation.getType()) {
+                case ASYM:
+                    rho = 1d / Math.hypot(dyEq, 1d + dxEq) / rhoInit; // the formula = 1d / Math.hypot(dyEq, 1d + dxEq) already takes into account rhoInit, so we divide by rhoInit that will be carried by the ratio tap changer
+                    alpha = Math.toDegrees(Math.atan2(dy, 1 + dx));
+                    break;
+
+                case SYMM:
```
annetill: @phiedw here in case of symmetrical phase tap changer, no difference with classical treatment.

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
JB-H: Agreed

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
JB-H: I checked with a Fresnel diagram and the composition of ratio + angle taps should only change the global ratio which is carried by the ratio tap changer. As I understand the Fresnel Diagram, this doesn't change the alpha.
The diffrence with the asymmetrical phase shifter is that the formula is true for all ratio taps for the symmetrical one.

annetill: Sorry indeed, I didn't understand that your work was not finished. So I will wait for symmetrical support to perform testing. As you see, I have simplified your equations to use only the current rho of the ratio tap changer. But I have to explain that in your documentation. Just tell me if it is okay for you.

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
```diff
@@ -497,55 +522,95 @@ private static void createRatioTapChanger(UctePhaseRegulation uctePhaseRegulatio
...
+        switch (ucteAngleRegulation.getType()) {
+            case ASYM:
+                if (currentRatioTapChangerRho == null) {
+                    rho = 1d / Math.hypot(dy, 1d + dx);
+                    alpha = Math.toDegrees(Math.atan2(dy, 1 + dx));
+                } else {
+                    double dyEq = dy / currentRatioTapChangerRho;
+                    double dxEq = (dx + 1) / currentRatioTapChangerRho - 1;
+                    rho = 1d / Math.hypot(dyEq, 1d + dxEq) / currentRatioTapChangerRho; // the formula = 1d / Math.hypot(dyEq, 1d + dxEq) already takes into account rhoInit, so we divide by rhoInit that will be carried by the ratio tap changer
+                    alpha = Math.toDegrees(Math.atan2(dyEq, 1 + dxEq));
```
phiedw, line 581: I don't understand what is trying to be achieved here.

Using the equations written here:
rho = 1 / hypot(dyEq, 1+dxEq) / rtcRho
      = 1 / hypot(dy/rtcRho, (dx+1)/rtcRho) / rtcRho
      = 1 / hypot(dy, dx+1) / rtcRho²
and 
alpha = atan2(dyEq, 1+dxEq)
          = atan2(dy/rtcRho, (dx+1)/rtcRho)
          = atan2(dy, dx+1)

Are these really the equations we wanted to use? (alpha would be unchanged, and rho would be divided by rtcRho squared)

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
JB-H: yes it is OK

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/UcteImporter.java
JB-H: you are right, something is wrong, I will check that

JB-H: by the way here, rho is not divided by rtcRho squared but is unchanged

geofjamg (approved):
Tested on real data, it looks like to work perfectly!

phiedw (approved):
I agree with these equations :)

PR #2329, opened 2022-10-18
CGMES. Fix conversion of equivalent branches with different nominal voltages

zamarrenolm:

Fix the electric characteristics of IIDM lines converted from equivalent branches that have different nominal voltages at its ends.

Bad flows and pu parameters were obtained for equivalent branches that had different nominal voltages at both ends.

The equivalent branch PI model in CGMES includes an implicit ideal ratio in one of the ends, determined by its `BaseVoltage`, a reference that is required for all equivalent branch objects.
 To consider this ideal ratio without having to change the `Line` model in IIDM we have to perform some adjustments in the electric characteristics of the mapped object.
With this adjustment, the calculation of flows and the OLF gateway will work without any further change. That is, with this change, the OLF code that _adds a ratio to lines with different nominal voltage at both ends_ it is correct and **required**.


marqueslanauja (approved):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/EquivalentBranchConversion.java
```diff
@@ -75,12 +80,74 @@ private void convertEquivalentBranchAtBoundary(int boundarySide) {
...
         double gch = 0;
-        // If we have created buses and substations for boundary nodes,
-        // convert as a regular line
-        if (context.config().convertBoundary()) {
-            convertBranch(r, x, bch, gch);
+        // When the parameter convertBoundary is active:
+        // The boundary nodes are processed and included in the Network with a fictitious substation.
+        // The branches that lie on boundaries can be mapped to the Network directly in the ::convert method.
+        // So, this method should never be invoked when convertBoundary is active.
+        assert !context.config().convertBoundary();
```
miovd, line 87: Could you throw here a more explicit exception in the (unlikely) case it happens?
```java
if (context.config().convertBoundary()) {
      throw new PowsyblException("Boundaries should be converted");
}
```

zamarrenolm: The way the conversion works after we introducer handling boundaries using delayed nodes means that we will never reach this code when conversion of boundaries is active. This is why we preferred to put an assert instead of throwing an exception.

miovd: But in this case, I'm not sure what the assert brings to the table? It can be simply deleted.

zamarrenolm: As we have discussed, we changed the assert to an exception with a detailed explanation, just in case at some point the code is used with a different flow in the Conversion process.

miovd (approved):

PR #2343, opened 2022-10-26
Fix remaining dangling switches when replacing tee point by voltage level on line

FranckLecuyer:

no

Bug fix

Some dangling switches remain in the attached voltage level when replacing a tee point by voltage level on line.
screenshot

These dangling switches are properly removed.

No


iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/ReplaceTeePointByVoltageLevelOnLine.java
```diff
@@ -335,7 +335,7 @@ public void apply(Network network, boolean throwException,
         removedLineReport(reporter, line1ZId);
         lineZ2.remove();
         removedLineReport(reporter, lineZ2Id);
-        lineZP.remove();
+        lineZP.remove(true);
```
miovd, line 338: From what I have understand, this feature is quite buggy and will be removed by @flo-dup . Maybe it is better to wait until the correction is made (it will call another modification)

miovd (changes requested):

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/ReplaceTeePointByVoltageLevelOnLine.java
flo-dup: indeed! see PR #2340

miovd: You can now use `RemoveFeederBay`!

flo-dup (approved):

PR #2345, opened 2022-10-27
Dangling line boundary implementation: get voltage level of network side

zamarrenolm:

Code clarification

A dangling line boundary can be queried for a voltage level.
The voltage level returned is always the voltage level of the network side of the dangling line.

The name of the method in `Boundary` interface has been changed to clarify this.


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Boundary.java
```diff
@@ -49,7 +49,5 @@ default Connectable getConnectable() {
     /**
      * Get the voltage level at network side.
      */
-    default VoltageLevel getVoltageLevel() {
-        return null;
-    }
+    VoltageLevel getNetworkSideVoltageLevel();
```
miovd, line 52: add a default impl
```java
default VoltageLevel getNetworkSideVoltageLevel() {
      if (getConnectable() instanceOf Injection) {
           return ((Injection) getConnectable()).getTerminal().getVoltageLevel();
      }
      if (getConnectable() instanceOf Branch) {
          return ((Branch) getConnectable()).getTerminal(getSide()).getVoltageLevel();
      }
      throw new AssertionError("Unexpected parent connectable: " + getConnectable().getId());
}
```

and deprecate the previous
```java
/**
* @deprecated No voltage level should be linked to Boundary. Use {@link Boundary#getNetworkSideVoltageLevel()} if you want to retrieve the parent object voltage level.
**/
@Deprecated(since = 5.1.0)
default VoltageLevel getVoltageLevel() {
     return null:
}
```

miovd: Actually, keeping the deprecated method is not useful, you can keep it deleted as you suggested!

zamarrenolm: I don't like the idea of adding a default method that checks the type of the parent and then casts ... but I realize that your proposal will simplify alternative iidm implementations, so let's got with it.

miovd (approved):

PR #2347, opened 2022-10-27
CGMES export: fix export of boundary data from dangling lines

zamarrenolm:

Bug fix

When dangling lines are exported to CGMES new fictitious containers (substation, voltage level) for the boundary side are created and written in the EQ instance file.
Also, an EnergyConsumer that is not needed is exported to the EQ.
If the dangling line voltage level topology is node/breaker a new ConnectivityNode is always created for the boundary. If it is bus/branch, the boundary TopologicalNode is written in the EQ as part of the IGM.

It is expected that the IGM export will be combined with a proper boundary definition instance file.
This was already assumed when exporting base voltages: if they were defined in boundary instance files, they were not exported in the generated EQ file.
So:
No fictitious containers are created for the boundary side of dangling lines.
Reference to boundary ConnectivityNode or TopologicalNode must be present as an alias in the DanglingLine. Only if they are not found new nodes will be created (but they will not be considered boundaries in a subsequent re-import).


miovd (commented):
Just for me, did you try to export and re-import networks "from scratch" (from IIDM or  another format) containing dangling lines? It seems to me some dangling lines' data would miss

miovd:
> Just for me, did you try to export and re-import networks "from scratch" (from IIDM or another format) containing dangling lines? It seems to me some dangling lines' data would miss

Maybe boundary nodes (in dangling lines, and later on in tie lines) should be tagged as in boundary files or not

zamarrenolm:
> Just for me, did you try to export and re-import networks "from scratch" (from IIDM or another format) containing dangling lines? It seems to me some dangling lines' data would miss

For networks created directly from IIDM (or any other non-CGMES formats) I've added three unit tests that show how to properly export (and recover) dangling lines. They are in `CgmesExportTest.java`: `testFromIidmDanglingLineBusBranch`, `testFromIidmDanglingLineNodeBreaker` and `testFromIidmDanglingLineNodeBreakerNoBoundaries`. 

To be able to export dangling lines to CGMES and recover them as frontier-related equipment in theory we need only the boundary node where the dangling line lies. This is what is done in these tests: the Network is read from a non-CGMES source, and before exporting it to CGMES, only the boundary node information is set (through the appropriate alias). When the CGMES exported data is imported, an IIDM dangling line is recovered.

The XNode code is read from the boundary definitions, that are external to the model being imported/exported.

The equivalent injection and terminal identifiers at boundaries are specific to the IGM, and we do not need to set them up.

zamarrenolm:
> > Just for me, did you try to export and re-import networks "from scratch" (from IIDM or another format) containing dangling lines? It seems to me some dangling lines' data would miss
> 
> Maybe boundary nodes (in dangling lines, and later on in tie lines) should be tagged as in boundary files or not

My proposal here (made explicit in the unit tests) is that the user should know, when the model is exported to CGMES, if a specific dangling line should be considered _lying_ on an external boundary. If yes, then the user must set the external boundary node identifier to use in the proper alias of the dangling line.

miovd:
I don't agree with it, there will be cases where no external boundaries set exist (at least not yet). Maybe the easiest way would be to create an option `export-boundaries`, `false` by default. If false, it is kept as you did, if true fictitious containers are created.

zamarrenolm:
> I don't agree with it, there will be cases where no external boundaries set exist (at least not yet). Maybe the easiest way would be to create an option `export-boundaries`, `false` by default. If false, it is kept as you did, if true fictitious containers are created.

I do not see the need for a specific additional parameters to control the behaviour of the boundary export. The dangling lines that have an alias with the boundary node will be exported using it; the boundary node will not be defined in the export, it must be defined in a external BD file that will be required to re-import. The dangling lines that do not have an alias for the boundary node are exported using a new fictitious node that will be written in the exported EQ. The user may want to export only some dangling lines linked to known, existing boundaries and leave the rest as connected to the fictitious nodes.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
```diff
@@ -305,6 +314,16 @@ public boolean isExportedEquipment(Identifiable<?> c) {
         return !ignored;
     }
 
+    public CgmesExportContext setBoundaryEqId(String boundaryEqId) {
```
miovd, line 317: Is it possible to store BD_EQ and BD_TP ids during import (as properties) and automatically filled the context during export? It can be done in another PR (if you prefer to do it later, I will create an issue)

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
```diff
@@ -625,21 +607,20 @@ private static String writeDanglingLineBaseVoltage(DanglingLine danglingLine, St
...
-            // New ConnectivityNode
-            connectivityNodeId = CgmesExportUtil.getUniqueId();
-            ConnectivityNodeEq.write(connectivityNodeId, danglingLine.getNameOrId() + "_NODE", voltageLevelId, cimNamespace, writer);
+        if (context.writeConnectivityNodes()) {
+            // We keep the connectivity node from the boundary definition as an alias in the dangling line
+            if (danglingLine.getAliasFromType(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.CONNECTIVITY_NODE).isPresent()) {
+                connectivityNodeId = context.getNamingStrategy().getCgmesIdFromAlias(danglingLine, Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.CONNECTIVITY_NODE);
+            } else {
+                // If no information about original boundary has been preserved in the IIDM model,
+                // we will create a new ConnectivityNode inside the voltage level of the dangling line network side
```
miovd, line 619: As discussed, I would rather create a fictitious new voltage level.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/AbstractConductingEquipmentConversion.java
```diff
@@ -251,6 +255,9 @@ public void convertToDanglingLine(int boundarySide, double r, double x, double g
             dl.addAlias(tn, Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.TOPOLOGICAL_NODE);
             dl.setProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + dl.getId() + "." + CgmesNames.TOPOLOGICAL_NODE, tn); // TODO: delete when aliases are correctly handled by mergedlines
         });
+        Optional.ofNullable(connectivityNodeId(boundarySide)).ifPresent(cn ->
+            dl.addAlias(cn, Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.CONNECTIVITY_NODE)
```
miovd, line 259: Can you just add "boundary" in the name?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
```diff
@@ -625,21 +607,20 @@ private static String writeDanglingLineBaseVoltage(DanglingLine danglingLine, St
...
+                connectivityNodeId = CgmesExportUtil.getUniqueId();
```
miovd, line 620: Do you need to add it as alias?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/TopologyExport.java
```diff
@@ -227,30 +227,33 @@ private static void writeTerminal(String terminalId, String topologicalNode, Str
...
+            if (topologicalNodeId.isEmpty()) {
+                // If no information about original boundary has been preserved in the IIDM model,
+                // we will create a new TopologicalNode inside the voltage level of the dangling line network side
                 String baseVoltage = context.getBaseVoltageByNominalVoltage(dl.getTerminal().getVoltageLevel().getNominalV()).getId();
-                // TODO(Luma) It is wrong to consider the container of boundary topological node to be the same of dangling line
-                //  To be fixed in a separate PR (in fact boundary TNs of dangling lines should not be exported)
                 String containerId = context.getNamingStrategy().getCgmesId(dl.getTerminal().getVoltageLevel());
-                writeTopologicalNode(topologicalNodeId.get(), dl.getNameOrId(), containerId, baseVoltage, cimNamespace, writer);
+                String fictTopologicalNodeId = CgmesExportUtil.getUniqueId();
+                dl.addAlias(fictTopologicalNodeId, Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.TOPOLOGICAL_NODE);
```
miovd, line 244: Can you add "boundary" in the name?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
miovd: I discussed with @annetill and she agrees that we need a fictitious voltage level **and** a fictitious substation in this case. It is too weird otherwise...

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
zamarrenolm: Yes, for networks coming from CGMES we could use the approach you suggest. It is ok for me to have a separate PR

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
zamarrenolm: Fictitious containers (substation and voltage level) are created.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
zamarrenolm: Yes, it is convenient to keep it in the corresponding alias once we have created it, although it is not strictly needed to complete the current export.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/AbstractConductingEquipmentConversion.java
zamarrenolm: added `_Boundary` suffix for the related alias types.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/TopologyExport.java
zamarrenolm: added `_Boundary` suffix for the related alias types.

miovd (approved):

PR #2350, opened 2022-11-02
Remove iidm converter api

geofjamg:

No

Redesign of network loading and saving API

We can import/export or load/save a network using utility classes Importers and Exporters. 

We have removed the iidm-converter-api module and moved everything to iidm-api and all loading and saving functions have been moved to `Network` class 

**Other information**:


miovd (commented):
It seems to me that a `save` method implies that it is saved in a database or an object, which is okay when saved in a datasource but less obvious when we want to write a network in a file. It is just a remark though and I don't have a better name idea. Maybe discriminate with `write` when passing a path or a file/directory name, `save` otherwise?

geofjamg:
> It seems to me that a `save` method implies that it is saved in a database or an object, which is okay when saved in a datasource but less obvious when we want to write a network in a file. It is just a remark though and I don't have a better name idea. Maybe discriminate with `write` when passing a path or a file/directory name, `save` otherwise?

I just wanted to be symmetrical with 'load' as load and save are often used together.
You are probably right, write is better but in that case I rename load to read ?

miovd (approved):

PR #2356, opened 2022-11-09
Add status to shortcircuit result

colinepiloquet:

No.

Adds status to the results of a short circuit analysis. The status is added in the FaultResult object. Then, for each Fault, it is now possible to specify the status of the computation between four options: CONVERGED, SOLVER_FAILED,  FAILED and NO_DATA.

Status is not in the results.


shortcircuit-api/src/main/java/com/powsybl/shortcircuit/json/ShortCircuitAnalysisResultDeserializer.java
```diff
@@ -61,6 +62,11 @@ public ShortCircuitAnalysisResult deserialize(JsonParser parser, Deserialization
                     });
                     break;
 
+                case "status":
```
miovd, line 65: Serialization (and deserialization) should be versioned following this change

miovd (commented):
Version for serialization should be 1.1

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FaultResult.java
```diff
@@ -127,4 +158,13 @@ public Duration getTimeConstant() {
     public List<ShortCircuitBusResults> getShortCircuitBusResults() {
         return shortCircuitBusResults;
     }
+
+    /**
+     * The computation status. Converged if computation went ok, no data if the transient reactance of generators are missing
```
annetill, line 163: Can you update the java doc?

annetill (approved):

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FaultResult.java
```diff
@@ -22,6 +22,15 @@
...
 public final class FaultResult extends AbstractExtendable<FaultResult> {
 
+    public enum Status {
+        SUCCESS,
+        NO_SHORTCIRCUIT_DATA,
+        SOLVER_FAILURE,
+        FAILURE
+    }
+
+    private Status faultResultStatus;
```
sylvlecl, line 32: Let's make it `final`.

`status` is enough, faultResultStatus sounds redundant (and actually we could say it's more the  status of the "computation", than the status of the "result"). I see it's already done this way in the getter and serialization --> let's be consistent here too :)

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FaultResult.java
```diff
@@ -127,4 +158,13 @@ public Duration getTimeConstant() {
     public List<ShortCircuitBusResults> getShortCircuitBusResults() {
         return shortCircuitBusResults;
     }
+
+    /**
+     * The computation status. SUCCESS if computation went ok, NO_SHORTCIRCUIT if the transient reactance of every generators is missing,
```
sylvlecl, line 163: I would propose to move the doc for the different values to the enum class itself (each description goes to the javadoc of the corresponding enum value).

This way the user can refer to the definition of the values from anywhere they are used, not only from this getter.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/json/FaultResultSerializer.java
```diff
@@ -20,6 +20,10 @@
  */
 public class FaultResultSerializer extends StdSerializer<FaultResult> {
 
+    // VERSION = 1.0 fault, shortCircuitPower, timeConstant, feederResult, limitViolations, current, voltage, shortCircuitBusResults
+    // VERSION = 1.1 status
+    private static final String VERSION = "1.1";
```
sylvlecl, line 25: I think it's better to not introduce a version at this level of detail:
let's keep a version at the global shortcircuit result level only (and increment it from 1.0 to to 1.1)

sylvlecl (changes requested):
Mainly: I think we should increment the global version of the result, not introduce a version at the fault level

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/json/FaultResultSerializer.java
sylvlecl: Note: before actually doing this change, I would like some confirmation / agreement from @miovd and @geofjamg .

I am not 100% sure there is a consensus about this approach of versioning at the "file" level instead of the "object" level.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FaultResult.java
```diff
@@ -52,13 +77,31 @@ public FaultResult(Fault fault, double shortCircuitPower, List<FeederResult> fee
...
 
     public FaultResult(Fault fault, double shortCircuitPower, List<FeederResult> feederResults,
-                       List<LimitViolation> limitViolations, FortescueValue current, Duration timeConstant) {
-        this(fault, shortCircuitPower, feederResults, limitViolations, current, null, Collections.emptyList(), timeConstant);
+                       List<LimitViolation> limitViolations, FortescueValue current, FortescueValue voltage, List<ShortCircuitBusResults> shortCircuitBusResults,
+                       Duration timeConstant, Status status) {
+        this.fault = Objects.requireNonNull(fault);
+        this.shortCircuitPower = shortCircuitPower;
+        this.feederResults = feederResults;
+        this.limitViolations = limitViolations;
```
miovd, line 85: Same thing

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/json/FaultResultSerializer.java
miovd: I agree

miovd (changes requested):
Some technical changes.

I also agree with @sylvlecl , since `faultResultSerializer` is always called in the context of shortcircuit serialization, I would put the incremented version in shortcircuit results.

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/FaultResult.java
```diff
@@ -40,25 +62,39 @@ public final class FaultResult extends AbstractExtendable<FaultResult> {
...
+        }
         this.current = Objects.requireNonNull(current);
         this.voltage = voltage;
-        this.shortCircuitBusResults = shortCircuitBusResults;
+        this.shortCircuitBusResults = new ArrayList<>();
+        if (shortCircuitBusResults != null) {
+            this.shortCircuitBusResults.addAll(shortCircuitBusResults);
+        }
         this.timeConstant = timeConstant;
+        this.status = status;
```
miovd, line 83: Should it be `Objects.requireNonNull`?

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/json/FaultResultDeserializer.java
```diff
@@ -100,7 +109,9 @@ public FaultResult deserialize(JsonParser parser, DeserializationContext deseria
                     throw new AssertionError("Unexpected field: " + parser.getCurrentName());
             }
         }
-        FaultResult faultResult = new FaultResult(fault, shortCircuitPower, feederResults, limitViolations, current, voltage, shortCircuitBusResults, timeConstant);
+
```
miovd, line 112: Add a check if status null and if there is no filled current, status is `FAILURE`, else `SUCCESS`

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/json/FaultResultDeserializer.java
```diff
@@ -91,6 +94,12 @@ public FaultResult deserialize(JsonParser parser, DeserializationContext deseria
                     shortCircuitBusResults = parser.readValueAs(new TypeReference<List<ShortCircuitBusResults>>() { });
                     break;
 
+                case "status":
+                    JsonUtil.assertGreaterOrEqualThanReferenceVersion(CONTEXT_NAME, "Tag: " + parser.getCurrentName(), ShortCircuitAnalysisResult.VERSION, "1.1");
```
miovd, line 98: Since you have to check the current version, it can not be a `StdDeserializer`, just an object deserializer called in `ShortCircuitAnalysisResultDeserializer` with the current version passed as parameter

miovd (changes requested):

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/json/FaultResultDeserializer.java
```diff
@@ -22,87 +21,111 @@
...
-
-                case "extensions":
-                    parser.nextToken();
-                    extensions = JsonUtil.readExtensions(parser, deserializationContext, SUPPLIER.get());
-                    break;
-
-                default:
-                    throw new AssertionError("Unexpected field: " + parser.getCurrentName());
+            FaultResult faultResult;
+            if (status == null) {
```
miovd, line 115: check the version here (must be <= 1.0)

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/json/FaultResultDeserializer.java
```diff
@@ -22,87 +21,111 @@
...
+
+    private static final String CONTEXT_NAME = "FaultResult";
 
     private static final Supplier<ExtensionProviders<ExtensionJsonSerializer>> SUPPLIER =
             Suppliers.memoize(() -> ExtensionProviders.createProvider(ExtensionJsonSerializer.class, "short-circuit-analysis"));
 
-    FaultResultDeserializer() {
-        super(FaultResult.class);
-    }
+    public List<FaultResult> deserialize(JsonParser parser, DeserializationContext deserializationContext, String version) throws IOException {
```
miovd, line 38: Do you need `DerserializationContext`?

miovd (commented):
Almost done! Just one final remark

miovd (approved):

PR #2380, opened 2022-11-28
Move AbstractXmlConverterTest to src to be used in other repositories

miovd:

No

Moving code following removal of test jars: `AbstractXmlConverterTest` is used in powsybl-rte-core

No

**Other information**:


miovd:
It has been decided that `AbstractXmlConverterTest` should be duplicated in repositories it is used (**only** for versioned extensions). For simple xml roundTrip, `AbstractConverterTest` can be used.

PR #2391, opened 2022-12-06
Add docker based local computation manager test utilities

geofjamg:

No

Feature

There is no easy way to test something that is based on local computation manager, because the command should be installed on the system.

We can specify a docker image that will be used for executing commands of a local computation manager.  

**Other information**:

### DIFF
--- a/computation-local-test/src/test/java/com/powsybl/computation/local/test/DockerLocalCommandExecutorTest.java
+++ b/computation-local-test/src/test/java/com/powsybl/computation/local/test/DockerLocalCommandExecutorTest.java
@@ -1,0 +1,75 @@
+/**
+ * Copyright (c) 2022, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.computation.local.test;
+
+import com.powsybl.commons.config.PlatformConfig;
+import com.powsybl.commons.test.DockerTests;
+import com.powsybl.computation.*;
+import com.powsybl.computation.local.LocalComputationConfig;
+import com.powsybl.computation.local.LocalComputationManager;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.rules.TemporaryFolder;
+
+import java.io.IOException;
+import java.nio.file.Files;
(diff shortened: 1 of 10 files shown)


computation-local-test/src/test/resources/com/powsybl/config/test/config.yml
```diff
@@ -0,0 +1,2 @@
+computationDocker:
```
sylvlecl, line 1: maybe snake-case instead, for more consistency?

computation-local-test/src/test/java/com/powsybl/computation/local/test/DockerLocalCommandExecutorTest.java
```diff
@@ -0,0 +1,75 @@
...
+    public TemporaryFolder tempFolder = new TemporaryFolder();
+
+    private ComputationManager computationManager;
+
+    @Before
+    public void setup() throws IOException {
+        Path localDir = tempFolder.getRoot().toPath();
+        Path dockerDir = Path.of("/tmp");
+        ComputationDockerConfig config = ComputationDockerConfig.load(PlatformConfig.defaultConfig());
+        computationManager = new LocalComputationManager(new LocalComputationConfig(localDir),
```
sylvlecl, line 47: Do we want to provide an easier way to create the computation manager, maybe with a subclass of `LocalComputationManager` ?

distribution-core/pom.xml
```diff
@@ -156,6 +156,12 @@
             <version>${project.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>powsybl-computation-local-test</artifactId>
```
sylvlecl, line 161: I wonder if we should be more explicit about the use of containers, in the name.

sylvlecl (commented):
Mainly some thoughts, possibly for later evolution

distribution-core/pom.xml
geofjamg: Like powsybl-computation-local-docker-test ?

computation-local-test/src/test/java/com/powsybl/computation/local/test/DockerLocalCommandExecutorTest.java
geofjamg: `DockerLocalComputationManager` added

computation-local-test/src/test/resources/com/powsybl/config/test/config.yml
geofjamg: Done

sylvlecl (approved):

PR #2396, opened 2022-12-08
ContingencyBuilder add identifiable

OpenSuze:

No

New feature

You have to know which type of element you want a contingency from.

You can provide an Identifiable that will be added as contingency element.
Instead of providing an Identifiable, you can also provide a String and a network to define a contingency element.

**Other information**:
Open to discussion

### DIFF
--- a/contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyBuilder.java
+++ b/contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyBuilder.java
@@ -7,6 +7,10 @@
 
 package com.powsybl.contingency;
 
+import com.powsybl.commons.PowsyblException;
+import com.powsybl.iidm.network.Identifiable;
+import com.powsybl.iidm.network.Network;
+
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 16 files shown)


contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyBuilder.java
```diff
@@ -113,4 +117,17 @@ public ContingencyBuilder addSwitch(String id) {
         elements.add(new SwitchContingency(id));
         return this;
     }
+
+    public ContingencyBuilder addIdentifiable(String id, Network network) {
```
EtienneLt, line 121: would it be better if it is static ?

contingency/contingency-api/src/test/java/com/powsybl/contingency/ContingencyTest.java
```diff
@@ -154,4 +161,59 @@ public void validationTestForTwt3() {
                         .map(Contingency::getId)
                         .collect(Collectors.toList()));
     }
+
+    @Test
+    public void testIdentifiableBuilder() {
```
EtienneLt, line 166: could you test maybe a contingency with 2 or more elements created with this new method

sylvlecl (approved):

EtienneLt (approved):

flo-dup (approved):

PR #2397, opened 2022-12-09
PSSE: Admit special characters

marqueslanauja:

Bug fix

Some special characters are not supported

Special characters are supported
For readability duplicated code is not removed.

**Other information**:


psse/psse-model/src/main/java/com/powsybl/psse/model/io/RecordGroupIOJson.java
```diff
@@ -119,6 +118,15 @@ private static String[] readFieldNames(JsonNode n) {
         return fields.toArray(new String[0]);
     }
 
+    private static String readArrayContent(JsonNode jrecord) {
+        String srecord = jrecord.toString();
+        if (jrecord.isArray()) {
+            // Remove square brackets
+            return srecord.substring(1, srecord.length() - 1);
```
geofjamg, line 125: There should be a cleaner way to read a json array using the jackson API instead of parsing the string?

zamarrenolm (approved):

geofjamg (approved):

psse/psse-model/src/main/java/com/powsybl/psse/model/io/RecordGroupIOJson.java
geofjamg: ok

PR #2402, opened 2022-12-09
Active Power Control extension: add participationFactor

jeandemanged:
add participationFactor attribute in Active Power Control extension

Signed-off-by: Damien Jeandemange <damien.jeandemange@artelys.com>

doc/website update for next release to be done

No

Feature: 
  - add participationFactor attribute to active power control iIDM extension (for use by LoadFlow)
  - add corresponding LoadFlow BalanceType

No support of participationFactor representation, for use by LoadFlow

  - participationFactor attribute can be represented via active power control iIDM extension
  - LoadFlow API BalanceType enum has new option to use participation factors

**Note sure ...**
I made the attribute optional in the XML serialization, so I do not think this qualifies as a breaking change. Please let me know if you prefer that I bump a new version of the active power control extension xsd...

**Other information**:

* Please le me know if you prefer a separate extension for this instead of reusing the existing ActivePowerControl extension (which seems a good place imho)

* Other PRs will follow to:
  - fill in these attributes in CGMES Importer
  - open-loadflow support of these attributes


iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/ActivePowerControl.java
```diff
@@ -25,8 +25,66 @@ default String getName() {
...
+     * @param longPF new long term economic participation factor value
+     */
+    void setLongPF(float longPF);
+
+    /**
+     * Priority for use as powerflow voltage phase angle reference bus selection.
+     * 0 = don't care (default) 1 = highest priority. 2 is less than 1 and so on.
+     * @return reference priority
+     */
+    int getReferencePriority();
```
geofjamg, line 83: We already have an extension SlackTerminal for that (see CGMES import how referencePriority is used to create a SlackTerminal extension)

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowParameters.java
```diff
@@ -36,6 +36,10 @@ public enum VoltageInitMode {
     public enum BalanceType {
         PROPORTIONAL_TO_GENERATION_P,
         PROPORTIONAL_TO_GENERATION_P_MAX,
+        PROPORTIONAL_TO_GENERATION_REMAINING_MARGIN,
```
geofjamg, line 39: Remaining margin means proportional to pmax - targetP ?

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/ActivePowerControl.java
jeandemanged: SlackTerminal could indeed work for this but not in the general case:
- there can be only one SlackTerminal per VoltageLevel. In the case where there are two generators in on VoltageLevel, gen1 with priority 1, gen2 with priority 2. Say both are connected initially, but then the user disconnects the generator in priority 1, the other generator with priority 2 cannot be picked up.
- SlackTerminal-s across different VoltageLevel-s do not have have any relative priority. So If I have two VoltageLevel-s, VL1 with gen1 of referencePriority1, VL2 with gen2 of referencePriority2, in this case it is not possible to represent that gen1 has higher priority than gen2 for slack slelection.

geofjamg: The problem with referencePriority on generators is that it does not cover the general case of slack bus  selection but only slack bus on PV buses. Only very old power flow, does not allow general slack bus selection and most of the time best slack bus is not on a PV bus (like when you select most meshed bus). Also we tend to have more and more generators modelling on MT in our cases (so in 15Kv behind a step up transformer) and having a slack of this kind of bus is not robust.

Maybe we should just try to improve/enrich existing SlackTerminal extension in order to meet your needs ?

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowParameters.java
geofjamg: PROPORTIONAL_TO_GENERATION_REMAINING_MARGIN seems ok to me

geofjamg: Maybe we should you add a comment to describe exactly what does it mean

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/ActivePowerControl.java
jeandemanged: just thinking... an alternative solution could be to enhance the existing SlackTerminal extension to allow 1/ for multiple SlackTerminal extensions to be present in a given VoltageLevel, and 2/ add priority to the extension. But This would cause issues for backward compatibility of existing tools relying on this extension, and 2/ the very same extension is used for slack selection results, and when it comes to slack selection results, priority don't really make sense any more.

jeandemanged: I de-scoped referencePriority from this PR, have to think about it

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowParameters.java
jeandemanged: I added javadoc to the options

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/ActivePowerControl.java
geofjamg: Why not, maybe in order to complete this PR, we can just remove the referencePriority and think about how to improve slack selection part in another issue/PR ?

jeandemanged: I removed referencePriority from this PR and entered issue #2412 to track this topic

miovd (approved):

PR #2407, opened 2022-12-12
PSSE: Parallel Two Terminal DC between the same AC Buses

marqueslanauja:

Bug fix

Only one Two Terminal DC is allowed between the same AC buses

More than one Two Terminal DC is allowed between the same AC buses

**Other information**:


psse/psse-converter/src/main/java/com/powsybl/psse/converter/TwoTerminalDcConverter.java
```diff
@@ -89,7 +89,7 @@ private static double getLccConverterPowerFactor(PsseTwoTerminalDcConverter conv
     }
 
     private static String getLccConverterId(PsseTwoTerminalDcTransmissionLine psseTwoTerminalDc, PsseTwoTerminalDcConverter converter) {
-        return "LccConverter-" + psseTwoTerminalDc.getRectifier().getIp() + "-" + psseTwoTerminalDc.getInverter().getIp() + "-" + converter.getIp();
+        return "LccConverter-" + psseTwoTerminalDc.getRectifier().getIp() + "-" + psseTwoTerminalDc.getInverter().getIp() + "-" + converter.getIp() + "-" + psseTwoTerminalDc.getName();
```
geofjamg, line 92: This end up to quite strange naming in some cases, what about just using a counter?

zamarrenolm: Agree. Removed the names from the identifiers are used Identifiables.getUniqueId to avoid duplicates. Also kept the name of the PSSE two-terminal DC at IIDM converters and IIDM DC HVDC line to let the user keep track of the different elements.

zamarrenolm (approved):

geofjamg (approved):

PR #2410, opened 2022-12-12
Fix UCTE importer conventions of PST active power flow regulation

murgeyseb:

No

Bug fix

Current implementation of UCTE importer has an issue in PST angle regulation import. UCTE and IIDM conventions regarding the regulation value are not consistent. This may be the reason why some divergence where observed with angle regulation, forcing to disable it by default.

UCTE importer fixed to give correct convention to the regulation value.


phiedw (approved):

annetill (approved):
@TheMaskedTurtle and @phiedw  Be careful about this PR! I imagine that you have to remove the fix on your side for the phase shifter in active power control.

PR #2413, opened 2022-12-12
Security Analysis Result : component count, loss of load/generation/elements

obrix:
Add lost synchronous component count, loss of load, loss of generation and set of element lost to post contingency result.

**Other information**:


security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResultBuilder.java
```diff
@@ -239,6 +239,14 @@ public class PostContingencyResultBuilder extends AbstractLimitViolationsResultB
 
         private PostContingencyComputationStatus status;
 
+        private int createdComponentCount;
```
annetill, line 242: In my opinion, here you must add two int:
- createdSynchronousComponentCount : when the created component is created after the loss of classical lines.
- createConnectedComponentCount : when the created component is created after the loss of an hvdc line.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResultBuilder.java
```diff
@@ -239,6 +239,14 @@ public class PostContingencyResultBuilder extends AbstractLimitViolationsResultB
...
+
+        private double lossOfLoad;
```
annetill, line 244: In my opinion, you must add somewhere "activePower" for load and for generation.

jeandemanged: agree

i would suggest not having total `lossOfLoad` nor  `lossOfGeneration`, only the details of the disconnected equipments, from which and API users can derive their own summaries (total counts, total counts by type, total activePower, totalActivePower by type). Something like:

```
"outagedInjections" : [
  {
    "type": "LOAD",
    "id": "Load1",
    "activePower": 12.34,
    "reactivePower": 2.34
  },
  {
    "type": "GEN",
    "id": "Gen1",
    "activePower": 12.34,
    "reactivePower": 2.34
  },
  {
    "type": "BATTERY",
    "id": "Batt1",
    "activePower": 12.34,
    "reactivePower": 2.34
  }
]
```

annetill: Yes, it could be nice of course. But keep in mind that it can decrease the performances and increase the complexity in the implementation:
- For loads, in case of slack distribution on loads, what are the values for `P0` and `Q0` ? The base case before slack distribution of after slack distribution ? Furthermore, we also have a mode that keeps the power factor constant, changing reactive power during slack distribution.
- For generators, in case of slack distribution on generation, what are the values of active power ? The values before or after slack distribution ? In case of a generator that controls voltage: we also have to update the reactive power during the security analysis. I imagine we have to put the reactive power at the end of the base case resolution. Right? 
So I think that we must, if we do that, ask for the state after the first load flow, all the regulations done. If yes, I prefer that we try to implement it before merging these PR.

jeandemanged: I am ok to leave as it is now for this PR. Just some thought for future work 
- for postcontingency results: active power values reported should be the 'cut values' at the time the contingency occurred. so these are basecase state solved value (after slack distribution)
- for post operator strategy results: active power values reported should be the 'cut values' at the time the operator strategy are applied. So these are contingency state solved value (after slack distribution)
- we could probably forget the reactivePower part which doesn't bring much value

annetill: Good idea for the reactive power part ;-) @geofjamg what do you prefer for active power losses? The detailed solution or the global one?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResultBuilder.java
obrix: I just renamed and added the missing attribute.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResultBuilder.java
geofjamg: We can start we global/sum values and see later if one really need detailed results

geofjamg: I think we have a naming problem: when first looking at loadActivePowerLoss I understand "loss" as heat loss not loss because of lost elements.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/results/ConnectivityResult.java
```diff
@@ -0,0 +1,55 @@
...
+
+    private final int createdConnectedComponentCount;
+
+    private final double loadActivePowerLoss;
+
+    private final double generationActivePowerLoss;
+
+    private final Set<String> lostElements;
+
+    public ConnectivityResult(int createdSynchronousComponentCount, int createdConnectedComponentCount, double loadActivePowerLoss,
```
annetill, line 27: As discuss in TSC yesterday, it seems that the naming is not the good one. 
- `loadActivePowerLoss` can be understand as network losses. We suggest `disconnectedLoadActivePower`.
- `generationActivePowerLoss` : we suggest `disconnectedGenerationActivePower`.

obrix: I renamed the two attributes.

annetill (approved):

Issue #2390, opened 2022-12-06
[IIDM] Default values for network element adders


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature

* **What is the current behavior?**

When creating networks programmatically, almost all attributes of all elements must be explicitly defined.

This makes the creation of simple networks very verbose and fastidious.

* **What is the expected behavior?**

Instead, we could make some attributes optional, and assign them default values.

A first pull request was drafted here: #2003 

Here is a tentative list of default values :

- Line/tie lines/dangling lines/two windings transformers and three windings transformers legs:
  - g/b/g1/b1/g2/b2 = 0
  - r = 0 ?
  - in bus/breaker topology, the voltage level is redundant with bus ID --> it should be optional
- Two windings transformers:
  - Use voltage level nominal voltage for rated voltages ?
- Ratio tap changers:
  - r, x, g, b = 0 for every step
- Phase tap changers 
  - r, x, g, b = 0 for every step
  - ratio = 1 for every step
- Shunt compensators:
  - g = 0 for each section

geofjamg:
Agree on all your proposition.
For r=0 with line, I am also not 100% convinced, let's keep it mandatory for a first version?
@annetill if you also agree on this list, we can use this spec to terminate your PR?

annetill:
For ratio tap changers, an alpha of 0 for every step?

PR #2417, opened 2022-12-15
Default values for network element adders

HugoKulesza:

#2390 

**Other information**:


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractBranchAdder.java
```diff
@@ -70,6 +76,18 @@ protected VoltageLevelExt checkAndGetVoltageLevel1() {
         return voltageLevel1;
     }
 
+    private String checkAndGetDefaultVoltageLevelId1() {
```
EtienneLt, line 79: I think you can mutualize these

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractBranchAdder.java
```diff
@@ -110,4 +133,16 @@ protected VoltageLevelExt checkAndGetVoltageLevel2() {
         return voltageLevel2;
     }
 
+    private String checkAndGetDefaultVoltageLevelId2() {
```
EtienneLt, line 136: with this

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ThreeWindingsTransformerAdderImpl.java
```diff
@@ -141,6 +146,18 @@ protected VoltageLevelExt checkAndGetVoltageLevel() {
             return voltageLevel;
         }
 
+        private String checkAndGetDefaultVoltageLevelId() {
```
EtienneLt, line 149: same here

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ThreeWindingsTransformerAdderImpl.java
```diff
@@ -131,7 +131,12 @@ protected TerminalExt checkAndGetTerminal() {
 
         protected VoltageLevelExt checkAndGetVoltageLevel() {
             if (voltageLevelId == null) {
-                throw new ValidationException(this, "voltage level is not set");
+                String defaultVoltageLevelId = checkAndGetDefaultVoltageLevelId();
+                if (defaultVoltageLevelId == null) {
+                    throw new ValidationException(this, "voltage level is not set");
```
EtienneLt, line 136: maybe it could be more explicit, something like "voltage level is not set and it has no default value"

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractBranchAdder.java
```diff
@@ -110,4 +121,24 @@ protected VoltageLevelExt checkAndGetVoltageLevel2() {
         return voltageLevel2;
     }
 
+    private String checkAndGetDefaultVoltageLevelId(String connectableBus) {
+        if (connectableBus != null) {
+            VoltageLevelExt defaultVoltageLevel = (VoltageLevelExt) getNetwork().getBusBreakerView()
```
annetill, line 126: In case of a string corresponding to connectableBus that is not an id of a bus, what is happening ? In open-loadflow, we always check that the bus is not null. Why we don't need it here?

annetill (approved):
Just a small question for my understanding.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractBranchAdder.java
HugoKulesza: Regarding the first question : if no voltage level is given and connectableBus is not a bus ID, the creation fails with a NullPointerException (when trying to getVoltageLevel() on the not found bus). Should I change this to make the origin of the error clearer ?

Regarding the second point : if I understood right, the bus is checked after when building the Terminal and is primarily used to say if the element is really connected to the connectableBus (if the bus is null, the element is not connected).

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractBranchAdder.java
```diff
@@ -110,4 +121,24 @@ protected VoltageLevelExt checkAndGetVoltageLevel2() {
...
+                    .getBus(connectableBus)
+                    .getVoltageLevel();
+            if (defaultVoltageLevel.getTopologyKind().equals(TopologyKind.BUS_BREAKER)) {
```
miovd, line 129: I don't think this is the check that should be done: if the voltage level is `NODE_BREAKER`, it should fail later, not here necessarily.

The think to check is: if the bus `connectableBus` exists i.e. if `getNetwork().getBusBreakerView().getBus(connectableBus) != null` (it can not exist: return null, otherwise you will have a NPE)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ThreeWindingsTransformerAdderImpl.java
```diff
@@ -141,6 +146,24 @@ protected VoltageLevelExt checkAndGetVoltageLevel() {
...
         }
 
+        private String checkAndGetDefaultVoltageLevelId() {
+            if (connectableBus != null) {
+                VoltageLevelExt defaultVoltageLevel = (VoltageLevelExt) getNetwork().getBusBreakerView()
+                        .getBus(connectableBus)
+                        .getVoltageLevel();
+                if (defaultVoltageLevel.getTopologyKind().equals(TopologyKind.BUS_BREAKER)) {
+                    return defaultVoltageLevel.getId();
+                }
```
miovd, line 156: Same remark as above

miovd (changes requested):
I asked some changes in check and get of voltage levels for branches and t3w

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractBranchAdder.java
miovd: Regarding the first question, I would prefer a clearer error than a NPE indeed. I put in my review that voltageLevel should return null in this case but we can go even deeper and state that the bus associated to the id does not exist.

miovd (approved):

miovd:
One checkstyle error to fix but otherwise it is ready to be merged for me!

PR #2429, opened 2023-01-02
Use contingency context in operator strategies

geofjamg:

No

Feature

We can only associate an operator strategy to one contingency.

We can thanks to contingency context associate an operator strategy to not only one contingency but also to non or all contingencies.

**Other information**:

### DIFF
--- a/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/OperatorStrategyDeserializer.java
+++ b/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/OperatorStrategyDeserializer.java
@@ -16,6 +16,8 @@ import com.powsybl.commons.extensions.Extension;
 import com.powsybl.commons.extensions.ExtensionJsonSerializer;
 import com.powsybl.commons.extensions.ExtensionProviders;
 import com.powsybl.commons.json.JsonUtil;
+import com.powsybl.contingency.ContingencyContext;
+import com.powsybl.contingency.ContingencyContextType;
 import com.powsybl.security.condition.Condition;
 import com.powsybl.security.strategy.OperatorStrategy;
 
@@ -37,6 +39,7 @@ public class OperatorStrategyDeserializer extends StdDeserializer<OperatorStrate
 
     private static class ParsingContext {
         String id;
+        ContingencyContextType contingencyContextType;
         String contingencyId;
         Condition condition;
         List<String> actionIds;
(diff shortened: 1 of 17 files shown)


security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/OperatorStrategyDeserializer.java
```diff
@@ -73,7 +79,9 @@ public OperatorStrategy deserialize(JsonParser parser, DeserializationContext de
                     return false;
             }
         });
-        OperatorStrategy operatorStrategy = new OperatorStrategy(context.id, context.contingencyId, context.condition, context.actionIds);
+        ContingencyContext contingencyContext = new ContingencyContext(context.contingencyId,
+                context.contingencyContextType != null ? context.contingencyContextType : ContingencyContextType.SPECIFIC);
```
jeandemanged, line 83: maybe just add a note that this is done in this way to support both V1.0 and V1.1 JSON format - could be easily forgotten a couple of months from now ...

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/OperatorStrategyListDeserializer.java
```diff
@@ -49,8 +49,8 @@ public OperatorStrategyList deserialize(JsonParser parser, DeserializationContex
                     return false;
             }
         });
-        if (context.version == null || !context.version.equals(OperatorStrategyList.VERSION)) {
-            throw new JsonMappingException(parser, "version is missing or not equal to 1.0");
+        if (context.version == null) {
+            throw new JsonMappingException(parser, "version is missing");
```
jeandemanged, line 53: maybe maintain the list of supported versions in OperatorStrategyList, and here continue checking if the version being read is supported ?

geofjamg: We have 2 problems to implement it like this (correctly indeed):
 - `OperatorStrategy` is used in 2 contexts: as an input (`OperatorStrategyList`) and as an output (inside `SecurityAnalysisResult`). Each of these have their own version and so how could we check version in the jackson deserializer (`OperatorStrategyDeserializer`)? Using which version?
  - there is a bug in jackson 2 that forbid using serialization/deserialization context to carry the version from root deserializer (`SecurityAnalysisResultDeserializer` or `OperatorStrategyListDeserializer`) to children deserializers (like `OperatorStrategyDeserializer`): https://github.com/FasterXML/jackson-databind/issues/1991

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/OperatorStrategyDeserializer.java
geofjamg: Done

jeandemanged (approved):

flo-dup (approved):

miovd (approved):

Issue #2428, opened 2023-01-02
PowerFactory: support DGS V6 and V7


geofjamg:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
We only support DGS V5 ASCII format which is an old format.

* **What is the expected behavior?**
In addition to DGS V5, we should be able to import V6 and V7 ones.

* **What is the motivation / use case for changing the behavior?**

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

geofjamg:
This has not been done

PR #2430, opened 2023-01-02
PowerFactory: prepare for DGS 6 and 7 support

geofjamg:

#2428 

Feature

We only support DGS ASCII 5 and we are not able to detect an unsupported version of DGS

We throw an exception when DGS version is 6 or 7. `DgsHandler` has been refactored to prepare 6 and 7 support so that ID read is abstracted (not anymore a normal String attribute as not the same name depending of the version).

**Other information**:


marqueslanauja (approved):
Code works properly only if the identifier column (`ID` or `FID` column) is the first column (The new  `DataObject `is created when the `IdElementData `is read). This consideration is true for `DGS Version 5.0` as it is remarked in the documentation file. I do not know if this consideration will also be true for `Version 6.0 and 7.0`, I have not found any reference in the documentation files.

zamarrenolm (approved):

geofjamg:
> Code works properly only if the identifier column (`ID` or `FID` column) is the first column (The new `DataObject` is created when the `IdElementData` is read). This consideration is true for `DGS Version 5.0` as it is remarked in the documentation files. I do not know if this consideration will also be true for `Version 6.0 and 7.0`, I have not found any reference in the documentation files.

Thanks, good to know, we will take care of this, in next PR with support to 6 and 7. There will be many other things to adapt as for instance, matrices and vector modelling which seem to be quite different as in version 5.

Issue #2401, opened 2022-12-09
Network reduction: support splitting HVDCs and 3 windings transformers


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature

* **What is the current behavior?**

When trying to "cut" the network on an HVDC or a 3 windings transfo, with network reduction feature, we get an exception such as:
```
Reduction of HVDC lines is not supported
```

* **What is the expected behavior?**

We should support this, by simply replacing the HVDC or the transfo with a load, the same as branches.

PR #2434, opened 2023-01-06
Allow reduction of HVDC lines

HugoKulesza:

Issue #2401


iidm/iidm-reducer/src/main/java/com/powsybl/iidm/reducer/DefaultNetworkReducer.java
```diff
@@ -178,6 +182,57 @@ private Load replaceBranchByLoad(Branch<?> branch, VoltageLevel vl, Terminal ter
         return load;
     }
 
+    private void replaceHvdcLine(HvdcLine hvdcLine, VoltageLevel vl, Terminal terminal, HvdcConverterStation.HvdcType type) {
+        if (type == HvdcConverterStation.HvdcType.VSC) {
```
geofjamg, line 186: This this the invert and only if the VSC has the boolean voltageRegulatorOn to true. LCC are always replaced by a load.

HugoKulesza: Changes are done (if I understood right).

geofjamg:
> Note: the DC converter station is not removed when the HVDC line is reduced, shouldn't it be? @geofjamg @sylvlecl

Yes it should.

miovd (approved):

iidm/iidm-reducer/src/main/java/com/powsybl/iidm/reducer/DefaultNetworkReducer.java
```diff
@@ -96,20 +96,24 @@ protected void reduce(ThreeWindingsTransformer transformer) {
...
-        VoltageLevel vl2 = hvdcLine.getConverterStation2().getTerminal().getVoltageLevel();
-        if (getPredicate().test(vl1) || getPredicate().test(vl2)) {
-            throw new UnsupportedOperationException("Reduction of HVDC lines is not supported");
+        Terminal terminal1 = hvdcLine.getConverterStation1().getTerminal();
+        Terminal terminal2 = hvdcLine.getConverterStation2().getTerminal();
+        VoltageLevel vl1 = terminal1.getVoltageLevel();
+        VoltageLevel vl2 = terminal2.getVoltageLevel();
+        HvdcConverterStation<?> station1 = hvdcLine.getConverterStation1();
+        HvdcConverterStation<?> station2 = hvdcLine.getConverterStation2();
+        if (getPredicate().test(vl1)) {
```
geofjamg, line 110: We should test first if vl1 AND vl1 have to be removed ?

geofjamg: My question is just: does it work if both voltage levels have to be removed?

HugoKulesza: If both voltage levels have to be removed, test(vl1) = test(vl2) = false so we remove the hvdcLine (l.115).
However I have to add here the removal of converter stations as well.

geofjamg: maybe in that case, it is automatically removed with the voltage level?

HugoKulesza: In fact the last comment is not necessary : converter stations are in this case already removed when removing the voltage levels.

geofjamg (approved):

PR #2439, opened 2023-01-11
Bbv cache not invalidated for a retained switch when opened/closed

flo-dup:

No

Feature

Bus breaker view cache invalidate when switch opened/closed even if switch is retained

If switch is retained, bus breaker view cache not invalidated when switch opened/closed

No


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/SwitchImpl.java
```diff
@@ -66,7 +66,13 @@ public void setOpen(boolean open) {
             this.open.set(index, open);
             String variantId = network.getVariantManager().getVariantId(index);
             network.getListeners().notifyUpdate(this, "open", variantId, oldValue, open);
-            voltageLevel.invalidateCache();
+            if (isRetained()) {
```
geofjamg, line 69: Are you sure it also work for a bus/breaker topo? It seems that all switches are set to retained.
Also busview whatever the topology is should be invalidated too at voltage level scope and so at network scope.

geofjamg (changes requested):

geofjamg (approved):

PR #2440, opened 2023-01-11
CGMES : move conversion test tools to a separate module

zamarrenolm:
**Please check if the PR fulfills these requirements** *(please use `'[x]'` to check the checkboxes, or submit the PR and then click the checkboxes)*
- [X] The commit message follows our guidelines
- [X] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**Does this PR already have an issue describing the problem ?** *If so, link to this issue using `'#XXX'` and skip the rest*
Related to #2370 

**What kind of change does this PR introduce?** *(Bug fix, feature, docs update, ...)*
Feature

**What is the current behavior?** *(You can also link to an open issue here)*
Some code in CGMES conversion test packages was used by external tools through test-jar.

**What is the new behavior (if this is a feature change)?**
Common code for CGMES conversion tests is placed in a separate module.

PR #2443, opened 2023-01-17
CGMES fix network id for CGMES3; conformity test configurations v3

zamarrenolm:

Bug fix

The network identifier is not read correctly for CGMES 3 (CIM100) files. The Network identifier is read as the `FullModel` `mRID` of the EQ instance file.

The network identifier is read correctly for CGMES 3 files. The profile name that identifies the EQ instance file has been updated for the CGMES 3 in the corresponding SPARQL query.

### DIFF
--- a/cgmes/cgmes-conversion/pom.xml
+++ b/cgmes/cgmes-conversion/pom.xml
@@ -34,13 +34,6 @@
                         </manifestEntries>
                     </archive>
                 </configuration>
-                <executions>
-                    <execution>
-                        <goals>
-                            <goal>test-jar</goal>
-                        </goals>
-                    </execution>
-                </executions>
             </plugin>
         </plugins>
     </build>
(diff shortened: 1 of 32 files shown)


cgmes/cgmes-conversion/pom.xml
```diff
@@ -34,6 +34,13 @@
                         </manifestEntries>
                     </archive>
                 </configuration>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>test-jar</goal>
```
miovd, line 40: Can you explain why this is needed?

zamarrenolm: sorry, that went without wanting. I will remove this change.
I needed to kept that to be able to use test code from conversion in some of my external modules, but I am splitting the test code from conversion that is reused in #2440, following the same schema done in #2370

miovd (approved):

PR #2446, opened 2023-01-17
Allowing to reduce three windings transformers

HugoKulesza:

Issue #2401


HugoKulesza:
The idea between the reduction of three windings transformers is the following :

- if no voltage level connected is kept it is removed, if all are kept nothing change.
- if only one voltage level is kept, the transformer is removed and replaced by a load on the kept voltage level.
- if two voltage levels are kept : the third one is added to the voltage levels to keep, and the transformer is kept entirely.

To do the last point, a list of voltage level ids to keep is added to the network reducer and built using the predicate.

sylvlecl (changes requested):
Small adaptations required

sylvlecl (approved):

PR #2452, opened 2023-01-23
Clean CGMES export of tie lines/dangling lines

miovd:

No

Bug fix + cleaning + adding comments about what should be done later

- Properties used in tie lines and dangling lines during CGMES import/export are not consistent with what is done during merging of properties
- There are some export we do not wish to have (reference to tie lines when we want references to half lines)

- Properties are now consistent
- Comments have been added on future export of tie lines as two AC line segments
- Ignore export with bad references to tie lines

**Other information**:


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
```diff
@@ -223,14 +223,20 @@ private static void writePowerFlows(Network network, String cimNamespace, XMLStr
...
                 TieLine tl = (TieLine) b;
-                writePowerFlowTerminalFromProperty(tl, Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + tl.getHalf1().getId() + ".Terminal_Network",
-                        tl.getTerminal1().getP(), tl.getTerminal1().getQ(), cimNamespace, writer, context);
-                writePowerFlowTerminalFromProperty(tl, Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + tl.getHalf2().getId() + ".Terminal_Network",
-                        tl.getTerminal2().getP(), tl.getTerminal2().getQ(), cimNamespace, writer, context);
-                writePowerFlowTerminalFromAlias(tl, Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + "HALF1." + CgmesNames.TERMINAL + "_Boundary", tl.getHalf1().getBoundary().getP(), tl.getHalf1().getBoundary().getQ(), cimNamespace, writer, context);
-                writePowerFlowTerminalFromAlias(tl, Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + "HALF2." + CgmesNames.TERMINAL + "_Boundary", tl.getHalf2().getBoundary().getP(), tl.getHalf2().getBoundary().getQ(), cimNamespace, writer, context);
+                if (!b.hasProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.TERMINAL1)) { // prevent duplicates
```
zamarrenolm, line 228: the lines above have written the flows if the **alias** `TERMINAL1/2` exist. And here we are checking for a **property** named `TERMINAL1/2`. If we want to avoid duplicates, shouldn't we check for the alias instead?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
```diff
@@ -436,6 +442,10 @@ private static void writeStatus(Network network, String cimNamespace, XMLStreamW
     }
 
     private static void writeConnectableStatus(Connectable<?> connectable, String cimNamespace, XMLStreamWriter writer, CgmesExportContext context) {
+        if (connectable instanceof TieLine && connectable.hasProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.TERMINAL + "_Boundary_1")
+                && connectable.hasProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.TERMINAL + "_Boundary_2")) {
```
zamarrenolm, line 446: when writing flows we use the **alias** `TERMINAL_Boundary_1/2`, but here we are checking if the **property** with same name exists. Do we have both the alias and a property defined ?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
miovd: No because for tie lines, both aliases and properties are added (with the same values). Properties are added to be consistent with merging (merging of aliases is not handled in destructive merge or in merging view).

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
miovd: You're right! I'm adding the properties in `ACLineSegmentConversion` as they are the ones handled by merging at the moment.

zamarrenolm: debugging a test case (CGMES cas 3 conformity BE+NL) I see that the following are defined:

```
properties
  CGMES.Terminal_1
  CGMES.Terminal_2
  CGMES.Terminal_Boundary_1
  CGMES.Terminal_Boundary_2
aliasesByType
  CGMES.Terminal2
  CGMES.Terminal1
```

I am checking all the values. I think the aliases for `Terminal1/2` are created automatically and maybe we should remove them ... let me check it.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
miovd: done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
zamarrenolm: With the current version (commit [6e975a3](https://github.com/powsybl/powsybl-core/pull/2452/commits/6e975a306a807203d369502189d87a4ceaabb81c included), when we export the power flows for the tie lines we are using the aliases created automatically when preparing the export (`CGMES.Terminal1/2`, created in `CgmesExportContext::addIidmMappingsTerminal`), instead of the terminals read from CGMES input and stored in properties (`CGMES.Terminal_1`).

I think that we should either store the original terminal info in the alias (record in `CGMES.Terminal1` alias instead or in addition to saving it in `CGMES.Terminal_1` property, like we do for any generic connectable) or we avoid creating the auto aliases for tie lines in `CgmesExportContext::addIidmMappings...` (treating these lines as an special case).

I think the best option would be to set the aliases `Terminal1/2`, so we could have "uniform" treatment in other parts of the code. What do you think ?

miovd: Yes, I agree in `addIidmMappings`, if there is a `CGMES.Terminal_1/2` property, we can set it in `CGMES.Terminal1/2`

zamarrenolm: I will do the change and re-test it

miovd: ah oops didn't see your comment in time sorry, i've already commit it....

zamarrenolm: no problem, I was out.
but the way you fixed is not adding the alias by type to the tie line, because the alias already exists in the network index. I am tracing it ...

zamarrenolm: the alias is already present in the index. It is stored in the map `idByAlias`, pointing to the id of the tie line. The attempt to add as new alias by type is silently ignored because it will point to the same object (the tie line). So we are are "silently" losing the alias by the new type.

zamarrenolm: I think the aliases (at least for terminals 1) are added by id, without type, from the dangling line when the merged line is created. In the method `NetworkImpl::replaceDanglingLineByLine`, at the point `mergedLine.aliases.forEach(l::addAlias)`

zamarrenolm: Proposal:
- rename method NetworkImpl::replaceDanglingLineByLine to NetworkImpl::createTieLinesFromMergedLines, 
- and take care of re-creating the proper aliases:
  * after mergedLine.aliases.forEach(l::addAlias), 
  * remove the (untyped) aliases of terminals and re-create them as alias with proper type

miovd: Yes, there are issues in the merging of aliases but I don't think we should resolve them in this PR as it should be resolved in the improvements of the merging. It prevents doing the work twice!

zamarrenolm: I made the changes proposed. It is a small fix in the implementation of tie line creation plus extending the unit tests ...

zamarrenolm: At least we will have related information in the point where we are creating the tie lines

zamarrenolm (approved):

zamarrenolm:
Agree on avoiding the change in network implementation (that should be kept generic and without references to CGMES).

Maybe we could keep the rename of the method for creating the tie lines from the merged lines. The reference to dangling lines in the current method name is confusing for me.

I also _fixed_ a sonar suggestion for marking the classes `MergedLine` and `HalfMergedLine` as `static`, maybe it could also be kept in this PR.

PR #2454, opened 2023-01-24
Add Hvdc action

EtienneLt:

feature

add a new action to manipulate hvdc in N-1 situations

### DIFF
--- a/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/HvdcAction.java
+++ b/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/HvdcAction.java
@@ -13,12 +13,16 @@ import java.util.Optional;
 import java.util.OptionalDouble;
 
 /**
- * An action to modify hvdc parameters or how an hvdc is operated
- * parameter enabled to true correspond to ac emulation operation, false to fix active power setpoint. Note that ac
- * emulation works only with VSC converter station.
- * droop and p0 are parameters used for ac emulation only.
- * activePowerSetpoint and converterMode are for fix active power setpoint operation only. Attribute relative value should
- * be used only to define the new active power setpoint.
+ * An action to modify HVDC parameters and/or operating mode.
+ * <ul>
+ *     <li>{@code acEmulationEnabled} parameter set to true correspond to AC emulation operation,
+ *     false to fixed active power setpoint.
+ *     <li>Note that AC emulation is relevant only for VSC converter station.
+ *     <li>{@code droop} and {@code p0} parameters are used for AC emulation only.<br>
+ *     <li>{@code activePowerSetpoint} and {@code converterMode} parameters are for fixed
+ *     active power setpoint operation only. The {@code relativeValue} attribute should be used only
+ *     when defining fixed active power setpoint.
+ * </ul>
  *
  * @author Etienne Lesot <etienne.lesot@rte-france.com>
  */
--- a/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/HvdcAction.java
+++ b/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/HvdcAction.java
@@ -13,12 +13,12 @@ import java.util.Optional;
 import java.util.OptionalDouble;
 
 /**
- * an action to modify hvdc parameters
- * parameter enabled corresponds to the regulation mode true correspond to ac emulation, false to fix target
+ * An action to modify hvdc parameters or how an hvdc is operated
+ * parameter enabled to true correspond to ac emulation operation, false to fix active power setpoint. Note that ac
+ * emulation works only with VSC converter station.
  * droop and p0 are parameters used for ac emulation only.
- * activePowerSetpoint is for fix target mode only.
- * ac emulation works only with vsc technology.
- * attribute relative value if true add target p to the previous target p if false it replaces it
+ * activePowerSetpoint and converterMode are for fix active power setpoint operation only. Attribute relative value should
+ * be used only to define the new active power setpoint.
  *
  * @author Etienne Lesot <etienne.lesot@rte-france.com>
  */
@@ -38,8 +38,8 @@ public class HvdcAction extends AbstractAction {
         this(id, hvdcId, false, activePowerSetpoint, converterMode, null, null, relativeValue);
     }
 
-    public HvdcAction(String id, String hvdcId, HvdcLine.ConvertersMode converterMode, Double droop, Double p0, Boolean relativeValue) {
-        this(id, hvdcId, true, null, converterMode, droop, p0, relativeValue);
+    public HvdcAction(String id, String hvdcId, Double droop, Double p0) {
+        this(id, hvdcId, true, null, null, droop, p0, null);
     }
 
     public HvdcAction(String id, String hvdcId, boolean acEmulationEnabled, Double activePowerSetpoint, HvdcLine.ConvertersMode converterMode, Double droop, Double p0, Boolean relativeValue) {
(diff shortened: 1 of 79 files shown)


security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/HvdcAction.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public OptionalDouble getTargetP() {
+        return targetP == null ? OptionalDouble.empty() : OptionalDouble.of(targetP);
+    }
+
+    public OptionalDouble getP0() {
+        return p0 == null ? OptionalDouble.empty() : OptionalDouble.of(p0);
+    }
+
+    public boolean isEnabled() {
```
annetill, line 77: `isEnabled` to not precise enough: you must include in the name that you are talking about AC emulation.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/HvdcAction.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public HvdcLine.ConvertersMode getConverterMode() {
+        return converterMode;
+    }
+
+    public OptionalDouble getDroop() {
+        return droop == null ? OptionalDouble.empty() : OptionalDouble.of(droop);
+    }
+
+    public OptionalDouble getTargetP() {
```
annetill, line 69: What do you mean by `targetP`? You are talking about `activePowerSetpoint`?

annetill: Just a remark: for generator and load, we have to possibility to define a change in active power setpoint relatively or absolutely. Here you don't need it?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/HvdcAction.java
EtienneLt: I agree but it is related to the enabled attribute of HvdcAngleDroopActivePowerControl so I thought it was more consistent

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/HvdcAction.java
EtienneLt: targetP is the target for fix target mode (!= ac emulation mode)

EtienneLt: I don't know for fix or relative change but I think this would be better

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/HvdcAction.java
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * @author Etienne Lesot <etienne.lesot@rte-france.com>
+ */
+public class HvdcAction extends AbstractAction {
+
+    public static final String NAME = "HVDC";
+
+    private final String hvdcId;
+    private final boolean acEmulationEnabled;
+    private final Double targetP;
```
annetill, line 30: I prefer that you use the name we have in the attributes or something similar. We have `activePowerSetpoint`.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/HvdcAction.java
```diff
@@ -0,0 +1,95 @@
...
+        this.hvdcId = hvdcId;
+        this.acEmulationEnabled = acEmulationEnabled;
+        this.targetP = targetP;
+        this.converterMode = converterMode;
+        this.droop = droop;
+        this.p0 = p0;
+        this.relativeValue = relativeValue;
+    }
+
+    public static HvdcAction activateAcEmulationMode(String id, String hvdcId, HvdcLine.ConvertersMode converterMode, Double droop, Double p0, boolean relativeValue) {
```
annetill, line 55: `activateAcEmulationMode`:
- `enabled` should be at `true` ;
- `droop` could be given (or the user knows that in the model the value is correct), so it could be optional ;
- `P0` could be given (or the user knows that in the model the vale is given), so it could be optional ;
`activateActivePowerSetpointMode`:
- `targetP` + `relativeValue` could be given,  so it could be optional ;
- `converterMode` can be given or unchanged.
- for sure in that case, it means that `enabled` is set to false.

annetill (changes requested):

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/HvdcAction.java
EtienneLt: I added a constructor for both modes with only mandatory attributes is that ok ?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/HvdcAction.java
```diff
@@ -0,0 +1,104 @@
...
+    private final HvdcLine.ConvertersMode converterMode;
+    private final Double droop;
+    private final Double p0;
+    private final Boolean relativeValue;
+
+    public HvdcAction(String id, String hvdcId, Double activePowerSetpoint, HvdcLine.ConvertersMode converterMode, Boolean relativeValue) {
+        this(id, hvdcId, false, activePowerSetpoint, converterMode, null, null, relativeValue);
+    }
+
+    public HvdcAction(String id, String hvdcId, HvdcLine.ConvertersMode converterMode, Double droop, Double p0, Boolean relativeValue) {
```
annetill, line 41: This constructor is for what? If it is for Ac emulation, you don't need the converter mode and the relative value (used only for active power set point). But you need the enabled boolean.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/HvdcAction.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private final String hvdcId;
+    private final boolean acEmulationEnabled;
+    private final Double activePowerSetpoint;
+    private final HvdcLine.ConvertersMode converterMode;
+    private final Double droop;
+    private final Double p0;
+    private final Boolean relativeValue;
+
+    public HvdcAction(String id, String hvdcId, Double activePowerSetpoint, HvdcLine.ConvertersMode converterMode, Boolean relativeValue) {
```
annetill, line 37: I think here you need also the enalbled boolean to deactivate the HVDC AC emulation.

jeandemanged:
That is fine for me, though some points for discussion:
- Looking at other action types, sometimes we have public constructors (SwitchAction), sometimes builders classes (LoadAction, GeneratorAction), sometimes builder methods (here for HvdcAction), we might want to unify everything (in a separate PR) in order to expose a consistent API
- I do not have an immediate need for VSC voltage regulation settings changes via action, we may want to do it later on (or now ?): voltageRegulatorOn, reactivePowerSetpoint, voltageSetpoint

jeandemanged:
- HvdcAction uses `OptionalDouble`, like PhaseTapChangerRegulationAction
- LoadAction and GeneratorAction uses `Optional<Double>`

maybe we should try to unify these api inconsistencies, but it is not so obvious to me what is the best choice in this context, any idea @EtienneLt @geofjamg @annetill ?

jeandemanged (approved):

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/HvdcAction.java
```diff
@@ -0,0 +1,85 @@
...
+
+    public OptionalDouble getActivePowerSetpoint() {
+        return activePowerSetpoint == null ? OptionalDouble.empty() : OptionalDouble.of(activePowerSetpoint);
+    }
+
+    public OptionalDouble getP0() {
+        return p0 == null ? OptionalDouble.empty() : OptionalDouble.of(p0);
+    }
+
+    public boolean isAcEmulationEnabled() {
```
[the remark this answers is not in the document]
jeandemanged: good catch, indeed should be optional too

annetill (approved):
Static functions will be done later, after a first integration in open-loadflow.

PR #2457, opened 2023-01-25
Add options to encode/decode CGMES IDs as they are MEMIri in triplestore and can be modified during import otherwise

miovd:

No

Add option for IDs conservation in IIDM during CGMES import (and encode in CGMES export): allow to debug/test networks from scratch (not CGMES) exported in CGMES more easily

No

**Other information**:


zamarrenolm:
Do we really need to encode/decode identifiers ? CGMES master resource identifiers are strongly recommended to be 
 UUIDs, and they are mapped to rdf IDs.

In which situation do we need to handle these identifiers ? can not they be handled in a different way ? (if they are URN of coordinate systems, these are not CGMES identifiers).

miovd:
> Do we really need to encode/decode identifiers ? CGMES master resource identifiers are strongly recommended to be UUIDs, and they are mapped to rdf IDs.
> 
> In which situation do we need to handle these identifiers ? can not they be handled in a different way ? (if they are URN of coordinate systems, these are not CGMES identifiers).

We need them for CGMES export debug: for example network from scratch that we want to check (not with CGMES IDs but with IIDM IDs). But I agree for "correct"  networks, RDF Ids should be used.

zamarrenolm:
> > Do we really need to encode/decode identifiers ? CGMES master resource identifiers are strongly recommended to be UUIDs, and they are mapped to rdf IDs.
> > In which situation do we need to handle these identifiers ? can not they be handled in a different way ? (if they are URN of coordinate systems, these are not CGMES identifiers).
> 
> We need them for CGMES export debug: for example network from scratch that we want to check (not with CGMES IDs but with IIDM IDs). But I agree for "correct" networks, RDF Ids should be used.

In this case, instead of "polluting" the CGMES model / export, could we consider an approach where we transform the original IIDM network, in a similar to what is done with the anonymizer ?

miovd:
> > > Do we really need to encode/decode identifiers ? CGMES master resource identifiers are strongly recommended to be UUIDs, and they are mapped to rdf IDs.
> > > In which situation do we need to handle these identifiers ? can not they be handled in a different way ? (if they are URN of coordinate systems, these are not CGMES identifiers).
> > 
> > 
> > We need them for CGMES export debug: for example network from scratch that we want to check (not with CGMES IDs but with IIDM IDs). But I agree for "correct" networks, RDF Ids should be used.
> 
> In this case, instead of "polluting" the CGMES model / export, could we consider an approach where we transform the original IIDM network, in a similar to what is done with the anonymizer ?

To look through CGMES files, it is a bit harder, don't you think?

But you're right, the clean way is to use CGMES naming strategy and import with the dictionary but I'm talking about debug for network development.

Since encoding/decoding does not change anything in terms of what we write, I don't really think it is polluting?

zamarrenolm:
> > > > Do we really need to encode/decode identifiers ? CGMES master resource identifiers are strongly recommended to be UUIDs, and they are mapped to rdf IDs.
> > > > In which situation do we need to handle these identifiers ? can not they be handled in a different way ? (if they are URN of coordinate systems, these are not CGMES identifiers).
> > > 
> > > 
> > > We need them for CGMES export debug: for example network from scratch that we want to check (not with CGMES IDs but with IIDM IDs). But I agree for "correct" networks, RDF Ids should be used.
> > 
> > 
> > In this case, instead of "polluting" the CGMES model / export, could we consider an approach where we transform the original IIDM network, in a similar to what is done with the anonymizer ?
> 
> To look through CGMES files, it is a bit harder, don't you think?
> 
> But you're right, the clean way is to use CGMES naming strategy and import with the dictionary but I'm talking about debug for network development.
> 
> Since encoding/decoding does not change anything in terms of what we write, I don't really think it is polluting?

by "polluting" I meant the code, that gets a little bit more complicated.

yes, maybe the naming strategy would be a good point to place the required transformation.

miovd:
> > > > > Do we really need to encode/decode identifiers ? CGMES master resource identifiers are strongly recommended to be UUIDs, and they are mapped to rdf IDs.
> > > > > In which situation do we need to handle these identifiers ? can not they be handled in a different way ? (if they are URN of coordinate systems, these are not CGMES identifiers).
> > > > 
> > > > 
> > > > We need them for CGMES export debug: for example network from scratch that we want to check (not with CGMES IDs but with IIDM IDs). But I agree for "correct" networks, RDF Ids should be used.
> > > 
> > > 
> > > In this case, instead of "polluting" the CGMES model / export, could we consider an approach where we transform the original IIDM network, in a similar to what is done with the anonymizer ?
> > 
> > 
> > To look through CGMES files, it is a bit harder, don't you think?
> > But you're right, the clean way is to use CGMES naming strategy and import with the dictionary but I'm talking about debug for network development.
> > Since encoding/decoding does not change anything in terms of what we write, I don't really think it is polluting?
> 
> by "polluting" I meant the code, that gets a little bit more complicated.
> 
> yes, maybe the naming strategy would be a good point to place the required transformation.

The issue with the naming strategy (or any solution with an anonymizer) is it doesn't answer the need of easy debug... It makes it more complicated to navigate in CGMES file and recognize equipments.

Not sure the code is more complex than before, since all that is changed is the ID reading/writing method but I might be wrong.

zamarrenolm:
> > > > > > Do we really need to encode/decode identifiers ? CGMES master resource identifiers are strongly recommended to be UUIDs, and they are mapped to rdf IDs.
> > > > > > In which situation do we need to handle these identifiers ? can not they be handled in a different way ? (if they are URN of coordinate systems, these are not CGMES identifiers).
> > > > > 
> > > > > 
> > > > > We need them for CGMES export debug: for example network from scratch that we want to check (not with CGMES IDs but with IIDM IDs). But I agree for "correct" networks, RDF Ids should be used.
> > > > 
> > > > 
> > > > In this case, instead of "polluting" the CGMES model / export, could we consider an approach where we transform the original IIDM network, in a similar to what is done with the anonymizer ?
> > > 
> > > 
> > > To look through CGMES files, it is a bit harder, don't you think?
> > > But you're right, the clean way is to use CGMES naming strategy and import with the dictionary but I'm talking about debug for network development.
> > > Since encoding/decoding does not change anything in terms of what we write, I don't really think it is polluting?
> > 
> > 
> > by "polluting" I meant the code, that gets a little bit more complicated.
> > yes, maybe the naming strategy would be a good point to place the required transformation.
> 
> The issue with the naming strategy (or any solution with an anonymizer) is it doesn't answer the need of easy debug... It makes it more complicated to navigate in CGMES file and recognize equipments.
> 
> Not sure the code is more complex than before, since all that is changed is the ID reading/writing method but I might be wrong.

Ok. I buy it for as the easiest way to handle and debug the scenarios you describe. You are right that the changes in the low-level classes are "surgical" and do not have a great impact.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
```diff
@@ -401,6 +403,11 @@ private void copyStream(ReadOnlyDataSource from, DataSource to, String fromName,
             "Source for IIDM identifiers",
             SOURCE_FOR_IIDM_ID_MRID,
             List.of(SOURCE_FOR_IIDM_ID_MRID, SOURCE_FOR_IIDM_ID_RDFID));
+    private static final Parameter UNESCAPE_IDENTIFIERS_PARAMETER = new Parameter(
```
zamarrenolm, line 406: Maybe is easier to read if we change to affirmative? (`USE_ESCAPED_IDENTIFIERS`?)

zamarrenolm: or `HANDLE_ESCAPED_IDENTIFIERS`

triple-store/triple-store-api/src/main/java/com/powsybl/triplestore/api/PropertyBag.java
```diff
@@ -19,14 +21,15 @@
  */
 public class PropertyBag extends HashMap<String, String> {
 
-    public PropertyBag(List<String> propertyNames) {
-        this(propertyNames, true);
+    public PropertyBag(List<String> propertyNames, boolean unescapeIdentifiers) {
```
zamarrenolm, line 24: Maybe use a factory method instead of adding a boolean to the constructor, to state the behaviour of the property bag ? `PropertyBag.createWithEscapedIdentifiers()`

zamarrenolm: Or `PropertyBag.createForEscapedIdentifiers()` ?

triple-store/triple-store-api/src/main/java/com/powsybl/triplestore/api/PropertyBag.java
```diff
@@ -44,15 +47,15 @@ public String getLocal(String property) {
         if (value == null) {
             return null;
         }
-        return removePrefix(value, false);
+        return modifyIdentifier(value, false);
```
zamarrenolm, line 50: consider rename this method to `extractIdenfier`

triple-store/triple-store-api/src/main/java/com/powsybl/triplestore/api/PropertyBag.java
```diff
@@ -138,16 +141,21 @@ private String tabulate(String title, BiFunction<PropertyBag, String, String> ge
         return "";
     }
 
-    private String removePrefix(String s, boolean isIdentifier) {
+    private String modifyIdentifier(String s, boolean isIdentifier) {
```
zamarrenolm, line 144: what about renaming this method to `extractIdentfier` ?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
miovd: Is `ESCAPE_IDENTIFIERS` okay for you?

triple-store/triple-store-api/src/main/java/com/powsybl/triplestore/api/PropertyBag.java
miovd: Same question, is `escapeIdentifiers` okay for you?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
miovd: Is `DECODE_ESCAPED_IDENTIFIERS` okay for you? `HANDLE` is a bit vague for me

triple-store/triple-store-api/src/main/java/com/powsybl/triplestore/api/PropertyBag.java
miovd: Same thing, `decodeEscapedIdentifiers`?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
zamarrenolm: ok for me

zamarrenolm (approved):

PR #2462, opened 2023-02-02
CGMES import: active power control extension is optional

annetill:

**Other information**:

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
@@ -244,6 +244,12 @@ public class CgmesImport implements Importer {
                                 getFormat(),
                                 p,
                                 STORE_CGMES_CONVERSION_CONTEXT_AS_NETWORK_EXTENSION_PARAMETER,
+                                defaultValueConfig))
+                .setCreateActivePowerExtension(
+                        Parameter.readBoolean(
+                                getFormat(),
+                                p,
+                                CREATE_ACTIVE_POWER_CONTROL_EXTENSION_PARAMETER,
                                 defaultValueConfig));
         String namingStrategy = Parameter.readString(getFormat(), p, ID_MAPPING_FILE_NAMING_STRATEGY_PARAMETER, defaultValueConfig);
         String idMappingFilePath = Parameter.readString(getFormat(), p, ID_MAPPING_FILE_PATH_PARAMETER, defaultValueConfig);
@@ -298,6 +304,7 @@ public class CgmesImport implements Importer {
     public static final String SOURCE_FOR_IIDM_ID = "iidm.import.cgmes.source-for-iidm-id";
     public static final String STORE_CGMES_MODEL_AS_NETWORK_EXTENSION = "iidm.import.cgmes.store-cgmes-model-as-network-extension";
     public static final String STORE_CGMES_CONVERSION_CONTEXT_AS_NETWORK_EXTENSION = "iidm.import.cgmes.store-cgmes-conversion-context-as-network-extension";
+    public static final String CREATE_ACTIVE_POWER_CONTROL_EXTENSION = "iidm.import.cgmes.create-active-power-extension";
 
     public static final String SOURCE_FOR_IIDM_ID_MRID = "mRID";
     public static final String SOURCE_FOR_IIDM_ID_RDFID = "rdfID";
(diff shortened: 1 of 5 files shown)


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/SynchronousMachineConversion.java
```diff
@@ -70,10 +70,14 @@ public void convert() {
         }
         double normalPF = p.asDouble("normalPF");
         if (!Double.isNaN(normalPF)) {
-            g.newExtension(ActivePowerControlAdder.class)
-                    .withParticipate(normalPF != 0.0)
-                    .withParticipationFactor(normalPF)
-                    .add();
+            if (context.config().createActivePowerControlExtension()) {
+                g.newExtension(ActivePowerControlAdder.class)
+                        .withParticipate(normalPF != 0.0)
```
jeandemanged, line 75: we should probably change this to .withParticipate(true) always

CGMES does not have an equivalent for this participate attribute

then load flow implementations can use distribution proportional to Pmax, to targetP, to remaining margin, independently of this participate flag for iIDMs created from CGMES

annetill: Yes, I was thinking about that too... One of our test is failing because only one generator has a `participating` to true.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
```diff
@@ -298,6 +304,7 @@ private void copyStream(ReadOnlyDataSource from, DataSource to, String fromName,
     public static final String SOURCE_FOR_IIDM_ID = "iidm.import.cgmes.source-for-iidm-id";
     public static final String STORE_CGMES_MODEL_AS_NETWORK_EXTENSION = "iidm.import.cgmes.store-cgmes-model-as-network-extension";
     public static final String STORE_CGMES_CONVERSION_CONTEXT_AS_NETWORK_EXTENSION = "iidm.import.cgmes.store-cgmes-conversion-context-as-network-extension";
+    public static final String CREATE_ACTIVE_POWER_CONTROL_EXTENSION = "iidm.import.cgmes.create-active-power-extension";
```
geofjamg, line 307: iidm.import.cgmes.create-active-power-control-extension

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -810,6 +810,15 @@ public Config setStoreCgmesConversionContextAsNetworkExtension(boolean storeCgme
             return this;
         }
 
+        public boolean createActivePowerControlExtension() {
+            return createActivePowerExtension;
```
geofjamg, line 814: createActivePowerControlExtension

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -810,6 +810,15 @@ public Config setStoreCgmesConversionContextAsNetworkExtension(boolean storeCgme
...
+        }
+
+        public Config setCreateActivePowerExtension(boolean createActivePowerExtension) {
```
geofjamg, line 817: setCreateActivePowerControlExtension

geofjamg (approved):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/SynchronousMachineConversion.java
zamarrenolm: Formal documentation about CGMES participation factor (`normalPF`) only says it is used for distributed slack: _Generating unit economic participation factor.  The sum of the participation factors across generating units does not have to sum to one.  It is used for representing distributed slack participation factor. The attribute shall be a positive value or zero._

It seems correct to set the flag `participating` to false when `normalPF` is zero, but if that introduces noise, I agree on setting it always to `true`, and let the Load Flow engines decide later if the unit is effectively used for distributed slack looking the `participationFactor` value.

annetill: Thanks @zamarrenolm for these clarifications and your approval :-)

zamarrenolm (approved):

miovd (approved):

Issue #2458, opened 2023-01-25
Enhance Robustness of CGMES Importer 


AnkurArohi:
* **Do you want to request a *feature* or report a *bug*?**

Feature

* **What is the current behavior?**
Currently Powsybl leads to an exceptions and fails to build the network model if any equipment has a Voltage Level or Base level Voltage containment which is not defined.

For example if there is a Voltage LEvel referenced to an equipment but not defined there will be an error and Powsybl model will not be built.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**

Since after a basic equipment import, topology is formed and in this state the nominal voltage can be distributed later considering the connected equioments with the exception of a trafo.
This means one can always extract/distribute the nominal voltage from the topological information.

In worst case as another alternative (not the expected/optimal one) powsybl could further process the other nodes and just leave a warning message in logs so that atleast the network model is correctly built.

* **What is the motivation / use case for changing the behavior?**
Robustness enhancement

* **Please tell us about your environment:**
  - PowSyBl Version: ... 0.20.0
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

alicecaron:
Fixed by [MR](https://github.com/powsybl/powsybl-core/pull/2463)

PR #2463, opened 2023-02-02
CGMES : Implement a pre-processor that completes the CGMES model with missing containers before starting the conversion

zamarrenolm:

#2458 

Feature

CGMES import fails if containers (voltage levels) are referred but not defined in the set of input XML files.

A CGMES pre-processor scheme is proposed, following a similar approach to current post-processors.
Here we want to add plugins that intervene in the CGMES model AFTER the given CGMES data has been read but BEFORE the conversion is started.
This provides a point where we can configure a potentially time consuming process. It will be called only if the user activates it.
We inspect the initial CGMES model looking for missing data and try to complete it before the conversion starts.

**Other information**:
The implementation will build a new XML file that will contain the missing definitions.
After reading it to add its contents to the CGMES model in memory, we have to invalidate all previous caches.


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/DefineMissingContainersPreProcessor.java
```diff
@@ -0,0 +1,91 @@
...
+import java.nio.file.Paths;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * @author Luma Zamarreño <zamarrenolm at aia.es>
+ */
+@AutoService(CgmesImportPreProcessor.class)
+public class DefineMissingContainersPreProcessor implements CgmesImportPreProcessor {
```
miovd, line 29: To be discussed: not sure we want this pre-processor in cgmes-conversion. In a new module, cgmes-validation?

cgmes/cgmes-model/src/main/resources/CIM16.sparql
```diff
@@ -264,7 +264,8 @@ SELECT *
 # query: connectivityNodeContainers
 SELECT *
 WHERE {
-{ GRAPH ?graphEQ {
+# FIXME(Luma) when we fix voltage levels VL definitions may be in a different graph from the original Bay
+# { GRAPH ?graphEQ {
```
miovd, line 268: Do we want to modify this file or introduce a new file with less strict requirements?

miovd (commented):
Some remarks to open a discussion

cgmes/cgmes-model/src/main/resources/CIM16.sparql
zamarrenolm: In this case we could directly modify the file, as we are just iterating over all connectivity node containers in all graphs.

We generally preserve graph clauses to limit the search of related data and improve performance. Here we were only implicitly enforcing that bays must be in the same graph of its voltage level, and removing the graph clause should not have a big impact. To be measured, anyway.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/DefineMissingContainersPreProcessor.java
zamarrenolm: I prefer the option of using a separate module instead of a different repo, as it would keep closer all the cgmes-related code.

I would like a name that, in addition to "validation", suggests that we could also be "fixing" the model ... cgmes-fixes ? cgmes-completion ?

zamarrenolm: I've moved this pre-processor to a new module named cmges-completion.

cgmes/cgmes-model/src/main/resources/CIM16.sparql
zamarrenolm: Ok to remove the GRAPH clause, and we will keep a comment that refers that the attributes queried in this clause are found in the EQ instance files.

zamarrenolm:
When no folder is specified as parameter, the completion files are written in a temp directory created from the current `user.dir` to avoid the security risk defined in https://rules.sonarsource.com/java/RSPEC-5443.

cgmes/cgmes-completion/pom.xml
```diff
@@ -0,0 +1,97 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (c) 2017, RTE (http://www.rte-france.com)
```
miovd, line 4: *copyright 2023

cgmes/cgmes-completion/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,26 @@
...
+        </encoder>
+    </appender>
+    <root level="error">
+        <appender-ref ref="STDOUT"/>
+    </root>
+    <logger name="com.bigdata" level="error"/>
+    <logger name="com.powsybl.triplestore.TripleStore" level="error"/>
+    <logger name="com.powsybl.cgmes.triplestore.CgmesModelTripleStore" level="error"/>
+    <logger name="com.powsybl.cgmes.conversion.Conversion" level="error"/>
+    <logger name="com.powsybl.cgmes.completion" level="info"/>
```
miovd, line 25: Why is the log level info here?

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/CgmesImportPostProcessorTest.java
```diff
@@ -28,6 +29,7 @@
 /**
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
  */
+// FIXME(Luma) rename to CgmesImportProcessorsTest (we will test both pre- and post- processors)
```
miovd, line 32: You can rename it now :)

cgmes/cgmes-completion/src/main/java/com/powsybl/cgmes/completion/DefineMissingContainersPreProcessor.java
```diff
@@ -0,0 +1,216 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipOutputStream;
+
+/**
+ * @author Luma Zamarreño <zamarrenolm at aia.es>
+ */
+@AutoService(CgmesImportPreProcessor.class)
+public class DefineMissingContainersPreProcessor implements CgmesImportPreProcessor {
```
miovd, line 48: Maybe `CreateMissingContainersPreProcessor` is more explicit?

miovd (commented):
I'm wondering if this pre-processor's place is in core at all.... The feature of CGMES pre-processors in itself is very nice and a good idea! But this one is, I think, about a use-case that is too specific to be in core. Maybe it can be in another repo? @annetill do you have any opinion?

cgmes/cgmes-completion/src/test/resources/logback-test.xml
zamarrenolm: For the unit tests, to have easy access to the relevant info about the run (location of the output file, list of missing voltage levels in the input) ...

cgmes/cgmes-completion/src/main/java/com/powsybl/cgmes/completion/DefineMissingContainersPreProcessor.java
zamarrenolm: done

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/CgmesImportPostProcessorTest.java
zamarrenolm: Yes, I forgot to do it. I have renamed the test class as `CgmesImportPreAndPostProcessorsTest`

zamarrenolm:
Generic scheme for pre-processors has been moved to #2513. Once it is merged in main, we will leave here only the specific code for completing CGMES model with missing containers using the pre-processor mechanism.

cgmes/cgmes-completion/src/main/java/com/powsybl/cgmes/completion/CreateMissingContainersPreProcessor.java
```diff
@@ -0,0 +1,216 @@
...
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipOutputStream;
+
+/**
+ * @author Luma Zamarreño <zamarrenolm at aia.es>
+ */
+@AutoService(CgmesImportPreProcessor.class)
+public class CreateMissingContainersPreProcessor implements CgmesImportPreProcessor {
+
+    public static final String NAME = "CreateMissingContainers";
```
miovd, line 50: For homogenisation purposes, you can simply name it `missingContainers`

cgmes/cgmes-completion/src/main/java/com/powsybl/cgmes/completion/CreateMissingContainersPreProcessor.java
```diff
@@ -0,0 +1,216 @@
...
+
+    private static String nameFor(CgmesModel cgmes) {
+        try {
+            return new URI(cgmes.getBasename()).getAuthority();
+        } catch (URISyntaxException e) {
+            return cgmes.modelId();
+        }
+    }
+
+    private void buildFixesOnFolder(CgmesModel cgmes, String basename, Path fixesFolder) {
```
miovd, line 76: Can this be static?

cgmes/cgmes-completion/src/main/java/com/powsybl/cgmes/completion/CreateMissingContainersPreProcessor.java
```diff
@@ -0,0 +1,216 @@
...
+    private static Set<String> findMissingVoltageLevels(CgmesModel cgmes) {
+        // check missing CN containers
+        Set<String> defined = cgmes.connectivityNodeContainers().stream().map(c -> c.getId(CgmesNames.CONNECTIVITY_NODE_CONTAINER)).collect(Collectors.toSet());
+        Set<String> referred = cgmes.connectivityNodes().stream().map(c -> c.getId(CgmesNames.CONNECTIVITY_NODE_CONTAINER)).collect(Collectors.toSet());
+        return referred.stream().filter(c -> !defined.contains(c)).collect(Collectors.toSet());
+    }
+
+    private static void buildZipFileWithFixes(Set<String> missingVoltageLevels, Path fixesFile, String basename) {
+        Network network = NetworkFactory.findDefault().createNetwork("empty", "CGMES");
+        // TODO(Luma) should we ensure that context CIM version uses the same version of the input files ?
```
miovd, line 126: Can you elaborate this `TODO`?

cgmes/cgmes-completion/src/main/java/com/powsybl/cgmes/completion/CreateMissingContainersPreProcessor.java
```diff
@@ -0,0 +1,216 @@
...
+
+        // TODO(Luma) In a first approach,
+        // we do not have additional information about the voltage level,
+        // we create a different substation and base voltage for every missing voltage level
+        String voltageLevelName = voltageLevelId + " VL";
+        String substationId = CgmesExportUtil.getUniqueId();
+        String substationName = voltageLevelId + "SUB for missing VL " + voltageLevelId;
+        String baseVoltageId = CgmesExportUtil.getUniqueId();
+        // An arbitrary, abnormal value
+        double nominalV = 10000.0;
```
miovd, line 170: Maybe put it as parametrisable? At minimum, it should be a constant to be clear, not hidden in the code

cgmes/cgmes-completion/src/main/java/com/powsybl/cgmes/completion/CreateMissingContainersPreProcessor.java
```diff
@@ -0,0 +1,216 @@
...
+            } catch (IOException e) {
+                LOG.warn("Failed to create a temp directory");
+            }
+        }
+    }
+
+    static class RegionContainers {
+        String subGeographicalRegionId;
+        String geographicalRegionId;
+    }
```
miovd, line 215: Can be private

miovd (changes requested):

cgmes/cgmes-completion/src/main/java/com/powsybl/cgmes/completion/CreateMissingContainersPreProcessor.java
zamarrenolm: But for example the IIDM modification post-processor that replaces tie lines by lines starts also with a verb (`replaceTieLinesByLines`). I prefer if we keep the action in the name. What do you think?

cgmes/cgmes-completion/src/main/java/com/powsybl/cgmes/completion/CreateMissingContainersPreProcessor.java
zamarrenolm: yes, done.

cgmes/cgmes-completion/src/main/java/com/powsybl/cgmes/completion/CreateMissingContainersPreProcessor.java
zamarrenolm: We have read CGMES files in a given version.
We should export CGMES files in the same version.
When the network is created applying conversion from CGMES files, the version of input data is stored in the extension `CimCharacteristics`, and when an export context is created from a `Network`, the version to use in the export is the same one that was read, unless explicitly changed.
Here we are creating an export context with an empty `Network` that will not carry information about specific CIM version to use, so the default value will be used (CIM16).

I will set the CIM version in the context explicitly, based on the version of the input files.

cgmes/cgmes-completion/src/main/java/com/powsybl/cgmes/completion/CreateMissingContainersPreProcessor.java
zamarrenolm: It is not very useful to parametrise it, the missing voltage levels may correspond to different points of the network. I added a constant.

cgmes/cgmes-completion/src/main/java/com/powsybl/cgmes/completion/CreateMissingContainersPreProcessor.java
zamarrenolm: done

cgmes/cgmes-completion/pom.xml
zamarrenolm: updated

cgmes/cgmes-completion/src/test/resources/logback-test.xml
zamarrenolm: log level updated to `error`

cgmes/cgmes-completion/src/test/java/com/powsybl/cgmes/completion/CgmesCompletionTest.java
```diff
@@ -0,0 +1,59 @@
...
+        //
+        // For tests, it is not possible to put a reference to a Jimfs folder,
+        // Even if we write the uri of the folder, and read it back as an uri,
+        // the default file system is used to interpret it
+        //
+        // An alternative would be to write in the config a temp folder that we create here in the tests,
+        // But that means writing in a resource file (the actual path of the temp folder), it is dirty.
+        //
+        // Instead, the CGMES completion processor creates itself a temp folder when no parameter is received.
+        // We won't receive directly the file with the fixes, but we do not care.
```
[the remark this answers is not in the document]
zamarrenolm: I did not want to use Jimfs directly in the preprocessor: in the rest of PowSyBl core this library is only included in test code. It will add a new dependency in the framework.

The problem is that we can not setup a Jimfs folder inside the test code that is then used in the preprocessor: no way to put a reference in the config file.

zamarrenolm:
Removed the "silent" write if the output folder for fixes is not defined in the configuration.
Adjusted the test to use Jimfs in this scenario.
When an output folder for the files containing the fixes is defined in the configuration: if it is an absolute path it is taken as it is. If it is relative, it is assumed to be relative the the config dir. This allows to create a test config on Jimfs and the pre-processor to use it.

cgmes/cgmes-completion/src/main/java/com/powsybl/cgmes/completion/CreateMissingContainersPreProcessor.java
```diff
@@ -0,0 +1,245 @@
...
+import java.nio.file.Paths;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipOutputStream;
+
+/**
+ * @author Luma Zamarreño <zamarrenolm at aia.es>
```
miovd, line 49: Maybe a (light) javadoc, just to explain what is done in this pre-processor?

cgmes/cgmes-completion/src/main/java/com/powsybl/cgmes/completion/CreateMissingContainersPreProcessor.java
miovd: I think there is a difference between modifications (which state the action they do) and processors (which state the context, particularly for CGMES processors) but both are okay in my opinion but it should start with a low case

zamarrenolm: The name now starts with lower case, `createMissingContainers`.
I think of pre/post-processors as specialized tools carrying out specific tasks before/after CGMES import, so for me it is ok if their name starts with a verb.

cgmes/cgmes-completion/src/main/java/com/powsybl/cgmes/completion/CreateMissingContainersPreProcessor.java
```diff
@@ -0,0 +1,216 @@
...
+        String fixesFolderName = Parameter.readString("CGMES", null, FIXES_FOLDER_NAME_PARAMETER, defaultValueConfig);
+        if (fixesFolderName != null) {
+            buildFixesOnFolder(cgmes, basename, Path.of(fixesFolderName));
+        } else {
+            LOG.warn("Missing fixes folder parameter {}, attempt to create a temp directory in user working directory.", FIXES_FOLDER_NAME_PARAMETER.getName());
+            Path current = Path.of(System.getProperty("user.dir"));
+            String prefix = "cgmes-fixes-for-missing-containers";
+            boolean debug = false;
+            try (WorkingDirectory workingDirectory = new WorkingDirectory(current, prefix, debug)) {
+                buildFixesOnFolder(cgmes, basename, workingDirectory.toPath());
```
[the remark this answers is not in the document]
zamarrenolm: I was able to solve the issue (and do proper testing) without the need of a explicit additional parameter for `FileSystem`.

When we obtain the folder parameter from config:

- if it is an absolute path the default file system is used.
- If the path is relative, it is assumed relative to the platform config dir.

In the tests, we create a platform config that lives in the Jimfs file system, following the same schema applied in other unit tests.

cgmes/cgmes-completion/src/main/java/com/powsybl/cgmes/completion/CreateMissingContainersPreProcessor.java
zamarrenolm: added a small javadoc

PR #2464, opened 2023-02-08
Remove voltage level and its equipments

etiennehomer:

Feature: this PR implements the 'remove voltage level' modification. When the modification is applied, the voltage level, its equipments and the branches it is connected to are removed.

### DIFF
--- a/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/RemoveVoltageLevel.java
+++ b/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/RemoveVoltageLevel.java
@@ -3,6 +3,7 @@
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
  */
 package com.powsybl.iidm.modification.topology;
 
--- a/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/RemoveVoltageLevel.java
+++ b/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/RemoveVoltageLevel.java
@@ -18,7 +18,6 @@ import java.util.Objects;
 
 import static com.powsybl.iidm.modification.topology.ModificationReports.notFoundVoltageLevelReport;
 
-
 /**
  * @author Etienne Homer <etienne.homer at rte-france.com>
  */
--- a/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/RemoveVoltageLevel.java
+++ b/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/RemoveVoltageLevel.java
@@ -1,0 +1,59 @@
+/**
+ * Copyright (c) 2023, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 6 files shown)


iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/RemoveVoltageLevel.java
```diff
@@ -0,0 +1,57 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(RemoveVoltageLevel.class);
+
+    private final String voltageLevelId;
+
+    public RemoveVoltageLevel(String voltageLevelId) {
+        this.voltageLevelId = Objects.requireNonNull(voltageLevelId);
+    }
+
+    @Override
+    public void apply(Network network, boolean throwException, ComputationManager computationManager, Reporter reporter) {
```
annetill, line 34: To everybody:
- This feature (but we have this issue also for a classical connectable remove) does not handle extensions? Or it is hidden in the remove of each connectable? A `VoltageLevel` can also have extensions. What is our position on that?
- I have an issue with the remove of connectables in general, that we do here. In some network, we are using `regulatingTerminal` that can be remote. So I think that we can remove a connectable which terminal can be used for a remote regulation.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/RemoveVoltageLevel.java
```diff
@@ -0,0 +1,57 @@
+/**
+ * Copyright (c) 2023, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
```
miovd, line 6: Can you add the SPDX? (example :
```/**
 * Copyright (c) 2023, RTE (http://www.rte-france.com/)
 */)

File: iidm/iidm-modification/src/test/java/com/powsybl/iidm/modification/topology/RemoveVoltageLevelTest.java [hunk 3]
```diff
@@ -0,0 +1,77 @@
...
+
+import java.util.HashSet;
+import java.util.Set;
+
+import static org.junit.Assert.*;
+
+/**
+ * @author Etienne Homer <etienne.homer at rte-france.com>
+ */
+public class RemoveVoltageLevelTest {
```
miovd, line 26: I think you can add a round trip test on an xml test just to test the rest of the network has not been touched by the modification

miovd (changes requested):

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/RemoveVoltageLevel.java
miovd: About the first point, it is already handled by the `remove` method!

About the second point: it is indeed an issue, to be discussed in another PR about removal in general.

iidm/iidm-modification/src/test/java/com/powsybl/iidm/modification/topology/RemoveVoltageLevelTest.java
etiennehomer: I added a round trip test for node breaker and bus breaker topology. Does it make sense to test both ?

miovd (approved):

geofjamg (approved):

PR #2465, opened 2023-02-13
Replace AssertionError by IllegalStateException.

colinepiloquet:

No.

Replace all AssertionErrors by IllegalStateExceptions.


shortcircuit-api/src/test/java/com/powsybl/shortcircuit/json/JsonFaultParametersTest.java
```diff
@@ -44,7 +45,7 @@ public void readUnexpectedField() throws IOException {
         Files.copy(getClass().getResourceAsStream("/FaultParametersFileInvalid.json"), fileSystem.getPath("/FaultParametersFileInvalid.json"));
 
         Path path = fileSystem.getPath("/FaultParametersFileInvalid.json");
-        AssertionError e = assertThrows(AssertionError.class, () -> FaultParameters.read(path));
-        assertEquals("Unexpected field: unexpected", e.getMessage());
+        UncheckedIOException e = assertThrows(UncheckedIOException.class, () -> FaultParameters.read(path));
```
colinepiloquet, line 48: @flo-dup, somehow here the transmission of the exception is different and the message is less clear. Do you know if there is a better way to do it? (it is the same problem with every exception message that I had to change)

flo-dup: I don't know if the message is less clear? but if you want to get rid of his, you could disable the exception wrapping by calling, when creating your objectMapper:
```java
objectMapper.disable(DeserializationFeature.WRAP_EXCEPTIONS)
```

miovd: I'm not sure I agree with this modification (which is not even useful now since the tests do not fail anymore with the passage to Junit5). The wrapped exception brings other information about the location of issues, are we sure we want to get rid of it?

flo-dup: @colinepiloquet I agree with @miovd, id adds more information about the exception, so we should probably keep it. Even if in your case it's not that useful, for all other exceptions which occur there you won't have the full context.

miovd:
@colinepiloquet Don't pay attention to the sonarcloud issues, but can you fix the DCO error?

miovd (approved):

flo-dup (commented):
Just wondering about two things:
- Weren't we concerned by the `assert` in code?
- You replaced all the `AssertionError`, wasn't there at least one which was justified?

miovd:
> Just wondering about two things:
> 
>     * Weren't we concerned by the `assert` in code?

Indeed, this should be treated as well.

>     * You replaced all the `AssertionError`, wasn't there at least one which was justified?

I'm pretty sure that all our `AssertionError` were not used wisely, only for forbidden values (functional).

flo-dup (changes requested):

flo-dup (approved):

PR #2472, opened 2023-02-17
CGMES Importer: optional creation of terminals fictitious switches

jeandemanged:

No

Feature

For node-breaker models, fictitious switches are always created for disconnected terminals

option `iidm.import.cgmes.create-fictitious-switches-for-disconnected-terminals`
defaults to false (=new behavior)
if set to true the fictitious switches are created (=legacy behavior)

No

**Other information**:
From slack discussions Feb 17th, 2023, #singlelinediagram channel

quick test on CGMES CAS 2.0 MicroGrid Type4 NB

option true (legacy behavior)
image

option false (default, new behavior)
image


annetill:
@zamarrenolm has started the work too...

zamarrenolm:
> @zamarrenolm has started the work too...

I was aware of the proposal @annetill, @jeandemanged shared it in the Slack #cgmes channel.

It looks ok for me to assume that for node/breaker network all equipment modeled must have proper switches with the right connection status. If a user knows that there could be inconsistencies (or equipment directly connected to busbars without switches) in its model, then the behaviour for creating fictitious switches could be activated through configuration.

The scope of the change is also ok for me: I don't think that we should analyze the connectivity for every disconnected terminal and check if we should open one (or more) switches: we may have situations where no such switches exist (equipment directly connected to busbars), many switches must be opened, or other equipment could be affected.

annetill:
@zamarrenolm @miovd @jeandemanged these fictitious switches creation was useful at the beginning, and maybe we will still found inconsistencies, but as a general philosophy in the organisation, we don't like to create fictitious elements. We can make some really complex developments to avoid having voltage levels at boundary nodes. So it is for me in contradiction with a `true` default behavior for this parameter.

zamarrenolm:
> @zamarrenolm @miovd @jeandemanged these fictitious switches creation was useful at the beginning, and maybe we will still found inconsistencies, but as a general philosophy in the organisation, we don't like to create fictitious elements. We can make some really complex developments to avoid having voltage levels at boundary nodes. So it is for me in contradiction with a `true` default behavior for this parameter.

Agree on by default NOT creating fictitious switches

zamarrenolm (approved):
Verified against "real-world" test cases, the change is ok for me.

A few observations after the review:

- Some modelling authorities may indicate their model is node/breaker but only switches between busbar sections are created, and all equipment in general are connected to busbars directly. For these situations the analyst will find interesting to create the fictitious switches.
- In a few cases, loads are modelled as connected directly to one end of a transformer, with the transformer itself having a breaker. This kind of modelling may be realistic (the only way to disconnected part of the load is disconnect the distribution transformer), and does not need creation of fictitious switches.
- Sometimes, even if there is a proper switching element that is open, its terminals themselves are also marked as disconnected. With this change we avoid the creation of redundant fictitious elements.

jeandemanged:
@zamarrenolm thanks!

> Some modelling authorities may indicate their model is node/breaker but only switches between busbar sections are created, and all equipment in general are connected to busbars directly.

Correct, reffered as "BB style model, modelling with CN" [here line 1233](https://eepublicdownloads.entsoe.eu/clean-documents/CIM_documents/Grid_Model_CIM/CGM_BUILDING_PROCESS_IMPLEMENTATION_GUIDE_AC_PART_1.3.pdf)

I thought that the adding of fictitous switches was hiding the fact that powsybl has no equivalent for cim:Equipment.inService (not to be confused with cim:SvStatus.inService). However after check cim:Equipment.inService is not part of CGMES 2.4.15 (but is part of CGMES 3.0).

I am still not 100% sure whether this PR is correct or not.
- Maybe we should use cim:Terminal.connected also for NB models after all.
- We may instead use the [QOCdC rule SwitchOpenVsConnected line 4014](https://eepublicdownloads.azureedge.net/clean-documents/digital/QualityOfCGMESdatasetsAndCalculations_v3_3.pdf) stating that Switches terminal are supposed to be always connected, and skip creation of fictitous switches _only for Switches_?

miovd:
> @zamarrenolm thanks!
> 
> > Some modelling authorities may indicate their model is node/breaker but only switches between busbar sections are created, and all equipment in general are connected to busbars directly.
> 
> Correct, reffered as "BB style model, modelling with CN" [here line 1233](https://eepublicdownloads.entsoe.eu/clean-documents/CIM_documents/Grid_Model_CIM/CGM_BUILDING_PROCESS_IMPLEMENTATION_GUIDE_AC_PART_1.3.pdf)
> 
> I thought that the adding of fictitous switches was hiding the fact that powsybl has no equivalent for cim:Equipment.inService (not to be confused with cim:SvStatus.inService). However after check cim:Equipment.inService is not part of CGMES 2.4.15 (but is part of CGMES 3.0).
> 
> I am still not 100% sure whether this PR is correct or not.
> 
>     * Maybe we should use cim:Terminal.connected also for NB models after all.
> 
>     * We may instead use the [QOCdC rule SwitchOpenVsConnected line 4014](https://eepublicdownloads.azureedge.net/clean-documents/digital/QualityOfCGMESdatasetsAndCalculations_v3_3.pdf) stating that Switches terminal are supposed to be always connected, and skip creation of fictitous switches _only for Switches_?

It is certainly safer :) @zamarrenolm do you think only checking if the terminal is on a switch would affect the performance a lot? I think it is should be okay but I would like to have your opinion.

miovd:
> @zamarrenolm thanks!
> 
> > Some modelling authorities may indicate their model is node/breaker but only switches between busbar sections are created, and all equipment in general are connected to busbars directly.
> 
> Correct, reffered as "BB style model, modelling with CN" [here line 1233](https://eepublicdownloads.entsoe.eu/clean-documents/CIM_documents/Grid_Model_CIM/CGM_BUILDING_PROCESS_IMPLEMENTATION_GUIDE_AC_PART_1.3.pdf)
> 
> I thought that the adding of fictitous switches was hiding the fact that powsybl has no equivalent for cim:Equipment.inService (not to be confused with cim:SvStatus.inService). However after check cim:Equipment.inService is not part of CGMES 2.4.15 (but is part of CGMES 3.0).
> 
> I am still not 100% sure whether this PR is correct or not.
> 
>     * Maybe we should use cim:Terminal.connected also for NB models after all.
> 
>     * We may instead use the [QOCdC rule SwitchOpenVsConnected line 4014](https://eepublicdownloads.azureedge.net/clean-documents/digital/QualityOfCGMESdatasetsAndCalculations_v3_3.pdf) stating that Switches terminal are supposed to be always connected, and skip creation of fictitous switches _only for Switches_?

@jeandemanged are you also sure that it actually reduces the number of switches in real case files? @zamarrenolm @annetill

jeandemanged:
Hi @miovd 

> @jeandemanged are you also sure that it actually reduces the number of switches in real case files?

yes, see image above from manual test, and I also had to enable the legacy behavior for a couple of existing unit tests that were relying on on these fictitious switches creation so that they do not break.

miovd (approved):

miovd (commented):
As you said, you can make the distinction of terminals of switches or not, I put "approved" as I approved the conclusion of the discussion!

zamarrenolm:
After review your suggestion about switches, I think we can make an improvement to avoid creating more fictitious switches than required:

Currently, a fictitious switch is created if the terminal is not connected and the equipment is considered as "connected". But when creating nodes for ends of switches, we are not passing its "open" attribute to the method. So we don't know if the switch is "connected" or not, resulting in always creating the fictitious switch if the terminal is disconnected, even if its switch is already open.

I will try to pass the "open" state of the switch to the node mapping method. It was already prepared to handle it, I think we just have to pass the data.

zamarrenolm:
> After review your suggestion about switches, I think we can make an improvement to avoid creating more fictitious switches than required:
> 
> Currently, a fictitious switch is created if the terminal is not connected and the equipment is considered as "connected". But when creating nodes for ends of switches, we are not passing its "open" attribute to the method. So we don't know if the switch is "connected" or not, resulting in always creating the fictitious switch if the terminal is disconnected, even if its switch is already open.
> 
> I will try to pass the "open" state of the switch to the node mapping method. It was already prepared to handle it, I think we just have to pass the data.

I have prepared this change, @jeandemanged, @miovd. It will reduce the number of fictitious switches created. For the reference example at the beginning of the conversation, no fictitious switches will be created as the breaker is already open. If you agree with it I can push the commit.

jeandemanged:
@zamarrenolm the [QOCdC rule SwitchOpenVsConnected line 4014](https://eepublicdownloads.azureedge.net/clean-documents/digital/QualityOfCGMESdatasetsAndCalculations_v3_3.pdf) states that Switche terminals must be always connected independently of open status.

I think you can skip creation to fictitious switches for switches always, independently of the switch open status.

miovd:
> @zamarrenolm the [QOCdC rule SwitchOpenVsConnected line 4014](https://eepublicdownloads.azureedge.net/clean-documents/digital/QualityOfCGMESdatasetsAndCalculations_v3_3.pdf) states that Switche terminals must be always connected independently of open status.
> 
> I think you can skip creation to fictitious switches for switches always, independently of the switch open status.
> 
> and yes, CGMES CAS 2.0 MicroGrid Type4 NB (picture above) violates the QOCdC rule, but that is a different issue ...

I agree with that. So in this case, the PR must be modified to:
- always import fictitious switches for terminals of equipment that are not real switches
- do not import fictitious switches for terminals of real switches when `CREATE_FICTITIOUS_SWITCHES_FOR_DISCONNECTED_TERMINALS_PARAMETER` is false @jeandemanged

jeandemanged:
> > @zamarrenolm the [QOCdC rule SwitchOpenVsConnected line 4014](https://eepublicdownloads.azureedge.net/clean-documents/digital/QualityOfCGMESdatasetsAndCalculations_v3_3.pdf) states that Switche terminals must be always connected independently of open status.
> > I think you can skip creation to fictitious switches for switches always, independently of the switch open status.
> > and yes, CGMES CAS 2.0 MicroGrid Type4 NB (picture above) violates the QOCdC rule, but that is a different issue ...
> 
> I agree with that. So in this case, the PR must be modified to:
> 
> * always import fictitious switches for terminals of equipment that are not real switches
> * do not import fictitious switches for terminals of real switches when `CREATE_FICTITIOUS_SWITCHES_FOR_DISCONNECTED_TERMINALS_PARAMETER` is false @jeandemanged

i might disaggree, I would rather do, `CREATE_FICTITIOUS_SWITCHES_FOR_DISCONNECTED_TERMINALS_PARAMETER` :
* false: **never** create fictitious switches
* true: create fictitious switches for disconnected terminals, **except** Switch terminals

miovd:
> > > @zamarrenolm the [QOCdC rule SwitchOpenVsConnected line 4014](https://eepublicdownloads.azureedge.net/clean-documents/digital/QualityOfCGMESdatasetsAndCalculations_v3_3.pdf) states that Switche terminals must be always connected independently of open status.
> > > I think you can skip creation to fictitious switches for switches always, independently of the switch open status.
> > > and yes, CGMES CAS 2.0 MicroGrid Type4 NB (picture above) violates the QOCdC rule, but that is a different issue ...
> > 
> > 
> > I agree with that. So in this case, the PR must be modified to:
> > 
> > * always import fictitious switches for terminals of equipment that are not real switches
> > * do not import fictitious switches for terminals of real switches when `CREATE_FICTITIOUS_SWITCHES_FOR_DISCONNECTED_TERMINALS_PARAMETER` is false @jeandemanged
> 
> @miovd i might disaggree, I would rather do, `CREATE_FICTITIOUS_SWITCHES_FOR_DISCONNECTED_TERMINALS_PARAMETER` :
> 
>     * false: **never** create fictitious switches
> 
>     * true: create fictitious switches for disconnected terminals, **except** Switch terminals

@jeandemanged I don't agree because it is not contrary to standard to express topology of equipment (not switches) with only terminals' connected status so I think it might be very dangerous to never create fictitious switches

Also I think we must keep an option that keeps the current exhaustive behavior, if only for debug or networks in development.

PR #2478, opened 2023-02-26
Fix Java 19 build

geofjamg:

No

Bug fix

We cannot build using a JDK 19

Mockito core and Groovy need to be updated bu there is still an issue with Mockito when mocking java.util.ByeBuffer.

**Other information**:


geofjamg:
https://github.com/mockito/mockito/issues/2927

geofjamg:
So the mockito issue in time series api unit tests won't be solved as it is not recommended to mock JDK classes.
We have to redesign theses tests to avoid directly mock java.util.ByteBuffer.

geofjamg:
It looks like jacoco is not working too. Maybe need to be updated?

geofjamg:
> It looks like jacoco is not working too. Maybe need to be updated?

Confirmed. We need to update powsybl-parent to jacoco 0.8.8

time-series/time-series-api/src/test/java/com/powsybl/timeseries/AbstractBigBufferTest.java
```diff
@@ -0,0 +1,58 @@
...
+        tempdir = Files.createTempDirectory("powsybltimeseriestest");
+        channels = new ArrayList<>();
+    }
+
+    @AfterEach
+    void after() throws Exception {
+        for (FileChannel channel : channels) {
+            channel.close();
+        }
+        Files.delete(tempdir);
```
geofjamg, line 55: @jonenst On Windows test failed because directory is not empty. So a real file has been created on the file system even in private map mode.

geofjamg:
@jonenst thinking again not sure this is the right testing strategy. I tested on a 4Go Windows virtual machine and all tests failed with 
```java
Caused by: java.io.IOException: Le fichier de pagination est insuffisant pour terminer cette opération
	at java.base/sun.nio.ch.FileChannelImpl.map0(Native Method)
	at java.base/sun.nio.ch.FileChannelImpl.mapInternal(FileChannelImpl.java:1305)
	at java.base/sun.nio.ch.FileChannelImpl.map(FileChannelImpl.java:1181)
	at com.powsybl.timeseries.AbstractBigBufferTest.testAllocator(AbstractBigBufferTest.java:39)
```
We should not require so much memory for running unit test and not really allocate so bug buffers.
Maybe we should revert back to old solution but not mocking ByteBuffer directly but instead introducing an intermediate abstraction (a wrapper of ByteBuffer) that will be the target of the mock.

jonenst (approved):

PR #2489, opened 2023-03-07
Create severity filter in Reporter API

caioluke:

Feature

It is currently not possible to filter reports based on their severity

This PR creates a method to be able to filter out reports according to a severity threshold in the Reporter API

**Other information**:

### DIFF
--- a/commons/src/main/java/com/powsybl/commons/reporter/Reporter.java
+++ b/commons/src/main/java/com/powsybl/commons/reporter/Reporter.java
@@ -131,6 +131,10 @@ public interface Reporter {
      */
     void report(Report report);
 
+    boolean isEnabled(TypedValue severityType);
+
+    void setSeverityThreshold(TypedValue severityThreshold);
+
     class NoOpImpl extends AbstractReporter {
         public NoOpImpl() {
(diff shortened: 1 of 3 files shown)


flo-dup (changes requested):
This is not what @geofjamg and @jeandemanged had in mind so I think we should have discussed this in a new issue before opening the PR

flo-dup:
Thanks for this pull request, but as discussed in TSC with @geofjamg and @jeandemanged,
- Report severity is not a very mature concept in current Reporter API. It used to be just a tag with no existing default value, as we wanted to have something richer than the 5 standard levels of logging. Later we added the standard log levels, but we still can add any tag, so ordering and threshold don't make so much sense.
- It makes more sense, at least in powsybl/powsybl-open-loadflow#740 usecase, to decide to bypass reports based on which part of the code is concerned, rather than based on the severity of the corresponding reports: we want either all the reports or none.
- There's no other identified usecase so far of bypassing some reports based on the severity (or not)

Hence, closing this pull request. Feel free of course to reopen it if one or more of the mentioned reasons are not valid any more.

PR #2493, opened 2023-03-09
Matpower importer: Support lines with different nominal voltage at both ends

marqueslanauja:

Feature

Conversion of electric characteristics for lines with different nominal voltage at both ends is incorrect. De-normalization from per-unit quantities to engineering units must take into account both nominal voltages.

De-normalization from per-unit quantities to engineering units for lines with different nominal voltage at both ends has been fixed. We have decided to keep them as transmission lines instead of transformers. If modeled as transformers, end nodes should have to be grouped in the same substation.

**Other information**:


zamarrenolm:
We have verified that OLF (1.0.1) solves correctly all Matpower samples taking into account the base voltages and performing the de-normalization implemented in this Pull Request. As expected, the solutions obtained are exactly the same ones that are obtained if base voltages are completely ignored in the IIDM model.

zamarrenolm (approved):

annetill (approved):

zamarrenolm (approved):
Verified running reference cases against OLF

PR #2497, opened 2023-03-10
Create action to change shunt compensator position

miovd:

No

Feature: create action to change shunt compensator position

No

**Other information**:

### DIFF
--- a/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/ShuntCompensatorPositionAction.java
+++ b/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/ShuntCompensatorPositionAction.java
@@ -1,0 +1,38 @@
+/**
+ * Copyright (c) 2023, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+package com.powsybl.security.action;
+
+/**
+ * @author Miora Vedelago <miora.ralambotiana at rte-france.com>
+ */
+public class ShuntCompensatorPositionAction extends AbstractAction {
+
+    public static final String NAME = "SHUNT_COMPENSATOR_POSITION";
+
+    private final String shuntCompensatorId;
+    private final int sectionCount;
+
+    ShuntCompensatorPositionAction(String id, String shuntCompensatorId, int sectionCount) {
+        super(id);
+        this.shuntCompensatorId = shuntCompensatorId;
+        this.sectionCount = sectionCount;
+    }
+
+    @Override
+    public String getType() {
+        return NAME;
+    }
+
+    public String getShuntCompensatorId() {
+        return shuntCompensatorId;
+    }
+
+    public int getSectionCount() {
+        return sectionCount;
+    }
+}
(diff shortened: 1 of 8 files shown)


security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/ShuntCompensatorPositionAction.java
```diff
@@ -0,0 +1,38 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+package com.powsybl.security.action;
+
+/**
+ * @author Miora Vedelago <miora.ralambotiana at rte-france.com>
+ */
+public class ShuntCompensatorPositionAction extends AbstractAction {
```
annetill, line 13: Position or section?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/ShuntCompensatorPositionActionBuilder.java
```diff
@@ -0,0 +1,43 @@
...
+
+    private String id;
+    private String shuntCompensatorId;
+    private Integer sectionCount = null;
+
+    public ShuntCompensatorPositionAction build() {
+        if (sectionCount == null) {
+            throw new IllegalArgumentException("sectionCount in undefined");
+        }
+        if (sectionCount < 0) {
```
annetill, line 23: I think this action could lead to a zero section count, it is a way to easily disconnect the shunt compensator. NB: fix typo.

annetill (changes requested):

security-analysis/security-analysis-api/src/test/resources/ActionFileTestV1.0.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version" : "1.0",
```
flo-dup, line 2: Why did you put this file in 1.1? It's now exactly the same file as ActionFileTest.json

flo-dup (changes requested):

security-analysis/security-analysis-api/src/test/resources/ActionFileTestV1.0.json
annetill: @miovd you cannot change "value" to "tapPosition" because it is not always a tap position. When relative, it is a delta. Value has been chosen in consequence.  I don't understand why we have this change...

miovd: Indeed, my mistake...

annetill: You should stay in 1.0.

annetill (approved):

security-analysis/security-analysis-api/src/test/resources/ActionFileTestV1.0.json
```diff
@@ -162,5 +162,10 @@
     "acEmulationEnabled" : false,
     "activePowerSetpoint" : 12.0,
     "relativeValue" : true
+  }, {
+    "type" : "SHUNT_COMPENSATOR_POSITION",
```
flo-dup, line 166: As discussed, it's a bit strange to have new type of actions in v1.0 (or in v1.1), but it seems that this kind of forward-compatibility is the way the versions are dealt with here.

flo-dup (approved):

PR #2501, opened 2023-03-13
Create a fictitious substation during CGMES export when a voltage level does not have a substation

miovd:

No

Bug fix

NPE when trying to export in CGMES a network containing a voltage level without substation.

A fictitious substation is created in the CGMES file containing all voltage levels without substation.

No

**Other information**:

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
@@ -223,6 +223,7 @@ public final class EquipmentExport {
     }
 
     private static void writeVoltageLevels(Network network, String cimNamespace, XMLStreamWriter writer, CgmesExportContext context, Set<Double> exportedBaseVoltagesByNominalV) throws XMLStreamException {
+        String fictSubstationId = null;
         for (VoltageLevel voltageLevel : network.getVoltageLevels()) {
             double nominalV = voltageLevel.getNominalV();
             BaseVoltageMapping.BaseVoltageSource baseVoltage = context.getBaseVoltageByNominalVoltage(nominalV);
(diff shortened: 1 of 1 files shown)


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
```diff
@@ -223,15 +223,25 @@ private static void writeSubGeographicalRegion(String subGeographicalRegionId, S
...
     private static void writeVoltageLevels(Network network, String cimNamespace, XMLStreamWriter writer, CgmesExportContext context, Set<Double> exportedBaseVoltagesByNominalV) throws XMLStreamException {
+        String fictSubstationId = null;
         for (VoltageLevel voltageLevel : network.getVoltageLevels()) {
             double nominalV = voltageLevel.getNominalV();
             BaseVoltageMapping.BaseVoltageSource baseVoltage = context.getBaseVoltageByNominalVoltage(nominalV);
             if (!exportedBaseVoltagesByNominalV.contains(nominalV) && baseVoltage.getSource().equals(Source.IGM)) {
                 BaseVoltageEq.write(baseVoltage.getId(), nominalV, cimNamespace, writer, context);
                 exportedBaseVoltagesByNominalV.add(nominalV);
             }
+            Optional<String> substationId = voltageLevel.getSubstation().map(s -> context.getNamingStrategy().getCgmesId(s));
```
zamarrenolm, line 234: Have you considered reusing the existing method `writeFictitiousSubstationFor(...)`. It could be called with the voltage level. It would create a different pair of sub/region for every voltage level, but we would not duplicate code ...

miovd: Good catch! I will do some refactoring in this sense

zamarrenolm (approved):

PR #2502, opened 2023-03-13
Add add method to the adders API

### LINKED ISSUE #2227: InjectionAdder::add
Feature

The add of `InjectionAdder` is not defined in the interface but in each implementations.

Having an `add` which would return an `Injection` (or a `U extends Injection`?).

**What is the motivation / use case for changing the behavior?**
For creating a bay, we would like to have an `InjectionAdder` as parameter of the `NetworkModification` to have a single java class for all the injections, but `add` cannot be called to create the corresponding injection.

Besides there's not a clear link to an `Injection` in the definition of the `InjectionAdder` (and to a lesser extent to an `Adder`)

**Other information**
Note that similarly there's no add in `IdentifiableAdder`.

flo-dup:

Only partly: fixes #2227 

Feature

Adder interfaces do not include the add method

Adder interfaces include the add method

Note that it is a breaking change only for users which have added their own equipments based on the existing interfaces.

### DIFF
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/BranchAdder.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/BranchAdder.java
@@ -10,23 +10,23 @@ package com.powsybl.iidm.network;
  *
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
  */
-public interface BranchAdder<T extends BranchAdder, B extends Branch<? super B>> extends IdentifiableAdder<T, B> {
+public interface BranchAdder<T extends Branch, A extends BranchAdder> extends IdentifiableAdder<T, A> {
 
-    T setVoltageLevel1(String voltageLevelId1);
+    A setVoltageLevel1(String voltageLevelId1);
 
-    T setNode1(int node1);
+    A setNode1(int node1);
 
-    T setBus1(String bus1);
+    A setBus1(String bus1);
 
-    T setConnectableBus1(String connectableBus1);
+    A setConnectableBus1(String connectableBus1);
 
-    T setVoltageLevel2(String voltageLevelId2);
+    A setVoltageLevel2(String voltageLevelId2);
 
-    T setNode2(int node2);
+    A setNode2(int node2);
 
-    T setBus2(String bus2);
+    A setBus2(String bus2);
 
-    T setConnectableBus2(String connectableBus2);
+    A setConnectableBus2(String connectableBus2);
 
-    B add();
+    T add();
 }
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/BranchAdder.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/BranchAdder.java
@@ -10,7 +10,7 @@ package com.powsybl.iidm.network;
  *
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
  */
-public interface BranchAdder<T extends Branch, A extends BranchAdder> extends IdentifiableAdder<T, A> {
+public interface BranchAdder<T extends Branch<? super T>, A extends BranchAdder> extends IdentifiableAdder<T, A> {
 
     A setVoltageLevel1(String voltageLevelId1);
 
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/BranchAdder.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/BranchAdder.java
@@ -10,7 +10,7 @@ package com.powsybl.iidm.network;
  *
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
  */
-public interface BranchAdder<T extends BranchAdder> extends IdentifiableAdder<T> {
+public interface BranchAdder<T extends BranchAdder, B extends Branch<? super B>> extends IdentifiableAdder<T, B> {
 
     T setVoltageLevel1(String voltageLevelId1);
 
@@ -28,4 +28,5 @@ public interface BranchAdder<T extends BranchAdder> extends IdentifiableAdder<T>
 
     T setConnectableBus2(String connectableBus2);
 
+    B add();
 }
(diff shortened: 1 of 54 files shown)


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/BranchAdder.java
```diff
@@ -10,22 +10,23 @@
  *
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
  */
-public interface BranchAdder<T extends BranchAdder> extends IdentifiableAdder<T> {
+public interface BranchAdder<T extends Branch<? super T>, A extends BranchAdder> extends IdentifiableAdder<T, A> {
```
miovd, line 13: Just for me, why did you write `T extends Branch<? super T>` instead of `T extends Branch<T>`?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/BranchAdder.java
```diff
@@ -10,22 +10,23 @@
...
-    T setNode2(int node2);
+    A setNode2(int node2);
 
-    T setBus2(String bus2);
+    A setBus2(String bus2);
 
-    T setConnectableBus2(String connectableBus2);
+    A setConnectableBus2(String connectableBus2);
 
+    T add();
```
miovd, line 31: I supposed that you added it here because of the generic class, but can you add `@Override`?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/HvdcConverterStationAdder.java
```diff
@@ -11,8 +11,9 @@
  *
  * @author Mathieu Bague <mathieu.bague at rte-france.com>
  */
-public interface HvdcConverterStationAdder<T extends HvdcConverterStationAdder> extends InjectionAdder<T> {
+public interface HvdcConverterStationAdder<T extends HvdcConverterStation<T>, A extends HvdcConverterStationAdder> extends InjectionAdder<T, A> {
 
-    T setLossFactor(float lossFactor);
+    A setLossFactor(float lossFactor);
 
+    T add();
```
miovd, line 18: Same remark

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/InjectionAdder.java
```diff
@@ -10,12 +10,13 @@
...
-    T setNode(int node);
+    A setNode(int node);
 
-    T setBus(String bus);
+    A setBus(String bus);
 
-    T setConnectableBus(String connectableBus);
+    A setConnectableBus(String connectableBus);
 
+    T add();
```
miovd, line 21: Same remark

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/AbstractIdentifiableAdderAdapter.java
```diff
@@ -7,13 +7,14 @@
...
 import com.powsybl.commons.PowsyblException;
+import com.powsybl.iidm.network.Identifiable;
 import com.powsybl.iidm.network.IdentifiableAdder;
 import com.powsybl.iidm.network.util.Identifiables;
 
 /**
  * @author Thomas Adam <tadam at silicom.fr>
  */
-abstract class AbstractIdentifiableAdderAdapter<I extends IdentifiableAdder<I>> extends AbstractAdapter<I> implements IdentifiableAdder<I> {
+abstract class AbstractIdentifiableAdderAdapter<T extends Identifiable<? super T>, I extends IdentifiableAdder<T, I>> extends AbstractAdapter<I> implements IdentifiableAdder<T, I> {
```
miovd, line 17: Same question as in `BranchAdder`

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractTransformerXml.java
```diff
@@ -19,7 +19,7 @@
 /**
  * @author Geoffroy Jamgotchian <geoffroy.jamgotchian at rte-france.com>
  */
-abstract class AbstractTransformerXml<T extends Connectable, A extends IdentifiableAdder<A>> extends AbstractConnectableXml<T, A, Container<? extends Identifiable<?>>> {
+abstract class AbstractTransformerXml<T extends Connectable<? super T>, A extends IdentifiableAdder<T, A>> extends AbstractConnectableXml<T, A, Container<? extends Identifiable<?>>> {
```
miovd, line 22: Same question as in `BranchAdder`

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/AbstractIdentifiableAdderAdapter.java
flo-dup: It's again because of `TieLine`, which makes BranchAdder have the same kind of parameter, and as BranchAdder extends AbstractIdentifiable we need the same here.

iidm/iidm-xml-converter/src/main/java/com/powsybl/iidm/xml/AbstractTransformerXml.java
flo-dup: This is no needed in fact, I'd better remove it. I left it in case we want to do something similar to TieLine, but do we really...?

miovd: I don't think we will

iidm/iidm-mergingview/src/main/java/com/powsybl/iidm/mergingview/AbstractIdentifiableAdderAdapter.java
flo-dup: After the merge I needed the same for AbstractSimpleIdentifiableXml

miovd (approved):

PR #2507, opened 2023-03-20
CGMES import: make substations property bags distinct when their region is defined twice (in BD and IGM)

miovd:

No

Bug fix

When the region of a substation is defined twice differently (different IDs in BD and IGM), two substations with the same ID are detected by the importer0

We keep only one substation with the same ID.

No

**Other information**:


zamarrenolm:
I prefer to keep the complexity of dealing with multiple references to regions and subregions in the SPARQL query.

Here is an updated version of the query for substations that works with the unit test:

```sparql
# query: substations
SELECT *
WHERE {
    ?Substation
        a cim:Substation ;
        cim:IdentifiedObject.name ?name .
    # Optional is required to keep the ?Substation variable already bounded in the sub-queries ...
    OPTIONAL {
        {
            SELECT ?SubRegion
            WHERE {
                ?Substation cim:Substation.Region ?SubRegion .
                ?SubRegion a cim:SubGeographicalRegion .
            }
            LIMIT 1
        }
        ?SubRegion cim:IdentifiedObject.name ?subRegionName .
        OPTIONAL {
            {
                SELECT ?Region
                WHERE {
                    ?SubRegion cim:SubGeographicalRegion.Region ?Region .
                    ?Region a cim:GeographicalRegion .
                }
                LIMIT 1
            }
            ?Region cim:IdentifiedObject.name ?regionName .
        }
    }
    # ... But we do not want results where the regions are missing
    FILTER (bound(?SubRegion) && bound(?Region))
}
```

zamarrenolm (approved):

PR #2510, opened 2023-03-21
Add loading from platform config and serialization for ScalingParameters

miovd:

No

To make entso-e processes work (balance computation and GLSK), we need serialization and loading

No

**Other information**:

### DIFF
--- a/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/scalable/ScalingParameters.java
+++ b/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/scalable/ScalingParameters.java
@@ -7,18 +7,29 @@
  */
 package com.powsybl.iidm.modification.scalable;
 
+import com.powsybl.commons.config.PlatformConfig;
+
+import java.util.Objects;
+
 /**
  * @author Coline Piloquet <coline.piloquet@rte-france.fr>
  */
 public class ScalingParameters {
 
-    private Scalable.ScalingConvention scalingConvention = Scalable.ScalingConvention.GENERATOR;
+    public static final String VERSION = "1.0";
 
-    private boolean reconnect;
+    public static final Scalable.ScalingConvention DEFAULT_SCALING_CONVENTION = Scalable.ScalingConvention.GENERATOR;
+    public static final boolean DEFAULT_CONSTANT_POWER_FACTOR = false;
+    public static final boolean DEFAULT_RECONNECT = false;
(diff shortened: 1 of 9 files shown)


iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/scalable/ScalingParameters.java
```diff
@@ -80,4 +91,20 @@ public ScalingParameters setIterative(boolean iterative) {
...
     }
+
+    public static ScalingParameters load() {
+        return load(PlatformConfig.defaultConfig());
+    }
+
+    public static ScalingParameters load(PlatformConfig platformConfig) {
+        Objects.requireNonNull(platformConfig);
+        ScalingParameters scalingParameters = new ScalingParameters();
+        platformConfig.getOptionalModuleConfig("scaling-default-parameters").ifPresent(config -> {
```
colinepiloquet, line 102: Is there a rule for property names? (between scaling-default-parameters and scalingConvention)

miovd: not really, I took inspiration from conventions in LoadFlowParameters... (camel for attributes and hyphens for the name)

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/scalable/ScalingParameters.java
```diff
@@ -80,4 +91,20 @@ public ScalingParameters setIterative(boolean iterative) {
...
+            scalingParameters.setScalingConvention(config.getEnumProperty("scalingConvention", Scalable.ScalingConvention.class, DEFAULT_SCALING_CONVENTION));
+            scalingParameters.setConstantPowerFactor(config.getBooleanProperty("constantPowerFactor", DEFAULT_CONSTANT_POWER_FACTOR));
```
flo-dup, line 104: To centralize the default parameter allocation you could also do something like:
```java
optModuleConfig.flatMap(c -> c.getOptionalBooleanProperty("xxx")).orElse(DEFAULT_VALUE);
```

miovd: The default boolean values might not be identical (it is coincidental that they are all `false` here), it might change in the future (or we might add other boolean attributes with a `true` default value) which is why I prefer to have one default value by attribute.

flo-dup (approved):

PR #2520, opened 2023-03-27
Fix identifiable stream

flo-dup:

No

Bug fix

`ClassCastException` when `network.getIdentifiableStream(IdentifiableType.BATTERY)`


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Network.java
```diff
@@ -1194,38 +1195,38 @@ default Network setMinimumAcceptableValidationLevel(ValidationLevel validationLe
         return this;
     }
 
-    default <I extends Identifiable<I>> Stream<I> getIdentifiableStream(IdentifiableType identifiableType) {
+    default Stream<Identifiable<?>> getIdentifiableStream(IdentifiableType identifiableType) {
```
miovd, line 1198: It doesn't work with previous signature?

PR #2537, opened 2023-04-11
CGMES import: import TATL shortName or name, not ID + CGMES export: fix on LoadingLimitType and LoadingLimit names

miovd:

No

Bug fix/enhancement on limits naming during CGMES import/export

No

**Other information**:


annetill (changes requested):
You must find a way to keep original ids (with alias) in order to export a CGMES file that comes from a CGMES file. It seems that there are lost.

miovd:
> You must find a way to keep original ids (with alias) in order to export a CGMES file that comes from a CGMES file. It seems that there are lost.

There were already lost but it does not seem to me that they are supposed to be persistent (same things for OperationalLimitSet)

zamarrenolm (approved):

annetill (approved):

PR #2546, opened 2023-04-13
Cgmes export fixes

miovd:

No

Bug fixes

Not consistent with quality of CGMES dataset

- when creating or referencing TN for disconnected switches, naming strategy is called
- Name is truncated to 32 characters in CGMES export (Quality check) 
- store all topological nodes in context to write all SvVoltage 
- Add netInterchange for control areas
- Add TieFlow.positiveFlowIn 
- neutralU in RTC and PTC is ratedU

No

**Other information**:

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportUtil.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportUtil.java
@@ -153,7 +153,7 @@ public final class CgmesExportUtil {
     public static void writeStartIdName(String className, String id, String name, String cimNamespace, XMLStreamWriter writer, CgmesExportContext context) throws XMLStreamException {
         writeStartId(className, id, true, cimNamespace, writer, context);
         writer.writeStartElement(cimNamespace, CgmesNames.NAME);
-        writer.writeCharacters(name);
+        writer.writeCharacters(name.length() > 32 ? name.substring(0, 32) : name); // name should not be longer than 32 characters
         writer.writeEndElement();
     }
 
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
@@ -7,6 +7,7 @@
 package com.powsybl.cgmes.conversion.export;
 
 import com.powsybl.cgmes.conversion.Conversion;
+import com.powsybl.cgmes.extensions.CgmesControlAreas;
 import com.powsybl.cgmes.extensions.CgmesTapChanger;
 import com.powsybl.cgmes.extensions.CgmesTapChangers;
 import com.powsybl.cgmes.model.CgmesNames;
@@ -219,9 +220,10 @@ public final class StateVariablesExport {
         });
         equivalentInjectionTerminalP.keySet().forEach(eiId -> writePowerFlow(eiId, equivalentInjectionTerminalP.get(eiId), equivalentInjectionTerminalQ.get(eiId), cimNamespace, writer, context));
 
+        CgmesControlAreas areas = network.getExtension(CgmesControlAreas.class);
         network.getBranchStream().forEach(b -> {
-            writePowerFlowTerminalFromAlias(b, Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.TERMINAL1, b.getTerminal1().getP(), b.getTerminal1().getQ(), cimNamespace, writer, context);
-            writePowerFlowTerminalFromAlias(b, Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.TERMINAL2, b.getTerminal2().getP(), b.getTerminal2().getQ(), cimNamespace, writer, context);
+            writeOptionalPowerFlowTerminalFromAlias(areas, b, Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.TERMINAL1, b.getTerminal1(), cimNamespace, writer, context);
+            writeOptionalPowerFlowTerminalFromAlias(areas, b, Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.TERMINAL2, b.getTerminal2(), cimNamespace, writer, context);
             if (b instanceof TieLine) {
                 TieLine tl = (TieLine) b;
                 if (context.exportBoundaryPowerFlows()) {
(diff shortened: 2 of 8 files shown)


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/elements/TieFlowEq.java
```diff
@@ -21,6 +21,9 @@ public static void write(String id, String controlAreaId, String terminalId, Str
         CgmesExportUtil.writeStartId("TieFlow", id, false, cimNamespace, writer, context);
         CgmesExportUtil.writeReference("TieFlow.ControlArea", controlAreaId, cimNamespace, writer, context);
         CgmesExportUtil.writeReference("TieFlow.Terminal", terminalId, cimNamespace, writer, context);
+        writer.writeStartElement(cimNamespace, "TieFlow.positiveFlowIn");
+        writer.writeCharacters("true"); // always true
```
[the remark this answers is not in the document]
zamarrenolm: About `positiveFlowIn`, it depends on the location of the terminal respect to the control area. For terminals of tie lines that are further away from the control area, this should be `true`. If the terminal is the near end of the tie line branch, it should be `false`. 

The typical configuration is that we have boundaries and the terminals of tie flows for control areas are defined at the boundary side, so for me it is ok to assume `true` by default. If we have to IGMs that share a boundary, and both define their tie flow at the boundary side of is own lines, both will set the `positiveFlowIn` to `true` in its tie flows. This can be seen in the conformity configuration MicroGrid BaseCase (https://github.com/powsybl/powsybl-core/tree/main/cgmes/cgmes-conformity/src/main/resources/conformity/cas-3-data-3.0.2/MicroGrid/BaseCase/MicroGrid-BaseCase-Merged)

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportUtil.java
```diff
@@ -153,7 +153,7 @@ public static void writeStartId(String className, String id, boolean writeMaster
     public static void writeStartIdName(String className, String id, String name, String cimNamespace, XMLStreamWriter writer, CgmesExportContext context) throws XMLStreamException {
         writeStartId(className, id, true, cimNamespace, writer, context);
         writer.writeStartElement(cimNamespace, CgmesNames.NAME);
-        writer.writeCharacters(name);
+        writer.writeCharacters(name.length() > 32 ? name.substring(0, 32) : name); // name should not be longer than 32 characters
```
zamarrenolm, line 156: I found empty names importing a real-world test case that would result in NPE, we should protect the check for length. Anyway, I have to check the import for that case

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
```diff
@@ -316,10 +299,6 @@ private static void writePowerFlowTerminalFromProperty(Identifiable<?> c, String
     }
 
     private static void writePowerFlow(String cgmesTerminalId, double p, double q, String cimNamespace, XMLStreamWriter writer, CgmesExportContext context) {
```
zamarrenolm, line 301: If we remove the condition for missing flows here we are writing as p=q=0 many power flows for terminals of lines and transformers that have not been initialized. I think we should avoid writing an SvPowerFlow object if both values are undefined. Writing power flow values for terminals of branches is not required.

miovd: SvPowerFlow are required in Quality requirements (at least, their absences are considered as warnings)

```xml
<qodc:RuleModel qodc:ruleID="SvPowerFlowInstances" qodc:severity="ERROR">
		<qodc:Level>5</qodc:Level>
		<qodc:RuleDetails>
		SvPowerFlow class is required to be instantiated for the following classes:
		subclasses of the RotatingMachine, subclasses of the EnergyConsumer, 
		EquivalentInjection, ShuntCompensator, StaticVarCompensator and EnergySource.
		</qodc:RuleDetails>
```

```xml
<qodc:RuleModel qodc:ruleID="SvPowerFlowBranchInstances" qodc:severity="WARNING">
		<qodc:Level>5</qodc:Level>
		<qodc:RuleDetails>
		SvPowerFlow class is required to be instantiated for branches if 
		referenced by a TieFlow.
		</qodc:RuleDetails>
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportUtil.java
miovd: I don't agree, name is here always from `nameOrId` in the code, it should never be empty (when name is empty, id is returned)

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
zamarrenolm: Yes, they are required in general for injections (both P and Q may be results of power flow calculations) and tie flows. but the quality requirements do not consider power flows to be present in SV for lines and transformers that are not tie flows. 

My problem here is that if we have NaNs in IIDM, it means we have not the set or calculated the value in IIDM. We will write the SV a value that may not be correct.

miovd: > quality requirements do not consider power flows to be present in SV for lines and transformers that are not tie flows

Are you sure? This is not what I gathered from the rule I quoted above...

> My problem here is that if we have NaNs in IIDM, it means we have not the set or calculated the value in IIDM. We will write the SV a value that may not be correct.

I agree but I'm not seeing another default value which would "more correct"

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportUtil.java
zamarrenolm: The NPE is raised when writing geographical region names. I will debug it.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
miovd: You're right! So we ignore lines and transfo that do not reference tie flows

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportUtil.java
miovd: cf e42901bf348ae8d35f5b606abc1a3308f6b5ff1c

zamarrenolm: Your fix in commit e42901b works ok. Sub-regions with duplicated names are preserved, although exported without name.

Maybe we could solve the handling of regions and sub-regions with duplicated names in a separate PR.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/TopologyExport.java
```diff
@@ -133,23 +133,21 @@ private static String findOrCreateTopologicalNode(VoltageLevel vl, int node) {
...
-        if (selectedNode.isEmpty()) {
-            throw new PowsyblException("nodeSet never can be empty");
+            return context.getNamingStrategy().getCgmesId(selectedBus.get());
         }
 
-        return vl.getId() + "_" + selectedNode.get();
+        return context.getNamingStrategy().getCgmesId(nodeSet.stream()
+                .sorted()
+                .findFirst()
+                .map(selectedNode -> "DISCONNECTED_" + vl.getId() + "_" + selectedNode)
```
zamarrenolm, line 142: At least some of the topological nodes created as "disconnected" miss the terminal definition in the EQ file. I'm checking it on real-world test cases.

miovd: Not sure why, the terminals are created after in `writeSwitchTerminal`, no?

zamarrenolm: Topological nodes and terminals should not be written for fictitious switches. `context.isExportedEquipment(sw)` is `false`, we do not have to export TN definitions and terminal references.

zamarrenolm: The same in SSH export.

zamarrenolm: If we remove the `DISCONNECTED_` prefix, the TNs will watch the ones used in SV export

zamarrenolm (approved):

PR #2551, opened 2023-04-18
[Network modifications] Add logs when exception is thrown

colinepiloquet:

No

Fix

Some exceptions are thrown without the associated log.

Adds logs when exception is thrown in network modifications.


miovd (commented):
Just for me, why do we need logs for exceptions in constructors? In this case, it has to fail and the message is in the exception... We put logs only when ignoring the issue was an option which is not the case here.

flo-dup:
> Just for me, why do we need logs for exceptions in constructors? In this case, it has to fail and the message is in the exception... We put logs only when ignoring the issue was an option which is not the case here.

Indeed. I think there's a misunderstanding, functional logging is missing in some cases when an exception is thrown. Here you added logs and functional logs with reporter.

miovd:
> > Just for me, why do we need logs for exceptions in constructors? In this case, it has to fail and the message is in the exception... We put logs only when ignoring the issue was an option which is not the case here.
> 
> Indeed. I think there's a misunderstanding, functional logging is missing in some cases when an exception is thrown. Here you added logs instead of functional logs with reporter.

Okay so
- Logs (slf4j) must be added in case of error that can be ignored (with `throwException`)
- Logs in **Reporter** must be added in any case

miovd (changes requested):
I don't agree with putting reporter in constructor rather than in API.
The issue was: in some places, when exceptions were thrown, logs were written but not everywhere. In practice, it means that if we want to catch exception and write logs, logs can be duplicated.
Two solutions are possible:
- only write reporters with `ERROR_SEVERITY ` when `throwException` is `false`: reporters are only filled when no exception is thrown
- no check in constructor: we move the checks in constructor inside `apply` method before processing the modification

I personally prefer the first option.

@flo-dup @geofjamg do you have any preference?

flo-dup:
Personally I prefer the second option... Indeed, if you catch exception and write logs based on the exception, you're losing the benefit of functional logs: you're just logging plain text. Moreover in all modifications' apply there's already some checks done: checking that the given ids correspond to an object in the network. So this would concentrate the checks in one place.

But maybe I'm missing the drawbacks of such an approach, could you elaborate why you prefer first option?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractLineConnectionModification.java
```diff
@@ -98,18 +98,26 @@ public String getLine2Name() {
         return line2Name;
     }
 
-    private static double checkPositionPercent(double positionPercent) {
+    private static boolean checkPositionPercent(double positionPercent, boolean throwException, Reporter reporter, Logger logger) {
```
flo-dup, line 101: Instead of passing the logger, you could use a logger defined in `AbstractLineConnectionModification`

flo-dup: Ok, that was the case already, and that's not the topic here.

PR #2557, opened 2023-04-24
Add new network modifications

nicolas-pierr:

To solve OpenReac issue [here](https://github.com/powsybl/powsybl-optimizer/issue/4), we need to extend the actual possibilities of IIDM extension with new Modifications classes.


iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/SectionModification.java
```diff
@@ -0,0 +1,58 @@
...
+import com.powsybl.iidm.network.ShuntCompensator;
+
+import java.util.Objects;
+
+/**
+ * Simple {@link NetworkModification} to change the section of a shunt compensator.
+ *
+ * @author Nicolas PIERRE <nicolas.pierre at artelys.com>
+ */
+public class SectionModification extends AbstractNetworkModification {
```
annetill, line 22: In the actions used in the security analysis API, we have chosen `ShuntCompensatorPositionAction`. I think finally we have to chose a name close to this first choice to remain coherent and uniform.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/SectionModification.java
```diff
@@ -0,0 +1,58 @@
...
+
+    private final String shuntCompensatorId;
+    private final int sectionCount;
+
+    public SectionModification(String shuntCompensatorId, int sectionCount) {
+        this.shuntCompensatorId = Objects.requireNonNull(shuntCompensatorId);
+        this.sectionCount = sectionCount;
+        if (sectionCount < 0) {
```
annetill, line 30: It is going to throw during the `apply`, so do we need it here too?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/SectionModification.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Override
+    public void apply(Network network, boolean throwException, ComputationManager computationManager,
+                      Reporter reporter) {
+        ShuntCompensator shuntCompensator = network.getShuntCompensator(shuntCompensatorId);
+        if (shuntCompensator == null) {
+            logOrThrow(throwException, "Shunt Compensator '" + shuntCompensatorId + "' not found");
+            return;
+        } else if (shuntCompensator.getMaximumSectionCount() < sectionCount) {
```
annetill, line 42: Same comment, this check already exists. I am not in favor of code duplication. @olperr1 what should we do?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/StaticVarCompensatorModification.java
```diff
@@ -0,0 +1,66 @@
...
+    private final OptionalDouble voltageSetPoint;
+    private final OptionalDouble reactivePowerSetPoint;
+
+    public StaticVarCompensatorModification(String svcId, OptionalDouble voltageSetPoint,
+                                            OptionalDouble reactivePowerSetPoint) {
+        this.svcId = svcId;
+        this.voltageSetPoint = Objects.requireNonNull(voltageSetPoint);
+        this.reactivePowerSetPoint = Objects.requireNonNull(reactivePowerSetPoint);
+        if (voltageSetPoint.isEmpty() && reactivePowerSetPoint.isEmpty()) {
+            LOGGER.warn("Creating a StaticVarCompensatorModification with no change !");
```
annetill, line 38: Change -> modification.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/StaticVarCompensatorModification.java
```diff
@@ -0,0 +1,66 @@
...
+        }
+        voltageSetPoint.ifPresent(svc::setVoltageSetpoint);
+        reactivePowerSetPoint.ifPresent(svc::setReactivePowerSetpoint);
+    }
+
+    public String getSvcId() {
+        return svcId;
+    }
+
+    public OptionalDouble getReactivePowerSetPoint() {
```
annetill, line 59: A long long discussion in the community: you must write "Setpoint" and not "SetPoint" :-)

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/TapPositionModification.java
```diff
@@ -0,0 +1,162 @@
...
+import java.util.Objects;
+import java.util.OptionalInt;
+
+/**
+ * Allow to modify a tap phase or ratio on two or three winding transformers.
+ * For three windings you must specify the Leg.
+ *
+ * @author Nicolas PIERRE <nicolas.pierre at artelys.com>
+ */
+public class TapPositionModification extends AbstractNetworkModification {
```
annetill, line 25: Indeed... I only need a modification on ratio tap changers. @olperr1 do we need to introduce genericity now of in a second time?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/VscConverterStationModification.java
```diff
@@ -0,0 +1,66 @@
...
+        }
+        voltageSetPoint.ifPresent(vsc::setVoltageSetpoint);
+        reactivePowerSetPoint.ifPresent(vsc::setReactivePowerSetpoint);
+    }
+
+    public String getVscId() {
+        return vscId;
+    }
+
+    public OptionalDouble getReactivePowerSetPoint() {
```
annetill, line 59: Setpoint

annetill (commented):
`TapPositionModification` should be think more deeply with @olperr1. We have just one modification but maybe it will be easier to have two. And indeed we don't need to modify tap position of PTC for the moment.

olperr1 (changes requested):

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/SectionModification.java
olperr1: This check exists in `setSectionCount` (same place as the check ` > 0`). It will be cleaner if it is removed in this `apply` method.

olperr1:
> `TapPositionModification` should be think more deeply with @olperr1. We have just one modification but maybe it will be easier to have two. And indeed we don't need to modify tap position of PTC for the moment.

@annetill is right, it will be more consistent with the existing code to have 2 classes to represent the modifications about Phase and Ratio taps. To achieve that, we could define an abstract class `AbstractTapPositionModification` and two child classes `PhaseTapPositionModification` and `RatioTapPositionModification`. These 2 children will each only contain their own implementation of the `apply` method.

olperr1 (changes requested):

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/StaticVarCompensatorModification.java
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @author Nicolas PIERRE <nicolas.pierre at artelys.com>
+ */
+public class StaticVarCompensatorModification extends AbstractNetworkModification {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(StaticVarCompensatorModification.class);
+
+    private final String svcId;
+    private final OptionalDouble voltageSetpoint;
+    private final OptionalDouble reactivePowerSetpoint;
```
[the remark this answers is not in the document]
nicolas-pierr: A simple principle to follow like: have Integer, Double _etc._ as attributes, handle null inside the class and use Optional in return type for public  functions. Is it a good understanding ?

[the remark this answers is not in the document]
olperr1: Yes, that's it. This is the intent of the Optional, and the way we use them in PowSyBl.

nicolas-pierr:
I split `TapPositionModifier` to have a Phase and a Ratio class. They extend a common abstract class.

Did some cleaning about Optionals and typos. @olperr1, can you tell me if there are some others things to change?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/tap/TapType.java
```diff
@@ -0,0 +1,14 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.iidm.modification.tap;
+
+/**
+ * @author Nicolas PIERRE <nicolas.pierre at artelys.com>
+ */
+public enum TapType {
```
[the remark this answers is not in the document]
nicolas-pierr: you are right, I went too fast on the split.

I don't think we will need to differenciate `PhaseTapPositionModification` and `RatioTapPositionModification` when aggregated as `AbstractTapPositionModification`, so I will move it to tests.

olperr1 (approved):

olperr1:
It's OK for me. Thank you, @nicolas-pierr.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/GeneratorModification.java
```diff
@@ -84,6 +84,14 @@ private void setTargetPWithinBoundaries(Generator g, double targetP, boolean ski
...
 
+    public Modifs getModifs() {
+        return modifs;
+    }
+
+    public String getGeneratorId() {
+        return generatorId;
+    }
+
     public static class Modifs {
```
annetill, line 95: How does it work if for a generator I want to apply a change only on `targetQ` and `targetV`?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/tap/RatioTapPositionModification.java
```diff
@@ -0,0 +1,73 @@
...
+        return new RatioTapPositionModification(transfoId,
+            TransformerType.THREE_WINDINGS_TRANSFORMER, tapPosition, legNumber);
+    }
+
+    /**
+     * @param transfoId   the ID of the three windings transformer, which holds the rtc
+     * @param tapPosition the new tap position
+     * @param leg         defines on which leg of the three winding transformer the modification will be done. null on two windings.
+     */
+    public RatioTapPositionModification(String transfoId, TransformerType element,
```
annetill, line 43: We cannot choose this design. When a user gives the id of a transformer, you cannot ask him to give the type of transformer : two or three windings. As you need the network to know, most of the checks you are doing must be in the apply. And for me you must have a method where we don't provide the leg number, without saying that it is a two windings transfromer.

annetill: And finally the leg number must be optional. Even for a three windings transformer, you need it when you found more than one RTC in the transformer.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/GeneratorModification.java
nicolas-pierr: The other values will not be set for our application in OpenReac. Thus only targetV and targetQ won't be null.

olperr1 (changes requested):

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/tapchanger/PhaseTapPositionModification.java
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * @author Nicolas PIERRE <nicolas.pierre at artelys.com>
+ */
+public class PhaseTapPositionModification extends AbstractTapPositionModification {
+
+    /**
+     * Creates a PTC tap modification for two windings transformers, or three windings transformer with a single PTC.
+     *
+     * @param transformerId the ID of the three windings transformer, which holds the PTC
```
olperr1, line 21:
```suggestion
     * @param transformerId the ID of the transformer, which holds the PTC
```

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/tapchanger/RatioTapPositionModification.java
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * @author Nicolas PIERRE <nicolas.pierre at artelys.com>
+ */
+public class RatioTapPositionModification extends AbstractTapPositionModification {
+
+    /**
+     * Creates a RTC tap modification for two windings transformers, or three windings transformer with a single RTC.
+     *
+     * @param transformerId the ID of the three windings transformer, which holds the rtc
```
olperr1, line 21:
```suggestion
     * @param transformerId the ID of the transformer, which holds the rtc
```

olperr1 (changes requested):

olperr1 (approved):

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/tapchanger/AbstractTapPositionModification.java
```diff
@@ -0,0 +1,115 @@
...
+     *                    Ignored on two windings transformers.
+     */
+    protected AbstractTapPositionModification(String transformerId, int tapPosition,
+                                              ThreeWindingsTransformer.Side legSide) {
+        this.transformerId = Objects.requireNonNull(transformerId);
+        this.tapPosition = tapPosition;
+        this.legSide = legSide;
+    }
+
+    abstract void applyTwoWindings(Network network, TwoWindingsTransformer twoWindingsTransformer,
```
annetill, line 49: applyTwoWindings to applyTwoWindingsTranformer and applyThreeWindings to applyThreeWindingsTransformer could be better.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ThreeWindingsTransformer.java
```diff
@@ -319,6 +319,19 @@ default Substation getNullableSubstation() {
...
+    default Leg getLeg(Side side) {
+        switch (side) {
+            case ONE:
+                return getLeg1();
+            case TWO:
+                return getLeg2();
+            case THREE:
+                return getLeg3();
+            default:
+                return null;
```
annetill, line 331: After a second thought, can you better throw here instead of returing null.

annetill (changes requested):
Finally small change.

annetill (approved):

PR #2563, opened 2023-05-10
[WIP] Activate control on Qmin and Qmax on ReactiveCapabilityCurve

etiennehomer:
PR open for discussion: do we need to keep those rows commented as it will raise an exception for many network files that contain incohent values Qmin and Qmax ? 
Was it an omission to remove the TODO after IIDM 1.1 as it is said in the comment ?

Remove an old TODO and activate the control on Qmin < Qmax

No control

An exception is raised if Qmin > Qmax


geofjamg (changes requested):
We need to keep backward compatibility
 - when loading an old xiidm < 1.11 with such an qmin > Qmax, it should be fixed automatically and not throw an error but warn the user.
 - we need to ensure that all importers are robust to this

alicecaron:
This pull request is duplicated by #3314

PR #2564, opened 2023-05-10
Add modification to replace tie lines by lines

flo-dup:

No

Feature

A tie line can be replaced by a line thanks to a network modification


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/TieLineUtil.java
```diff
@@ -77,28 +84,37 @@ public static void mergeProperties(DanglingLine dl1, DanglingLine dl2, Propertie
                 properties.setProperty(prop, dl1.getProperty(prop));
             } else if (dl1.getProperty(prop).isEmpty()) {
                 LOGGER.debug("Inconsistencies of property '{}' between both sides of merged line. Side 1 is empty, keeping side 2 value '{}'", prop, dl2.getProperty(prop));
+                propertyOnlyOnOneSide(reporter, prop, dl2.getProperty(prop), 1, dl1.getId(), dl2.getId());
                 properties.setProperty(prop, dl2.getProperty(prop));
             } else if (dl2.getProperty(prop).isEmpty()) {
                 LOGGER.debug("Inconsistencies of property '{}' between both sides of merged line. Side 2 is empty, keeping side 1 value '{}'", prop, dl1.getProperty(prop));
+                propertyOnlyOnOneSide(reporter, prop, dl1.getProperty(prop), 2, dl1.getId(), dl2.getId());
```
flo-dup, line 91: Shouldn't it be:
```suggestion
                propertyOnlyOnOneSide(reporter, prop, dl1.getProperty(prop), 2, dl2.getId(), dl1.getId());
```

miovd: No it should be always side 1 first then side 2

miovd:
@flo-dup you deleted MergedXnode and it doesn't work anymore :( Should we revert it?

miovd:
> @flo-dup you deleted MergedXnode and it doesn't work anymore :( Should we revert it?

It had been decided to keep it deleted and only add aliases

annetill:
We should remain coherent with the previous network modifications. In `ReplaceTeePointByVoltageLevelOnLine` and in `ConnectVoltageLevelOnLine`, we remove lines without making any treatment on extension, aliases or properties. It was a question of course at that moment, for example for short-circuit extensions, positions extensions, etc. As we decided not to do it at a first approach, I think that you have to get ride of everything at a first approach. I just have a doubt with dangling lines id that could be added as aliases.

miovd:
> We should remain coherent with the previous network modifications. In `ReplaceTeePointByVoltageLevelOnLine` and in `ConnectVoltageLevelOnLine`, we remove lines without making any treatment on extension, aliases or properties. It was a question of course at that moment, for example for short-circuit extensions, positions extensions, etc. As we decided not to do it at a first approach, I think that you have to get ride of everything at a first approach. I just have a doubt with dangling lines id that could be added as aliases and maybe the Xnode name.

Contrary to `ReplaceTeePointByVoltageLevelOnLine` or `ConnectVoltageLevelOnLine`, this modification is foremost dedicated to keep backwards compatibility, which is why I made it reproduce former behavior with aliases and properties merging. Extensions are not kept, only a warning is ensued.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/ReplaceTieLinesByLinesPostProcessor.java
```diff
@@ -0,0 +1,50 @@
...
+import com.powsybl.iidm.network.ImportPostProcessor;
+import com.powsybl.iidm.network.Network;
+
+/**
+ * @author Miora Vedelago <miora.ralambotiana at rte-france.com>
+ */
+@AutoService(ImportPostProcessor.class)
+public class ReplaceTieLinesByLinesPostProcessor implements ImportPostProcessor {
+
+    public static final String THROW_EXCEPTION = "iidm.import.post-processor.replace-tie-lines-by-lines.throw-exception";
```
annetill, line 26: @annetill @flo-dup @olperr1  to be documented in the migration guide.

annetill (approved):

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/ReplaceTieLinesByLines.java
```diff
@@ -0,0 +1,131 @@
...
+            double p1 = dl1.getTerminal().getP();
+            double q1 = dl1.getTerminal().getQ();
+            double p2 = dl2.getTerminal().getP();
+            double q2 = dl2.getTerminal().getQ();
+            Properties properties = new Properties();
+            tl.getPropertyNames().forEach(pn -> properties.put(pn, tl.getProperty(pn)));
+            mergeProperties(dl1, dl2, properties, reporter);
+            Map<String, String> aliases = new HashMap<>();
+            tl.getAliases().forEach(alias -> aliases.put(alias, tl.getAliasType(alias).orElse("")));
+            mergeIdenticalAliases(dl1, dl2, aliases, reporter);
```
flo-dup, line 84: There cannot be identical aliases, dl1 and dl2 are in the same network. An exception is thrown when an existing alias is added in the network index.

PR #2574, opened 2023-05-17
Fixes in CGMES export

colinepiloquet:

No

Fixes related to SUV

- **Remove regulation status from EQ**
-  **Double dependency to SSH is removed in SV**
- **Dependency to TPBD is added to SV**
- **ReactiveCapabilityCurve CurveStyle changed from ConstantYValue to StraightLineYValues**
- **default value for exportPowerFlowsForSwitches is true**
- **Add LoadArea and subload area in load groups**
- **Add custom modeling authority set**
- **Add energyArea in controlArea**
- **Format maximum double values using scientific notation**.
- **TopologicalIslandId needs to be maximum 32 characters**
- **Terminals must have a name**
- **RatedS for PowerTransformers and RotatingMachine must be written and cannot be equal to 0, so it is set to 100 for PowerTransformerEnd but as optional for SynchronousMachine we do nothing (it should be done at iidm level).**
- **Fix terminals and busbar sections have the same ID**
- **switch retained and connected to same bus on both side are ignored** (I am not sure anymore why we have to do that)
- GeneratingUnit.maxP cannot be equal to 0.
- add interpolation if ReactiveCapabilityCurve has only 2 points -> TODO at IIDM level
- export all SvPowerFlow for switches
- **write TieFlow for Boundary**
- **Return non conform load by default if the extension LoadDetail is not null**
- minOperatingP and maxOperatingP takes 0 in the interval (for the interpolation)
- **fix kind synchronous machine so that it fits the mode in SSH**

**Other information**:


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportUtil.java
```diff
@@ -150,13 +156,17 @@ public static void writeStartId(String className, String id, boolean writeMaster
         }
     }
 
-    public static void writeStartIdName(String className, String id, String name, String cimNamespace, XMLStreamWriter writer, CgmesExportContext context) throws XMLStreamException {
-        writeStartId(className, id, true, cimNamespace, writer, context);
+    public static void writeStartIdName(String className, String id, String name, boolean writeMasterResourceId, String cimNamespace, XMLStreamWriter writer, CgmesExportContext context) throws XMLStreamException {
```
colinepiloquet, line 159: I am not sure this is useful anymore.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/ExportXmlCompare.java
```diff
@@ -578,6 +578,44 @@ static ComparisonResult ignoringOperationalLimitIds(Comparison comparison, Compa
         return result;
     }
 
+    static ComparisonResult ignoringLoadAreaIds(Comparison comparison, ComparisonResult result) {
```
colinepiloquet, line 581: @miovd I am absolutely not sure that this is how it should be done. Maybe for ID but I feel like there should be another way for resource?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -203,7 +210,7 @@ public String getFormat() {
             EXPORT_POWER_FLOWS_FOR_SWITCHES,
             ParameterType.BOOLEAN,
             "Export power flows for switches",
-            Boolean.FALSE);
+            Boolean.TRUE);
```
annetill, line 213: @miovd are you sure about this change?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -227,6 +234,18 @@ public String getFormat() {
             "Boundary TP model identifier",
             null);
 
+    private static final Parameter MODELING_AUTHORITY_SET_PARAMETER = new Parameter(
+            MODELING_AUTHORITY_SET,
+            ParameterType.STRING,
+            "Modeling authority set",
+            "powsybl.org");
+
+    private static final Parameter MAXIMUM_DOUBLE_VALUE_PARAMETER = new Parameter(
```
annetill, line 243: Why do we need that?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
```diff
@@ -339,6 +340,13 @@ public boolean isExportedEquipment(Identifiable<?> c) {
         boolean ignored = c.isFictitious() &&
                 (c instanceof Load
                         || c instanceof Switch && "true".equals(c.getProperty(Conversion.PROPERTY_IS_CREATED_FOR_DISCONNECTED_TERMINAL)));
+        if (c instanceof Switch) {
+            Switch ss = (Switch) c;
+            VoltageLevel.BusBreakerView view = ss.getVoltageLevel().getBusBreakerView();
+            if (ss.isRetained() && view.getBus1(ss.getId()).equals(view.getBus2(ss.getId()))) {
```
annetill, line 346: @zamarrenolm do you know why we have to do this?

miovd:
> RatedS for PowerTransformers and RotatingMachine must be written and cannot be equal to 0,so it is set to Double.MAX_VALUE
> GeneratingUnit.maxP cannot be equal to 0, so it is set to Double.MAX_VALUE

For these, default values must be debated @annetill @zamarrenolm 

> write positive minP and maxP for SynchronousMachine

Not sure this should be kept... @colinepiloquet can you make test just without this fix?

zamarrenolm:
> > RatedS for PowerTransformers and RotatingMachine must be written and cannot be equal to 0,so it is set to Double.MAX_VALUE

For power transformers, `ratedS` is not read during CGMES import. It could be read. 
For rotating machines, it is an optional attribute both in CGMES 2 and 3. Maybe the validation rules have additional constraints over the standard?. 
Maybe we can think of a default value that is not so extreme.

> > GeneratingUnit.maxP cannot be equal to 0, so it is set to Double.MAX_VALUE
> 
> For these, default values must be debated @annetill @zamarrenolm
> 
> > write positive minP and maxP for SynchronousMachine
> 
> Not sure this should be kept... @colinepiloquet can you make test just without this fix?

zamarrenolm:
> > > RatedS for PowerTransformers and RotatingMachine must be written and cannot be equal to 0,so it is set to Double.MAX_VALUE
> 
> For power transformers, `ratedS` is not read during CGMES import. It could be read. For rotating machines, it is an optional attribute both in CGMES 2 and 3. Maybe the validation rules have additional constraints over the standard?. Maybe we can think of a default value that is not so extreme.
> 
> > > GeneratingUnit.maxP cannot be equal to 0, so it is set to Double.MAX_VALUE
> > 
> > 
> > For these, default values must be debated @annetill @zamarrenolm
> > > write positive minP and maxP for SynchronousMachine
> > 
> > 
> > Not sure this should be kept... @colinepiloquet can you make test just without this fix?

To obtain a reasonable default value for the `ratedS` of rotating machines, we could obtain the maximum absolute value of `P` and `Q` for the generator (eventually using points from reactive capability curve) and multiply it by a reasonability factor (10), like it is described in the rule `SMRatedSunrealistic` of the ENTSO-E document about quality of CGMES dataset calculations (v3.3).

marqueslanauja (approved):

PR #2576, opened 2023-05-17
[Network modifications] Add check on order position value in CreateConnectableBay

colinepiloquet:

Fixes https://github.com/powsybl/powsybl-core/pull/2532.

Fix

Order position value is not checked, meaning that a value that is too low or too big can be given and as long as it is not equal to an order position that is not taken, the extension will be created. After creation of the extension, the methods getUnusedOrderPositionsAfter and getUnusedOrderPositionsBefore return an empty range.

The value of the order position given as an attribute is checked and if it is wrong, the connectable is created but not the extension and a warning is returned (same behaviour as if the order position is already taken). This check fails if the order position is negative, if it is higher than the lowest order position on the next slice or if it is slower than the highest order position in the previous slice.


iidm/iidm-modification/pom.xml
```diff
@@ -100,6 +100,11 @@
             <artifactId>powsybl-iidm-xml-converter</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
annetill, line 103: to be removed.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeederBays.java
```diff
@@ -91,6 +98,33 @@ private boolean checkOrders(int side, VoltageLevel voltageLevel, Reporter report
         return true;
     }
 
+    private boolean checkOrderValue(int side, BusbarSection busbarSection) {
+        VoltageLevel voltageLevel = busbarSection.getTerminal().getVoltageLevel();
+        Integer positionOrder = getPositionOrder(side);
+        BusbarSectionPosition positionExtension = busbarSection.getExtension(BusbarSectionPosition.class);
+        if (positionExtension != null) {
+            Optional<Range<Integer>> unusedOrderPositionsAfter = getUnusedOrderPositionsAfter(busbarSection);
+            Optional<Range<Integer>> unusedOrderPositionsBefore = getUnusedOrderPositionsBefore(busbarSection);
```
flo-dup, line 107: Here you get a range whereas you only need one of the two values. And calling the two methods makes you calculating the map of position twice. We could avoid this by calculating directly the desired values.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/TopologyModificationUtils.java
```diff
@@ -350,6 +350,38 @@ static void createTopologyFromBusbarSectionList(VoltageLevel voltageLevel, int f
         });
     }
 
+    /**
+     * Get the maximum order position between the maximum for the given busbar section and the maximum for all the
+     * busbar sections that have the same index.
+     */
+    public static Optional<Integer> getMaxOrderPositionSlice(BusbarSection bbs) {
+        BusbarSectionPosition busbarSectionPosition = bbs.getExtension(BusbarSectionPosition.class);
```
flo-dup, line 358: Duplicate code, we can factorize this

flo-dup (changes requested):

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/CreateFeederBay.java
```diff
@@ -42,7 +42,7 @@ public class CreateFeederBay extends AbstractCreateConnectableFeederBays {
         super(0);
         this.injectionAdder = Objects.requireNonNull(injectionAdder);
         this.busOrBbsId = Objects.requireNonNull(busOrBbsId);
-        this.injectionPositionOrder = injectionPositionOrder;
+        this.injectionPositionOrder = checkPositionOrderPositive(injectionPositionOrder);
```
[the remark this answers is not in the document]
flo-dup: and if injectionPositionOrder is null it will throw a NPE when casting to int

flo-dup (approved):

PR #2581, opened 2023-05-22
add new method checkLimitsDc with dcPowerFactor param

dbraquart:

This PR adds a new utility method checkLimitDc() to check the current limit violations on a network where no intensity is computed (dc power flow mode). When I is not set, we compute it using the active power and a cosinus factor we have to pass (cf P = sqrt(3) x U x I x cosphi).

**Other information**:


security-analysis/security-analysis-api/src/test/java/com/powsybl/security/SecurityTest.java
```diff
@@ -125,14 +124,27 @@ void printLimitsViolations() {
     @Test
     void checkLimits() {
         List<LimitViolation> violations = Security.checkLimits(network);
-
-        assertEquals(5, violations.size());
```
EtienneLt, line 129: it does not pass ?

dbraquart: yes it does, but it is also checked by assertViolations()

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/detectors/AbstractContingencyBlindDetector.java
```diff
@@ -50,9 +63,25 @@ public void checkCurrent(Branch branch, Consumer<LimitViolation> consumer) {
...
     @Override
     public void checkAll(Network network, Consumer<LimitViolation> consumer) {
         network.getBranchStream().forEach(b -> checkCurrent(b, consumer));
+        checkVoltageLevels(network, consumer);
+    }
+
+    @Override
+    public void checkAllDc(Network network, double dcPowerFactor, Consumer<LimitViolation> consumer) {
+        network.getBranchStream().forEach(b -> checkCurrentDc(b, dcPowerFactor, consumer));
+        checkVoltageLevels(network, consumer);
```
geofjamg, line 81: To remove because there is no voltage magnitude calculation in DC LF calculation

geofjamg (changes requested):

geofjamg (approved):

EtienneLt (approved):

annetill (approved):

PR #2589, opened 2023-05-29
SV: support of DC approximation

annetill:

Feature

Dc approximation is not supported in the SV class.

Dc approximation is supported in the SV class. When there is not enough information to calculate all the values on the other side, the Dc approximation is calculated if the `p` and `a` are defined.

**Other information**:

### DIFF
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
@@ -74,9 +74,16 @@ public class SV {
     }
 
     public SV otherSide(double r, double x, double g1, double b1, double g2, double b2, double rho, double alpha) {
-        LinkData.BranchAdmittanceMatrix adm = LinkData.calculateBranchAdmittance(r, x, 1 / rho, -alpha, 1.0, 0.0,
-            new Complex(g1, b1), new Complex(g2, b2));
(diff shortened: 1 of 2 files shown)


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SV.java
```diff
@@ -348,6 +268,9 @@ private SV otherSide(LinkData.BranchAdmittanceMatrix adm) {
             s = flowAtEnd1(adm, v, v2);
             otherSide = Branch.Side.ONE;
         }
+        if (Double.isNaN(q) || Double.isNaN(u)) { // DC approximation
```
annetill, line 271: @zamarrenolm and @marqueslanauja this approximation is totally coherent with what we are doing in open-loadflow for DC approximation.

annetill: Only a test is missing.

marqueslanauja: I think we must change the alpha (`a`) on the other side. I implemented and committed the necessary code to calculate the alpha according with the DC loadflow inside the open-Loadflow. I also added unitary tests.

zamarrenolm (approved):

PR #2618, opened 2023-06-15
CGMES export: use region and base voltage references from given BD file

zamarrenolm:

Feature.

When exporting an IIDM network, region and base voltages may be created in the EQ output with unique IDs if no aliases/properties are present in substation/voltage levels.

If no specific aliases/properties are given for region/base voltage ids in the IIDM network, they could be obtained from a Boundary file given as input to the export process.

**Other information**:

To solve GeographicalRegionBD issue reported by validation tool that considers quality of [Quality of CGMES datasets and calculations v3.3](https://eepublicdownloads.azureedge.net/clean-documents/digital/QualityOfCGMESdatasetsAndCalculations_v3_3.pdf)


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
```diff
@@ -305,12 +314,44 @@ private void addIidmMappingsSubstations(Network network) {
         }
     }
 
+    private Pair<String, String> getCreateRegion(Substation substation) {
```
annetill, line 317: The name of the method is a bit weird.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -239,18 +269,21 @@ public String getFormat() {
...
             "Modeling authority set",
-            "powsybl.org");
-
+            DEFAULT_MODELING_AUTHORITY_SET_VALUE);
     private static final Parameter MODEL_DESCRIPTION_PARAMETER = new Parameter(
             MODEL_DESCRIPTION,
             ParameterType.STRING,
             "Model description",
             null);
+    private static final Parameter SOURCING_ACTOR_PARAMETER = new Parameter(
```
annetill, line 282: This new parameter is really confusing for me and I don't see why we need it. 
- If we don't come from a CGMES file and we want to export a CGMES file from IIDM, I understand. 
- If we come from a CGMES file (what we do for European Merging Function), I think most of the time we have enough information in the input file to know what is the sourcing actor. We can store the information in a property or more.
- Sourcing actor or modeling authority set : what is the difference? And do we really 2 fields?

annetill: I wonder about having an extension with 
````
  <cim:GeographicalRegion rdf:ID="_7e6dab44-5e92-4698-92d1-dcc3421cce1a">
    <cim:IdentifiedObject.description>Geographical Region of ENTSO-E</cim:IdentifiedObject.description>
    <cim:IdentifiedObject.name>EU</cim:IdentifiedObject.name>
    <entsoe:IdentifiedObject.shortName>EU</entsoe:IdentifiedObject.shortName>
    <entsoe:IdentifiedObject.energyIdentCodeEic>10Y1001C--00037Z</entsoe:IdentifiedObject.energyIdentCodeEic>
    <cgmbp:MARegion.ModelingAuthority rdf:resource="#_6f010c0a-a735-49ec-99b1-76fed42ea948"/>
    <cgmbp:NetworkRegion.sourcingActor>ENTSOE</cgmbp:NetworkRegion.sourcingActor>
    <cgmbp:NetworkRegion.masUri>http://www.entsoe.eu/OperationalPlanning</cgmbp:NetworkRegion.masUri>
  </cim:GeographicalRegion>
````

Comment (zamarrenolm) [reply cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java:282]: > This new parameter is really confusing for me and I don't see why we need it.
> 
> * If we don't come from a CGMES file and we want to export a CGMES file from IIDM, I understand.
> * If we come from a CGMES file (what we do for European Merging Function), I think most of the time we have enough information in the input file to know what is the sourcing actor. It seems that we have in the IGM the id of the GeographicalRegion, that is fully described in boundary profiles. We must use that reference in priority.

You are right, when we come from CGMES files we do not need (and we do not expect) this parameter. The information needed for the CGMES export is already present in the IIDM network as either extensions, aliases or properties. 

The parameter is only relevant when we are not coming from CGMES. In this situation we have to inform the export module "who" (which actor) is publishing the data.

As discussed, there is an additional step we could apply to simplify the mechanism: if the IIDM network contains only one country, we could determine the rest of required data based in its code: for "FR", we can determine that we have to use the sourcing actor "RTEFRANCE" looking in the reference data. This way, for networks with only country and one TSO, there will be no need to specify the "sourcing actor" as a parameter, it would be automatically detected from the IIDM country.

Comment (zamarrenolm) [reply cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java:282]: > I wonder about having an extension with
> 
> ```
>   <cim:GeographicalRegion rdf:ID="_7e6dab44-5e92-4698-92d1-dcc3421cce1a">
>     <cim:IdentifiedObject.description>Geographical Region of ENTSO-E</cim:IdentifiedObject.description>
>     <cim:IdentifiedObject.name>EU</cim:IdentifiedObject.name>
>     <entsoe:IdentifiedObject.shortName>EU</entsoe:IdentifiedObject.shortName>
>     <entsoe:IdentifiedObject.energyIdentCodeEic>10Y1001C--00037Z</entsoe:IdentifiedObject.energyIdentCodeEic>
>     <cgmbp:MARegion.ModelingAuthority rdf:resource="#_6f010c0a-a735-49ec-99b1-76fed42ea948"/>
>     <cgmbp:NetworkRegion.sourcingActor>ENTSOE</cgmbp:NetworkRegion.sourcingActor>
>     <cgmbp:NetworkRegion.masUri>http://www.entsoe.eu/OperationalPlanning</cgmbp:NetworkRegion.masUri>
>   </cim:GeographicalRegion>
> ```
> 
> when exporting from scratch

Having an extension for the information required to link with reference data is a good idea, but the details about the "reference data" itself are defined and published in the boundary file. I think that in the extension we should store only information on how to link the IIDM network entities with that reference data.

Additionally, we probably should design the extension in a way that allows to express that the IIDM network contains multiple countries/TSOs/modelling authorities. Either that, or we assume one Network corresponds exactly to one TSO and when merged (using Subnetworks) all extensions (one for each network) will be considered ...

File: cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java [hunk 3]
```diff
@@ -67,20 +68,49 @@ public void export(Network network, Properties params, DataSource ds, Reporter r
         String filenameTp = baseName + "_TP.xml";
         String filenameSsh = baseName + "_SSH.xml";
         String filenameSv = baseName + "_SV.xml";
+
+        // Reference data (if required) will come from imported boundaries
+        // We may have received a sourcing actor as a parameter
+        String sourcingActorName = Parameter.readString(getFormat(), params, SOURCING_ACTOR_PARAMETER, defaultValueConfig);
+        String countryName = null;
+        if (sourcingActorName == null || sourcingActorName.isEmpty()) {
```
annetill, line 76: I disagree with that logic. For me, it is better to rely on the country present in the file. If you find one country and with the boundary files you are able to find the information about geographical region, that is nice. If you are not able, because the country leads to several sourcing actors for example and you are not able to choose, so you use the parameter. But for me, using the parameter is only a back up. Maybe I was not clear on that sorry.

annetill:
@colinepiloquet has tested this feature on our network:
- If several country are present, it does not work. But it is the normal behavior for the moment. @nemanja-st tells me yesterday that in that case, we have two behaviors 1) Relying on parameter (EU, ENTSO-E) or 2) Exporting all the georgraphical regions (SWE).
- If we put a bad sourcing actor, evein if we have one country, it does not work. I think it should work.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
colinepiloquet, line 76: What about if SourcingActor given as parameter is wrong? Do we not want to log a warning but still check if there's only one country ?

annetill: @flo-dup confirms me that it is an override, even if the parameter is wrong.

annetill (approved):

PR #2623, opened 2023-06-19
Invalid components after dangling lines merging and tie lines removal

geofjamg:

No

Bug fix

When merging 2 networks or when removing tie lines, connected and synchronous components are not correctly invalidated.

When merging 2 networks or when removing tie lines, connected and synchronous components are now correctly invalidated.

**Other information**:


miovd (commented):
In some **weird fringe cases that should not happen** (e.g. tie line inside a voltage level), it might cause a topology problem if the view is not invalidated. It might be safer to also invalidate topological cache (as it is done when a connectable is removed)

miovd:
> In some **weird fringe cases that should not happen** (e.g. tie line inside a voltage level), it might cause a topology problem if the view is not invalidated. It might be safer to also invalidate topological cache (as it is done when a connectable is removed)

After discussion, it seems that topological calculation does not consider buses linked by lines as the same so this is good for me!

miovd (approved):

PR #2627, opened 2023-06-27
[CGMES] Export loads with negative p0 as EnergySource

colinepiloquet:

No

Bug fix

All loads are exported as ConformLoads or NonConformLoads

In QoCDC, it is specified that ConformLoads and NonConformLoads cannot have a negative p. These loads are exported as EnergySource.


zamarrenolm:
To align better with the description of the affected rule I propose to use EnergySource class instead of an EquivalentInjection. 

The justification of the rule ValidLoad in the [QoCDC document v3.3.1](https://eepublicdownloads.azureedge.net/clean-documents/digital/QualityOfCGMESdatasetsAndCalculations_v3.3.1.pdf) is:

_Due to the load sign convention, all loads should be positive or zero. Decentralized generation should be modelled explicitly.  See IEC TS 61970-600-2:2017 section 7.8.5._ 

Section 7.8.5 of the 61970-600-2 standard talks about EnergySource, defined as: _A generic equivalent for an energy supplier on a transmission or distribution voltage level_.

zamarrenolm:
EnergySource is used in QoCDC to talk about distributed energy sources (DER). The rule ValidDER matches with the interpretation of p < 0 for IIDM loads: _For every instance of a DistributedEnergyResource (DER), e.g. cim:EnergySource, the value of cim:EnergySource.activePower should be lower than or equal to zero_.

zamarrenolm:
Furthermore, we would not have to specify values for min/max P, Q, that are required for equivalent injections because of the potential regulation capabilities.

zamarrenolm (approved):
Agree with the changes. Maybe consider adding a specific unit test.

annetill (approved):

Issue #2498, opened 2023-03-13
Operator strategies multiple conditions


geofjamg:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
With operator strategies, all the actions of a single operator strategy are triggered from a unique condition.
```java
Condition condition = new AtLeastOneViolationCondition(List.of("B1"));
List<String> actions = List.of("A1", "A2");
OperatorStrategy strategy = new OperatorStrategy("OS", ContingencyContext.all(), condition, actions);
```

So here there is no way to have a different condition for A1 and A2 actions. And if we split the operator strategy, we won't have during the security analysis a violation detection potentially based of A1 and A2.

* **What is the expected behavior?**
We should be able to group actions in an operator strategy and each group should have its own condition.
It could look something like:
```java
Condition condition1 = new AtLeastOneViolationCondition(List.of("B1"));
Condition condition2 = new AtLeastOneViolationCondition(List.of("B2"));
List<SubStrategy> subStrategies = List.of(new SubStrategy("A1", condition1), new SubStrategy("A2", condition2));
OperatorStrategy strategy = new OperatorStrategy("OS", ContingencyContext.all(), subStrategies);
```

* **What is the motivation / use case for changing the behavior?**

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

geofjamg:
@annetill @Hadrien-Godard @obrix @jeandemanged what could be a better name that SubStrategy?

PR #2635, opened 2023-07-05
[WIP] Operator Strategy as a list of conditional sub strategies

obrix:
Rework operator strategy to handle several actions stages with specific conditions.

See #2498.

**Other information**:


jeandemanged:
I think it would be better readable and understandable with this renaming:

- OperatorStrategy -> OperatorStrateg**ies**
- OperatorStrategyStage -> OperatorStrategy

but this change might lead to confusion for existing API users ... any advice @annetill @geofjamg @Hadrien-Godard ?

Also thinking about it I would avoid using 'stage' terminology, this term is used in Entsoe CSA profiles, but this is something different, e.g. an SPS in multiples stages 1, 2, 3, ... where e.g. conditions of stage 2 must be evaluated after applications of actions from stage 1 (so 'stage' is different topic in CSA, see cim:Stage.priority)

annetill:
I understand from that PR that the results do not change. Even if we have several stages in an operator strategy, we only report the violations after applying all the actions and we monitor the network state after applying all the actions too. If yes (sorry for asking many times), stage is a bit confusing. All conditions are checked on post contingency state ? In that case, we just need an object to associate a list of actions to a condition. I am not sure it is a stage or a sub strategy at all.

We have the chance to have a great powsybl-open-loadflow to test our API. Let's do the implementation in order to clearly understand what we are doing and if we do what is needed, no ?

obrix:
@annetill Yes we do not have results for stages, we will still have one result per operator strategy, stages are only there to define conditions for packet of actions within an operator strategy.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/strategy/OperatorStrategy.java
```diff
@@ -28,14 +28,31 @@
...
-        this.condition = Objects.requireNonNull(condition);
-        this.actionIds = ImmutableList.copyOf(Objects.requireNonNull(actionIds));
+        this.stages = List.of(new ConditionalActions("default", condition, actionIds));
+    }
+
+    /**
+     * Multiple stage operator strategy
+     * @param id The id of the operator strategy
+     * @param contingencyContext The contingency context in which to apply the operator strategy
+     * @param stages The list of stages for this operator strategy
```
annetill, line 50: FIXME Stages twice

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/strategy/OperatorStrategy.java
```diff
@@ -52,17 +69,7 @@ public ContingencyContext getContingencyContext() {
...
-    public Condition getCondition() {
-        return condition;
-    }
-
-    /**
-     * An ordered list of actions, which will be simulated if the condition holds true.
-     */
-    public List<String> getActionIds() {
-        return actionIds;
+    public List<ConditionalActions> getStages() {
```
annetill, line 72: FIXME stages twice here too.

security-analysis/security-analysis-api/src/test/resources/OperatorStrategyFileExtensionsTest.json
```diff
@@ -4,10 +4,13 @@
     "id" : "id1",
     "contingencyContextType" : "SPECIFIC",
     "contingencyId" : "contingencyId1",
-    "condition" : {
-      "type" : "TRUE_CONDITION"
-    },
-    "actionIds" : [ "actionId1", "actionId2", "actionId3" ],
+    "stages" : [ {
```
annetill, line 7: stages here too.

annetill:
Do we have a json compatibility?

geofjamg (approved):

PR #2636, opened 2023-07-05
Permanent limit is now mandatory in IIDM

annetill:

No

Feature

In IIDM model, it is possible to create an operational limit set with temporary limits only. This leads to a great confusion on what the modeling means. What does it mean being under a temporary limit without any permanent limit?

It is now forbidden to create an operational limit set with temporary limits only: an exception is thrown.

The permanent limit is mandatory in the new v1.12 IIDM version.
To provide backward compatibility with previous IIDM versions, when the permanent limit is missing and temporary limits are present, the following policy is applied:
- if there is a temporary limit with an infinite acceptable duration, its value is used as permanent limit and the said temporary limit is removed;
- else, the permanent limit is computed as a certain percentage of the minimum temporary limits' value:
  - the default percentage is 100%;
  - for IIDM imports, it could be changed using the `iidm.import.xml.missing-permanent-limit-percentage` parameter;
  - for CGMES imports, it could be changed using the `iidm.import.cgmes.missing-permanent-limit-percentage` parameter.

For UCTE imports, since no temporary limits are present, no LoadingLimits is defined if the permanent limit is missing.

- [X] Yes
- [ ] No

## Loading limits

If your project defines custom IIDM implementations, you should implement the following methods:
- In `LoadingLimitsAdder` implementations:
 - `double getTemporaryLimitValue(String name)`
 - `int getTemporaryLimitAcceptableDuration(String name)`
 - `double getLowestTemporaryLimitValue()`
 - `Collection<String> getTemporaryLimitNames()`
 - `void removeTemporaryLimit(String name)`
 - `String getOwnerId()`

`ValidationUtil.checkPermanentLimit` now takes an additional parameter:
- `Collection<LoadingLimits.TemporaryLimit> temporaryLimits`: the temporary limits defined for the same `LoadingLimits` object.

## Serialization / Deserialization

The `readActivePowerLimits`, `readApparentPowerLimits` and `readCurrentLimits` methods of `ConnectableSerDeUtil` now take 2 additional parameters:
- `IidmVersion iidmVersion`: the IIDM version of the network being deserialized;
- `ImportOptions options`: the import options to apply when deserializing the network.

**Other information**:


zamarrenolm (approved):

annetill:
For me or @zamarrenolm : @geofjamg thinks that we have to be resilient to a `NaN` permanent limit if all temporary limits are undefined too. I am not sure to know where I have to put this check. I imagine that in this case, all the operation limit set is dropped in IIDM.

zamarrenolm:
> For me or @zamarrenolm : @geofjamg thinks that we have to be resilient to a `NaN` permanent limit if all temporary limits are undefined too. I am not sure to know where I have to put this check. I imagine that in this case, all the operation limit set is dropped in IIDM.

I am not sure to understand the request. With this change we only accept a set of limits for IIDM if it contains a valid value for permanent limit (not `NaN`), no matter what temporary values the set contains.

annetill:
> > For me or @zamarrenolm : @geofjamg thinks that we have to be resilient to a `NaN` permanent limit if all temporary limits are undefined too. I am not sure to know where I have to put this check. I imagine that in this case, all the operation limit set is dropped in IIDM.
> 
> I am not sure to understand the request. With this change we only accept a set of limits for IIDM if it contains a valid value for permanent limit (not `NaN`), no matter what temporary values the set contains.

Yes for the moment, but @geofjamg asks me to apply this check only if one temporary limit is defined and not `NaN`. Furthermore, after a talk with him, I think we have to check that we don't have any temporary limit with infinite acceptable duration. In that case, we have to merge the permanent limit and this temporary limit by taking the smallest one as permanent limit. All of that directly in IIDM?

So-Fras:
@annetill @zamarrenolm @geofjamg what is an infinite acceptable duration for you? `MAX_VALUE`?
If there is such a temporary limit in the `temporaryLimits` TreeMap, would you like to remove it from `temporaryLimits`?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -1017,6 +1026,8 @@ public boolean disconnectNetworkSideOfDanglingLinesIfBoundaryIsDisconnected() {
         private Xfmr3RatioPhaseInterpretationAlternative xfmr3RatioPhase = Xfmr3RatioPhaseInterpretationAlternative.NETWORK_SIDE;
         private Xfmr3ShuntInterpretationAlternative xfmr3Shunt = Xfmr3ShuntInterpretationAlternative.NETWORK_SIDE;
         private Xfmr3StructuralRatioInterpretationAlternative xfmr3StructuralRatio = Xfmr3StructuralRatioInterpretationAlternative.STAR_BUS_SIDE;
+
+        private double lowestTatlCoefficient = 1;
```
annetill, line 1030: You must now expose this parameter to the cgmes importer. It should be a percentage with default value 100%.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -585,9 +585,8 @@ public static void checkConnected(Validable validable, Boolean connected) {
     }
 
     public static void checkPermanentLimit(Validable validable, double permanentLimit) {
-        // TODO: if (Double.isNaN(permanentLimit) || permanentLimit <= 0) {
         if (permanentLimit <= 0) {
-            throw new ValidationException(validable, "permanent limit must be defined and be > 0");
+            throw new ValidationException(validable, "permanent limit must be > 0");
```
annetill, line 589: Now, you must rebase your PR on #2816 and add the same kind of treatment as what you have done for CGMES conversion for XIIDM/JIIDM version under version 1.12.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -584,9 +585,8 @@ public static void checkConnected(Validable validable, Boolean connected) {
         }
     }
 
-    public static void checkPermanentLimit(Validable validable, double permanentLimit) {
-        // TODO: if (Double.isNaN(permanentLimit) || permanentLimit <= 0) {
-        if (permanentLimit <= 0) {
+    public static void checkPermanentLimit(Validable validable, double permanentLimit, Collection<LoadingLimits.TemporaryLimit> temporaryLimits) {
+        if (Double.isNaN(permanentLimit) && temporaryLimits.isEmpty() || permanentLimit <= 0) {
```
olperr1, line 589: Isn't it rather?
```suggestion
        if (Double.isNaN(permanentLimit) && !temporaryLimits.isEmpty() || permanentLimit <= 0) {
```
(A NaN permanent limit is not allowed when there are temporary limits.)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractLoadingLimits.java
```diff
@@ -3,21 +3,28 @@
...
 import java.util.Objects;
 import java.util.TreeMap;
+import java.util.stream.Collectors;
 
 /**
  * @author Miora Ralambotiana {@literal <miora.ralambotiana at rte-france.com>}
  */
 abstract class AbstractLoadingLimits<L extends AbstractLoadingLimits<L>> extends AbstractOperationalLimits implements LoadingLimits {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractLoadingLimitsAdder.class);
```
olperr1, line 26:
```suggestion
    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractLoadingLimits.class);
```

olperr1 (changes requested):

rolnico (changes requested):

rolnico (approved):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/LoadingLimitsMapping.java
```diff
@@ -34,17 +32,12 @@ public class LoadingLimitsMapping {
...
-                    limits.setPermanentLimit(fixedPermanentLimit);
-                }
+                entry.getValue()
+                        .fixLimits(context.config().getMissingPermanentLimitPercentage(), context::fixed)
+                        .add();
             }
         }
         adders.clear();
     }
+
```
annetill, line 42: Just for me: what is the rule? A space or no space at the end?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractLoadingLimits.java
```diff
@@ -95,4 +103,35 @@ public double getTemporaryLimitValue(int acceptableDuration) {
         TemporaryLimit tl = getTemporaryLimit(acceptableDuration);
         return tl != null ? tl.getValue() : Double.NaN;
     }
+
+    private void checkTemporaryLimits() {
+        // check temporary limits are consistent with permanent
+        if (LOGGER.isDebugEnabled()) {
+            double previousLimit = Double.NaN;
+            boolean wrongOrderMessageAlreadyLogged = false;
+            for (LoadingLimits.TemporaryLimit tl : temporaryLimits.values()) { // iterate in ascending order
```
annetill, line 112: Where do we have the rule (or the check) that values must be ordered? Because it was a question I had, without any clear answer that we have a check on that.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/LoadingLimitsMapping.java
olperr1: I don't think there is a convention about this.
This empty line was added because a new method was created then removed in the successive commits of the PR.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractLoadingLimits.java
```diff
@@ -95,4 +103,35 @@ public double getTemporaryLimitValue(int acceptableDuration) {
...
+                }
+                if (!wrongOrderMessageAlreadyLogged && !Double.isNaN(previousLimit) && tl.getValue() <= previousLimit) {
+                    LOGGER.debug("{} : temporary limits should be in ascending value order", owner.getMessageHeader());
+                    wrongOrderMessageAlreadyLogged = true;
+                }
+                previousLimit = tl.getValue();
+            }
+        }
+        // check name unicity
+        temporaryLimits.values().stream()
```
annetill, line 124: Here it is a breaking change, no? Because as I saifd in my previous comment, nothing prevent from this before your PR.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractLoadingLimitsAdder.java
```diff
@@ -144,31 +142,38 @@ public boolean hasTemporaryLimits() {
         return !temporaryLimits.isEmpty();
     }
 
-    private void checkTemporaryLimits() {
```
annetill, line 147: Ok you just moved some code.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/LoadingLimitsUtil.java
```diff
@@ -0,0 +1,84 @@
...
+            limitFixLogger.log("Operational Limit Set of " + ownerId,
+                    "An operational limit set without permanent limit is considered with permanent limit " +
+                            "equal to lowest TATL value with infinite acceptable duration",
+                    Double.NaN, lowestTemporaryLimitWithInfiniteAcceptableDuration);
+            adder.setPermanentLimit(lowestTemporaryLimitWithInfiniteAcceptableDuration);
+        } else {
+            double firstTemporaryLimit = adder.getLowestTemporaryLimitValue();
+            double percentage = missingPermanentLimitPercentage / 100.;
+            double fixedPermanentLimit = firstTemporaryLimit * percentage;
+            limitFixLogger.log("Operational Limit Set of " + ownerId,
```
annetill, line 77: Can you remove "set" please? Because we have groups/group and limits.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/LoadingLimitsUtil.java
```diff
@@ -0,0 +1,84 @@
...
+            if (adder.getTemporaryLimitAcceptableDuration(name) == MAX_VALUE) {
+                hasTemporaryLimitWithInfiniteAcceptableDuration = true;
+                lowestTemporaryLimitWithInfiniteAcceptableDuration =
+                        Math.min(lowestTemporaryLimitWithInfiniteAcceptableDuration, adder.getTemporaryLimitValue(name));
+                adder.removeTemporaryLimit(name);
+            }
+        }
+
+        if (hasTemporaryLimitWithInfiniteAcceptableDuration) {
+            limitFixLogger.log("Operational Limit Set of " + ownerId,
```
annetill, line 68: Same here, remove "set" that is a CGMES wording.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractLoadingLimits.java
olperr1: Temporary limits are automatically ordered by descending order of their acceptable durations (it's the `temporaryLimit` variable of `AbstractLoadingLimitsAdder`, a `TreeMap` initialized with the `AbstractLoadingLimitsAdder.ACCEPTABLE_DURATION_COMPARATOR` comparator, which is responsible for that).
Therefore, acceptable durations are ordered by descending order, but nothing ensures that values are ordered by ascending order.

PR #2640, opened 2023-07-07
[WIP] CGMES: keep operational limits defined at boundary terminals

zamarrenolm:

Bug fix

Operational limits defined at boundary terminals are ignored.

Operational limits defined at boundary terminals are kept, stored at the corresponding dangling line.
In a first version, no modification is made to the value of the limit.

**Other information**:

To complete the change, we have two alternatives:
  1. Modify the dangling line API to support loading limits at both sides.
  2. Store at the network side (do not modify the dangling line API) but translate the value if needed.

The value must be translated for limits of current if the dangling line is a transformer or a line with different nominal values


nemanja-st:
> @zamarrenolm, @nemanja-st and @rcourtier what is the status of the PR as for me, an operational limit at boundary means nothing. What says QoCDC for that?

@annetill there was an update in QoCDC last year regarding this, it is allowed to have a limit only on one side of TieLine:
image

So it is OK to have on your side of the network and not have to copy it to the X node side.

annetill:
> > @zamarrenolm, @nemanja-st and @rcourtier what is the status of the PR as for me, an operational limit at boundary means nothing. What says QoCDC for that?
> 
> @annetill there was an update in QoCDC last year regarding this, it is allowed to have a limit only on one side of TieLine: image
> 
> So it is OK to have on your side of the network and not have to copy it to the X node side.

Here, the question is: if we have one on boundary side and nothing at network side, do we have a plan to transfer the limit to network side in order to be stored in iidm model?

zamarrenolm:
> > > @zamarrenolm, @nemanja-st and @rcourtier what is the status of the PR as for me, an operational limit at boundary means nothing. What says QoCDC for that?
> > 
> > 
> > @annetill there was an update in QoCDC last year regarding this, it is allowed to have a limit only on one side of TieLine: image
> > So it is OK to have on your side of the network and not have to copy it to the X node side.
> 
> Here, the question is: if we have one on boundary side and nothing at network side, do we have a plan to transfer the limit to network side in order to be stored in iidm model?

That was exactly the purpose of the change: try to handle the case where the limit is defined only at boundary and it has to be "transferred" to the network side to allow evaluation inside PowSyBl. It has to be "transferred" if we do not want to extend the DanglingLine model to have operational limits at boundary side.

About QoCDC: I think that the rule is not considering that one of the equipment at boundary may be a transformer. Current limits can not be directly transferred to the other side in a transformer, they should be to be adjusted by the voltage for comparison. When checking that both TSOs agree on the same limits, the rule should not directly comparing the values at network sides of the same boundary, if one of the segments is a transformer.

annetill:
I close this PR because, reading the QoCDC, the operational limit missing at boundary side is not a big issue and in that case, the network side must be used. I understand that we can have this issue exporting a CGMES file from iidm network, but this document provides a rule to deal with. Note that the rule cannot be applied to transformers, only tie lines.

PR #2641, opened 2023-07-07
[WIP] Slack terminals priority

### LINKED ISSUE #2412: SlackTerminal priorities
Feature

  - there can be only one SlackTerminal per VoltageLevel.
  - over a network, all SlackTerminal-s have the same priority

  - being able to have multiple SlackTerminal-s per VoltageLevel.
  - SlackTerminal-s can be prioritized (over the entire network, not just within a VoltageLevel), e.g. by specifying an integer value (1: highest priority for selection, 2: second priority, and so on ...)

Example use case: cim:SynchronousMachine.referencePriority (in CGMES import).
  - In the case where there are two generators in on VoltageLevel, gen1 with priority 1, gen2 with priority 2, current CGMES importer will put extension only on gen1. If the user disconnects the generator in priority 1, the other generator with priority 2 cannot be picked up with today's implementation of SlackTerminal.
  - In the case of two VoltageLevel-s, VL1 with gen1 of referencePriority1, VL2 with gen2 of referencePriority2, today it is not possible to represent that gen1 has higher priority than gen2 for slack slelection.

  - PowSyBl Version: 5.0.0
  - OS Version: ...

issue raised during discussion on #2402

gverger:

Fixes #2412 

This PR allows several terminals to be defined as "slack terminals" instead of just one.

It is possible to have one slack terminal per voltage level.

Several slack terminals can be set for each voltage level, and priorities define which terminal should be considered as the current slack terminal.

The CGMES importer should work as intended, with `referencePriority` as the slack terminal priority.

I tried to keep it consistent with the current API, but marked some methods as deprecated.

**Other information**:

I am not sure about what I should do with the iidm (de)serialization.
I feel like I am lacking some tests: I added the same tests as there was for the AbstractSlackTerminal, and I keep the old ones to be sure to be compatible, but I doubt it guarantees it works as intended.

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/SynchronousMachineConversion.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/SynchronousMachineConversion.java
@@ -14,8 +14,11 @@ import com.powsybl.cgmes.model.PowerFlow;
 import com.powsybl.iidm.network.EnergySource;
 import com.powsybl.iidm.network.Generator;
 import com.powsybl.iidm.network.GeneratorAdder;
+import com.powsybl.iidm.network.Terminal;
 import com.powsybl.iidm.network.extensions.ActivePowerControlAdder;
 import com.powsybl.iidm.network.extensions.SlackTerminal;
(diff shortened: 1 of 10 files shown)


obrix:
The work on this will be put aside for the time being, we will return to it if the need comes back. Thank you !

PR #2644, opened 2023-07-11
CgmesExport. Export transformers with sorted ends

marqueslanauja:

Feature

In Cgmes export, impedances are specified at `powerTransformerEnd1` and powerTransformer preserves the orientation of the transformer in the iidm model.

In Cgmes export, high voltage winding must always be `powerTransformerEnd1`. Impedances must always be specified at `powerTransformerEnd1`.
In twoWindingsTransformers oriented, in the iidm model, from low to high nominal voltage, the powerTransformerEnd1 must be mapped to the side 2 of the transformer and must specify the impedances. As the shunt admittances of the transformer are defined, in iidm model, on the side 1, the exported cgmes and iidm models will not be equivalent when the shunt admittances are not zero.
 In threeWindingsTransformers endNumber 1, 2, and 3 are mapped to high, medium and low voltage.

**Other information**:


zamarrenolm (approved):

colinepiloquet:
I tested it with the SUV and this fixed most of the warnings. 
Note that in QoCDC v3.3, rules PowerTransformerEndX and PowerTransformerEndR indicated that R and X values on end 1 should be greater than 0.01, which is not always the case in our IGM.
This feature should also probably be optional and by default not activated.

marqueslanauja:
> I tested it with the SUV and this fixed most of the warnings. Note that in QoCDC v3.3, rules PowerTransformerEndX and PowerTransformerEndR indicated that R and X values on end 1 should be greater than 0.01, which is not always the case in our IGM. This feature should also probably be optional and by default not activated.

We can do it in a separated PR if you agree.

colinepiloquet:
> > I tested it with the SUV and this fixed most of the warnings. Note that in QoCDC v3.3, rules PowerTransformerEndX and PowerTransformerEndR indicated that R and X values on end 1 should be greater than 0.01, which is not always the case in our IGM. This feature should also probably be optional and by default not activated.
> 
> We can do it in a separated PR if you agree.

Do you mean adding a parameter to make it optional?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -252,6 +253,12 @@ public String getFormat() {
             "Model description",
             null);
 
+    private static final Parameter EXPORT_TRANSFORMERS_WITH_HIGHEST_VOLTAGE_AT_END1_PARAMETER = new Parameter(
+            EXPORT_TRANSFORMERS_WITH_HIGHEST_VOLTAGE_AT_END1,
+            ParameterType.BOOLEAN,
+            "Export transformers with highest voltage at end1",
+            CgmesExportContext.EXPORT_POWER_FLOWS_FOR_SWITCHES_DEFAULT_VALUE);
+
```
zamarrenolm, line 261: we should define a separate variable for the default value of this parameter

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
```diff
@@ -57,11 +57,11 @@ public class CgmesExportContext {
 
     public static final boolean EXPORT_BOUNDARY_POWER_FLOWS_DEFAULT_VALUE = true;
     public static final boolean EXPORT_POWER_FLOWS_FOR_SWITCHES_DEFAULT_VALUE = true;
+    public static final boolean EXPORT_TRANSFORMERS_WITH_HIGHEST_VOLTAGE_AT_END1_DEFAULT_VALUE = false;
```
zamarrenolm, line 60: Ok, the specific variable is defined, but not used in the definition of the `Parameter`.

zamarrenolm (changes requested):
In `CgmesExport` we should set the value for the new parameter when we build the context, like we do with the rest of configurable behaviour (add a line following the pattern for lines 72-77 of `CgmesExport.java`). If we do not set it in the context, the parameter specified through configuration files or properties will not arrive at the equipment export module.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
annetill: Yes, right.

zamarrenolm (approved):

annetill (approved):

PR #2647, opened 2023-07-12
Tie line: improve remove method

annetill:

No.

When a tie line is removed, the underlying dangling lines `P0` and `Q0` can be updated or not. 

**Other information**:

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/TieLineImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/TieLineImpl.java
@@ -124,8 +124,20 @@ class TieLineImpl extends AbstractIdentifiable<TieLine> implements TieLine {
 
     @Override
     public void remove() {
+        remove(false);
+    }
+
+    @Override
+    public void remove(boolean updateDanglingLines) {
         NetworkImpl network = getNetwork();
         network.getListeners().notifyBeforeRemoval(this);
+
+        if (updateDanglingLines) {
+            danglingLine1.setP0(danglingLine1.getBoundary().getP());
+            danglingLine1.setQ0(danglingLine1.getBoundary().getQ());
+            danglingLine2.setP0(danglingLine2.getBoundary().getP());
+            danglingLine2.setQ0(danglingLine2.getBoundary().getQ());
+        }
 
         // Remove dangling lines
         danglingLine1.removeTieLine();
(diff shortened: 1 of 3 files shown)


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/TieLineImpl.java
```diff
@@ -124,9 +124,21 @@ public double getB2() {
...
+        remove(false);
+    }
+
+    @Override
+    public void remove(boolean updateDanglingLines) {
         NetworkImpl network = getNetwork();
         network.getListeners().notifyBeforeRemoval(this);
 
+        if (updateDanglingLines) {
+            danglingLine1.setP0(danglingLine1.getBoundary().getP());
```
zamarrenolm, line 136: There should be a change in sign. It should be `setP0(-danglingLine1.getBoundary().getP())` (you can check it looking at `DanglingLineBoundaryImpl.java::getP`). `P0, Q0` is a load (consumption) located at the boundary node, and `boundary.getP` is interpreted as the flow of the line seen at the boundary terminal.

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractTieLineTest.java
```diff
@@ -272,6 +273,33 @@ public void testRemove() {
...
+        assertEquals(0.0, line1.getDanglingLine1().getQ0());
+        assertEquals(0.0, line1.getDanglingLine2().getP0());
+        assertEquals(0.0, line1.getDanglingLine2().getQ0());
+        assertEquals(0.0, line2.getDanglingLine1().getP0());
+        assertEquals(0.0, line2.getDanglingLine1().getQ0());
+        assertEquals(0.0, line2.getDanglingLine2().getP0());
+        assertEquals(0.0, line2.getDanglingLine2().getQ0());
+        line1.remove(true);
+        line2.remove(true);
+        assertEquals(-301.316, line1.getDanglingLine1().getP0(), 0.001);
```
zamarrenolm, line 293: Signs of the expected values must be updated according to the change in the implementation.

zamarrenolm (changes requested):

zamarrenolm:
Additionally, we could take into account the generation defined at the dangling line, keeping its targets and assigning the remainder to `(p0, q0)`. OR, simply resetting them (`targetP, targetQ` of `danglineLine.getGeneration()`).

The generation part of the dangling line only makes sense when the dangling line is not paired (not inside a tie line), to simulate the fact that the boundary at that point has support for voltage. When the dangling line is paired, its generation should be ignored for calculations, because the "other" part of the network is already modelled in detail.

When we remove tie lines and update its inner dangling lines, we could try keep the same "setpoints" that we had originally, or reset them (and let the user further update them as needed).

zamarrenolm (approved):

marqueslanauja (approved):

marqueslanauja (approved):

marqueslanauja (approved):

PR #2648, opened 2023-07-12
CGMES : import optional ratedS value of transformers

zamarrenolm:

Bug fix

CGMES attribute `ratedS` of transformer ends is not read.

CGMES attribute `ratedS` of transformer is read and saved in the IIDM model.
The attribute is optional in CGMES.
CGMES specifies that for 2-winding transformers the values for the high and low voltage sides shall be identical.

**Other information**:


marqueslanauja (approved):

annetill (approved):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/InterpretedT3xModel.java
```diff
@@ -66,6 +66,7 @@ static class InterpretedWinding {
         final TapChangerConversion.InterpretedEnd end1;
         final InterpretedEnd2 end2;
         final boolean structuralRatioAtEnd2;
+        final double ratedS;
```
annetill, line 69: @zamarrenolm we have a regression here because of this `double` instead of `Double`. I am going to open a PR for this bug fix and it will lead to a patch release of core.

PR #2660, opened 2023-08-02
add getBusCount() to network and voltage level busbreakerview for symetry

jonenst:

NO

Feature

no getBusCount(), but in all other places where we have a getXStream() we also have getXCount()

add getBusCount in network and voltagelevel busbreakerview

no

**Other information**:
another PR to follow, getBusCount on the busview


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Network.java
```diff
@@ -410,6 +411,16 @@ interface BusBreakerView {
...
 
+        /**
+         * Get the bus count.
+         * <p>
+         * Depends on the working variant.
+         * @see VariantManager
+         */
+        default int getBusCount() {
+            return Ints.checkedCast(getBusStream().count());
+        }
```
olperr1, line 422: To be consistent with the other `getXxxCount()` methods in this interface, we should not propose a default implementation. This way the existing Network's implementation's developers will be aware of the existence of this new method and could provide a more performant implementation.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/VoltageLevel.java
```diff
@@ -707,6 +708,17 @@ interface SwitchAdder extends IdentifiableAdder<Switch, SwitchAdder> {
...
+        /**
+         * Get the bus count.
+         * <p>
+         * Depends on the working variant if topology kind is NODE_BREAKER.
+         *
+         * @see VariantManager
+         */
+        default int getBusCount() {
+            return Ints.checkedCast(getBusStream().count());
+        }
```
olperr1, line 720: Same comment as for `Network.getBusCount()`.

olperr1 (changes requested):

olperr1 (approved):

geofjamg (approved):

PR #2661, opened 2023-08-03
Add getConfiguredBus to network and busbreakervoltagelevel.busbreaker…

jonenst:

no

add APIs to get configured bus, symmetric to getBusbarSection

no APIs to get configured buses, but in some use cases you want to use ids of busbarsections and ids of configuredbuses equivalently.

NO

**Other information**:
uses a hack to implement default methods, but we can remove it, or make it normal by adding APIs to views to identify if the view is the same as the voltage level definition ?


geofjamg (commented):
Is it a good idea to expose in the API the concept of "configured bus"? At least if answer is yes, we should use it everywhere when relevant.
For instance when creating a new bus in a bus/breaker voltage level:
```java
   vl.getBusBreakerView().newConfiguredBus()
       .setId("X")
       .add();
```

PR #2664, opened 2023-08-06
Add busbreakerview::isComputed

jonenst:

no

feature

the busbreakerview has no API to know if it matches the voltage level definition or if it's computed.

Add API


jonenst:
This corresponds to the documentation of https://javadoc.io/doc/com.powsybl/powsybl-core/latest/com/powsybl/iidm/network/VoltageLevel.html, should we use the same words ? view::getStatus -> enum with N/A, readonly, modifiable instead of isComputed only on the busbreakerview ?
```
Depending on the topology model kind of the voltage level a view can have
 the status:

   N/A, it doesn't make sense to take view that is more detailed than the
       model. An exception is thrown when a method is called on an N/A view
   
   modifiable, when the view has the same level of detail than the model
   
   readable only, because the view is a result of a computation on the
       topology model

```

PR #2668, opened 2023-08-07
Fix the Maven Wrapper doc (proxy configuration)

olperr1:

No

Doc fix

The the proxy configuration is false in the Maven Wrapper documentation.

Fixes the proxy configuration in the Maven Wrapper documentation

**Other information**:

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ Notes:
 
 This second step is required to download the Maven distribution.
 
-Once both distributions are retrieved, you can just use `./mvwn` or `mvnw.cmd`, without all this configuration, instead of `mvn`.
+Once both distributions are retrieved, the proxy configuration isn't needed anymore to use `./mvnw` or `mvnw.cmd` commands.
 
 
 ##### Checking your access configuration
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -109,20 +109,34 @@ scripts provided. They will download a compatible maven distribution and use it
...
+./mvnw -DproxyHost=XXX -DproxyPort=XXX -Dhttp.proxyUser=XXX -Dhttp.proxyPassword=XXX -Djdk.http.auth.tunneling.disabledSchemes= clean
+```
+Notes:
+- The 4 `XXX` occurrences should be replaced with your configuration;
+- The `-Djdk.http.auth.tunneling.disabledSchemes=` option should be left empty;
+- Windows users should use `mvnw.cmd` instead of `./mwn`.
+
+This second step is required to download the Maven distribution.
+
+Once both distributions are retrieved, you can just use `./mvwn` or `mvnw.cmd`, without all this configuration, instead of `mvn`.
```
colinepiloquet, line 137: I don't understand "without all this configuration". Maybe it can just be removed?

colinepiloquet: Ah yes ok, I think it is clearer like that

colinepiloquet (approved):

Issue #2373, opened 2022-11-17
Remove limit violation filter from security analysis SPI


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Removal of obsolete component.

* **What is the current behavior?**

The security analysis providers are expected to take into account a `LimitViolationFilter` which can remove some violations from the result.

I would propose to remove that parameter from the SPI because:

- its behaviour can, and probably should, be implemented as post processing of the result
- it's redundant with the violation detector (which design and presence in the SPI is also questionable, indeed)

**This is a proposition, to be discussed.**

geofjamg:
Ok, I would also remove the violation detector as it is impl related

alicecaron:
The  `LimitViolationDetector` has been removed from the `SecurityAnalysis` since this issue: https://github.com/powsybl/powsybl-core/pull/2943

PR #2675, opened 2023-08-11
Dynamic security analysis

Lisrte:


dynamic-security-analysis/src/main/java/com/powsybl/security/dynamic/DynamicSecurityAnalysis.java
```diff
@@ -0,0 +1,295 @@
...
+     */
+    public static final class Runner implements Versionable {
+
+        private final DynamicSecurityAnalysisProvider provider;
+
+        public Runner(DynamicSecurityAnalysisProvider provider) {
+            this.provider = Objects.requireNonNull(provider);
+        }
+
+        public CompletableFuture<SecurityAnalysisReport> runAsync(Network network,
```
annetill, line 59: List<LimitReduction> is missing.The `LimitViolationFilter` must be removed too because it has been decided that it is the role of the application (with users expections) and not the engine (https://github.com/powsybl/powsybl-core/issues/2373). The `LimitViolationDetector` should be removed too because it is not a generic use: some simulators can do it without any provided detector.

annetill: In general: all these `runAsync` and `run` are providing noise. The only reasons for me to provide 2 different constructors is that some engine do not support operator strategies at all. We cannot imagine a commun interface with static security analysis and some kind of a builder. It is what we do when we have not mandatory objects and a longue list.

dynamic-security-analysis/src/main/java/com/powsybl/security/dynamic/DynamicSecurityAnalysisProvider.java
```diff
@@ -0,0 +1,149 @@
...
+     * @param dynamicModelsSupplier supplies list of dynamic models for the dynamic simulation part
+     * @param eventModelsSupplier supplies list of event models for the dynamic simulation part
+     * @param contingenciesProvider provides list of contingencies
+     * @param runParameters runner parameters
+     * @return a {@link CompletableFuture} on {@link SecurityAnalysisResult} that gathers security factor values
+     */
+    CompletableFuture<SecurityAnalysisReport> run(Network network,
+                                                  String workingVariantId,
+                                                  DynamicModelsSupplier dynamicModelsSupplier,
+                                                  EventModelsSupplier eventModelsSupplier,
```
flo-dup, line 90: The EventModelsSupplier should be removed, there is no event we want to add outside the contingencies event.

flo-dup (changes requested):

dynamic-simulation/dynamic-simulation-api/src/main/java/com/powsybl/dynamicsimulation/DynamicSimulationParameters.java
```diff
@@ -38,7 +38,7 @@ public interface ConfigLoader<E extends Extension<DynamicSimulationParameters>>
     public static final String VERSION = "1.0";
 
     public static final int DEFAULT_START_TIME = 0;
-    public static final int DEFAULT_STOP_TIME = 1;
+    public static final int DEFAULT_STOP_TIME = 10;
```
flo-dup, line 41: So, this new value is needed to be able to have a contingency start time between those default start and stop. This highlights an issue: these start and stop times shouldn't be integers but doubles. Besides, dynawo expects float numbers according to the jobs.xsd file. Could we take the opportunity to break this?

annetill (approved):

PR #2685, opened 2023-08-29
CGMES fix export SV: export phase and ratio tap changer if both present

zamarrenolm:

Bug fix

If a transformer has both phase and ratio tap changer, only the tap step for the phase tap changer is exported in the SV file. The SSH file correctly contains information about both tap changers.

Both tap steps are exported to the SV.

**Other information**:

Issue reported on slack cgmes channel by Kristjan Vilgo.


annetill (approved):

colinepiloquet (approved):

So-Fras (approved):

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/StateVariablesExportTest.java
```diff
@@ -237,6 +239,58 @@ void microGridBEWithHiddenTapChangers() throws XMLStreamException {
...
+                .setR(0.0)
+                .setX(0.0)
+                .setB(0)
+                .setG(0)
+                .setRho(1.0)
+                .setAlpha(20)
+                .endStep()
+                .add();
+
+        // as all the tap positions are unique we check that RTC and PTC tap positions are in the sv file
```
annetill, line 285: It is just a unit test but the assertion are not really clean.

zamarrenolm: it was the simplest option to avoid re-importing the case and checking the tap position (costly in testing time), as the change in the code was very simple and only affected the SV output, but we will implement a more clear alternative.

PR #2690, opened 2023-09-01
Reporter for xml conversion - Rework

olperr1:

No

Feature

When importing a XIIDM network, no reporter can be used.

It introduces a `ReporterContext`, working as a `Reporter` pile, in the `Network`. During XIIDM network imports, a reporter is added on the pile and used to report element creations.
A multi-thread version of the context is also available.

<br/>

**Other information**:

***For the migration guide***
If you had define your own `Network` implementation, you should implement the following methods:

- `public void allowReporterContextMultiThreadAccess(boolean allow)`
- `public ReporterContext getReporterContext()`

The simplest way is to copy the ones used by `NetworkImpl`:

```java
    private AbstractReporterContext reporterContext;

    @Override
    public void allowReporterContextMultiThreadAccess(boolean allow) {
        this.reporterContext = Networks.allowReporterContextMultiThreadAccess(this.reporterContext, allow);
    }

    @Override
    public ReporterContext getReporterContext() {
        return this.reporterContext;
    }
```


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/AbstractReporterContext.java
```diff
@@ -0,0 +1,35 @@
...
+            pushReporter(it.next());
+        }
+    }
+
+    /**
+     * <p>Returns an iterator on the elements to copy when creating a new ReporterContext from another one.</p>
+     *
+     * @return an Iterator on the elements to copy
+     */
+    protected abstract Iterator<Reporter> copyIterator();
```
flo-dup, line 34: I think it would be clearer to have a `iterator()` abstract method instead  and to do the skip first in copyReporters method above.

flo-dup (changes requested):

flo-dup (approved):

PR #2702, opened 2023-09-14
[WIP] Upgrade all runtime dependencies to the most recent version

jonenst:
The following remains and is left as is (I did a manual check that there was nothing better available)

$ mvn versions:display-dependency-updates
[INFO]   com.google.code.findbugs:annotations ................ 3.0.1 -> 3.0.1u2
[INFO]   com.powsybl:powsybl-build-tools ............................. 13 -> 14
[INFO]   org.apache.groovy:groovy ..................... 4.0.14 -> 5.0.0-alpha-2
[INFO]   org.apache.groovy:groovy-console ............. 4.0.14 -> 5.0.0-alpha-2
[INFO]   org.apache.groovy:groovy-groovysh ............ 4.0.14 -> 5.0.0-alpha-2
[INFO]   org.apache.groovy:groovy-nio ................. 4.0.14 -> 5.0.0-alpha-2
[INFO]   org.apache.maven:maven-core ................... 3.8.5 -> 4.0.0-alpha-7
[INFO]   org.eclipse.rdf4j:rdf4j-common-transaction ......... 4.3.6 -> 5.0.0-M1
[INFO]   org.eclipse.rdf4j:rdf4j-repository-sail ............ 4.3.6 -> 5.0.0-M1
[INFO]   org.eclipse.rdf4j:rdf4j-rio-rdfxml ................. 4.3.6 -> 5.0.0-M1
[INFO]   org.eclipse.rdf4j:rdf4j-sail-memory ................ 4.3.6 -> 5.0.0-M1

TODO: there's at least one relocation I know of that I didn't handle: findbugs -> spotbugs
TODO: are there other relocations ?

**Please check if the PR fulfills these requirements**

- [x] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**Does this PR already have an issue describing the problem?**
NO

**What kind of change does this PR introduce?**

Feature

**What is the current behavior?**

Old deps

**What is the new behavior (if this is a feature change)?**
new deps !

**Does this PR introduce a breaking change or deprecate an API?**

- [ ] The *Breaking Change* or *Deprecated* label has been added
- [ ] The migration guide has been updated in the github wiki *(What changes might users need to make in their application due to this PR?)*
Not sure, maybe worth mentionning 

**Other information**:
based on code from https://github.com/powsybl/powsybl-core/pull/2124

PR #2704, opened 2023-09-18
Upgrade2023 notall2

jonenst:

NO

Upgrade all the depdencies

old runtime deps

recent runtime deps

Biggest change is the groovy package migration ?

**Other information**:
see #2702 
Same but rdf4j and univocity are not upgraded


zamarrenolm:
I've checked that the performance degradation of RDF4J was introduced in 4.3.3.

I've used version 4.3.2 successfully with all CGMES conversion unit tests and with similar performance to the old version we were using (4.2.2).

jonenst:
> I've checked that the performance degradation of RDF4J was introduced in 4.3.3.
good to know. Then probably we'll upgrade to 4.2.2 -> 4.3.2 in this PR, is that what you suggested ?

zamarrenolm:
> > I've checked that the performance degradation of RDF4J was introduced in 4.3.3.
> 
> good to know. Then probably we'll upgrade to 4.2.2 -> 4.3.2 in this PR, is that what you suggested ?

Yes, that is my proposal

geofjamg:
> I've checked that the performance degradation of RDF4J was introduced in 4.3.3.
> 
> I've used version 4.3.2 successfully with all CGMES conversion unit tests and with similar performance to the old version we were using (4.2.2).

We should build a test case showing the performance regression et submit it to RDF4J team

PR #2710, opened 2023-09-20
Proposal for handling UpDown scalable get steady state power instead of throwing exception

zamarrenolm:

`UpDownScalable` throws an exception in method `getSteadyStatePower`.
GSLK module in powsybl-entsoe creates up/down scalables. Some of the features in GSLK fail if we do not allow scaling on these objects.

Because `UpDownScalable` delegates in its `upScalable` or `downScalable` inner objects to perform the scaling, it seems to make sense to delegate also the `getSteadyStatePower` method to the proper sub-object. The sub-object is selected depending on the `asked` parameter of the scaling, so to do the proper delegation we must pass it around also in `getSteadyStatePower`.

**Other information**:


rolnico (approved):
I don't see a better and still logical way to fix it

Issue #2707, opened 2023-09-18
CGMES export bugs 


annetill:
### Describe the current behavior

- If reactive capability curve present, no need of `minQ` and `maxQ`.
- Missing `AcLineSegment` bch: maybe because there are equal to zero (we have this condition in the code).
- Missing `FossilFuel` and `HydroPowerPlant`.
- `TapChangerControl` missing (phase control and voltage control).
- SynchronousMachine changes `operatingMode` in SSH even though the injection is 0. This is not that relevant unless the machine has control enabled.

To be continued.

marqueslanauja:
`minQ` and `maxQ` are required in `ExternalNetworkInjection` and are optional in `SynchronousMachine` and `EquivalentInjection`.
In concordance with specifications, if a reactive capability curve is present, any provided `minQ` / `maxQ` limit attributes shall be ignored. In my opinion, it is not a problem to always export `minQ` and `maxQ`. 

`gch` it is optional and it is considered zero if it is not present. In my opinion, it is ok not to export it when is zero.

I will fix the rest of the issues in separated PRs. I will start with the `TapChangerControl` issue.

zamarrenolm:
About preserving `FossilFuel` information:

We plan to store (in a property) only the `FossilFuel` type (coal, oil, gas, ...), that is the only attribute visible through the EQ profile. The generic CIM definition, and other ENTSO-E profiles related to market and transparency, have other detailed attributes (cost, dispatch cost, handling cost, efficiency factor, heat content, mixture, ...). We assume we won't need to keep these attributes in IIDM in the near future.

zamarrenolm:
About preserving `HydroPowerPlant` information:

We plan to store (in a property) only the storage type of hydro power plant (run of river, pumped storage, storage).

geofjamg:
Using property is very weak, we should think about enriching the generator to have something better than just energy source with just a few values

zamarrenolm:
Both `FossilFuel` and `HydroPowerPlant` characteristics are defined in CGMES using independent objects associated with the generating unit. We assume we do not need to preserve the identifiers of these objects, to reduce the amount of data attached to the IIDM Network. This means that, when re-exported, the relationship between synchronous machines and their extended attributes will be re-created through new identifiers.

For the use-case that we want to support, is this valid ?

zamarrenolm:
[...]

PR #2716, opened 2023-09-22
Cgmes export fix tap changer control issue

marqueslanauja:

Fixes one of the items of #2707 (TapChangerControl missing (phase control and voltage control))

Bug fix

The information required to export a tap changer control to CGMES is stored in the `CgmesTapChanger` extension in IIDM.
Starting from an IIDM Network not imported from CGMES, or making changes to the tap changer control after importing from CGMES, may result in the extension not being created or incomplete.

Before export, it is ensured that all required information about tap changer regulation exists in the `CgmesTapChanger` extension.

**Other information**:


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportUtil.java
```diff
@@ -308,6 +309,86 @@ public static <C extends Connectable<C>> void setCgmesTapChangerType(C eq, Strin
         }
     }
 
+    // tap changer is exported as it is modelled in IIDM, always at end 1
+    public static void addTapChangerExtensionsNecessaryForExporting(TwoWindingsTransformer twt) {
```
zamarrenolm, line 313: naming suggestion: use `addUpdateCgmesTapChangerExtension` instead of `addTapChangerExtensionsNecessaryForExporting`

marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportUtil.java
```diff
@@ -308,6 +309,86 @@ public static <C extends Connectable<C>> void setCgmesTapChangerType(C eq, Strin
...
+            String newTapChangerId = CgmesExportUtil.getUniqueId();
+            twt.addAlias(newTapChangerId, aliasType);
+            return newTapChangerId;
+        } else {
+            return optionalTapChangerId.get();
+        }
+    }
+
+    private static boolean regulatingControlIsDefined(RatioTapChanger rtc) {
+        return !Double.isNaN(rtc.getTargetV()) && rtc.getTargetV() > 0.0;
```
zamarrenolm, line 364: we should also check that regulating terminal is not null

zamarrenolm: and/or verify the flag for load tap changing capabilities

marqueslanauja: Done. Only the regulating terminal checking is needed

zamarrenolm (approved):

zamarrenolm:
> Merging `TapChanger` on `PowerTransformerEnd` might lead to loss of `TapChanger` and `TapChangerControl` in exported `SSH` and `SV` cimxml files. This poses a risk that `SSH` and `SV` are unusable to user for power flow analysis because not all equipment can be mapped.

For power flow analysis we need only to export `SSH` and `SV` (no need to re-export the `EQ`file). In this situation, data about the "hidden" tap changer is still exported in the `SSH`, with the attribute `controlEnabled` set to `false`. 

It is true that the tap changer control data is still not exported, but we have to consider that this will only happen if there are two tap changer controls in the transformer, one on each end. That is, the transformer is able to regulate changing positions in two tap changers. If there is only one tap changer control in the transformer, we keep it as the "main" tap changer, and the other tap changer (without control) will be the "hidden" one.

annetill:
> > Merging `TapChanger` on `PowerTransformerEnd` might lead to loss of `TapChanger` and `TapChangerControl` in exported `SSH` and `SV` cimxml files. This poses a risk that `SSH` and `SV` are unusable to user for power flow analysis because not all equipment can be mapped.
> 
> For power flow analysis we need only to export `SSH` and `SV` (no need to re-export the `EQ`file). In this situation, data about the "hidden" tap changer is still exported in the `SSH`, with the attribute `controlEnabled` set to `false`.
> 
> It is true that the tap changer control data is still not exported, but we have to consider that this will only happen if there are two tap changer controls in the transformer, one on each end. That is, the transformer is able to regulate changing positions in two tap changers. If there is only one tap changer control in the transformer, we keep it as the "main" tap changer, and the other tap changer (without control) will be the "hidden" one.

Import and export are consistent, and our power flow cannot handle this kind of rare configuration. So I agree to not support that.

annetill (approved):

PR #2717, opened 2023-09-26
CGMES option to force import of node/breaker as bus/breaker

### LINKED ISSUE #2700: Force CGMES import to bus/breaker topology

Sometimes, IGMs in CIM-CGMES are in node/breaker topology. But in these files, we have just a small number of breakers (in general just between bus bar sections). The consequence are multiple. For example, a line contingency with propagation could lead to the loose of many equipments. The line cannot be opened also. In IIDM, this kind of description are typically the bus/breaker topology. 

We would like a parameter in CGMES import that forces to import a file in bus/breaker topology. We have a solution (removing the Operation profile) but it is time costly. 

### Describe the motivation

_No response_

### Extra Information

_No response_

zamarrenolm:

Fixes #2700 

A new parameter has been introduced. Set `iidm.import.cgmes.import-node-breaker-as-bus-breaker = true` to force the model be imported as bus/breaker (TP instance file is required).

**Other information**:


annetill (approved):

jeandemanged (approved):
thanks !
don't forget to document this option on powsybl web site after the release

PR #2730, opened 2023-10-02
Disconnect dangling lines disconnected at boundary

zamarrenolm:

Feature

When an equipment (ACLS, switch, 2w transformer, equivalent branch) located at boundary has the terminal at boundary side disconnected, the equivalent injection is set to zero, but there is not way to indicate that it should not be connected at boundary side.

This approach works for an individual grid model (IGM): leaving p0, q0 of dangling line set to zero when the boundary terminal is disconnected is electrically equivalent (a reactive injection at network side would be seen, if this side is connected).

But is creates a problem when the network is merged with another one: the tie line will be considered as connected. 

To ensure a correct merging is made, we introduce a parameter (active by default) that forces the disconnection of the network side of dangling lines when its boundary terminals are disconnected. Calculation with the IGM introduces a (minor) approximation, but the merging will be more correct.

The parameter to control this behaviour is `iidm.import.cgmes.disconnect-dangling-line-if-boundary-side-is-disconnected`. It is `true` by default. 

**Other information**:

### DIFF
--- a/cgmes/cgmes-conformity/src/main/resources/conformity-modified/cas-1.1.3-data-4.0.3/MicroGrid/BaseCase/BC_BE_v2_line_disconnected_at_boundary_node/MicroGridTestConfiguration_BC_BE_SSH_V2.xml
+++ b/cgmes/cgmes-conformity/src/main/resources/conformity-modified/cas-1.1.3-data-4.0.3/MicroGrid/BaseCase/BC_BE_v2_line_disconnected_at_boundary_node/MicroGridTestConfiguration_BC_BE_SSH_V2.xml
@@ -1,0 +1,286 @@
+<?xml version="1.0" encoding="utf-8"?>
+<rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns:cim="http://iec.ch/TC57/2013/CIM-schema-cim16#" xmlns:entsoe="http://entsoe.eu/CIM/SchemaExtension/3/1#" xmlns:md="http://iec.ch/TC57/61970-552/ModelDescription/1#">
+  <md:FullModel rdf:about="urn:uuid:52b712d1-f3b0-4a59-9191-79f2fb1e4c4e">
+    <md:Model.created>2014-10-24T11:42:40</md:Model.created>
+    <md:Model.scenarioTime>2014-06-01T10:30:00</md:Model.scenarioTime>
+    <md:Model.version>2</md:Model.version>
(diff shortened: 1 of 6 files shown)


cgmes/cgmes-conformity/src/main/resources/conformity-modified/cas-1.1.3-data-4.0.3/MicroGrid/BaseCase/BC_BE_v2_line_disconnected_at_boundary_node/MicroGridTestConfiguration_BC_BE_SSH_V2.xml
```diff
@@ -0,0 +1,286 @@
+<?xml version="1.0" encoding="utf-8"?>
+<rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns:cim="http://iec.ch/TC57/2013/CIM-schema-cim16#" xmlns:entsoe="http://entsoe.eu/CIM/SchemaExtension/3/1#" xmlns:md="http://iec.ch/TC57/61970-552/ModelDescription/1#">
+  <md:FullModel rdf:about="urn:uuid:52b712d1-f3b0-4a59-9191-79f2fb1e4c4e">
```
annetill, line 3: We have an issue with CGMES conversion tests. The CI spends 20 minutes to run them. It seems that they are too big in general and more comparable to integration tests than to unit tests. We have to think about this issue with @flo-dup and maybe think about smallest network as we have the export now.

annetill (approved):

cgmes/cgmes-conformity/src/main/resources/conformity-modified/cas-1.1.3-data-4.0.3/MicroGrid/BaseCase/BC_BE_v2_line_disconnected_at_boundary_node/MicroGridTestConfiguration_BC_BE_SSH_V2.xml
zamarrenolm: > We have an issue with CGMES conversion tests. The CI spends 20 minutes to run them. It seems that they are too big in general and more comparable to integration tests than to unit tests. We have to think about this issue with @flo-dup and maybe think about smallest network as we have the export now.

We will think about it, there are multiple alternatives: we either could "strip down" the test cases, removing the non-essential data, run all tests only on some builds/under a specific configuration, or move some of them to a separate repo.

PR #2739, opened 2023-10-05
Cgmes export: Preserve original classes of loads

marqueslanauja:

Feature

An IIDM load is exported to CGMES as an `EnergySource`, `EnergyConsumer`, `ConformLoad`, or `NonConformLoad` based on certain rules.
The original classes, `AsynchronousMachine` and `StationSupply` are never preserved.
LoadResponseCharacteristicModel is not exported.

The original class is recorded as a property during the import process.
If only the `SSH` file is exported, the original class is always preserved in the `SSH` file to be coherent with the original `EQ` file.
If both files are exported the original class will be preserved if it is coherent with the current attributes of the load. 
Loads exported as `AsynchronousMachine` loose the regulation control data.
LoadResponseCharacteristicModel is now exported.

**Other information**:


nemanja-st:
How is imported `SvInjection` exported?

Importing `SvInjection` as a load can lead to issues with `KirchhoffsFirstLaw` after power flow results export since the object is not defined in equipment (this is an issue only for Entsoe exchanges where this check is applied).

marqueslanauja:
> `AsynchronousMachine` can have a link to `RegulatingControl`.
> 
> If it's imported as a load, is the reference to `RegulatingControl` (and the object) lost on re-export?

Yes, all the RegulatingControl data is lost during the export process.

marqueslanauja:
> How is imported `SvInjection` exported?
> 
> Importing `SvInjection` as a load can lead to issues with `KirchhoffsFirstLaw` after power flow results export since the object is not defined in equipment (this is an issue only for Entsoe exchanges where this check is applied).

Loads created from SvInjections are not exported.

nemanja-st:
`AsynchronousMachine` can have a reference to `RegulatingControl`.
Modeling ASM as a load and not keeping any info about RC will lead to loss of RC in exported `SSH` cimxml file.
This poses a risk that `SSH` is unusable to user for power flow analysis because RC cannot be mapped.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
```diff
@@ -546,24 +548,42 @@ private static void writeEquivalentInjection(String cgmesId, double p, double q,
...
-                } else {
-                    writeSshEnergyConsumer(context.getNamingStrategy().getCgmesId(load), load.getP0(), load.getQ0(), load.getExtension(LoadDetail.class), cimNamespace, writer, context);
+                String className = obtainLoadClassName(load, context);
+                switch (className) {
+                    case CgmesNames.ASYNCHRONOUS_MACHINE ->
+                            writeAsynchronousMachine(context.getNamingStrategy().getCgmesId(load), load.getP0(), load.getQ0(), cimNamespace, writer, context);
+                    case CgmesNames.ENERGY_SOURCE ->
+                            writeEnergySource(context.getNamingStrategy().getCgmesId(load), load.getP0(), load.getQ0(), cimNamespace, writer, context);
+                    case CgmesNames.ENERGY_CONSUMER, CgmesNames.CONFORM_LOAD, CgmesNames.NONCONFORM_LOAD, "StationSupply" ->
+                            writeSshEnergyConsumer(context.getNamingStrategy().getCgmesId(load), load.getP0(), load.getQ0(), load.getExtension(LoadDetail.class), cimNamespace, writer, context);
```
colinepiloquet, line 561: This class should also be modified, because it reassess the class for the Load with the LoadDetails.

marqueslanauja: Done. Agree with you.

zamarrenolm (approved):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportUtil.java
```diff
@@ -213,6 +233,22 @@ public static String loadClassName(LoadDetail loadDetail) {
         return CgmesNames.ENERGY_CONSUMER;
     }
 
+    private static boolean isConformLoad(LoadDetail loadDetail) {
```
annetill, line 236: Are you sure about this definition? The fixed powers can be non zero, no?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
```diff
@@ -278,24 +277,72 @@ private static String writeLoadGroups(String networkId, Collection<LoadGroup> fo
     private static void writeLoads(Network network, LoadGroups loadGroups, String cimNamespace, XMLStreamWriter writer, CgmesExportContext context) throws XMLStreamException {
         for (Load load : network.getLoads()) {
             if (context.isExportedEquipment(load)) {
-                LoadDetail loadDetail = load.getExtension(LoadDetail.class);
-                String className = CgmesExportUtil.loadClassName(loadDetail);
+                String className = CgmesExportUtil.loadClassName(load);
```
annetill, line 280: obtainLoadClassName

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportUtil.java
zamarrenolm: The definition has not been changed from the previous one. The condition just has been moved to a specific method. If the load is not considered as "pure conform" (only variable part) or "pure non-conform" (only fixed part) it is declared as "non-conform", but this is not seen here now.

I agree that, if we decide to consider as "conform" also loads with variable part that may have a fixed part, it should be explicit in these conditions.

Then we should express it as: 

`isConform === loadDetail != null && (loadDetail.variableP != 0 || loadDetail.variableQ != 0)`

no matter which values of `fixedP`, `fixedQ`.

What do yo think @marqueslanauja ?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
zamarrenolm: When we decide that we export the EQ profile we determine the "best" match for CGMES load class that corresponds to the current IIDM load (that may have been modified after import). If the call `obtainLoadClassName` here we will be always exporting the original class name.

To be checked with @marqueslanauja

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportUtil.java
marqueslanauja: As you said the definition has not been changed. If we decide to change it I think is better to do it in a separated PR.

annetill: @nemanja-st what is the good definition of conform loads?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
annetill: Ok.

annetill:
> > `AsynchronousMachine` can have a link to `RegulatingControl`.
> > If it's imported as a load, is the reference to `RegulatingControl` (and the object) lost on re-export?
> 
> Yes, all the RegulatingControl data are lost during the export process.

Can we write an issue to keep this in mind? I am in a hurry for this PR so it will be managed in a separate PR.

annetill:
> How is imported `SvInjection` exported?
> 
> Importing `SvInjection` as a load can lead to issues with `KirchhoffsFirstLaw` after power flow results export since the object is not defined in equipment (this is an issue only for Entsoe exchanges where this check is applied).

@marqueslanauja do you have the answer?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportUtil.java
nemanja-st: @annetill `ConformLoad` is determined by the class name.
`EnergyConsumer.p` and `q` are the setpoints for all load instances so they cannot be used to differentiate if a load is fixed or variable.
Also, [CGM AC IG 1.3](https://eepublicdownloads.entsoe.eu/clean-documents/CIM_documents/Grid_Model_CIM/CGM_BUILDING_PROCESS_IMPLEMENTATION_GUIDE_AC_PART_1.3.pdf) **10.21** specifies that `EnergyConsumer.pfixed` and `qfixed` are not used in ENTSOe exchanges.

marqueslanauja:
> > How is imported `SvInjection` exported?
> > Importing `SvInjection` as a load can lead to issues with `KirchhoffsFirstLaw` after power flow results export since the object is not defined in equipment (this is an issue only for Entsoe exchanges where this check is applied).
> 
> @marqueslanauja do you have the answer?

When a `SvInjection` is imported a fictitious load is created in the IIDM model. Fictitious loads are not exported as equipment, but they are exported as `SvInjections` in the SV file. Considering the exported SvInjections the exported case is coherent with power flow results.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportUtil.java
annetill: Yes okay, so maybe we should throw or log and keep it like this.

zamarrenolm: Agree.

And by the way, the IIDM `LoadDetail` fixed and variable parts are NOT related to CGMES `EnergyConsumer.pfixed,qfixed` values.

When importing from CGMES, if the energy consumer is a `ConformLoad`, its `p,q` setpoints are set in the `LoadDetail` variable part. If the energy consumer is a `NonConformLoad`, they are copied to the fixed part. See `EnergyConsumerConversion::setLoadDetail`.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
```diff
@@ -278,24 +277,72 @@ private static String writeLoadGroups(String networkId, Collection<LoadGroup> fo
...
-                    // as equivalent injections.
-                    writeEnergySource(loadId, load.getNameOrId(), cimNamespace, writer, context);
-                } else {
-                    String loadGroup = loadGroups.groupFor(className);
-                    EnergyConsumerEq.write(className, loadId, load.getNameOrId(), loadGroup, context.getNamingStrategy().getCgmesId(load.getTerminal().getVoltageLevel()), cimNamespace, writer, context);
+                switch (className) {
+                    case CgmesNames.ASYNCHRONOUS_MACHINE ->
+                            writeAsynchronousMachine(loadId, load.getNameOrId(), cimNamespace, writer, context);
+                    case CgmesNames.ENERGY_SOURCE -> writeEnergySource(loadId, load.getNameOrId(), cimNamespace, writer, context);
+                    case CgmesNames.ENERGY_CONSUMER, CgmesNames.CONFORM_LOAD, CgmesNames.NONCONFORM_LOAD, "StationSupply" -> {
```
annetill, line 286: StationSupply -> use constant

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
```diff
@@ -546,24 +547,42 @@ private static void writeEquivalentInjection(String cgmesId, double p, double q,
...
-                    writeEnergySource(cgmesId, load.getP0(), load.getQ0(), cimNamespace, writer, context);
-                } else {
-                    writeSshEnergyConsumer(context.getNamingStrategy().getCgmesId(load), load.getP0(), load.getQ0(), load.getExtension(LoadDetail.class), cimNamespace, writer, context);
+                String className = obtainLoadClassName(load, context);
+                switch (className) {
+                    case CgmesNames.ASYNCHRONOUS_MACHINE ->
+                            writeAsynchronousMachine(context.getNamingStrategy().getCgmesId(load), load.getP0(), load.getQ0(), cimNamespace, writer, context);
+                    case CgmesNames.ENERGY_SOURCE ->
+                            writeEnergySource(context.getNamingStrategy().getCgmesId(load), load.getP0(), load.getQ0(), cimNamespace, writer, context);
+                    case CgmesNames.ENERGY_CONSUMER, CgmesNames.CONFORM_LOAD, CgmesNames.NONCONFORM_LOAD, "StationSupply" ->
```
annetill, line 559: Same, use constant

annetill (approved):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
marqueslanauja: Done

PR #2755, opened 2023-10-24
[Shortcircuit API] Add detailedReport in parameters

colinepiloquet:

No

Feature

No parameter

Adds a parameters indicating if the functional logs (from reporter) should be detailed or aggregated

**Other information**:

### DIFF
--- a/shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitParameters.java
+++ b/shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitParameters.java
@@ -41,7 +41,8 @@ public class ShortCircuitParameters extends AbstractExtendable<ShortCircuitParam
     // VERSION = 1.1 withVoltageMap -> withFortescueResult and withVoltageResult
     // VERSION = 1.2 subTransientCoefficient, withLoads, withShuntCompensators, withVSCConverterStations, withNeutralPosition,
     //                initialVoltageProfileMode, voltageRanges
-    public static final String VERSION = "1.2";
+    // VERSION = 1.3 detailedLog
+    public static final String VERSION = "1.3";
 
     private boolean withLimitViolations = DEFAULT_WITH_LIMIT_VIOLATIONS;
     private boolean withFortescueResult = DEFAULT_WITH_FORTESCUE_RESULT;
@@ -56,6 +57,7 @@ public class ShortCircuitParameters extends AbstractExtendable<ShortCircuitParam
     private boolean withNeutralPosition = DEFAULT_WITH_NEUTRAL_POSITION;
     private InitialVoltageProfileMode initialVoltageProfileMode = DEFAULT_INITIAL_VOLTAGE_PROFILE_MODE;
     private List<VoltageRange> voltageRanges;
+    private boolean detailedReport = DEFAULT_DETAILED_REPORT;
 
     /**
      * Load parameters from platform default config.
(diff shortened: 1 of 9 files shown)


shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitParameters.java
```diff
@@ -82,7 +84,8 @@ public static ShortCircuitParameters load(PlatformConfig platformConfig) {
                         .setWithVSCConverterStations(config.getBooleanProperty("with-vsc-converter-stations", DEFAULT_WITH_VSC_CONVERTER_STATIONS))
                         .setWithNeutralPosition(config.getBooleanProperty("with-neutral-position", DEFAULT_WITH_NEUTRAL_POSITION))
                         .setInitialVoltageProfileMode(config.getEnumProperty("initial-voltage-profile-mode", InitialVoltageProfileMode.class, DEFAULT_INITIAL_VOLTAGE_PROFILE_MODE))
-                        .setVoltageRanges(getVoltageRangesFromConfig(config)));
+                        .setVoltageRanges(getVoltageRangesFromConfig(config))
+                        .setDetailedReport(config.getBooleanProperty("detailed-report", DEFAULT_DETAILED_REPORT)));
```
olperr1, line 88: This parameter can be tested in adding:

- in `src/test/resources/config.yml`:
```yml
detailed-report: false
```
- in `ShortCircuitParametersTest.testConfigLoader()`:
```java
assertFalse(parameters.isDetailedReport());
```

shortcircuit-api/src/test/java/com/powsybl/shortcircuit/ShortCircuitParametersTest.java
```diff
@@ -278,7 +278,35 @@ void readVersion12() {
...
+        assertTrue(parameters.isWithNeutralPosition());
+        assertEquals(InitialVoltageProfileMode.CONFIGURED, parameters.getInitialVoltageProfileMode());
+        List<VoltageRange> voltageRanges = parameters.getVoltageRanges();
+        assertEquals(3, voltageRanges.size());
+        assertEquals(1.05, voltageRanges.get(0).getRangeCoefficient());
+        assertEquals(Range.between(380., 410.), voltageRanges.get(0).getRange());
+        assertEquals(1.1, voltageRanges.get(1).getRangeCoefficient());
+        assertEquals(Range.between(0., 225.), voltageRanges.get(1).getRange());
+        assertEquals(1.09, voltageRanges.get(2).getRangeCoefficient());
+        assertEquals(Range.between(230., 375.), voltageRanges.get(2).getRange());
```
olperr1, line 294: The default value can be tested, for instance here:
```suggestion
        assertEquals(Range.between(230., 375.), voltageRanges.get(2).getRange());
        assertTrue(parameters.isDetailedReport());
```

olperr1 (changes requested):

shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitParameters.java
colinepiloquet: You are right, I will add that right now, thank you

shortcircuit-api/src/test/java/com/powsybl/shortcircuit/ShortCircuitParametersTest.java
colinepiloquet: Actually this is just an improvement of the test of version 1.2 that does not have this new parameter. The new parameters introduced in this version were not introduced so I added them, like that the test make more sense, testing all the new parameters. 
To test `detailedReport`, I created a test that is right below

olperr1: I thought that the default value of the parameter wasn't covered by any test. But in fact, `writeExtension()` will fail if the default value is changed. 
=> No need for another test.

olperr1 (approved):

Issue #1131, opened 2020-01-24
GroundDisconnector not supported during CGMES import


MioRtia:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
`GroundDisconnector` are ignored during CGMES import.

* **What is the expected behavior?**
`GroundDisconnector` objects should be imported as switches during CGMES conversion.

* **What is the motivation / use case for changing the behavior?**
It is in CGMES documentation and could be used.

* **Please tell us about your environment:**
  - PowSyBl Version: 3.2.0-SNAPSHOT

* **Other information** 
It might be necessary to add another `SwitchKind` (either `GROUND_DISCONNECTOR` or `OTHER`) for switches that do not have defined type.

So-Fras:
Point of attention:
At first glance, nothing prevents someone from creating a switch with the "GROUND_DISCONNECTOR" type between 2 busbar sections. However, it does not make any sense. 
Thus, using a new type of disconnector could be seen as a weak way to model ground disconnectors.

So-Fras:
Another way to model ground disconnectors is to add a new equipment "Ground" that will be connected to the topology and a ground disconnector will just be a disconnector that connect a Ground to the topology.

Question: what kind of equipment would the "Ground" be? It does not look like an Identifiable or an Extendable. Would it be a Singleton?

NB: in the CGMES import and export, one would have to be careful when converting ground disconnectors. One object in CGMES becomes two objects in IIDM (disconnector + ground)

So-Fras:
There is another question about the loadflow calculation: what happen if we decide to close the ground disconnector?
Do we take it into account? If yes, how?

NB: in the current configuration, I think that opening or closing the ground disconnector will have no effect on the loadflow since one side of the ground disconnector is connected to "nothing".

olperr1:
Already fixed.

PR #2757, opened 2023-10-24
Add GROUND_DISCONNECTOR to existing kinds of Switch

So-Fras:

#1131
But, in fact, ground disconnectors are not ignored but classified as disconnectors.

Bug fix

No distinction is made between ground disconnectors and disconnectors in IIDM.

Ground disconnectors and disconnectors are distinct.
Moreover, no internal connection should be added to the node of the ground disconnector that is connected "to the ground".


geofjamg:
I don't understand why we add a new type of switch. Ground disconnectors are just disconnectors and grounding is related to the way it is connected. Nothing related to a new switch type.

geofjamg (changes requested):
Adding a new type is a weak way to model ground disconnectors. What about if you create a switch with ground disconnector between 2 busbar sections ? Nothing prevent to do it but it does not make any sense. A correct way to model this is to add a new equipment "Ground" that will be connected to the topology and a ground disconnector will just be a disconnector that connect a Ground to the topology.

So-Fras:
Thank you @geofjamg for your remarks, I have added them to the #1131 issue.
@annetill we need to talk about it on your return!

annetill:
In order to enrich the modelling choice, that is directly linked to short-circuit analysis, we have to have in mind our "RTE" modelling but also open data modeling. For example:
Screenshot 2023-11-02 at 15 02 42

annetill:
Rework in #2837

PR #2759, opened 2023-10-27
Correct getCountryCount in SubnetworkImpl

colinepiloquet:

No

Bug fix

If two subnetworks have the same country, then this country is counted twice.

The method should return the exact amount of countries.

**Other information**:

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/SubnetworkImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/SubnetworkImpl.java
@@ -93,7 +93,7 @@ public class SubnetworkImpl extends AbstractNetwork {
 
     @Override
     public int getCountryCount() {
-        return (int) getCountryStream().count();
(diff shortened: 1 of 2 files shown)


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/SubnetworkImpl.java
```diff
@@ -93,7 +93,7 @@ public Set<Country> getCountries() {
 
     @Override
     public int getCountryCount() {
-        return (int) getCountryStream().count();
+        return getCountries().size();
     }
 
     private Stream<Country> getCountryStream() {
```
colinepiloquet, line 99: Maybe here we also want to have distinct countries?

olperr1:
Could you add a simple test on this fix (in `AbstractSubnetworksCreationTest`, for instance)?
Note that the case with several countries is already covered in `AbstractSubnetworksExplorationTest`.

olperr1 (approved):

PR #2764, opened 2023-11-03
Add a listener on terminal connection

EtienneLt:

No

Feature

We are notified when a terminal is connected of disconnected.

**Other information**:

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTerminal.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTerminal.java
@@ -17,7 +17,6 @@ import java.util.List;
 import java.util.function.Predicate;
 
 /**
- *
  * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
  */
 abstract class AbstractTerminal implements TerminalExt {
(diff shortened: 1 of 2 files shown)


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/NetworkListener.java
```diff
@@ -36,6 +36,14 @@ default void onElementRemoved(Identifiable identifiable, String attribute, Objec
         // empty default implementation
     }
 
+    default void onElementConnected(Connectable<?> connectable) {
+        // empty default implementation
+    }
+
+    default void onElementDisconnected(Connectable<?> connectable) {
```
EtienneLt, line 43: I'm not sure of what is needed here ? Connectable is enough ?

geofjamg: Why do we not use an attribute like fir the other notifications? Why a specific method?

EtienneLt: done

geofjamg (changes requested):
Could you add a unit test? (one with a bus/breaker network and one with a node/breaker network?)

geofjamg (approved):

PR #2767, opened 2023-11-03
Load Flow Results API enhancements

jeandemanged:

No

Feature

- LoadFlowResult JSON format version bump v1.3 -> v1.4
- support reporting multiple slack buses per component (instead of only one)
- support reporting angle reference bus per component
- support reporting metrics per component
- removed component status SOLVER_FAILED (was too specific, now should use FAILED instead)
- added component statusText field to allow implementations to add additional info about component status
- added new component status NOT_CALCULATED (e.g. for components without PV nodes)
- deprecated LoadFlowResult level `isOk()` method
- added LoadFlowResult global status as a replacement of `isOk()`, with values FULLY_CONVERGED, PARTIALLY_CONVERGED, FAILED
- Load Flow Tool table export updated with new infos

Yes, but with backward compatibility support. Flagging as breaking change solely to encourage LoadFlow implementations to migrate.

**Other information**:


geofjamg (approved):
Looks good to me.

Maybe some additional cleaning could be done (not necessarily in this PR, could be later):
 - isOk() : as far as we enrich the loadflow result model, this method has lower and lower interest. Furthermore being "ok" if one of the connected component is converging is probably just wrong. I would make sense to me to be ok if "all synchronous components of main connected components are converging" . But maybe we should just deprecate this method. Defining that a computation is ok depends on the context and maybe could not be simply summarized by just a ok.
 -  Do we deprecate to metrics at the LoadFlowResult level?

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowResult.java
```diff
@@ -14,19 +15,56 @@
...
             SOLVER_FAILED,
-            FAILED
+            /**
+             * Non-convergence
+             */
+            FAILED,
+            /**
+             * Component was not calculated, e.g. component does not have any PV node
+             */
+            NOT_CALCULATED
```
annetill, line 67: In open load flow, we have for NR status ` NO_CALCULATION`.

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowResult.java
```diff
@@ -47,22 +85,44 @@ enum Status {
...
+         * Get the reference bus id (angle reference)
+         * @return the reference bus id
+         */
+        String getReferenceBusId();
+
+        /**
+         * get slack results of all slack buses
+         * @return list of slack results
+         */
+        List<SlackResult> getSlackResults();
```
annetill, line 110: Just a detail but we provide a list because we can have several slack buses. So why you don't call it `SlackBusResult` and inside a `SlackBusResut`, you just provide a `getId()` and a `getActivePowerMismatch()` ?

jeandemanged:
> Looks good to me.
> 
> Maybe some additional cleaning could be done (not necessarily in this PR, could be later):
> 
> * isOk() : as far as we enrich the loadflow result model, this method has lower and lower interest. Furthermore being "ok" if one of the connected component is converging is probably just wrong. I would make sense to me to be ok if "all synchronous components of main connected components are converging" . But maybe we should just deprecate this method. Defining that a computation is ok depends on the context and maybe could not be simply summarized by just a ok.
> * Do we deprecate to metrics at the LoadFlowResult level?

@geofjamg 

We may indeed deprecate isOk(), instead of just deprecating and then removing it, we could add a status enum instead with 'Valid', 'PartiallyValid', 'Invalid'. isOk() deprecated with legacy behavior would be a synonym of 'PartiallyValid'. Will try to work this in this PR.

I'd leave metrics at LoadFlowResult level, can have its use, e.g. some implementations may be able to only provide global metrics but not component metrics. Also for backward compatibility.

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowResult.java
jeandemanged: Will change for NO_CALCULATION. Also realizing javadoc is wrong, no PV node should be failed. NO_CALCULATION should be for not-energized components, e.g. no generation. Component with generation but without PV node should be flagged FAILED.

UNREALISTIC_STATE is in my opinion OpenLoadFlow specific, and should translate to just FAILED. Also I am not so comfortable with the SOLVER_FAILED status, which is implementation specific. In LoadFlow API I would just leave FAILED.

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowResult.java
jeandemanged: good catch, will change this

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowResult.java
annetill: Just a question: what is exactly "No **more** PV bus"? No PV bus at beginning or during the load flow calculation?
What is the exact criteria of a de-energized component for you?

annetill: Agree with the text status, great idea.

geofjamg: yes, good idea to remove the too specific solver failed and add a text to give more details

geofjamg:
> We may indeed deprecate isOk(), instead of just deprecating and then removing it, we could add a status enum instead with 'Valid', 'PartiallyValid', 'Invalid'. isOk() deprecated with legacy behavior would be a synonym of 'PartiallyValid'. Will try to work this in this PR.

Ok. Maybe valid is not right term.

geofjamg:
> I'd leave metrics at LoadFlowResult level, can have its use, e.g. some implementations may be able to only provide global metrics but not component metrics. Also for backward compatibility.

Ok

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/json/LoadFlowResultDeserializer.java
```diff
@@ -68,7 +68,12 @@ public LoadFlowResult.ComponentResult deserializeComponentResult(JsonParser pars
                 }
                 case "status" -> {
                     parser.nextToken();
-                    status = LoadFlowResult.ComponentResult.Status.valueOf(parser.getValueAsString());
+                    String tempStatus = parser.getValueAsString();
+                    if (version.compareTo("1.4") < 0 && "SOLVER_FAILED".equals(tempStatus)) {
+                        // SOLVER_FAILED removed in v1.4, translated to FAILED
+                        tempStatus = LoadFlowResult.ComponentResult.Status.FAILED.name();
+                    }
+                    status = LoadFlowResult.ComponentResult.Status.valueOf(tempStatus);
```
geofjamg, line 76: @jeandemanged in this case you can maybe fill the statusText with the solver failed information?

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowResult.java
jeandemanged: > Just a question: what is exactly "No **more** PV bus"? No PV bus at beginning or during the load flow calculation? What is the exact criteria of a de-energized component for you?

Bad example ... Should be "No PV bus".

de-energized is a component without any connected equipment with voltage support. In this case Load Flow implementations shoud report NO_CALCULATION. The component is dead / not viable.

If at start of calculation or if during calculation all equipments with voltage support are PQ or change to PQ due to reactive limits -> Component should be flagged FAILED

jeandemanged:
> > We may indeed deprecate isOk(), instead of just deprecating and then removing it, we could add a status enum instead with 'Valid', 'PartiallyValid', 'Invalid'. isOk() deprecated with legacy behavior would be a synonym of 'PartiallyValid'. Will try to work this in this PR.
> 
> Ok. Maybe valid is not right term.

proposed "Fully Valid", "Partially Valid", "Failed"

geofjamg (approved):

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowResult.java
```diff
@@ -14,19 +15,52 @@
...
 
+    /**
+     * Result for one slack bus
+     */
+    interface SlackBusResult {
+        /**
+         * Get the slack bus id.
+         * @return the slack bus id
+         */
+        String getId();
```
annetill, line 30: Indeed small hesitation with `getBusId()`

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowResult.java
```diff
@@ -74,11 +137,70 @@ enum Status {
     }
 
     /**
-     * Get the global status. It is expected to be ok if at least one component has converged.
+     * Load Flow global calculation status, derived from all {@link ComponentResult}, using
+     * {@link ComponentResult.Status}, but excluding non-calculated component i.e. excluding
+     * components having status {@code NO_CALCULATION}
+     */
+    enum Status {
```
annetill, line 144: @jeandemanged these statuses are required in operation in your case?

PR #2768, opened 2023-11-06
CGMES export shunt compensators: do not create new regulating controls if not needed

zamarrenolm:

Bug fix

When a CGMES export is prepared, a regulating control identifier is created for every shunt compensator.

When we export a network, we create regulating control identifiers in advance if we do not already have them from a potential CGMES input and the equipment has regulating capabilities active. For shunt compensators, the regulating control identifiers were always created.

Following the same strategy followed for generators and static var compensators, a regulating control identifier is created for a shunt compensator only if voltage regulation is on, or if the regulating terminal is different of the shunt's own terminal.

**Other information**:


annetill (approved):
Thanks.

PR #2770, opened 2023-11-08
CGMES. Preserve static var compensator regulating control data even if disabled

zamarrenolm:

Bug fix

When importing from CGMES, information about disabled regulation controls for static var compensators is not preserved in the IIDM Network. When the Network is exported back to CGMES (only SSH or full export) the regulating control is not present in the output.

Information about the disabled regulating controls for static var compensators is preserved in the IIDM Network and used in the export. Specifically, its identifier and target value.

**Other information**:

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForStaticVarCompensators.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForStaticVarCompensators.java
@@ -106,7 +106,7 @@ public class RegulatingControlMappingForStaticVarCompensators {
         boolean okSet = false;
         if (!control.enabled && rc.controlEnabled) {
             context.fixed("SVCControlEnabledStatus", () -> String.format("Regulating control of %s is disabled but controlEnabled property is set to true." +
-                    "Equipment and regulatin control properties are used to set local default regulation if local default regulation is reactive power. Else, regulation is disabled.", svc.getId()));
+                    "Equipment and regulating control properties are used to set local default regulation if local default regulation is reactive power. Else, regulation is disabled.", svc.getId()));
             setDefaultRegulatingControl(rc, svc, true, control);
             return false;
         }
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForStaticVarCompensators.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForStaticVarCompensators.java
@@ -75,14 +75,14 @@ public class RegulatingControlMappingForStaticVarCompensators {
         String controlId = rc.regulatingControlId;
         if (controlId == null) {
             LOG.trace("Regulating control Id not present for static var compensator {}", svc.getId());
-            setDefaultRegulatingControl(rc, svc, false);
+            setDefaultRegulatingControl(rc, svc, false, null);
             return;
         }
 
         RegulatingControl control = parent.cachedRegulatingControls().get(controlId);
         if (control == null) {
             context.missing(String.format("Regulating control %s", controlId));
-            setDefaultRegulatingControl(rc, svc, false);
+            setDefaultRegulatingControl(rc, svc, false, null);
             return;
         }
 
@@ -90,6 +90,9 @@ public class RegulatingControlMappingForStaticVarCompensators {
     }
 
     private boolean setRegulatingControl(CgmesRegulatingControlForStaticVarCompensator rc, RegulatingControl control, StaticVarCompensator svc) {
+        // Always save the original RC id,
+        // even if we can not complete setting all the regulation data
+        svc.setProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + "RegulatingControl", rc.regulatingControlId);
 
         // Take default terminal if it has not been defined in CGMES files (it is never null)
         Terminal regulatingTerminal = RegulatingTerminalMapper
@@ -103,8 +106,8 @@ public class RegulatingControlMappingForStaticVarCompensators {
         boolean okSet = false;
         if (!control.enabled && rc.controlEnabled) {
             context.fixed("SVCControlEnabledStatus", () -> String.format("Regulating control of %s is disabled but controlEnabled property is set to true." +
-                    "Equipment properties are used to set local default regulation if local default regulation is reactive power. Else, regulation is disabled.", svc.getId()));
-            setDefaultRegulatingControl(rc, svc, true);
+                    "Equipment and regulatin control properties are used to set local default regulation if local default regulation is reactive power. Else, regulation is disabled.", svc.getId()));
+            setDefaultRegulatingControl(rc, svc, true, control);
             return false;
         }
         if (RegulatingControlMapping.isControlModeVoltage(control.mode.toLowerCase())) {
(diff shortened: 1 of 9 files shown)


colinepiloquet:
I tested it and it works well, thanks.
Just a question to be sure. If in SSH RegulatingCondEq.controlEnabled is true but StaticVarCompensator.q is 0, we want to correct it to controlEnabled false right?

zamarrenolm:
> I tested it and it works well, thanks. Just a question to be sure. If in SSH RegulatingCondEq.controlEnabled is true but StaticVarCompensator.q is 0, we want to correct it to controlEnabled false right?

There are two flags that we have to consider: the `RegulatingCondEq.controlEnabled` and the `RegulatingControl.enabled`. We activate SVC regulation only if both flags are `true`. If the regulation mode is for reactive power output, if there is a regulating control, the `RegulationControl.targetValue` will be used as setpoint. For me a target value of `q = 0` would be valid. Why discard the regulation in this situation ?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
```diff
@@ -343,10 +351,10 @@ private static void writeStaticVarCompensators(Network network, String cimNamesp
...
                 String multiplier;
-                if (regulationMode == StaticVarCompensator.RegulationMode.VOLTAGE) {
+                if (regulationMode == StaticVarCompensator.RegulationMode.VOLTAGE || isValidSvcVolatgeSetpoint(svc.getVoltageSetpoint())) {
                     targetValue = svc.getVoltageSetpoint();
                     multiplier = "k";
-                } else if (regulationMode == StaticVarCompensator.RegulationMode.REACTIVE_POWER) {
+                } else if (regulationMode == StaticVarCompensator.RegulationMode.REACTIVE_POWER || isValidSvcReactivePowerSetpoint(svc.getReactivePowerSetpoint())) {
                     targetValue = svc.getReactivePowerSetpoint();
                     multiplier = "M";
                 } else {
```
marqueslanauja, line 360: Maybe in this way will be more clear

```
if (regulationMode == StaticVarCompensator.RegulationMode.VOLTAGE) {
                    targetValue = svc.getVoltageSetpoint();
                    multiplier = "k";
} else if (regulationMode == StaticVarCompensator.RegulationMode.REACTIVE_POWER) {
                    targetValue = svc.getReactivePowerSetpoint();
                    multiplier = "M";
} else if (regulationMode == StaticVarCompensator.RegulationMode.OFF && isValidSvcVolatgeSetpoint(svc.getVoltageSetpoint())) {
                    targetValue = svc.getVoltageSetpoint();
                    multiplier = "k";
} else if (regulationMode == StaticVarCompensator.RegulationMode.OFF && isValidSvcReactivePowerSetpoint(svc.getReactivePowerSetpoint())) {
                    targetValue = svc.getReactivePowerSetpoint();
                    multiplier = "M";
} else {
                    targetValue = 0;
                    multiplier = "k";
}
```

marqueslanauja (approved):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
zamarrenolm: Agree, I will apply the suggestion

colinepiloquet:
> > I tested it and it works well, thanks. Just a question to be sure. If in SSH RegulatingCondEq.controlEnabled is true but StaticVarCompensator.q is 0, we want to correct it to controlEnabled false right?
> 
> There are two flags that we have to consider: the `RegulatingCondEq.controlEnabled` and the `RegulatingControl.enabled`. We activate SVC regulation only if both flags are `true`. If the regulation mode is for reactive power output, if there is a regulating control, the `RegulationControl.targetValue` will be used as setpoint. For me a target value of `q = 0` would be valid. Why discard the regulation in this situation ?

Ok, thank you for the precision, I did not have in my the `RegulatingCondEq.controlEnabled`. It's clear for me now.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
```diff
@@ -343,10 +351,12 @@ private static void writeStaticVarCompensators(Network network, String cimNamesp
                 // Regulating control could be reactive power or voltage
                 double targetValue;
                 String multiplier;
-                if (regulationMode == StaticVarCompensator.RegulationMode.VOLTAGE) {
+                if (regulationMode == StaticVarCompensator.RegulationMode.VOLTAGE
```
annetill, line 354: I don't understand here... if the static var compensator is `OFF`, the target value can be valid for VOLTAGE and for REACTIVE POWER. For example, '225' can match for both. What we are doing is totally arbitrary...

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForStaticVarCompensators.java
```diff
@@ -127,22 +130,37 @@ private boolean setRegulatingControl(CgmesRegulatingControlForStaticVarCompensat
...
         double targetReactivePower = Double.NaN;
         StaticVarCompensator.RegulationMode regulationMode;
 
+        // Before using the default target value,
+        // We try to keep data from original regulating control if available.
+        // Even if the regulating control was disabled it may have defined a valid target value
         if (RegulatingControlMapping.isControlModeVoltage(rc.defaultRegulationMode.toLowerCase())) {
             regulationMode = onlyReactivePowerReg ? StaticVarCompensator.RegulationMode.OFF : StaticVarCompensator.RegulationMode.VOLTAGE;
-            targetVoltage = rc.defaultTargetVoltage;
+            targetVoltage = isValidVoltageFromRegulatingControl(control) ? control.targetValue : rc.defaultTargetVoltage;
```
annetill, line 160: When you are here, you already know that it is a VOLTAGE control mode (on or off).

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForStaticVarCompensators.java
```diff
@@ -127,22 +130,37 @@ private boolean setRegulatingControl(CgmesRegulatingControlForStaticVarCompensat
...
         } else if (isControlModeReactivePower(rc.defaultRegulationMode.toLowerCase())) {
             regulationMode = StaticVarCompensator.RegulationMode.REACTIVE_POWER;
-            targetReactivePower = rc.defaultTargetReactivePower;
+            targetReactivePower = isValidReactivePowerFromRegulatingControl(control) ? control.targetValue : rc.defaultTargetReactivePower;
```
annetill, line 163: Same here.

annetill (changes requested):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
zamarrenolm: We are assuming if the svc has a valid voltage setpoint (there is a typo in the name of the method) the regulating control is define for regulating voltage. If there  is a valid value for reactive power setpoint, we assume the RC is defined for regulating reactive power.

zamarrenolm: You are right that we are assuming the RC is defined (in the EQ) in a consistent way with the current values of the setpoints. To be reviewed ...

annetill: I don't understand. When we are `OFF`, it could be a change in IIDM network. We can be voltage or reactive power at import, right? For me the method `isValidSvcVoltageSetpoint` is not enough to say that it was regulating voltage. And it does not rely on import data. Am I wrong?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForStaticVarCompensators.java
zamarrenolm: when defining the default regulating control we may be establishing a different kind of regulation that was defined in the original control. This is why we check if the mode defined in the RC is consistent with the one being defined now. I ran into an inconsistency with one of the unit tests we had, this is why this condition has been added.

jeandemanged:
Hi @zamarrenolm @colinepiloquet @annetill @marqueslanauja ,

I am discovering the PR, sorry for being late.
Just note that:
- As per [CGM BP section 10.15.2](https://eepublicdownloads.entsoe.eu/clean-documents/CIM_documents/Grid_Model_CIM/CGM_BUILDING_PROCESS_IMPLEMENTATION_GUIDE_AC_PART_1.3.pdf), the local SVC control attributes should not be used, only the RegulatingControl attributes should be used
- As per QoCDC rule SVCVoltage, RegulatingControl is required (level ERROR)

This should clarify what needs to be done and perhaps simplify the code.

image

image

zamarrenolm:
> Hi @zamarrenolm @colinepiloquet @annetill @marqueslanauja ,
> 
> I am discovering the PR, sorry for being late. Just note that:
> 
> * As per [CGM BP section 10.15.2](https://eepublicdownloads.entsoe.eu/clean-documents/CIM_documents/Grid_Model_CIM/CGM_BUILDING_PROCESS_IMPLEMENTATION_GUIDE_AC_PART_1.3.pdf), the local SVC control attributes should not be used, only the RegulatingControl attributes should be used
> * As per [QoCDC](https://eepublicdownloads.azureedge.net/clean-documents/digital/QualityOfCGMESdatasetsAndCalculations_v3.3.1.pdf) rule SVCVoltage, RegulatingControl is required (level ERROR)
> 
> This should clarify what needs to be done and perhaps simplify the code.
> 
> And, to be clear:
> 
> * yes, some TSOs are not compliant, and you may end up switching OFF an SVC from CGMES inputs where SV indicates it is on (becuase the TSO was using the local SVC control attributes while he should not have)
> * ymaybe some EntsoE test configurations are not compliant too
> 
> image
> 
> image

Thanks for the feedback @jeandemanged. Never too late. We will review it in detail

zamarrenolm:
@jeandemanged: we think we have to provide support for the regulation data defined in the equipment, to allow other non-ENTSO-E processes and users. But we will keep using regulating control information with priority, if it is present.

annetill:
> @jeandemanged: we think we have to provide support for the regulation data defined in the equipment, to allow other non-ENTSO-E processes and users. But we will keep using regulating control information with priority, if it is present.

In the code, we have 
````
        if (control.enabled && rc.controlEnabled) {
            svc.setRegulationMode(regulationMode);
        }
````
Reading the code, I wonder about an enabled regulating control and a disabled property.

zamarrenolm:
> > @jeandemanged: we think we have to provide support for the regulation data defined in the equipment, to allow other non-ENTSO-E processes and users. But we will keep using regulating control information with priority, if it is present.
> 
> In the code, we have
> 
> ```
>         if (control.enabled && rc.controlEnabled) {
>             svc.setRegulationMode(regulationMode);
>         }
> ```
> 
> Reading the code, I wonder about an enabled regulating control and a disabled property.

The regulating control can be shared by multiple equipment. One of the "participating" equipment may be not available for participating at a given moment, but the regulating control could still be enabled, with the rest of the equipment try to "enforcement the setpoint.

In this situation, the RC will be enabled, but the equipment that cannot participate at this time will have controlEnabled set to false. The result is that for this equipment we can not activate the regulation.

jeandemanged:
> > In the code, we have
> > ```
> >         if (control.enabled && rc.controlEnabled) {
> >             svc.setRegulationMode(regulationMode);
> >         }

> The regulating control can be shared by multiple equipment. One of the "participating" equipment may be not available for participating at a given moment, but the regulating control could still be enabled, with the rest of the equipment try to "enforcement the setpoint.
> 
> In this situation, the RC will be enabled, but the equipment that cannot participate at this time will have controlEnabled set to false. The result is that for this equipment we can not activate the regulation.

Good point - I forgot about possibility of RC driving by multiple equipments ...

PR #2773, opened 2023-11-13
CGMES: fix CGM export adding equivalent injection power flows at tie lines

zamarrenolm:

Bug fix

The SV export of a CGM did not include power flow values for the equivalent injections of tie lines. Only equivalent injections of dangling lines adjacent to the external boundaries were considered.

The SV export of a CGM also contains power flows for the equivalent injections corresponding to dangling lines that were grouped in tie lines.
Due to current CGMES importer behaviour, the information about equivalent injections and equivalent injection terminals in a tie line is only available if the CGM has been built from separately imported Networks joined as subnetworks. If the CGM is imported as "assembled" (using all the EQ and SSH files for all IGMs in a single import operation) this information has not been preserved, and thus cannot be used in the SV CGM export. 

**Other information**:

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
@@ -210,7 +210,6 @@ public final class StateVariablesExport {
             equivalentInjectionTerminalP.compute(context.getNamingStrategy().getCgmesIdFromProperty(dl, Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + "EquivalentInjectionTerminal"), (k, v) -> v == null ? -dl.getBoundary().getP() : v - dl.getBoundary().getP());
             equivalentInjectionTerminalQ.compute(context.getNamingStrategy().getCgmesIdFromProperty(dl, Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + "EquivalentInjectionTerminal"), (k, v) -> v == null ? -dl.getBoundary().getQ() : v - dl.getBoundary().getQ());
         });
-        equivalentInjectionTerminalP.keySet().forEach(eiId -> writePowerFlow(eiId, equivalentInjectionTerminalP.get(eiId), equivalentInjectionTerminalQ.get(eiId), cimNamespace, writer, context));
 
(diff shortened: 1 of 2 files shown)


jeandemanged:
The PR mentions as new behavior:

> If the CGM is imported as "assembled" (using all the EQ and SSH files for all IGMs in a single import operation) this information is not available, and thus cannot be used in the SV CGM export.

I might disagree, the information is available.
E.g. using [EntsoE CAS v2.0](https://www.entsoe.eu/Documents/CIM_documents/Grid_Model_CIM/TestConfigurations_packageCASv2.0.zip) the assembled micro grid basecase still contains the 10 Equivalent Injections of the 5 tie lines.
Or do you mean that today the CGMES importer loses the information of these equivalent injections ?

zamarrenolm:
> The PR mentions as new behavior:
> 
> > If the CGM is imported as "assembled" (using all the EQ and SSH files for all IGMs in a single import operation) this information is not available, and thus cannot be used in the SV CGM export.
> 
> I might disagree, the information is available. E.g. using [EntsoE CAS v2.0](https://www.entsoe.eu/Documents/CIM_documents/Grid_Model_CIM/TestConfigurations_packageCASv2.0.zip) the assembled micro grid basecase still contains the 10 Equivalent Injections of the 5 tie lines. Or do you mean that today the CGMES importer loses the information of these equivalent injections ?

Yes, I didn't mean the information is not available in CGMES. It is only that we do not import it in this case. Sorry for the confusion.

zamarrenolm:
I've edited the description in the PR.

I will try to fix the import for assembled models in a separate PR. But be aware that with assembled models will not be able to export separately the EQ, SSH files for each IGM.

annetill (approved):

annetill:
> I've edited the description in the PR.
> 
> I will try to fix the import for assembled models in a separate PR. But be aware that with assembled models will not be able to export separately the EQ, SSH files for each IGM.

We have to re-think about this use case, if we can add sub-neworks during CGMES import for assembled case.

zamarrenolm:
> > I've edited the description in the PR.
> > I will try to fix the import for assembled models in a separate PR. But be aware that with assembled models will not be able to export separately the EQ, SSH files for each IGM.
> 
> We have to re-think about this use case, if we can add sub-neworks during CGMES import for assembled case.

Good idea. It could simplify the logic in the importer. Although it will be complex to identify and remove the unused code if we switch to this strategy.

jeandemanged (approved):

PR #2774, opened 2023-11-13
CGMES import: keeping references to equivalent injections of tie lines in assembled model

zamarrenolm:

Bug fix

When an assembled model was imported, the references to equivalent injections were not stored in the dangling lines corresponding to "inner" ite lines. In an assembled import, there are multiple IGMs (multiple EQ, SSH pairs) loaded together and converted to a single network.

The dangling lines of tie lines created during the import of an assembled model keep a references to the associated equivalent injection and equivalent injection terminal.

**Other information**:

We may re-design how assembled models are imported: a single user call to the CMES importer could analyze the list of instance files to be loaded and partition them by IGM, import each IGM separately and combine them in a single IIDM Network with subnetworks.

### DIFF
--- a/iidm/iidm-reducer/src/main/java/com/powsybl/iidm/reducer/IdentifierNetworkPredicate.java
+++ b/iidm/iidm-reducer/src/main/java/com/powsybl/iidm/reducer/IdentifierNetworkPredicate.java
@@ -15,7 +15,7 @@ import java.util.*;
  */
 public class IdentifierNetworkPredicate implements NetworkPredicate {
 
-    private final Set<String> ids = new HashSet<>();
+    private final Set<String> ids = new LinkedHashSet<>();
 
     public static IdentifierNetworkPredicate of(String... ids) {
         Objects.requireNonNull(ids);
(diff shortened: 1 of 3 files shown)


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/AbstractConductingEquipmentConversion.java
```diff
@@ -351,6 +350,32 @@ private EquivalentInjectionConversion getEquivalentInjectionConversionForDanglin
         }
     }
 
+    protected static EquivalentInjectionConversion getEquivalentInjectionConversionForAssembledDanglingLine(Context context, String boundaryNode, BoundaryLine boundaryLine) {
```
annetill, line 353: What is your definition of assembled dangling lines? This is a wording really close to CGMES world, do you mean paired here?

annetill (approved):

annetill (approved):

annetill:
> I agree with your proposal to unify the two methods to obtain the right `EquivalentInjection` conversion of a given dangling line. I initially wrote two separate functions to make explicit that we have to deal with two very different situations, but the code that you have left, together with the comments, is clear enough. Thanks!

Thank YOU for this feature !

PR #2775, opened 2023-11-14
CGMES. Import input with multiple grid models as a network with subnetworks

zamarrenolm:

Feature

Input data sources that contain multiple individual models are imported all together in a single network. Tie lines between individual models are created during CGMES conversion. Information about the original individual models is not explicitly preserved in the resulting network or its equipment (indirectly, it could be inferred in some scenarios from each country substation, but that is not precise).

When the contents of the data source to be imported are detected as belonging to multiple grid models, the individual networks are imported separately and then merged. The resulting network contains as many subnetworks as individual grid models were assembled. The tie lines are defined during the merge of the imported networks.

The input data source is subdivided based on instance files names or modeling authority. 

When separating input files by modeling authority, we consider individual grid models only for the modeling authorities that have an EQ instance file.

If there is a common SV, it should have the modeling authority of the merging agent.

**Other information**:

This feature can be enabled/disabled and adjusted through configuration using the following parameters:
- `iidm.import.cgmes.cgm-with-subnetworks`. `BOOLEAN`. Set to `true` to import input data source with multiple grid models using subnetworks.
- `iidm.import.cgmes.cgm-with-subnetworks-defined-by`. `STRING`. Set to `FILENAME` to use instance file names. Set to `MODELING_AUTHORITY` to use `ModelingAuthoritySet` from `FullModel` object.


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
```diff
@@ -38,18 +42,26 @@
...
 
-    enum FictitiousSwitchesCreationMode {
+    public enum FictitiousSwitchesCreationMode {
         ALWAYS,
         ALWAYS_EXCEPT_SWITCHES,
-        NEVER;
+        NEVER
+    }
+
+    public enum AssembledSeparatingBy {
```
annetill, line 62: As I said in the previous PR, "assembled" is a name choice hard to understand for the community and for non CGMES expert. In Powsybl, assembled is a network. Here the question is just : do we want to add subnetworks in a network and how do we define a subnetwork. `NAME` is too global, maybe something like `FILE_NAME` (and you can keep `MODELING_AUTHORITY`).

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
```diff
@@ -357,6 +556,8 @@ private void copyStream(ReadOnlyDataSource from, DataSource to, String fromName,
     public static final String STORE_CGMES_CONVERSION_CONTEXT_AS_NETWORK_EXTENSION = "iidm.import.cgmes.store-cgmes-conversion-context-as-network-extension";
     public static final String IMPORT_NODE_BREAKER_AS_BUS_BREAKER = "iidm.import.cgmes.import-node-breaker-as-bus-breaker";
     public static final String DISCONNECT_DANGLING_LINE_IF_BOUNDARY_SIDE_IS_DISCONNECTED = "iidm.import.cgmes.disconnect-dangling-line-if-boundary-side-is-disconnected";
+    public static final String IMPORT_ASSEMBLED_AS_SUBNETWORKS = "iidm.import.cgmes.assembled-as-subnetworks";
```
annetill, line 559: See above, for me `assembled` must not be exposed as parameters.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/CgmesConversionContextExtensionTest.java
```diff
@@ -25,6 +25,7 @@ class CgmesConversionContextExtensionTest {
     void test() {
         Properties properties = new Properties();
         properties.put(CgmesImport.STORE_CGMES_CONVERSION_CONTEXT_AS_NETWORK_EXTENSION, "true");
+        properties.put(CgmesImport.IMPORT_ASSEMBLED_AS_SUBNETWORKS, "false");
```
annetill, line 28: Have you run all the test with `true` behaviour? I understand that for performance issue, we prefer to run unit tests with false, but when I run them in local with true, I fall into errors.

annetill (changes requested):

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/CgmesConversionContextExtensionTest.java
zamarrenolm: some expected results in the unit tests were only valid if we import with the "single network" configuration. I do not think we have to adapt old unit tests to the new feature. We have already added specific unit tests for the new feature

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
zamarrenolm: I see you already made the change

zamarrenolm:
Updated the name of the inner class, and edited the PR description and the documentation for the parameters.

rcourtier (approved):
Tested successfully with both parameters (by Modeling Authority or by FileName).
The merging by filename create Subnetworks regardless of the date (only the sourcingActor is considered). Date handling can be enhanced in a later change.

annetill (approved):

PR #2778, opened 2023-11-15
Convert from Joda-Time to java.time

rolnico:

No

Dependency change

Powsybl-core was still using [Joda-Time](https://www.joda.org/joda-time/) library, a library which is asking its users to move to [java.time](https://docs.oracle.com/javase/8/docs/api/java/time/package-summary.html) since it was implemented in Java 1.8.

Powsybl-core now uses java.time features.

**Other information**:

A [website](https://blog.joda.org/2014/11/converting-from-joda-time-to-javatime.html) showing a comparison between classes from the two libraries.


cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/triplestore/CgmesModelTripleStore.java
```diff
@@ -304,6 +309,32 @@ private DateTime queryDate(String propertyName, DateTime defaultValue) {
         return d;
     }
 
+    /**
+     * Parse a date in ISO format. If the offset is not present at the end (ie. no "Z" nor "+xx:xx" or "+xxxx"), it is
```
flo-dup, line 313: Just wondering why you did need this... joda was doing it automatically?

commons/pom.xml
```diff
@@ -92,22 +89,25 @@
...
+        </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.ow2.sirocco</groupId>
-            <artifactId>sirocco-text-table-formatter</artifactId>
+            <groupId>org.threeten</groupId>
+            <artifactId>threeten-extra</artifactId>
```
flo-dup, line 102: This is needed for Inteval, but I'm wondering if we do really need it? To my knowledge ModuleConfig::getIntervalProperty and ModuleConfig::getOptionalIntervalProperty are not used by any of our users.

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/triplestore/CgmesModelTripleStore.java
geofjamg: agree, there is something to investigate here

rolnico: The difference with Joda is that, while there was a unique class DateTime in Joda, there is now multiple classes in java.time : LocalDateTime, ZonedDateTime, OffsetDateTime.
With Joda, if you parsed "yyyy-MM-dd'T'HH:mm:ss.SSSS+xx:xx" or "yyyy-MM-dd'T'HH:mm:ss.SSSS", you would get the same type of data: a DateTime.
With Java.Time, if you parse "yyyy-MM-dd'T'HH:mm:ss.SSSS+xx:xx", you can get a ZonedDateTime (date + time + offset to UTC) while if you parse "yyyy-MM-dd'T'HH:mm:ss.SSSS", you get only a LocalDateTime (date + time), which has less information. If you want to convert the LocalDateTime to a ZonedDateTime, you have to specify which offset to UTC you are assuming to have.

flo-dup: To avoid this annoying if, you'd better use parseBest, which is meant exactly for that purpose:
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/time/format/DateTimeFormatter.html#parseBest(java.lang.CharSequence,java.time.temporal.TemporalQuery...)

flo-dup (changes requested):

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/triplestore/CgmesModelTripleStore.java
rolnico: done

flo-dup (approved):

PR #2781, opened 2023-11-16
Add a chart of iidm versions and powsybl core versions

EtienneLt:

**Other information**:


README.md
```diff
@@ -159,3 +159,20 @@ You can revert this configuration with the following command:
 
 ### Usage
 Once the configuration is done, you just need to use `./mvnw` instead of `mvn` in your commands.
+
+### IIDM version vs PowSyBl core version
```
olperr1, line 163: This should be a subsection of `Using Maven Wrapper`:

```suggestion
## IIDM version vs PowSyBl core version
```

EtienneLt: ok

EtienneLt: I think it is already because ## Using Maven Wrapper

olperr1: Sorry, I mean "it should **not** be"

README.md
flo-dup, line 1: I don't know if you should put the chart at the end. But anyway the whole readme is such a mess that it's ok :tired_face:

flo-dup (commented):
Couldn't we use this opportunity to start a new README.md from scratch? What you added is the most interesting info of the whole file. There's almost nothing describing what powsybl-core is and contains, whereas the install section is only for very specific users, and the maven wrapper section could be located in a wiki help page.

annetill:
Solved with information in the wiki https://github.com/powsybl/powsybl-core/wiki

PR #2782, opened 2023-11-20
CGMES. SV export load flow status for each topological island

zamarrenolm:

Feature

Load flow status is not exported in SV instance file.

To comply with [QoCDC 3.31](https://eepublicdownloads.azureedge.net/clean-documents/digital/QualityOfCGMESdatasetsAndCalculations_v3.3.1.pdf) rule TIConvergenceStatMissing: _The cim:IdentifiedObject.description of cim:TopologicalIsland shall have one the following string values: “converged” and “diverged” which represents the convergence status of the cim:TopologicalIsland_.

In addition, to comply with QoCDC 3.1 rule KirchhoffsFirstLaw, we have added `SvInjection` objects to compensate the mismatch of slack buses.  

We have considered (and discarded) adding the load flow results object to the exporters, to have the information available during export. Instead, we consider a topological island is converged if all its buses have a valid voltage/angle and its mismatch in q and q is close to zero. That is, the sum of flows at all connected terminals is according to Kirchhoff's first law.

Considerations about checking the sum of flows for connected terminals:

- CGMES topological nodes are created from IIDM bus/breaker buses, 
- Sum of flows is checked at the `BusView` bus obtained from the bus/breaker buses, where all potential retained switches have been discarded.
- If we cannot obtain a `BusView` bus for the given topological node (bus/breaker bus), we consider the bus not solved, hence the case not converged. Should not happen, but an error is logged.
- We keep a cache of buses already analyzed (many bus/breaker buses could point to the same bus view bus).
- If the bus does not contain connected terminals, we consider it solved.
- If the bus is a slack (determined through `SlackTerminal` extension) we do not check the sum (mismatches at slacks are allowed).
- If there is at least a terminal with finite value (not NaN, not Infinite), we check the sum.
- If all terminals have non-finite values, we consider the bus not solved.

Considerations about mismatches of slack buses:
- Mismatch of a slack bus is found at its bus view bus.
- SvInjection compensation must be associated to a Topological Node, that is, a bus/breaker bus.
- We have to ensure that all the mismatch in the slack bus view bus is assigned to the bus/breaker bus that is labeled as slack, to establish the right association.
- This has impact on the way that flows through retained switches are computed. A modification has been made to ensure that potential mismatches in a set of bus/breaker buses connected through retained switches are left at the proper bus/breaker bus (if there is one labeled as slack, it should receive the total mismatch).

**Other information**:

There are 4 new parameters for the CGMES importer that control the behaviour of this feature:
- `iidm.export.cgmes.export-load-flow-status`: A `BOOLEAN` to enabled/disable the export load flow status of topological islands. Default is `true`.
- `iidm.export.cgmes.max-p-mismatch-converged`: A `DOUBLE` value to specify the max mismatch in active power to consider a bus converged (by default 0.1 following QoCDC 3.3.1).
- `iidm.export.cgmes.max-q-mismatch-converged`: A `DOUBLE` value to specify the max mismatch in reactive power to consider a bus converged (by default 0.1).
- `iidm.export.cgmes.export-sv-injections-for-slacks`: A `BOOLEAN` to enabled/disable the export of SV injection objects to compensate the mismatch of slack buses. Default is `true`.


jeandemanged:
@zamarrenolm default max mismatch from QoCDC (and also in PR code) is 0.1, not 0.01.

I adjusted PR description.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
```diff
@@ -60,9 +60,17 @@ public class CgmesExportContext {
     public static final boolean EXPORT_POWER_FLOWS_FOR_SWITCHES_DEFAULT_VALUE = true;
     public static final boolean EXPORT_TRANSFORMERS_WITH_HIGHEST_VOLTAGE_AT_END1_DEFAULT_VALUE = false;
     public static final boolean ENCODE_IDS_DEFAULT_VALUE = true;
+    public static final boolean EXPORT_LOAD_FLOW_STATUS_DEFAULT_VALUE = false;
```
jeandemanged, line 63: i would put a true default value, but I am probably biased

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
```diff
@@ -67,18 +68,23 @@ public static void write(Network network, XMLStreamWriter writer, CgmesExportCon
...
         Map<String, String> angleRefs = buildAngleRefs(network, context);
-        Map<String, List<String>> islands = buildIslands(network, context);
+        List<TopologicalIsland> islands = buildIslands(network, context);
         String cimNamespace = context.getCim().getNamespace();
-        for (Map.Entry<String, List<String>> island : islands.entrySet()) {
-            if (!angleRefs.containsKey(island.getKey())) {
-                Supplier<String> log = () -> String.format("Synchronous component  %s does not have a defined slack bus: it is ignored", island.getKey());
+        for (TopologicalIsland island : islands) {
+            if (!angleRefs.containsKey(island.key)) {
+                Supplier<String> log = () -> String.format("Synchronous component  %s does not have a defined slack bus: it is ignored", island.key);
```
jeandemanged, line 75: rather "does not have a defined angle reference bus"

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
```diff
@@ -125,19 +131,115 @@ private static void buildAngleRefs(Bus bus, Map<String, String> angleRefs, Cgmes
...
-        Map<String, List<String>> islands = new HashMap<>();
+    private static final class TopologicalIsland {
+        static final String CONVERGED = "converged";
+        static final String DIVERGED = "diverged";
+
+        // The key can be a synchronous component number or a topological node identifier
+        final String key;
+        final List<String> topologicalNodes;
+        // We need to export the load flow status for the island,
+        // from QoCDC 3.31 rule TIConvergenceStatMissing:
```
jeandemanged, line 142: 3.31 -> 3.3.1

jeandemanged (changes requested):
minor comments

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
zamarrenolm: Agree with you. I've checked the impact on the export time running the unit tests and it is negligible.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
zamarrenolm: fixed

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
zamarrenolm: fixed

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
```diff
@@ -125,19 +131,115 @@ private static void buildAngleRefs(Bus bus, Map<String, String> angleRefs, Cgmes
...
+            // Instead of checking switch flows, we check that the corresponding bus view bus is balanced
+            // Maybe we will check the same bus view bus more than once
+            Optional<Bus> bus0 = getBusViewBus(bus);
+            if (bus0.isEmpty()) {
+                LOG.warn("Can not check if bus is in accordance with Kirchoff's first law. No BusView bus can be found for: {}", bus);
+                return false;
+            }
+            Complex sumTerminalFlows = bus0.get().getConnectedTerminalStream()
+                    .filter(this::isValidTerminal)
+                    .map(t -> new Complex(t.getP(), t.getQ()))
```
jeandemanged, line 200: @zamarrenolm just wondering what happens when t.getP() and/or t.getQ() is Double.NaN ?

zamarrenolm: The sum will be NaN and the check against max p/q will fail, considering the bus as not according to Kirchoff's first law. I had to add conditions to ignore terminals for static var compensators with regulation OFF and busbar sections to avoid its NaN values.

zamarrenolm: For the rest of situations where a terminal has `NaN`, I think it should be ok to consider the case as `diverged`.

jeandemanged (approved):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -304,6 +310,22 @@ public String getFormat() {
...
+    private static final Parameter MAX_P_MISMATCH_CONVERGED_PARAMETER = new Parameter(
+            MAX_P_MISMATCH_CONVERGED,
+            ParameterType.DOUBLE,
+            "Max mismatch in active power to consider a bus converged when exporting load flow status of topological islands",
+            CgmesExportContext.MAX_P_MISMATCH_CONVERGED_DEFAULT_VALUE);
+    private static final Parameter MAX_Q_MISMATCH_CONVERGED_PARAMETER = new Parameter(
+            MAX_Q_MISMATCH_CONVERGED,
+            ParameterType.DOUBLE,
+            "Max mismatch in reactive power to consider a bus converged when exporting load flow status of topological islands",
+            CgmesExportContext.MAX_P_MISMATCH_CONVERGED_DEFAULT_VALUE);
```
marqueslanauja, line 327: Must be  MAX_Q_MISMATCH_ONVERGED_DEFAULT_VALUE

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
```diff
@@ -706,6 +714,33 @@ public CgmesExportContext setExportTransformersWithHighestVoltageAtEnd1(boolean
...
+
+    public double getMaxQMismatchConverged() {
+        return maxQMismatchConverged;
+    }
+
+    public CgmesExportContext setMaxQMismatchConverged(double maxPMismatchConverged) {
+        this.maxQMismatchConverged = maxPMismatchConverged;
+        return this;
+    }
+
```
marqueslanauja, line 743: Better maxQMismatchConverded

marqueslanauja (approved):
Some minor changes to do

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
zamarrenolm: fixed

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
```diff
@@ -125,19 +131,115 @@ private static void buildAngleRefs(Bus bus, Map<String, String> angleRefs, Cgmes
...
+            return Double.isFinite(a);
+        }
+
+        boolean isInAccordanceWithKirchoffsFirstLaw(Bus bus) {
+            // Instead of checking switch flows, we check that the corresponding bus view bus is balanced
+            // Maybe we will check the same bus view bus more than once
+            Optional<Bus> bus0 = getBusViewBus(bus);
+            if (bus0.isEmpty()) {
+                LOG.warn("Can not check if bus is in accordance with Kirchoff's first law. No BusView bus can be found for: {}", bus);
+                return false;
```
annetill, line 196: So false? It can happen indeed?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
annetill: Ok so I disagree. A `NaN` terminal should be filtered and excluded from the sum. And I think you have to separate the P treatment that the Q treatment. Because we can have a NaN for P and not for Q. If something has be sum up because filtered, so we can say that it is a divergence. 
We have to be coherent, a NaN at terminal is exported as zero value.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
```diff
@@ -125,19 +131,115 @@ private static void buildAngleRefs(Bus bus, Map<String, String> angleRefs, Cgmes
...
+
+        Optional<Bus> getBusViewBus(Bus bus) {
+            if (bus.getVoltageLevel().getTopologyKind().equals(TopologyKind.BUS_BREAKER)) {
+                return Optional.of(bus.getVoltageLevel().getBusView().getMergedBus(bus.getId()));
+            } else {
+                return bus.getConnectedTerminalStream().map(t -> t.getBusView().getBus()).filter(Objects::nonNull).findFirst();
+            }
+        }
+
+        boolean isValidTerminal(Terminal t) {
```
annetill, line 221: We can leave that...

annetill: Excepted for bus bar section. If we have an issue with static var compensator, the issue is in the power flow engine. Or in a more clever treatment of NaN values.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
zamarrenolm: I've extended the look-up for bus view bus from bus/breaker bus. If we still are unable  to find a bus view bus, that topological island should not be considered as converged.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
zamarrenolm: separate p and q sums, changed by sum over non-NaN values and check not all terminals have NaN value.

jeandemanged (approved):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
```diff
@@ -125,19 +131,161 @@ private static void buildAngleRefs(Bus bus, Map<String, String> angleRefs, Cgmes
...
+            if (optionalBusViewBus.isEmpty()) {
+                LOG.error("Can not check if bus is in accordance with Kirchhoff's first law. No BusView bus can be found for: {}", bus);
+                return false;
+            }
+            Bus busViewBus = optionalBusViewBus.get();
+            // We do not check the same bus view bus more than once
+            if (checkedBusViewBuses.contains(busViewBus)
+                    || busViewBus.getConnectedTerminalCount() == 0
+                    || isSlack(busViewBus)
+                    || isAngleReference(busViewBus)) {
```
annetill, line 205: Why do you have to filter the angle reference bus?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
```diff
@@ -125,19 +131,161 @@ private static void buildAngleRefs(Bus bus, Map<String, String> angleRefs, Cgmes
...
+                bus.getConnectedTerminalStream()
+                        .forEach(t -> LOG.info(String.format("  %7.2f  %7.2f  %s %s %s", t.getP(), t.getQ(), t.getConnectable().getType(), t.getConnectable().getNameOrId(), t.getConnectable().getId())));
+            }
+        }
+
+        Optional<Bus> getBusViewBus(Bus bus) {
+            if (bus.getVoltageLevel().getTopologyKind().equals(TopologyKind.BUS_BREAKER)) {
+                return Optional.of(bus.getVoltageLevel().getBusView().getMergedBus(bus.getId()));
+            } else {
+                if (bus.getConnectedTerminalCount() > 0) {
```
annetill, line 265: Mmmm a bus without any connected terminal is not a bus in iidm topology computation, no?

zamarrenolm: At node/breaker level, the computed bus/breaker buses may not have terminals because the lie in nodes connected through retained switches. The following diagram shows an example (taken from a real-world network, the model is defined as node/breaker):

Screenshot 2023-11-28 at 16 13 27

The nodes 0, 15, 20 are grouped in a calculated bus/breaker bus that do not have terminals (only ends of retained switches); the same for 4, 16, 24. The color of the nodes is the indication of the scope of the bus/breaker bus. To locate the corresponding `BusView` bus we can not go through terminals, we find it checking which bus view bus from the voltage level contains the bus/breaker view bus.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
zamarrenolm: Right. Fixed. Buses for angle reference terminals must also be solved, they should not have a mismatch.

annetill (approved):
So nice indeed, thanks a lot!

marqueslanauja (approved):
Reviewed and checked

PR #2784, opened 2023-11-20
Renaming XML classes/artifactId/module/package following JSON support

flo-dup:

No

"Bug" fix

The serializing classes/artifactId/module/package are still named with the `xml` keyword

The `xml` keyword in serializing classes/artifactId/module/package is replaced with `serde` or `SerDe` keyword, or removed if not needed (e.g. `IidmVersion` instead of `IidmXmlVersion`)

**Does this PR introduce a breaking change or deprecate an API?** No


flo-dup:
We should also rename the properties for importing/exporting with backward compatibilities

PR #2787, opened 2023-11-21
Factorizing breadth/depth-first traversals

flo-dup:

No

Refactoring/Factorizing code between breadth-first and depth-first traversals.

One iterative traversal, one recursive traversal with very similar code

The common code is factorized and both traversals are recursive

No


math/src/main/java/com/powsybl/math/graph/UndirectedGraphImpl.java
```diff
@@ -543,14 +467,32 @@ public boolean traverse(int v, TraversalType traversalType, Traverser traverser,
...
+                case DEPTH_FIRST -> edgesToTraverse.pollLast();
+                case BREADTH_FIRST -> edgesToTraverse.pollFirst();
+            };
+            Edge<E> edge = edges.get(e);
+            if (encountered[edge.getV1()] && encountered[edge.getV2()]) {
+                continue; // traversing parallel edges only once
+            }
 
+            int vOrigin = encountered[edge.getV1()] ? edge.getV1() : edge.getV2();
+            int vDest = encountered[edge.getV1()] ? edge.getV2() : edge.getV1();
```
rolnico, line 486: Since it's the same condition, it might be easier to read (and it reduces the complexity) if you used an if statement, even if that means a bit more lines of code ?

rolnico (changes requested):

geofjamg (approved):

rolnico (approved):

PR #2788, opened 2023-11-22
Dynamic simulation results

### LINKED ISSUE #2772: DynamicSimulationResult

_No response_

Remove log field  -> use DynamicSimulationProvider reporter instead
Replace status field -> use Status enum with description of the error copied from dynawo error report
Replace StringTimeSeries field with a adequate TimeSeries 

### Describe the motivation

Match Dynawaltz implementation

### Extra Information

_No response_

Lisrte:

Fixes #2772

Feature


dynamic-simulation/dynamic-simulation-api/src/main/java/com/powsybl/dynamicsimulation/DynamicSimulationResult.java
```diff
@@ -6,35 +6,53 @@
...
+
 /**
  * @author Marcos de Miguel {@literal <demiguelm at aia.es>}
+ * @author Laurent Issertial {@literal <laurent.issertial at rte-france.com>}
  */
 public interface DynamicSimulationResult {
 
-    boolean isOk();
+    enum Status {
+        SUCCEED,
```
flo-dup, line 24: Both value should have the same grammatically class:  `SUCCEEDED` and `FAILED`, or `SUCCEED` and `FAIL`, or `SUCCESS` and `FAILURE`. As we chose success and failure in the other APIs, we should stick with that.

dynamic-simulation/dynamic-simulation-api/src/main/java/com/powsybl/dynamicsimulation/DynamicSimulationResultImpl.java
```diff
@@ -7,37 +7,47 @@
...
 /**
  * @author Marcos de Miguel {@literal <demiguelm at aia.es>}
+ * @author Laurent Issertial {@literal <laurent.issertial at rte-france.com>}
  */
 public class DynamicSimulationResultImpl implements DynamicSimulationResult {
 
-    private final boolean status;
-    private final String logs;
+    private final Status status;
+    private final String error;
```
flo-dup, line 23: Even if we're using it only for error on powsybl-dynawo side, we could keep the more general name of statusText if later we change that practice (or if somebody else uses this implementation).

dynamic-simulation/dynamic-simulation-api/src/main/java/com/powsybl/dynamicsimulation/DynamicSimulationResultImpl.java
```diff
@@ -7,37 +7,47 @@
...
     private final Map<String, TimeSeries> curves;
-    private final StringTimeSeries timeLine;
+    private final List<TimelineEvent> timeLine;
 
-    public DynamicSimulationResultImpl(boolean status, String logs, Map<String, TimeSeries> curves, StringTimeSeries timeLine) {
-        this.status = status;
-        this.logs = logs;
+    public DynamicSimulationResultImpl(Status status, String error, Map<String, TimeSeries> curves, List<TimelineEvent> timeLine) {
+        this.status = Objects.requireNonNull(status);
+        this.error = error;
```
flo-dup, line 29: I think we could also require that input to be non-null.

We could also replace the null input with the name of the input, as it's done on `LoadFlowResultImpl`: `Objects.requireNonNullElse(statusText, status.name());`, but I'm not so much convinced by adding this useless info, are you?

flo-dup (changes requested):
Just very minor comments

dynamic-simulation/dynamic-simulation-api/src/main/java/com/powsybl/dynamicsimulation/DynamicSimulationResultImpl.java
```diff
@@ -7,37 +7,47 @@
...
         this.curves = Objects.requireNonNull(curves);
         this.timeLine = Objects.requireNonNull(timeLine);
+        timeLine.forEach(Objects::requireNonNull);
+    }
+
+    public DynamicSimulationResultImpl(Status status, Map<String, TimeSeries> curves, List<TimelineEvent> timeLine) {
+        this(status, "", curves, timeLine);
+    }
+
+    public static DynamicSimulationResultImpl createSucceededResult(Map<String, TimeSeries> curves, List<TimelineEvent> timeLine) {
```
flo-dup, line 39: Now that succeed has been renamed to success, we should probably rename this into `createSuccessResult`.

dynamic-simulation/dynamic-simulation-api/src/main/java/com/powsybl/dynamicsimulation/DynamicSimulationResultImpl.java
```diff
@@ -7,37 +7,47 @@
...
  * @author Marcos de Miguel {@literal <demiguelm at aia.es>}
+ * @author Laurent Issertial {@literal <laurent.issertial at rte-france.com>}
  */
 public class DynamicSimulationResultImpl implements DynamicSimulationResult {
 
-    private final boolean status;
-    private final String logs;
+    private final Status status;
+    private final String statusText;
     private final Map<String, TimeSeries> curves;
```
flo-dup, line 24: Just realizing by seeing the code smells that we didn't change anything for curves: shouldn't we take this opportunity to ask for `Map<String, DoubleTimeSeries>`? I don't think any other TimeSeries would make sense for a curve.

flo-dup (changes requested):

flo-dup (approved):

PR #2789, opened 2023-11-22
Loadflow specific parameters to map

geofjamg:

No

Feature

A method in `LoadFlowProvider` has been added to converter specific parameters to a map to be able to get specific parameters values in a non typed way.
I also removed deprecated methods and default methods from `LoadFlowProvider`.

**Other information**:


loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/AbstractLoadFlowProvider.java
flo-dup, line 1: This should be used only for loadflow providers which don't have specific parameters, so shouldn't we put that in the name? Couldn't find anything better than `AbstractNoSpecificParametersLoadFlowProvider` though, or `AbstractBasicLoadFlowProvider` :expressionless:

flo-dup (approved):

PR #2792, opened 2023-11-24
Serialization/Deserialization fixes

flo-dup:

No

Bug fix

- JSON deserialization fails when CGMESControlArea contains a boundary with a danglingLine and a side (side was not read as expected / not useful)
- some methods used by other repositories were removed from XmlUtil
- IidmSerDeConstants.CURRENT_IIDM_XML_VERSION
- ConnectableSerDeUtil.::writeCurrentLimits does not accept null currentLimits
- No TreeDataWriter::writeStringArray
- ReferencePriorities/ReferenceTerminals SerDe does not support JSON as it does not implement getArrayNameToSingleNameMap

- CGMESControlArea JSON deserialization OK
- The removed XmlUtil methods are back
- IidmSerDeConstants.CURRENT_IIDM_VERSION
- ConnectableSerDeUtil.::writeCurrentLimits accepts null currentLimits
- TreeDataWriter::writeStringArray added
- ReferencePriorities/ReferenceTerminals SerDe supports JSON

No


iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -488,10 +489,22 @@ public static Anonymizer writeAndValidate(Network n, Path xmlFile) {
         return writeAndValidate(n, new ExportOptions(), xmlFile);
     }
 
+    /**
+     * Writes given network to JSON file, then reads the resulting file, then write the resulting network
+     * to the given path in XML format and validates the result with iidm .
+     */
     public static Anonymizer writeAndValidate(Network n, ExportOptions options, Path xmlFile) {
```
geofjamg, line 496: This testing method should not be here as writeAndValidate and not intended to be only used for unit testing. It could be used by anyone willing to load and apply xml validation. So when used with Json format, we should just print a warning to inform that validation is not supported for json.
The roundtrip xml + json code should be in testing code (so in AbstractSerderTest)

geofjamg (changes requested):

geofjamg (approved):

PR #2794, opened 2023-11-24
Fix CGMES date parsing

flo-dup:

No

Bug fix

In some CGMES files, the fractions of second may be written with more than 6 digits, for instance:
`2023-07-31T00:30:00.0000000Z`
Such a date leads to a DateTimeParseException, and then the caseDate is overwritten with ZonedDateTime.now()

Increasing to maximum the max width (9).

- [ ] Yes
- [x] No

### DIFF
--- a/cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/DateParsingTest.java
+++ b/cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/DateParsingTest.java
@@ -26,14 +26,13 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 class DateParsingTest {
 
     @Test
-
     void test() {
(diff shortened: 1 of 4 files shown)


flo-dup:
Could it be more than 7 digits? I didn't find the specification about the date format in CGMES

flo-dup:
Any kind of precision should be allowed according to the documentation (thanks @colinepiloquet!), hence I raised the width to the maximum value (9).
> Date and time as "yyyy-mm-ddThh:mm:ss.sss", which conforms with ISO 8601. UTC time zone is
specified as "yyyy-mm-ddThh:mm:ss.sssZ". A local timezone relative UTC is specified as "yyyy -
mm-ddThh:mm:ss.sss-hh:mm". The second component (shown here as "ss.sss") could have any
number of digits in its fractional part to allow any kind of precision beyond seconds.

olperr1 (approved):

PR #2814, opened 2023-12-01
Matpower export: fix low impedance branches and branches connected to same bus

geofjamg:
… same bus

No

Bug fix.

It seems that Matpower LF does not like:
 - branches connected to same bus
 - branches with a low impedance

 - we skip branches connected to same bus
 - we cut too low impedances

- [ ] Yes
- [ ] No

**Other information**:


matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerExporter.java
```diff
@@ -388,39 +389,44 @@ private void createTransformers2(Network network, MatpowerModel model, Context c
...
-                    b *= 1 + rtc.getCurrentStep().getB() / 100;
-                }
-                var ptc = twt.getPhaseTapChanger();
-                if (ptc != null) {
-                    mBranch.setPhaseShiftAngle(-ptc.getCurrentStep().getAlpha());
-                    rho *= ptc.getCurrentStep().getRho();
-                    r *= 1 + ptc.getCurrentStep().getR() / 100;
-                    x *= 1 + ptc.getCurrentStep().getX() / 100;
-                    b *= 1 + ptc.getCurrentStep().getB() / 100;
+                if (!bus1.getId().equals(bus2.getId())) {
```
annetill, line 392: Do you really think that it can happen with transformers?

geofjamg: I never had the case indeed

annetill (approved):

PR #2820, opened 2023-12-05
Add new type of NodeCalc to compute Min/Max between timeseries

rolnico:

No

Feature

Using NodeCalc, it is only possible to get `timeseries["foo"].min(comparedValue)` the minimum between `comparedValue` and the values of the timeseries.

It is now possible to get the minimal value(s) between two timeseries using `min(timeseries["foo"], timeseries["bar"])` in a groovy script

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/NodeCalcVisitor.java
+++ b/time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/NodeCalcVisitor.java
@@ -56,4 +56,8 @@ public interface NodeCalcVisitor<R, A> {
     R visit(BinaryMinCalc nodeCalc, A arg, R left, R right);
 
     Pair<NodeCalc, NodeCalc> iterate(BinaryMinCalc nodeCalc, A arg);
+
+    R visit(BinaryMaxCalc nodeCalc, A arg, R left, R right);
+
+    Pair<NodeCalc, NodeCalc> iterate(BinaryMaxCalc nodeCalc, A arg);
 }
--- a/time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/NodeCalcVisitor.java
+++ b/time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/NodeCalcVisitor.java
@@ -52,4 +52,8 @@ public interface NodeCalcVisitor<R, A> {
     R visit(TimeSeriesNameNodeCalc nodeCalc, A arg);
 
     R visit(TimeSeriesNumNodeCalc nodeCalc, A arg);
+
+    R visit(BinaryMinCalc nodeCalc, A arg, R left, R right);
+
+    Pair<NodeCalc, NodeCalc> iterate(BinaryMinCalc nodeCalc, A arg);
 }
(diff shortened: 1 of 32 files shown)


time-series/time-series-api/src/test/java/com/powsybl/timeseries/CalculatedTimeSeriesTest.java
```diff
@@ -84,21 +86,33 @@ public Set<Integer> getTimeSeriesDataVersions(String timeSeriesName) {
...
         // check versions of the data available for the calculated time series
         tsCalc.setTimeSeriesNameResolver(resolver);
+        tsCalcMin.setTimeSeriesNameResolver(resolver);
+        tsCalcMax.setTimeSeriesNameResolver(resolver);
         assertEquals(Sets.newHashSet(1), tsCalc.getVersions());
 
-        List<DoubleTimeSeries> tsLs = Arrays.asList(ts, tsCalc);
+        List<DoubleTimeSeries> tsLs = Arrays.asList(ts, tsCalc, tsCalcMin, tsCalcMax);
         String json = TimeSeries.toJson(tsLs);
         String jsonRef = String.join(System.lineSeparator(),
```
geofjamg, line 117: Now that we have text blocks in Java 17, we can do it much simpler and cleaner.
Maybe should be done in another PR.

time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/NodeCalcVisitor.java
```diff
@@ -52,4 +48,8 @@ public interface NodeCalcVisitor<R, A> {
     R visit(TimeSeriesNameNodeCalc nodeCalc, A arg);
 
     R visit(TimeSeriesNumNodeCalc nodeCalc, A arg);
+
+    R visit(AbstractBinaryNodeCal nodeCalc, A arg, R left, R right);
```
geofjamg, line 52: You should continue with same design that previous by having in the visitor only child node and not abstracts. As you can see, as doing that you need to lot of instanceof check in all implementations.

time-series/time-series-dsl/src/main/groovy/com/powsybl/timeseries/dsl/NodeCalcGroovyExtensionModule.groovy
```diff
@@ -8,6 +8,7 @@ package com.powsybl.timeseries.dsl
 
 import com.powsybl.commons.PowsyblException
 import com.powsybl.timeseries.ast.BigDecimalNodeCalc
+import com.powsybl.timeseries.ast.BinaryMinCalc
```
geofjamg, line 11: unused import?

geofjamg (changes requested):

geofjamg (approved):

time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/NodeCalcVisitor.java
rolnico: done

time-series/time-series-dsl/src/main/groovy/com/powsybl/timeseries/dsl/NodeCalcGroovyExtensionModule.groovy
rolnico: done

PR #2827, opened 2023-12-06
CGMES import: fix loading limits conversion

annetill:

- [ ] Yes
- [ ] No

**Other information**:


murgeyseb:
It seems to fix our issue ! I let @WalAmeni check it also, and it should be ok us.

zamarrenolm:
The fix solves the problem of the wrong assignment of an operational limit to an equipment, but some operational limits may be lost during conversion.

A way to keep loading limits defined over terminals of switches is to adapt/reuse a similar code developed for keeping regulating control terminals in `RegulatingTerminalMapper::mapForFlowControl`. We propose to merge this fix and prepare an additional PR with this alternative.

zamarrenolm (approved):

PR #2831, opened 2023-12-12
Network modification connect and disconnect

rolnico:

No

feature

When connecting/disconnecting a terminal, you get a confirmation boolean after the modification is done, which means that if you try to disconnect a connectable with 2 terminals, you will disconnect the first then the second. The problem is that if the 2nd disconnection fails, the first is still done so you have a connectable half-way disconnected

By using the connect/disconnect methods from the connectable class, you will connect/disconnect the different terminals if and only if all the terminals can be connected/disconnected.
If some terminals were already connected/disconnected, it will still work and the other terminals will be connected/disconnected.
If all the terminals were already connected/disconnected, it will fail and return false

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractConnectable.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractConnectable.java
@@ -7,16 +7,15 @@
 package com.powsybl.iidm.network.impl;
 
 import com.powsybl.commons.PowsyblException;
-import com.powsybl.iidm.network.Bus;
-import com.powsybl.iidm.network.Connectable;
-import com.powsybl.iidm.network.TopologyKind;
-import com.powsybl.iidm.network.TopologyPoint;
+import com.powsybl.iidm.network.*;
 import com.powsybl.iidm.network.impl.util.Ref;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Objects;
+import java.util.*;
+import java.util.function.Predicate;
 import java.util.function.Supplier;
+
+import static com.powsybl.iidm.network.TopologyKind.BUS_BREAKER;
+import static com.powsybl.iidm.network.TopologyKind.NODE_BREAKER;
 
 /**
  *
@@ -137,7 +136,7 @@ abstract class AbstractConnectable<I extends Connectable<I>> extends AbstractIde
         }
 
         // check bus topology
-        if (voltageLevel.getTopologyKind() != TopologyKind.NODE_BREAKER) {
+        if (voltageLevel.getTopologyKind() != NODE_BREAKER) {
             String msg = String.format(
                     "Trying to move connectable %s to node %d of voltage level %s, which is a bus breaker voltage level",
                     getId(), node, voltageLevel.getId());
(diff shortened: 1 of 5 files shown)


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevel.java
```diff
@@ -1197,22 +1202,38 @@ public boolean connect(TerminalExt terminal) {
...
+        // Check the topology kind
+        checkTopologyKind(terminal);
+
         // already connected?
         if (terminal.isConnected()) {
             return false;
         }
 
+        // Initialisation of a list to open in case some terminals are in node-breaker view
+        Set<SwitchImpl> switchForDisconnection = new HashSet<>();
```
flo-dup, line 1214:
```suggestion
        Set<SwitchImpl> switchForConnection = new HashSet<>();
```

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/ConnectableConnection.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.function.Predicate;
+
+/**
+ * @author Nicolas Rol {@literal <nicolas.rol at rte-france.com>}
+ */
+public class ConnectableConnection extends AbstractNetworkModification {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ConnectableConnection.class);
+    final String connectableId;
+    Predicate<Switch> isTypeSwitchToOperate = SwitchPredicates.IS_NON_NULL;
```
flo-dup, line 30: It is only set in the constructor so we can make it final.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/AbstractDisconnection.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.util.Objects;
+import java.util.function.Predicate;
+
+/**
+ * @author Nicolas Rol {@literal <nicolas.rol at rte-france.com>}
+ */
+public abstract class AbstractDisconnection extends AbstractNetworkModification {
+    final String connectableId;
+    Predicate<Switch> openableSwitches;
```
flo-dup, line 21: can you put this final and get it from the constructor arguments?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/UnplannedDisconnection.java
```diff
@@ -0,0 +1,51 @@
...
+    public void apply(Network network, NamingStrategy namingStrategy, boolean throwException,
+                      ComputationManager computationManager, Reporter reporter) {
+        // Get the connectable
+        Connectable<?> connectable = network.getConnectable(connectableId);
+
+        // Disconnect the connectable
+        boolean hasBeenDisconnected = connectable.disconnect(openableSwitches, reporter);
+
+        if (hasBeenDisconnected) {
+            LOG.info("Connectable {} has been disconnected.", connectableId);
```
flo-dup, line 46: We usually put reports also in the network modifications

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractConnectable.java
```diff
@@ -190,4 +193,135 @@ private void attachTerminal(TerminalExt oldTerminal, TopologyPoint oldTopologyPo
...
+    }
+
+    public boolean disconnect(Reporter reporter) {
+        return disconnect(SwitchPredicates.IS_CLOSED_BREAKER, reporter);
+    }
+
+    public boolean disconnect(Predicate<Switch> isSwitchOpenable, Reporter reporter) {
+        // Booleans
+        boolean isAlreadyDisconnected = true;
+        boolean isNowDisconnected = true;
```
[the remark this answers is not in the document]
rolnico: I merged two if (and inverted one) in order to have this boolean only used for one goal

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractConnectable.java
```diff
@@ -190,4 +193,135 @@ private void attachTerminal(TerminalExt oldTerminal, TopologyPoint oldTopologyPo
...
+            // Exit if the terminal cannot be disconnected
+            if (!isNowDisconnected) {
+                return false;
+            }
+        }
+
+        // Exit if the connectable is already fully disconnected
+        if (isAlreadyDisconnected) {
+            return false;
+        }
```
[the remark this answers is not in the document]
flo-dup: After some discussion, this breaking change should not be done here.
But could we put in the javadoc that returning false means no changes?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/util/ModificationReports.java
```diff
@@ -645,4 +645,31 @@ public static void scalingReport(Reporter reporter, String type, ScalingType sca
...
+            .withValue("connectable", connectable.getId())
+            .withSeverity(TypedValue.INFO_SEVERITY)
+            .build());
+    }
+
+    public static void connectableDisconnectionReport(Reporter reporter, Connectable<?> connectable, boolean disconnectionSuccessful, boolean isPlanned) {
+        String defaultMessage = disconnectionSuccessful ?
+            "Connectable ${connectable} has been disconnected." :
+            "Connectable ${connectable} has NOT been disconnected.";
+        String key = isPlanned ? "planned" : "unplanned";
```
flo-dup, line 666: isPlanned doesn't change the message so either add it to the message or remove it from the key

flo-dup (changes requested):

flo-dup (approved):

PR #2836, opened 2023-12-15
Cgmes: Fix GeographicalRegion creation if field is absent

colinepiloquet:

No

Bug fix

Regions are not added if they are retrieved from reference data

Regions are created

- [ ] Yes
- [X] No


colinepiloquet:
@zamarrenolm, I do not know if this is what should be done but it works

colinepiloquet:
It's ok, we don't export GeographicalRegions because there is a reference to Boundary set.

PR #2837, opened 2023-12-18
New "Ground" injection

rolnico:

https://github.com/powsybl/powsybl-core/issues/1131
But, in fact, ground disconnectors are not ignored but classified as disconnectors.

See for information first try in #2757 

Bug fix / feature

No distinction is made between ground disconnectors and disconnectors in IIDM and no ground element exists

A new Ground identifiable is created and can be added next to a disconnector to modelise a "ground disconnector"

- [ ] Yes
- [ ] No

**Other information**:


iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractGroundTest.java
```diff
@@ -0,0 +1,398 @@
...
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * @author Nicolas Rol {@literal <nicolas.rol at rte-france.com>}
+ */
+public abstract class AbstractGroundTest {
+    private Network network;
+
+    @BeforeEach
+    public void initNetwork() {
```
So-Fras, line 31: It looks like this network is a duplicate of the network created in `GroundSerDeTest`, at least if we omit the ground and the associated disconnector parts. If it is indeed the case, could it be possible to factorize this code in the iidm-test package?

So-Fras (changes requested):
Thnak you for this PR.
I left 2 minor remarks.
In order to make the examples in tests more realistic, the disconnectors (or the bus in bus-breaker view) associated with the grounds could be linked with a line.

So-Fras:
I am thinking about something: should we allow the ground creation in bus/breaker view? Will the associated disconnector be present on a bus/breaker network?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerVoltageLevel.java
```diff
@@ -464,29 +464,13 @@ public boolean isValid(UndirectedGraph<? extends TerminalExt, SwitchImpl> graph,
...
-                            break;
-
-                        case BUSBAR_SECTION:
-                            busbarSectionCount++;
-                            break;
-
-                        default:
-                            throw new IllegalStateException();
+                        }
+                        case LOAD, GENERATOR, BATTERY, SHUNT_COMPENSATOR, STATIC_VAR_COMPENSATOR, GROUND -> feederCount++;
```
flo-dup, line 471: adding the ground here means that the following cases are considered as a valid bus:
- a busbar with a ground
- a line and two grounds

I think we don't want that

flo-dup (changes requested):

flo-dup (approved):

So-Fras (approved):

flo-dup (changes requested):
Just realized we forgot to put it in TopologyVisitor!

flo-dup (approved):

PR #2848, opened 2024-01-11
CGMES ground conversion

zamarrenolm:

Feature

CGMES Ground objects are ignored.

CGMES Ground objects are imported into IIDM.

**Other information**:

It is based on #2837


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -181,6 +181,7 @@ public Network convert(Reporter reporter) {
             convert(cgmes.busBarSections(), bbs -> new BusbarSectionConversion(bbs, context));
         }
 
+        convert(cgmes.grounds(), g -> new GroundConversion(g, context));
```
annetill, line 184: After a talk with @flo-dup, this feature can add many buses in the bus/breaker view coming from the node/breaker view. To avoid performance issue, I think we have to test: the difference of bus number with an without the feature on a real network. If the difference is huge, a parameter must be added, to false by default.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/GroundConversionTest.java
```diff
@@ -0,0 +1,29 @@
...
+ * @author Sophie Frasnedo {@literal <sophie.frasnedo at rte-france.com>}
+ */
+
+class GroundConversionTest {
+
+    @Test
+    void groundConversionTest() {
+        Network network = Network.read("groundTest.xml", getClass().getResourceAsStream("/groundTest.xml"));
+        long groundCount = network.getVoltageLevelStream().map(VoltageLevel::getGrounds).count();
+        assertEquals(2, groundCount);
```
olperr1: You can call `network.getGroundCount()` to simplify the test.

Also, the followings can also be tested:
- the id and names of the grounds are imported (`GroundAdderImpl` should be modified to use the names);
- the grounds are attached to the right voltage levels ("CV" and "OU" are both attached to voltage level "S");
- the grounds' terminals are not null and not connected (both ground disconnectors are open).

olperr1 (changes requested):

annetill (approved):
@zamarrenolm is going to reduce a little bit the network. If not ready today, I agree to merge like this.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/GroundConversionTest.java
annetill: Ok, finally @zamarrenolm do not reduce the file, it is okay like this.

olperr1 (approved):

PR #2852, opened 2024-01-15
Secondary voltage control notifications

geofjamg:

No

Feature

When a pilot point target voltage or a control unit status is modified, network listeners are not notified.

We added an extension notification feature and secondary voltage control has been fully redesigned to support notifications.

- [ ] Yes
- [ ] No

**Other information**:


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/NetworkListener.java
```diff
@@ -6,33 +6,40 @@
...
 
     void afterRemoval(String id);
 
-    void onUpdate(Identifiable identifiable, String attribute, Object oldValue, Object newValue);
+    void onUpdate(Identifiable<?> identifiable, String attribute, Object oldValue, Object newValue);
 
-    default void onUpdate(Identifiable identifiable, String attribute, String variantId, Object oldValue,
-            Object newValue) {
-        // empty default implementation
-    }
```
olperr1, line 25: Why did you remove this default implementation but keep the ones for `onElementAdded`, `onElementReplaced` and `onElementRemoved`?

geofjamg: We could also. I will open a PR to also rename them as element is not a right term (this is in really only properties add or remove)

olperr1 (approved):

PR #2854, opened 2024-01-15
Connection disconnection notification

geofjamg:

No

Feature

connection, disconnection notification are mixed with more elementary switch opening or terminal connection status notification so that it is hard to handle.

We have one notification at connection/disconnection beginning and another one at ending. 

- [X] Yes
- [ ] No

When using the default IIDM implementation, at terminal connection/disconnection:
- the `"connection"` notifications that were emitted are replaced:
  - at connection, by `"beginConnect"` and `"endConnect"` notifications;
  - at disconnection, by `"beginDisconnect"` and `"endDisconnect"` notifications;
- the `"connected"` notifications are replaced by `"connected" + side number` notifications.

If you have defined your own IIDM implementation, you should implement the following methods:
 - in your `Terminal` implementations:
   - `ThreeSides getSide()`

**Other information**:


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTerminal.java
```diff
@@ -172,12 +177,12 @@ public boolean connect(Predicate<Switch> isTypeSwitchToOperate) {
...
+        String variantId = getVariantManagerHolder().getVariantManager().getVariantId(variantIndex);
+        boolean connectedBefore = isConnected();
+        connectable.notifyUpdate("beginConnect", variantId, connectedBefore, null);
         boolean connected = voltageLevel.connect(this, isTypeSwitchToOperate);
-        if (connected) {
-            int variantIndex = getVariantManagerHolder().getVariantIndex();
-            String variantId = getVariantManagerHolder().getVariantManager().getVariantId(variantIndex);
-            connectable.notifyUpdate("connection", variantId, false, true);
-        }
+        connectable.notifyUpdate("endConnect", variantId, !connected, connected);
```
olperr1, line 185: When the terminal was already connected, the `voltageLevel.connect(...)` method returns `false`.
Thus, the "endConnect" notification has the following values:
- `oldValue`: `true`;
- `newValue`: `false`.

Since `connect(...)` throws an `Exception` when a problem occurs, we may always use `true` for `newValue`.
It isn't a real update when the terminal is already connected, but I think it may be more important to have a notification when the process is finished:
```suggestion
        connectable.notifyUpdate("endConnect", variantId, connectedBefore, true);
```
<br/>

Another solution would be to use `null` as `newValue` when `connected` is `false`:
```suggestion
        connectable.notifyUpdate("endConnect", variantId, connectedBefore, connected ? true : null);
```

(To improve readability, I think it is better to use `connectedBefore` as `oldValue` parameter.)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTerminal.java
```diff
@@ -198,12 +203,12 @@ public boolean disconnect(Predicate<Switch> isSwitchOpenable) {
...
+        String variantId = getVariantManagerHolder().getVariantManager().getVariantId(variantIndex);
+        boolean disconnectedBefore = !isConnected();
+        connectable.notifyUpdate("beginDisconnect", variantId, disconnectedBefore, null);
         boolean disconnected = voltageLevel.disconnect(this, isSwitchOpenable);
-        if (disconnected) {
-            int variantIndex = getVariantManagerHolder().getVariantIndex();
-            String variantId = getVariantManagerHolder().getVariantManager().getVariantId(variantIndex);
-            connectable.notifyUpdate("connection", variantId, true, false);
-        }
+        connectable.notifyUpdate("endDisconnect", variantId, !disconnected, disconnected);
```
olperr1, line 211: Same comment as for `"endConnect"` notification.

We could either always use `true` as `newValue`:
```suggestion
        connectable.notifyUpdate("endDisconnect", variantId, disconnectedBefore, true);
```

or use `null` when the terminal was already disconnected:
```suggestion
        connectable.notifyUpdate("endDisconnect", variantId, disconnectedBefore, disconnected ? true : null);
```

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractConnectionNotificationTest.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
+ */
+public abstract class AbstractConnectionNotificationTest {
```
olperr1, line 25: Could you also add tests connecting an already connected terminal and disconnecting an already disconnected terminal?

olperr1 (changes requested):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTerminal.java
olperr1: Ok, you are right. I was looking  at the Bus/Breaker implementation and not at the Node/Breaker one (which I should have done). It returns `false` if nothing was performed.
Maybe we can use the following values, then:
- `oldValue`: `connectedBefore`;
- `newValue`: `connectedBefore || connected` (true if it was already connected or if the connection succeeded)

olperr1: Ok, I just saw your change. This new policy is ok for me.

geofjamg: I used the oldValue for the beginConnection and newValue for the endConnection and null for the other mainly because there is a filter of notif when oldValue = newValue. So I'm not really happy of that but maybe this is enough at the moment?

PR #2862, opened 2024-01-19
Optimisation of TimeSeries computing

rolnico:

No

Feature

Every node in the tree forming the timeseries is computed, even if it has already been computed before (in the case where a node appears multiple times in the tree).

Each node is computed only once.

- [ ] Yes
- [ ] No

**Other information**:


rolnico:
Using an abstract class seems to make the code slower, we might want to stick with the code duplicated in the different classes

time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/BinaryMaxCalc.java
```diff
@@ -52,6 +52,6 @@ public boolean equals(Object obj) {
 
     @Override
     public int hashCode() {
-        return left.hashCode() + right.hashCode();
+        return hash;
```
olperr1, line 55: There is a problem here, and in the other classes extending `AbstractBinaryNodeCalc`.

`hash` is only computed in the constructors of these classes, but `left` and `right` are not final, so they can change.
If `setLeft(...)` or `setRight(...)` are called, the hash doesn't match anymore with the content.

This leads to a major problem: `hashCode()` doesn't respect its general contract:

> If two objects are equal according to the [equals](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Object.html#equals(java.lang.Object)) method, then calling the hashCode method on each of the two objects must produce the same integer result.

([source](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Object.html#hashCode()))

For instance, the following test fails on the last assertion:

```java
    @Test
    void hashTest() {
        NodeCalc node1 = new IntegerNodeCalc(1);
        NodeCalc node2 = new IntegerNodeCalc(2);
        NodeCalc node3 = new IntegerNodeCalc(3);

        AbstractBinaryNodeCalc binary1 = new BinaryMaxCalc(node1, node3);
        binary1.setRight(node2);
        NodeCalc binary2 = new BinaryMaxCalc(node1, node2);

        assertEquals(binary2, binary1);
        assertEquals(binary2.hashCode(), binary1.hashCode());
    }
```

time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/UnaryOperation.java
```diff
@@ -56,26 +58,41 @@ public static UnaryOperation positive(NodeCalc child) {
...
     @Override
-    public <R, A> R accept(NodeCalcVisitor<R, A> visitor, A arg, int depth) {
+    public <R, A> R accept(NodeCalcVisitor<R, A> visitor, A arg, int depth, Map<Integer, R> cache) {
         if (depth < NodeCalcVisitors.RECURSION_THRESHOLD) {
-            NodeCalc child = visitor.iterate(this, arg);
-            R childValue = null;
-            if (child != null) {
-                childValue = child.accept(visitor, arg, depth + 1);
+            if (visitor instanceof NodeCalcEvaluator) {
+                if (cache.containsKey(this.hashCode())) {
```
geofjamg, line 72: Using the hashcode for caching is not correct as we could have has code collisions. You should use System.identity() or even better an IdentityHashMap. 
Maybe also we should consider a different approach:
 - using a node number in the AST to have a cache based on an array direct access
 - even better and a totally different approach, modeling directly in the AST a NodeCalc reference at the building phase so that at eval phase only caches value where there is reference NodeCalc.

geofjamg (changes requested):

rolnico:
Covered by https://github.com/powsybl/powsybl-core/pull/2899

PR #2870, opened 2024-01-23
Fix the visibility of the XxxLimitsAdderImpl classes

olperr1:

No

Bug fix

It is not possible/easy to directly instantiate a `CurrentLimitsAdderImpl` (for instance to add limits in an extension) because its constructor takes a `OperationalLimitsGroupImpl` which is protected since the operational limits refactoring.

The following classes are now public:
- `ActivePowerLimitsAdderImpl`;
- `ApparentPowerLimitsAdderImpl`;
- `CurrentLimitsAdderImpl`;
- `OperationalLimitsGroupImpl`;
- `OperationalLimitsGroupImpl.PermanentLimitInfo` (record);
- `OperationalLimitsGroupImpl.OperationalLimitsInfo` (record).

- [ ] Yes
- [X] No

**Other information**:

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/OperationalLimitsGroupImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/OperationalLimitsGroupImpl.java
@@ -15,21 +15,30 @@ import java.util.Optional;
 /**
  * @author Pauline Jean-Marie {@literal <pauline.jean-marie at artelys.com>}
  */
-class OperationalLimitsGroupImpl implements OperationalLimitsGroup, Validable {
+public class OperationalLimitsGroupImpl implements OperationalLimitsGroup, Validable {
 
     private final String id;
     private CurrentLimits currentLimits;
     private ActivePowerLimits activePowerLimits;
     private ApparentPowerLimits apparentPowerLimits;
-    private final AbstractIdentifiable<?> identifiable;
+    private final Validable validable;
+    private final Identifiable<?> identifiable;
     private final String attributeName;
     private String selectedGroupId;
 
-    OperationalLimitsGroupImpl(String id, AbstractIdentifiable<?> identifiable, String attributeName, String selectedGroupId) {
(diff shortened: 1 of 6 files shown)


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/OperationalLimitsGroupImpl.java
```diff
@@ -101,26 +110,34 @@ public void setApparentPowerLimits(ApparentPowerLimits limits) {
...
+    }
+
+    protected boolean notificationEnabled() {
+        return true;
+    }
+
+    private void doNotify(String attribute, Object oldValue, Object newValue) {
+        if (notificationEnabled() && identifiable instanceof AbstractIdentifiable<?> ident) {
+            ident.getNetwork().getListeners().notifyUpdate(identifiable, attribute, oldValue, newValue);
+        }
```
flo-dup, line 135: I'm not very fond of the condition. If we really ask in the constructor a parameter for each functionality, we could add a NetworkListenerList as parameter.

flo-dup (commented):
It's ok to open an implementation in iidm-impl. Just one point, if we extend we're forced to use iidm-impl. Meaning we could not use iidm-network-store as implementation with the class extending OperationlLimitsGroupImpl.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/OperationalLimitsGroupImpl.java
flo-dup: And then if we do that, we only need the id for the identifiable, hence the three fields: String ownerId, NetworkListenerList listeners, Validable validable

flo-dup: Identifiable is still needed for the notify

olperr1: You are right, this is way cleaner like this. I just pushed a change.

PR #2872, opened 2024-01-25
CGMES Import. Create tie lines when no subnetworks are present (assembled models). Simplify the code.

marqueslanauja:

Feature

When a cgmes model is imported without using subnetworks (assembled model) the paired `danglingLines` are created using specific methods based on the `BoundaryLine` class. These `danglingLines` are created at the same time as the corresponding `TieLine`. We have two options to create a danglingLine. 

The code has been simplified. 
Now there is only one alternative to create a `danglingLine`. `DanglingLines` are created always in the same way independently of the imported option (Importing using subnetworks or without using subnetworks).
In assembled mode, after creating the `danglingLines`, a `TieLine` is created for each pair of `danglingLines` that have the same `paringkey` and have different `regionName`.
When the `using subnetworks option` is true, the `TieLines` are created in the merging process. 

BoundaryLine class has been deleted.

- [ ] Yes
- [ ] No

**Other information**:


cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/conformity/modified/CgmesConformity1ModifiedConversionTest.java
```diff
@@ -562,12 +562,12 @@ void microAssembledTransformerAtBoundary() {
                 NetworkFactory.findDefault(), importParams);
         TieLine m = network.getTieLine("17086487-56ba-4979-b8de-064025a6b4da + 8fdc7abd-3746-481a-a65e-3df56acd8b13");
 
-        assertEquals(4.848348287766889, m.getR(), tolerance);
-        assertEquals(80.20069732770635, m.getX(), tolerance);
-        assertEquals(0.000006589225754810883, m.getG1(), tolerance);
-        assertEquals(-0.00003533205998631356, m.getB1(), tolerance);
-        assertEquals(0.00002420000000000048, m.getG2(), tolerance);
-        assertEquals(0.00008984950000000269, m.getB2(), tolerance);
+        assertEquals(4.788687265185187, m.getR(), tolerance);
```
annetill, line 565: You present your work as a refactoring, and indeed the code looks really simpler: I need to spend time to have a good understanding. But, why do you get these changes?

marqueslanauja: In this test, there is a `danglingLine` created from a `twoWindingTransformer`.

In the main version: 

- when the cgmes model is imported using subnetworks, the danglingLine associated to the twoWindingTransformer is created fixing the ratio to 1.0 and the angle to 0.0 (Ratio is not considered). 
- when the cgmes model is imported without using subnetworks,  the danglingLine associated to the twoWindingTransformer is created using the real ratio  and fixing only the angle to 0.0. 

After this PR there is only one way to create the danglingLines, the first one. I decided to keep this option as it was.

As the danglingLine does not allow to model properly the twoWindingTransformer, both options are reasonable. Considering the ratio is more precise, but as we can not considerer the angle, both options are valid, in my opinion.

rcourtier (approved):

annetill (approved):

PR #2888, opened 2024-02-02
[CGMES export] UUID of fullModel on each file should be unique

colinepiloquet:

No

Bug fix

When introducing the naming strategy for CGMES export, the creation of the `FullModel` UUIDs of each file (EQ, TP, SSH, SV) was changed to be persistent. They are always the same and based on the type of file. 

The creation of this UUID is now back as before so that we get a unique ID for each export for each file.

- [ ] Yes
- [x] No

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportUtil.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportUtil.java
@@ -116,9 +116,14 @@ public final class CgmesExportUtil {
 
     public static void writeModelDescription(Network network, CgmesSubset subset, XMLStreamWriter writer, ModelDescription modelDescription, CgmesExportContext context) throws XMLStreamException {
         // The ref to build a unique model id must contain:
-        // the network, the subset (EQ, SSH, SV, ...), and the FULL_MODEL part
+        // the network, the subset (EQ, SSH, SV, ...), the time of the scenario, the version and the FULL_MODEL part
         // If we use name-based UUIDs this ensures that the UUID for the model will be specific enough
-        CgmesObjectReference[] modelRef = {refTyped(network), ref(subset), Part.FULL_MODEL};
+        CgmesObjectReference[] modelRef = {
+                refTyped(network),
+                ref(subset),
+                ref(DATE_TIME_FORMATTER.format(context.getScenarioTime())),
+                ref(format(modelDescription.getVersion())),
+                Part.FULL_MODEL};
         String modelId = "urn:uuid:" + context.getNamingStrategy().getCgmesId(modelRef);
         modelDescription.setIds(modelId);
         context.updateDependencies();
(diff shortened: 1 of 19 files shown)


zamarrenolm:
The network was passed to the writeHeader method to provide a unique id based on the network id itself, the instance file profile, and the fact that the ID was made for "FullModel" output. 

The call to `refTyped(network)` was intended to return a unique ID based on the network id PLUS the network object type, in this case "Network", suffixed as "_N". 

If we want stable IDs for network models and instance files we must keep the existing way.

if we need to export revisions of the same network model / instance files we could use the `version` attribute of the `FullModel`.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportUtil.java
```diff
@@ -114,12 +113,8 @@ public static void writeRdfRoot(String cimNamespace, String euPrefix, String euN
         writer.writeNamespace("md", MD_NAMESPACE);
     }
 
-    public static void writeModelDescription(Network network, CgmesSubset subset, XMLStreamWriter writer, ModelDescription modelDescription, CgmesExportContext context) throws XMLStreamException {
-        // The ref to build a unique model id must contain:
-        // the network, the subset (EQ, SSH, SV, ...), and the FULL_MODEL part
-        // If we use name-based UUIDs this ensures that the UUID for the model will be specific enough
-        CgmesObjectReference[] modelRef = {refTyped(network), ref(subset), Part.FULL_MODEL};
```
annetill, line 121: This is not enough to have a unique id for several time step and different model version (when we re-import model after fixes): 
- We need the case date somwhere ;
- We need the model version. 
Where are these informations to generate a unique id?

zamarrenolm: If the full model identifier has to be dependent on the `scenarioTime` and the `version`, we can pass these references to the naming strategy, instead of building new random identifiers for every export.

zamarrenolm: Both attributes are available in the method that is building the identifier for the full model object.

zamarrenolm: We could just add items to the list of object references that will be used as a seed for stable identifiers in the naming strategy:

```java
        CgmesObjectReference[] modelRef = {
                refTyped(network),
                ref(subset),
                ref(DATE_TIME_FORMATTER.format(context.getScenarioTime())),
                ref(format(modelDescription.getVersion())),
                Part.FULL_MODEL};
        String modelId = "urn:uuid:" + context.getNamingStrategy().getCgmesId(modelRef);
```

zamarrenolm: Business processes involving CGMES exports typically will require the exchanged data to be unique by:

- Effective date time (or scenario time).
- Business process (market, operation, ...).
- Sourcing actor (who is producing the file).
- Model part (our CGMES subset).
- Version of the file (sequential, as an integer).

If we consider that the IIDM network id includes information about the sourcing actor (is different for each sourcing actor), then with the lines above we would be covering the requirement for unique identifiers, with the exception of business process: we would be generating the same ID for all business processess.

colinepiloquet: I will open a new PR with the introduction of a parameter for the business process.

zamarrenolm: for the purpose of generating unique model ids, I think we can rely on the IIDM network identifier to give different results for different sourcing actors; no need to explicitly use an additional reference to a sourcing actor

zamarrenolm (approved):

Issue #2868, opened 2024-01-23
In RAO: replace RAO Contingency by Core Contingency, In Core: Extends Contingency to cover open-rao needs


pjeanmarie:
### Describe the current behavior

Core project and open-rao project have each have their own definition of contingency:
- core: class com.powsybl.contingency.Contingency
- rao: interface com.powsybl.openrao.data.cracapi.Contingency, impl com.powsybl.openrao.data.cracimpl.ContingencyImpl, used in particular in Crac

### Describe the expected behavior

Core project and open-rao project should share a common definition and implementation of a contingency and a network-action which should be in core project only.

The idea is:
- when there is similar elements -> modify rao to use the one in core
- when there is necessary elements in rao which do not exist in core -> add them in core

**In fact rao contingency define an apply method which create a core contingency and apply it to the iidm network, so the link between rao and core contingency is well defined here.**

Changes details:
| rao | core | what to do? | change is validated in core or rao |
|--------|--------|--------|--------|
| name and id | only id | name is needed on rao side so getName which return an Optional name wil be added on core side | yes core (for the contingency and its builder) |
| network elements is a Set | network elements is a List | change to List on rao side | yes rao |
| getNetworkElements | getElements | rename on rao side since we should avoid regression on core side | yes rao |
| rao contingency contains NetworkElement | core contingency contains ContingencyElement | rao will use the core contingency  so it will contains ContingencyElement  and no more NetworkElement  | yes rao |
| NetworkElement has getName | ContingencyElement has not | contingency elements' name not used in rao, no need to add a getName in core | yes rao |
| NetworkElement has getLocation | ContingencyElement has not | not called on contingecy elements so it can be removed | yes rao |
| equals | equals | keep core | yes rao |
| hash with id and network elements | hash with id | keep core | yes rao |
| ContingencyAdder | ContingencyBuilder  | ContingencyBuilder need to be instantiated with contingency id, and add aslo non primitive type for some optional ".add", which is to much a change compare to ContingencyAdder and other rao adder so ContingencyAdder  is keeped but we use ContingencyBuilder inside as much as possible | yes rao and core for the optional Name |
| NetworkElement add in Crac |  | Do not create a network element for each contingecy element to add in crac expecially beacuse they have no name so can create discrepences with same network-element with a name | yes rao (adder) |
[...]

PR #2890, opened 2024-02-06
Add name in contingency for open rao

pjeanmarie:
This PR add an optional Name in core Contingency for powsybl-open-rao to be able to use directly core Contingency knowing that a name is mandatory for rao. 

This is part of the issue: https://github.com/powsybl/powsybl-core/issues/2868

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/contingency/contingency-api/src/main/java/com/powsybl/contingency/Contingency.java
+++ b/contingency/contingency-api/src/main/java/com/powsybl/contingency/Contingency.java
@@ -8,9 +8,9 @@ package com.powsybl.contingency;
 
 import com.powsybl.commons.extensions.AbstractExtendable;
 import com.powsybl.contingency.contingency.list.ContingencyList;
-import com.powsybl.iidm.network.*;
-import com.powsybl.iidm.modification.NetworkModificationList;
 import com.powsybl.iidm.modification.NetworkModification;
+import com.powsybl.iidm.modification.NetworkModificationList;
+import com.powsybl.iidm.network.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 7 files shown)


pjeanmarie:
Add serializer. NB. Contingency DSL not updated.

contingency/contingency-api/src/main/java/com/powsybl/contingency/Contingency.java
```diff
@@ -30,12 +30,25 @@ public class Contingency extends AbstractExtendable<Contingency> {
...
+        this.elements = new ArrayList<>(Objects.requireNonNull(elements));
+    }
+
+    public Contingency(String id, String name, ContingencyElement... elements) {
+        this(id, name, Arrays.asList(elements));
+    }
+
+    public Contingency(String id, String name, List<ContingencyElement> elements) {
+        this.id = Objects.requireNonNull(id);
+        this.name = Objects.requireNonNull(name);
```
obrix, line 51: The use of requireNonNull seems a bit strange here because name can be null in a instance. Would it not be simpler to just drop the requireNonNull and use this constructor at line 40 ?
You can then drop the test in ContingencyDeserializer at line 63.

contingency/contingency-api/src/main/java/com/powsybl/contingency/json/ContingencyDeserializer.java
```diff
@@ -58,8 +60,9 @@ public Contingency deserialize(JsonParser parser, DeserializationContext deseria
                 default -> throw new IllegalStateException("Unexpected field: " + parser.getCurrentName());
             }
         }
-
-        Contingency contingency = new Contingency(id, elements);
+        Contingency contingency = (name != null) ?
```
obrix, line 63: See upper comment about null name.

contingency/contingency-api/src/main/java/com/powsybl/contingency/Contingency.java
pjeanmarie: ok

contingency/contingency-api/src/main/java/com/powsybl/contingency/json/ContingencyDeserializer.java
pjeanmarie: done

contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyBuilder.java
```diff
@@ -24,12 +24,17 @@ public class ContingencyBuilder {
...
 
+    private String name;
+
     ContingencyBuilder(String id) {
         this.id = Objects.requireNonNull(id);
         this.elements = new ArrayList<>();
     }
 
     public Contingency build() {
+        if (name != null) {
```
annetill, line 35: I understand that now you can simplify the code to call just once the constructor with a null name as you has allowed it. Just to be sure: in open-rao, we always check that the name is not null before any use?

pjeanmarie: I will remove that. No previous rao core was "fine" with null name input, but get name was never null because it was using id in case of null name. Now the name can be optional but when there is a get name, a get Id is done if empty (to have same behavior).

obrix (approved):

annetill:
Release note not updated after merge. Is it a breaking change?

PR #2897, opened 2024-02-09
LoadScalable: Do not apply constant power factor if P0 is zero

vmouradian:

No

Bug fix (handle the case when Scaling a load with P0=0 and constantPowerFactor = true).

The calculated new Q0 value will be NaN because of dividing by P0=0. Which will make setQ0() throw an error.

The value of Q0 is not updated ant stays as is, instead of throwing an error.

- [ ] Yes
- [x] No

**Other information**:


jeandemanged (approved):
Thank you, I'd argue this is a bugfix, not a "_small feature_"

colinepiloquet (approved):
I agree that it's more a bug fix

m-guibert (approved):

PR #2898, opened 2024-02-13
CGMES metadata models extension

zamarrenolm:

Feature + bug because of dependencies management during export.

Multiple extensions to keep CGMES metadata:
- cgmesSshMetadata
- cgmesSvMetedata

CGMES metadata is kept in a single IIDM Network extension.

- [X] Yes
- [ ] No

**Other information**:

Bring to current development branch the initial work made in #2034


flo-dup:
After discussing with @annetill, the backward compatibility for extensions CgmesSshMetadata and CgmesSvMetedata does not need to be addressed. It seems that there's no usecase which needed to keep iidm files with those extensions (see slack question in the #powsybl channel). You can thus delete these extensions and their SerDe. If a (strong) need is expressed later, we'll deal with it with a corrective release.

cgmes/cgmes-extensions/src/main/java/com/powsybl/cgmes/extensions/CgmesMetadataModelsSerDe.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    @Override
+    public Map<String, String> getArrayNameToSingleNameMap() {
+        return Map.of(
+                MODELS, MODEL,
+                PROFILES, PROFILE,
+                DEPENDENT_ONS, DEPENDENT_ON,
+                SUPERSEDESS, SUPERSEDES
+                );
```
flo-dup, line 52: Just to recap, this map is used to translate the tag which might occur more than once in xiidm within the same tag to a list/plural tag in jiidm, as we cannot put the same attribute name twice in json. In this sense, `DependentOns` and Superserdess` are not correct.

Even if it's merely a matter of grammar (hence no reason to be different from one usage to the other) we should protect this by associating it with a namespace. Nonetheless this is not straightforward, so we need to put something else while waiting for it, for instance:
- DependentOnModel / DependentOnModels
- SupersedesModel / SupersedesModels
If we change it later, this means we will not be able to read the jiidm files produced previously. We will then deal with it by adding a 2.0 version to this SerDe.

cgmes/cgmes-extensions/src/main/resources/xsd/cgmesMetadataModels.xsd
flo-dup, line 1: Don't forget to remove the upper case for the first letter of each name

cgmes/cgmes-extensions/src/main/java/com/powsybl/cgmes/extensions/CgmesMetadataModelsSerDe.java
zamarrenolm: The names have been fixed following your proposal

cgmes/cgmes-extensions/src/main/resources/xsd/cgmesMetadataModels.xsd
zamarrenolm: The names have been adjusted.

zamarrenolm:
To support the use-case of CGM export we need to be able to build an SV model with "dependent on" references to its IGM updated SSH models and also to the original IGM TP models.

In a CGM IIDM `Network` we have each IGM as a `Subnetwork`. 

When a user wants to make a CGM export, first he has to export the SSH of each `Subnetwork`, then he has to export the SV at the `Network` level.

To keep the required information in the extension (the data about the exported SSH), we plan to update the extension every time an export is made. To avoid the extension growing indefinitely, we will only keep the last model for each part and modelling authority.

zamarrenolm:
After an internal discussion we prefer not to use the metadata extension to store results of export. 

Instead, to support the use case of CGM SSH, SV export, we will use an export context.

zamarrenolm:
I think that we should further unify the class ModelDescription used in the CGMES export context with the `CgmesMetadataModels.Model`. I propose to have a unique simple class for both uses in the `cgmes-model` module, that will be named `CgmesMetadataModel`. The `CgmesMetadataModels` extension will only be a container for objects of that class.

cgmes/cgmes-model/src/main/resources/CIM16.sparql
```diff
@@ -28,6 +28,37 @@ SELECT *
     OPTIONAL { ?FullModel md:Model.description ?description }
 }}
 
+# query: fullModels
+prefix md: <http://iec.ch/TC57/61970-552/ModelDescription/1#>
```
annetill, line 32: Why do you need new querys?

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/triplestore/CgmesModelTripleStore.java
```diff
@@ -257,6 +257,11 @@ public PropertyBags fullModel(String cgmesProfile) {
         return namedQuery("fullModel", cgmesProfile);
```
colinepiloquet, line 257: Should the "fullModel" query not be deleted? I think it will not be used anymore.

zamarrenolm: yes, it has been removed in the last commit

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/InMemoryCgmesModel.java
```diff
@@ -277,11 +277,6 @@ private void fakeObjectsFromIdentifiers(String propertyNameId, String[] ids, Pro
         numObjectsByType.add(p);
     }
 
-    @Override
-    public PropertyBags fullModel(String cgmesProfile) {
-        return new PropertyBags();
-    }
-
```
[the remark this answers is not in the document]
zamarrenolm: This was used to obtain the CGMES metadata model for a specific part. Now we want to make a single query to the CGMES database to obtain all the metadata models. We do not need anymore the method `fullModel(String cgmesProfile)`. We have replaced it by `fullModels()`.

cgmes/cgmes-model/src/main/resources/CIM16.sparql
```diff
@@ -28,6 +28,37 @@ SELECT *
     OPTIONAL { ?FullModel md:Model.description ?description }
```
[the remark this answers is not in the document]
zamarrenolm: Yes, it is not used anymore. I have removed it.

cgmes/cgmes-extensions/src/main/java/com/powsybl/cgmes/extensions/CgmesMetadataModelsSerDe.java
```diff
@@ -0,0 +1,128 @@
+/**
+ * Copyright (c) 2022, RTE (http://www.rte-france.com)
```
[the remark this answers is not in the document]
zamarrenolm: yes, fixed.

cgmes/cgmes-extensions/src/main/resources/xsd/cgmesMetadataModels.xsd
```diff
@@ -0,0 +1,31 @@
...
+        <xs:attribute name="part" type="xs:string"/>
+        <xs:attribute name="modelingAuthoritySet" use="required" type="cmm:nonEmptyString"/>
+        <xs:attribute name="id" type="xs:string"/>
+        <xs:attribute name="version" use="required" type="cmm:nonEmptyString"/>
+        <xs:attribute name="description" type="xs:string"/>
+    </xs:complexType>
+    <xs:element name="cgmesMetadataModels">
+        <xs:complexType>
+            <xs:sequence>
+                <xs:element name="model" type="cmm:Model" maxOccurs="unbounded"/>
```
flo-dup, line 27: Instead of a unbounded list of elements with a `part` enum, it would be clearer to restrict to 4 optional named elements (SV / SSH / TP / EQ).

zamarrenolm: We want the extension to be generic enough to store metadata for other parts. The complete list is defined in `CgmesSubset`. What do you think about defining the part as a simple enum type with the proper values? (like the Side in the branch observability IIDM extension)

```
EQUIPMENT
TOPOLOGY
STATE_VARIABLES
STEADY_STATE_HYPOTHESIS
DYNAMIC
DIAGRAM_LAYOUT
GEOGRAPHICAL_LOCATION
EQUIPMENT_BOUNDARY
TOPOLOGY_BOUNDARY
UNKNOWN
```

zamarrenolm: We are storing in the extension only the information about the models that have been read during the import of the network. We could limit it even more, and store only some of the models we have read.
I think that having a limited list of named elements instead of a property would make future modifications more difficult. What do you think?

flo-dup: Ok, I wasn't aware there are more! Ok to keep the list with a `part` enum then.

flo-dup: And about defining the `part` as a simple enum type in the xsd, it's a good idea yes

zamarrenolm: the part is now an enum in the xsd

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
```diff
@@ -59,10 +61,10 @@ public class CgmesExportContext {
...
     private String businessProcess = DEFAULT_BUSINESS_PROCESS;
 
-    private final ModelDescription eqModelDescription = new ModelDescription("EQ Model", cim.getProfileUri("EQ"));
-    private final ModelDescription tpModelDescription = new ModelDescription("TP Model", cim.getProfileUri("TP"));
-    private final ModelDescription svModelDescription = new ModelDescription("SV Model", cim.getProfileUri("SV"));
-    private final ModelDescription sshModelDescription = new ModelDescription("SSH Model", cim.getProfileUri("SSH"));
+    private final CgmesMetadataModel exportedEQModel = new CgmesMetadataModel(CgmesSubset.EQUIPMENT, DEFAULT_MODELING_AUTHORITY_SET_VALUE);
+    private final CgmesMetadataModel exportedTPModel = new CgmesMetadataModel(CgmesSubset.TOPOLOGY, DEFAULT_MODELING_AUTHORITY_SET_VALUE);
+    private final CgmesMetadataModel exportedSVModel = new CgmesMetadataModel(CgmesSubset.STATE_VARIABLES, DEFAULT_MODELING_AUTHORITY_SET_VALUE);
+    private final CgmesMetadataModel exportedSSHModel = new CgmesMetadataModel(CgmesSubset.STEADY_STATE_HYPOTHESIS, DEFAULT_MODELING_AUTHORITY_SET_VALUE);
```
rolnico, line 67: Do we want to keep those models hardcoded here? If another model is added, we'll have to add it here, add the corresponding methods (like `getExportedEQModel()`), etc.

Would it not be better to store them for example in a map, using the `CgmesSubset` enum as a key? It would allow to have only a `getExporterModel(CgmesSubset cmgesSubset)` method instead of one for each model.

rolnico (changes requested):
In addition to the other comment, could you also add some javadoc? At least to the public methods that have been modified

annetill (approved):
Really nice, thanks all!

rolnico (approved):

rolnico:
My comments will be covered by a future PR

PR #2899, opened 2024-02-13
Improving NodeCalc computing

### LINKED ISSUE #2891: Optimize processing of NodeCalc

When a NodeCalc tree is created, some sub-trees can be duplicated multiple times inside the main tree. 
For example:
```java 
NodeCalc node_C = BinaryOperation.plus(
            new TimeSeriesNameNodeCalc("A"),
            BinaryOperation.multiply(
                BinaryOperation.div(
                    new TimeSeriesNameNodeCalc("A"),
                    new TimeSeriesNameNodeCalc("B")),
                new DoubleNodeCalc(-1000.0)
            )
        );
NodeCalc node_D = new TimeNodeCalc(node_C);
NodeCalc node_E = new TimeSeriesNameNodeCalc("E")
NodeCalc node_F = BinaryOperation.plus(
            BinaryOperation.multiply(
                BinaryOperation.greaterThan(node_E, node_D),
                BinaryOperation.minus(node_E, node_D)
            ),
            BinaryOperation.multiply(
                BinaryOperation.lessThan(node_E, UnaryOperation.negative(node_D)),
                BinaryOperation.minus(node_E, UnaryOperation.negative(node_D))
            )
        );
```
In this case, the subtree corresponding to node_D is present at multiple places in the main tree (node_F).

Meanwhile, in the current implementations of the NodeCalcVisitor, each call to a node through a visitor will call the children using the same visitor, even if the subtree corresponding to one of the children has already been visited.
For example, when using `NodeCalcEvaluator`, the value of each child node is computed in order to compute the value of the parent node, even when the value of the child node has already been computed elsewhere in the tree. In our example, the value of node_D would be computed 4 times.

Another issue is that the serialisation/deserialisation in json (using `writeJson` and `parseJson`) or using `NodeCalcPrinter` and rebuilding the tree from the output does not keep in memory the fact that some nodes are present in multiple places, multiple identical nodes are simply created.

This leads to redundant calls, higher computing durations, higher memory consumption, etc.

A subtree present at multiple places should only be visited (as in computed) once for a single tree. Any subsequent visit to this node should then get the result previously computed.

In the same idea, serialisation/deserialisation should keep the link between the nodes: a node present at multiple places should still be the same single node present in the different places after Ser/De.

### Describe the motivation

_No response_

### Extra Information

_No response_

rolnico:

Fixes #2891

- [ ] Yes
- [x] No

**Other information**:


time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/AbstractMinMaxNodeCalc.java
```diff
@@ -12,30 +12,19 @@
...
 import java.io.IOException;
-import java.util.Objects;
 
 /**
  * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
  */
-public abstract class AbstractMinMaxNodeCalc implements NodeCalc {
-
-    protected NodeCalc child;
+public abstract class AbstractMinMaxNodeCalc extends AbstractSingleChildNodeCalc {
```
geofjamg, line 19: Same, you can add AbstractSingleChildNodeCalc in another PR as a refacto that we will merge before.

time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/BinaryMaxCalc.java
```diff
@@ -52,6 +53,6 @@ public boolean equals(Object obj) {
 
     @Override
     public int hashCode() {
-        return left.hashCode() + right.hashCode();
+        return Objects.hash(left, right, NAME);
```
geofjamg, line 56: Why do you need to add the NAME to the hash calculation?

time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/NodeCalcEvaluator.java
```diff
@@ -93,6 +103,23 @@ public NodeCalc iterate(MaxNodeCalc nodeCalc, DoubleMultiPoint multiPoint) {
         return nodeCalc.getChild();
     }
 
+    @Override
+    public Double visit(CachedNodeCalc nodeCalc, DoubleMultiPoint arg, Double child) {
+        double childValue;
+        if (child == null) {
```
geofjamg, line 109: Looking at the way it is implemented, do you really need to add CachedNodeCalc to the tree ? As far as you have the set of NodeCalc that are duplicated you can directly manage the caching in for instance the binary op node ?

time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/BinaryMaxCalc.java
rolnico: If I don't, a BinaryMinCalc and a BinaryMaxCalc with the same child nodes would have the same hashcode

time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/AbstractMinMaxNodeCalc.java
rolnico: new PR for this modification : #2906

time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/NodeCalcEvaluator.java
rolnico: I think it might be useful to keep it like this since it allows the users to do other things specifically on those nodes, without having to keep the set of duplicated nodes in memory. For example, if a user wants to change how the nodes are printed, or if s.he adds a new visitor to export the tree and wants to avoid duplication in the export, etc.

rolnico: Or else, we can indeed choose to not use this specific type of nodecalc and force the users to compute again the set of duplicated nodes if they need it somewhere

geofjamg: ok

time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/AbstractMinMaxNodeCalc.java
geofjamg: merged

time-series/time-series-api/src/main/java/com/powsybl/timeseries/DoubleTimeSeries.java
```diff
@@ -68,7 +68,12 @@ public int getTimeSeriesNum() {
             }
         }
 
-        Iterator<Map.Entry<Integer, List<DoublePointExt>>> it = points.entrySet().iterator();
+        List<Map.Entry<Integer, List<DoublePointExt>>> entrySet = new ArrayList<>();
```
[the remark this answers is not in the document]
geofjamg: Ok. Anyway looking at code it's not obvious for me that new code is faster as there is now a lot of hashmap lookup.

time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/NodeCalcEvaluator.java
```diff
@@ -9,13 +9,23 @@
...
 
+import java.util.IdentityHashMap;
+import java.util.Map;
+
 /**
  * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
  */
 public class NodeCalcEvaluator implements NodeCalcVisitor<Double, DoubleMultiPoint> {
 
+    private Map<NodeCalc, Double> cache;
```
geofjamg, line 20: So now the NodeCalcEvaluator is not stateless anymore. Maybe we could have use the visitor context for that. So instead of DoubleMultiPoint could be a 

```java
class EvalContext {
      DoubleMultiPoint point;
      Map<NodeCalc, Double> cache;
}
```

rolnico: modified

time-series/time-series-api/src/main/java/com/powsybl/timeseries/DoubleTimeSeries.java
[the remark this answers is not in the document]
geofjamg: Is this one faster (at least shorter) ?
```java
        Iterator<Map.Entry<Integer, List<DoublePointExt>>> it = points.entrySet().stream()
                .sorted(Comparator.comparingInt(Map.Entry::getKey))
                .iterator();
```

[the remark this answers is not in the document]
rolnico: Nice! It is indeed shorter even if not really faster (but not slower either)

PR #2903, opened 2024-02-22
Dangling Line Action

annetill:

No, but soon. `powsybl-open-rao` has an injection action for loads, generators and dangling lines (maybe more because very generic). To remove this action in order to use the ones of `powsybl-core`, we need at least a dangling line action, very close to the load action. 

We now have a dangling line action, separated to load action because:
- not really used in general compared to load action.
- the simulator will do a different treatment, so two objects seems better choice.

- [ ] Yes
- [ ] No

**Other information**:

Serialization/DeSerialization missing.


obrix (approved):

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/action/AbstractLoadActionDeserializer.java
```diff
@@ -0,0 +1,65 @@
+package com.powsybl.security.json.action;
```
EtienneLt, line 1: missing author and headers

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/action/AbstractLoadActionSerializer.java
```diff
@@ -0,0 +1,42 @@
+package com.powsybl.security.json.action;
```
EtienneLt, line 1: missing author and headers

EtienneLt (approved):

PR #2909, opened 2024-02-26
Export RegulatingControl only when generator has regulation capability.

### LINKED ISSUE #2908: [WIP] CGMES export RegulatingControl only when generator has regulation capability

Currently, `RegulatingControl` is exported for `SynchronousMachines` which have:
- `minQ` == `maxQ` (`ReactiveCapabilityCurve` not provided)
- `y1value` == `y2value` for every `xvalue` (`ReactiveCapabilityCurve` provided)

Since reactive limits are equal, these machines do not have the capability to regulate reactive power or voltage.

`RegulatingControl` should be exported only when:
- `minQ` != `maxQ` (`ReactiveCapabilityCurve` not provided)
- `y1value` != `y2value` for any `xvalue` (`ReactiveCapabilityCurve` provided)

### Describe the steps

_No response_

### Environment

_No response_

### Relevant Log Output

_No response_

### Extra Information

_No response_

nemanja-st:

Fixes #2908 

- [ ] Yes
- [x] No


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
```diff
@@ -525,13 +525,36 @@ private void addIidmMappingsGenerators(Network network) {
...
+                        break;
+                    }
+                }
+            } else if (reactiveLimits.getKind() == ReactiveLimitsKind.MIN_MAX) {
+                MinMaxReactiveLimits mmrl = (MinMaxReactiveLimits) reactiveLimits;
+                if (mmrl.getMaxQ() != mmrl.getMinQ()) {
+                    hasReactiveCapability = true;
+                }
+            }
+        }
```
annetill, line 554: As far as I understand, your proposal looks nice, every generator with a correct reactive capability curve will have a regulating control. To be sure: here, there is no more the idea of a difference between local and remote regulating control. Many more regulating control, is it what you expect? Then I wonder about a check on a not Double.NaN targetV.

nemanja-st: Currently, we're exporting more controls (generators which cannot regulate) so initial assumption was that this will reduce the number of exported controls.

With your suggestion in mind, indeed we might export more controls for generators which are controlling remotely, but I think this is ok, you want to export the true state of the grid. There is no reason not to export remote controls.

`targetValue` is mandatory attribute for `RegulatingControl` so I'll add the check for `Double.NaN` `targetV`.

nemanja-st: Fixed with the latest commit.

colinepiloquet (approved):
Note that I tested it with the SUV and on my test case, everything is fine with this modification.

zamarrenolm (changes requested):
Agree with the proposed changes. A generator can only regulate if it has reactive capability.

Just one comment about remote regulation: previously generators with regulating terminal DIFFERENT than its own (that is, with a define remote regulating terminal) were considered as regulating (without looking at reactive capability or currently defined target voltage) and a regulating control was created for them.

The reasoning for that was that if someone in IIDM explicitly defines a remote regulating terminal, it is a strong indication that the generation has regulation capabilities.

So I will suggest to complete the PR with that condition: if the generator has a remote regulating terminal, and has regulating capability, create the corresponding regulating control. Even if in the current state of the network the regulation is disabled (targetV not defined).

zamarrenolm (approved):

PR #2915, opened 2024-03-04
Add new field Name in equals and hash code

pjeanmarie:

- [ ] Yes
- [x] No

No breaking changes is expected if no name was used in contingency before (no specialization of Contingency.class with a field name) or if the name used was already a unique identifier of the contingency as the field id is.

**Other information**:


contingency/contingency-api/src/main/java/com/powsybl/contingency/Contingency.java
```diff
@@ -60,13 +60,13 @@ public List<ContingencyElement> getElements() {
...
     public int hashCode() {
-        return Objects.hash(id, elements);
+        return Objects.hash(id, name, elements);
     }
 
     @Override
     public boolean equals(Object obj) {
         if (obj instanceof Contingency other) {
-            return id.equals(other.id) && elements.equals(other.elements);
+            return id.equals(other.id) && Objects.equals(name, other.name) && elements.equals(other.elements);
```
EtienneLt, line 69:
```java
name.equals(other.name)
```
is it not better ?

annetill (changes requested):
Unit test missing.

contingency/contingency-api/src/main/java/com/powsybl/contingency/Contingency.java
pjeanmarie: name can be null

obrix (approved):

annetill (approved):

PR #2916, opened 2024-03-04
Allow task interruption in groovy scripts

rolnico:

No

Feature - same modification as https://github.com/powsybl/powsybl-metrix/pull/129

If a task is interrupted by a user, it's handled differently depending on whether the task contains a groovy script evaluation or not:
* If it's only a Java task, the interruption is well catched and the task is interrupted
* If it contains a Groovy script evaluation, the Java thread stops but the groovy evaluation continues, which can cause performance issues if multiple big groovy scripts are running even when they are supposed to be interrupted.

1. In the groovy CompilerConfiguration, a new ASTTransformationCustomizer on [ThreadInterrupt](https://docs.groovy-lang.org/latest/html/gapi/groovy/transform/ThreadInterrupt.html) has been added: it automatically adds a check on ThreadInterrupt in every loop (for, while) in the groovy scripts so that they stop if interrupted.
2. Right before the script evaluation begins, a check is done on the thread interruption in order to not evaluate the script if it has already been interrupted. 

- [ ] Yes
- [x] No


contingency/contingency-dsl/src/test/java/com/powsybl/contingency/dsl/GroovyContingencyListTest.java
```diff
@@ -156,4 +164,27 @@ void testExtension() throws IOException {
...
+    void testCancelTask(boolean isDelayed) throws Exception {
+
+        // Counters
+        waitForStart = new CountDownLatch(1);
+        waitForFinish = new CountDownLatch(1);
+        waitForInterruption = new CountDownLatch(1);
+
+        // Booleans
+        config = new AtomicBoolean(false);
+        interrupted = new AtomicBoolean(false);
```
olperr1, line 180: Why these values are not initialized directly in `testCancelShortTask`?

rolnico: no reason, it could (I just changed it)

olperr1 (approved):

PR #2919, opened 2024-03-06
Add VoltageRegulation extension

colinepiloquet:

No

Feature

New extension: VoltageRegulation

- [ ] Yes
- [x] No


iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/extensions/VoltageRegulationSerDe.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.Objects;
+
+/**
+ * @author Coline Piloquet {@literal <coline.piloquet@rte-france.fr>}
+ */
+@AutoService(ExtensionSerDe.class)
+public class VoltageRegulationSerDe<T extends Injection<T> & ReactiveLimitsHolder> extends AbstractVersionableNetworkExtensionSerDe<T, VoltageRegulation<T>> {
+
+    public VoltageRegulationSerDe() {
+        super("voltageRegulation", VoltageRegulation.class, "vr",
```
colinepiloquet, line 36: I don't know if we want it just for IIDM 1.12/ if we can add it for the other versions. Also do we want in core the second version of the extension for IIDM 1.0?

annetill: @geofjamg help :-)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/VoltageRegulationImpl.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+    }
+
+    @Override
+    public Terminal getRegulatingTerminal() {
+        return regulatingTerminal;
+    }
+
+    @Override
+    public void setRegulatingTerminal(Terminal regulatingTerminal) {
```
flo-dup, line 56: Just thinking at the behaviour of the SlackTerminal extension... It's strange to support multivariant and have a setter here but not supporting multivariant for that parameter.

flo-dup (approved):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/VoltageRegulationImpl.java
annetill: The regulating terminal is not variant dependent...

PR #2925, opened 2024-03-12
Default version number for unversioned TimeSeries

rolnico:

No

Feature

The default version number for unversioned TimeSeries is hardcoded in a method and unavailable from anywhere else.
Current value is 0.

The default version number for unversioned TimeSeries is now set in a variable in the `TimeSeries` interface so that it is available to use if needed. The value is now -1.
TimeSeries version allowed values are now:
- -1 for the unversioned TimeSeries
- Any positive integer (0 included) for versioned TimeSeries

The parsing will now check that the versions given by the users for their different TimeSeries are not equals to the default version number for unversioned TimeSeries.
A new boolean parameter in `TimeSeriesCsvConfig` allows to choose if the parsing is strict or not. If a TimeSeries version number is equals to the default version number for unversioned TimeSeries:
- If it's strict (`strictVersioningImport == true`), the parsing will throw a TimeSeriesException
- If it's not strict (`strictVersioningImport == false`), the parsing will only log a warning.
By default, the strict parsing is selected.

- [x] Yes
- [ ] No

- If you used the default `TimeSeriesCsvConfig` by calling `new TimeSeriesCsvConfig()` and imported some versioned TimeSeries with a version number at 0 (old default value for unversioned TimeSeries), it will now consider them as versioned TimeSeries since the new default value has been set to -1. 
- If you still want to import TimeSeries with a version number and have them considered as unversioned TimeSeries, you have to use the version number -1. However, it will now throw an exception. Replace your call from `new TimeSeriesCsvConfig()` to `new TimeSeriesCsvConfig(ZoneId.systemDefault(), TimeSeriesConstants.DEFAULT_SEPARATOR, true, TimeFormat.DATE_TIME, DEFAULT_MAX_COLUMNS, false)` in order to keep it working (you will still get warnings) or change your TimeSeries version number to another value if it's not mean to be an unversioned one.


rolnico:
This PR would be useful for the PR https://github.com/powsybl/powsybl-metrix/pull/130

rolnico:
Is it logical/prractical to have 0 as default value ? Should we change it to -1 ?
I'm questioning it since there is nothing that could prevent users to import a TS with a version at 0 only, which would make no difference with an unversioned TS.
Maybe a better idea would be to add a warning and/or throw an exception if that's the case.

time-series/time-series-api/src/main/java/com/powsybl/timeseries/TimeSeries.java
```diff
@@ -213,8 +222,31 @@ private List<String> createStringValues() {
             return stringValues;
         }
 
-        int getVersion(String[] tokens) {
-            return timeSeriesCsvConfig.versioned() ? Integer.parseInt(tokens[1]) : 0;
+        int getVersion(String[] tokens, ReportNode reportNode) {
```
marifunf, line 225: we could allow the case versioned == true and version == DEFAULT_VERSION_NUMBER if DEFAULT_VERSION_NUMBER is the unique version of csv file

time-series/time-series-api/src/main/java/com/powsybl/timeseries/TimeSeries.java
```diff
@@ -406,7 +438,8 @@ static void checkCsvHeader(TimeSeriesCsvConfig timeSeriesCsvConfig, String[] tok
         }
     }
 
-    static Map<Integer, List<TimeSeries>> parseCsv(BufferedReader reader, TimeSeriesCsvConfig timeSeriesCsvConfig) {
+    static Map<Integer, List<TimeSeries>> parseCsv(BufferedReader reader, TimeSeriesCsvConfig timeSeriesCsvConfig,
```
marifunf, line 441: perhaps we could also use this PR to change all TimeSeries with omitted type parameters (sonar indicates to provide the parametrized type)
concerned functions are at least : parseCsv, createTimeSeries, readCasValues, writeJson, toJson, parseChunks, parseJson

rolnico: It would be better to create another PR for this since it's a different issue (but it might be a good idea)

time-series/time-series-api/src/test/java/com/powsybl/timeseries/TimeSeriesTest.java
```diff
@@ -231,6 +161,67 @@ void testNoVersion() {
...
+
+    @Test
+    void testVersionedAtDefaultNumberStrictCSV() {
+        String csv = String.join(System.lineSeparator(),
+            "Time;Version;ts1;ts2",
+            "0.000;0;1.0;",
+            "0.001;0;;a",
+            "0.002;0;3.0;b",
+            "0.000;2;4.0;c",
+            "0.001;2;5.0;",
```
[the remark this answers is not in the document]
rolnico: yes, the point of the test is to check how the boolean parameter `strictVersioningImport` is used when importing this CSV

time-series/time-series-api/src/main/java/com/powsybl/timeseries/TimeSeries.java
rolnico: Since `DEFAULT_VERSION_NUMBER_FOR_UNVERSIONED_TIMESERIES` is supposed to be for unversioned TimeSeries, I think it would not be a good idea to allow that case. If a user really wants to do it, s.he just has to set `strictVersioningImport` to false.
It's debatable but it's my point of view

PR #2926, opened 2024-03-12
add network element identifiers for action elements

EtienneLt:

no

feature

actions target a network element with only an id, with these pr now networkElementIdentifiers can be set

- [ ] Yes
- [ ] No

**Other information**:


security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/AbstractAction.java
```diff
@@ -16,13 +20,24 @@
 public abstract class AbstractAction implements Action {
 
     private final String id;
+    private final List<NetworkElementIdentifier> networkElementIdentifiers;
```
rolnico, line 23: Why are you using a list here? Should we not use a single NetworkElementIdentifier in replacement of the element id we had previously in the subclasses?

rolnico (changes requested):
Some points:
- There are breaking changes!
- Why are you using a list of NetworkElementIdentifier and not a single NetworkElementIdentifier?
- Javadoc has to be updated at some places

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/AbstractAction.java
EtienneLt: because sometimes the network source is unknown and to be sur to get the right element you could get two identifiers 
for exemple one in ucte and one in another format

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/DanglingLineAction.java
```diff
@@ -22,22 +24,15 @@ public class DanglingLineAction extends AbstractLoadAction {
...
-    DanglingLineAction(String id, String danglingLineId, boolean relativeValue, Double activePowerValue, Double reactivePowerValue) {
-        super(id, relativeValue, activePowerValue, reactivePowerValue);
-        this.danglingLineId = Objects.requireNonNull(danglingLineId);
-    }
-
-    public String getDanglingLineId() {
-        return danglingLineId;
+    DanglingLineAction(String id, List<NetworkElementIdentifier> networkElementIdentifiers, boolean relativeValue, Double activePowerValue, Double reactivePowerValue) {
+        super(id, networkElementIdentifiers, relativeValue, activePowerValue, reactivePowerValue);
     }
```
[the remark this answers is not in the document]
EtienneLt: well it passes by the builder and it assures compatibility, the only problem could be on serialization Am I wrong ?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/GeneratorAction.java
```diff
@@ -58,10 +58,6 @@ public String getType() {
         return NAME;
     }
 
-    public String getGeneratorId() {
-        return generatorId;
-    }
-
```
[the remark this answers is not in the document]
EtienneLt: yes I dont know how to assumes compatibility with the getters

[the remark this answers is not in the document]
EtienneLt: maybe I can leave the string variable with the list of identifiers ?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/DanglingLineAction.java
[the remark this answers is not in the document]
rolnico: I went a bit fast on this one, it's not a public method so not a breaking change

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/action/AbstractAction.java
rolnico: In that case, it would be better to change how you did it and to keep the same principles as for the contingencies, where the list of contingencies are created with the right network elements in the `IdentifierContingencyList`.
It would mean:

- keeping the actions as they were before (no breaking change)
- adding a new IdentifierActionList class which would have a map of Action builders and lists of networkElementIdentifiers, plus a method to create the list of actions, just like `getContingencies(Network network)` in `IdentifierContingencyList`.

EtienneLt: ok I see

EtienneLt: indeed not exactly

EtienneLt:
refactor in #2952

PR #2760, opened 2023-10-29
[WIP] Limit reduction API

So-Fras:
**Please check if the PR fulfills these requirements**

- [x] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**Does this PR already have an issue describing the problem?**
#2751 

**Does this PR introduce a breaking change or deprecate an API?**

- [ ] The *Breaking Change* or *Deprecated* label has been added
- [ ] The migration guide has been updated in the github wiki *(What changes might users need to make in their application due to this PR?)*

PR #2935, opened 2024-03-18
Use double format for limit reduction values

olperr1:

No

Feature

Limit reduction values are in **float** format in:
- `LimitViolation`;
- `LimitReduction`;
- the methods taking a limit reduction value in `LimitViolationUtils`, the action DSL and in the API's equipment classes.

Limit reduction values are in **double** format.

- [X] Yes
- [ ] No

- `LimitViolation.getLimitReduction` now returns a `double` instead of a `float`;

- If you have defined your own IIDM implementation, you should change the type of the `limitReductionValue` parameter from `float` to `double` in the `checkPermanentLimit…` and `checkTemporaryLimit…` methods of your `Branch`'s and `ThreeWindingTransformer`'s implementations;

- If you have a custom `LimitViolationDetector`'s implementation, you should change the type of the `limitReductionValue` parameter from `float` to `double` in the `checkPermanentLimit` and `checkTemporaryLimit` methods;

- `AbstractBranchActionExpressionNode.getLimitReduction()` now returns a `double` instead of a `float`.

**Other information**:

The `PR: do-not-merge` tag was added since this PR relies on #2760. This latter should be merged prior to the current one.

### DIFF
--- a/action/action-dsl/src/main/java/com/powsybl/action/dsl/ast/AbstractBranchActionExpressionNode.java
+++ b/action/action-dsl/src/main/java/com/powsybl/action/dsl/ast/AbstractBranchActionExpressionNode.java
@@ -16,9 +16,9 @@ public abstract class AbstractBranchActionExpressionNode extends AbstractActionE
 
     private final List<String> branchIds;
 
-    private final float limitReduction;
+    private final double limitReduction;
 
-    public AbstractBranchActionExpressionNode(List<String> branchIds, float limitReduction) {
+    public AbstractBranchActionExpressionNode(List<String> branchIds, double limitReduction) {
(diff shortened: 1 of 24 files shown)


rolnico (approved):
Looks good !
New issues spotted by Sonar are unrelated, they probably come from the parent branch.

Could you maybe just add in the migration steps some notes for the methods from the action module that now take double parameters instead of float, since they are public ?

olperr1:
> Could you maybe just add in the migration steps some notes for the methods from the action module that now take double parameters instead of float, since they are public ?

`float` values can be used for `double` parameters when calling a method. Thus, the parameters' format change for the methods you are pointing out is not a breaking change.

annetill (approved):

PR #2938, opened 2024-03-19
Conditional actions results

obrix:
Support results for each conditional actions list instead of a single result per operator strategy.

- [x] Yes
- [ ] No

OperatorStrategyResult object now has a list of ConditionalActionsResult objects instead of only the final state results. Each ConditionActionsResult represents the result of the ConditionalActions action package application.

* For an operator strategy with a single ConditionalActions object ConditionalActionsResult will have a single object containing the same results as before this change.
* For operator strategy with multiple ConditionalActions objects, ConditionalActionsResult will have one result per ConditionalActions  object and the last one in the list will be the same as the single one that was present before this change.
* ConditionalActionsResult can be associated to ConditionalActions using conditionalActionsId attribute.

**Other information**:


obrix:
I kept compatibility in result serializer / deserializer with versioning but there is a breaking change in the OperatorStrategyResult getters API.

geofjamg (commented):
Which such a design, how can we if are not interested by intermediate results after each conditional action, only get final result ?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResultBuilder.java
```diff
@@ -280,28 +280,21 @@ public SecurityAnalysisResultBuilder endContingency() {
...
-            Objects.requireNonNull(limitViolationContext);
-            violations.add(Objects.requireNonNull(violation));
-            //TODO: call to interceptors
-            return this;
-        }
-
-        public OperatorStrategyResultBuilder setStatus(PostContingencyComputationStatus status) {
-            this.status = status;
-            return this;
+        public ConditionalActionsResultBuilder addConditionalActionsResult(String conditionalActionsId) {
```
geofjamg, line 297: Should be newConditionalActionsResult to be consistent with the others and because at this stage we do not add anything

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResultBuilder.java
```diff
@@ -280,28 +280,21 @@ public SecurityAnalysisResultBuilder endContingency() {
         }
     }
 
-    public class OperatorStrategyResultBuilder extends AbstractLimitViolationsResultBuilder<OperatorStrategyResultBuilder> {
+    public class OperatorStrategyResultBuilder {
 
         private final OperatorStrategy strategy;
 
-        private PostContingencyComputationStatus status = PostContingencyComputationStatus.CONVERGED;
+        private final List<OperatorStrategyResult.ConditionalActionsResult> conditionalActionsResult = Collections.synchronizedList(new ArrayList<>());
```
geofjamg, line 288: Is it on purpose? Do we really support concurrent build of operator strategies?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisParameters.java
```diff
@@ -182,6 +188,15 @@ public SecurityAnalysisParameters setIncreasedViolationsParameters(IncreasedViol
         return this;
     }
 
+    public SecurityAnalysisParameters setIntermediateResultsInOperatorStrategy(boolean enable) {
+        intermediateResultsInOperatorStrategy = enable;
```
geofjamg, line 193: enable -> intermediateResultsInOperatorStrategy

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/ConditionalActionsResultDeserializer.java
```diff
@@ -0,0 +1,51 @@
+package com.powsybl.security.json;
```
geofjamg, line 1: License and authorship are missing

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/ConditionalActionsResultSerializer.java
```diff
@@ -0,0 +1,25 @@
+package com.powsybl.security.json;
```
geofjamg, line 1: License and authorship are missing

geofjamg (changes requested):

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/ConditionalActionsResultDeserializer.java
```diff
@@ -0,0 +1,51 @@
...
+
+public class ConditionalActionsResultDeserializer extends StdDeserializer<OperatorStrategyResult.ConditionalActionsResult> {
+
+    public ConditionalActionsResultDeserializer() {
+        super(OperatorStrategyResult.ConditionalActionsResult.class);
+    }
+
+    @Override
+    public OperatorStrategyResult.ConditionalActionsResult deserialize(JsonParser parser, DeserializationContext deserializationContext) throws IOException {
+        String conditionalActionsId = null;
```
EtienneLt, line 24: should a builder be better ? I don know

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResultBuilder.java
obrix: postContingencyResults and operatorStrategyResults both use synchronizedList so I figured there was some concurrent use somewhere and followed the same pattern. Maybe unecessary ?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/ConditionalActionsResultDeserializer.java
obrix: I could reuse ConditionalActionsResultBuilder here but I don't think it really brings something more interesting than a simple call to the constructor.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/results/OperatorStrategyResult.java
```diff
@@ -7,46 +7,110 @@
...
+     * @return The Operator strategy associated to this result
+     */
+    public OperatorStrategy getOperatorStrategy() {
+        return operatorStrategy;
+    }
+
+    /**
+     * @return The list of ConditionalActionsResult for each conditional actions of the associated operator strategy
+     */
+    public List<ConditionalActionsResult> getConditionalActionsResult() {
```
annetill, line 102: Why not `getConditionalActionsResults()` here? And a good method could be `getConditionalActionsResult(String id)`

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisResultBuilder.java
geofjamg: yes unnecessary

geofjamg (approved):

PR #2939, opened 2024-03-19
Support of validation level for missing permanent limit

annetill:

No

Feature

If a `LoadingLimit` has temporary limits, the permanent limit is mandatory. This is imposed by the way the limit violations detection works. When importing an IIDM network, if a permanent limit is missing :
- the network is rejected for IIDM version >= v1.12
- the permanent limit is computed from the 1st temporary limit's value.

At this subject, see #2636

With this principle, it is not possible to import a network (especially of an older IIDM version) and prompt a user for the missing permanent limits.

With this PR, it is possible have IIDM networks with missing permanent limits (either created via the API or imported) as long as the validation level is `EQUIPMENT`.

When importing an IIDM network, it is possible to override its minimum validation level. To do this, you should use the following import parameter: `iidm.import.minimal-validation-level: EQUIPMENT`.

Operational limits are now considered by the validation level mechanism:

- When adding operational limits on a network element with a missing permanent limit, or when settting a permanent limit to Double.NaN, the operation will:
  - succeed if the validation level is `EQUIPMENT`;
  - fail if the minimum validation level is not `EQUIPMENT`;
  - succeed if the validation level is not `EQUIPEMENT`, but the **minimum** validation level is. In this case, the validation level will be lowered to `EQUIPMENT`.

- When only a missing permanent limit prevents a network to be in `STEADY_STATE_HYPOTHESIS` level, using the `LoadingLimits.setPermanentLimit(...)` method to define it a value will increase the validation level of the network to `STEADY_STATE_HYPOTHESIS`. 

Note that all operational limits are taken into account to determine the validation level. It is also the case for the operational limits defined in a non-selected `OperationalLimitsGroup` (the group is not currently used for the corresponding network element).

- [X] Yes
- [ ] No

The following utilitary methods:
- `ValidationUtil.checkLoadingLimits(...)`;
- `ValidationUtil.checkPermanentLimit(...)`;

now take two additional parameters:
- `boolean throwException`: should the method throw an exception when a missing permanent limit is detected? 
- `ReportNode reportNode`: the reportNode to use to log potential errors. You can use `ReportNode.NO_OP` if you are not interested by this log.

**Other information**:


iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractNetworkTest.java
```diff
@@ -626,4 +626,16 @@ public void testScadaNetwork() {
...
+    @Test
+    public void testPermanentLimit() {
+        Network network = EurostagTutorialExample1Factory.createWithFixedCurrentLimits();
+        assertEquals(ValidationLevel.STEADY_STATE_HYPOTHESIS, network.getValidationLevel());
+        ValidationException e = assertThrows(ValidationException.class, () -> network.getLine("NHV1_NHV2_1").getCurrentLimits2().orElseThrow().setPermanentLimit(Double.NaN));
+        assertTrue(e.getMessage().contains("AC line 'NHV1_NHV2_1': permanent limit must be defined if temporary limits are present"));
+        network.setMinimumAcceptableValidationLevel(ValidationLevel.EQUIPMENT);
+        network.getLine("NHV1_NHV2_1").getCurrentLimits2().orElseThrow().setPermanentLimit(Double.NaN);
+        assertTrue(Double.isNaN(network.getLine("NHV1_NHV2_1").getCurrentLimits2().orElseThrow().getPermanentLimit()));
+        assertEquals(ValidationLevel.STEADY_STATE_HYPOTHESIS, network.getValidationLevel()); // not supported through setters...
```
olperr1, line 639: I think it is important to have this.
This relies on the `ValidationUtil.checkIdentifiable(...)` method which should now consider the limits to determine the validation level of a network.

olperr1 (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -589,15 +589,30 @@ public static void checkConnected(Validable validable, Boolean connected) {
...
         ValidationUtil.checkTemporaryLimits(validable, permanentLimit, temporaryLimits);
+        return validationLevel;
     }
 
-    public static void checkPermanentLimit(Validable validable, double permanentLimit, Collection<LoadingLimits.TemporaryLimit> temporaryLimits) {
-        if (Double.isNaN(permanentLimit) && !temporaryLimits.isEmpty() || permanentLimit <= 0) {
-            throw new ValidationException(validable, "permanent limit must be defined and be > 0");
+    public static ValidationLevel checkPermanentLimit(Validable validable, double permanentLimit, Collection<LoadingLimits.TemporaryLimit> temporaryLimits,
+                                                      boolean throwException, ReportNode reportNode) {
+        ValidationLevel validationLevel = ValidationLevel.STEADY_STATE_HYPOTHESIS;
```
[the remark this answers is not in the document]
annetill: It is the way it is implemented everywhere else.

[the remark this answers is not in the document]
olperr1: Ok, you are right.
Since we never directly use this method's result as the voltageLevel of the network, but via `setValidationLevelIfGreaterThan(...)`, it is OK.

(We take the highest possible value and retain the min between the previous value and the newly computed one at every step to finally use the min between this value and the network's value. Thus, it is equivalent.)

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractNetworkTest.java
olperr1: I pushed modifications to take the limits into account when determining the validation level of the network.

I choose to consider all the `OperationalLimitsGroup`s (the selected and the unselected) since it is easier to do: when considering only the selected ones, the validation level should be checked every time the selected operationalLimitsGroup is changed for any limits holder.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTreeDataImporter.java
```diff
@@ -54,6 +54,8 @@ public abstract class AbstractTreeDataImporter implements Importer {
 
     public static final String MISSING_PERMANENT_LIMIT_PERCENTAGE = "iidm.import.xml.missing-permanent-limit-percentage";
 
+    public static final String MINIMAL_VALIDATION_LEVEL = "iidm.import.xml.minimal-validation-level";
```
annetill, line 57: import -> serde

olperr1: I don't know if it is a good idea to remove the "import" part of the parameter. We can have parameters dealing with the same concept for import and export and we may want them to have different values (for instance, we have both `iidm.export.xml.extensions` and `iidm.import.xml.extensions` parameters).

We can maybe use something like "serde.import", or just "import"?

annetill: @flo-dup proposes `iidm.import.minimal-validation-level`

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTreeDataImporter.java
```diff
@@ -69,6 +71,10 @@ public abstract class AbstractTreeDataImporter implements Importer {
             ParameterType.DOUBLE, "Percentage applied to lowest temporary limit to compute the permanent limit when missing (for IIDM < 1.12 only)",
             100.);
 
+    public static final Parameter MINIMAL_VALIDATION_LEVEL_PARAMETER = new Parameter(MINIMAL_VALIDATION_LEVEL,
```
miovd, line 74: It should be added in `getParameters()`

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ImportOptions.java
```diff
@@ -79,4 +83,21 @@ public ImportOptions setMissingPermanentLimitPercentage(double missingPermanentL
     public double getMissingPermanentLimitPercentage() {
         return missingPermanentLimitPercentage;
     }
+
+    public ImportOptions setMinimalValidationLevel(String minimalValidationLevel) {
+        if (minimalValidationLevel != null) {
+            // no check?
```
miovd, line 89: Maybe a check would be cleaner? But not mandatory, in any case, it should throw an exception if the value is not supported

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -679,12 +679,21 @@ private static Network initNetwork(NetworkFactory networkFactory, NetworkDeseria
         network.setCaseDate(date);
         network.setForecastDistance(forecastDistance);
 
-        ValidationLevel[] minValidationLevel = new ValidationLevel[1];
-        minValidationLevel[0] = ValidationLevel.STEADY_STATE_HYPOTHESIS;
-        IidmSerDeUtil.runFromMinimumVersion(IidmVersion.V_1_7, context, () -> minValidationLevel[0] = reader.readEnumAttribute(MINIMUM_VALIDATION_LEVEL, ValidationLevel.class));
+        ValidationLevel minValidationLevel;
+        if (context.getOptions().getMinimalValidationLevel().isPresent()) {
+            minValidationLevel = context.getOptions().getMinimalValidationLevel().orElseThrow();
```
miovd, line 684: Specify the exception thrown in `orElseThrow`

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTreeDataImporter.java
olperr1: The parameter was changed accordingly.

miovd (approved):
Mostly okay for me, only small remarks

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTreeDataImporter.java
olperr1: Done

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
olperr1: The else part couldn't be reached because of the previous check.
I introduced a variable (to avoid a Sonar issue) and used `.get()` instead of `.orElseThrow()`.

miovd (approved):

olperr1 (approved):

PR #2943, opened 2024-03-19
Remove violation detector from SecurityAnalysis API

flo-dup:

No

Refactor

LimitViolationDetector in SecurityAnalysis API, not in line wit

- `LimitViolationDetector` removed from `SecurityAnalysis`, `SecurityAnalysisProvider` and `SecurityAnalysisInput`
- interface `LimitViolationDetector` moved to dedicated package `com.powsybl.security.detectors`
- `LimitViolationDetector` parameter in `DefaultSecurityAnalysis` now optional (nullable)

- [x] Yes
- [ ] No

Starting from this release the `LimitViolationDetector` is removed from security analysis API. The default implementations remains available, and the `DefaultSecurityAnalysis` default implementation still handles a `LimitViolationDetector` parameter through its constructor, although it is now becoming nullable. The `DefaultSecurityAnalysisProvider` now builds a `DefaultSecurityAnalysis` with a null `LimitViolationDetector`.

If you want to still use the `LimitViolationDetector` you need to pass one to your security analysis (or to the `DefaultSecurityAnalysis`). Hence, you cannot use directly the `DefaultSecurityAnalysisProvider` anymore. It is thus not possible to call `SecurityAnalysis.find()` to create the `SecurityAnalysis.Runner` anymore. Instead, you should declare your own `SecurityAnalysisProvider` (which may or not extends `DefaultSecurityAnalysisProvider`) which will create a `DefaultSecurityAnalysis` using the constructor taking a `LimitViolationDetector`. As a consequence, the provider won't be retrieved dynamically from the `default-impl-name` property of the configuration file.


security-analysis/security-analysis-default/src/main/java/com/powsybl/security/impl/DefaultSecurityAnalysis.java
```diff
@@ -284,4 +292,22 @@ private ThreeWindingsTransformerResult createThreeWindingsTransformerResult(Thre
...
+        } else {
+            LimitViolationDetection.checkAll(network, EnumSet.allOf(LoadingLimitType.class), 1., consumer);
+        }
+    }
+
+    protected void checkPostContingencyViolations(Contingency contingency, Network network, Consumer<LimitViolation> consumer) {
+        if (violationDetector != null) {
+            violationDetector.checkAll(contingency, network, consumer);
+        } else {
+            // TODO For now, contingencies are ignored in the default security analysis
```
flo-dup, line 307: Maybe remove as already discussed the `TODO For now` and `This should change to fully support LimitReductions`, and add something like `This blocks a full support of LimitReductions`. Indeed we won't probably change this behaviour in the future.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/detectors/LimitViolationDetector.java
```diff
@@ -483,7 +476,7 @@ default void checkPermanentLimit(ThreeWindingsTransformer transformer, ThreeSide
     /**
      * Generic implementation for temporary limit checks
```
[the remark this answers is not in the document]
olperr1: You are right.
I changed the returned value (`true` when no overload is detected) and updated the javadoc.

security-analysis/security-analysis-default/src/main/java/com/powsybl/security/impl/DefaultSecurityAnalysis.java
olperr1: I removed the `TODO` and added a clearer comment.

flo-dup (commented):
The PR is now ok for me!

olperr1 (approved):
PR OK with @flo-dup (see previous comment)

PR #2946, opened 2024-03-21
PSSE. copy of the imported PSSE model to make the updates and export

marqueslanauja:
…copy)

Feature

The original psse records are updated before exporting them. 

Original psse records are always preserved. A partial copy, only the updated record groups are copied, is made before updating them.

- [ ] Yes
- [x] No

**Other information**:


zamarrenolm:
I think it would be better to deep copy all the record groups when preparing the updated data, instead of deep copy some of them and keeping references for others. If later we have to make updates in a record group that is only saved as a reference we have to remember to make the deep copy first.

marqueslanauja:
> I think it would be better to deep copy all the record groups when preparing the updated data, instead of deep copy some of them and keeping references for others. If later we have to make updates in a record group that is only saved as a reference we have to remember to make the deep copy first.

We can think about it.

geofjamg:
> > I think it would be better to deep copy all the record groups when preparing the updated data, instead of deep copy some of them and keeping references for others. If later we have to make updates in a record group that is only saved as a reference we have to remember to make the deep copy first.
> 
> We can think about it.

This is true that it would be maybe more robust to deep copy everything but let's do it later in another PR.

geofjamg (approved):

PR #2950, opened 2024-03-22
Add a builder for LimitReductions

olperr1:

No

Feature + doc update

When creating a `LimitReduction` with either a contingency context or a criterion list, every characteristics of a `LimitReduction` must be given as parameter for the constructor. This forces the user to give empty lists to the constructor for the unneeded parameters. 

- Two small constructors are now present for simple cases:
  - `LimitReduction(LimitType limitType, double value)`;
  - `LimitReduction(LimitType limitType, double value, boolean monitoringOnly)`.
- For more complex cases, a builder was added.

- [X] Yes
- [ ] No

When you used the constructor with all the parameters to create a "LimitReduction", you should use the builder.
Default values for the characteristics are:
- `monitoringOnly`: false;
- `contingencyContext`: `ContingencyContext.all()`;
- `networkElementCriteria`: `Collections.emptyList()`;
- `limitDurationCriteria`: `Collections.emptyList()`.

Thus, you can omit setting one of the characteristics if you used its default value.

For instance:

```java
       LimitReduction limitReduction = new LimitReduction(LimitType.APPARENT_POWER, 0.5, false,
                ContingencyContext.all(),
                Collections.emptyList(),
                List.of(durationCriterion1, durationCriterion2));
```

could be rewritten in

```java
       LimitReduction limitReduction = LimitReduction.builder(LimitType.APPARENT_POWER, 0.5)
                .withLimitDurationCriteria(durationCriterion1, durationCriterion2)
                .build();
```

Note:
- `monitoringOnly`, `networkElementCriteria` and `contingencyContext` are not specified as their default values are used;
- the `with...Criteria(...)` methods take either a `List` or varArgs. You could thus omit the encapsulating `List.of()` for ease of use. But in either case, the previously definied criteria are replaced by the given ones. 

**Other information**:

### DIFF
--- a/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/limitreduction/LimitReductionDeserializer.java
+++ b/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/limitreduction/LimitReductionDeserializer.java
@@ -12,13 +12,12 @@ import com.fasterxml.jackson.databind.DeserializationContext;
 import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
 import com.powsybl.commons.json.JsonUtil;
 import com.powsybl.contingency.ContingencyContext;
-import com.powsybl.iidm.network.LimitType;
 import com.powsybl.iidm.criteria.NetworkElementCriterion;
-import com.powsybl.security.limitreduction.LimitReduction;
 import com.powsybl.iidm.criteria.duration.LimitDurationCriterion;
+import com.powsybl.iidm.network.LimitType;
+import com.powsybl.security.limitreduction.LimitReduction;
 
 import java.io.IOException;
-import java.util.Collections;
 import java.util.List;
 
 /**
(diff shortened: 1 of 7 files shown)


security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/limitreduction/LimitReductionDeserializer.java
```diff
@@ -76,9 +75,17 @@ public LimitReduction deserialize(JsonParser parser, DeserializationContext dese
                 }
             }
         });
-        return new LimitReduction(context.limitType, context.value, context.monitoringOnly,
-                context.contingencyContext,
-                context.networkElementCriteria != null ? context.networkElementCriteria : Collections.emptyList(),
-                context.durationCriteria != null ? context.durationCriteria : Collections.emptyList());
+        LimitReduction.Builder builder = LimitReduction.builder(context.limitType, context.value)
+                .withMonitoringOnly(context.monitoringOnly);
+        if (context.contingencyContext != null) {
```
annetill, line 80: What is the meaning of a null contingency context?

olperr1: We are in the deserializer here, the contingency context may not have been specified in the JSON file to import.

annetill: Why do we have these two ways? Null and ALL ?

olperr1: Ok, I will set the ContingencyContext as mandatory in the JSON file.

annetill (approved):

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/json/limitreduction/LimitReductionDeserializer.java
olperr1: The ContingencyContext is now mandatory in the JSON files to import.

PR #2951, opened 2024-03-22
More robust VoltageInterval

olperr1:

No

Feature

When importing a `...NominalVoltageCriterion`, if a bound of the interval is missing, no exception is thrown, but the interval is considered as "null" and is not taken into account.

- `SingleNominalVoltageCriterion.VoltageInterval` was extracted in a dedicated class:`VoltageInterval`;
- The concept of "null interval" is removed. as a consequence:
   - `getVoltageInterval...()` method of `TwoNominalVoltageCriterion` and `ThreeNominalVoltageCriterion` now return an `Optional<VoltageInterval>` instead of a `VoltageInterval`
- When one of its bound is missing, the interval is no more considered as a "null interval". Instead, it represents:
    - [0 ; maxValue|: if only a max value is defined (the upper bound is opened or closed depending on a parameter); 
    - |minValue ; Double.MAX_VALUE]: if only a min value is defined (the lower bound is opened or closed depending on a parameter).
-  The existing constructor was replaced by:
   - `build()` static method, returning a builder;
   - 3 convenient static methods `lowerThan(...)`, `greaterThan(...)` and `between(...)`, creating an interval with one or both bounds.
- Since a bound can be missing, `getNominalVoltageLowBound()` and `getNominalVoltageHighBound()` now return an `Optional<Double>` instead of a `Double`;
- A new method `VoltageInterval.asRange()` returns a `DoubleRange` representation of the VoltageInterval.

- [X] Yes
- [ ] No

Generalities:

`SingleNominalVoltageCriterion.VoltageInterval` was renommed as `VoltageInterval`.

The following methods of `VoltageInterval` were renamed to match the bean methods syntax:
- `getLowClosed()` => `isLowClosed()`
- `getHighClosed()` => `isHighClosed()`

<br/>

Since interval with only one bound are now considered as valid:

`VoltageInterval.getNominalVoltageLowBound()` and `VoltageInterval.getNominalVoltageHighBound()` now return an `Optional<Double>` instead of a `Double` to indicated if a bound was defined or not.

The concept of "null interval" was removed, as for the `VoltageInterval.isNull()` method. If you need to indicate that an interval was not set, use a "null" value instead.

As a consequence, `getVoltageInterval...()` methods of `TwoNominalVoltageCriterion` and `ThreeNominalVoltageCriterion` now return an `Optional<VoltageInterval>` instead of a `VoltageInterval`.

<br/>

New ways to create an interval:

The historical constructor was removed. Instead, you can use either the `build()` static method which returns a builder, or one of the `greaterThan`, `lowerThan` and `between` static methods which return an interval with one or both bounds. 

For instance, you can replace:

```java
new VoltageInterval(200.0, 230.0, true, true)
```
by either

``` java
VoltageInterval interval = VoltageInterval.builder()
    .setLowBound(200., true)
    .setHighBound(230.0, true)
    .build();
```

or

``` java
VoltageInterval.between(200.0, 230.0, true, true)
```

**Other information**:

### DIFF
--- a/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisParameters.java
+++ b/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisParameters.java
@@ -26,9 +26,14 @@ public class SecurityAnalysisParameters extends AbstractExtendable<SecurityAnaly
 
     // VERSION = 1.0
     // VERSION = 1.1 IncreasedViolationsParameters adding.
-    public static final String VERSION = "1.1";
+    // VERSION = 1.2 intermediateResultsInOperatorStrategy
+    public static final String VERSION = "1.2";
 
     private LoadFlowParameters loadFlowParameters = new LoadFlowParameters();
+
+    static final boolean DEFAULT_INTERMEDIATE_RESULTS_IN_OPERATOR_STRATEGY = false;
+
+    private boolean intermediateResultsInOperatorStrategy = DEFAULT_INTERMEDIATE_RESULTS_IN_OPERATOR_STRATEGY;
 
     public static class IncreasedViolationsParameters {
 
@@ -155,6 +160,7 @@ public class SecurityAnalysisParameters extends AbstractExtendable<SecurityAnaly
 
         platformConfig.getOptionalModuleConfig("security-analysis-default-parameters")
                 .ifPresent(config -> {
+                    parameters.setIntermediateResultsInOperatorStrategy(config.getBooleanProperty("intermediate-results-in-operator-strategy", DEFAULT_INTERMEDIATE_RESULTS_IN_OPERATOR_STRATEGY));
                     parameters.getIncreasedViolationsParameters()
                             .setFlowProportionalThreshold(config.getDoubleProperty("increased-flow-violations-proportional-threshold", IncreasedViolationsParameters.DEFAULT_FLOW_PROPORTIONAL_THRESHOLD))
                             .setLowVoltageProportionalThreshold(config.getDoubleProperty("increased-low-voltage-violations-proportional-threshold", IncreasedViolationsParameters.DEFAULT_LOW_VOLTAGE_PROPORTIONAL_THRESHOLD))
@@ -181,6 +187,15 @@ public class SecurityAnalysisParameters extends AbstractExtendable<SecurityAnaly
     public SecurityAnalysisParameters setIncreasedViolationsParameters(IncreasedViolationsParameters increasedViolationsParameters) {
         this.increasedViolationsParameters = Objects.requireNonNull(increasedViolationsParameters);
         return this;
+    }
+
+    public SecurityAnalysisParameters setIntermediateResultsInOperatorStrategy(boolean intermediateResultsInOperatorStrategy) {
+        this.intermediateResultsInOperatorStrategy = intermediateResultsInOperatorStrategy;
+        return this;
+    }
+
+    public boolean getIntermediateResultsInOperatorStrategy() {
+        return intermediateResultsInOperatorStrategy;
     }
 
     public LoadFlowParameters getLoadFlowParameters() {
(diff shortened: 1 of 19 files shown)


iidm/iidm-criteria/src/main/java/com/powsybl/iidm/criteria/SingleNominalVoltageCriterion.java
```diff
@@ -101,19 +127,39 @@ public boolean checkIsBetweenBound(Double value) {
             }
         }
 
-        public Double getNominalVoltageLowBound() {
+        /**
+         * <p>Return a {@link DoubleRange} representation of the interval.</p>
+         * <p>This method throws an exception for {@link #nullInterval()} result.</p>
+         * @return the interval as a {@link DoubleRange}
+         */
+        public DoubleRange asRange() {
```
annetill, line 135: Thanks!

iidm/iidm-criteria/src/main/java/com/powsybl/iidm/criteria/SingleNominalVoltageCriterion.java
```diff
@@ -67,15 +68,34 @@ public VoltageInterval getVoltageInterval() {
...
+            this.nominalVoltageLowBound = Double.NaN;
+            this.nominalVoltageHighBound = Double.NaN;
         }
 
-        public VoltageInterval(Double nominalVoltageLowBound, Double nominalVoltageHighBound, Boolean lowClosed, Boolean highClosed) {
+        public static VoltageInterval nullInterval() {
+            return NULL_INTERVAL;
+        }
+
+        public VoltageInterval(double nominalVoltageLowBound, double nominalVoltageHighBound, boolean lowClosed, boolean highClosed) {
```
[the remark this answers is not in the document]
olperr1: The problem here was that at least one of the bounds was not defined, the interval was considered as a "null interval"  and was simply ignored.
Thus, the following criterion was true for every network element!

```xml
 "nominalVoltageCriterion" : {
      "type" : "SINGLE_NOMINAL_VOLTAGE",
      "voltageInterval" : {
        "nominalVoltageHighBound" : 230.0,
        "highClosed" : true
      }
    }
```
I choose to stick with the existing principle that both bounds should be defined for the voltageInterval to be defined, but
I secured the class. Now, we can still have a null interval, but it must be created explicitly to prevent errors.

EtienneLt (approved):

iidm/iidm-criteria/src/main/java/com/powsybl/iidm/criteria/SingleNominalVoltageCriterion.java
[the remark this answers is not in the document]
olperr1: Seen with @annetill and @EtienneLt: I will adapt this PR to have the same behaviour as for the duration interval. It will thus be possible to have an interval with a single bound. In the case of the previous comment's example, the criteria will return `true` for values in [0; 230] (0 is the min value for valid nominal voltages).

iidm/iidm-criteria/src/main/java/com/powsybl/iidm/criteria/VoltageInterval.java
```diff
@@ -0,0 +1,224 @@
...
+                .build();
+    }
+
+    /**
+     * <p>Convenient method to easily create a {@link VoltageInterval} with only a upper bound.</p>
+     * @param value the upper bound of the interval to create (it corresponds to the <code>nominalVoltageHighBound</code> attribute of the interval)
+     * @param valueIncluded is the bound included in the interval (it corresponds to the <code>highClosed</code> attribute of the interval)
+     * @return an interval
+     */
+    public static VoltageInterval lowerThan(double value, boolean valueIncluded) {
```
annetill, line 127: Just a detail here but do you change for `valueIncluded`? Closed was nice.

annetill (approved):
Just a small remark but globally nicer than before!

iidm/iidm-criteria/src/main/java/com/powsybl/iidm/criteria/VoltageInterval.java
olperr1: I used this term because I think that when someone says "something lower than a value", it is more natural to indicate if the value is "valid" or included ("or equals") than to say that the corresponding interval is closed.

But the intent behind this term is maybe not that clear.

I will change it to "closed" for consistency.

annetill (approved):

PR #2955, opened 2024-03-27
Generic Area model in IIDM

### LINKED ISSUE #2883: Generic Area model in iIDM

iIDM (today) v1.12 supports the following which somewhat mimics areas:
- Specifying the Country of a Substation
- Specifying the TSO of a Substation
- Adding geographical tags per Substation (simple list of string)
- Subnetworks (available since v1.11)

Although above data model fits most simple use cases, it comes with limitations for real world cases:
- smallest granularity is a Substation
- the area boundaries are not explicitly listed
- a geographical tag is nothing more than a string
- no possibility to distinguish between a Dangling Line representing an AC interconnection vs. a DC Converter modeled through a simple Equivalent Injection

We propose to introduce a generic "area" concept in iIDM v1.13, separated from the above-mentioned features which will remain as is.

### Describe the motivation

- Area Interchange Control in LoadFlow
- Security Analysis, Rebalancing in contingency state based on control blocks
- Security Analysis, Countertrading Remedial Actions based on Bidding Zones
- Zonal automatons

#### AreaType

AreaType-s are modeled by a simple string attribute in Area. 

PowSyBl does not pre-define any AreaType, user can adapt to their specific use case.

Examples of AreaTypes are:
- "ControlArea",
- "SchedulingArea",
- "BiddingZone"
- "ControlBlock"
- "OrganizationalRegion" such as GeographicalRegion, SubGeographicalRegion, Capacity Coordination Region, Outage Coordination Region, ...
- "AutomatonProtectedZone"
- (Note that Country & TSO could as well fit in there)

Note: for a later iIDM version, AreaType-s could be an `iidm:Identifiable` (using a new `iidm:IdentifiableType.AREA_TYPE`)and defined/contained in an iidm:Network and allow data engineer to describe the type of areas that will be instantiated in the Network. Those instances would not follow the very hierarchical geographic relationship that other Identifiable-s have, and could exist in several subnetworks and in a mergedNetwork. How to handle properly such transverse instance remains to be defined.

#### Area

An Area is defined by:
- (unique) ID and other attributes from `iidm:AbstractIdentifiable`
- An areaType (String, mandatory)
- a target AC Interchange (MW, load sign convention, optional)

Then depending on use case, an Area may be used to:
- define area boundaries (using the AreaBoundary datamodel)
- and/or, assign the area to voltage levels

##### AreaBoundary

An AreaBoundary is defined by
- Either a Connectable's Terminal, OR a DanglingLine's Boundary (these two options are mutually exclusive)
- and and ac boolean attribute to indicate if the area boundary is to be considered AC or DC

##### Area assignment

VoltageLevel-s may be assigned zero, one, or multiples areas. 

A VoltageLevel object may have only one Area of a given AreaType.
Defining two areas of the same AreaType on the same VoltageLevel is forbidden and rejected.

Networks and Subnetworks are not assigned to area. Their getAreas methods return all of their known Area-s, which may include several Area-s of the same areaType.

##### Network merge
Two network with the same Area (ie Area-s with the same ID) cannot be merged.
The merge will fail with an error, as it does for all Identifiable-s with conflicting IDs.

#### iIDM (de)serialization

serialisation to iIDM versions <= 1.12 will not export area related objects/fields.

m-guibert:

Fixes #2883 

- [X] Yes
- [ ] No

If you had define your own IIDM implementation, you should implement the following methods:
- in your `Network` implementations:
  - `Iterable<String> getAreaTypes();`
  - `Stream<String> getAreaTypeStream();`
  - `int getAreaTypeCount();`
  - `AreaAdder newArea();`
  - `Iterable<Area> getAreas();`
  - `Stream<Area> getAreaStream();`
  - `Area getArea(String id);`
  - `int getAreaCount();`
- in your `VoltageLevel` implementations:
  - `Iterable<Area> getAreas();`
  - `Stream<Area> getAreasStream();`
  - `Optional<Area> getArea(String areaType);`
  - `void addArea(Area area);`
  - `void removeArea(Area area);`


olperr1:
In the current state, there is a problem with the `Subnetwork`s. Since `Area` and `AreaType` don't have a real `parentNetwork` management, they will always be attached to the root network. The `attach` and `detach` methods won't work as we could expect.

For instance, if we add the following code at the end of the `AbstractAreaTest.checkSubnetworkNewAreas()` unit test:

```java
        Network detachedNetwork = subnetwork.detach();
        Assertions.assertAll(
                () -> assertTrue(Iterables.contains(detachedNetwork.getAreaTypes(), areaType)),
                () -> assertTrue(Iterables.contains(detachedNetwork.getAreas(), area)),
                () -> assertFalse(Iterables.contains(network.getAreaTypes(), areaType)),
                () -> assertFalse(Iterables.contains(network.getAreas(), area))
        );
```

Since `area` and `areaType` were created in the subnetwork, we expect that they will be still accessible from it when it is detached. But all these assertions fail.

Note that a special management should be introduced. If an `AreaType` or an `Area` is created (or imported) from a subnetwork, it would be available for the whole network and it would be possible to use it from another subnetwork.
The **used** `AreaType`s and `Area`s should maybe be cloned at detach (in the original network or in the detached network).

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Area.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * @author Marine Guibert {@literal <marine.guibert at artelys.com>}
+ * @author Valentin Mouradian {@literal <valentin.mouradian at artelys.com>}
+ */
+public interface Area extends Identifiable<Area> {
+
+    AreaType getAreaType();
+
+    void setAreaType(AreaType areaType);
```
olperr1, line 23: This method should not be available in the interface.
It was added only to manage the merge case, but having it in the interface may indicate that this is a "regular" operation (plus, in the actual implementation, nothing unsure that no VoltageLevel of the current Area is not already linked to another Area of the new AreaType).

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Area.java
```diff
@@ -0,0 +1,61 @@
...
+     */
+    Optional<Double> getAcNetInterchangeTarget();
+
+    /**
+     * Get the optional net interchange tolerance in MW
+     * @return the net interchange tolerance
+     */
+    Optional<Double> getAcNetInterchangeTolerance();
+
+    boolean isMergeable(Area other);
```
olperr1, line 58: I don't think it is a good idea to have this method exposed in the interface.
There's no real Areas merging usage except during the networks merging operation.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/AreaType.java
```diff
@@ -0,0 +1,23 @@
...
+ * @author Valentin Mouradian {@literal <valentin.mouradian at artelys.com>}
+ */
+public interface AreaType extends Identifiable<AreaType> {
+
+    @Override
+    default IdentifiableType getType() {
+        return IdentifiableType.AREA_TYPE;
+    }
+
+    boolean isMergeable(AreaType other);
```
olperr1, line 21: Same here, I don't think it is a good idea to have this method exposed in the interface.

olperr1 (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Network.java
```diff
@@ -618,6 +618,24 @@ static PrettyNetworkFactory with(String name) {
...
+
+    Iterable<AreaType> getAreaTypes();
+
+    Stream<AreaType> getAreaTypeStream();
+
+    AreaType getAreaType(String id);
+
+    AreaAdder newArea();
+
+    /** @return all existing Area-s of the Network, which may include several areas for each AreaType */
```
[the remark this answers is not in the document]
m-guibert: It has been modified

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AreaImpl.java
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * @author Marine Guibert {@literal <marine.guibert at artelys.com>}
+ * @author Valentin Mouradian {@literal <valentin.mouradian at artelys.com>}
+ */
+public class AreaImpl extends AbstractIdentifiable<Area> implements Area {
+    private final Ref<NetworkImpl> networkRef;
+    private final AreaType areaType;
+    private final Optional<Double> acNetInterchangeTarget;
+    private final Optional<Double> acNetInterchangeTolerance;
```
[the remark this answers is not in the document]
m-guibert: Thank you for the feedback. It is fixed

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/AbstractAreaAdder.java
```diff
@@ -0,0 +1,21 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+package com.powsybl.iidm.network;
+
+/**
+ * @author Marine Guibert {@literal <marine.guibert at artelys.com>}
+ * @author Valentin Mouradian {@literal <valentin.mouradian at artelys.com>}
+ */
+public interface AbstractAreaAdder<T extends Area, A extends AbstractAreaAdder<T, A>> extends IdentifiableAdder<T, A> {
```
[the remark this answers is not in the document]
m-guibert: It was useful at some point of the dev but is now obsolete. The classes are merged now.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
```diff
@@ -254,6 +254,46 @@ public int getCountryCount() {
...
     }
 
+    @Override
+    public AreaTypeAdder newAreaType() {
+        return new AreaTypeAdderImpl(ref);
+    }
+
+    @Override
+    public Iterable<AreaType> getAreaTypes() {
+        return Iterables.concat(index.getAll(AreaTypeImpl.class));
```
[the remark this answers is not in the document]
m-guibert: Modified

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/SubnetworkImpl.java
```diff
@@ -106,6 +106,46 @@ public int getCountryCount() {
         return (int) getCountryStream().count();
     }
 
+    @Override
```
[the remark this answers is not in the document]
m-guibert: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/BoundaryPointType.java
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * @author Marine Guibert {@literal <marine.guibert at artelys.com>}
+ * @author Valentin Mouradian {@literal <valentin.mouradian at artelys.com>}
+ */
+public enum BoundaryPointType {
+    HVDC,
+    PAIRED_AC,
+    UNPAIRED_AC,
+    IGNORED_AC
+}
```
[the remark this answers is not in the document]
m-guibert: The whole "boundaryPoint" model was a WIP and has since been modified. 
We now use a simpler structure: a BoundaryTerminal record, with just a boolean to indicate if the terminal if on an AC or DC line. The paired/unpaired info can be computed when needed according to the pairing state. The IGNORED_AC value was meant to flag boundaries that we want to exclude from a computation (when computing the Net Interchange value of the Area eg). But in the end, such ignored elements just won't be added to the Area-s.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Area.java
vmouradian: I agree, I changed the merge implementation so that we don't need this anymore

vmouradian:
> In the current state, there is a problem with the `Subnetwork`s. Since `Area` and `AreaType` don't have a real `parentNetwork` management, they will always be attached to the root network. The `NetworkFactory.merge(...)` and `Network.detach()` methods won't work as we could expect.
> 
> Note that a special management should be introduced. If an `AreaType` or an `Area` is created (or imported) from a subnetwork, it would be available for the whole network and it would be possible to use it from another subnetwork. The **used** `AreaType`s and `Area`s should maybe be cloned at detach (in the original network or in the detached network).

The implementation has been updated for the subnetworks, and Areas are handled for the detach.
Areas are considered related to a subnetwork if they contain voltagelevels that are contained in the subnetwork

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractAreaTest.java
```diff
@@ -0,0 +1,201 @@
...
+        assertEquals(List.of(biddingZoneA), subnetwork.getAreaStream().toList());
+
+        assertEquals(1, Iterables.size(subnetwork.getAreas()));
+        assertTrue(Iterables.contains(subnetwork.getAreas(), biddingZoneA));
+        assertEquals(1, Iterables.size(subnetwork.getAreaTypes()));
+        assertTrue(Iterables.contains(subnetwork.getAreaTypes(), biddingZone));
+    }
+
+    @Test
+    void checkSubnetworkNewAreas() {
```
olperr1, line 112: Could you add a test with the following characteristics?
- the network has an area;
- this Area is not "visible" from the subnetwork;
- we try to create the "same" area from the subnetwork.

What is the expected behavior in this case? I think it should not fail, but the "new" area should be the same as the existing?

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/AreaSerDeTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+import static com.powsybl.iidm.serde.IidmSerDeConstants.CURRENT_IIDM_VERSION;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * @author Marine Guibert {@literal <marine.guibert at artelys.com>}
+ * @author Valentin Mouradian {@literal <valentin.mouradian at artelys.com>}
+ */
+class AreaSerDeTest extends AbstractIidmSerDeTest {
```
olperr1, line 26: There should be a SerDe test with subnetworks to check that it works as expected.

olperr1 (changes requested):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AreaAdderImpl.java
```diff
@@ -53,11 +57,32 @@ public AreaAdder setAcNetInterchangeTolerance(Double acNetInterchangeTolerance)
...
     }
 
+    public AreaAdder addProperty(String key, String value) {
+        properties.setProperty(key, value);
+        return this;
+    }
+
+    public AreaAdder copy(Area otherArea) {
+        this.setId(otherArea.getId())
+                .setName(otherArea.getNameOrId())
```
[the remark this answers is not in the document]
m-guibert: Thank you for pointing that. It has been fixed

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AreaTypeAdderImpl.java
```diff
@@ -22,6 +23,14 @@ public class AreaTypeAdderImpl extends AbstractIdentifiableAdder<AreaTypeAdderIm
         this.networkRef = networkRef;
     }
 
+    @Override
+    public AreaTypeAdder copy(AreaType otherAreaType) {
+        this.setId(otherAreaType.getId())
+                .setName(otherAreaType.getNameOrId())
```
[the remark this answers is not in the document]
m-guibert: Same, it is fixed

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/MergeTest.java
```diff
@@ -199,6 +201,260 @@ void mergeNetworksWithSameCaseDates() {
         assertEquals(zonedDateTime1, networkMerged.getCaseDate());
     }
 
+    @Test
+    void mergeNodeBreakerWithAreasTestPass1() {
```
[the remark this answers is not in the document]
m-guibert: Done. The moved tests were renamed and refactored to be more efficient and understandable.

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/MergeTest.java
```diff
@@ -199,6 +201,260 @@ void mergeNetworksWithSameCaseDates() {
...
+        n1Bza.addVoltageLevel(n1S1VL1);
+        n2Bz2.addVoltageLevel(n2S1VL1);
+
+        Terminal g1Terminal = n1.getGenerator(id("S1VL1-Generator", "1")).getTerminal();
+        n1Bza.addBoundaryTerminal(g1Terminal, true);
+
+        // Merge
+        Network merged = Network.merge(n1, n2);
+
+        assertEquals(List.of(n1Bz), merged.getAreaTypeStream().toList());
```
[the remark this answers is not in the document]
m-guibert: The refactored tests in AbstractMergeNetworkTest are now using assertions on id and names.

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/MergeTest.java
```diff
@@ -199,6 +201,260 @@ void mergeNetworksWithSameCaseDates() {
...
+        VoltageLevel n2S1VL1 = n2.getVoltageLevel(id("S1VL1", "2"));
+
+        n1Bza.addVoltageLevel(n1S1VL1);
+        n2Bzb.addVoltageLevel(n2S1VL1);
+
+        // Merge
+        Network merged = Network.merge(n1, n2);
+
+        assertEquals(List.of(n1Bz), merged.getAreaTypeStream().toList());
+        Area mergedBzb = merged.getArea("bzb"); // not the same object as n2Bzb but a duplicate
```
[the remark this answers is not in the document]
m-guibert: Comment deleted during the refactor of those tests. The refactored tests in AbstractMergeNetworkTest are now using assertions on id and names to work whether objects are reused or not.

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/MergeTest.java
```diff
@@ -199,6 +201,260 @@ void mergeNetworksWithSameCaseDates() {
...
+
+        // Detached Network
+        AreaType detachedBz = n1Detached.getAreaType("bz");
+        Area detachedBza = n1Detached.getArea("bza");
+
+        assertEquals(List.of(detachedBz), n1Detached.getAreaTypeStream().toList());
+        assertEquals(List.of(detachedBza), n1Detached.getAreaStream().toList());
+        assertEquals(Set.of(n1S1VL1), detachedBza.getVoltageLevels());
+
+        // Previous Network
```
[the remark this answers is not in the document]
m-guibert: These tests were refactored to limit the redundancy and clarify what is tested.

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/MergeTest.java
```diff
@@ -227,8 +227,20 @@ private static Network createNodeBreakerWithVoltageAngleLimit(String nid) {
     }
 
     private static Network createNodeBreakerWithVoltageAngleLimit(NetworkFactory networkFactory, String nid, String valId) {
```
olperr1, line 229: Since finally no test was added / changed in this class, could you revert it in its initial state?

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/MergeTest.java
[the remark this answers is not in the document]
olperr1: Thank you, it is much clearer now.

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/MergeTest.java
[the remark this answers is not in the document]
olperr1: Thanks

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/MergeTest.java
m-guibert: It is fixed, thank you.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/AreaBoundary.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+public interface AreaBoundary {
+
+    Optional<Terminal> getTerminal();
+
+    Optional<Boundary> getBoundary();
+
+    boolean isAc();
+
+    double getP();
```
annetill, line 65: Why do you need getP() and getQ() ? You have all you need in `Terminal` and `Boundary`, no?

annetill: Ok if not serialized.

iidm/iidm-serde/src/test/resources/V1_12/areaRoundTripRef.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
olperr1, line 1: Could you adapt this file and the V1_11 one to use the same network as for v1.13?
It would make it easier to spot the differences between each version.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -350,6 +350,7 @@ private static void writeBaseNetwork(Network n, NetworkSerializerContext context
         writeLines(n, context);
         writeTieLines(n, context);
         writeHvdcLines(n, context);
+        writeAreas(n, context);
```
olperr1, line 353: You need to indicate that `Areas` should only be exported for 1.13+ versions of IIDM:
```suggestion
        IidmSerDeUtil.runFromMinimumVersion(IidmVersion.V_1_13, context, () -> writeAreas(n, context));
```
Or else it won't be possible to export networks containing areas with a previous IIDM version (which could be useful if you have downstream softwares using an older version of PowSyBl).

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/AreaSerDeTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @Test
+    void testNetworkAreas() throws IOException {
+        Network network = createBaseNetworkWithAreas();
+        allFormatsRoundTripTest(network, "/areaRoundTripRef.xml", CURRENT_IIDM_VERSION);
```
olperr1, line 27: Could you also add:
```java
allFormatsRoundTripTest(network, "/areaRoundTripRef.xml", IidmVersion.V_1_12);
```
to check that the network is also exportable with a prior version.

(This is not checked by the following assertion.)

olperr1 (changes requested):
I still have the unit tests to review, but here is some feedback.

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractAreaTest.java
```diff
@@ -0,0 +1,294 @@
...
+        // Deleting equipment from the network should automatically delete their AreaBoundary-s
+        network.getLine(EurostagTutorialExample1Factory.NHV1_NHV2_1).remove();
+        network.getGenerator("GEN").remove();
+        network.getDanglingLine("danglingLine1").remove();
+
+        assertEquals(0, aicA.getAreaBoundaryStream().count());
+        assertBoundaries(Map.of(), aicA);
+    }
+
+    void assertBoundaries(Map<Object, Boolean> expectedBoundaries, Area area) {
```
olperr1, line 250: I don't know if this method, relying on objects equalities, will work with every IIDM implementations. It may be useful to have it in public.

olperr1 (changes requested):

iidm/iidm-serde/src/main/resources/xsd/iidm_V1_13.xsd
```diff
@@ -703,18 +702,13 @@
...
         <xs:attribute name="id" use="required" type="iidm:nonEmptyString"/>
     </xs:complexType>
-    <xs:complexType name="BoundaryTerminal">
-        <xs:sequence>
-            <xs:element name="terminalRef" type="iidm:TerminalRef"/>
-        </xs:sequence>
+    <xs:complexType name="AreaBoundary">
+        <xs:attribute name="type" use="required" type="iidm:nonEmptyString"/>
+        <xs:attribute name="id" use="required" type="iidm:nonEmptyString"/>
+        <xs:attribute name="side" type="iidm:Side"/>
```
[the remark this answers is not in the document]
vmouradian: It is needed in case of a Terminal used as a boundary. We identify the terminal by the id of the connectable and a side if it is a branch. 
We use the methods of the terminalRef serde but wanted to avoid nesting for just one TerminalRef.

docs/grid_model/network_subnetwork.md
```diff
@@ -95,6 +95,26 @@ When defining the model, the user has to specify how the different equipment con
...
+## Area
+[![Javadoc](https://img.shields.io/badge/-javadoc-blue.svg)](https://javadoc.io/doc/com.powsybl/powsybl-core/latest/com/powsybl/iidm/network/Area.html)
+
+An Area represents geographical zone defined by the [voltage levels](#voltage-level) that it contains and by its boundaries, that can be terminals of equipments or `Boundary` objects from [dangling lines](#dangling-line).
+
+Areas can represent various granularities depending on their types. For instance: control areas, bidding zones, countries...
+
+A [voltage level](#voltage-level) can belong to several Areas, as long as they all have different types.
+
+The powerflow through the area boundaries can be summed and used to perform calculations for area interchange control.
```
annetill, line 107: Please add convention. What is the definition of ac net interchange and dc net interchange with and without any boundaries in the list.

iidm/iidm-serde/src/main/resources/xsd/iidm_V1_13.xsd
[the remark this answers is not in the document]
olperr1: OK, you are right. I missed this.
Thanks.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AreaBoundaryAdderImpl.java
```diff
@@ -0,0 +1,84 @@
...
+    @Override
+    public AreaBoundaryAdder setAc(boolean ac) {
+        this.ac = ac;
+        return this;
+    }
+
+    @Override
+    public void add() {
+        if (isAc() == null) {
+            throw new PowsyblException("AreaBoundary must have a non-null attribute 'ac' to be added");
```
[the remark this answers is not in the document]
vmouradian: adressed in f1f2d0b18b106f55b23b5d541da74a5e8ac138df

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AreaBoundaryAdderImpl.java
```diff
@@ -0,0 +1,84 @@
...
+        }
+        if (getBoundary() != null) {
+            getArea().addAreaBoundary(new AreaBoundaryImpl(getBoundary(), isAc()));
+        } else if (getTerminal() != null) {
+            getArea().addAreaBoundary(new AreaBoundaryImpl(getTerminal(), isAc()));
+        } else {
+            throw new PowsyblException("AreaBoundary must have a non-null 'terminal' or 'boundary' attribute be added");
```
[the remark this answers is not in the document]
vmouradian: adressed in f1f2d0b18b106f55b23b5d541da74a5e8ac138df

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AreaImpl.java
```diff
@@ -80,13 +112,8 @@ public Optional<Double> getAcNetInterchangeTolerance() {
...
     @Override
     public void addVoltageLevel(VoltageLevel voltageLevel) {
-        voltagelevels.add(voltageLevel);
         voltageLevel.addArea(this);
-    }
-
-    @Override
-    public void removeBoundaryTerminal(Terminal terminal) {
-        boundaryTerminals.removeIf(boundaryTerminal -> boundaryTerminal.terminal().equals(terminal));
+        voltagelevels.add(voltageLevel);
```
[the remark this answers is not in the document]
vmouradian: adressed in b933ca87357e36d209f9aeb33130b27d3a6fa623

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AreaImpl.java
```diff
@@ -98,17 +125,42 @@ public void removeVoltageLevel(VoltageLevel voltageLevel) {
...
     @Override
-    public void addBoundaryTerminal(Terminal terminal, boolean ac) {
-        boundaryTerminals.add(new Area.BoundaryTerminal(terminal, ac));
+    public AreaBoundaryAdderImpl newAreaBoundary() {
+        return new AreaBoundaryAdderImpl(this);
+    }
+
+    @Override
+    public void removeAreaBoundary(Terminal terminal) {
+        areaBoundaries.remove(areaBoundaries.stream().filter(b -> Objects.equals(b.getTerminal().orElse(null), terminal)).findFirst().orElse(null));
```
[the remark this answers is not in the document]
vmouradian: adressed in b933ca87357e36d209f9aeb33130b27d3a6fa623

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AreaImpl.java
```diff
@@ -98,17 +125,42 @@ public void removeVoltageLevel(VoltageLevel voltageLevel) {
...
+    }
+
+    @Override
+    public void removeAreaBoundary(Boundary boundary) {
+        areaBoundaries.remove(areaBoundaries.stream().filter(b -> Objects.equals(b.getBoundary().orElse(null), boundary)).findFirst().orElse(null));
```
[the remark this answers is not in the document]
vmouradian: adressed in b933ca87357e36d209f9aeb33130b27d3a6fa623

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
```diff
@@ -255,28 +255,23 @@ public int getCountryCount() {
...
     @Override
-    public Stream<AreaType> getAreaTypeStream() {
-        return index.getAll(AreaTypeImpl.class).stream().map(Function.identity());
-    }
-
-    @Override
-    public AreaType getAreaType(String id) {
-        return index.get(id, AreaTypeImpl.class);
+    public int getAreaTypeCount() {
+        return Iterables.size(getAreaTypes());
```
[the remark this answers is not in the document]
vmouradian: adressed in b933ca87357e36d209f9aeb33130b27d3a6fa623

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/TerminalRefSerDe.java
```diff
@@ -79,16 +79,20 @@ public static Terminal readTerminal(NetworkDeserializerContext context, Network
         return TerminalRefSerDe.resolve(id, side, n);
     }
 
-    public static void readTerminalRef(NetworkDeserializerContext context, Network network, Consumer<Terminal> endTaskTerminalConsumer) {
+    public static void readTerminalRefAttributes(NetworkDeserializerContext context, Network network, Consumer<Terminal> endTaskTerminalConsumer) {
```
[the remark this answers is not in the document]
vmouradian: Thanks to your comment I understood that this method was not really needed. 
I made a small refactor that deletes it.
see 993baaea9c0647aabac94c56e55b720a54ab5582

iidm/iidm-serde/src/test/resources/V1_12/areaRoundTripRef.xml
vmouradian: adressed in e8eaddeddefc0cbe61a1af7eb490f0d23305bc5c

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
vmouradian: Adressed in 77ae1f322f10dae4af6d17219c9a4586e05ca194

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/AreaSerDeTest.java
vmouradian: adressed in 77ae1f322f10dae4af6d17219c9a4586e05ca194

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractSubnetworksExplorationTest.java
```diff
@@ -277,6 +283,28 @@ public void testExploreCountries() {
         assertCollection(List.of(Country.FR, Country.BE, Country.DE), subnetwork2.getCountries());
     }
 
+    @Test
+    public void testExploreAreaTypes() {
+        assertEquals(2, merged.getAreaTypeCount());
+        assertEquals(1, subnetwork1.getAreaTypeCount());
+        assertEquals(1, subnetwork2.getAreaTypeCount());
+        assertCollection(List.of(id("areaType1", ID_1), id("areaType1", ID_2)), merged.getAreaTypeStream().toList());
```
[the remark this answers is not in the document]
vmouradian: adressed in 2796ed468ad8bb5c9ab62e89f3a2163925888bab

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractAreaTest.java
```diff
@@ -0,0 +1,294 @@
...
+        final Function<AreaBoundary, ?> getBoundaryOrTerminalFunction = areaBoundary -> areaBoundary.getTerminal().isPresent() ? areaBoundary.getTerminal().get() : areaBoundary.getBoundary().get();
```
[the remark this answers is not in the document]
vmouradian: adressed in 2796ed468ad8bb5c9ab62e89f3a2163925888bab

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractAreaTest.java
```diff
@@ -0,0 +1,294 @@
...
+    void setUp() {
+        network = EurostagTutorialExample1Factory.createWithAreas();
+        biddingZoneA = network.getArea("bza");
+        biddingZoneB = network.getArea("bzb");
+        regionA = network.getArea("rga");
+        aicA = network.getArea("aic_a");
+    }
+
+    @Test
+    void areaAttributesTest() {
```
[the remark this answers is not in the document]
vmouradian: adressed in 2796ed468ad8bb5c9ab62e89f3a2163925888bab

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractAreaTest.java
vmouradian: The method has been updated to check ids (and sides for terminals) rather that object equalities.
see 2796ed468ad8bb5c9ab62e89f3a2163925888bab
I let it private, but maybe we would still like to make it public ?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AreaImpl.java
```diff
@@ -21,35 +21,57 @@
...
     private final Double acNetInterchangeTarget;
-    private final Double acNetInterchangeTolerance;
-    private final List<Area.BoundaryTerminal> boundaryTerminals;
+    private final List<AreaBoundary> areaBoundaries;
+    private final Set<VoltageLevel> voltagelevels;
+
+    private final class AreaListener extends DefaultNetworkListener {
+        @Override
+        public void beforeRemoval(Identifiable identifiable) {
+            if (identifiable instanceof DanglingLine danglingLine) {
```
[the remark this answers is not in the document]
vmouradian: added in 2c766e3b0791da95a9bee2421ff2c10de4776abe

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AreaImpl.java
```diff
@@ -69,8 +91,18 @@ public Optional<Double> getAcNetInterchangeTarget() {
...
     @Override
-    public Optional<Double> getAcNetInterchangeTolerance() {
-        return Optional.ofNullable(acNetInterchangeTolerance);
+    public Double getAcNetInterchange() {
+        return areaBoundaries.stream().filter(AreaBoundary::isAc).mapToDouble(AreaBoundary::getP).filter(p -> !Double.isNaN(p)).sum();
+    }
+
+    @Override
+    public Double getDcNetInterchange() {
+        return areaBoundaries.stream().filter(areaBoundary -> !areaBoundary.isAc()).mapToDouble(AreaBoundary::getP).filter(p -> !Double.isNaN(p)).sum();
```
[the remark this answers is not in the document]
vmouradian: adressed in 2c766e3b0791da95a9bee2421ff2c10de4776abe

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AreaImpl.java
```diff
@@ -21,35 +21,57 @@
...
+    private final String areaType;
     private final Double acNetInterchangeTarget;
-    private final Double acNetInterchangeTolerance;
-    private final List<Area.BoundaryTerminal> boundaryTerminals;
+    private final List<AreaBoundary> areaBoundaries;
+    private final Set<VoltageLevel> voltagelevels;
+
+    private final class AreaListener extends DefaultNetworkListener {
+        @Override
+        public void beforeRemoval(Identifiable identifiable) {
```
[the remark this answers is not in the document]
vmouradian: adressed in 2c766e3b0791da95a9bee2421ff2c10de4776abe

docs/grid_model/network_subnetwork.md
vmouradian: The doc has been updated in d40f8ba4d8776fff3162f66b8a178e1290cb0456

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AreaImpl.java
```diff
@@ -27,16 +28,19 @@ public class AreaImpl extends AbstractIdentifiable<Area> implements Area {
...
-        public void beforeRemoval(Identifiable identifiable) {
+        public void beforeRemoval(Identifiable<?> identifiable) {
             if (identifiable instanceof DanglingLine danglingLine) {
                 // if dangling line removed from network, remove its boundary from this extension
                 AreaImpl.this.removeAreaBoundary(danglingLine.getBoundary());
             } else if (identifiable instanceof Connectable<?> connectable) {
                 // if connectable removed from network, remove its terminals from this extension
                 connectable.getTerminals().forEach(AreaImpl.this::removeAreaBoundary);
             }
+            // If the identifiable removed is a VoltageLevel: the area voltageLevels attribute will be updated by VoltageLevel::remove()
```
olperr1, line 43: I find this comment difficult to understand. Maybe something like this will be clearer:
```suggestion
            // When a VoltageLevel is removed, its areas' voltageLevels attributes are directly updated. This is not managed via this listener.
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/TerminalRefSerDe.java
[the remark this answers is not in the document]
olperr1: You are right, it is indeed much better like this.

Just one thing: before this PR,
`context.getReader().readEndNode();`
was before
` context.getEndTasks().add(() -> ...`
in `TerminalRefSerDe.readTerminalRef`. Do you need to move it or could it be restored at its previous place?

If it could be moved, maybe could you also move it in `BoundaryRefSerDe.readBoundaryRef` to be consistent.

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractAreaTest.java
olperr1: Thanks for having rewritten this method.
I think you can leave it with this visibility. The checks are only done using String and enum values so it shouldn't depend on the implementation.  :+1:

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Area.java
```diff
@@ -0,0 +1,127 @@
...
+ *     </tbody>
+ * </table>
+ *
+ * @author Marine Guibert {@literal <marine.guibert at artelys.com>}
+ * @author Valentin Mouradian {@literal <valentin.mouradian at artelys.com>}
+ * @see VoltageLevel
+ * @see AreaBoundary
+ * @see AreaAdder
+ */
+public interface Area extends Identifiable<Area> {
```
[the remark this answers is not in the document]
vmouradian: I forgot to remove the listener from the network, fixed here aa3972fcec3253b8c105174a13ae6d83aefa8b25

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/TerminalRefSerDe.java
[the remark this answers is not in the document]
vmouradian: It can be restored, you are rigth.
fixed in f6a9fbf0fc509b121c08b4005ad4e2cbac251fed

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Area.java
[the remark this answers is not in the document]
olperr1: Thanks

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AreaImpl.java
```diff
@@ -0,0 +1,282 @@
...
+        return OptionalDouble.of(target);
+    }
+
+    @Override
+    public Area setAcInterchangeTarget(double acInterchangeTarget) {
+        NetworkImpl n = getNetwork();
+        int variantIndex = n.getVariantIndex();
+        double oldValue = this.acInterchangeTarget.set(variantIndex, acInterchangeTarget);
+        String variantId = n.getVariantManager().getVariantId(variantIndex);
+        n.invalidateValidationLevel();
```
olperr1, line 122: There is no need to invalidate the validation level because there is no check about the AC interchange target regarding the validation level (see `ValidationUtil.validate(…)`.

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractAreaTest.java
```diff
@@ -0,0 +1,465 @@
...
+        } catch (Exception ignored) {
+            // ignore
+        }
+
+        // Remove observer changes
+        network.removeListener(mockedListener);
+    }
+
+    @Test
+    void testGetAreaBoundary() {
```
olperr1, line 185: Could you set the visibility of every test method to public in this class?

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractAreaTest.java
```diff
@@ -0,0 +1,465 @@
...
+
+        // Change values
+        controlAreaA.setAcInterchangeTarget(123.4);
+
+        // Check update notification
+        Mockito.verify(mockedListener, Mockito.times(1))
+                .onUpdate(controlAreaA, "acInterchangeTarget", VariantManagerConstants.INITIAL_VARIANT_ID, Double.NaN, 123.4);
+
+        // At this point
+        // no more changes is taking into account
```
olperr1, line 123:
```suggestion
        // After this point, no more changes are taken into account.
```

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractAreaTest.java
```diff
@@ -0,0 +1,487 @@
...
+
+        // Simulate exception for onUpdate calls
+        Mockito.doThrow(new PowsyblException()).when(mockedListener)
+                .onUpdate(controlAreaA, "acInterchangeTarget", VariantManagerConstants.INITIAL_VARIANT_ID, oldValue, 123.4);
+
+        // Case when same value is set
+        controlAreaA.setAcInterchangeTarget(123.4);
```
olperr1, line 130: Isn't an assertion missing here?

olperr1 (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/AreaAdder.java
```diff
@@ -0,0 +1,64 @@
...
+     */
+    AreaAdder addAreaBoundary(Boundary boundary, boolean ac);
+
+    /**
+     * Build the Area object.
+     * These are the checks that are performed before creating the object :
+     * <p/
+     *      - areaType is not null
+     * @return {@link Battery}
+     */
```
olperr1, line 61:
```suggestion
    /**
     * Build the Area object.
     * <p>These are the checks that are performed before creating the object :</p>
     * <ul>
     *     <li>areaType is not null;</li>
     * </ul>
     * @return {@link Area}
     */
```

olperr1 (changes requested):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AreaImpl.java
jeandemanged: thanks, I wasn't fully aware on how this feature works, indeed doesn't apply here, this is fixed now.

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractAreaTest.java
jeandemanged: done, sorry

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractAreaTest.java
jeandemanged: done

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractAreaTest.java
jeandemanged: the relevant assertion is the one at the very end: mockito's "no interaction wanted".

there was a bug in the test, fixed now, see below comment HERE--->

```java
        // Case when same value is set
        controlAreaA.setAcInterchangeTarget(123.4);
        // Case when no listener is registered
        network.removeListener(mockedListener);
        controlAreaA.setAcInterchangeTarget(456.7); // HERE---> was 123.4 before, so not really testing listener removal, this is fixed now

        // Check no notification
        Mockito.verifyNoMoreInteractions(mockedListener);

```

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractAreaTest.java
```diff
@@ -0,0 +1,489 @@
...
+
+    @Test
+    public void areaInterchangeComputation() {
+        assertEquals(-602.631, controlAreaA.getAcInterchange(), DELTA);
+        assertEquals(0.0, controlAreaA.getDcInterchange());
+        assertEquals(-602.631, controlAreaA.getTotalInterchange(), DELTA);
+
+        assertEquals(+603.563, controlAreaB.getAcInterchange(), DELTA);
+        assertEquals(0.0, controlAreaB.getDcInterchange());
+        assertEquals(+603.563, controlAreaB.getTotalInterchange(), DELTA);
```
jeandemanged, line 204: @olperr1 @annetill @geofjamg 

This is not so nice here, because the two values should be equal of opposite sign.
`Control Area A: -602.631 MW`
`Control Area B: +603.563 MW`

The discrepancy is caused by `EurostagTutorialExample1Factory#createWithTieLines` which introduces a non-zero G for the dangling lines, although `EurostagTutorialExample1Factory#create` has zero G-s for the 2 lines, and the power flow results in here seem to be built using zero G-s.

If I modify `createWithTieLines` with zero G-s for the dangling lines, then I find much better values:
`Control Area A: -602.949 MW`
`Control Area B: +602.944 MW`
Which fits with powerflow tolerances...

Was this G introduction on DanglingLines made on purpose ? Do we need to update the powerflow results specifically for the `createWithTieLines` case ?
Or, should we fix to use zero G-s within this MR or another MR ?

Thanks

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractAreaTest.java
olperr1: OK, that's clear!

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractAreaTest.java
olperr1: The non-zero G-s come from #2599. They were set to avoid problems when executing the tests in Java 17 and in Java >= 19, where the results are not exactly the same.
As you suggested, setting the G-s to zero in your specific test case may be a solution. But it should be tested with Java 17 and >=19 versions.

jeandemanged: Thanks, 
Here what I did in 68dbd546b59ef717b340e9635192129d3dd723b9:
- The previously existing `createWithTieLines` remains unchanged (with non-zero DL G's because of #2599)
- The new `createWithTieLinesAndAreas` introduced in this PR has zero DL G's

Because `createWithTieLinesAndAreas` is not used in any SerDe related test, no issue introduced (tested also locally with Java 19 and Java 21).

olperr1: It's OK for me. Thanks.

olperr1 (approved):

docs/grid_model/network_subnetwork.md
```diff
@@ -95,6 +95,35 @@ When defining the model, the user has to specify how the different equipment con
...
+[![Javadoc](https://img.shields.io/badge/-javadoc-blue.svg)](https://javadoc.io/doc/com.powsybl/powsybl-core/latest/com/powsybl/iidm/network/Area.html)
+
+An Area is a geographical zone of a given type.
+
+An Area is composed of a collection of [voltage levels](#voltage-level), and a collection of area boundaries.
+Area boundaries can be terminals of equipments or `Boundary` objects from [dangling lines](#dangling-line).
+
+The area type is used to distinguish between various area concepts of different granularity. For instance: control areas, bidding zones, countries...
+
+A [voltage level](#voltage-level) can belong to several Areas, as long as all areas are of different type.
```
annetill, line 108: Areas -> areas.

docs/grid_model/network_subnetwork.md
```diff
@@ -95,6 +95,35 @@ When defining the model, the user has to specify how the different equipment con
...
+
+The area boundaries define how interchange are to be calculated for the area.  
+Area interchange is calculated by summing the power flows across the area boundaries and can be obtained for AC part only (considering only AC boundaries),
```
annetill, line 111: active power flows

annetill, line 111: active power flows

docs/grid_model/network_subnetwork.md
m-guibert: Fixed by 73dbdaa761751dc7ed9d97e325ce36b5f2650301

docs/grid_model/network_subnetwork.md
m-guibert: Fixed by 73dbdaa761751dc7ed9d97e325ce36b5f2650301

docs/grid_model/network_subnetwork.md
```diff
@@ -95,6 +95,35 @@ When defining the model, the user has to specify how the different equipment con
...
+Note that this target interchange is for only the AC part of the interchange.
+
+All area interchange values use the load sign convention: positive values indicate that the area is importing, negative values that the area is exporting.
+
+**Characteristics**
+
+| Attribute             | Unit | Description                                                    |
+|-----------------------|------|----------------------------------------------------------------|
+| $AreaType$            |      | To specify the type of Area (eg. ControlArea, BiddingZone ...) |
+| $AcInterchangeTarget$ | MW   | Target AC active power interchange                             |
```
annetill, line 125: My last remark ;-) Here, even if it is not always done in other sections, I would like to know that I can have the list of voltage levels and the list of boundaries.

m-guibert: Thank you for the review! This has been modified in 74a3b3d4d2d17f8644e4943e25f4483a861a929d

m-guibert: Note: Damien suggested we add a second table with the AreaBoundary attributes to help to understand the AreaBoundaries attribute. It has been done in 5f417e15b72a847d5b9afc8508af8e3837e4aeb6

docs/grid_model/network_subnetwork.md
```diff
@@ -99,6 +99,56 @@ When defining the model, the user has to specify how the different equipment con
...
+| $VoltageLevels$       |      | List of voltage levels of the area                             |
+| $AreaBoundaries$      |      | List of area boundaries of the area                            |
+
+**Characteristics of an AreaBoundary**
+
+An area boundary is modelled by an `AreaBoundary` instance.
```
olperr1, line 136: Typo (we use the American English spelling):

```suggestion
An area boundary is modeled by an `AreaBoundary` instance.
```

PR #2956, opened 2024-03-27
Fix reference terminal extension managing cloned variants

zamarrenolm:

Bug fix

An extension for reference terminals does not exist in in the original network, but is added to a cloned variant.
NPE when trying to copy the cloned variant with the extension information to a variant created when it was not yet defined.
Also NPE if we try to clone the initial variant (without extension) to a variant with extension info.

Initializs all variants with empty set of reference terminals instead of null.

- [ ] Yes
- [X] No

**Other information**:

Observed running balance adjustments on powsybl-entsoe repo.


jeandemanged (approved):
Indeed the code was not managing well the use case where the extension is created _after_ the variants.

sorry for this bug, thanks for the fix.

Issue #989, opened 2019-10-24
[DataSource API] Design issues and/or missing documentation


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Design issues, or missing documentation.

* **What is the current behavior?**

The `DataSource` API, which is used in particular in the core `Importer` API, is very hard for third-party users to use and understand.

An original issue seems to be that it is used for 2 different things:
1. A common abstraction over different storages (folder, zip, AFS, memory ...)
2. The representation of a `Network` as a set of binary contents which can be read by an `Importer`. To help for this, it has the concept of "base name" which has nothing to do with item 1. However, the actual resource names are not known by the data source, only by the importer which has the ability to import it.

A consequence of this is that the manipulation of `DataSource` objects as described in item 2 is not convenient, because it's not autonomous.
For example, if you want to simply copy the binary content representing a network from a DataSource to an in-memory `Map<String, byte[]>`, you will need :
  - An `Importer` which knows what files in the DataSource are part of the network representation. If you want to find an adequate `Importer`, you can use `Importers.findImporter` but you need to provide unexpected parameters, for example a `ComputationManager`
  - An in-memory DataSource implementation which exposes the actual map of content it contains (kind of extended `MemDataSource`) : otherwise, at the end you don't know what files have been actually copied, only the `Importer` knows.

Or for example you could want to copy the contents to an in-memory zip. Once again, you need to provide your own DataSource implementation, and this is not obvious since the interface is not documented.

* **What is the expected behavior?**

The `DataSource` API is used a lot in the framework, and should be documented, and simple to understand and use for third-party users.

Possible guidelines for a clearer API would be to separate the 2 concerns described above into 2 different objects or interfaces. One would be in charge of abstracting away the storage, while the other one would carry the actual set of binary contents which serve for the `Network` representation.

* **What is the motivation / use case for changing the behavior?**

This central part of the API really feels like a blocker for adoption of the framework by third-party users.

* **Please tell us about your environment:**
  - PowSyBl Version: 3.0.0

sylvlecl:
**References**
 * issue #121 describes problems with data source extension handling
 * cancelled PR #471 tried to adress some of the problems described here or in #121 
In particular, this PR removed the extension handling from `DataSource`, leaving it to the `Importer` API, but left an optional "main file name" in place of the "base name" as part of the object.
It also externalized the handling of compression formats in a dedicated interface.
[...]

Issue #121, opened 2017-11-24
DataSourceUtil extension handling


sylvlecl:
The handling of file extensions in DataSources is not satisfying today for several reasons.
The current mechanism, based on the creation of a DataSource based on a file name:

1/ cannot handle successive compression extensions (.gz.zip ...), although it makes not much sense...

2/ does not take into account the extension defined at the time of creation, but only at the time of writing the output

3/ deletes everything after the first dot in the name when creating the DataSource, even though it may not be part of the actual extension. It makes it impossible, for instance to read from a file "toto.tata.xiidm.gz" passed to the command line

**The DataSourceUtil and DataSource implementations need to be refactored to cope with those problems.**

mathbagu:
4/ if a folder contains two files toto.uct and toto.UCT with different contents, it's impossible to load toto.UCT because the UcteImporter will always load the first matching extension. This assessment is true with all the importers which support several extensions.

mathbagu:
This bug will be solved with the datasource refactoring #471

alicecaron:
Obsolete since the DataSource refacto

PR #2959, opened 2024-03-28
Datasources

rolnico:

No

Bug fix + feature

_to be completed_

Two types of `DataSources` now exists:

- `DirectoryDataSource` (inherited by `GzDataSource`, `Bzip2DataSource`, `ZstdDataSource`, `XZDataSource`): considers files in a directory
- `AbstractArchiveDataSource` (`ZipDataSource`, `TarDataSource`): considers files in an archive

Directory DataSources consider the following parameters:

- A directory (where the files are located)
- A base name (start of the file names to consider)
- A source format (end of the file names to consider, excluding the compression format)
- A compression format (files to use through the datasource should have this format)
- An observer

Archive DataSources consider the following parameters:

- A directory (where the archive is located)
- A base name (start of the file names to consider)
- A source format (end of the file names to consider, excluding the compression format)
- An archive format (format of the archive: `zip` or `tar`)
- A compression format (compression of the archive: `zip`, `gz`, `xz`, etc.)
- The archive file
- An observer

The parameter `sourceFormat` corresponds to the extension of the files the users want to consider in their datasources. For example, if `sourceFormat == ".xiidm"`, only files ending with `".xiidm"` will be considered. The value can be anything. If the parameter is empty, it won't be considered in the file listing.

The method `listNames(String)` now works like this:

- For `DirectoryDataSource`, it lists the files located in the directory, starting with the base name, having the same source format and the same compression format as the datasource and respecting the regex given as parameter
- For Archive DataSources, it lists the files located in the archive starting with the base name, having the same source format as the datasource and respecting the regex given as parameter

A `DataSourceBuilder` is now provided in addition to usual `DataSourceUtil.createDataSource` methods. 

_Note:_ when creating a datasource by giving a filename, for example via `DataSourceUtil.createDataSource(Path directory, String fileName, DataSourceObserver observer)`, the different parameters are extracted from the filename. If the source format is empty or not a usual one (list defined in the new `FileInformation` class), a warning will be raised.

- [x] Yes
- [ ] No

Two types of `DataSources` now exists:

- `DirectoryDataSource` (inherited by `GzDataSource`, `Bzip2DataSource`, `ZstdDataSource`, `XZDataSource`): considers files in a directory
- `AbstractArchiveDataSource` (`ZipDataSource`, `TarDataSource`): considers files in an archive

Directory DataSources consider the following parameters:

- A directory (where the files are located)
- A base name (start of the file names to consider)
- A source format (end of the file names to consider, excluding the compression format)
- A compression format (files to use through the datasource should have this format)
- An observer

Archive DataSources consider the following parameters:

- A directory (where the archive is located)
- A base name (start of the file names to consider)
- A source format (end of the file names to consider, excluding the compression format)
- An archive format (format of the archive: `zip` or `tar`)
- A compression format (compression of the archive: `zip`, `gz`, `xz`, etc.)
- The archive file
- An observer

The parameter `sourceFormat` corresponds to the extension of the files the users want to consider in their datasources. For example, if `sourceFormat == ".xiidm"`, only files ending with `".xiidm"` will be considered. The value can be anything. If the parameter is empty, it won't be considered in the file listing.

Therefore, multiple classes have changed, please update your code if you used them directly:
| Old classes | New classes     |
|----------|------------- |
| `Bzip2FileDataSource` | `Bzip2DataSource`     |
| `FileDataSource` | `DirectoryDataSource`     |
| `GzFileDataSource` | `GzDataSource`     |
|   | `TarDataSource`     |
| `XZFileDataSource` | `XZDataSource`     |
| `ZipFileDataSource` | `ZipDataSource`     |
| `ZstdFileDataSource` | `ZstdDataSource`     |

*Note:* Their constructors are not the same so you will have to adapt your code to the new ones, depending on which DataSource you use.

`ReadOnlyDataSource`, `ResourceDataSource` and `MultipleReadOnlyDataSource` now extend `AbstractReadOnlyDataSource`.

The methods `DataSourceUtil.createDataSource()` have been reworked to be more consistent with each other and to provide more variations depending on the parameters you might want to use. Be careful regarding the parameters you use:
- If you use only one `String` parameter, it will be considered as a `fileName`
- If you use two, they will be considered as a `baseName` and a `sourceFormat`
- If you use three, they will be considered as an `archiveName`, a `baseName` and a `sourceFormat`

<br/>

For instance, `new FileDataSource(path, "fileName")` could be replaced by one of the followings:
- `new DirectoryDataSource(path, "fileName", "")`;
- `DataSourceUtil.createDataSource(path, "fileName", "")`.

**Other information**:


rolnico:
Should solve  #121 (except successive compressions)

rolnico:
Might solve #989 with some more documentation (will be done once the modifications are verified and accepted)

zamarrenolm:
One common usage is to create a datasource based on a path and a basename. After the refactoring, the simplest way to build it is to use `DataSourceUtil.createDataSource(path, "", baseName)`. I am ok with reserving the overloaded `createDataSource(Path,String)` for path + filename, but I wonder if we could have something like `createDataSourceForBasename(Path,String)` to avoid having to pass an empty string to the version with three parameters.

To reduce the length of these method names, we could remove the `DataSource` suffix.

```java
DataSourceUtil.create(path, filename)
DataSourceUtil.createForBasename(path, basename)
...
```

zamarrenolm:
> One common usage is to create a datasource based on a path and a basename. After the refactoring, the simplest way to build it is to use `DataSourceUtil.createDataSource(path, "", baseName)`. I am ok with reserving the overloaded `createDataSource(Path,String)` for path + filename, but I wonder if we could have something like `createDataSourceForBasename(Path,String)` to avoid having to pass an empty string to the version with three parameters.
> 
> To reduce the length of these method names, we could remove the `DataSource` suffix.
> 
> ```java
> DataSourceUtil.create(path, filename)
> DataSourceUtil.createForBasename(path, basename)
> ...
> ```

This use case corresponds to the import of multiple CGMES instance files: the data source is specified as a folder plus a basename (the prefix common to all CGMES instance file filenames). All instance files should be XML.

To avoid the potential problem of allowing an empty extension, in this use case we could ask only for files with ".xml" extension.

commons/src/main/java/com/powsybl/commons/datasource/DataSourceBuilder.java
```diff
@@ -0,0 +1,94 @@
...
+                new TarDataSource(directory, archiveFileName, baseName, compressionFormat, sourceFormat, observer);
+        } else if (compressionFormat == null) {
+            return new DirectoryDataSource(directory, baseName, sourceFormat, observer);
+        } else {
+            return switch (compressionFormat) {
+                case BZIP2 -> new Bzip2DataSource(directory, baseName, sourceFormat, observer);
+                case GZIP -> new GzDataSource(directory, baseName, sourceFormat, observer);
+                case XZ -> new XZDataSource(directory, baseName, sourceFormat, observer);
+                case ZSTD -> new ZstdDataSource(directory, baseName, sourceFormat, observer);
+                default -> new DirectoryDataSource(directory, baseName, sourceFormat, observer);
```
jonenst, line 90: since compressionFormat==null is already checked above, this is for unknown compression format ? Maybe add a warning or throw ?

commons/src/main/java/com/powsybl/commons/datasource/FileInformation.java
```diff
@@ -0,0 +1,102 @@
...
+        // Last dot index
+        currentDotIndex = fileNameWithoutCompressionExtension.lastIndexOf('.');
+
+        // Archive extension
+        String fileNameWithoutCompressionNorArchive;
+        if (compressionFormat == CompressionFormat.ZIP) {
+            archiveFormat = ArchiveFormat.ZIP;
+            fileNameWithoutCompressionNorArchive = fileNameWithoutCompressionExtension;
+        } else {
+            archiveFormat = !fileNameWithoutCompressionExtension.substring(currentDotIndex + 1).isEmpty() && "tar".contains(fileNameWithoutCompressionExtension.substring(currentDotIndex + 1)) ? ArchiveFormat.TAR : null;
```
jonenst, line 66: why "tar".contains ? "foo.ta" or "foo.t"  will be accepted but shouldn't ?

commons/src/main/java/com/powsybl/commons/datasource/MultipleReadOnlyDataSource.java
```diff
@@ -41,41 +41,41 @@ public String getBaseName() {
...
         });
     }
 
     @Override
-    public InputStream newInputStream(String suffix, String ext) throws IOException {
+    public InputStream newInputStream(String suffix, String ext, boolean checkConsistencyWithDataSource) throws IOException {
         for (var dataSource : dataSources) {
-            if (dataSource.exists(suffix, ext)) {
+            if (dataSource.exists(suffix, ext, checkConsistencyWithDataSource)) {
                 return dataSource.newInputStream(suffix, ext);
```
jonenst, line 69: should use the same strategy as GenericReadOnlyDataSource. After the exists check, either pass checkConsistencyWithDataSource to newInputStream  or not

commons/src/main/java/com/powsybl/commons/datasource/MultipleReadOnlyDataSource.java
```diff
@@ -41,41 +41,41 @@ public String getBaseName() {
...
         }
         return null;
     }
 
     @Override
-    public InputStream newInputStream(String fileName) throws IOException {
+    public InputStream newInputStream(String fileName, boolean checkConsistencyWithDataSource) throws IOException {
         for (var dataSource : dataSources) {
-            if (dataSource.exists(fileName)) {
+            if (dataSource.exists(fileName, checkConsistencyWithDataSource)) {
```
jonenst, line 78: same as https://github.com/powsybl/powsybl-core/pull/2959/files#r1613391360

commons/src/main/java/com/powsybl/commons/datasource/ReadOnlyMemDataSource.java
```diff
@@ -63,24 +64,27 @@ public String getBaseName() {
...
+    public boolean exists(String fileName, boolean checkConsistencyWithDataSource) throws IOException {
         Objects.requireNonNull(fileName);
-        return data.containsKey(fileName);
+        return (!checkConsistencyWithDataSource || isConsistentWithDataSource(fileName)) && data.containsKey(fileName);
     }
 
     @Override
-    public InputStream newInputStream(String suffix, String ext) throws IOException {
+    public InputStream newInputStream(String suffix, String ext, boolean checkConsistencyWithDataSource) throws IOException {
         return newInputStream(DataSourceUtil.getFileName(baseName, suffix, ext));
```
jonenst, line 79: should pass checkConsistencyWithDataSource ?

commons/src/main/java/com/powsybl/commons/datasource/TarDataSource.java
```diff
@@ -0,0 +1,316 @@
...
+     * {@inheritDoc}
+     *
+     * <p>Files are here located in the archive.</p>
+     */
+    @Override
+    public Set<String> listNames(String regex) throws IOException {
+        // Initialize variables
+        Pattern p = Pattern.compile(regex);
+        Set<String> names = new HashSet<>();
+        Path zipFilePath = getArchiveFilePath();
```
jonenst, line 60: rename variable "zip" but this is "tar" ?

commons/src/main/java/com/powsybl/commons/datasource/TarDataSource.java
```diff
@@ -0,0 +1,316 @@
...
+            throw new UnsupportedOperationException("append not supported in tar file data source");
+        }
+        Path tarFilePath = getArchiveFilePath();
+        OutputStream os = new TarEntryOutputStream(tarFilePath, fileName, compressionFormat);
+        return observer != null ? new ObservableOutputStream(os, tarFilePath + ":" + fileName, observer) : os;
+    }
+
+    @Override
+    public InputStream newInputStream(String suffix, String ext, boolean checkConsistencyWithDataSource) throws IOException {
+        return newInputStream(DataSourceUtil.getFileName(baseName, suffix, ext));
```
jonenst, line 149: should pass checkConsistencyWithDataSource ?

commons/src/main/java/com/powsybl/commons/datasource/DataSourceBuilder.java
rolnico: added a warning

commons/src/main/java/com/powsybl/commons/datasource/FileInformation.java
rolnico: switched it with equals

commons/src/main/java/com/powsybl/commons/datasource/MultipleReadOnlyDataSource.java
rolnico: done

commons/src/main/java/com/powsybl/commons/datasource/ReadOnlyMemDataSource.java
rolnico: fixed

commons/src/main/java/com/powsybl/commons/datasource/TarDataSource.java
rolnico: fixed

commons/src/main/java/com/powsybl/commons/datasource/TarDataSource.java
rolnico: fixed

commons/src/test/java/com/powsybl/commons/datasource/TarArchiveDataSourceTest.java
```diff
@@ -0,0 +1,238 @@
...
+            default -> os;
+        };
+    }
+
+    protected void createArchiveAndFiles(String fileName) throws IOException {
+
+        // File information
+        FileInformation fileInformation = new FileInformation(fileName);
+
+        // Create the Zip archive and add the files
```
jonenst, line 164: Zip -> Tar

rolnico:
With the last modifications, using the following command in a folder containing the files from this archive [MicroGrid.zip](https://github.com/user-attachments/files/16013768/MicroGrid.zip) should create a new json file with the network from `MicroGrid.xml` (so it does not use the CGMES importer):
`itools convert-network --input-file MicroGrid.xml --output-format JIIDM --output-file test.json`

flo-dup:
PR splitted in parts (see #3101, #3102, #3103)

PR #2963, opened 2024-04-02
Launch CI when new version tag is pushed

flo-dup:

No

Feature

CI is often not launched on version commit, as the "prepare release" pull request contains two commits which are often pushed at once. The CI is then launched only at the latest commit ("Bump to X.Y.Z-SNAPSHOT version") and not the the first one ("Bump to X.Y.Z version")

The CI is launched when the added tag "vX.Y.Z" is pushed

- [ ] Yes
- [x] No

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -7,7 +7,7 @@ on:
       - 'release-v**'
       - 'full-sonar-analysis-**'
     tags:
-      - 'v*.*.*'
+      - 'v[0-9]+.[0-9]+.[0-9]+*'
   pull_request:
 
 permissions: {}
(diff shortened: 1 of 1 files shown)


.github/workflows/maven.yml
```diff
@@ -6,6 +6,8 @@ on:
       - 'main'
       - 'release-v**'
       - 'full-sonar-analysis-**'
+    tags:
+      - 'v*.*.*'
```
olperr1, line 10: We should maybe use a more restrictive rule, such as:

```suggestion
      - 'v[0-9]+.[0-9]+.[0-9]+*'
```

flo-dup: I checked whether that syntax is supported before, and it seems so! https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

flo-dup: I copied it from powsybl-distribution, maybe we could backport that there

olperr1 (approved):

PR #2969, opened 2024-04-08
Fix LoadFlowResult status for retro-compatibility and replace uses of deprecated isOk

rolnico:

No

Bug fix

When initializing a new `LoadFlowResultImpl(boolean ok, Map<String, String> metrics, String logs)` with `ok` set to `true`, since no componentResults is given, the status is still set to FAILED. Therefore, there is an incoherence between `isOk` and `status`.
This case appears for example when deserializing an old loadflow result where no component result is defined.

If no component result is given during the initialisation of a new `LoadFlowResultImpl` and if `ok` is set to `true`, then `status` is set to `DEFAULT_OK_STATUS_FOR_RETROCOMPATIBILITY`, which has been defined equals to `PARTIALLY_CONVERGED`.

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/action-ial/action-ial-simulator/src/main/java/com/powsybl/action/ial/simulator/loadflow/LoadFlowActionSimulator.java
+++ b/action-ial/action-ial-simulator/src/main/java/com/powsybl/action/ial/simulator/loadflow/LoadFlowActionSimulator.java
@@ -305,7 +305,7 @@ public class LoadFlowActionSimulator implements ActionSimulator {
         } catch (Exception e) {
             throw new PowsyblException(e);
         }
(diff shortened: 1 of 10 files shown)


rolnico:
> I don't think this is a good id to add a new value just for compatibility. It just cluter the api for few added value.Is there another solution ?

It's not really a new value: `DEFAULT_OK_STATUS_FOR_RETROCOMPATIBILITY` is only an internal (and private) variable equals to `Status.PARTIALLY_CONVERGED`

rolnico:
This PR is just about choosing a middle ground between `ok` set to true (so the loadflow should have converged) and no result being given (so, according to our implementation, it should be set to `Status.FAILED)

PR #2987, opened 2024-04-24
CGMES unit tests for transformer conversion

zamarrenolm:

Bug fix

Unit tests for transformer conversion are looking for specific transformers in the network by id. Hardcoded ids with a prefixed underscore are used.
But current conversion uses CGMES mRIDs as identifiers, that do not use prefix underscore.
As a result, transformers are not found in the network.
The tests tried to compare expected and actual flows in terminals of transformers. When the transformer is not found, the flows are initialized to `NaN` values.
The comparison between expected and actual flows does not take into account potential `NaN` values.

The right identifiers are used in the unit tests.
The comparison between expected and actual flows has been fixed to fail if any of the compared values is `NaN`.

- [ ] Yes
- [X] No

### DIFF
--- a/cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/TransformerConversionTest.java
+++ b/cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/TransformerConversionTest.java
@@ -559,7 +559,11 @@ class TransformerConversionTest {
             case SPLIT:
                 t3wtSplitShuntAdmittance = true;
         }
-        boolean twtSplitShuntAdmittance;
+         if (t2wtSplitShuntAdmittance != t3wtSplitShuntAdmittance) {
+            throw new PowsyblException(String.format("Unexpected SplitShuntAdmittance configuration %s %s",
+                t2wtSplitShuntAdmittance, t3wtSplitShuntAdmittance));
+        }
+        boolean twtSplitShuntAdmittance = t2wtSplitShuntAdmittance;
         if (!t2wtSplitShuntAdmittance && !t3wtSplitShuntAdmittance) {
             twtSplitShuntAdmittance = false;
         } else if (t2wtSplitShuntAdmittance && t3wtSplitShuntAdmittance) {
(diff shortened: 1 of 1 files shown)


cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/TransformerConversionTest.java
```diff
@@ -521,26 +519,22 @@ private static T3xFlow threeWindingsTransformerFlow(ThreeWindingsTransformer twt
...
-            Math.abs(expected.p2 - actual.p2) > tol ||
-            Math.abs(expected.q2 - actual.q2) > tol) {
-            return false;
-        }
-        return true;
+        // Comparison fails if actual has NaN values
+        return Math.abs(expected.p1 - actual.p1) < tol &&
+                Math.abs(expected.q1 - actual.q1) < tol &&
+                Math.abs(expected.p2 - actual.p2) < tol &&
+                Math.abs(expected.q2 - actual.q2) < tol;
```
rolnico, line 526: Shouldn't it be `<=` ?

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/TransformerConversionTest.java
```diff
@@ -565,7 +559,7 @@ private LoadFlowParameters defineLoadflowParameters(LoadFlowParameters loadFlowP
             case SPLIT:
                 t3wtSplitShuntAdmittance = true;
         }
-        boolean twtSplitShuntAdmittance = false;
+        boolean twtSplitShuntAdmittance;
```
rolnico, line 562: You can simplify it even more:
```suggestion
         if (t2wtSplitShuntAdmittance != t3wtSplitShuntAdmittance) {
            throw new PowsyblException(String.format("Unexpected SplitShuntAdmittance configuration %s %s",
                t2wtSplitShuntAdmittance, t3wtSplitShuntAdmittance));
        }
        boolean twtSplitShuntAdmittance = t2wtSplitShuntAdmittance;
```

rolnico (changes requested):

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/TransformerConversionTest.java
zamarrenolm: yes, although it is not really relevant here

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/TransformerConversionTest.java
zamarrenolm: yes, thanks!

rolnico (approved):

PR #2989, opened 2024-04-26
Connect transformers during generator connection

WalAmeni:

During the shift of GeneratorScalable, the generator should be connected before targetP update. Is some cases, the generator is connected to the main network with a transformer. We should connect this transformer so that the generator connection is effective and taken into account by loadflow computation.
If even connecting the transformer, the generator remains disconnected, we recommend reverting the modification and leaving the generator at targetP = 0. This way, the shift will be made with other scalable generators. 

This feature solves a real case problem encountered when shifting Spain during capacity calculation process.

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/scalable/GeneratorScalable.java
+++ b/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/scalable/GeneratorScalable.java
@@ -123,8 +123,14 @@ class GeneratorScalable extends AbstractInjectionScalable {
         Terminal t = g.getTerminal();
         if (!t.isConnected()) {
             if (parameters.isReconnect()) {
-                new ConnectGenerator(g.getId()).apply(n);
                 LOGGER.info("Connecting {}", g.getId());
+                new ConnectGenerator(g.getId()).apply(n);
+                if (!t.isConnected() || !g.getTerminal().getBusBreakerView().getConnectableBus().isInMainConnectedComponent()) {
+                    //If the generator still disconnected, we should not change the active power setPoint
+                    LOGGER.info("Generator {} could not be connected, discarded from scaling", g.getId());
+                    return 0.;
+                }
+
             } else {
                 LOGGER.info("Generator {} is not connected, discarded from scaling", g.getId());
                 return 0.;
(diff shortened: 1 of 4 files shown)


iidm/iidm-modification/src/test/java/com/powsybl/iidm/modification/ConnectGeneratorTest.java
```diff
@@ -101,4 +107,67 @@ void testConnectGeneratorWithNoNetworkInformation() {
...
+        });
+    }
+
+    private void checkInitalStatusTerminalDisconnected(Network network, List<String> ids) {
+        ids.forEach(id -> {
+            Generator gen = network.getGenerator(id);
+            assertFalse(gen.getTerminal().isConnected());
+            assertTrue(gen.getTerminal().getBusBreakerView().getConnectableBus().isInMainSynchronousComponent());
+        });
+    }
```
rolnico, line 171: You might want to replace those 4 methods with one:
```suggestion
    private void checkTerminalStatus(Network network, List<String> ids, boolean shouldBeConnected, boolean shouldBeInMainSynchronousComponent) {
        ids.forEach(id -> {
            Generator gen = network.getGenerator(id);
            assertEquals(shouldBeConnected, gen.getTerminal().isConnected());
            assertEquals(shouldBeInMainSynchronousComponent, gen.getTerminal().getBusBreakerView().getConnectableBus().isInMainSynchronousComponent());
        });
    }
```
and fix the call to this method in the test

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/scalable/GeneratorScalable.java
```diff
@@ -123,8 +123,14 @@ public double scale(Network n, double asked, ScalingParameters parameters) {
...
             if (parameters.isReconnect()) {
-                new ConnectGenerator(g.getId()).apply(n);
                 LOGGER.info("Connecting {}", g.getId());
+                new ConnectGenerator(g.getId()).apply(n);
+                if (!t.isConnected() || !g.getTerminal().getBusBreakerView().getConnectableBus().isInMainConnectedComponent()) {
+                    //If the generator still disconnected, we should not change the active power setPoint
+                    LOGGER.info("Generator {} could not be connected, discarded from scaling", g.getId());
+                    return 0.;
+                }
+
```
rolnico, line 133: Please extract this part and the few surrounding lines in a new method to reduce the method complexity and avoid a brain overload issue

rolnico (changes requested):
In addition to the various comments, could you add some more tests to cover the few remaining lines? It should be feasible to reach 100%, seeing what is missing.

rolnico:
Additional question: you managed the TwoWindingsTransformers in this PR but should the ThreeWindingsTransformers be also reconnected the same way?

flo-dup (changes requested):
The way of detecting the transformers which connect the generator to the network seem very specific to the usecase. It's not clear to me how to make it more generic, we could add the switch part I mention, but what about three-winding transformers, what if the generator is behind two (or more) transformers in line, should we connect all transformers as for the unit test case?

PR #2990, opened 2024-04-26
Change the reduction's type from float to double in the LimitViolation builder

EtienneLt:

no

Quality (type harmonisation)

The `LimitViolation` builder now takes a `double` instead of a `float` for the reduction as the `reduction` attribute in `LimitViolation` is a `double`.

- [ ] Yes
- [x] No


olperr1 (changes requested):
Can you also modify `LimitViolationBuilderTest.buildCurrentViolation()` to use doubles instead of floats? It would be cleaner.

olperr1 (approved):

PR #2927, opened 2024-03-12
CGMES export. Handle CGM export dependencies

zamarrenolm:
**Please check if the PR fulfills these requirements**

- [X] The commit message follows our guidelines
- [X] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**Does this PR already have an issue describing the problem?**

**What kind of change does this PR introduce?**

Feature.

**What is the current behavior?**

CGMES export only considers IGM (Individual Grid Model) dependencies when exporting.

**What is the new behavior (if this is a feature change)?**
CGMES export will consider the dependencies required for CGM (Common Grid Model) export.

A Merging Agent using PowSyBl will export first an updated SSH for each IGM,
that should contain a `Supersedes` reference to the IGM’s original SSH CIMXML file.

Then, the Merging Agent should export a single SV for the whole CGM. This SV file must contain a `DependentOn` reference to each IGM’s updated SSH. The power flow of a CGM is calculated without topology processing. No new TP CIMXML files are created. This means that the CGM SV file must contain a `DependentOn` reference to each IGM’s original TP.

**Does this PR introduce a breaking change or deprecate an API?**
- [ ] Yes
- [ ] No

**If yes, please check if the following requirements are fulfilled**

- [ ] The *Breaking Change* or *Deprecated* label has been added
- [ ] The migration steps are described in the following section

**What changes might users need to make in their application due to this PR? (migration steps)**

**Other information**:

This change is based on #2898

PR #2991, opened 2024-04-26
CGMES export: fix regression on external explicit SV dependencies

zamarrenolm:

Bug fix

Some users of the PowSyBl framework were setting explicitly the dependencies for CGMES SV exported file through properties stored in the network.
The recent refactoring of the CGMES metadata models (#2898) introduced a regression in this feature. 

The feature for explicitly setting SV dependencies through network properties is kept. 
It has been implemented for legacy support. The users of this feature do not need to change their code.
In the future, it is expected that users of CGMES CGM SV export rely directly on the CGMES CGM export that is being developed in #2927.

- [ ] Yes
- [X] No

**Other information**:


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
```diff
@@ -124,10 +125,23 @@ public void updateDependencies() {
                 .clearDependencies()
                 .addDependentOn(eqModelId);
 
-        getExportedSVModel()
-                .clearDependencies()
-                .addDependentOn(getExportedTPModel().getId())
-                .addDependentOn(getExportedSSHModel().getId());
+        getExportedSVModel().clearDependencies();
+        List<String> tpIds = legacyIdsForSvDependencies.get(CgmesSubset.TOPOLOGY);
```
annetill, line 129: I understand your solution but I was thinking of something different:
- No properties anymore ;
- We keep information in DependentOn fields ;
- I was thinking about a simple parameter at export that will just say if we export using the extension entirely defined by the user in SV part, meaning that we trust the user. If not, we are allowed to clear the dependencies of SV in extension and to use the other models' dependencies (TP and SSH).

zamarrenolm: That was the second option we discussed, but it will require some changes in the code where the issue was raised. I first wanted to provide a fix that did not required changes in the code already written by the users.

Additionally, building the extension for SV model entirely from scratch requires the user to know some information that now is managed by the CGMES export in a transparent way (mainly, the profile URLs, that are different for each version of CGMES exported).

zamarrenolm: Anyway, I am already preparing a PR that lets the user give the metadata models as inputs, as you suggest.

annetill: So, with the option relying on a parameter, we must keep the properties?

zamarrenolm: #2993 implements your proposal about respecting the info about dependencies present in the extension. It requires that the external user builds the metadata model in the extension. An example of this, following the same coding scheme seen in the software that raised the issue, can be found in the unit test `LegacyCommonGridModelExportTest::buildSvDependenciesManaginMetadataModels`.

I would keep (for now) the two ways of solving the problem:

- Accept updated dependencies through the legacy mechanism of properties (so the external users do not have to change their code).
- Let the user manage dependencies through metadata model (with an example provided in our code).

Both PRs can be merged  without conflicts. #2993 is based on this one and can be merged after it.

annetill: Properties management should be removed for next version, here is only to ease transition.

annetill (approved):
Properties management should be removed for next version (2024.2.0).

PR #2992, opened 2024-04-26
Fix CGMES GL import

flo-dup:

No

Bug fix

GL import requires the `CoordinateSystem` to have `crsName` equal to `"WGS84"`, whereas this field is optional

GL import does not query for that parameter (and therefore does not check it)

- [ ] Yes
- [x] No


zamarrenolm (approved):

zamarrenolm:
the name can be kept for export ...

annetill (approved):

flo-dup:
> the name can be kept for export ...

Yes indeed, we could keep the name for export. This requires some more work though, as in your commit you're writing it whatever the input name was. Is that really worth the effort?

PR #2993, opened 2024-04-29
CGMES export. Allow external metadata model definition. Flag to disable update of dependencies

zamarrenolm:

Feature

When an export is made, metadata model dependencies are always updated.

Let the user specify if export dependencies must be updated or not based on a parameter `iidm.export.cgmes.update-dependencies`, `True` by default. If it is `True`, dependencies will be updated automatically during the export. `False` means that the user has already put them in the extension for metadata models and they should not be updated.

- [ ] Yes
- [X] No

**Other information**:

After this change the user can build the metadata model that must be used in the export. The dependencies in the exported files will not be automatically prepared by the export module. The dependencies that the user has put in the metadata model present in the network extension will be used.


cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/LegacyCommonGridModelExportTest.java
```diff
@@ -0,0 +1,145 @@
...
+        String exportBasename = "tmp-micro-bc-from-CGM";
+        network.write("CGMES", exportParams, tmpDir.resolve(exportBasename));
+
+        Set<String> deps = findAll(REGEX_DEPENDENT_ON, Files.readString(tmpDir.resolve(exportBasename + "_SV.xml")));
+        System.out.println("dependencies : " + Arrays.toString(deps.toArray()));
+        assertEquals(Set.of("ssh-updated-dep1", "ssh-updated-dep2", "ssh-updated-dep3", "tp-initial-dep1", "tp-initial-dep2", "tp-initial-dep3"), deps);
+    }
+
+    @Test
+    void testExportCgmSvDependenciesOnMetadataModelsExtension() throws IOException {
```
annetill, line 61: This way never relies on properties, right?

zamarrenolm: Yes, this test uses only metadata models. No properties are used or required.

WalAmeni (approved):
The dependentOn are correctly exported by adding the CgmesMetadataModels extension

annetill (approved):
The default value of the new parameter could be false.

PR #2994, opened 2024-04-29
Add side in FeederResult

colinepiloquet:

No

Fix

In case of FeederResult on a branch, there is no way to know if it is on side 1 or 2.

Side is added as a variable.

- [ ] Yes
- [x] No


shortcircuit-api/src/main/java/com/powsybl/shortcircuit/AbstractFeederResult.java
```diff
@@ -15,13 +18,27 @@
...
+    protected AbstractFeederResult(String connectableId, ThreeSides side) {
         this.connectableId = Objects.requireNonNull(connectableId);
+        this.side = side;
     }
 
     public String getConnectableId() {
         return connectableId;
     }
 
+    public ThreeSides getSide() {
```
flo-dup, line 36: Could you add the missing `@Override` for the 3 methods? it should be seen as a code smell from sonar cloud, but I noticed lately it's not. Maybe because the class is abstract?
https://sonarcloud.io/organizations/powsybl-ci-github/rules?activation=true&open=java%3AS1161&q=override&qprofile=AYonwccxzxfrDz7ASaP1

flo-dup (approved):

PR #3006, opened 2024-05-13
CGMES export equipment fix: equivalent injection base voltage reference

zamarrenolm:

bug fix

When an IIDM generator is exported to CGMES EQ file as an equivalent injection the reference to its base voltage is wrong. Instead of the identifier of a base voltage, the identifier of a voltage level is exported.
This only happens for equivalent injections not located inside boundaries.

The correct identifier for the equivalent injection base voltage is exported in the CGMES EQ file.

- [ ] Yes
- [X] No


annetill (approved):
Thanks!

PR #3011, opened 2024-05-14
ReportNode V2.1

flo-dup:

No

Feature / bug fix

- ReportNode v2.0
- context of included ReportNode is not replaced by context of new root ReportNode
- included ReportNode is silently replacing messageTemplate if key conflicts occur
- Deserialization loads the whole file before deserializing it
- No timestamps (except if added manually)
- messageTemplate needed to construct ReportNodes

- ReportNode v2.1
- context of included ReportNode is replaced by context of new root ReportNode through `RefChain`
- if messageTemplate key conflicts occur when including ReportNode, templates are not overridden, + warn log
- Deserializes the JSON reportNode file while streaming it
- Timestamp value added for each ReportNode with a specific pattern if enabled at root ReportNode creation
- messageTemplate can be missing at ReportNode construction (to anticipate the dictionaries questions which will later arise)

- [x] Yes
- [ ] No

- Starting from this release, the reportNode JSON files from API v2.0 cannot be read anymore. Please switch to JSON files generated with API V2.1.
- `ReportConstants.REPORT_SEVERITY_KEY` has been renamed into `ReportConstants.SEVERITY_KEY`


commons/src/main/java/com/powsybl/commons/report/ReportNodeBuilder.java
```diff
@@ -7,12 +7,31 @@
...
  * @author Florian Dupuy {@literal <florian.dupuy at rte-france.com>}
  */
 public interface ReportNodeBuilder extends ReportNodeAdderOrBuilder<ReportNodeBuilder> {
+
+    /**
+     * Enable/disable timestamps on build ReportNode and all descendants
+     * @param timestampPattern: the pattern to use for the timestamp (see {@link DateTimeFormatter#ofPattern}), or null to disable the timestamp
+     * @return a reference to this object
+     */
+    ReportNodeBuilder withTimestamps(String timestampPattern);
```
geofjamg, line 24: I would have add 2 methods, one for activation or not the timestamped logging and another to change the default date pattern.

commons/src/test/java/com/powsybl/commons/report/ReportNodeTest.java
```diff
@@ -12,15 +12,76 @@
...
-import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.*;
 
 /**
  * @author Florian Dupuy {@literal <florian.dupuy at rte-france.com>}
  */
 class ReportNodeTest extends AbstractSerDeTest {
 
+    @Test
+    void testValues() throws IOException {
```
geofjamg, line 28: I do not see any unit test for timestamps. Maybe it is because hard to compare to a reference? Maybe just searching for the right nodes and matching the date with a pattern would be enough?

flo-dup: that's more or less what I'm doing in unit test `ReportNodeTest::testTimestamps`. I don't check that the pattern is the right one though. I'll add that!

flo-dup: Well I was checking the pattern in fact, but with the default iso-like from java.time.format...

geofjamg (approved):

PR #3019, opened 2024-05-16
TCK: Remove reference comparisons for Ratio/PhaseTapChangers

AbdelHedhili:

changes TCK test

There are tests that compare Ratio/PhaseTapChanger by reference and expect that we keep the same reference for these objects

Remove these checks that are implementation dependent

- [ ] Yes
- [X] No


iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractThreeWindingsTransformerTest.java
```diff
@@ -115,11 +115,9 @@ public void baseTests() {
         leg3.setRatedS(1.3);
         assertEquals(1.3, leg3.getRatedS(), 0.0);
 
-        RatioTapChanger ratioTapChangerInLeg1 = createRatioTapChanger(leg1,
-            transformer.getTerminal(ThreeSides.ONE));
+        createRatioTapChanger(leg1, transformer.getTerminal(ThreeSides.ONE));
```
rolnico, line 118: Instead of deleting the existing tests, it might be better to make them optional by adding a new parameter `public static boolean shouldTestOnIdentity = true;` to the TCK test classes and to add a setup method to modify it in the implementation test classes if you don't want to do those tests
```
    @BeforeAll
    static void setup() {
        shouldTestOnIdentity = false;
    }
 ```

rolnico (changes requested):

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractThreeWindingsTransformerTest.java
AbdelHedhili: I don't really agree on this, it isn't really the goal of a TCK to test implementation dependent details no ?  
- Either the reference comparison is something that is needed for the api to work as expected and so it must be tested in the TCK 
- Or it's not something that is needed and is just implementation dependent (if i'm not mistaken it's the case here) and so it shouldn't be in the TCK 
That's what I understand of it anyway. 
So should we really make your asked changes ?

AbdelHedhili: Ok it makes sense. The modifications seems better since no matter the implementation it's supposed to work. Should I go for that then ? Do you agree @rolnico ?

rolnico: go for it :)

rolnico (approved):

PR #3020, opened 2024-05-16
Remove order constraint for streams in tck

AbdelHedhili:

Modify TCK test

When retrieving equipment from a network with streams we check for a specific order in the TCK.

Only check that the content is the same, we don't care about the order

- [X] Yes
- [ ] No

Custom IIDM implementation maintainers should add the following dependency in their TCK implementation's `pom.xml`:

```xml
<dependency>
    <groupId>org.assertj</groupId>
    <artifactId>assertj-core</artifactId>
    <scope>test</scope>
</dependency>
```

**Other information**:


iidm/iidm-tck/pom.xml
```diff
@@ -54,6 +54,11 @@
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
```
olperr1, line 61: You also need to add this dependency in the `powsybl-iidm-impl`'s `pom.xml` for the tests to succeed.

olperr1 (changes requested):

olperr1 (approved):

PR #3022, opened 2024-05-17
TCK: Remove reference comparisons for CurrentLimits

AbdelHedhili:

changes TCK test

There are tests that compare currentLimits by reference and expect that we keep the same reference for these objects

Remove these checks that are implementation dependent

- [ ] Yes
- [X] No


iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractLineTest.java
```diff
@@ -114,26 +114,22 @@ public void baseAcLineTests() {
         acLine.setB2(b2);
         assertEquals(b2, acLine.getB2(), 0.0);
 
-        CurrentLimits currentLimits1 = acLine.newCurrentLimits1()
+        acLine.newCurrentLimits1()
```
rolnico, line 117: Instead of deleting the existing tests, it might be better to make them optional by adding a new parameter `public static boolean shouldTestOnIdentity = true;` to the TCK test classes and to add a setup method to modify it in the implementation test classes if you don't want to do those tests
```
    @BeforeAll
    static void setup() {
        shouldTestOnIdentity = false;
    }
 ```
(same as in the other PR)

rolnico (changes requested):

rolnico (approved):

PR #3029, opened 2024-05-21
CGMES export retained switch with same TN at both ends as non-retained

zamarrenolm:

Fix for export. Required to comply with rule SwitchTN1 of ENTSO-E QoCDC.

A retained switch may have the same bus/breaker view buses at both ends. For example: a retained coupler joining two busbars that are also connected through the closed disconnectors of a line or other equipment in scheme similar to this one:
Screenshot 2024-05-21 at 19 28 28

Here the COUPLER joins busbar sections BBS0 and BBS1 and is marked as retained.
But the busbar sections can also be connected through disconnectors LOAD_DIS0 and LOAD_DIS1, when both are closed. In this situation, both ends of the COUPLER are the same bus, in the bus/breaker view topology level.

When the network is exported to CGMES, bus/breaker view buses are used to define Topological Nodes (TNs).

But the ENTSO-E QoCDC rule SwitchTN1 states that TNs at ends of retained switches must be different.

We fix the export to CGMES unmarking the COUPLER as retained.
It is not clear how to re-define the Topological Nodes in these configurations to avoid the QoCDC rule being raised.

- [ ] Yes
- [X] No


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
```diff
@@ -189,11 +189,21 @@ private static void writeSwitches(Network network, String cimNamespace, XMLStrea
...
+                // To ensure we do not violate rule SwitchTN1 of ENTSO-E QoCDC,
+                // we only export as retained a switch if it will be exported with different TNs at both ends
+                boolean exportAsRetained = sw.isRetained() && hasDifferentTNsAtBothEnds(sw);
+                SwitchEq.write(context.getNamingStrategy().getCgmesId(sw), sw.getNameOrId(), switchType, sw.getKind(), context.getNamingStrategy().getCgmesId(vl), sw.isOpen(), exportAsRetained, cimNamespace, writer, context);
             }
         }
     }
 
+    private static boolean hasDifferentTNsAtBothEnds(Switch sw) {
+        // The exported Topological Nodes como from IIDM bus/breaker view buses
```
annetill, line 201: como ? :-)

annetill (approved):
Thanks a lot!

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
zamarrenolm: "come" :-)
funny typo

colinepiloquet (approved):
I have tested in my use case and it solved the SwitchTN1 issue.

PR #3030, opened 2024-05-22
Refactoring of newRootReportNode from ReportNode interface

caioluke:

Bug (incoherence) fix

Currently, the ReportNode interface allows the implementation of different reporters. However, ReportNode's static method [newRootReportNode ](https://github.com/powsybl/powsybl-core/blob/main/commons/src/main/java/com/powsybl/commons/report/ReportNode.java#L61) defines a default ReportNode builder (ReportNodeImpl) that cannot be overriden.

newRootReportNode is a method of the class ReportNode, therefore different types of ReportNodes can be created. Also, the builders now have a constructor.

- [X] Yes
- [ ] No

Instead of calling the previous static method from ReportNode, one would have to create an instance of a ReportNodeRootBuilder, e.g ReportNodeRootBuilderImpl, and then build.

Before:
```java
ReportNode root = ReportNode.newRootReportNode().build();
```

After:
```java
ReportNode root = new ReportNodeRootBuilderImpl().build();
```


jeandemanged:
Won't fix - Custom implementation of ReportNode is not a supported use case.

PR #3037, opened 2024-05-28
Fix: the fictitious load SvInjection should refer to TopologicalNode UUID

colinepiloquet:

No

Bug fix

When writing a SvInjection for a fictitious load, the Id of the bus is used for the topological node reference, without regards to the naming strategy.

The topological node reference is the one of the bus and takes into account the naming strategy.

- [ ] Yes
- [x] No


zamarrenolm:
> @zamarrenolm, @annetill, I am not 100% sure my fix is correct but I am 100% that there is a problem here. I exported a file with the `cgmes` naming strategy and for the SvInjection associated to a fictitious load, the topological node reference was not a UUID.

You are totally right, there was a missing call to the naming strategy when exporting the topological node of the `SvInjection`.

zamarrenolm:
Updated the unit test to verify that the exported SV file contains a proper reference to a topological node that is present in the exported TP file.

zamarrenolm (approved):

Issue #3026, opened 2024-05-19
Network reducer


annetill:
### Describe the current behavior

![image](https://github.com/powsybl/powsybl-core/assets/48123713/b962f209-ec11-4c11-8da9-d3d3f5015dfd)

And no reduce method for tie lines...

PR #3038, opened 2024-05-29
Tie lines in network reduction

geofjamg:

#3026 

Bug fix

When network reduction impact the removal of one of a tie line voltage levels, we remove the tie line (so we un-merge the dangling lines). 

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/iidm/iidm-reducer/src/test/java/com/powsybl/iidm/reducer/TieLineReductionTest.java
+++ b/iidm/iidm-reducer/src/test/java/com/powsybl/iidm/reducer/TieLineReductionTest.java
@@ -10,6 +10,7 @@ package com.powsybl.iidm.reducer;
 import com.powsybl.iidm.network.Identifiable;
 import com.powsybl.iidm.network.Network;
 import com.powsybl.iidm.network.test.EurostagTutorialExample1Factory;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import java.util.List;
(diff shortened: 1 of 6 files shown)


iidm/iidm-reducer/src/test/java/com/powsybl/iidm/reducer/TieLineReductionTest.java
```diff
@@ -0,0 +1,38 @@
...
+ * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
+ */
+class TieLineReductionTest {
+
+    @Test
+    void test() {
+        var observer = new NetworkReducerObserverImpl();
+        Network network = EurostagTutorialExample1Factory.createWithTieLine();
+        var reducer = NetworkReducer.builder()
+                .withNetworkPredicate(new IdentifierNetworkPredicate(List.of("VLGEN", "VLHV1", "VLLOAD")))
```
flo-dup, line 29: To improve coverage we could also add within this test a reducer with following predicates:
for a  "doubly" deleted tieline
 ```java
new IdentifierNetworkPredicate(List.of("VLHV1", "VLHV2"));
```
for a  non-deleted tielines
```java
new IdentifierNetworkPredicate(List.of("VLGEN"));
```

geofjamg: for first one that would be a really fake test as it would break this simple network into part with just 2 voltage levels remaining with a load and a generator

flo-dup: No, not exactly (my comment was misleading you, as I mixed up things...).
- The first predicate should lead to two voltage levels with a busbar, linked with two tielines. Without any other equipment.
- The second one is a bit strange yes as it should lead to a single voltage level with a generator. 
 
Both cases are not stranger than current unit test, which breaks the network into 2 parts with the voltage level VLLOAD left alone. And even if the unit tests are strange they are still checking that the reduction works as expected.

geofjamg: Done

iidm/iidm-reducer/src/test/java/com/powsybl/iidm/reducer/TieLineReductionTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @BeforeEach
+    void setUp() {
+        network = EurostagTutorialExample1Factory.createWithTieLine();
+        observer = new NetworkReducerObserverImpl();
+    }
+
+    @Test
+    void testOneSideTieLineReduction() {
+        Network network = EurostagTutorialExample1Factory.createWithTieLine();
```
flo-dup, line 37: You can delete this line
```suggestion
```

flo-dup (changes requested):

flo-dup (approved):

PR #3044, opened 2024-05-30
CGMES export: output max number values as float, not double

zamarrenolm:

Bug fix

IIDM uses Java Double for floating point numbers.
CGMES CIMXML defines its floating point numbers as `xsd:float`.
Regular values do not represent a problem, but extreme values (`+/-Double.MAX_VALUE`) are reported as errors during validation.

To fix the error in validation we decide to output `+/-Float.MAX_VALUE`.
We decide to write this instead of the `INF` or `-INF` literals, to avoid potential problems with the interpretation of that string values.


colinepiloquet (approved):
I tested

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportUtil.java
```diff
@@ -70,7 +70,9 @@ public static String format(double value) {
     public static String format(double value, double defaultValue) {
         // Always use scientific format for extreme values
         if (value == Double.MAX_VALUE || value == -Double.MAX_VALUE) {
-            return scientificFormat(value, defaultValue);
+            // CIMXML expects xsd:float values
+            float value1 = value == Double.MAX_VALUE ? Float.MAX_VALUE : -Float.MAX_VALUE;
+            return scientificFormat(value1, defaultValue);
         }
```
jeandemanged, line 76: not sure it would work as intended for values between float max and double max ?

zamarrenolm: Values greater than float max are not expected in "regular" calculations of a power grid.
This change is only to adapt the extreme values used as an equivalent for "infinity" in PowSyBl, that considers its basic floating point type as `double` to the CIMXML standard, that relies on `float`.

jeandemanged: I fully agree that what is made here will work 99.9999999999999999999999999...999999999% of the time.

But in CGMES export we may redefine 'infinity' to be anything exceeding +/- `Float.MAX_VALUE`, instead of just being equality with +/-`Double.MAX_VALUE`.

But not blocking for me, can be merged as is.

jeandemanged (approved):

PR #3045, opened 2024-05-31
Documentation : load flow and load flow validation

annetill:

No.

Just a migration from github.io.

- [ ] Yes
- [ ] No

**Other information**:


docs/README.md
```diff
@@ -5,7 +5,7 @@ They are published on powsybl.readthedocs.io/ and pull requests are built and pr
 
 To build the docs locally, run the following commands:
 ~~~bash
-pip install -r docs/requirements.txt
```
annetill, line 8: I don't know what is the right thing to put here.

docs/README.md
```diff
@@ -5,7 +5,7 @@ They are published on powsybl.readthedocs.io/ and pull requests are built and pr
...
-sphinx-build -a docs ./build-docs
+pip install -r requirements.txt
```
So-Fras, line 8:
```suggestion
pip install -r docs/requirements.txt
```
Leave it as it was!

docs/README.md
```diff
@@ -5,7 +5,7 @@ They are published on powsybl.readthedocs.io/ and pull requests are built and pr
...
+sphinx-build -a . ../build-docs
```
So-Fras, line 9:
```suggestion
sphinx-build -a docs ./build-docs
```
Same here

docs/README.md
```diff
@@ -5,7 +5,7 @@ They are published on powsybl.readthedocs.io/ and pull requests are built and pr
...
 ~~~
-Then open `build-docs/index.html` in your browser.
\ No newline at end of file
+Then open `../build-docs/index.html` in your browser.
```
So-Fras, line 11:
```suggestion
Then open `build-docs/index.html` in your browser.
```
Same here.

I think that the reason you add problems launching your commands locally is that you were in the "wrong" directory (i.e. the `docs` directory instead of being in `powsybl-core`).
Should we explain that better?

docs/README.md
annetill: Why? The README is located in repositiory docs/.

docs/README.md
annetill: The issue is the location of your README.md file for me.

docs/README.md
annetill: image

docs/README.md
So-Fras: The README.md is about the documentation so I think its location makes sense.

Maybe we can add the sentence you can find in the powsybl-open-rao docs/README.md to make it less ambiguous:

When modifying the website content, you can easily preview the result on your PC **by navigating to the root of the project** and running:

pip install -r docs/requirements.txt
sphinx-build -a docs ./build-docs

Then open build-docs/index.html in your browser.

So-Fras: Ok, now I get the problem when running the commands on an IDE.

I will amend the README.md so that the commands can be run from an IDE too.

So-Fras (approved):

docs/README.md
```diff
@@ -3,9 +3,27 @@ These are the documentation sources for PowSybl core features.
 Please keep them up to date with your developments.  
 They are published on powsybl.readthedocs.io/ and pull requests are built and previewed automatically.
 
-To build the docs locally, run the following commands:
+When modifying the website content, you can easily preview the result on your PC.
+
+**First option: running commands from a terminal**
+
+Navigate to the root of the project and run:
```
rolnico, line 10:
```suggestion
**First option - in a terminal, navigate to the root of the project and run the following commands:**
```

docs/README.md
```diff
@@ -3,9 +3,27 @@ These are the documentation sources for PowSybl core features.
...
+
 ~~~bash
 pip install -r docs/requirements.txt
 sphinx-build -a docs ./build-docs
 ~~~
-Then open `build-docs/index.html` in your browser.
\ No newline at end of file
+
+**Second option: running commands from your IDE HMI**
```
rolnico, line 17:
```suggestion
**Second option - run the following commands directly from your IDE GUI**
```

PR #3050, opened 2024-06-03
Consider sides when applying limitReductions with IdentifiableCriterion

olperr1:

No

Bug fix

When a `LimitReduction` is defined using an `IdentifiableCriterion` (to define a reduction on a voltage level for instance), the reduction is applied on each side of the network element if one of its sides verify the criterion.

For `IdentifiableCriterion`, the reduction is applied only on the side(s) verifying the criterion.

- [ ] Yes
- [X] No

**Other information**:


iidm/iidm-criteria/src/main/java/com/powsybl/iidm/criteria/AtLeastOneCountryCriterion.java
```diff
@@ -55,8 +62,10 @@ private List<Country> getCountriesToCheck(Identifiable<?> identifiable, Identifi
         }
     }
 
-    private List<Country> getCountriesToCheck(NetworkElement networkElement) {
-        return Arrays.asList(networkElement.getCountry1().orElse(null), networkElement.getCountry2().orElse(null));
+    private List<Country> getCountriesToCheck(NetworkElement networkElement, ThreeSides side) {
+        return side != null ?
+                Collections.singletonList(networkElement.getCountry(side).orElse(null)) :
+                Arrays.asList(networkElement.getCountry1().orElse(null), networkElement.getCountry2().orElse(null));
```
flo-dup, line 68: Even if it doesn't really make sense with current equipments, shouldn't we add `getCountry3`? Because your `NetworkElementTest` for instance leads to unexpected behaviour: the `Country.BE` is ignored by this criterion.

iidm/iidm-criteria/src/main/java/com/powsybl/iidm/criteria/AtLeastOneCountryCriterion.java
flo-dup, line 1: The AtLeast name is a bit unexpected now that there is this sides consideration.

flo-dup (changes requested):

iidm/iidm-criteria/src/main/java/com/powsybl/iidm/criteria/AtLeastOneCountryCriterion.java
olperr1: You are right. I forgot to add it when I introduced `getCountry3()` in `NetworkElement`.

flo-dup (approved):

PR #3054, opened 2024-06-03
Add the ability to override minP and maxP to reflect generator temporary limits

vidaldid-rte:

Some generators have limit stricter than pMin and pMax for compensation operations or other short term active power control operations.
In order to capture these limites, the ActivePowerControl extension can store optional values for minPOverride and maxPOverride. 

New optional fields in the ActivePowerControl extension 

- [ ] Yes
- [X] No

In order to take into account the overriden pMin and pMax, applications should check the value of the extension.
If the overridden limits are provided they should be used (instead of minP or maxP) for short term active power control operations.

**Other information**:


iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/ActivePowerControl.java
```diff
@@ -54,4 +56,25 @@ default String getName() {
      */
     void setParticipationFactor(double participationFactor);
 
+    /**
+     * @return if present, provides the overridden value of pmin to be used for active power control operations.
```
annetill, line 60: Maybe you should be more precise here in the Javadoc to explain that it is used "only" during slack distribution if performed on generators ?

docs/grid_model/extensions.md
```diff
@@ -28,7 +30,7 @@ generator.newExtension(ActivePowerControlAdder.class)
     .add();
 ```
 
-The participation status and the participation factor are multi-variants: they can vary from one variant to another.
+The participation status, the participation factor, the maxP override and the minP override are multi-variants: they can vary from one variant to another.
```
annetill, line 33: I think that maxP and minP are not in the variant, why do you need them to be in the variant?

annetill: In general, in documentation we must be homogeneous: in the same table, we write Pmax and maxP.

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/ActivePowerControl.java
```diff
@@ -54,4 +56,25 @@ default String getName() {
...
+    Optional<Double> getMinPOverride();
+
+    /**
+     * Sets the overridden minimal active power.
+     * @param pMinOverride  The overridden value of Pmin. A Nan value removes the override.
+     */
+    void setMinPOverride(double pMinOverride);
+
+    /**
+     * @return if present, provides the overridden value of pmax to be used for active power control operations.
```
annetill, line 71: pmax -> Pmax

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/ActivePowerControl.java
annetill: pmin -> Pmin

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/ActivePowerControlImpl.java
```diff
@@ -26,22 +27,38 @@ public class ActivePowerControlImpl<T extends Injection<T>> extends AbstractMult
...
+                                  double participationFactor,
+                                  double minPOverride,
+                                  double maxPOverride) {
         super(component);
         int variantArraySize = getVariantManagerHolder().getVariantManager().getVariantArraySize();
         this.participate = new TBooleanArrayList(variantArraySize);
         this.droop = new TDoubleArrayList(variantArraySize);
         this.participationFactor = new TDoubleArrayList(variantArraySize);
-        this.allTDoubleArrayLists = List.of(this.droop, this.participationFactor);
+        this.minPOverride = new TDoubleArrayList(variantArraySize);
```
annetill, line 53: Why do you need a variant dependency?

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/extensions/ActivePowerControlSerDe.java
```diff
@@ -45,11 +45,12 @@ public ActivePowerControlSerDe() {
                         .put(IidmVersion.V_1_9, ImmutableSortedSet.of("1.0", "1.1"))
                         .put(IidmVersion.V_1_10, ImmutableSortedSet.of("1.0", "1.1"))
                         .put(IidmVersion.V_1_11, ImmutableSortedSet.of("1.0", "1.1"))
-                        .put(IidmVersion.V_1_12, ImmutableSortedSet.of("1.0", "1.1"))
+                        .put(IidmVersion.V_1_12, ImmutableSortedSet.of("1.0", "1.1", "1.2"))
                         .build(),
                 new ImmutableMap.Builder<String, String>()
                         .put("1.0", "http://www.itesla_project.eu/schema/iidm/ext/active_power_control/1_0")
                         .put("1.1", "http://www.powsybl.org/schema/iidm/ext/active_power_control/1_1")
+                        .put("1.2", "http://www.powsybl.org/schema/iidm/ext/active_power_control/1_2")
```
annetill, line 53: @flo-dup can you check serialization for me please?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/ActivePowerControlImpl.java
vidaldid-rte: @annetill 
Well... Why not ? You may imagine scenarios where you play with the compensation margin.

But the reality is that I discover now the concept of variants possible for some fields and not others. A strange concept.
Especially if it can vary in the fields of a same extension. Is this concept driven by real needs or by the memory heavy implementation of variants ?

Anyway, let me know what you want. My intention was juts to be homogenous with the existing local code and I had not strong reflections on variants.

annetill: Because it is not with minP and maxP as attributes of a generator, so it is not homogeneous.

docs/grid_model/extensions.md
vidaldid-rte: @annetill  this pMax/maxP story is indeed confusing.
I used the OLF convention (generator.setMaxP). The droop definition seems to use the ADN convention (pMax).
Let me know which one should be corrected (the droop definition or the use of the API convention fro the override)

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/ActivePowerControl.java
vidaldid-rte: well, I corrected with maxP / minP, using the Generator API convention as well as the java bean convention. Hope it works.

annetill: Just in doc !

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/ActivePowerControl.java
vidaldid-rte: I used minP to be like the API.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/ActivePowerControlImpl.java
vidaldid-rte: Let's discuss. I do not like the homogenous argument.
First because, making decision based on consistency leads nowhere.
Second because we have the choice between two inconsisencies. In the extension or vs the pMax/pMin of the generator.
I'd prefer to discuss based on need and tradeoff to make because of the high cost of current variant implementation.
Let's discuss on Monday.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/extensions/ActivePowerControlXmlTest.java
```diff
@@ -44,13 +46,36 @@ public void setUp() throws IOException {
...
+        assertNotNull(activePowerControl1);
+        assertEquals(100, activePowerControl1.getMaxPOverride().get());
+        assertTrue(activePowerControl1.getMinPOverride().isEmpty());
+
+        Battery bat2 = network2.getBattery("BAT");
+        assertNotNull(bat2);
+        ActivePowerControl<Battery> activePowerControl2 = bat2.getExtension(ActivePowerControl.class);
+        assertNotNull(activePowerControl2);
+        assertTrue(activePowerControl2.getMaxPOverride().isEmpty());
+        assertEquals(10, activePowerControl2.getMinPOverride().get());
```
flo-dup, line 66: To avoid the `get()` which might be seen as a code smell you could do so:
```suggestion
        assertEquals(OptionalDouble.of(10), activePowerControl2.getMinPOverride());
```

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/ActivePowerControl.java
vidaldid-rte: we agreed on minP with Anne, for the doc, after reviewing the doc.

docs/grid_model/extensions.md
vidaldid-rte: For variants - no strong argument for or against (use case may exist / but memory overhead) but we agreed to keep it in team meeting.
For doc, as said earlier, we agreed to keep the javaBean convention in the doc and use maxP / minP when talking about these quantities.

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/ActivePowerControl.java
vidaldid-rte: maxP / minP used in the doc as agreed.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/ActivePowerControlImpl.java
```diff
@@ -94,8 +111,29 @@ public void deleteVariantArrayElement(int index) {
     public void allocateVariantArrayElement(int[] indexes, int sourceIndex) {
         for (int index : indexes) {
             participate.set(index, participate.get(sourceIndex));
-            droop.set(index, droop.get(sourceIndex));
             allTDoubleArrayLists.forEach(dl -> dl.set(index, dl.get(sourceIndex)));
         }
     }
+
+    @Override
+    public Optional<Double> getMinPOverride() {
```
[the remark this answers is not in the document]
vidaldid-rte: Done. Still unsure why and when OptionalDouble should be used instead of Optional<Double>. Seems to be a less convenient version. Any advantage ?

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/extensions/ActivePowerControlXmlTest.java
vidaldid-rte: Done. But don't see a strong smell difference, especially in an assert. Makes the expected more complex to me.

flo-dup: if empty the result of the test is a bit clearer that way

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/ActivePowerControlImpl.java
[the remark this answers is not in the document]
flo-dup: less convenient indeed. only advantage is that there's no boxing

PR #3061, opened 2024-06-07
Allow zero value for limits in loading limits

nao1345678:

No

Feature

You could not add permanent or temporary limits to 0 (for `LoadingLimits`).

It is now allowed to set a value equal to zero in a `LoadingLimits` (for permanent and temporary limits).

- [x] Yes
- [ ] No

This is not really a breaking change, but you should be aware that `LoadingLimits` can now be imported/created with a zero value for permanent or temporary limits.
As a consequence, you can now have a `Double.NaN` if you divide something by one of these limits. 

For custom IIDM implementation maintainers:
You should now allow 0 as a value for permanent and temporary limits.


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -562,9 +562,9 @@ private static ValidationLevel checkPermanentLimit(Validable validable, double p
             throwExceptionOrLogError(validable, "permanent limit must be defined if temporary limits are present", throwException, reportNode);
             validationLevel = ValidationLevel.min(validationLevel, ValidationLevel.EQUIPMENT);
         }
-        if (!Double.isNaN(permanentLimit) && permanentLimit <= 0) {
+        if (!Double.isNaN(permanentLimit) && permanentLimit < 0) {
```
annetill, line 565: Maybe a log if zero?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractLoadingLimitsAdder.java
```diff
@@ -74,8 +74,8 @@ public B endTemporaryLimit() {
             if (Double.isNaN(value)) {
                 throw new ValidationException(validable, "temporary limit value is not set");
             }
-            if (value <= 0) {
-                throw new ValidationException(validable, "temporary limit value must be > 0");
```
annetill, line 78: Same.

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractCurrentLimitsTest.java
```diff
@@ -678,4 +679,21 @@ public void testAdderSetPermanentLimitWithInfiniteDurationValue() {
...
+                .beginTemporaryLimit()
+                    .setName("TEST")
+                    .setAcceptableDuration(Integer.MAX_VALUE)
+                    .setValue(0)
+                .endTemporaryLimit();
+        adder.add();
+        Optional<CurrentLimits> optionalLimits = line.getCurrentLimits(TwoSides.ONE);
+        assertTrue(optionalLimits.isPresent());
+        CurrentLimits limits = optionalLimits.get();
+        assertEquals(0, limits.getTemporaryLimit(Integer.MAX_VALUE).getValue());
```
olperr1, line 697: In your second commit, you have removed the assertion on the permanent limit's value.
Could you restore it?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -562,10 +562,15 @@ private static ValidationLevel checkPermanentLimit(Validable validable, double p
...
             // because it is forbidden for SSH and EQ validation levels.
-            throw new ValidationException(validable, "permanent limit must be > 0");
+            throw new ValidationException(validable, "permanent limit must be >= 0");
         }
+        if (!Double.isNaN(permanentLimit) && permanentLimit == 0) {
+            // log if zero
+            LOGGER.info("permanent limit is set to 0");
```
olperr1, line 571: Can you also add the prefix here?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractLoadingLimitsAdder.java
```diff
@@ -74,8 +78,11 @@ public B endTemporaryLimit() {
...
+            if (value < 0) {
+                throw new ValidationException(validable, "temporary limit value must be >= 0");
+            }
+            if (value == 0) {
+                LOGGER.info("temporary limit value is set to 0");
```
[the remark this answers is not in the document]
nao1345678: With `LOGGER.info` the program doesn't pass the tests if I put `validable` as a prefix

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractLoadingLimitsAdder.java
```diff
@@ -31,7 +30,7 @@ abstract class AbstractLoadingLimitsAdder<L extends LoadingLimits, A extends Loa
 
     public class TemporaryLimitAdderImpl<B extends LoadingLimitsAdder<L, B>> implements TemporaryLimitAdder<B> {
 
-        private static final Logger LOGGER = LoggerFactory.getLogger(ValidationUtil.class);
+        private final Logger logger = LoggerFactory.getLogger(TemporaryLimitAdderImpl.class);
```
flo-dup, line 33: Why not static anymore?

nao1345678: Sonar's rule stated I should use `TemporaryLimitAdderImpl` instead of `ValidationUtil` for the logger because it is its current class, but when I did I got this error : 

`[ERROR] /home/elcheikhnal/projects/powsybl-core/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractLoadingLimitsAdder.java:[33,70] non-static class com.powsybl.iidm.network.impl.AbstractLoadingLimitsAdder.TemporaryLimitAdderImpl cannot be referenced from a static context
`

olperr1: It is better to define the logger as a static attribute instead of as an instance attribute because there is no need to create a new one each time we want to add a new `TemporaryLimit`.
I suggest to declare it as a static attribute of `AbstractLoadingLimitsAdder` (and not in the inner class).

```java
private static final Logger LOGGER = LoggerFactory.getLogger(AbstractLoadingLimitsAdder.class);
```

This should solve both problems.

olperr1 (approved):

PR #3062, opened 2024-06-10
BatteryShortCircuit extension

colinepiloquet:

- [ ] Yes
- [ ] No

**Other information**:


iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/AbstractShortCircuit.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+package com.powsybl.iidm.network.extensions;
+
+import com.powsybl.commons.extensions.Extendable;
+import com.powsybl.commons.extensions.Extension;
+
+/**
+ * @author Coline Piloquet {@literal <coline.piloquet@rte-france.fr>}
+ */
+public interface AbstractShortCircuit<T extends Extendable<T>> extends Extension<T> {
```
olperr1, line 16: It may be misleading to have an interface name starting with `Abstract` since we reserve it for abstract classes.
Can you rename it (as `ShortCircuitExtension` for instance)?

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/AbstractShortCircuitAdder.java
```diff
@@ -0,0 +1,30 @@
...
+
+import com.powsybl.commons.PowsyblException;
+import com.powsybl.commons.extensions.Extendable;
+import com.powsybl.commons.extensions.ExtensionAdder;
+
+/**
+ *
+ * @author Coline Piloquet {@literal <coline.piloquet@rte-france.fr>}
+ */
+public interface AbstractShortCircuitAdder<T extends Extendable<T>, C extends AbstractShortCircuit<T>> extends ExtensionAdder<T, C> {
```
olperr1, line 18: Idem here, the `Abstract` in the name is misleading.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/AbstractShortCircuitImpl.java
```diff
@@ -0,0 +1,70 @@
...
+        this.stepUpTransformerX = stepUpTransformerX;
+    }
+
+    @Override
+    public double getDirectSubtransX() {
+        return directSubtransX;
+    }
+
+    @Override
+    public AbstractShortCircuitImpl<T> setDirectSubtransX(double directSubtransX) {
```
olperr1, line 41: It would be better to have the implementation type as a generic parameter of the class in order to return an well-cast object here instead of an object typed with the abstract class.

iidm/iidm-extensions/src/main/java/com/powsybl/iidm/network/extensions/AbstractShortCircuit.java
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * Get the direct-axis sub-transient reactance (also known as X''d)
+     */
+    double getDirectSubtransX();
+
+    /**
+     * Set the direct-axis sub-transient reactance (also known as X''d)
+     */
+    AbstractShortCircuit<T> setDirectSubtransX(double directSubtransX);
```
olperr1, line 26: Using the implementation class as a generic parameter of the interface, it is possible to have the real type as return type of these setters.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/BatteryShortCircuitAdderImpl.java
```diff
@@ -0,0 +1,61 @@
...
+import com.powsybl.commons.extensions.AbstractExtensionAdder;
+import com.powsybl.iidm.network.Battery;
+import com.powsybl.iidm.network.extensions.BatteryShortCircuit;
+import com.powsybl.iidm.network.extensions.BatteryShortCircuitAdder;
+
+/**
+ *
+ * @author Coline Piloquet {@literal <coline.piloquet@rte-france.fr>}
+ */
+public class BatteryShortCircuitAdderImpl extends AbstractExtensionAdder<Battery, BatteryShortCircuit>
```
olperr1, line 20: It is possible to share this code with `GeneratorShortCircuitAdderImpl`.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/extensions/BatteryShortCircuitXmlSerDeTest.java
```diff
@@ -0,0 +1,53 @@
...
+        Network network2 = allFormatsRoundTripTest(network, "/shortcircuits/batteryShortCircuitRef.xml");
+
+        Battery bat2 = network2.getBattery("BAT");
+        assertNotNull(bat2);
+        BatteryShortCircuit batteryShortCircuits2 = bat2.getExtension(BatteryShortCircuit.class);
+        assertNotNull(batteryShortCircuits2);
+
+        assertEquals(batteryShortCircuits.getDirectTransX(), batteryShortCircuits2.getDirectTransX(), 0f);
+        assertEquals(batteryShortCircuits.getDirectSubtransX(), batteryShortCircuits2.getDirectSubtransX(), 0f);
+        assertEquals(batteryShortCircuits.getStepUpTransformerX(), batteryShortCircuits2.getStepUpTransformerX(), 0f);
```
olperr1, line 51: I don't think you'll have a problem here since `1.0` can be precisely stored in binary, but it is a good practice to use a non null value for the delta (such as `0.001d`).

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/extensions/BatteryShortCircuitXmlSerDeTest.java
```diff
@@ -0,0 +1,53 @@
...
+    @Test
+    void test() throws IOException {
+        Network network = BatteryNetworkFactory.create();
+        network.setCaseDate(ZonedDateTime.parse("2017-06-25T17:43:00.000+01:00"));
+        Battery bat = network.getBattery("BAT");
+        assertNotNull(bat);
+        bat.newExtension(BatteryShortCircuitAdder.class)
+            .withDirectTransX(1.0)
+            .withDirectSubtransX(1.0)
+            .withStepUpTransformerX(1.0)
```
olperr1, line 38: Could you use different values for all the attributes in order to check that they are not mixed up when serialized?

olperr1 (changes requested):

olperr1 (approved):

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/extensions/BatteryShortCircuitSerDe.java
```diff
@@ -8,43 +8,96 @@
...
+                        .put(IidmVersion.V_1_7, ImmutableSortedSet.of("0.1"))
+                        .put(IidmVersion.V_1_8, ImmutableSortedSet.of("0.1"))
+                        .put(IidmVersion.V_1_9, ImmutableSortedSet.of("0.1"))
+                        .put(IidmVersion.V_1_10, ImmutableSortedSet.of("0.1"))
+                        .put(IidmVersion.V_1_11, ImmutableSortedSet.of("0.1"))
+                        .put(IidmVersion.V_1_12, ImmutableSortedSet.of("0.1"))
+                        .put(IidmVersion.V_1_13, ImmutableSortedSet.of("0.1", "1.0"))
+                        .build(),
+                new ImmutableMap.Builder<String, String>()
+                        .put("0.1", "http://www.itesla_project.eu/schema/iidm/ext/battery_short_circuits/1_0")
```
flo-dup, line 52: As the namespace URI gives a 1.0 reference, shouldn't we call this the 1.0, and the other one 1.1 instead? Or, another way could be to call it something like 1.0-legacy?

flo-dup (approved):

PR #3065, opened 2024-06-18
Support of condensers

### LINKED ISSUE #2999: IIDM 1.13: support of condensers

For `powsybl-open-loadflow` and `powsybl-open-reac`, we need a way to make the difference between a generator that is not started and a generator that works as synchronous condenser. It is not possible for the moment, except forcing Pmin and Pmax of the condenser to be equal to zero. 

With a new version of xiidm (1.13, 2024.2.0), we propose to add a field `operatingAs` or any proposal with following enum values:
- `GENERATOR` ;
- `SYNCHRONOUS_CONDENSER` ;
- `MOTOR`.

`STATCOM` not in that PR because it is a technology that we may want to model entirely.

### Describe the motivation

_No response_

### Extra Information

- Note that we all agree to say that this field is not dynamic and does not depend on `targetP` value of a generator for example.
- Note that an equipment can be `GENERATOR` and `MOTOR` depending on the exploitation time: this is the example of pumping stations.
- A generator could be `GENERATOR` and `SYNCHRONOUS_CONDENSER` in theory.
- A generator could be only `GENERATOR`, meaning that a strictly negative `targetP` is forbidden?
- A generator could be only `MOTOR`, meaning that a strictly positive `targetP` is forbidden?
- A generator could be only `SYNCHRONOUS_CONDENSER`, meaning that `targetP` is always zero.
- Could we imagine a generator `MOTOR` and `SYNCHRONOUS_CONDENSER` or `MOTOR` and `GENERATOR` and `SYNCHRONOUS_CONDENSER`?

flo-dup:

Yes, fixes #2999

Feature

### DIFF
--- a/docs/grid_model/network_subnetwork.md
+++ b/docs/grid_model/network_subnetwork.md
@@ -109,18 +109,19 @@ A generator is an equipment that injects or consumes active power, and injects o
 
 **Characteristics**
 
-| Attribute | Unit | Description                                                 |
-| --------- | ---- |-------------------------------------------------------------|
-| $MinP$ | MW | Minimum generator active power output                       |
-| $MaxP$ | MW | Maximum generator active power output                       |
-| $ReactiveLimits$ | MVar | Operational limits of the generator (P/Q/V diagram)         |
-| $RatedS$ | MVA | The rated nominal power                                     |
-| $TargetP$ | MW | The active power target                                     |
-| $TargetQ$ | MVAr | The reactive power target at local terminal                 |
-| $TargetV$ | kV | The voltage target at regulating terminal                   |
-| $RegulatingTerminal$ |  | Associated node or bus for which voltage is to be regulated |
-| $VoltageRegulatorOn$ |  | True if the generator regulates voltage                     |
-| $EnergySource$ |  | The energy source harnessed to turn the generator           |
+| Attribute            | Unit | Description                                                                                                                            |
+|----------------------|------|----------------------------------------------------------------------------------------------------------------------------------------|
+| $MinP$               | MW   | Minimum generator active power output                                                                                                  |
+| $MaxP$               | MW   | Maximum generator active power output                                                                                                  |
+| $ReactiveLimits$     | MVar | Operational limits of the generator (P/Q/V diagram)                                                                                    |
+| $RatedS$             | MVA  | The rated nominal power                                                                                                                |
+| $TargetP$            | MW   | The active power target                                                                                                                |
+| $TargetQ$            | MVAr | The reactive power target at local terminal                                                                                            |
+| $TargetV$            | kV   | The voltage target at regulating terminal                                                                                              |
+| $RegulatingTerminal$ |      | Associated node or bus for which voltage is to be regulated                                                                            |
+| $VoltageRegulatorOn$ |      | True if the generator regulates voltage                                                                                                |
+| $EnergySource$       |      | The energy source harnessed to turn the generator                                                                                      |
+| $IsCondenser$        |      | True if the generator may behave as a synchronous condenser, that is, if it may control voltage even if its $TargetP$ is equal to zero |
 
 **Specifications**
 
--- a/docs/grid_model/network_subnetwork.md
+++ b/docs/grid_model/network_subnetwork.md
@@ -109,19 +109,19 @@ A generator is an equipment that injects or consumes active power, and injects o
 
 **Characteristics**
 
-| Attribute            | Unit | Description                                                                                                                            |
-|----------------------|------|----------------------------------------------------------------------------------------------------------------------------------------|
-| $MinP$               | MW   | Minimum generator active power output                                                                                                  |
-| $MaxP$               | MW   | Maximum generator active power output                                                                                                  |
-| $ReactiveLimits$     | MVar | Operational limits of the generator (P/Q/V diagram)                                                                                    |
-| $RatedS$             | MVA  | The rated nominal power                                                                                                                |
-| $TargetP$            | MW   | The active power target                                                                                                                |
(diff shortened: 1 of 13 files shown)


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/GeneratorAdderImpl.java
```diff
@@ -121,7 +129,7 @@ id, getName(), isFictitious(), energySource,
                                     minP, maxP,
                                     voltageRegulatorOn, regulatingTerminal != null ? regulatingTerminal : terminal,
                                     targetP, targetQ, targetV,
-                                    ratedS);
+                                    ratedS, isCondenser);
```
annetill, line 132: @geofjamg I think that we have to add a validity checks or not, need to think about it: 
- targetP must be zero if condenser = true.
- or we consider that all active power related attributes must be ignored? For example, the generator should be excluded form active power distribution so it is natural to ignore related data.
What do you think?

docs/grid_model/network_subnetwork.md
```diff
@@ -104,18 +104,19 @@ A generator is an equipment that injects or consumes active power, and injects o
...
+| $TargetP$            | MW   | The active power target                                                                                                                |
+| $TargetQ$            | MVAr | The reactive power target at local terminal                                                                                            |
+| $TargetV$            | kV   | The voltage target at regulating terminal                                                                                              |
+| $RegulatingTerminal$ |      | Associated node or bus for which voltage is to be regulated                                                                            |
+| $VoltageRegulatorOn$ |      | True if the generator regulates voltage                                                                                                |
+| $EnergySource$       |      | The energy source harnessed to turn the generator                                                                                      |
+| $IsCondenser$        |      | True if the generator may behave as a synchronous condenser, that is, if it may control voltage even if its $TargetP$ is equal to zero |
```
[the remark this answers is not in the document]
jeandemanged: not sure to understand what you plan to change, but as long as `True if the generator may control voltage even if its $TargetP$ is equal to zero` remains, then ok for me.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/GeneratorAdderImpl.java
```diff
@@ -33,6 +33,8 @@ class GeneratorAdderImpl extends AbstractInjectionAdder<GeneratorAdderImpl> impl
 
     private double ratedS = Double.NaN;
 
+    private Boolean isCondenser = false;
```
jeandemanged, line 36: could be a primitive `boolean`

docs/grid_model/network_subnetwork.md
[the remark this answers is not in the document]
annetill: The main issue is "synchronous". About of course the comment about voltage control is right, but not exclusive that is whay I prefer to put in the description and not in the table.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/GeneratorAdderImpl.java
jeandemanged: > * targetP must be zero if condenser = true.

I'd be against that, or, for a practical use, it would require `isCondenser` to be renamed to `isCondensing` + make it variant dependent.

> * or we consider that all active power related attributes must be ignored? For example, the generator should be excluded form active power distribution so it is natural to ignore related data.
What do you think?

My call would be that isCondenser should be used/checked only when targetP is zero.
Conversely, if targetP is not zero, then isCondenser should not be used/checked.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/AbstractIidmSerDeTest.java
```diff
@@ -103,6 +103,16 @@ protected void allFormatsRoundTripFromVersionedXmlFromMinToCurrentVersionTest(St
                 .toArray(IidmVersion[]::new));
     }
 
+    /**
+     * Execute a round trip test on the test resource IIDM-XML file with a given file name for all IIDM versions
+     * equals or more recent than a given minimum IIDM version.
+     */
+    protected void allFormatsRoundTripFromVersionedXmlFromMinVersionTest(String file, IidmVersion minVersion) throws IOException {
```
jeandemanged, line 110: not sure here here ... 

why not re-using existing `allFormatsRoundTripFromVersionedXmlFromMinToCurrentVersionTest` used already in many places ?

jeandemanged: oh I see, thanks for the explanation

docs/grid_model/network_subnetwork.md
```diff
@@ -180,6 +181,9 @@ The `VoltageRegulatorOn` attribute is required. It voltage regulation is enabled
 
 Target values for generators (`TargetP` and `TargetQ`) follow the generator sign convention: a positive value means an injection into the bus. Positive values for `TargetP` and `TargetQ` mean negative values at the flow observed at the generator `Terminal`, as `Terminal` flow always follows load sign convention. The following diagram shows the sign convention of these quantities with an example.
 
+The `isCondenser` value corresponds for instance to generators which can control voltage even if their targetP is equal to zero.
```
annetill, line 184: @jeandemanged or under Pmin?

jeandemanged: in OLF we have another parameter to disable voltage control if outside Pmin/Pmax

jeandemanged (approved):
thanks :)

PR #3069, opened 2024-06-19
Fix exception for connectable not found in Network Modification connection and disconnection

rolnico:

No

Bug fix

If you use for example a ConnectableConnection on a non-existing connectable or on a TieLine (which is not a Connectable), the application of the NetworkModification will crash due to a null pointer exception

You will now get a PowsyblException if you apply with `throwException==true` or the Network Modification will not be applied and the error will be logged

- [ ] Yes
- [X] No


rolnico:
https://github.com/powsybl/powsybl-core/pull/3072 could also be another solution

annetill (approved):

PR #3072, opened 2024-06-19
Add connect/disconnect on Tie Lines and HVDC Lines

rolnico:

No

Bug fix + new Feature

If you use for example a ConnectableConnection on a non-existing connectable or on a TieLine (which is not a Connectable), the application of the NetworkModification will crash due to a null pointer exception. This is due to the fact that only Connectables are managed since Tie Lines and HVDC Lines do not have connect/disconnect methods.

Tie Lines and HVDC Lines now have connect and disconnect methods implemented.
Network modification for connection and disconnection now also work with Tie Lines and HVDC Lines.

- [X] Yes
- [ ] No

If you had your own implementation for HVDC Lines, you now need to implement the following methods:
```java
    boolean connectConverterStations();
    boolean connectConverterStations(Predicate<Switch> isTypeSwitchToOperate);
    boolean connectConverterStations(Predicate<Switch> isTypeSwitchToOperate, TwoSides side);
    boolean disconnectConverterStations();
    boolean disconnectConverterStations(Predicate<Switch> isSwitchOpenable);
    boolean disconnectConverterStations(Predicate<Switch> isSwitchOpenable, TwoSides side);
```

If you had your own implementation for Tie Lines, you now need to implement the following methods:
```java
    boolean connectDanglingLines();
    boolean connectDanglingLines(Predicate<Switch> isTypeSwitchToOperate);
    boolean connectDanglingLines(Predicate<Switch> isTypeSwitchToOperate, TwoSides side);
    boolean disconnectDanglingLines();
    boolean disconnectDanglingLines(Predicate<Switch> isSwitchOpenable);
    boolean disconnectDanglingLines(Predicate<Switch> isSwitchOpenable, TwoSides side);
```


rolnico:
This is another possible solution that would close https://github.com/powsybl/powsybl-core/pull/3069

rolnico:
Should the Tie Line and HVDC Line become Connectables ?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/HvdcLine.java
```diff
@@ -204,6 +207,20 @@ default HvdcConverterStation<?> getConverterStation(TwoSides side) {
...
+
+    boolean connect(Predicate<Switch> isTypeSwitchToOperate, ThreeSides side);
+
+    boolean disconnect();
+
+    boolean disconnect(Predicate<Switch> isSwitchOpenable);
+
+    boolean disconnect(Predicate<Switch> isSwitchOpenable, ThreeSides side);
+
+    List<Terminal> getTerminals(ThreeSides side);
```
flo-dup, line 222: We don't need to expose `getTerminals` so you could remove it here

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/HvdcLine.java
```diff
@@ -204,6 +207,20 @@ default HvdcConverterStation<?> getConverterStation(TwoSides side) {
      */
     void remove();
 
+    boolean connect();
+
+    boolean connect(Predicate<Switch> isTypeSwitchToOperate);
+
+    boolean connect(Predicate<Switch> isTypeSwitchToOperate, ThreeSides side);
```
flo-dup, line 214: `TwoSides` would make more sense here and in disconnect

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/TieLine.java
flo-dup, line 1: Same comments for TieLine

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/HvdcLineImpl.java
```diff
@@ -231,6 +241,145 @@ public void remove() {
         network.getListeners().notifyAfterRemoval(id);
     }
 
+    @Override
+    public boolean connect() {
+        return connect(SwitchPredicates.IS_NONFICTIONAL_BREAKER);
```
flo-dup, line 246: didn't realize, but isn't it fictitious rather than fictional?

flo-dup (changes requested):
Could you add a utility method to avoid the code duplication?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/HvdcLineImpl.java
rolnico: both can be used, IMO.
If we want to change it, we should do it in another PR

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractConnectable.java
```diff
@@ -206,59 +205,13 @@ public boolean connect(Predicate<Switch> isTypeSwitchToOperate) {
...
-
-        // Connect all bus-breaker terminals
-        for (TerminalExt terminal : getTerminals(side)) {
-            if (!terminal.isConnected()
-                && terminal.getVoltageLevel().getTopologyKind() == BUS_BREAKER) {
-                // At this point, isNowConnected should always stay true but let's be careful
-                isNowConnected = isNowConnected && terminal.connect(isTypeSwitchToOperate);
-            }
-        }
+        List<Terminal> terminalsToConnect = new ArrayList<>(getTerminals(side));
```
flo-dup, line 208: I don't think you need to copy the list, you only iterate over it in the connect. You did it for the case of the connectable being removed while connecting it?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractConnectable.java
```diff
@@ -206,59 +205,13 @@ public boolean connect(Predicate<Switch> isTypeSwitchToOperate) {
...
 
-        // Disconnect all switches on node-breaker terminals
-        switchForDisconnection.forEach(sw -> sw.setOpen(false));
-        return isNowConnected;
+        return ConnectDisconnectUtil.connectAllTerminals(
+            this,
+            terminalsToConnect,
+            isTypeSwitchToOperate,
+            this.getNetwork().getReportNodeContext().getReportNode());
```
flo-dup, line 214: You can remove the `this` here (and in the other calls to connect/disconnect)
```suggestion
            getNetwork().getReportNodeContext().getReportNode());
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/HvdcLineImpl.java
```diff
@@ -231,6 +235,51 @@ public void remove() {
...
+        return connectConverterStations(SwitchPredicates.IS_NONFICTIONAL_BREAKER, null);
+    }
+
+    @Override
+    public boolean connectConverterStations(Predicate<Switch> isTypeSwitchToOperate) {
+        return connectConverterStations(isTypeSwitchToOperate, null);
+    }
+
+    @Override
+    public boolean connectConverterStations(Predicate<Switch> isTypeSwitchToOperate, TwoSides side) {
```
flo-dup, line 249: Shouldn't we rename this to `connectConverterStation`? It's strange to give a side and have the plural form, and also to interpret the null value as both sides.

rolnico: For the null value, it's already like this for the connectables so I kept the same logic. Same for the plural form : it's `getTerminals(ThreeSides side)` even when we give a ThreeSide so I tried to keep it coherent

olperr1 (changes requested):
@annetill: Should we also revert the javadoc changes of #3079?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractConnectable.java
olperr1: Some are still present in `HvdcLineImpl` and `TieLineImpl`.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
```diff
@@ -28,6 +28,15 @@ public final class ConnectDisconnectUtil {
     private ConnectDisconnectUtil() {
     }
 
+    /**
+     * Connect the specified terminals of the identifiable. It will connect all the specified terminals or none if at
```
annetill, line 32: terminals of the identifiable does not sound good. Terminals are related to connectable, exception for HVDC line and tie line only.

olperr1 (approved):

PR #3076, opened 2024-06-20
Fix ReferenceTerminals and SlackTerminal iIDM extension with merged networks and subnetworks

jeandemanged:
**Please check if the PR fulfills these requirements**

- [X] The commit message follows our guidelines
- [X] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**Does this PR already have an issue describing the problem?**

No

**What kind of change does this PR introduce?**

Bug fix

**What is the current behavior?**

ReferenceTerminals and SlackTerminal iIDM extensions were never revisited to work properly with subnetworks.
In particular:
- We get an exception when merging a network having a ReferenceTerminals extension, writing it, and reading it.
- `ReferenceTerminals.reset` and `SlackTerminal.reset` do not reset the subnetworks

**What is the new behavior (if this is a feature change)?**
ReferenceTerminals fixed to properly check the network owning the terminal.
`reset` methods will also reset the subnetworks.

**Does this PR introduce a breaking change or deprecate an API?**
- [x] No

PR #3081, opened 2024-06-24
Fix area boundary auto-removal with merged and detached Networks

jeandemanged:

No

Bug fix

In #2955 listeners were implemented to automatically delete area boundaries when the undelying equipment is removed from the network (e.g. DanglingLine removal). This was done using "cleanup" listeners. However when performing network merge or detach, the listeners are lost / not transfered to the new root network of the Areas.

The "cleanup" listeners are properly moved to the new root Network when merging and detaching.

- [x] No

**Other information**:

This is very similar to #3076, although here it is not about iIDM extensions.
I originally wanted to add notification mechanism for merging and detaching, but obviously this wouldn't work because manipulating listeners within listeners isn't the right approach (ConcurrentModificationException).
I made few specific code dedicated to Area only. When iIDM will be improved to deal with equipment removal & cleanup in a generic way, this would be revisited.

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AreaImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AreaImpl.java
@@ -48,9 +48,19 @@ public class AreaImpl extends AbstractIdentifiable<Area> implements Area {
         }
     }
 
+    /**
+     * Must be called whenever the Area is moved to a different root Network when merging and detaching.
+     * @param fromNetwork previous root network
+     * @param toNetwork new root network
+     */
+    void moveListener(NetworkImpl fromNetwork, NetworkImpl toNetwork) {
+        fromNetwork.removeListener(this.areaListener);
+        toNetwork.addListener(this.areaListener);
+    }
+
     private final NetworkListener areaListener;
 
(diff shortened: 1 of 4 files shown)


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AreaImpl.java
```diff
@@ -48,9 +48,19 @@ public void beforeRemoval(Identifiable<?> identifiable) {
...
+     */
+    void moveListener(NetworkImpl fromNetwork, NetworkImpl toNetwork) {
+        fromNetwork.removeListener(this.areaListener);
+        toNetwork.addListener(this.areaListener);
+    }
+
     private final NetworkListener areaListener;
 
-    public AreaImpl(Ref<NetworkImpl> ref, Ref<SubnetworkImpl> subnetworkRef, String id, String name, boolean fictitious, String areaType,
+    AreaImpl(Ref<NetworkImpl> ref, Ref<SubnetworkImpl> subnetworkRef, String id, String name, boolean fictitious, String areaType,
```
jeandemanged, line 63: shoudn't have been made `public` in the first place

m-guibert: Good catch, thank you !

m-guibert (approved):

PR #3082, opened 2024-06-25
Datasource: add mainExtension, existsStrict and use it in exists() methods of importers [WIP]

jonenst:
This allows to import the correct network when a path is provided, the importers will respect the full path and not declare that they can import from the datasource TODO explain more

**Please check if the PR fulfills these requirements**

- [x] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**Does this PR already have an issue describing the problem?**

**What kind of change does this PR introduce?**

**What is the current behavior?**

**What is the new behavior (if this is a feature change)?**

**Does this PR introduce a breaking change or deprecate an API?**
- [ ] Yes
- [ ] No

**If yes, please check if the following requirements are fulfilled**

- [ ] The *Breaking Change* or *Deprecated* label has been added
- [ ] The migration steps are described in the following section

**What changes might users need to make in their application due to this PR? (migration steps)**

**Other information**:

PR #3085, opened 2024-06-26
Disambiguate files on import

jonenst:

https://github.com/powsybl/powsybl-core/issues/989
#3082
#2959
others...

bug fix

```shell
$ itools convert-network --input-file foo.xiidm # will wrongly use foo.xml from cgmes if it exists
```

always use the file specified by the user in the command line

- [ ] Yes
- [x] No

### DIFF
--- a/ieee-cdf/ieee-cdf-converter/src/main/java/com/powsybl/ieeecdf/converter/IeeeCdfImporter.java
+++ b/ieee-cdf/ieee-cdf-converter/src/main/java/com/powsybl/ieeecdf/converter/IeeeCdfImporter.java
@@ -79,6 +79,12 @@ public class IeeeCdfImporter implements Importer {
 
     @Override
     public boolean exists(ReadOnlyDataSource dataSource) {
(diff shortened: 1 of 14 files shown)


jonenst:
Need more tests

ieee-cdf/ieee-cdf-converter/src/main/java/com/powsybl/ieeecdf/converter/IeeeCdfImporter.java
```diff
@@ -79,6 +79,12 @@ public String getComment() {
 
     @Override
     public boolean exists(ReadOnlyDataSource dataSource) {
+        // if given a main file with an extension and it's not compatible, don't import
+        // to avoid importing a file when the user specified another
+        if (!dataSource.getBaseExtension().isEmpty()) {
```
jonenst, line 84: bug here if getBaseExtension.equals(EXT) we must not return false...

flo-dup:
Replaced by #3102

PR #3086, opened 2024-06-26
Allow to import folders

jonenst:

no

feature

impossible to import all files in a folder, and impossible to import files in a folder that don't have the same basename as the folder

possible

- [ ] Yes
- [x] No


annetill:
Hi @jonenst, for me it is not a use case. If this feature is for CGMES users, we don't need it. The current design, where we load in triple store, is working fine and the community is used to it. I think that we don't have to create new use cases with datasource, just fix the current ones.

jonenst:
https://github.com/powsybl/powsybl-core/pull/3092/files is a more explicit and better choice I think

PR #3092, opened 2024-07-05
Importfolders2

jonenst:

NO

feature

impossible to import all files in a folder, and impossible to import files in a folder that don't have the same basename as the folder

possible

- [ ] Yes
- [x] No

**Other information**:

alternative to #3086


flo-dup (approved):

rolnico (approved):

docs/grid_exchange_formats/going_further/datasources.md
```diff
@@ -91,8 +91,14 @@ Even if `DirectoryDataSource` integrates the notions of base name and data exten
...
 extension if there is one.
 
-In addition to filtering with the regex parameter, in directory datasources the method `listNames(String regex)` filters
-filenames to only keep those starting with the base name.
+Note that in directory datasources, there are two behaviours for the method `listNames(String regex)`, which is used 
+to filter the files within the datasource. In addition to the filtering using the regex parameter, it can also 
+filter filenames to only keep those starting with the base name. This behaviour basename-filtering only occurs if 
+`allFiles` is set to `false` in the datasource.
+Setting `allFiles` to `true` is  for example very useful for the CGMES use case, in which a network is defined by 
+several files (TP, EQ, SV, SSH, GL, ...), often without a common prefix (base name).
```
jonenst, line 99:
```suggestion
several files (TP, EQ, SV, SSH, GL, ...), often without a common prefix (base name). This makes the DirectoryDataSource behave just like an ArchiveDataSource. 
```

commons/src/main/java/com/powsybl/commons/datasource/DataSourceUtil.java
```diff
@@ -62,20 +63,41 @@ static DataSource createDataSource(Path directory, String basename, String dataE
...
-    static DataSource createDataSource(Path directory, String fileNameOrBaseName, DataSourceObserver observer) {
-        Objects.requireNonNull(directory);
-        Objects.requireNonNull(fileNameOrBaseName);
+    static DataSource createDataSource(Path fileOrDirectory, DataSourceObserver observer) {
+        Objects.requireNonNull(fileOrDirectory);
+
+        DataSourceBuilder dataSourceBuilder = new DataSourceBuilder().withObserver(observer);
+        if (Files.isDirectory(fileOrDirectory)) {
+            dataSourceBuilder.withDirectory(fileOrDirectory)
+                .withBaseName(fileOrDirectory.getFileName().toString())
```
jonenst, line 72:
```suggestion
// we use the full directoryname here instead of a parsed representation using FileInformation
// this is required for now to allow directories with dots but without the dataextension (like "microgrid.v1.complete/")
// otherwise Fileinformation interprets the part after the dot as the dataextension. This has the drawback that
// directories with a dataextension (e.g. "network.xiidm/") will be rejected/mishandled because we will use
// a file named "network.xiidm.xiidm" inside it
// NOTE: this does not occur if the directory has a dot and also has the dataextension: "microgrid.v1.complete.xml/" works
// NOTE2: Maybe this can be fixed in the future by looking more at what is inside the directory and we can remove this
                .withBaseName(fileOrDirectory.getFileName().toString())
```

rolnico: This should go in the javadoc more than in a simple comment

PR #3096, opened 2024-07-09
[Short-circuit API] Add some javadoc.

colinepiloquet:

No

Docs update

A lot of javadoc is missing.

I've tried to add a short description for every public class or method.

- [ ] Yes
- [X] No


shortcircuit-api/src/main/java/com/powsybl/shortcircuit/ShortCircuitParameters.java
```diff
@@ -304,6 +308,9 @@ public ShortCircuitParameters setDetailedReport(boolean detailedReport) {
         return this;
     }
 
+    /**
+     * Validates the ShortCircuitParameters. If the initial voltage profile mode is set to configured, then the voltage ranges should not be empty.
```
So-Fras, line 312:
```suggestion
     * Validates the ShortCircuitParameters. If the initial voltage profile mode is set to CONFIGURED, then the voltage ranges should not be empty.
```

So-Fras (approved):

PR #3099, opened 2024-07-10
CGMES metadata models: allow clear of supersedes

zamarrenolm:

Bug fix

Supersedes set in CGMES metadata models (SSH) can not be emptied by the user.
This creates a problem when trying to export CGMES data for a network that is read from CGMES inputs that already specify superseding info. 

In CGM quick export, the previous supersede and the one added automatically during export are present in the output.

In manual export, the user may add a supersede to the model, to prepare the export, but the previous supersede can not be removed. The exported files contains two supersede items.

The CGMES metada model supersedes set in the network extension can be emptied with the method `CgmesMetadataModel::clearDependencies`.

The quick CGM export clears the existing SSH supersedes before adding the current one.

For the manual export of single files, the user is now being able to clear previous supersedes before adding the requested one:
```java
network.getExtension(CgmesMetadataModels.class)
    .getModelForSubset(CgmesSubset.STEADY_STATE_HYPOTHESIS).orElseThrow()
        .clearSupersedes()
        .addSupersedes("manual-supersede");
network.write("CGMES", exportParams, ...);
```

- [X] Yes
- [ ] No

If you had define a custom `CgmesMetadataModel` implementation, you need to define a new method:
`CgmesMetadataModelImpl clearSupersedes()`.

**Other information**:

### DIFF
--- a/cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CommonGridModelExportTest.java
+++ b/cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CommonGridModelExportTest.java
@@ -470,6 +470,7 @@ class CommonGridModelExportTest extends AbstractSerDeTest {
             String country = n.getSubstations().iterator().next().getCountry().orElseThrow().toString();
             CgmesMetadataModel sshModel = n.getExtension(CgmesMetadataModels.class).getModelForSubset(CgmesSubset.STEADY_STATE_HYPOTHESIS).orElseThrow();
             sshModel.clearDependencies()
+                    .clearSupersedes()
                     .addDependentOn("myDependency")
                     .addSupersedes("mySupersede")
                     .setVersion(exportedVersion)
@@ -578,6 +579,7 @@ class CommonGridModelExportTest extends AbstractSerDeTest {
                 .setVersion(version)
                 .setModelingAuthoritySet("http://elia.be/CGMES/2.4.15")
                 .addDependentOn("BE EQ model ID")
+                .addSupersedes("BE SSH previous ID")
                 .addProfile("http://entsoe.eu/CIM/SteadyStateHypothesis/1/1")
                 .add()
                 .add();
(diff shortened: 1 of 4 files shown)


cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CommonGridModelExportTest.java
```diff
@@ -470,6 +470,7 @@ void testFaraoUseCaseManualExport() throws IOException {
             String country = n.getSubstations().iterator().next().getCountry().orElseThrow().toString();
             CgmesMetadataModel sshModel = n.getExtension(CgmesMetadataModels.class).getModelForSubset(CgmesSubset.STEADY_STATE_HYPOTHESIS).orElseThrow();
             sshModel.clearDependencies()
+                    .clearSupersedes()
```
annetill, line 473: The test has no assertion linked to our changes.

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesMetadataModel.java
```diff
@@ -166,4 +166,12 @@ public interface CgmesMetadataModel {
      * @return The model with an empty set of values this model depends on.
      */
     CgmesMetadataModelImpl clearDependencies();
+
+    /**
+     * Remove all the model ids this model supersedes.
+     * @return The model with an empty set of values this model supersedes.
+     */
+    default CgmesMetadataModelImpl clearSupersedes() {
```
annetill, line 174: Why do you need a default method here?

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CommonGridModelExportTest.java
zamarrenolm: In the line 510 below we check that the exported XML file contains exactly only one "supersede", that is the one that has been added by the user. The model used in the test contains a previous "supersede", so we are checking that the previous has been removed.

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesMetadataModel.java
zamarrenolm: Because the interface was already published in a previous version, and we are extending it. If other implementations have been created, they will fail to compile if we do not add a default.

annetill (approved):

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesMetadataModel.java
annetill: The patch release has other breaking changes, so after talking with @olperr1 we prefer the default behavior to be removed.

PR #3103, opened 2024-07-17
Datasources - part 3: TarArchiveDataSource

rolnico:

No

Feature

Tar files cannot be used

Tar files can now be used as datasource

- [ ] Yes
- [X] No

**Other information**:


commons/src/main/java/com/powsybl/commons/datasource/TarArchiveDataSource.java
```diff
@@ -0,0 +1,330 @@
...
+            baseName, null, compressionFormat, null);
+    }
+
+    public TarArchiveDataSource(Path directory, String baseName) {
+        this(directory,
+            baseName + ".tar",
+            baseName, null, null, null);
+    }
+
+    public TarArchiveDataSource(Path tarFile) {
```
jonenst, line 78: use private constructor with a FileInformation argument to avoid calling 3 times ?

commons/src/main/java/com/powsybl/commons/datasource/TarArchiveDataSource.java
```diff
@@ -0,0 +1,330 @@
...
+        Set<String> names = new HashSet<>();
+        Path tarFilePath = getArchiveFilePath();
+
+        // Explore the archive
+        try (BufferedInputStream inputStream = new BufferedInputStream(Files.newInputStream(tarFilePath));
+             InputStream cis = getCompressedInputStream(inputStream, compressionFormat);
+             TarArchiveInputStream tar = new TarArchiveInputStream(cis)) {
+            ArchiveEntry entry;
+            while ((entry = tar.getNextEntry()) != null) {
+                // Consider only files located at the root of the archive
```
jonenst, line 102: Maybe as a first version do the same as zip ? only filter on the matcher. (in the zip code the comment is wrong). Then in a separate PR we can decide if we implement this behavior for both ?

jonenst: https://github.com/powsybl/powsybl-core/blob/feeb6bc5d68427419939c40b7bd53d554003fe0c/commons/src/main/java/com/powsybl/commons/datasource/ZipFileDataSource.java#L212

but

https://github.com/powsybl/powsybl-core/blob/feeb6bc5d68427419939c40b7bd53d554003fe0c/commons/src/main/java/com/powsybl/commons/datasource/ZipFileDataSource.java#L222

commons/src/main/java/com/powsybl/commons/datasource/TarArchiveDataSource.java
```diff
@@ -0,0 +1,329 @@
...
+
+                    // Useful parameter
+                    taos.setBigNumberMode(TarArchiveOutputStream.BIGNUMBER_POSIX);
+                    taos.setLongFileMode(TarArchiveOutputStream.LONGFILE_POSIX);
+
+                    // Copy content of temporary stream file into an entry of the temporary archive
+                    try (InputStream is = Files.newInputStream(getTmpStreamFilePath(tarFilePath))) {
+
+                        // Content of the stream
+                        byte[] streamContent = is.readAllBytes();
```
flo-dup, line 264: isn't this a bit dangerous? That's only to get the size, couldn't you get it from `Files.size()`?

flo-dup (changes requested):

commons/src/main/java/com/powsybl/commons/datasource/TarArchiveDataSource.java
rolnico: not sure if it's dangerous but I can change it

flo-dup: the problem is if the file is huge, readAllBytes will require a huge byte array, which could be avoided in case of a copy. And if the file is bigger than  `Integer.MAX_NUMBER`, it will be truncated.

commons/src/main/java/com/powsybl/commons/datasource/TarArchiveDataSource.java
```diff
@@ -0,0 +1,329 @@
...
+
+                        // New tar entry
+                        TarArchiveEntry entry = new TarArchiveEntry(fileName);
+                        entry.setSize(Files.size(tmpStreamFilePath));
+
+                        // New file to add
+                        taos.putArchiveEntry(entry);
+
+                        // Write the data in the entry
+                        taos.write(is.readAllBytes());
```
flo-dup, line 274: The problem remains the same here. We could call the same method as in `ZipArchiveDataSource`
```suggestion
                        ByteStreams.copy(is, taos);
```

flo-dup (approved):

jonenst (approved):

PR #3108, opened 2024-07-24
Docs: fix bad edge 7-8 and make edges follow the underlying equipment in nodeBreakerTopologyGraph.svg

jonenst:

NO

Docs update

hard to understand how the green edges and pink edge correspond to the underlying disconnectors/breakers

Hopefully easier to follow ?

- [ ] Yes
- [x] No


jonenst:
would it be even better with the rest of the diagonal edges following the underlying equipment ?
nodeBreakerTopologyGraph

So-Fras (changes requested):
Thanks for fixing the 7-8 edge! 
There is also another mistake to fix for the 5-6 edge (it should be a dotted line). Could you fix it too?
I think that the diagonal lines are more visible. I can visualize the connections at a glance without needing to zoom in. What do you think?

jonenst:
nodeBreakerTopologyGraph
better like this ? I pushed this version

olperr1 (approved):

PR #3114, opened 2024-08-02
Datasources - part 5: fixes

rolnico:

No

Bug fix

- In `CgmesOnDataSource.checkIfMainFileNotWithCgmesData()`, a stream was open without first checking if the file existed, which could result in an unexpected exception.

- In `CgmesOnDataSource.checkIfMainFileNotWithCgmesData()`, we first check if the file exists before trying to open a stream on it.

- [X] Yes
- [ ] No

The following methods of `CgmesOnDataSource` now throw an `IOException`:
 - `public boolean exists()`
 - `public boolean existsCim14()`

If you call them, you should catch or propagate this exception. To keep the same behavior as before, you can catch this exception and propagate it in a new `UncheckedIOException`.

**Other information**:

### DIFF
--- a/cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesOnDataSource.java
+++ b/cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesOnDataSource.java
@@ -38,11 +38,12 @@ public class CgmesOnDataSource {
         if (dataSource.getDataExtension() == null || dataSource.getDataExtension().isEmpty()) {
             return false;
         } else if (EXTENSION.equals(dataSource.getDataExtension())) {
-            try (InputStream is = dataSource.exists(null, EXTENSION) ? dataSource.newInputStream(null, EXTENSION) : null) {
-                if (is == null) {
-                    return true;
+            try {
+                if (dataSource.exists(null, EXTENSION)) {
+                    try (InputStream is = dataSource.newInputStream(null, EXTENSION)) {
+                        return isCim14 ? !existsNamespacesCim14(NamespaceReader.namespaces(is)) : !existsNamespaces(NamespaceReader.namespaces(is));
+                    }
                 }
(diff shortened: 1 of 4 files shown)


jonenst (approved):

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesOnDataSource.java
```diff
@@ -37,8 +37,11 @@ public ReadOnlyDataSource dataSource() {
     private boolean checkIfMainFileNotWithCgmesData(boolean isCim14) {
         if (dataSource.getDataExtension() == null || dataSource.getDataExtension().isEmpty()) {
             return false;
-        } else if (EXTENSIONS.equals(dataSource.getDataExtension())) {
-            try (InputStream is = dataSource.newInputStream(null, EXTENSIONS)) {
+        } else if (EXTENSION.equals(dataSource.getDataExtension())) {
+            try (InputStream is = dataSource.exists(null, EXTENSION) ? dataSource.newInputStream(null, EXTENSION) : null) {
+                if (is == null) {
+                    return true;
+                }
```
flo-dup, line 44: why not checking beforehand?
```suggestion
            if (dataSource.exists(null, EXTENSION) == null) {
                return true;
            }
            try (InputStream is = dataSource.newInputStream(null, EXTENSION)) {
```

jonenst: because datasource.exists is declared to throw IOexception, so we need to rewrap with an extra try catch

flo-dup: Well I'd rather have 2 nested try than current code. Or we could catch the `IOException` only in `CgmesImport::exists`
```suggestion
            try {
                if (!dataSource.exists(null, EXTENSION)) {
                    return true;
                }
                try (InputStream is = dataSource.newInputStream(null, EXTENSION)) {
```

olperr1 (approved):

PR #3115, opened 2024-08-08
Flatten network

olperr1:

No

Feature

When a network has subnetworks, there is no way to delete the subnetwork partitioning without also removing all the content of these subnetworks.

A new method `Network.flatten()` is introduced. When called on a network containing subnetworks, all the elements contained in these latter are attached to the root network and the subnetworks are removed.

- [X] Yes
- [ ] No

Custom IIDM implementation maintainers should add the following method in their `Network` implementations:
- `Network.flatten()`.

**Other information**:


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
```diff
@@ -1195,6 +1195,27 @@ public boolean isBoundaryElement(Identifiable<?> identifiable) {
...
 
+    @Override
+    public void flatten() {
+        if (subnetworks.isEmpty()) {
+            // Nothing to do
+            return;
+        }
+        subnetworks.values().forEach(subnetwork -> {
+            // The subnetwork ref chain should point to the current network's subnetworkRef.
+            // This way, all its network elements (using this ref chain) will have a reference to the current network
```
flo-dup, line 1206: One thing I missed at first was that the reference is through a double RefChain. Is it worth mentioning it or would it make it more confusing?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
```diff
@@ -1083,7 +1083,7 @@ private void checkMergeability(NetworkImpl otherNetwork) {
     private static void createSubnetwork(NetworkImpl parent, NetworkImpl original) {
         // The root network reference should point to parent and not original anymore
```
flo-dup, line 1084: Could you just mention like in flatten why you need to have a ref pointing to the same RefChain as the parent network?

flo-dup (approved):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
olperr1: I added more details in the comment.

PR #3116, opened 2024-08-12
Limits copier

nao1345678:

It allows the creation of current limits by copying already existing ones.

The current behavior forces to write the whole limits by hand.

Instead of calling a `new...Limits()` method to get a new limit adder, you can call the same method with an existing limit as parameter to get a new adder already initialized with the values of the said limit.

For instance:

```java
        line.newCurrentLimits2(limits1).add();
```
will copy `limits1` on side 2 of `line`.

- [ ] Yes
- [x] No

### DIFF
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/FlowsLimitsHolder.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/FlowsLimitsHolder.java
@@ -127,6 +127,14 @@ public interface FlowsLimitsHolder {
      */
     CurrentLimitsAdder newCurrentLimits();
 
+    /**
+     * <p>Create an adder to add a copy of the {@link CurrentLimits} in parameters in the selected {@link OperationalLimitsGroup}. </p>
+     * <p>If there's no selected group, the adder will also create a new group with the default name and set it as selected.
+     * This operation is performed when the limits are created via {@link CurrentLimitsAdder#add()}, only if the limits to add
+     * are valid.</p>
+     * @param limits a set of existing current limits.
+     * @return an adder allowing to create a new {@link CurrentLimits} initialized from the limits in parameters in the selected {@link OperationalLimitsGroup}.
+     */
     default CurrentLimitsAdder newCurrentLimits(CurrentLimits limits) {
         CurrentLimitsAdder adder = newCurrentLimits();
         return initializeFromLoadingLimits(adder, limits);
@@ -141,6 +149,14 @@ public interface FlowsLimitsHolder {
      */
     ApparentPowerLimitsAdder newApparentPowerLimits();
 
+    /**
+     * <p>Create an adder to add a copy of the {@link ApparentPowerLimits} in parameters in the selected {@link OperationalLimitsGroup}. </p>
+     * <p>If there's no selected group, the adder will also create a new group with the default name and set it as selected.
+     * This operation is performed when the limits are created via {@link CurrentLimitsAdder#add()}, only if the limits to add
+     * are valid.</p>
+     * @param limits a set of existing apparent power limits.
+     * @return an adder allowing to create a new {@link ApparentPowerLimits} initialized from the limits in parameters in the selected {@link OperationalLimitsGroup}.
+     */
     default ApparentPowerLimitsAdder newApparentPowerLimits(ApparentPowerLimits limits) {
         ApparentPowerLimitsAdder adder = newApparentPowerLimits();
         return initializeFromLoadingLimits(adder, limits);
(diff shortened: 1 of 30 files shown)


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/LoadingLimitsAdder.java
```diff
@@ -102,6 +102,13 @@ interface TemporaryLimitAdder<A> {
      */
     String getOwnerId();
 
+    /**
+     * <p>Creates a limit that is the exactly same as the one given in parameters.</p>
+     * @param limits the loading limit getting copied.
+     * @return a new limit
+     */
+    A copyFromCurrentLimits(CurrentLimits limits);
```
olperr1, line 110: It would be better to simply call

```java
CurrentLimitsAdder adder2 = line.newCurrentLimits2(limits1);
```
instead of 

```java
CurrentLimitsAdder adder2 = line.newCurrentLimits2()
                .copyFromCurrentLimits(limits1);
```
since it doesn't really make sense to call another method before `copyFromCurrentLimits(...)`.

To achieve this, "shortcut" methods should be declared each time an `add...Limits...()` method is defined.

In order to avoid breaking changes, I think the following strategy could be applied:

1. Move the `copyFromCurrentLimits` implementation in `LoadingLimitsUtil`. It should be possible since it relies only on API method calls (note that it should be renamed and should use generic types to ensure that the return's type is compatible with the argument's one — see the previous comments).
2. In all the interfaces where an `add...Limits...()` method is defined, define a **default** method `add...Limits...(...Limits)` which will call the method without argument then call the method defined in 1. on its result.

This way, custom IIDM implementation maintainers won't have to do anything to support this new feature, but it will be possible to them to define a custom implementation if really needed (by overriding the default methods).

olperr1 (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/LoadingLimitsAdder.java
olperr1: With the implementations in `AbstractConnectableBranch` and `TieLineImpl`, there is a breaking change for custom IIDM implementation maintainers since they don't use these classes but only the interface (`Branch`).

If you remove these implementations and define the default methods in `Branch`, at is was done for `FlowsLimitsHolder`, there will be no breaking change.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/FlowsLimitsHolder.java
```diff
@@ -125,6 +127,11 @@ default ApparentPowerLimits getNullableApparentPowerLimits() {
      */
     CurrentLimitsAdder newCurrentLimits();
 
+    default CurrentLimitsAdder newCurrentLimits(CurrentLimits limits) {
```
olperr1, line 130: Can you add javadoc comments here too?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/TieLineImpl.java
```diff
@@ -279,16 +279,31 @@ public CurrentLimitsAdder newCurrentLimits1() {
...
+    }
+
     @Override
     public ActivePowerLimitsAdder newActivePowerLimits1() {
         return danglingLine1.newActivePowerLimits();
     }
 
+    @Override
+    public ActivePowerLimitsAdder newActivePowerLimits1(ActivePowerLimits activePowerLimits) {
+        return danglingLine1.newActivePowerLimits();
```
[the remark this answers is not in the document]
nao1345678: Should I do it with every `new...Limits1...(limits)` methods in `TieLineImpl.java` ?

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractThreeWindingsTransformerTest.java
```diff
@@ -1038,4 +1041,69 @@ private void createThreeWindingsTransformerWithLeg3(double r, double x, double g
...
+                .setAcceptableDuration(20 * 60)
+                .setValue(1200.0)
+                .endTemporaryLimit();
+        apparentPowerLimitsAdder1.add();
+
+        CurrentLimits currentLimits1 = leg1.getCurrentLimits().get();
+        ActivePowerLimits activePowerLimits1 = leg1.getActivePowerLimits().get();
+        ApparentPowerLimits apparentPowerLimits1 = leg1.getApparentPowerLimits().get();
+
+        CurrentLimitsAdder currentLimitsAdder2 = leg1.newCurrentLimits(currentLimits1);
```
olperr1, line 1083: It would be better to add the new limits on leg2 or leg3.
In the current state, if the adders don't do anything, the test will still succeed.

olperr1 (changes requested):

olperr1 (approved):

PR #3122, opened 2024-08-27
Allow alternative serde for extensions

olperr1:

No

Feature

When an extension is added in `powsybl-core`, it sometimes cover the same functional perimeter as an existing extension declared in a custom project. A good practice is then to use the open-source extension instead of the custom one.
To allow working with previously generated IIDM files (containing the old extension name), a compatibility deserializer can 
easily be written (it should read the data of the old extension tag and create the new extension on the network being imported).

But at this time there is no way to export the open-source extension using the old extension name, which could be needed for compatibility with other softwares.

Introduce a mechanism to provide an alternative SerDe for an extension.

- [ ] Yes
- [X] No

**Other information**:

To use an alternative SerDe, you need to create a new SerDe class:
- It should be annotated with `@AutoService({AlternativeExtensionSerDe.class, ExtensionSerDe.class})`.
- It should extends either:
  - `AbstractAlternativeExtensionSerDe`, for a non-versioned SerDe;
  - `AbstractAlternativeVersionableNetworkExtensionSerDe`, for a versioned SerDe.

By default, the importer is always active, but the exporter is disable.
To use the alternative SerDe at export, you should specify the version `"alternative"` (or `ExportOptions.ALTERNATIVE_VERSION`) in your export options for the real extension.

If your SerDe is versioned, the used extension version will be the default for the IIDM version. If you want another version, you should also add this version **associated with the alternative extension name**.

For instance, if the original extension name is "foo" and the alternative extension name is "altFoo", the configuration will look like:

```java
var exportOptions = new ExportOptions()
                .addExtensionVersion("altFoo", "1.0")
                .addExtensionVersion("foo", ExportOptions.ALTERNATIVE_VERSION);
```

You can find examples in this PR:

| Original SerDe | Alternative SerDe | Type |
| --- | --- | --- |
| `ActivePowerControlSerDe` | `LegacyActivePowerControlSerDe` | Versioned |
| `OperatingStatusSerDe` | `LegacyOperatingStatusSerDe` | Non-versioned |

They are both used by the test class `AlternativeExtensionXmlTest`.


olperr1:
Superseded by #3282

PR #3123, opened 2024-08-30
Dry run - 3rd proposal - full dry run only

rolnico:

#3001 

- [x] Yes
- [ ] No

The classes extending the class `AbstractNetworkModification` now have to implement a new method `getName` returning the name of the network modification.

**Other information**:


iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/NetworkModification.java
```diff
@@ -19,45 +19,79 @@ public interface NetworkModification {
...
      * exception will be thrown. Otherwise, computation will continue but the injection will not be added to the network
      * in case of error.
      */
     void apply(Network network, NamingStrategy namingStrategy, boolean throwException, ReportNode reportNode);
 
+    /**
+     * Applies the modification to the given network. If throwException is set to true, then in case of error, an
+     * exception will be thrown. Otherwise, computation will continue but the injection will not be added to the network
+     * in case of error.
+     */
```
olperr1, line 81: It would be useful to add elements about the `dryRun` flag:

```suggestion
    /**
     * <p>Applies the modification to the given network. If throwException is set to true, then in case of error, an
     * exception will be thrown. Otherwise, computation will continue but the injection will not be added to the network
     * in case of error.</p>
     * <p>When <code>dryRun</code> is set to true, the modification is applied on a copy of the network and
     * no exception is thrown if an application error is encountered. Instead, the error is logged in <code>reportNode</code>
     * and the method returns <code>false</code>. If no application error is encountered, the method returns <code>true</code>.</p>
     */
```

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/NetworkModification.java
```diff
@@ -19,45 +19,79 @@ public interface NetworkModification {
...
      * exception will be thrown. Otherwise, computation will continue but the injection will not be added to the network
      * in case of error.
      */
     void apply(Network network, NamingStrategy namingStrategy, boolean throwException, ComputationManager computationManager, ReportNode reportNode);
+
+    /**
+     * Applies the modification to the given network. If throwException is set to true, then in case of error, an
+     * exception will be thrown. Otherwise, computation will continue but the injection will not be added to the network
+     * in case of error.
+     */
```
olperr1, line 95: Idem here:

```suggestion
    /**
     * <p>Applies the modification to the given network. If throwException is set to true, then in case of error, an
     * exception will be thrown. Otherwise, computation will continue but the injection will not be added to the network
     * in case of error.</p>
     * <p>When <code>dryRun</code> is set to true, the modification is applied on a copy of the network and
     * no exception is thrown if an application error is encountered. Instead, the error is logged in <code>reportNode</code>
     * and the method returns <code>false</code>. If no application error is encountered, the method returns <code>true</code>.</p>
     */
```

olperr1 (changes requested):

olperr1 (approved):

PR #3127, opened 2024-09-02
Add checks on Network Modifications

rolnico:

No

Feature

In order to see if a network modification can be applied to a network, you have to apply it and see if an exception is thrown.
In order to know if it had an impact on the network, you have to manually compare the values.

The new method `hasImpactOnNetwork(network)` provides for each Network Modification an enum telling if the network modification can be applied to the network and, if so, if it will have an impact on the network.

- [ ] Yes
- [x] No

**Other information**:


iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/NetworkModificationList.java
```diff
@@ -36,4 +36,21 @@ public void apply(Network network, NamingStrategy namingStrategy, boolean throwE
...
+            if (unitaryImpact == NetworkModificationImpact.CANNOT_BE_APPLIED) {
+                // If a sub-modification cannot be applied, the whole modification cannot be applied too
+                impact = unitaryImpact;
+                return impact;
+            } else if (hasImpact || unitaryImpact == NetworkModificationImpact.HAS_IMPACT_ON_NETWORK) {
+                hasImpact = true;
+            }
+        }
+        impact = hasImpact ? NetworkModificationImpact.HAS_IMPACT_ON_NETWORK : NetworkModificationImpact.NO_IMPACT_ON_NETWORK;
+        return impact;
```
pet-mit, line 54: I think the intention could be clearer if we simplify the code a little bit. Something like:
```suggestion
        Set<NetworkModificationImpact> unitaryImpacts = modificationList.stream().map(m -> m.hasImpactOnNetwork(network)).distinct().toSet();
        if (unitaryImpacts.contains(CANNOT_BE_APPLIED)) {
            return CANNOT_BE_APPLIED;
        }
        if (unitaryImpacts.contains(HAS_IMPACT_ON_NETWORK)) {
            return HAS_IMPACT_ON_NETWORK;
        }
        return NO_IMPACT_ON_NETWORK;
```

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/PhaseShifterShiftTap.java
```diff
@@ -55,4 +55,23 @@ private void adjustTapPosition(PhaseTapChanger phaseTapChanger) {
...
+    public NetworkModificationImpact hasImpactOnNetwork(Network network) {
+        impact = DEFAULT_IMPACT;
+        TwoWindingsTransformer phaseShifter = network.getTwoWindingsTransformer(phaseShifterId);
+        if (phaseShifter == null || !phaseShifter.hasPhaseTapChanger()) {
+            impact = NetworkModificationImpact.CANNOT_BE_APPLIED;
+        } else {
+            PhaseTapChanger phaseTapChanger = phaseShifter.getPhaseTapChanger();
+            int tapPosition = Math.min(Math.max(phaseTapChanger.getTapPosition() + tapDelta,
+                phaseTapChanger.getLowTapPosition()), phaseTapChanger.getHighTapPosition());
+            if (Math.abs(tapPosition - phaseTapChanger.getTapPosition()) < EPSILON
```
pet-mit, line 69:
```suggestion
            if (compareValues(tapPosition, phaseTapChanger.getTapPosition(), false)
```

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/PhaseShifterSetAsFixedTap.java
```diff
@@ -42,4 +42,18 @@ public void apply(Network network, NamingStrategy namingStrategy, boolean throwE
...
         phaseShifter.getPhaseTapChanger().setRegulationMode(PhaseTapChanger.RegulationMode.FIXED_TAP);
     }
+
+    @Override
+    public NetworkModificationImpact hasImpactOnNetwork(Network network) {
+        impact = DEFAULT_IMPACT;
+        TwoWindingsTransformer phaseShifter = network.getTwoWindingsTransformer(phaseShifterId);
+        if (phaseShifter == null || !phaseShifter.hasPhaseTapChanger()) {
+            impact = NetworkModificationImpact.CANNOT_BE_APPLIED;
+        } else if (Math.abs(tapPosition - phaseShifter.getPhaseTapChanger().getTapPosition()) < EPSILON
```
pet-mit, line 52: same

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/ShuntCompensatorModification.java
```diff
@@ -59,6 +56,19 @@ public void apply(Network network, NamingStrategy namingStrategy, boolean throwE
         }
     }
 
+    @Override
+    public NetworkModificationImpact hasImpactOnNetwork(Network network) {
+        impact = DEFAULT_IMPACT;
+        ShuntCompensator shuntCompensator = network.getShuntCompensator(shuntCompensatorId);
+        if (shuntCompensator == null) {
+            impact = NetworkModificationImpact.CANNOT_BE_APPLIED;
```
pet-mit, line 64: do we also want to verify that sectionCount is in the allows [minSectionCount, maxSectionCount] range of the shunt compensator ?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/PhaseShifterShiftTap.java
```diff
@@ -55,4 +55,23 @@ private void adjustTapPosition(PhaseTapChanger phaseTapChanger) {
         phaseTapChanger.setTapPosition(Math.min(Math.max(phaseTapChanger.getTapPosition() + tapDelta,
                 phaseTapChanger.getLowTapPosition()), phaseTapChanger.getHighTapPosition()));
     }
+
+    @Override
+    public NetworkModificationImpact hasImpactOnNetwork(Network network) {
+        impact = DEFAULT_IMPACT;
+        TwoWindingsTransformer phaseShifter = network.getTwoWindingsTransformer(phaseShifterId);
+        if (phaseShifter == null || !phaseShifter.hasPhaseTapChanger()) {
+            impact = NetworkModificationImpact.CANNOT_BE_APPLIED;
```
pet-mit, line 64: do we also want to verify that the tap is in the PST's actual tap range?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/PhaseShifterSetAsFixedTap.java
```diff
@@ -42,4 +42,18 @@ public void apply(Network network, NamingStrategy namingStrategy, boolean throwE
         phaseShifter.getPhaseTapChanger().setRegulating(false);
         phaseShifter.getPhaseTapChanger().setRegulationMode(PhaseTapChanger.RegulationMode.FIXED_TAP);
     }
+
+    @Override
+    public NetworkModificationImpact hasImpactOnNetwork(Network network) {
+        impact = DEFAULT_IMPACT;
+        TwoWindingsTransformer phaseShifter = network.getTwoWindingsTransformer(phaseShifterId);
+        if (phaseShifter == null || !phaseShifter.hasPhaseTapChanger()) {
+            impact = NetworkModificationImpact.CANNOT_BE_APPLIED;
```
pet-mit, line 51: same

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/tapchanger/PhaseTapPositionModification.java
```diff
@@ -93,4 +94,28 @@ public void apply(PhaseTapChangerHolder ptcHolder, boolean throwException) {
...
+    @Override
+    public NetworkModificationImpact hasImpactOnNetwork(Network network) {
+        impact = DEFAULT_IMPACT;
+        TwoWindingsTransformer twoWindingsTransformer = network.getTwoWindingsTransformer(getTransformerId());
+        ThreeWindingsTransformer threeWindingsTransformer = network.getThreeWindingsTransformer(getTransformerId());
+        if (twoWindingsTransformer == null && threeWindingsTransformer == null) {
+            impact = NetworkModificationImpact.CANNOT_BE_APPLIED;
+        } else if (twoWindingsTransformer != null) {
+            if (!twoWindingsTransformer.hasPhaseTapChanger()) {
+                impact = NetworkModificationImpact.CANNOT_BE_APPLIED;
```
pet-mit, line 107: same

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/tapchanger/PhaseTapPositionModification.java
```diff
@@ -93,4 +94,28 @@ public void apply(PhaseTapChangerHolder ptcHolder, boolean throwException) {
...
+            } else if (compareValues(getTapPosition(), twoWindingsTransformer.getPhaseTapChanger().getTapPosition(), isRelative)) {
+                impact = NetworkModificationImpact.NO_IMPACT_ON_NETWORK;
+            }
+        } else {
+            PhaseTapChangerHolder ptcHolder = getLeg(threeWindingsTransformer, PhaseTapChangerHolder::hasPhaseTapChanger, false);
+            if (!ptcHolder.hasPhaseTapChanger()) {
+                impact = NetworkModificationImpact.CANNOT_BE_APPLIED;
```
pet-mit, line 114: same

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/tapchanger/RatioTapPositionModification.java
```diff
@@ -66,4 +67,27 @@ public void apply(RatioTapChangerHolder rtcHolder, boolean throwException) {
...
+    @Override
+    public NetworkModificationImpact hasImpactOnNetwork(Network network) {
+        impact = DEFAULT_IMPACT;
+        TwoWindingsTransformer twoWindingsTransformer = network.getTwoWindingsTransformer(getTransformerId());
+        ThreeWindingsTransformer threeWindingsTransformer = network.getThreeWindingsTransformer(getTransformerId());
+        if (twoWindingsTransformer == null && threeWindingsTransformer == null) {
+            impact = NetworkModificationImpact.CANNOT_BE_APPLIED;
+        } else if (twoWindingsTransformer != null) {
+            if (!twoWindingsTransformer.hasRatioTapChanger()) {
+                impact = NetworkModificationImpact.CANNOT_BE_APPLIED;
```
pet-mit, line 79: same

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/tapchanger/RatioTapPositionModification.java
```diff
@@ -66,4 +67,27 @@ public void apply(RatioTapChangerHolder rtcHolder, boolean throwException) {
...
+            } else if (compareValues(getTapPosition(), twoWindingsTransformer.getRatioTapChanger().getTapPosition(), false)) {
+                impact = NetworkModificationImpact.NO_IMPACT_ON_NETWORK;
+            }
+        } else {
+            RatioTapChangerHolder rtcHolder = getLeg(threeWindingsTransformer, RatioTapChangerHolder::hasRatioTapChanger, false);
+            if (!rtcHolder.hasRatioTapChanger()) {
+                impact = NetworkModificationImpact.CANNOT_BE_APPLIED;
```
pet-mit, line 86: same

pet-mit (commented):
Some comments. I'm not very familiar with topology modifications, so I suggest you ask for another opinion

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/NetworkModificationList.java
rolnico: good point

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/PhaseShifterShiftTap.java
rolnico: In the present case it would be impossible since tapPosition is computed as: `tapPosition = Math.min(Math.max(phaseTapChanger.getTapPosition() + tapDelta, phaseTapChanger.getLowTapPosition()), phaseTapChanger.getHighTapPosition())` so the tap is always in the PST's actual tap range

pet-mit: I wonder if it would not be cleaner to remove min/max and return a CANNOT_BE_APPLIED. Because from a user perspective, they won't know that they set, for example, a tap position that is too high, and they'll think that the tap position was accepted as is.
What do you think?

pet-mit: maybe that's not the idea behind this action which is an iterative one. if that's the case then it's ok for me

rolnico: Not sure about the idea behind this network modification.
But if we want to change how it works, it has to be done in another PR since the present one is only to add the checks according to how the network modifications currently work.

pet-mit (approved):

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/AbstractNetworkModification.java
```diff
@@ -91,4 +95,33 @@ protected void logOrThrow(boolean throwException, String message) {
...
+        }
+        return vl != null;
+    }
+
+    protected boolean areValuesEqual(Double newValue, double currentValue, boolean isRelativeValue) {
+        return newValue == null || Math.abs(newValue - (isRelativeValue ? 0 : currentValue)) < EPSILON;
+    }
+
+    protected boolean areValuesEqual(Integer newValue, int currentValue, boolean isRelativeValue) {
+        return newValue == null || Math.abs(newValue - (isRelativeValue ? 0 : currentValue)) < EPSILON;
```
pjeanmarie, line 121: EPSILON not needed for integer

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/AbstractNetworkModification.java
```diff
@@ -91,4 +95,33 @@ protected void logOrThrow(boolean throwException, String message) {
...
+    }
+
+    protected boolean isValueOutsideRange(Integer newValue, int minValue, int maxValue) {
+        return newValue == null || newValue < minValue || newValue > maxValue;
```
pjeanmarie, line 125: newValue == null : here  => isValueOutsideRange true so probably can be applied false while if no value can be applied should be true. (or we can remove the null case and used int)

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/NetworkModificationImpact.java
```diff
@@ -0,0 +1,17 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+package com.powsybl.iidm.modification;
+
+/**
+ * @author Nicolas Rol {@literal <nicolas.rol at rte-france.com>}
+ */
+public enum NetworkModificationImpact {
```
annetill, line 13: Can you add some javadoc here? I can do it if you prefer.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/PhaseShifterShiftTap.java
```diff
@@ -55,4 +55,23 @@ private void adjustTapPosition(PhaseTapChanger phaseTapChanger) {
         phaseTapChanger.setTapPosition(Math.min(Math.max(phaseTapChanger.getTapPosition() + tapDelta,
                 phaseTapChanger.getLowTapPosition()), phaseTapChanger.getHighTapPosition()));
     }
+
+    @Override
+    public NetworkModificationImpact hasImpactOnNetwork(Network network) {
```
annetill, line 60: Indeed I don't understand why we have this `PhaseShifterShiftTap` as we can use the generic `PhaseTapPositionModification`... this is not the object of the PR, but not really understandable.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/ConnectableConnection.java
```diff
@@ -77,6 +77,20 @@ public void apply(Network network, NamingStrategy namingStrategy, boolean throwE
         }
     }
 
+    @Override
+    public NetworkModificationImpact hasImpactOnNetwork(Network network) {
```
annetill, line 81: I still don't understand your check (and for disconnection too). For me, the `side` attribute must be used in this method.

annetill (changes requested):
Some explantations or changes required. But as a general comment, including @pet-mit in the conversation, we have to clearly define what "has impact on network" means. For me it could be:
- IIDM network before applying action is identical as IIDM network after. It is what you have implemented most of the time.
- Or, in a load flow perspective, applying the action will not change anything in the result. 
What is the need in `powsybl-open-rao` ?

annetill:
> Some explantations or changes required. But as a general comment, including @pet-mit in the conversation, we have to clearly define what "has impact on network" means. For me it could be:
> 
> * IIDM network before applying action is identical as IIDM network after. It is what you have implemented most of the time.
> * Or, in a load flow perspective, applying the action will not change anything in the result.
>   What is the need in `powsybl-open-rao` ?



> Some explantations or changes required. But as a general comment, including @pet-mit in the conversation, we have to clearly define what "has impact on network" means. For me it could be:
> 
> * IIDM network before applying action is identical as IIDM network after. It is what you have implemented most of the time.
> * Or, in a load flow perspective, applying the action will not change anything in the result.
>   What is the need in `powsybl-open-rao` ?

Update after a talk with @pet-mit: the scope of this PR is "IIDM network before applying action is identical as IIDM network after". Another value of the enum could be imagined later if needed for LF calculations.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/ConnectableConnection.java
annetill: After a talk with @pet-mit, side must be taken into account, even if there is no impact on powsybl-open-rao.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/NetworkModificationImpact.java
```diff
@@ -8,6 +8,19 @@
...
+ * This enum represents the potential impact of a network modification on a network.
+ * <p>
+ *     The possible values for the enum are:
+ *     <ul>
+ *         <li>{@code CANNOT_BE_APPLIED}: the network modification cannot be applied on this network. If it applied,
+ *         an exception will be thrown.</li>
+ *         <li>{@code NO_IMPACT_ON_NETWORK}: the network modification can be applied to the network but, once the
+ *         network modification is applied, the network will still be the same as before.</li>
+ *         <li>{@code HAS_IMPACT_ON_NETWORK}: the network modification can be applied to the network and, once the
+ *         network modification is applied, the network will be different in its topology or physical characteristics.</li>
```
annetill, line 20: topology, steady states hypothesis or physical characteristics

annetill (approved):

PR #3128, opened 2024-09-03
Fix CGM export dependencies issues

rcourtier:

Bug fix

Currently, when the user performs a CGM export by setting the CGM_EXPORT parameter to true, a common SV model is exported for the main network and updated SSH models for the subnetworks.
However, it has been reported that 2 dependencies are missing:
- The common SV model should be dependentOn the original TP_BD model(s).
- Each of the updated SSH should be dependentOn the original EQ model.

The missing dependencies have been added.
Note that if the same TP_BD model is used by multiple IGM (which is most likely the case), it will appear only once in the SV model as a dependentOn. Also, if the TP_BD model is provided both through the parameter and via the network extension, the parameter will prevail.

- [ ] Yes
- [x] No

**Other information**:


cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CommonGridModelExportTest.java
```diff
@@ -375,8 +384,11 @@ void testCgmExportWithModelsAndProperties() throws IOException {
         String updatedNlSshId = "urn:uuid:Network_NL_N_STEADY_STATE_HYPOTHESIS_2022-03-04T05:30:00Z_4_1D__FM";
         String originalBeTpId = "urn:uuid:Network_BE_N_TOPOLOGY_2022-03-04T05:30:00Z_1_1D__FM";
         String originalNlTpId = "urn:uuid:Network_NL_N_TOPOLOGY_2022-03-04T05:30:00Z_1_1D__FM";
+        String originalBeTpBdId = "urn:uuid:Network_BE_N_TOPOLOGY_BOUNDARY_2022-03-04T05:30:00Z_1_1D__FM";
+        String originalNlTpBdId = "urn:uuid:Network_NL_N_TOPOLOGY_BOUNDARY_2022-03-04T05:30:00Z_1_1D__FM";
         String additionalDependency = "Additional dependency";
-        Set<String> expectedDependencies = Set.of(updatedBeSshId, updatedNlSshId, originalBeTpId, originalNlTpId, additionalDependency);
+        Set<String> expectedDependencies = Set.of(updatedBeSshId, updatedNlSshId, originalBeTpId,
```
zamarrenolm, line 390: here we could add a value for the parameter `BOUNDARY_TP_ID` in the properties and ensure that is also considered in the exported dependencies

zamarrenolm (approved):

PR #3130, opened 2024-09-04
Create adders from existing objects: ReactiveCapabilityCurve, PhaseTapChanger, RatioTapChanger

bc-rte:

This PR introduces adder initialized by copying an existing component.

- [X] Yes
- [ ] No

the `NetworkImpl::newLine` method now has two parametric implementations with one parameter.
Users that used `network.newLine(null)` need to cast null `network.newLine((String) null)` in order to pass compilation

**Other information**:

This PR is inspired from [previous work there](https://github.com/powsybl/powsybl-core/pull/1793/).


flo-dup (changes requested):
The whole mechanism is ok, but the interfaces introduced to think it as a generic solution makes it obscure or too complex I think (and besides it's not completely covering the usecase I think).

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/TemporaryLimitNestedAdderImpl.java
```diff
@@ -0,0 +1,38 @@
+package com.powsybl.iidm.network.impl;
+
+import com.powsybl.iidm.network.AdderUser;
+import com.powsybl.iidm.network.CurrentLimitsAdder;
+import com.powsybl.iidm.network.LoadingLimitsAdder;
+import com.powsybl.iidm.network.TemporaryLimitNestedAdder;
+
+public class TemporaryLimitNestedAdderImpl<ParentAdder extends AdderUser<TemporaryLimitNestedAdder<ParentAdder>>>
+        implements TemporaryLimitNestedAdder<ParentAdder> {
```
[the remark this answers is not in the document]
bc-rte: Moved PR to a smaller subset.

flo-dup (changes requested):
In this PR, we'd better stick with the basic but tedious cases of PhaseTapChanger/RatioTapChanger/ReactiveCapabilityCurve.

.gitignore
```diff
@@ -32,3 +32,6 @@ powsybl.log
 
 # Generated readthedocs pages
 build-docs/
+
+# dotfile for environment variables
+.envrc
```
flo-dup, line 37: Could you remove this from current PR? It's not related to current PR. 

Besides it seems that it's specific to your workflow. In that case you need to add this pattern in your `$GIT_DIR/info/exclude` or in your `~/.gitconfig` (`core.excludesFile` property). If not, feel free to open a PR to add it!

flo-dup (approved):

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/tck/TapChangerHolderTest.java
```diff
@@ -0,0 +1,6 @@
+package com.powsybl.iidm.network.impl.tck;
```
flo-dup, line 1: Author and copyright missing

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractTapChangerHolderTest.java
```diff
@@ -0,0 +1,195 @@
+/**
+ * Copyright (c) 2024, All partners of the iTesla project (http://www.itesla-project.eu/consortium)
```
flo-dup, line 2:
```suggestion
 * Copyright (c) 2024, RTE (http://www.rte-france.com)
```

flo-dup (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/RatioTapChangerHolder.java
```diff
@@ -21,6 +21,34 @@ public interface RatioTapChangerHolder {
...
+                .setRegulationValue(ratioTapChanger.getRegulationValue())
+                .setLoadTapChangingCapabilities(ratioTapChanger.hasLoadTapChangingCapabilities())
+                .setTargetV(ratioTapChanger.getTargetV())
+                .setLowTapPosition(ratioTapChanger.getLowTapPosition())
+                .setTapPosition(ratioTapChanger.getTapPosition())
+                .setRegulating(ratioTapChanger.isRegulating())
+                .setTargetDeadband(ratioTapChanger.getTargetDeadband());
+        for (int tapPosition = ratioTapChanger.getLowTapPosition(); tapPosition <= ratioTapChanger.getHighTapPosition(); tapPosition++) {
+            RatioTapChangerStep step = ratioTapChanger.getStep(tapPosition);
+            adder.beginStep()
```
flo-dup, line 41: wouldn't it be interesting to add a `A beginStep(D step)` in `TapChangerAdder` interface? It's probably not very useful, but in that way each level would have its copy method.

bc-rte: I would prefer to stick to the perimeter you defined previously.

If the need for copy-adder on steps solidifies, they could be added in a separate PR.

bc-rte: After looking back into it, the issue I have with it is that `beginStep` carries a state that would be unrelated to existing step: the order in which steps are created currently encodes the tap position. Being able to initialize `beginStep` by copying a step with a different tap position may be confusing for users.

A helper class could be used to mutualize the code, but it is unclear to me whether this would be better than the current situation.

flo-dup (approved):

PR #3131, opened 2024-09-05
Fix overload detection on reduced limits via LimitViolationUtils

EtienneLt:

No

Bug fix

There are problems about overload detection on reduced limits via `LimitViolationUtils`:
- The temporary limits are sorted in the wrong order.
- In the resulting `Overloading` object:
  - the violated limit's value is the reduced one instead of the original one;
  - the reduction value is not the good one (reduction of the next temporary limit instead of reduction the violated limit).

Fix these problems

- [ ] Yes
- [X] No


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/LimitViolationUtils.java
```diff
@@ -84,16 +82,32 @@ private static Overload getOverload(LimitsContainer<LoadingLimits> limitsContain
         if (Double.isNaN(i) || Double.isNaN(permanentLimit)) {
             return null;
         }
-        Collection<LoadingLimits.TemporaryLimit> temporaryLimits = limitsContainer.getLimits().getTemporaryLimits();
+        List<LoadingLimits.TemporaryLimit> temporaryLimits = limitsContainer.getLimits().getTemporaryLimits()
+            .stream().sorted(Comparator.comparing(LoadingLimits.TemporaryLimit::getValue)
+                .thenComparing(LoadingLimits.TemporaryLimit::getAcceptableDuration, Comparator.reverseOrder())).toList();
```
EtienneLt, line 87: To be checked, I'm not sure we want to have limits with same value

olperr1 (changes requested):

olperr1 (approved):

PR #3132, opened 2024-09-05
Add parameter in NetworkSerDe.copy to choose between formats (xml, binary, etc.)

rolnico:

No

Feature

`NetworkSerDe.copy` currently uses XML format to import/export the network.

A new `NetworkSerDe.copy` taking an additional parameter is introduced. This new parameter `TreeDataFormat format` allows to specify the format to import/export the network.
Using the new parameter with `TreeDataFormat.JSON` or `TreeDataFormat.BIN` as a value, it takes less time to copy network of various sizes:

| Network size | Time with TreeDataFormat.XML | Time with TreeDataFormat.JSON | Time with TreeDataFormat.BIN |
|------------------|-------------------------------------------|--------------------------------------------|--------------------------------------------|
| Very small | 7.1 ms | 3.9 ms | 2.4 ms |
| Small | 351 ms | 180 ms | 170 ms |
| Medium | 784 ms | 412 ms | 267 ms |
| Big | 4526 ms | 1969 ms | 1277 ms |

The copy is now performed by default using the JSON format. (The binary format, which is the most performant, is not yet used because it is still a beta feature.)

- [ ] Yes
- [x] No

**Other information**:


iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -874,14 +874,45 @@ public static Network copy(Network network, NetworkFactory networkFactory) {
...
+    }
+
+    public static Network copy(Network network, NetworkFactory networkFactory, ExecutorService executor, TreeDataFormat format) {
         Objects.requireNonNull(network);
         Objects.requireNonNull(networkFactory);
         Objects.requireNonNull(executor);
+        ExportOptions exportOptions = new ExportOptions();
+        exportOptions.setFormat(format);
+        ImportOptions importOptions = new ImportOptions();
+        importOptions.setFormat(format);
```
flo-dup, line 910: no need to define the options here, you can define them later, only when needed. Plus you can even create them inline,  for instance for export:
```java
write(network, new ExportOptions().setFormat(format), pos);
```

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/NetworkSerDeTest.java
```diff
@@ -116,6 +117,21 @@ void testGzipGunzip() throws IOException {
...
 
+    @Test
+    void testCopyFormat() throws IOException {
+        Network network = createEurostagTutorialExample1();
+        Path file1 = tmpDir.resolve("n.xml");
+        NetworkSerDe.write(network, file1);
+        Network network2 = NetworkSerDe.copy(network);
+        Path file2 = tmpDir.resolve("n2.xml");
+        NetworkSerDe.write(network2, file2);
+        assertArrayEquals(Files.readAllBytes(file1), Files.readAllBytes(file2));
```
flo-dup, line 128: The mismatch would be more understandable if you use `ComparisonUtils.assertTxtEquals` instead.

flo-dup (changes requested):

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -874,14 +874,41 @@ public static Network copy(Network network, NetworkFactory networkFactory) {
     }
 
     public static Network copy(Network network, NetworkFactory networkFactory, ExecutorService executor) {
+        return copy(network, networkFactory, executor, TreeDataFormat.XML);
```
flo-dup, line 877: Binary's still a beta feature, but shouldn't we default to JSON? any opinion @rolnico @geofjamg?

olperr1: JSON by default seems to be a good idea.

flo-dup (approved):

PR #3134, opened 2024-09-06
Add bus id in voltage limit violation

EtienneLt:

no

feature

for voltage violation detection in Security Api, it returns the voltagelevelid

for voltage violation detection in Security Api, returns bus id instead of voltagelevelid
it will be desynchronised with olf security analysis

- [x] Yes
- [ ] No

for voltage limit violation the detection in security will return bus id instead of voltage level id for the subject id


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/LimitViolationUtils.java
```diff
@@ -12,9 +12,7 @@
 import com.powsybl.iidm.network.limitmodification.result.AbstractDistinctLimitsContainer;
 import com.powsybl.iidm.network.limitmodification.result.LimitsContainer;
 
-import java.util.Collection;
-import java.util.Objects;
-import java.util.Optional;
+import java.util.*;
```
vidaldid-rte, line 15: Is this on purpose ?
If not maybe use these IntelliJ settings https://www.baeldung.com/intellij-disable-wildcard-import

It is usually cleaner to explicitely import classes in case of synonims in different packages.

EtienneLt: ok I'll change that

security-analysis/security-analysis-api/pom.xml
```diff
@@ -44,6 +44,14 @@
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/org.junit.platform/junit-platform-launcher -->
+        <dependency>
+            <groupId>org.junit.platform</groupId>
+            <artifactId>junit-platform-launcher</artifactId>
+            <version>1.10.2</version>
+            <scope>test</scope>
```
vidaldid-rte, line 52: Is it possible to avoid a hardcoded version ? Either by using a dependency management entry in a parent POM or a variable ?

EtienneLt: I forgot to remove that 
it is to launch the tests in intellij

vidaldid-rte:
I disagree that it is not a breaking change. It will change the behaviour for users of the security API since they receive a Bus instead of a Voltage Level.
I think that the change is good.
But 
  - it needs to be visible in the documentation.
  - and in the release note.

EtienneLt:
I updated the headers of the merge request

vidaldid-rte:
One more thing about the decision to attach a violation to a bus:
    In a security analysis the list of bus in a contingence can be different from the list of bus in the N state. For example if a swtich is open in a contingence, two bus bars would then belong to different busses.
So this change may produce weird results in node breaker topologies.

Looks like in node breaker, the only valid choice is to attach the violation:
     - To a Voltage Level but with a loss of information
     - To a BusBarSection

EtienneLt:
> One more thing about the decision to attach a violation to a bus: In a security analysis the list of bus in a contingence can be different from the list of bus in the N state. For example if a swtich is open in a contingence, two bus bars would then belong to different busses. So this change may produce weird results in node breaker topologies.
> 
> Looks like in node breaker, the only valid choice is to attach the violation: - To a Voltage Level but with a loss of information - To one or more BusBarSections

Yes but this pr will modify only for loadflow, for security analysis it is not defined yet because the buses on the variant are different 
and the idea is if the is a problem on a violation a loadflow will be computed on the contingency

vidaldid-rte:
> Yes but this pr will modify only for loadflow, for security analysis it is not defined yet because the buses on the variant are different and the idea is if there is a problem on a contingency a loadflow will be computed on the contingency

Security.checkLimits calls LimitViolationDetection.checkAll that calls the modified checkVoltage that attaches the limits to a bus instead of a VL.
So the change impacts security analysis. And it seems to me it may provide wrong results in node breaker topologies.

EtienneLt:
> > Yes but this pr will modify only for loadflow, for security analysis it is not defined yet because the buses on the variant are different and the idea is if there is a problem on a contingency a loadflow will be computed on the contingency
> 
> Security.checkLimits calls LimitViolationDetection.checkAll that calls the modified checkVoltage that attaches the limits to a bus instead of a VL. So the change impacts security analysis. And it seems to me it may provide wrong results in node breaker topologies.

Is it a problem to have the wrong id for security analysis ? 
with the bus id we have the voltage level id so it can be found
and if further information is needed, a modification on the network can be done and a loadflow can be computed

rosiereflo:
I fully agree with @vidaldid-rte remarks on this topic: returning the wrong id for SA is questionable. Moreover external simulation tools like DynaFlow will have hard time providing the proper data here.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/LimitViolationHelper.java
```diff
@@ -26,6 +26,12 @@ private static VoltageLevel getVoltageLevel(LimitViolation limitViolation, Netwo
         Objects.requireNonNull(limitViolation);
 
         Identifiable<?> identifiable = network.getIdentifiable(limitViolation.getSubjectId());
+        if (identifiable == null) {
+            Bus bus = network.getBusView().getBus(limitViolation.getSubjectId());
```
rosiereflo, line 30: External tools such as Hades and DynaFlow will have hard time retrieving the bus name in the NODE_BREAKER case. We need a solution that works both for integrated tools and external tools.

rosiereflo (changes requested):

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/LimitViolationHelper.java
EtienneLt: well I talk with Didier and I'm coding a solution with an enum I will close this mr and open a new one

rosiereflo: ok thanks

EtienneLt:
rework in https://github.com/powsybl/powsybl-core/pull/3153/

EtienneLt:
rework in https://github.com/powsybl/powsybl-core/pull/3153

PR #3135, opened 2024-09-06
CompletableFutureTask, fix spurious error logging when cancelling

jonenst:

NO

bug fix 

When calling cancel(true), the CompletableFutureTask was throwing a NullPointerException in its run method because cancel(true) causes future.get() to throw a CancellationException without a cause (null) and completeExceptionally requires a non null Exception

the exception was thrown to the executor and the very end. Executors usually catch and ignore everything (just logs to stdout) so the problem was just ugly logs and a slight performance drop. But also maybe in some cases an exception with a cause was thrown and the bad exception was used to completeexceptionally the completablefuturetask, not sure.

no exception, and we are sure that we always completeexceptionally with the correct exception

- [ ] Yes
- [x] No

**Other information**:

This was a regression from 5bef72fa5ee, the actual code is reverted to the previous version (except changing throwable to exception to please avoid sonar warnings I guess)

### DIFF
--- a/computation/src/test/java/com/powsybl/computation/CompletableFutureTaskTest.java
+++ b/computation/src/test/java/com/powsybl/computation/CompletableFutureTaskTest.java
@@ -40,8 +40,37 @@ class CompletableFutureTaskTest {
                 Executors.newSingleThreadExecutor(),
                 Executors.newCachedThreadPool(),
                 Executors.newWorkStealingPool(),
+                new MyTestExecutorWithException(),
                 ForkJoinPool.commonPool()
         );
+    }
+
+    // Very basic, spawn a new thread
+    // and allow to wait for the end of the command
+    // just keep exception to be able to assert,
+    // you should only it to launch one command
+    // because it has just one latch and one exception
+    private static class MyTestExecutorWithException implements Executor {
+
+        Exception exception = null;
+        CountDownLatch waitForDone;
+
+        @Override
+        public void execute(Runnable command) {
+            (new Thread() {
+                @Override
+                public void run() {
+                    waitForDone = new CountDownLatch(1);
+                    try {
+                        command.run();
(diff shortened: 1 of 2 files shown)


geofjamg (approved):

computation/src/test/java/com/powsybl/computation/CompletableFutureTaskTest.java
```diff
@@ -40,10 +40,39 @@ private static List<Executor> executors() {
...
+        @Override
+        public void execute(Runnable command) {
+            (new Thread() {
+                @Override
+                public void run() {
+                    waitForDone = new CountDownLatch(1);
+                    try {
+                        command.run();
+                    } catch (Exception exception) {
+                        MyTestExecutorWithException.this.exception = exception;
```
rolnico, line 67:
```suggestion
                    } catch (Exception e) {
                        MyTestExecutorWithException.this.exception = e;
```

computation/src/test/java/com/powsybl/computation/CompletableFutureTaskTest.java
```diff
@@ -40,10 +40,39 @@ private static List<Executor> executors() {
...
+                new MyTestExecutorWithException(),
                 ForkJoinPool.commonPool()
         );
     }
 
+    // Very basic, spawn a new thread
+    // and allow to wait for the end of the command
+    // just keep exception to be able to assert,
+    // you should only it to launch one command
+    // because it has just one latch and one exception
```
rolnico, line 52:
```suggestion
    // Very basic executor that spawns a new thread
    // and allows to wait for the end of the command.
    // It just keeps an exception to be able to assert it.
    // You should use it to launch only one command
    // because it has just one latch and one exception
```

rolnico (changes requested):

PR #3147, opened 2024-09-19
CgmesImporter, allow importing with wrong dataextension when the mainfile doesn't exist

jonenst:

NO

Bug fix

can't import a cgmes zip with a dot in the name and not ".xml.zip" as the extension (e.g. microgrid.v2.complete.zip)

can import

- [ ] Yes
- [x] No

**Other information**:
This happens when filenames have the 2 following properties:
- the filename contains a dot:
- the filename does not contain the dataextension For example this filename triggers the bug:  "microgrid.v2.complete.zip" but these work:
   "microgrid_v2_complete.zip" (because empty dataextension)
or "microgrid.v2.complete.xml.zip" (because ".xml" dataextension)

The problem is that the main extension is derived as ".complete" but the cgmes importer only accepts an empty dataextension or ".xml" (it checks the dataextension to avoid importing as cgmes from a datasource that was meant by the user to be used as another format). But for now no other importer works without a main file, so when the mainfile doesn't exist so we shouldn't block the import just like when the dataextension is not specified

NOTE: this happens because we want to allow archive filenames to be very flexible (with or without the dataextension, and for cgmes even with a totally different basename as what is inside).

NOTE: if we ever have another importer that works like the cgmes importer, this means we will not be able to handle the following case: archive/network_EQ.xml
archive/network_TP.xml # cgmes files
archive/network_foo.other
archive/network_bar.other # some new importer format and be able to import from a single directory containing both networks by specifying the dataextension. But this should be a rare case, and moving the data to separate directories or using separate basenames works around the problem.

### DIFF
--- a/cgmes/cgmes-model/src/test/java/com/powsybl/cgmes/model/CgmesOnDataSourceTest.java
+++ b/cgmes/cgmes-model/src/test/java/com/powsybl/cgmes/model/CgmesOnDataSourceTest.java
@@ -33,7 +33,7 @@ import static org.junit.jupiter.api.Assertions.assertFalse;
  */
 class CgmesOnDataSourceTest {
 
-    private static String XIIDM_XML_NOT_CGMES = "<?xml version='1.0' encoding='UTF-8'?><some></some>";
+    private final static String XIIDM_XML_NOT_CGMES = "<?xml version='1.0' encoding='UTF-8'?><some></some>";
 
     static Stream<Arguments> provideArguments() {
         return Stream.of(
--- a/cgmes/cgmes-model/src/test/java/com/powsybl/cgmes/model/CgmesOnDataSourceTest.java
+++ b/cgmes/cgmes-model/src/test/java/com/powsybl/cgmes/model/CgmesOnDataSourceTest.java
@@ -13,7 +13,6 @@ import com.powsybl.commons.datasource.ReadOnlyDataSource;
 import com.powsybl.commons.datasource.ResourceDataSource;
 import com.powsybl.commons.datasource.ResourceSet;
 import com.powsybl.commons.datasource.ZipArchiveDataSource;
-import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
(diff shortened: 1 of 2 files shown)


cgmes/cgmes-model/src/test/java/com/powsybl/cgmes/model/CgmesOnDataSourceTest.java
```diff
@@ -58,26 +59,127 @@ void testExists(String testName, String filename, String cimVersion, boolean exp
         assertEquals(expectedExists, exists);
     }
 
-    @Test
-    void testFileDoesNotExist() throws IOException {
+    static Stream<Arguments> provideArgumentsFortestXmlMainFileXiidmZip() {
```
flo-dup, line 62:
```suggestion
    static Stream<Arguments> provideArgumentsForTestXmlMainFileXiidmZip() {
```

cgmes/cgmes-model/src/test/java/com/powsybl/cgmes/model/CgmesOnDataSourceTest.java
```diff
@@ -34,6 +33,8 @@
  */
 class CgmesOnDataSourceTest {
 
+    private static String XIIDM_XML_NOT_CGMES = "<?xml version='1.0' encoding='UTF-8'?><some></some>";
```
flo-dup, line 36:
```suggestion
    private final static String XIIDM_XML_NOT_CGMES = "<?xml version='1.0' encoding='UTF-8'?><some></some>";
```

cgmes/cgmes-model/src/test/java/com/powsybl/cgmes/model/CgmesOnDataSourceTest.java
```diff
@@ -58,26 +59,127 @@ void testExists(String testName, String filename, String cimVersion, boolean exp
...
+    static Stream<Arguments> provideArgumentsFortestXmlMainFileXiidmZip() {
+        return Stream.of(
+                Arguments.of("foo", "xml"),
+                Arguments.of("foo", null),
+                Arguments.of("foo", ""),
+                Arguments.of("foo", "notexists"),
+                Arguments.of("bar", "xml"),
+                Arguments.of("bar", null),
+                Arguments.of("bar", ""),
+                Arguments.of("bar", "notexists")
```
flo-dup, line 71: are they useful? I think each one corresponds to the same logic as one of the previous arguments.

cgmes/cgmes-model/src/test/java/com/powsybl/cgmes/model/CgmesOnDataSourceTest.java
```diff
@@ -58,26 +59,127 @@ void testExists(String testName, String filename, String cimVersion, boolean exp
...
+                Arguments.of("kop", "xml", true),
+                Arguments.of("kop", "xiidm", false),
+                Arguments.of("kop", "notexists", true),
+                Arguments.of("kop", "", true),
+                Arguments.of("kop", null, true),
+                Arguments.of("notexist", "xml", true),
+                Arguments.of("notexist", "xiidm", true),
+                Arguments.of("notexist", "notexists", true),
+                Arguments.of("notexist", "", true),
+                Arguments.of("notexist", null, true)
```
flo-dup, line 124: Not sure those are needed

flo-dup (approved):

PR #3168, opened 2024-10-04
Add getI() method to DanglingLine's boundary

caioluke:

Feature

Boundary nodes do not have a `getI()` method, whereas terminals do. Therefore, for paired dangling lines, we can not get the current from both sides of the half line easily.

Add the `getI()` method for boundary nodes.

- [X] Yes
- [ ] No

Custom IIDM implementations that are implementing the `Boundary` interface (which is not necessarily the case) should add the following method to their implementation:
- `double getI();`


jeandemanged:
Hi @olperr1 ,
One question, I am not so sure this should be flagged "Breaking Change". Unless there are iIDM implementations around overriding the default "utils" calculations that are built-in the iIDM API. But I don't think implementations should do that ...
Your opinion ?

jeandemanged (approved):

olperr1:
I double-checked the `Boundary` interface usage and I found that `powsybl-network-store` has [its own implementation](https://github.com/powsybl/powsybl-network-store/blob/6e9498557fef4241b75ac6897dd078d5ce8a42aa/network-store-iidm-impl/src/main/java/com/powsybl/network/store/iidm/impl/DanglingLineBoundaryImpl.java#L18) (which is very similar to the powsybl-core one, but its an other problem). So I will report this change as a breaking change.

olperr1 (approved):

PR #3170, opened 2024-10-07
Add tgz option to iTools packager Maven plugin

gverger:

This PR adds an option for the packager to use a tar.gz compressor instead of zip. Zip is still the default.

There is currently no option to chose the file type when packaging itools, it is zipped.

A new option in the maven plugin allows for compressing with tar gz.

- [ ] Yes
- [x] No

**Other information**:

I don't know how to test this. The only thing I did was testing on our project depending on powsybl-core.


itools-packager/src/main/java/com/powsybl/itools/ItoolsPackagerMojo.java
```diff
@@ -64,6 +67,9 @@ public void setFiles(File[] files) {
         }
     }
 
+    @Parameter(defaultValue = "zip")
+    private String fileType;
```
jeandemanged, line 71: I would rename to `packageType`

itools-packager/src/main/java/com/powsybl/itools/ItoolsPackagerMojo.java
```diff
@@ -181,7 +217,13 @@ public void execute() {
 
             getLog().info("Zip package");
```
jeandemanged, line 218: move this log under the `zip` case, add a new `getLog().info("Tgz package");` for the tgz case

jeandemanged (changes requested):

jeandemanged (dismissed):

olperr1:
> thanks, tested manually/locally on my side and works good, @olperr1 let us know if you have ideas about automated testing of this maven plugin ...

It should be possible to test the plugin by using [Maven Plugin Testing Harness](https://maven.apache.org/plugin-testing/maven-plugin-testing-harness/getting-started/index.html). With this mechanism, one could retrieve the Mojo loaded with a test `pom.xml` configuration, and then call the `execute()` method on it.

I tried a little, but didn't succeed. Maybe the problem is related to the `package-zip` goal used by our plugin...

itools-packager/src/main/java/com/powsybl/itools/ItoolsPackagerMojo.java
```diff
@@ -100,6 +107,37 @@ public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) th
         }
     }
 
+    private void targz(Path dir, Path baseDir, Path zipFilePath) throws IOException {
+        getLog().info("Targz package");
+        try (TarArchiveOutputStream zos =
+                new TarArchiveOutputStream(new GzipCompressorOutputStream(Files.newOutputStream(zipFilePath)))) {
+            zos.setLongFileMode(TarArchiveOutputStream.LONGFILE_POSIX); // allow long file paths
```
olperr1, line 114: The test is failing on my computer with the following error:
```
java.lang.IllegalArgumentException: group id '17.......4' is too big ( > 2097151 ). Use STAR or POSIX extensions to overcome this limit
```
<br/>

I have to add the following option to successfully run the test. Is it OK for you to add it?

```java
zos.setBigNumberMode(TarArchiveOutputStream.BIGNUMBER_POSIX); // allow large numbers (for instance a big GID)
```

olperr1 (changes requested):

itools-packager/src/main/java/com/powsybl/itools/ItoolsPackagerMojo.java
gverger: Nicely spotted, I'll add it yes 👍

olperr1 (approved):

itools-packager/pom.xml
```diff
@@ -67,6 +67,34 @@
             <version>${maven.plugin.annotations.version}</version>
             <scope>provided</scope>
         </dependency>
+        <!-- Test dependencies -->
+        <dependency>
+            <groupId>org.apache.maven.plugin-testing</groupId>
+            <artifactId>maven-plugin-testing-harness</artifactId>
+            <version>3.3.0</version>
```
olperr1, line 74: Could you use a variable, defined in the main `pom.xml`, instead of using directly the version number?

itools-packager/pom.xml
```diff
@@ -67,6 +67,34 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.maven</groupId>
+            <artifactId>maven-compat</artifactId>
+            <version>3.9.9</version>
```
olperr1, line 95: Same here.

olperr1 (changes requested):

flo-dup (approved):

PR #3176, opened 2024-10-08
fix: OverloadManagementSystem did not serialize to binary when a trip…

bc-rte:

No

Bug fix for serialization issue on overload management systems.

When an `OverloadManagementSystem` with a nameless tripping is serialized to binary and deserialized again, the deserialization crashes when trying to read the `name attribute`.

This is caused by the writer not writing `name`, thus not creating a "placeholder" for this field in the binary format.
This in turn causes the BinReader trying to read the number of bytes for the `name` string, to read aberrant data instead and throw.

`name` attribute is written even when it is null.

Also improved error message for future generations :)

- [ ] Yes
- [X] No

### DIFF
--- a/commons/src/main/java/com/powsybl/commons/binary/BinReader.java
+++ b/commons/src/main/java/com/powsybl/commons/binary/BinReader.java
@@ -76,7 +76,8 @@ public class BinReader implements TreeDataReader {
             }
             byte[] stringBytes = dis.readNBytes(stringNbBytes);
             if (stringBytes.length != stringNbBytes) {
-                throw new PowsyblException("Cannot read the full string, bytes missing: " + (stringNbBytes - stringBytes.length));
+                throw new PowsyblException("Cannot read the full string, bytes missing: " + (stringNbBytes - stringBytes.length)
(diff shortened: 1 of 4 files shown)


commons/src/main/java/com/powsybl/commons/binary/BinReader.java
```diff
@@ -76,7 +76,8 @@ private String readString() {
             }
             byte[] stringBytes = dis.readNBytes(stringNbBytes);
             if (stringBytes.length != stringNbBytes) {
-                throw new PowsyblException("Cannot read the full string, bytes missing: " + (stringNbBytes - stringBytes.length));
+                throw new PowsyblException("Cannot read the full string, bytes missing: " + (stringNbBytes - stringBytes.length)
+                        + " (this may happen when the attribute wasn't written in the first place, causing string length to be an aberrant number)");
```
olperr1, line 80: This should not be reported in the exception. But it would be useful to keep it as a comment in the code.

olperr1 (changes requested):

olperr1 (approved):

PR #3153, opened 2024-09-23
Add the violation location in voltage LimitViolations

EtienneLt:
**Please check if the PR fulfills these requirements**

- [X] The commit message follows our guidelines
- [X] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**Does this PR already have an issue describing the problem?**
No

**What kind of change does this PR introduce?**
Feature

**What is the new behavior (if this is a feature change)?**

Add the violation location in `LimitViolation` for voltage limits violation:
- in NodeBreaker: the voltage level id + the ids of the busbars;
- in BusBreaker: the voltage level id + the bus id.

\+ fill these informations when using Security api.

**Does this PR introduce a breaking change or deprecate an API?**
- [ ] Yes
- [X] No

PR #3178, opened 2024-10-09
Return stable ids/nodes in ViolationLocation classes

olperr1:

No

Bug fix

#3153 introduces ViolationLocations for voltage violation.
- In the bus/breaker topology case, the bus id in the `BusBreakerViolationLocation` corresponds to a merged bus. This is not stable depending on the network implementation.
- In the node/breaker topology case, the `NodeBreakerViolationLocation` contains a list of busbar section ids. But in some cases there are no busbar section where the violation is detected.

- The `busId` of `BusBreakerViolationLocation` was replaced by a list of ids corresponding to the configured buses of the merge bus.
- The `busbarIds` of `NodeBreakerViolationLocation` was replaced by the list of the nodes where the violation was detected (always present).

Some unnecessary attributes / methods were also removed (for quality purpose):
- `getId()` (from `VoltageLocation` interface + both implementations)
- `getVoltageLevelId()` (from `VoltageLocation` + `BusBreakerViolationLocation`)
- `getBusId()` (from `VoltageLocation` interface + both implementations)
- `getBusbarIds()` (from `VoltageLocation` + both implementations)

- [X] Yes: There's breaking change compared to `powsybl:6.5.0-RC1`, **but not compared to any production release**.
- [ ] No

Only for migrating from 6.5.0-RC1:
- To retrieve data from a `ViolationLocation`, you now have to check it's type and cast it in the corresponding implementation.
- The `getBusId()` method from `BusBreakerViolationLocation` is replaced by a `getBusIds()` method returning the list of the ids of the **configured** buses instead of the sole id of the **merged** bus (which is not stable).
- The `getBusbarIds()` method from `NodeBreakerViolationLocation` is replaced by a `getNodes()` method returning the list of the nodes where the violation was detected. If needed, it is possible to retrieve the list of all concerned busbar sections using the following code:
```java
NodeBreakerViolationLocation vloc = (NodeBreakerViolationLocation) violationLocation;
VoltageLevel vl = network.getVoltageLevel(vloc.getVoltageLevelId());
List<String> busbarSectionIds = vloc.getNodes().stream()
        .map(node -> vl.getNodeBreakerView().getTerminal(node))
        .filter(Objects::nonNull)
        .map(Terminal::getConnectable)
        .filter(t -> t.getType() == IdentifiableType.BUSBAR_SECTION)
        .map(Identifiable::getId)
        .toList();
```

**Other information**:

The serialization versions of the short-circuits and security analysis results were increased in #3153.
The current version doesn't increase them since the said PR isn't already in a production release.

### DIFF
--- a/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/BusBreakerViolationLocation.java
+++ b/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/BusBreakerViolationLocation.java
@@ -17,7 +17,7 @@ public class BusBreakerViolationLocation implements ViolationLocation {
     private final List<String> busIds;
 
     public BusBreakerViolationLocation(List<String> busIds) {
-        this.busIds = Objects.requireNonNull(busIds, "'busIds' should not be null.");
+        this.busIds = Objects.requireNonNull(busIds, "busIds should not be null.");
     }
 
     public List<String> getBusIds() {
--- a/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/BusBreakerViolationLocation.java
+++ b/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/BusBreakerViolationLocation.java
@@ -7,6 +7,8 @@
  */
 package com.powsybl.security;
 
+import com.powsybl.iidm.network.Network;
+
 import java.util.List;
 import java.util.Objects;
 
@@ -43,4 +45,20 @@ public class BusBreakerViolationLocation implements ViolationLocation {
             "busIds='" + busIds + '\'' +
             '}';
     }
+
+    @Override
+    public BusView getBusView(Network network) {
+        return () -> busIds.stream()
+                .map(id -> network.getBusBreakerView().getBus(id))
+                .filter(b -> b.getConnectedTerminalCount() > 0)
+                .map(b -> b.getConnectedTerminals().iterator().next().getBusView().getBus())
+                .distinct();
+
+    }
+
+    @Override
+    public BusView getBusBreakerView(Network network) {
+        return () -> busIds.stream().map(id -> network.getBusBreakerView().getBus(id));
+    }
+
 }
--- a/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/BusBreakerViolationLocation.java
+++ b/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/BusBreakerViolationLocation.java
@@ -16,10 +16,18 @@ import java.util.Objects;
 public class BusBreakerViolationLocation implements ViolationLocation {
     private final List<String> busIds;
 
+    /**
+     * Create a ViolationLocation for a violation detected in a voltage level in bus/breaker topology.
+     * @param busIds The ids of the <b>configured</b> buses (of the bus/breaker view) where the violation was detected.
+     */
     public BusBreakerViolationLocation(List<String> busIds) {
         this.busIds = Objects.requireNonNull(busIds, "busIds should not be null.");
     }
 
+    /**
+     * Get the ids of the <b>configured</b> buses (of the bus/breaker view) where the violation was detected.
+     * @return the configured bus ids
+     */
     public List<String> getBusIds() {
         return busIds;
     }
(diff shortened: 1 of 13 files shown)


security-analysis/security-analysis-api/src/main/java/com/powsybl/security/NodeBreakerViolationLocation.java
```diff
@@ -18,26 +18,18 @@ public class NodeBreakerViolationLocation implements ViolationLocation {
     private final List<String> busBarIds;
 
     public NodeBreakerViolationLocation(String voltageLevelId, List<String> busBarIds) {
-        Objects.requireNonNull(voltageLevelId, "voltageLevelId");
-        this.voltageLevelId = voltageLevelId;
-        this.busBarIds = busBarIds;
+        this.voltageLevelId = Objects.requireNonNull(voltageLevelId, "voltageLevelId should not be null");
```
EtienneLt, line 21: do not we remove the voltageLevelId in Location ?

geofjamg: We could as it is redundant we busbar section Ids which allows to get the voltage level directly BUT 
 - using the busbar section is a design issue too
 - we should use nodes (integer) instead to covers all use cases and be symmetrical with what is in the  BusBreakerViolationLocation

So maybe we should keep it and in a next PR, remove busbar section and replace by nodes.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/NodeBreakerViolationLocation.java
```diff
@@ -18,26 +18,18 @@ public class NodeBreakerViolationLocation implements ViolationLocation {
     private final List<String> busBarIds;
```
geofjamg, line 18: busbarSectionIds

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/NodeBreakerViolationLocation.java
geofjamg: Or maybe we should do it now?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/BusBreakerViolationLocation.java
```diff
@@ -7,35 +7,21 @@
...
 package com.powsybl.security;
 
+import java.util.List;
 import java.util.Objects;
-import java.util.Optional;
 
 /**
  * @author Étienne Lesot {@literal <etienne.lesot at rte-france.com>}
  */
 public class BusBreakerViolationLocation implements ViolationLocation {
```
geofjamg, line 16: We should write in the Java doc that it is Ids of configured buses of the bus/breaker view that are expected.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/NodeBreakerViolationLocation.java
geofjamg: To avoid a painful migration

geofjamg (approved):

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/ViolationLocation.java
```diff
@@ -21,18 +17,5 @@ enum Type {
...
-
-    default Optional<String> getBusId() {
-        return Optional.empty();
-    }
-
-    default List<String> getBusBarIds() {
-        return Collections.emptyList();
-    }
-
 }
```
vidaldid-rte, line 21: Sans méthode, l'API sera dure à utiliser sans interface.

On pourrait ajouter getBusView(Network).getBusStream()
et getBusBreakerView(Network).getBusStream()

vidaldid-rte: @olperr1 @geofjamg small addition to make API more user friendly
https://github.com/powsybl/powsybl-core/pull/3178/commits/aa833ab7c1f2ea35599f613d39c31f8fd7c872c9

Can you have a look ?

olperr1: Thanks for your addition!

vidaldid-rte (approved):

PR #3188, opened 2024-10-22
Fix proportional scalable infinite loop

jeandemanged:

No

Bug fix

Infinite loop in proportional scalable under the following condition:
- asking a little bit more power than the full available power range over all scalables
- scalable percentages are low enough so that the difference of asked power minus done power on one scalable falls below the 0.01 MW epsilon

In this case the iteration percentage of the scalable is never set to zero, there is remaining asked power, and the scaling is stuck in an infinite loop.

May sound like a marginal case, but was observed on real data. On large networks a given load may have a small percentage.

We set iteration percentage to zero when a scalable is saturated, in addition to the existing condition checking that not all asked power has been done.

- [x] No

**Other information**:

### DIFF
--- a/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/scalable/ProportionalScalable.java
+++ b/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/scalable/ProportionalScalable.java
@@ -244,7 +244,7 @@ public class ProportionalScalable extends AbstractCompoundScalable {
     private double iterativeScale(Network n, double asked, ScalingParameters parameters) {
         double done = 0;
         double doneChange = Double.MAX_VALUE;
-        while ((Math.abs(asked - done) > EPSILON && notSaturated()) && (Math.abs(doneChange) > EPSILON)) {
+        while (Math.abs(asked - done) > EPSILON && notSaturated() && Math.abs(doneChange) > EPSILON) {
             checkIterationPercentages();
             doneChange = scaleIteration(n, asked - done, parameters);
             done += doneChange;
--- a/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/scalable/ProportionalScalable.java
+++ b/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/scalable/ProportionalScalable.java
@@ -243,11 +243,9 @@ public class ProportionalScalable extends AbstractCompoundScalable {
 
     private double iterativeScale(Network n, double asked, ScalingParameters parameters) {
         double done = 0;
-        double doneChange = Double.MAX_VALUE;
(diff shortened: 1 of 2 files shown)


phiedw (approved):
Thank you very much for this pull request

jeandemanged:
> What should the behaviour be in the following case? (maybe too much of a marginal case for us to worry about)
> 
> Initially load1 is at 0MW, load2 and load3 are at 100MW, and the coefficients on the different loads are ~100, 1e-5 and 1e-5.
> 
> If we ask for a 100MW reduction, with the new code, we will scale by 2e-3MW at the first iteration and will exit the loop. Before, we would have scaled by 2e-3MW at the first iteration, renormalized the coefficients (since load1 is saturated), and then at the second iteration, scaled by 50MW on both loads 2 and 3.

indeed reverted the change causing this, was a bad idea, but maybe could be made different, e.g. if difference `asked-done` doesn't 'move' anymore by more than EPSILON between two iterations, then break the loop.

I am a bit worried of remaining very small values summed, summing many small floats can be surprising sometimes.

You opinion @phiedw ?

phiedw:
I think this fix seems to be a decent way to fix the issue. I can't think of any cases where the behaviour would not be the expected one.
I was wondering if it could be possible to add to the scalable interface a method that checks some conditions to know if the scalable is saturated/disabled/disconnected etc to know if we should set its percentage to 0, but I'm not sure it would be cleaner.

jeandemanged:
> I think this fix seems to be a decent way to fix the issue. I can't think of any cases where the behaviour would not be the expected one. I was wondering if it could be possible to add to the scalable interface a method that checks some conditions to know if the scalable is saturated/disabled/disconnected etc to know if we should set its percentage to 0, but I'm not sure it would be cleaner.

yes I am not sure as well refactoring would make things easier

I was also wondering if to be on the safe side we should do as below, but cannot build a unit test for that.

saving it here so it doesn't get lost:

```java
    private double iterativeScale(Network n, double asked, ScalingParameters parameters) {
        double done = 0;
        boolean first = false;
        while (Math.abs(asked - done) > EPSILON && notSaturated()) {
            checkIterationPercentages();
            double previousAskedMinusDone = asked - done;
            done += scaleIteration(n, asked - done, parameters);
            double newAskedMinusDone = asked - done;
            if (Math.abs(previousAskedMinusDone - newAskedMinusDone) < EPSILON) {
                if (!first) {
                    first = true;
                } else {
                    LOGGER.warn("Proportional iterative scale did not distribute more than {} MW between 2 consecutive iterations, " +
                            "stopping iterative scale. asked={} MW, done={} MW", EPSILON, asked, done);
                    break;
                }
            } else {
                first = false;
            }
            updateIterationPercentages();
        }
        return done;
    }
```

olperr1 (approved):

phiedw (approved):

PR #3192, opened 2024-10-25
Network modification to set generator regulation to local instead of remote

rcourtier:

No

Feature: Network modification to change the regulation mode for a generator.

Generators can regulate remotely.

When the modification is applied, the specified generator regulates locally. It keeps the same target value in per unit.

- [ ] Yes
- [X] No

**Other information**:

See https://github.com/powsybl/powsybl-core/pull/3198 for setting the regulation mode to local only in the CGMES exported files and not in the IIDM model.

### DIFF
--- a/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/SetGeneratorToLocalRegulation.java
+++ b/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/SetGeneratorToLocalRegulation.java
@@ -45,10 +45,7 @@ public class SetGeneratorToLocalRegulation extends AbstractNetworkModification {
     @Override
     public void apply(Network network, NamingStrategy namingStrategy, boolean throwException, ComputationManager computationManager, ReportNode reportNode) {
         Generator generator = network.getGenerator(generatorId);
-        if (generator != null
-                && generator.getTerminal() != null
-                && generator.getRegulatingTerminal() != null
-                && generator.getTerminal() != generator.getRegulatingTerminal()) {
+        if (generator != null && !generator.getId().equals(generator.getRegulatingTerminal().getConnectable().getId())) {
             setLocalRegulation(generator, reportNode);
         }
     }
@@ -77,9 +74,9 @@ public class SetGeneratorToLocalRegulation extends AbstractNetworkModification {
     @Override
     public NetworkModificationImpact hasImpactOnNetwork(Network network) {
         Generator generator = network.getGenerator(generatorId);
-        if (generator == null || generator.getTerminal() == null || generator.getRegulatingTerminal() == null) {
+        if (generator == null) {
             impact = NetworkModificationImpact.CANNOT_BE_APPLIED;
-        } else if (generator.getTerminal() == generator.getRegulatingTerminal()) {
+        } else if (generator.getId().equals(generator.getRegulatingTerminal().getConnectable().getId())) {
             impact = NetworkModificationImpact.NO_IMPACT_ON_NETWORK;
         } else {
             impact = DEFAULT_IMPACT;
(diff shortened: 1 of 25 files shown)


colinepiloquet:
It looks good to me and it's perfect for my use case. I am just thinking, in case we want to be even more generic, that the network modification could take a list of generator IDs. Like that, not all the generators would be impacted but as I don't know if there is a use-case for this, I don't know if it's worth the effort, what do you think?

rcourtier:
> It looks good to me and it's perfect for my use case. I am just thinking, in case we want to be even more generic, that the network modification could take a list of generator IDs. Like that, not all the generators would be impacted but as I don't know if there is a use-case for this, I don't know if it's worth the effort, what do you think?

I don't think it's a big effort. I will work on it.

colinepiloquet (approved):

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/SetGeneratorToLocalRegulation.java
```diff
@@ -0,0 +1,89 @@
...
+        return "SetGeneratorToLocalRegulation";
+    }
+
+    @Override
+    public void apply(Network network, NamingStrategy namingStrategy, boolean throwException, ComputationManager computationManager, ReportNode reportNode) {
+        Generator generator = network.getGenerator(generatorId);
+        if (generator != null
+                && generator.getTerminal() != null
+                && generator.getRegulatingTerminal() != null
+                && generator.getTerminal() != generator.getRegulatingTerminal()) {
```
olperr1, line 51: Tests on references may fail depending on the used IIDM implementation. To avoid problems, these tests should be done using IDs. In this case, you could use the terminal's connectable ID.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/SetGeneratorToLocalRegulation.java
```diff
@@ -0,0 +1,89 @@
...
+        LOG.info("Changed regulation for generator: {} to local instead of remote", generator.getId());
+        generatorLocalRegulationReport(reportNode, generator.getId());
+    }
+
+    @Override
+    public NetworkModificationImpact hasImpactOnNetwork(Network network) {
+        Generator generator = network.getGenerator(generatorId);
+        if (generator == null || generator.getTerminal() == null || generator.getRegulatingTerminal() == null) {
+            impact = NetworkModificationImpact.CANNOT_BE_APPLIED;
+        } else if (generator.getTerminal() == generator.getRegulatingTerminal()) {
```
olperr1, line 82: Same thing here.

olperr1 (changes requested):

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/SetGeneratorToLocalRegulation.java
rcourtier: Done.
I have also removed the ```generator.getRegulatingTerminal() != null``` checks since the regulatingTerminal is never null (if not set at initialization or when explicitely attempting to set it to null, it is always reset to the generator's own terminal)

olperr1 (changes requested):

iidm/iidm-modification/src/test/java/com/powsybl/iidm/modification/SetGeneratorToLocalRegulationTest.java
```diff
@@ -0,0 +1,149 @@
+package com.powsybl.iidm.modification;
```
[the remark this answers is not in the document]
rcourtier: fixed!

iidm/iidm-modification/src/test/java/com/powsybl/iidm/modification/SetGeneratorToLocalRegulationTest.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    @Test
+    void getNameTest() {
+        SetGeneratorToLocalRegulation modification = new SetGeneratorToLocalRegulation("GEN1");
+        assertEquals("SetGeneratorToLocalRegulation", modification.getName());
+    }
+
+    private Network createTestNetwork() {
+        Network network = Network.create("test_network", "test");
```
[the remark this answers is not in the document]
rcourtier: fixed!

olperr1 (approved):

PR #3199, opened 2024-11-05
Voltage level topology model conversion

geofjamg:

No

Feature

We are able to change the topology kind of voltage level. For now we are only capable to convert a node/breaker voltage level topology to a bus/breaker one.

- [ ] Yes
- [x] No

**Other information**:


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractConnectable.java
```diff
@@ -178,7 +178,20 @@ protected void move(TerminalExt oldTerminal, TopologyPoint oldTopologyPoint, Str
         attachTerminal(oldTerminal, oldTopologyPoint, (VoltageLevelExt) bus.getVoltageLevel(), terminalExt);
     }
 
-    private void attachTerminal(TerminalExt oldTerminal, TopologyPoint oldTopologyPoint, VoltageLevelExt voltageLevel, TerminalExt terminalExt) {
+    void replaceTerminal(TerminalExt oldTerminal, TopologyPoint oldTopologyPoint, TerminalExt newTerminalExt, boolean notify) {
```
flo-dup, line 181: This could probably be merged with attachTerminal, at the cost of adding one boolean and renaming it (I think it's more a replace than an attach). The referrer part in VoltageLevelImpl should also be included - that would fix the revealed bug when moving a connectable!

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AreaBoundaryImpl.java
```diff
@@ -19,9 +19,9 @@ public class AreaBoundaryImpl implements AreaBoundary {
 
     final Area area;
 
-    final Terminal terminal;
+    Terminal terminal;
```
flo-dup, line 22: we could add package protected methods replaceTerminal / replaceBoundary instead, to avoid direct modification of the field from AreaImpl.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VoltageLevelImpl.java
```diff
@@ -618,4 +619,91 @@ public void allocateVariantArrayElement(int[] indexes, int sourceIndex) {
         super.allocateVariantArrayElement(indexes, sourceIndex);
         topologyModel.allocateVariantArrayElement(indexes, sourceIndex);
     }
+
+    @Override
+    public void convertToTopology(TopologyKind newTopologyKind) {
+        Objects.requireNonNull(newTopologyKind);
+        if (newTopologyKind != topologyModel.getTopologyKind()) {
+            if (newTopologyKind == TopologyKind.NODE_BREAKER) {
+                throw new PowsyblException("Topology model conversion from bus/breaker to node/breaker not yet supported");
```
flo-dup, line 628: To decrease complexity and indentations, you could add a return if the two kinds are the same, a return if the newTopologyKind is node breaker and remove the else (as the enum only has two values).

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VoltageLevelImpl.java
```diff
@@ -618,4 +619,91 @@ public void allocateVariantArrayElement(int[] indexes, int sourceIndex) {
...
+                }
+
+                // reconnect all connectable to new topology model
+                // first store all bus/breaker topological infos associated to this terminal because we will start moving
+                // terminal from old mode to new one, it will modify the old topology model
+                List<Triple<TerminalExt, Bus, Boolean>> oldTopologyModelInfos = new ArrayList<>();
+                for (Terminal oldTerminal : topologyModel.getTerminals()) {
+                    Bus connectableBus = oldTerminal.getBusBreakerView().getConnectableBus();
+                    boolean connected = oldTerminal.isConnected();
+                    oldTopologyModelInfos.add(Triple.of((TerminalExt) oldTerminal, connectableBus, connected));
```
flo-dup, line 654: you could use a record instead, and as the bus is only needed for the id, you'd better store the id string directly.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VoltageLevelImpl.java
```diff
@@ -618,4 +619,91 @@ public void allocateVariantArrayElement(int[] indexes, int sourceIndex) {
...
+                    boolean connected = triple.getRight();
+
+                    // if there is no way to find a connectable bus, remove the connectable
+                    // an alternative would be to connect them all to a new trash configured bus
+                    if (connectableBus != null) {
+                        AbstractConnectable<?> connectable = oldTerminalExt.getConnectable();
+
+                        // create the new terminal with new type
+                        TerminalExt newTerminalExt = null;
+                        if (oldTerminalExt.getConnectable().getType() != IdentifiableType.BUSBAR_SECTION) {
```
flo-dup, line 669: In case of a BBS, its terminal is set to null but it's still in the connectable index (this can be checked in the unit test), so maybe you could move this condition in the previous if above `if (connectableBus != null) {`

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VoltageLevelImpl.java
```diff
@@ -618,4 +619,91 @@ public void allocateVariantArrayElement(int[] indexes, int sourceIndex) {
...
+                            newTerminalExt = new TerminalBuilder(networkRef, this, oldTerminalExt.getSide())
+                                    .setBus(connected ? connectableBus.getId() : null)
+                                    .setConnectableBus(connectableBus.getId())
+                                    .build();
+                        }
+
+                        // first attach new terminal to connectable and to new topology model
+                        if (newTerminalExt != null) {
```
flo-dup, line 677: if the BBS is dealt with in the other if, `newTerminalExt` cannot be null

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VoltageLevelImpl.java
```diff
@@ -618,4 +619,91 @@ public void allocateVariantArrayElement(int[] indexes, int sourceIndex) {
...
+                        for (Referrer<Terminal> referrer : oldTerminalExt.getReferrerManager().getReferrers()) {
+                            referrer.onReferencedReplacement(oldTerminalExt, newTerminalExt);
+                        }
+                    } else {
+                        // here keep the removal notification
+                        oldTerminalExt.getConnectable().remove();
+                    }
+                }
+
+                // also here keep the notification for remaining switches and internal connection removal
```
flo-dup, line 699: There are no notifications for internal connection removal
```suggestion
                // also here keep the notification for remaining switches removal
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VoltageLevelImpl.java
```diff
@@ -618,4 +619,91 @@ public void allocateVariantArrayElement(int[] indexes, int sourceIndex) {
...
+                }
+
+                for (var triple : oldTopologyModelInfos) {
+                    TerminalExt oldTerminalExt = triple.getLeft();
+                    Bus connectableBus = triple.getMiddle();
+                    boolean connected = triple.getRight();
+
+                    // if there is no way to find a connectable bus, remove the connectable
+                    // an alternative would be to connect them all to a new trash configured bus
+                    if (connectableBus != null) {
```
flo-dup, line 664: Isn't there's always a non-null connectableBus? A connectable is tied to a node, and a CalculatedBus is always created for a non-empty set of nodes:
https://github.com/powsybl/powsybl-core/blob/218eb5055c456a913c28199f435a0b49e468ce92/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java#L491-L496

flo-dup (changes requested):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractConnectable.java
geofjamg: Done

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AreaBoundaryImpl.java
geofjamg: Done, I switched all attributes to private

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VoltageLevelImpl.java
geofjamg: Done, now the complexity is < 15

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VoltageLevelImpl.java
geofjamg: Done

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VoltageLevelImpl.java
geofjamg: There was a bug indead, busbar sections were still reacheable from the network index. I now start the conversion by removing the busbar sections.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VoltageLevelImpl.java
geofjamg: Yes

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VoltageLevelImpl.java
geofjamg: I'm almost sure I already seen null connectable bus ID in the bus/breaker view of a node/breaker topology when all switches are open.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractConnectable.java
```diff
@@ -175,22 +175,40 @@ protected void move(TerminalExt oldTerminal, TopologyPoint oldTopologyPoint, Str
...
+        if (newTerminalExt != null && notify) {
+            notifyUpdate("terminal" + (iSide + 1), oldTopologyPoint, newTerminalExt.getTopologyPoint());
+        }
+    }
+
+    void replaceTerminal(TerminalExt oldTerminal, TopologyModel newTopologyModel, TerminalExt newTerminalExt, boolean notify) {
         // first, attach new terminal to connectable and to voltage level of destination, to ensure that the new terminal is valid
-        terminalExt.setConnectable(this);
-        voltageLevel.getTopologyModel().attach(terminalExt, false);
+        if (newTerminalExt != null) {
```
flo-dup, line 196: Couldn't we require newTerminalExt to be non-null? we don't need it anymore in this PR, and besides the referrer call for replacement is not working well in the implementations if newTerminalExt is null.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractConnectable.java
```diff
@@ -175,22 +175,40 @@ protected void move(TerminalExt oldTerminal, TopologyPoint oldTopologyPoint, Str
...
-    private void attachTerminal(TerminalExt oldTerminal, TopologyPoint oldTopologyPoint, VoltageLevelExt voltageLevel, TerminalExt terminalExt) {
+    void replaceTerminal(TerminalExt oldTerminal, TopologyPoint oldTopologyPoint, TerminalExt newTerminalExt, boolean notify) {
+        Objects.requireNonNull(oldTerminal);
+        int iSide = terminals.indexOf(oldTerminal);
+        if (iSide == -1) {
+            throw new PowsyblException("Terminal to replace not found");
+        }
+        terminals.set(iSide, newTerminalExt);
+
+        if (newTerminalExt != null && notify) {
```
flo-dup, line 189: Then to be in line with the other replace we could remove this check also, at the cost of a Objects.requireNonNull

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractConnectable.java
geofjamg: Yes cleaned

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractConnectable.java
geofjamg: Done

flo-dup (approved):

PR #3203, opened 2024-11-07
Add network events to IIDM API

geofjamg:

No

Feature

Network notifications unit testing is hard and not done the same way everywhere.

We added in the API a network event data model and an network listener responsible for event recording that will be used at the end in every notification unit tests and also maybe for other use cases.  

- [x] Yes
- [ ] No

NetworkListener has been modified with breaking changes:
 - onUpdate overload without variant ID has been removed, we expect now to one with variant ID to use everywhere and with a null variantId in case of attribute that do not depend on variant.
 - onElementAdded has been renamed onPropertyAdded and default empty implementation has been removed
 - onElementReplaced has been renamed onPropertyReplaced and default empty implementation has been removed 
 - onElementRemoved has been renamed onPropertyRemoved and default empty implementation has been removed 
 - default empty implementation of onVariantCreated, onVariantOverwritten, onVariantRemoved have been removed

Substation geographicalTags attribute is no more an (old) element type notification but a standard attrbute notification
 

**Other information**:


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/NetworkEventRecorder.java
```diff
@@ -0,0 +1,100 @@
...
+        events.add(new ExtensionUpdateNetworkEvent(((Identifiable<?>) extension.getExtendable()).getId(), extension.getName(), attribute, variantId, oldValue, newValue));
+    }
+
+    @Override
+    public void onPropertyAdded(Identifiable<?> identifiable, String attribute, Object newValue) {
+        events.add(new PropertyUpdateNetworkEvent(identifiable.getId(), attribute, PropertyUpdateNetworkEvent.PropertyUpdateType.ADDED, null, newValue));
+    }
+
+    @Override
+    public void onPropertyReplaced(Identifiable<?> identifiable, String attribute, Object oldValue, Object newValue) {
```
flo-dup, line 77: In case of properties we should speak of `key` instead of `attribute`

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/events/PropertyUpdateNetworkEvent.java
```diff
@@ -0,0 +1,33 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+package com.powsybl.iidm.network.events;
+
+import java.util.Objects;
+
+/**
+ * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
+ */
+public record PropertyUpdateNetworkEvent(String id, String attribute, PropertyUpdateType updateType, Object oldValue, Object newValue) implements NetworkEvent {
```
flo-dup, line 15: Same comment regarding `attribute` parameter, rather `key`.

And shouldn't we call this `PropertiesUpdateNetworkEvent`? Indeed, the update is not on the propert**y** (a bit strange to talk about update with a new key or a removed key), but on the propert**ies** with a specific key.

Then we should speak  of `PropertiesUpdateType`, and maybe `ADDED_PROPERTY`, `REMOVED_PROPERTY`, `UPDATED_PROPERTY`?

flo-dup (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/NetworkEventRecorder.java
geofjamg: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/events/PropertyUpdateNetworkEvent.java
geofjamg: Done

flo-dup (approved):

PR #3208, opened 2024-11-13
Adder by copy for lines 

nao1345678:

It allows the lines to be initialized from another already existing one.

Lines are not allowed to be duplicated.

- [X] Yes
- [ ] No

Custom IIDM implementation maintainers should add the following method in their `Network` implementations:
- `LineAdder newLine(Line line)`.

**Other information**:


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/LineAdderImpl.java
```diff
@@ -110,6 +122,26 @@ public LineImpl add() {
         line.addTerminal(terminal1);
         line.addTerminal(terminal2);
 
+        if (copiedLine != null) {
+            Collection<Extension<Line>> extensions = copiedLine.getExtensions();
+            extensions.forEach(lineExtension ->
+                   line.addExtension((Class<? super Extension<Line>>) lineExtension.getClass(), lineExtension));
```
olperr1, line 128: Beware: here it does not copy the extension. It modifies it to point to `line` and adds it to `line. Thus, after calling `add()` on the adder, `line` and `copiedLine` will have the same Java object as extension.

olperr1 (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Branch.java
```diff
@@ -97,7 +97,7 @@
  *         </tr>
  *     </tbody>
  * </table>
- *
+
```
HugoKulesza, line 100: To clean

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/LoadingLimitsUtil.java
```diff
@@ -96,6 +96,9 @@ public static <L extends LoadingLimits, A extends LoadingLimitsAdder<L, A>> void
       * @param limits the limits to copy
       */
     public static <L extends LoadingLimits, A extends LoadingLimitsAdder<L, A>> A initializeFromLoadingLimits(A adder, L limits) {
+        if (limits == null) {
+            throw new PowsyblException("Cannot initialize new limits from null limits");
+        }
```
HugoKulesza, line 101: I think in the end that it is better to just warn the user (using a logger) and return an empty adder than to stop everything with a throw

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/SubnetworkImpl.java
```diff
@@ -232,6 +232,11 @@ public LineAdder newLine() {
         return getNetwork().newLine(id);
     }
 
+    @Override
+    public LineAdder newLine(Line line) {
+        return getNetwork().newLine(line.getId());
```
HugoKulesza, line 237: I think that it is more :
```suggestion
        return getNetwork().newLine(id, line);
```
(using the method newLine(String subnetwork, Line line) )
What you wrote here is creating a new (empty) line adder in the subnetwork that is named line.getId() which is not what we want

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractCurrentLimitsTest.java
```diff
@@ -738,6 +739,7 @@ public void testAdderByCopy() {
         adder.add();
 
         assertTrue(areLimitsIdentical(limits1, limits2));
+        assertThrows(PowsyblException.class, () -> line.newCurrentLimits1(null));
```
HugoKulesza, line 742: To change accordingly with the change in behaviour with null limits

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractIdenticalLinesTest.java
```diff
@@ -0,0 +1,25 @@
...
+                    && line1.getG1() == line2.getG1()
+                    && line1.getG2() == line2.getG2()
+                    && line1.getB1() == line2.getB1()
+                    && line1.getB2() == line2.getB2()
+                    && Objects.equals(line1.getTerminal1().getVoltageLevel().getId(), line2.getTerminal1().getVoltageLevel().getId())
+                    && Objects.equals(line1.getTerminal2().getVoltageLevel().getId(), line2.getTerminal2().getVoltageLevel().getId());
+        }
+        return areIdentical;
+    }
+}
```
HugoKulesza, line 23: I don't think it is necessary to add a new class for this method only, you can add it directly to AbstractLineTest

HugoKulesza (changes requested):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
```diff
@@ -356,13 +356,29 @@ public VoltageLevelExt getVoltageLevel(String id) {
...
+    @Override
+    public LineAdderImpl newLine(Line line) {
+        return newLine(null, line).setR(line.getR())
+                .setX(line.getX())
+                .setG1(line.getG1())
+                .setG2(line.getG2())
+                .setB1(line.getB1())
+                .setB2(line.getB2())
+                .setVoltageLevel1(line.getTerminal1().getVoltageLevel().getId())
+                .setVoltageLevel2(line.getTerminal2().getVoltageLevel().getId());
```
olperr1, line 375: The setters should be called in the adder's constructor (or in a utility method, but called in the constructor).
In the current state, these variables are not set when `newLine(Line)` is called from a subnetwork.

olperr1 (changes requested):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
olperr1: `BranchUtil` is in the `iidm-impl` module here. I think it would be better to have this method in the `iidm-api` module, so it could be reused by custom IIDM implementations (such as powsybl-network-store).

I think you can define it as a static method directly in `LineAdder`.

Another think. It would be better to call it in `LineAdderImpl(NetworkImpl network, String subnetwork, Line copiedLine)`. It would be counterintuitive that this method initializes the limits, but not the line attributes.

HugoKulesza (approved):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/LineAdderImpl.java
```diff
@@ -36,6 +35,15 @@ class LineAdderImpl extends AbstractBranchAdder<LineAdderImpl> implements LineAd
...
         this.subnetwork = subnetwork;
+        this.copiedLine = null;
+    }
+
+    LineAdderImpl(NetworkImpl network, String subnetwork, Line copiedLine) {
+        this.network = network;
+        this.subnetwork = subnetwork;
+        this.copiedLine = copiedLine;
+        LineAdderImpl lineAdderImpl = this;
+        LineAdder.fillLineAdder(lineAdderImpl, copiedLine);
```
olperr1, line 46: This could be simplified:

```suggestion
        LineAdder.fillLineAdder(this, copiedLine);
```

olperr1 (changes requested):

olperr1 (approved):

Issue #3157, opened 2024-09-25
Add Area interchange target action


jeandemanged:
### Describe the current behavior

We cannot model an action consisting in modifying the interchangeTarget of an Area

### Describe the expected behavior

We can model an action consisting in modifying the interchangeTarget of an Area

### Describe the motivation

For load flows with area interchange control enabled, we can model operator strategies consisting in changes in area interchanges.

PR #3209, opened 2024-11-13
Add area interchange target update action

obrix:

#3157

Introduce an action for area interchange target update. To be used in combination with generator and load action in a security analysis to implement a full countertrading process that impact the area interchange targets.

- [ ] Yes
- [X] No

**Other information**:

### DIFF
--- a/action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetAction.java
+++ b/action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetAction.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, Coreso SA (https://www.coreso.eu/) and TSCNET Services GmbH (https://www.tscnet.eu/)
+ * Copyright (c) 2024, Coreso SA (https://www.coreso.eu/) and TSCNET Services GmbH (https://www.tscnet.eu/)
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
--- a/action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetAction.java
+++ b/action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetAction.java
@@ -19,14 +19,14 @@ import java.util.Objects;
  *  </ul>
  * @author Bertrand Rix {@literal <bertrand.rix at artelys.com>}
  */
-public class AreaInterchangeTargetUpdateAction extends AbstractAction {
+public class AreaInterchangeTargetAction extends AbstractAction {
 
-    public static final String NAME = "AREA_INTERCHANGE_TARGET_UPDATE_ACTION";
+    public static final String NAME = "AREA_INTERCHANGE_TARGET_ACTION";
 
     private final String areaId;
     private final double interchangeTarget;
 
-    public AreaInterchangeTargetUpdateAction(String id, String areaId, double interchangeTarget) {
+    public AreaInterchangeTargetAction(String id, String areaId, double interchangeTarget) {
         super(id);
         this.areaId = Objects.requireNonNull(areaId);
         this.interchangeTarget = interchangeTarget;
@@ -56,7 +56,7 @@ public class AreaInterchangeTargetUpdateAction extends AbstractAction {
         if (!super.equals(o)) {
             return false;
         }
-        AreaInterchangeTargetUpdateAction that = (AreaInterchangeTargetUpdateAction) o;
+        AreaInterchangeTargetAction that = (AreaInterchangeTargetAction) o;
         return Objects.equals(areaId, that.areaId) && (interchangeTarget == that.interchangeTarget || (Double.isNaN(interchangeTarget) && Double.isNaN(that.interchangeTarget)));
     }
 
--- a/action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetAction.java
+++ b/action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetAction.java
@@ -15,7 +15,7 @@ import java.util.Objects;
 /**
  *  An action to:
  *  <ul>
- *      <li>Change the interchange target of an Area by specifying a new interchange target in MW.</li>
+ *      <li>Change the interchange target of an area by specifying a new interchange target in MW.</li>
  *  </ul>
  * @author Bertrand Rix {@literal <bertrand.rix at artelys.com>}
  */
--- a/action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetAction.java
+++ b/action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetAction.java
@@ -57,7 +57,7 @@ public class AreaInterchangeTargetAction extends AbstractAction {
             return false;
         }
         AreaInterchangeTargetAction that = (AreaInterchangeTargetAction) o;
-        return Objects.equals(areaId, that.areaId) && (interchangeTarget == that.interchangeTarget || (Double.isNaN(interchangeTarget) && Double.isNaN(that.interchangeTarget)));
+        return Objects.equals(areaId, that.areaId) && (interchangeTarget == that.interchangeTarget || Double.isNaN(interchangeTarget) && Double.isNaN(that.interchangeTarget));
     }
 
     @Override
--- a/action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetAction.java
+++ b/action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetAction.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024, RTE (http://www.rte-france.com)
+ * Copyright (c) 2023, Coreso SA (https://www.coreso.eu/) and TSCNET Services GmbH (https://www.tscnet.eu/)
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
--- a/action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetActionBuilder.java
+++ b/action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetActionBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, Coreso SA (https://www.coreso.eu/) and TSCNET Services GmbH (https://www.tscnet.eu/)
+ * Copyright (c) 2024, Coreso SA (https://www.coreso.eu/) and TSCNET Services GmbH (https://www.tscnet.eu/)
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
--- a/action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetActionBuilder.java
+++ b/action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetActionBuilder.java
@@ -16,7 +16,7 @@ public class AreaInterchangeTargetActionBuilder implements ActionBuilder<AreaInt
 
     private String areaId;
 
-    private Double target = Double.NaN;
+    private double target = Double.NaN;
 
     @Override
     public String getType() {
--- a/action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetActionBuilder.java
+++ b/action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetActionBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024, RTE (http://www.rte-france.com)
+ * Copyright (c) 2023, Coreso SA (https://www.coreso.eu/) and TSCNET Services GmbH (https://www.tscnet.eu/)
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
--- a/action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetActionBuilder.java
+++ b/action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetActionBuilder.java
@@ -10,7 +10,7 @@ package com.powsybl.action;
 /**
  * @author Bertrand Rix {@literal <bertrand.rix at artelys.com>}
  */
-public class AreaInterchangeTargetUpdateActionBuilder implements ActionBuilder<AreaInterchangeTargetUpdateActionBuilder> {
+public class AreaInterchangeTargetActionBuilder implements ActionBuilder<AreaInterchangeTargetActionBuilder> {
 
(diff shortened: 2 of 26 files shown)


action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetUpdateActionBuilder.java
```diff
@@ -0,0 +1,49 @@
...
+    public String getId() {
+        return id;
+    }
+
+    @Override
+    public AreaInterchangeTargetUpdateActionBuilder withNetworkElementId(String elementId) {
+        return this;
+    }
+
+    public AreaInterchangeTargetUpdateActionBuilder withTarget(Double target) {
```
jeandemanged, line 32: primitive double here

```suggestion
    public AreaInterchangeTargetUpdateActionBuilder withTarget(double target) {
```

olperr1 (changes requested):

action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetUpdateAction.java
```diff
@@ -0,0 +1,50 @@
...
+public class AreaInterchangeTargetUpdateAction extends AbstractAction {
+
+    public static final String NAME = "AREA_INTERCHANGE_TARGET_UPDATE_ACTION";
+
+    private final String areaId;
+    private final double target;
+
+    public AreaInterchangeTargetUpdateAction(String id, String areaId, double target) {
+        super(id);
+        this.areaId = areaId;
```
jeandemanged, line 14: requireNonNull
```suggestion
        this.areaId = Objects.requireNonNull(areaId);
```

olperr1:
Shouldn't the corresponding `NetworkModification` be created?

action-api/src/test/java/com/powsybl/action/json/JsonActionTest.java
```diff
@@ -83,6 +83,7 @@ void actionRoundTrip() throws IOException {
                 .withStaticVarCompensatorId("svc").withRegulationMode(StaticVarCompensator.RegulationMode.REACTIVE_POWER)
                 .withReactivePowerSetpoint(120.0).build());
         actions.add(new TerminalsConnectionAction("id4", "transformerId25", ThreeSides.THREE, true)); // only one side.
+        actions.add(new AreaInterchangeTargetUpdateAction("id99", "AreaA", 101.0));
```
jeandemanged, line 86: please add also a test with a target set to NaN (= remove the target), this should be supported.

action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetUpdateAction.java
```diff
@@ -0,0 +1,50 @@
+package com.powsybl.action;
+
+import java.util.Objects;
+
+public class AreaInterchangeTargetUpdateAction extends AbstractAction {
+
+    public static final String NAME = "AREA_INTERCHANGE_TARGET_UPDATE_ACTION";
+
+    private final String areaId;
+    private final double target;
```
jeandemanged, line 10: rename to `interchangeTarget` for consistency with iIDM

jeandemanged (changes requested):

action-api/src/test/java/com/powsybl/action/json/JsonActionTest.java
obrix: NaN is already correctly interpreted on AIC side ?

obrix:
> Shouldn't the corresponding `NetworkModification` be created?

Yes it is coming !

action-api/src/test/java/com/powsybl/action/json/JsonActionTest.java
jeandemanged: yes in OLF

action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetUpdateAction.java
jeandemanged, line 1: Rename to just `AreaAction` for consistency with other actions (today only interchangeTarget in the action indeed, but we can't predict the future)

jeandemanged (changes requested):

action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetUpdateAction.java
obrix: Are you sure ? I would prefer having a small unit action that only do simple things rather that a big one the does everything like GeneratorAction wich is a pain to handle everywhere

jeandemanged: I'd favor consistency over choosing between introducing an inconsistency or rewriting all the other actions.

obrix: Well we already have separated small actions for ratio / phase tap changer and a lot of actions are already small in scope even if they target a class of equipment. So I think calling that an inconsistency goes a bit far. 
I think keeping it as simple as we can will make futur implementation easier with no need to handle null / optional parameters everywhere.

jeandemanged: Tap changers from my point of view are separate equipment from transformers. There are transformers equipped with and transformers not equipped with tap changers. Making separate action classes for tap changers makes sense in my opinion. Also in CGMES there are also separate classes for the transformers and the tap changers.
I see that there are separate classes for "tap regulation" part and "tap position" part. Here however I don't know why it was made this way, to be honest I don't like it. For a phase tap changer, for example:
- I want to set it on active power control -> OK I have to use one `PhaseTapChangerRegulationAction` for mode and value.
- I want to set it on fixed tap position 12 -> I have to use both `PhaseTapChangerRegulationAction` (for the mode, but not the value) and `PhaseTapChangerTapPositionAction` (for the tap position).

So I think it would have been better to have a single `PhaseTapChangerAction` here.

obrix: @olperr1 Do you have an opinion to help use sort it out ? We have to globally choose between a global action per network element vs a small action per "unit change" on an element.
I stand on the fact it is easier to handle small actions in implementation, especially in OLF SA, from an user point of view I think it is less clear.

Impact on network modification is probably null because we can have a relation of 1 action = a list of modification.

action-api/src/test/resources/ActionFileTest.json
```diff
@@ -189,5 +189,15 @@
...
+  }, {
+    "type" : "AREA_INTERCHANGE_TARGET_UPDATE_ACTION",
+    "id" : "id99",
+    "areaId" : "AreaA",
+    "target" : 101.0
+  }, {
+    "type" : "AREA_INTERCHANGE_TARGET_UPDATE_ACTION",
+    "id" : "idDisabledTarget",
+    "areaId" : "AreaA",
+    "target" : NaN
```
jeandemanged, line 201: I think it would be better not to write the field. `NaN` is not standard in JSON / `NaN` support is JSON parser dependent.

obrix: Agreed I setup NaN as the default value in the builder and simply do not serialize it when the value is NaN.

action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetUpdateAction.java
jeandemanged: > from an user point of view I think it is less clear

That's the whole point: Today we have:
- `Load` has `LoadAction`
- `Generator` has `GeneratorAction`
- `DanglingLine` has `DanglingLineAction`
- `Hvdc` has `HvdcAction`
- `StaticVarCompensator` has `StaticVarCompensatorAction`
- `Switch` has `SwitchAction`

Despite the few outliers:
- `RatioTapChanger` has `RatioTapChanger<Regulation|TapPosition>Action`
- `PhaseTapChanger` has `PhaseTapChanger<Regulation|TapPosition>Action`
- `Terminal` has `TerminalConnectionAction`

I think from user point of view it is much more natural to expect an `AreaAction` for an `Area`.

geofjamg: consistency vs design flaw fix...
I think that large action scope (one per element) is not a good choice and does not allow clean and maintanable code especially on integration side (OLF for instance). We should have split for example the generation action into smaller peace maybe grouped by theme (active power change, voltage control change, etc).

action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetUpdateAction.java
```diff
@@ -0,0 +1,50 @@
+package com.powsybl.action;
```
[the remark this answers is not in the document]
jeandemanged: @obrix unresolving here, all copyrights to be changed to
```
Copyright (c) 2024, Coreso SA (https://www.coreso.eu/) and TSCNET Services GmbH (https://www.tscnet.eu/)
```

action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetUpdateAction.java
jeandemanged: OK that is fine for me

action-api/src/test/java/com/powsybl/action/EqualsActionTest.java
```diff
@@ -546,4 +546,34 @@ void multipleActionsAction() {
                 .testEquals();
     }
 
+    @Test
+    void interchangeTargetAction() {
```
jeandemanged, line 550: NaN case to be checked for == and !=

iidm/iidm-modification/src/test/java/com/powsybl/iidm/modification/AreaInterchangeModificationTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @BeforeEach
+    public void setUp() {
+        network = EurostagTutorialExample1Factory.createWithTieLinesAndAreas();
+        assertTrue(network.getAreaCount() > 0);
+        area = network.getArea("ControlArea_A");
+    }
+
+    @Test
+    void testBatteryModification() {
```
jeandemanged, line 37:
```suggestion
    void testModification() {
```

iidm/iidm-modification/src/test/java/com/powsybl/iidm/modification/AreaInterchangeModificationTest.java
```diff
@@ -0,0 +1,75 @@
...
+        modification2.apply(network);
+        assertTrue(area.getInterchangeTarget().isEmpty());
+
+        AreaInterchangeTargetModification modification3 = new AreaInterchangeTargetModification("AREA_UNKNOWN", 2.0);
+        PowsyblException exception = assertThrows(PowsyblException.class, () -> modification3.apply(network, true, ReportNode.NO_OP));
+        assertEquals("Area 'AREA_UNKNOWN' not found", exception.getMessage());
+    }
+
+    @Test
+    void testBatteryModificationGetters() {
```
jeandemanged, line 54:
```suggestion
    void testModificationGetters() {
```

jeandemanged (changes requested):

action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetUpdateAction.java
obrix: Yes probably thanks

action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetAction.java
```diff
@@ -0,0 +1,72 @@
+/*
+ * Copyright (c) 2023, Coreso SA (https://www.coreso.eu/) and TSCNET Services GmbH (https://www.tscnet.eu/)
```
jeandemanged, line 2: 2024

jeandemanged (changes requested):
Copyright year to be changed to 2024 in new files

action-api/src/main/java/com/powsybl/action/AreaInterchangeTargetActionBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * @author Bertrand Rix {@literal <bertrand.rix at artelys.com>}
+ */
+public class AreaInterchangeTargetActionBuilder implements ActionBuilder<AreaInterchangeTargetActionBuilder> {
+
+    private String id;
+
+    private String areaId;
+
+    private Double target = Double.NaN;
```
olperr1, line 19:
```suggestion
    private double target = Double.NaN;
```

olperr1 (changes requested):

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/AreaInterchangeTargetModification.java
```diff
@@ -0,0 +1,69 @@
...
+    @Override
+    public NetworkModificationImpact hasImpactOnNetwork(Network network) {
+        impact = DEFAULT_IMPACT;
+        Area area = network.getArea(areaId);
+        if (area == null) {
+            impact = NetworkModificationImpact.CANNOT_BE_APPLIED;
+        } else if (area.getInterchangeTarget().isPresent() && Math.abs(interchangeTarget - area.getInterchangeTarget().getAsDouble())
+            < EPSILON || area.getInterchangeTarget().isEmpty() && Double.isNaN(interchangeTarget)) {
+            impact = NetworkModificationImpact.NO_IMPACT_ON_NETWORK;
+        }
```
olperr1, line 58: To improve readability, I suggest this formatting:
```suggestion
        } else if (area.getInterchangeTarget().isPresent()
                    && Math.abs(interchangeTarget - area.getInterchangeTarget().getAsDouble()) < EPSILON
                || area.getInterchangeTarget().isEmpty() && Double.isNaN(interchangeTarget)) {
            impact = NetworkModificationImpact.NO_IMPACT_ON_NETWORK;
        }
```

olperr1 (changes requested):

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/AreaInterchangeTargetModification.java
obrix: Well, hard to see the difference because of github formatting but I'll try something...

olperr1: You can copy-paste the code if you want to see the suggestion.
But basically the suggested changes were to insert:
- new lines before the first `&&` and the first `||` in order to have the whole comparison to `EPSILON` on the same line;
- 4 additional spaces before the first `&&`, so it has not the same indent as the next line.

olperr1: But I saw your modification, it looks fine to me. Thanks!

olperr1 (approved):

olperr1 (changes requested):
Sorry, I missed another thing.
Can you add the action description in the list of the supported actions in the documentation?
([in this section](https://github.com/powsybl/powsybl-core/blob/a088d03adbebe79b960d08be6905dff6b1bc8157/docs/simulation/security/index.md?plain=1#L37))

olperr1 (approved):

PR #3210, opened 2024-11-13
[CGMES] Less internal connections created at import

flo-dup:

No

Feature

Each connectable and switch is created with two nodes at each end, one for the cim:ConnectivityNode, one for the cim:Terminal. The two nodes are connected with an InternalConnection. This leads to a verbose iidm file with an excessive number of InternalConnections.

When the configuration is the following:
- `iidm.import.cgmes.create-fictitious-switches-for-disconnected-terminals-mode` is `NEVER` (default value is `ALWAYS`)
- and `iidm.import.cgmes.create-busbar-section-for-every-connectivity-node` is `false` (which is the default value)

The InternalConnection is created only if needed:
- if there's only one Connectable at the ConnectivityNode, no InternalConnection
- if there are two Connectables at the ConnectivityNode, only one InternalConnection is created, for the first Connectable encountered (instead of one for each)
- if there are more than two Connectables, one InternalConnection is created between each terminal and the ConnectivityNode, except for the BusbarSection case:
   - if there's exactly one BusbarSection among the connectables, the InternalConnection are created between this BusbarSection and the other connectables
   - if there's more than one BusbarSection (does not make so much sense, but exists in real data), the InternalConnections are created between the first BusbarSection and all the other Connectables, including the other BusbarSections.

- [ ] Yes
- [x] No


zamarrenolm:
Just one quick consideration: could all these rules/simplification of the underlying connectivity model be moved to a post-processor after import (or a network modification). This way it could be useful for all importers, in a centralised way, and we should not need to consider them in every importer that have to handle. For example, recent versions of PSSE (35 and later) support node/breaker models where we could face similar problems.

flo-dup:
> Just one quick consideration: could all these rules/simplification of the underlying connectivity model be moved to a post-processor after import (or a network modification). This way it could be useful for all importers, in a centralised way, and we should not need to consider them in every importer that have to handle. For example, recent versions of PSSE (35 and later) support node/breaker models where we could face similar problems.

Thanks for pointing out that similarity with PSSE! We should rather centralize it in a utility method, post-processing is not the good way to go, in terms of performance, robustness, and readability. That's at the cost of handling the utility method call (and map building beforehand) in each importer, and having an abstraction which allows it. If that's not possible, I think we should rather duplicate the logic than putting it in a common postprocessing.

flo-dup:
> Just one quick consideration: could all these rules/simplification of the underlying connectivity model be moved to a post-processor after import (or a network modification). This way it could be useful for all importers, in a centralised way, and we should not need to consider them in every importer that have to handle. For example, recent versions of PSSE (35 and later) support node/breaker models where we could face similar problems.

Thanks for pointing out that similarity with PSSE! But we should rather centralize it in a utility method. Post-processing is not the good way to go I think (in terms of performance, robustness, and readability). That's at the cost of handling the utility method call (and map building beforehand) in each importer, and having an abstraction which allows it, sure. If that's not possible, I think we should rather duplicate the logic than putting it in a common postprocessing.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/NodeMapping.java
```diff
@@ -47,47 +49,123 @@ public int iidmNodeForTerminal(CgmesTerminal t, boolean isSwitchEnd, VoltageLeve
...
+            sw.setProperty(Conversion.PROPERTY_IS_CREATED_FOR_DISCONNECTED_TERMINAL, "true");
         }
+    }
+
+    private boolean isFirstBbsAtConnectivityNode(CgmesTerminal t, List<CgmesTerminal> connectivityNodeTerminals) {
+        return connectivityNodeTerminals.stream()
+                .filter(t1 -> t != t1)
+                .filter(NodeMapping::isBusbarSectionTerminal)
+                .noneMatch(t1 -> cgmes2iidm.containsKey(t1.id()));
+    }
```
olperr1, line 165: This method is not used anymore.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/NodeMapping.java
```diff
@@ -47,47 +49,123 @@ public int iidmNodeForTerminal(CgmesTerminal t, boolean isSwitchEnd, VoltageLeve
...
+                CgmesTerminal t1 = connectivityNodeTerminals.get(1);
+                if (!isBusbarSectionTerminal(t0) && isBusbarSectionTerminal(t1)) {
+                    oneIidmNodeForBothTerminalAndConnectivityNode(t1, vl);
+                    createInternalConnection(t0, vl);
+                } else {
+                    oneIidmNodeForBothTerminalAndConnectivityNode(t0, vl);
+                    createInternalConnection(t1, vl);
+                }
+            }
+            default -> {
```
olperr1, line 104: Can you add a unit test with more than 2 terminals at one ConnectivityNode?

olperr1 (changes requested):

flo-dup:
Tested on ENTSO-E CGMES sample data SmallGrid/NodeBreaker
|| Without this PR    | With this PR |
|:----------------:|:----------------:|:------------------------:|
|       iidm export file size       |       630ko       |          450ko           |
|       internal connections       |       2834       |          136           |
|       read time on my laptop (averaged over 100 iterations)       |       0.81094s       |          0.80416s           |

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/NodeMapping.java
```diff
@@ -47,47 +49,123 @@ public int iidmNodeForTerminal(CgmesTerminal t, boolean isSwitchEnd, VoltageLeve
...
+                for (CgmesTerminal t : connectivityNodeTerminals) {
+                    if (isBusbarSectionTerminal(t) && !bbsEncountered) {
+                        oneIidmNodeForBothTerminalAndConnectivityNode(t, vl);
+                        bbsEncountered = true;
+                    } else {
+                        // Add internal connection from terminal to connectivity node as iidm can only handle one terminal per node
+                        // For node-breaker models we create an internal connection between the terminal and its connectivity node
+                        createInternalConnection(t, vl);
+                    }
+                }
```
rcourtier, line 119: In case there are more than 2 non-busbarsection terminals, you could place the first terminal at the connectivity node. The same way you say if it is a busbar, I place it at the connectivity node, you could do that for the first element of the list if none is a busbarsection. That would create one less node and internal connection.
I would also understand you prefer to keep it that way, as your current implementation keeps the essence of multiple elements (all having their own node) connected through a common node coupling them all.

rcourtier (changes requested):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/NodeMapping.java
flo-dup: Unit test added covering all cases.

olperr1 (approved):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/NodeMapping.java
```diff
@@ -47,47 +51,121 @@ public int iidmNodeForTerminal(CgmesTerminal t, boolean isSwitchEnd, VoltageLeve
...
+            } else if (!cgmes2iidm.containsKey(t.id())) {
+                // Create internal connections but only if too many terminals on connectivity node
+                createInternalConnectionsIfNeeded(t.connectivityNode(), vl);
+            }
+        } else {
+            boolean connected = t.connected() && equipmentIsConnected;
+            if (!connected && createFictitiousSwitch(fictSwCreationMode, isSwitchEnd)) {
+                createFictitiousSwitch(t, vl);
+            } else {
+                createInternalConnection(t, vl);
```
rcourtier, line 74: You don't have to systematically create a node for the terminal and an internal connection between the terminal's node and the connectivitynode's node. You could do it only if needed, that is if 2 or more connected non-switch equipments share that terminal's connectivity node.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/CgmesImportMultiTerminalConnectivityNodesTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+    @Test
+    void testNever() {
+        Properties cgmesImportParameters = new Properties();
+        cgmesImportParameters.put(CgmesImport.CREATE_FICTITIOUS_SWITCHES_FOR_DISCONNECTED_TERMINALS_MODE, "NEVER");
+        Network network = Network.read("multiTerminalConnectivityNodes.xml",
+                getClass().getResourceAsStream("/multiTerminalConnectivityNodes.xml"),
+                LocalComputationManager.getDefault(),
+                ImportConfig.load(),
+                cgmesImportParameters);
```
rcourtier, line 38: To simplify the import, you may use a method from ConversionUtil class in com.powsybl.cgmes.conversion.test package:
```java
public static Network readCgmesResources(Properties properties, String dir, String... files)
```
That may be especially handy when you need to read several files (like EQ and SSH instance).

rcourtier (changes requested):

zamarrenolm:
Unit test fails: `EquipmentExportTest.miniGridCgmesExportPreservingOriginalClassesOfGenerators`.

The test converts a CGMES network to IIDM, writes it back to CGMES and re-imports to verify that specific classes of generators have been preserved.

The number of nodes in IIDM in the first import and the second one are different. This must be understood; it is not clear why it happens.

If the behaviour is ok and both networks are equivalent, the attribute that makes the comparison fail (`network/substation/voltageLevel/generator/@node`) could be ignored using the `DifferenceEvaluator`, in the same way other non relevant differences are skipped.

olperr1:
Impacted by the "1 EQ / 24 SSH" PRs.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/AbstractConductingEquipmentConversion.java
```diff
@@ -198,6 +198,13 @@ public DanglingLine convertToDanglingLine(String eqInstance, int boundarySide, d
             throw new PowsyblException(String.format("Unexpected boundarySide and modelSide at boundaryNode: %s", boundaryNode));
         }
 
+        // The IIDM dangling line comes from a CGMES switch. Its terminals haven't been added to the mapping.
```
marqueslanauja: Maybe clarify the comment, since not all dangling lines come from a switch

`When the IIDM dangling line comes from a CGMES switch, its terminals haven't been added to the mapping.`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/TerminalConversion.java
```diff
@@ -130,9 +131,17 @@ private static void moveSwitchNode(Switch sw, int node, int newNode) {
 
     private static void createSwitchForTerminal(Terminal terminal, String cgmesTerminalId, Context context) {
         int node = terminal.getNodeBreakerView().getNode();
-        int newNode = terminal.getVoltageLevel().getNodeBreakerView().getMaximumNodeIndex() + 1;
-        createSwitch(terminal.getVoltageLevel(), cgmesTerminalId, node, newNode, context);
-        terminal.getNodeBreakerView().moveConnectable(newNode, terminal.getVoltageLevel().getId());
+        List<Integer> nodesInternalConnectedTo = terminal.getVoltageLevel().getNodeBreakerView().getNodesInternalConnectedTo(node);
+        if (nodesInternalConnectedTo.size() == 1) {
```
marqueslanauja: This is only valid if there are no switches connected to the node (i.e., only one internal connection). Theoretically, this is possible, but depending on your implementation, it might no longer hold. Still, it's better to check to avoid future malfunctions after potential changes to your implementation


image

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/AbstractConductingEquipmentConversion.java
rcourtier: Change applied!

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/TerminalConversion.java
rcourtier: You are right. Thanks for catching that!
Change applied!

rcourtier:
> I've tested with real networks, and it works fine. In some cases, there is an unnecessary internal connection that always appears under the same configuration: a node without connectables acting as the starting point for all internal connections to the connectables. In all these cases, it's possible to remove one internal connection and move its connectable to the initial point.
> 
> Might this configuration be intentional to simplify graphical visualization?

Well actually no. The need to simplify graphical visualization and graph traversals mainly appears when one of the connectable of the 3+ connection to the node is a bus bar section. When that is the case, the bus bar section and its node become the central elements, and other equipments get connected to it through an internal connection.

We didn't want to have other kinds of connectable be as central connection-wise as the bus bar sections, so we created on purpose this central node with no connectable directly attached to it, but with internal connections to the other equipments.

In addition of having no other central equipments than bus bar sections, one advantage is that in case of the network being manually modified, it's possible to remove any one of the connectable sharing the central node (and its internal connection to it) without changing anything else.

jeandemanged:
Hello @flo-dup , @marqueslanauja , @rcourtier ,

Just wondering, won't this change prevent us in the future to have persistent / stable connectivity nodes mRIDs that are preserved on an import/export round-trip?

rcourtier:
> Hello @flo-dup , @marqueslanauja , @rcourtier ,
> 
> Just wondering, won't this change prevent us in the future to have persistent / stable connectivity nodes mRIDs that are preserved on an import/export round-trip?

Hi @jeandemanged 
It won't prevent us to have stable ConnectivityNode rdf:Id. The CGMES ConnectivityNode are still always mapped to IIDM nodes. The main change performed here is to not systematically create an IIDM node corresponding to the CGMES Terminal (and that goes with that, not systematically create the internal connection), which reduces the total number of IIDM nodes and internal connections.

jeandemanged:
> > Hello @flo-dup , @marqueslanauja , @rcourtier ,
> > Just wondering, won't this change prevent us in the future to have persistent / stable connectivity nodes mRIDs that are preserved on an import/export round-trip?
> 
> Hi @jeandemanged It won't prevent us to have stable ConnectivityNode rdf:Id. The CGMES ConnectivityNode are still always mapped to IIDM nodes. The main change performed here is to not systematically create an IIDM node corresponding to the CGMES Terminal (and that goes with that, not systematically create the internal connection), which reduces the total number of IIDM nodes and internal connections.

Ha indeed, you are correct, very good then. Thank you.

PR #3218, opened 2024-11-20
Fix non linear shunt conversion performances issue

rcourtier:

No

Fix performances issue.

Query of cim:NonlinearShuntCompensatorPoint is slow as it is using a FILTER clause which isn't good performance wise.

Query and java code have been rewritten to convert nonLinearShuntCompensatorPoint without using the FILTER clause, thus improving the performances.

- [X] Yes
- [ ] No

Any call to ```CgmesModel.nonlinearShuntCompensatorPoints(String id)``` should be replaced by ```CgmesModel.groupedShuntCompensatorPoints().getOrDefault(id, new PropertyBags())```

**Other information**:


cgmes/cgmes-model/src/main/resources/CIM16.sparql
```diff
@@ -826,8 +826,7 @@ WHERE {
         cim:NonlinearShuntCompensatorPoint.NonlinearShuntCompensator ?Shunt ;
         cim:NonlinearShuntCompensatorPoint.sectionNumber ?sectionNumber ;
         cim:NonlinearShuntCompensatorPoint.b ?b ;
-        cim:NonlinearShuntCompensatorPoint.g ?g  .
-    FILTER REGEX ( STR (?Shunt), "{0}")
+        cim:NonlinearShuntCompensatorPoint.g ?g ;
```
zamarrenolm, line 829: minor syntax thing: predicate-object lists should end in a dot (or nothing, if it is the last triple pattern in the group)

rcourtier: fixed

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesModel.java
```diff
@@ -109,7 +109,19 @@ default PropertyBags fullModels() {
 
     PropertyBags equivalentShunts();
 
-    PropertyBags nonlinearShuntCompensatorPoints(String id);
+    PropertyBags shuntCompensatorPoints();
+
+    // Non-linear shunt compensator points grouped by shunt compensator
+    Map<String, PropertyBags> groupedShuntCompensatorPoints();
```
olperr1, line 115: I don't think this method should be exposed in the API instead of the old `nonlinearShuntCompensatorPoints(String id)` method.
When we look at the usage, the new method is always called to retrieve the data for a single id and it is not trivial to refactor the code to call the method only once for all the IDs.
So I think this method is more an implementation detail of `CgmesModelTripleStore`.

Note that using the new method also introduces breaking changes that are not really needed.

olperr1 (changes requested):

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesModel.java
zamarrenolm: Providing this method makes the API of the CGMES model consistent with other cached data, like grouped transformer ends.

rcourtier: I'll refactor the other methods to have a consistent behaviour:
Provide in the API (CgmesModel) the methods:
- xxx() that will query all data
- xxx(String linkedObjectId) that will query the data associated to the linkedObjectId
In the implementation (AbstractCgmesModel) the same methods where:
- xxx() runs the query
- xxx(String linkedObjectId) that will:
  - If run for the 1st time run the query, build a Map and cache it
  - Returns the data associated to the linkedObjectId from the cached Map

olperr1: Seen with @rcourtier:

- The `nonlinearShuntCompensatorPoints(String id)` was restored and `groupedShuntCompensatorPoints()` is no more exposed in the `CgmesModel` interface.
- A general refactoring of `CgmesModel`, providing more consistency, will be done in another PR.

olperr1 (approved):

PR #3225, opened 2024-11-25
Store the CGMES id/name of OperationalLimitSet in a unique property

rcourtier:

No.

Feature change.

At CGMES import, OperationalLimitSet names and ids are stored in a property of the corresponding identifiable as follows:
- Property name: "CGMES.OperationalLimitSet_\<rdfid\>"
- Property value: "\<name\>"

Since rdfid are globally uniques, this results in the model creating overall as many property keys as OperationalLimitSet. This is an issue in pypowsybl where a huge number of property keys severly alters performances.

At CGMES import, OperationalLimitSet names and ids are stored in a property of the corresponding identifiable as follows:
- Property name: "CGMES.OperationalLimitSet"
- Property value: Serialized json of all "\<rdfid\>:\<name\>"

That way there is only one property key for storing limit sets id/name overall.

CGMES export has been adapted accordingly.

- [ ] Yes
- [X] No

**Other information**:

This is a refactoring of: https://github.com/powsybl/powsybl-core/pull/3139

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
@@ -94,7 +94,7 @@ public class OperationalLimitConversion extends AbstractIdentifiedObjectConversi
      * @param limitSetId The OperationalLimitSet id to store.
      * @param limitSetName The OperationalLimitSet name to store.
      */
-    private void storeCgmesId(Identifiable<?> identifiable, String limitSetId, String limitSetName) {
+    private void storeOperationalLimitSetIdentifiers(Identifiable<?> identifiable, String limitSetId, String limitSetName) {
         try {
             ObjectMapper mapper = new ObjectMapper();
             JsonNode node;
@@ -121,12 +121,12 @@ public class OperationalLimitConversion extends AbstractIdentifiedObjectConversi
     private void createLimitsAdder(int terminalNumber, String limitSubClass, String limitSetId, String limitSetName, Branch<?> b) {
         if (terminalNumber == 1) {
             OperationalLimitsGroup limitsGroup = b.getOperationalLimitsGroup1(limitSetId).orElseGet(() -> {
-                storeCgmesId(b, limitSetId, limitSetName);
+                storeOperationalLimitSetIdentifiers(b, limitSetId, limitSetName);
                 return b.newOperationalLimitsGroup1(limitSetId); });
             loadingLimitsAdder1 = context.loadingLimitsMapping().getLoadingLimitsAdder(limitsGroup, limitSubClass);
         } else if (terminalNumber == 2) {
             OperationalLimitsGroup limitsGroup = b.getOperationalLimitsGroup2(limitSetId).orElseGet(() -> {
-                storeCgmesId(b, limitSetId, limitSetName);
+                storeOperationalLimitSetIdentifiers(b, limitSetId, limitSetName);
                 return b.newOperationalLimitsGroup2(limitSetId); });
             loadingLimitsAdder2 = context.loadingLimitsMapping().getLoadingLimitsAdder(limitsGroup, limitSubClass);
         } else {
@@ -143,7 +143,7 @@ public class OperationalLimitConversion extends AbstractIdentifiedObjectConversi
      */
     private void createLimitsAdder(String limitSubClass, String limitSetId, String limitSetName, DanglingLine dl) {
         OperationalLimitsGroup limitsGroup = dl.getOperationalLimitsGroup(limitSetId).orElseGet(() -> {
-            storeCgmesId(dl, limitSetId, limitSetName);
+            storeOperationalLimitSetIdentifiers(dl, limitSetId, limitSetName);
             return dl.newOperationalLimitsGroup(limitSetId); });
         loadingLimitsAdder = context.loadingLimitsMapping().getLoadingLimitsAdder(limitsGroup, limitSubClass);
     }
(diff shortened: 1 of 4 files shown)


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
```diff
@@ -1203,10 +1206,17 @@ private static void writeLimitsGroup(Identifiable<?> identifiable, OperationalLi
...
         String operationalLimitSetId;
         String operationalLimitSetName;
-        String propertyKey = Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.OPERATIONAL_LIMIT_SET + "_" + limitsGroup.getId();
+        String propertyKey = Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.OPERATIONAL_LIMIT_SET;
         if (identifiable.hasProperty(propertyKey)) {
             operationalLimitSetId = limitsGroup.getId();
-            operationalLimitSetName = identifiable.getProperty(propertyKey);
+            try {
+                ObjectMapper mapper = new ObjectMapper();
+                JsonNode propertyNode = mapper.readTree(identifiable.getProperty(propertyKey));
```
alicecaron, line 1214: Should the check on `identifiable.hasProperty(propertyKey)` be now equivalent to the test on `identifiable.hasProperty(propertyKey)` + `null != propertyNode && null != limitsGroupNode` or equivalent? If so the else condition should include also this case?  
Maybe we should also include the cases when `propertyNode` is null to avoid NPE exception (need to really validate if that exception is thrown if the JsonNode is not found)

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
```diff
@@ -81,6 +86,30 @@ private void setVoltageLevelForVoltageLimit(Terminal terminal) {
...
 
+    /**
+     * Store the CGMES OperationalLimitSet id/name pair in a property of the identifiable.
+     * If the property already exists, meaning it has been created for another limit set of that identifiable,
+     * then append the id/name pair to the property value (which actually represents a serialized json).
+     * @param identifiable The Branch, DanglingLine, ThreeWindingsTransformer where the limit set id/name are stored.
+     * @param limitSetId The OperationalLimitSet id to store.
+     * @param limitSetName The OperationalLimitSet name to store.
+     */
+    private void storeCgmesId(Identifiable<?> identifiable, String limitSetId, String limitSetName) {
```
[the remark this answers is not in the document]
rcourtier: I'm not sure about this renaming.
Cgmes and OperationalLimitSet is redundant. Also we don't store the whole object but only 2 of its identifiers (id and name).
Would you agree upon ```storeOperationalLimitSetIdentifiers``` ?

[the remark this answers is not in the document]
alicecaron: yes totally

alicecaron (approved):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
```diff
@@ -28,7 +33,7 @@ public class OperationalLimitConversion extends AbstractIdentifiedObjectConversi
     private static final String OPERATIONAL_LIMIT_SUBCLASS = "OperationalLimitSubclass";
     private static final String OPERATIONAL_LIMIT_SET_ID = "OperationalLimitSet";
     private static final String OPERATIONAL_LIMIT_SET_NAME = "OperationalLimitSetName";
-    private static final String PROPERTY_PREFIX = Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.OPERATIONAL_LIMIT_SET + "_";
+    private static final String PROPERTY_PREFIX = Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.OPERATIONAL_LIMIT_SET;
```
flo-dup, line 36: Could you rename it? it's not a property prefix anymore.
Besides, could you put it public? so that the equipement export can directly refer to this variable.
And maybe add "_identifiers" at the end?

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/OperationalLimitsGroupTest.java
```diff
@@ -51,6 +54,11 @@ void importMultipleLimitsGroupsOnSameLineEndTest() {
         // When an end has only 1 set, this set gets selected, otherwise none is
         assertTrue(line.getSelectedOperationalLimitsGroup1().isPresent());
         assertFalse(line.getSelectedOperationalLimitsGroup2().isPresent());
+
+        // The CGMES id/name have been correctly imported
+        String propertyKey = Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.OPERATIONAL_LIMIT_SET;
+        String propertyValue = "{\"OLS_1\":\"SPRING\",\"OLS_2\":\"SPRING\",\"OLS_3\":\"WINTER\"}";
```
flo-dup, line 60:
```suggestion
        String propertyValue = """
                {"OLS_1":"SPRING","OLS_2":"SPRING","OLS_3":"WINTER"}""";
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
```diff
@@ -81,6 +86,30 @@ private void setVoltageLevelForVoltageLimit(Terminal terminal) {
...
+    private void storeOperationalLimitSetIdentifiers(Identifiable<?> identifiable, String limitSetId, String limitSetName) {
+        try {
+            ObjectMapper mapper = new ObjectMapper();
+            JsonNode node;
+            if (identifiable.hasProperty(PROPERTY_PREFIX)) {
+                node = mapper.readTree(identifiable.getProperty(PROPERTY_PREFIX));
+            } else {
+                node = mapper.createObjectNode();
+            }
+            ((ObjectNode) node).put(limitSetId, limitSetName);
```
flo-dup, line 106: it's a bit unexpected to use the id as the fieldName, even though that's the easiest way to do this...

flo-dup (changes requested):

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/OperationalLimitsGroupTest.java
rcourtier: Fixed

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
rcourtier: Fixed

flo-dup (approved):

PR #3245, opened 2024-12-04
Fix fork disconnection

### LINKED ISSUE #3236: Be able to disconnect a line from another one

When disconnecting a Line the graph traversal do not look for connections between two Connectable (2 lines) therfore do not disconnect themselves.

In the case of a fork like the image below if we want to disconnect the first line it is still connected to the second line.
Fork

In the case of a fork like the image above if we want to disconnect the first line then it should no longer be connected to the second line. The highlighted switch should be open.

### Describe the motivation

Be able to isolate a line from another one.

rolnico:

Fixes #3236

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/time-series/time-series-api/src/main/java/com/powsybl/timeseries/IrregularTimeSeriesIndex.java
+++ b/time-series/time-series-api/src/main/java/com/powsybl/timeseries/IrregularTimeSeriesIndex.java
@@ -15,10 +15,7 @@ import gnu.trove.list.array.TLongArrayList;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.time.Instant;
-import java.util.Arrays;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Objects;
+import java.util.*;
 import java.util.stream.Stream;
 
 /**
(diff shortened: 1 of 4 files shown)


flo-dup (changes requested):
Great!

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractForkConnectDisconnectTest.java
```diff
@@ -0,0 +1,247 @@
...
+
+        // D_L1 should be open but not D_L2
+        assertTrue(disconnectorL1.isOpen());
+        assertFalse(disconnectorL2.isOpen());
+    }
+
+    private void assertTerminalsStatus(List<? extends Terminal> terminals, boolean terminalsShouldBeConnected) {
+        if (terminalsShouldBeConnected) {
+            terminals.forEach(terminal -> assertNotNull(terminal.getBusView().getBus()));
+            terminals.forEach(terminal -> assertTrue(terminal.isConnected()));
```
flo-dup, line 241: these two lines are equivalent in node breaker topology (at least in the known implementations)

rolnico: yes but I preferred to make it obvious and visible since it's part of the issue

flo-dup (approved):
I remove the direct link to the issue as it's only solving disconnection. The connection part might also need a change.

alicecaron (approved):

PR #3246, opened 2024-12-04
Fix missing buses in bus breaker view

flo-dup:

No

Bug fix

Connectables which are not connected to any switch / internal connection in node breaker do not have a calculated bus in bus breaker view, whereas connectables which are connected to a dangling internal connection do have a calculated bus.

All connectables of a node breaker voltage level do have a calculated bus in bus breaker view (in particular connectables which are not connected to any switch / internal connection)

- [x] Yes
- [ ] No


geofjamg (approved):
@flo-dup thanks for this looooooooonnnnng time topology issue. Thanks to this we are now allowed to clean many bus/breaker buses (coming from node/breaker model) null buses which cannot exist anymore. 
 - One example (from myself...) of a bad workaround is the utility class https://github.com/powsybl/powsybl-core/blob/main/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Equipments.java which does not seem to be used in core itself but maybe in other repositories. 
 - Also we need to search for now unnecessary null checks.
 - Maybe there are some places like in CGMES importer where we try to find some equivalent terminals (to check I'm not 100% sure) to also workaround this issue when importing a bus/breaker model.

olperr1 (approved):

flo-dup:
I marked this PR as draft as some cleaning could be done following this (the connectableBus getter becomes non-nullable in bus breaker view). That is now in a separate PR, hence we can merge this.

PR #3247, opened 2024-12-04
Remove unused test files

flo-dup:

No

Cleaning

Many reference files are unused in powsybl-iidm-serde unit tests

The unused files are deleted.

- [ ] Yes
- [x] No


iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/TopologyLevelTest.java
```diff
@@ -43,13 +42,13 @@ void testConversion() throws IOException {
 
     private void testConversion(Network network) throws IOException {
 
-        ExportOptions options = new ExportOptions();
-        testWriteVersionedXml(network, options.setTopologyLevel(TopologyLevel.NODE_BREAKER), "fictitiousSwitchRef.xml", CURRENT_IIDM_VERSION);
+        testWriteXml(network, "/fictitiousSwitchRef.xml");
```
olperr1, line 45: Do we really want to have no tests on previous versions?

I think the problem here is essentially the fact that only the last version is checked.
Shouldn't we have something like this instead?
```java
testForAllVersionsSince(IidmVersion.V_1_0, v -> {
    try {
        testWriteVersionedXml(network, options.setTopologyLevel(TopologyLevel.NODE_BREAKER), "fictitiousSwitchRef.xml", v);
    } catch (IOException e) {
        throw new RuntimeException(e);
    }
});
```

PR #3256, opened 2024-12-19
Add fix to Matpower importer to import some pglib files

geofjamg:

No

Feature

Some matpower files like pglib_opf_case1354_pegase.m from https://github.com/power-grid-lib/pglib-opf cannot be imported because the generator column count correspond to v1 format and not v2. 

We accept to read the file anyway even with missing columns.

- [ ] Yes
- [x] No

**Other information**:


matpower/matpower-model/src/main/java/com/powsybl/matpower/model/MatpowerReader.java
```diff
@@ -145,17 +157,19 @@ private static void readGenerators(Matrix generators, MatpowerModel model) {
...
-            gen.setQc1Min(generators.getDouble(row, 12));
-            gen.setQc1Max(generators.getDouble(row, 13));
-            gen.setQc2Min(generators.getDouble(row, 14));
-            gen.setQc2Max(generators.getDouble(row, 15));
-            gen.setRampAgc(generators.getDouble(row, 16));
-            gen.setRampTenMinutes(generators.getDouble(row, 17));
-            gen.setRampThirtyMinutes(generators.getDouble(row, 18));
-            gen.setRampQ(generators.getDouble(row, 19));
-            gen.setApf(generators.getDouble(row, 20));
+            if (generators.getDimensions()[1] > MATPOWER_V1_GENERATORS_COLUMNS) {
```
caioluke, line 160: generatorColumns  > MATPOWER_V1_GENERATORS_COLUMNS ?

olperr1: `generatorColumns` is not defined in this method (`readGenerators`) but in `checkNumberOfColumns`.

olperr1, line 160: As discussed together, it would be safer to use the following test here:
`if (generators.getDimensions()[1] >= MATPOWER_V2_GENERATORS_COLUMNS)`

=> 

- if dimension < `MATPOWER_V1_GENERATORS_COLUMNS`: error
- if dimension in `[ MATPOWER_V1_GENERATORS_COLUMNS ; MATPOWER_V2_GENERATORS_COLUMNS [`: considered as V1 format;
- if dimension >= `MATPOWER_V2_GENERATORS_COLUMNS`: considered as V2 format.

matpower/matpower-model/src/main/java/com/powsybl/matpower/model/MatpowerReader.java
```diff
@@ -145,17 +162,19 @@ private static void readGenerators(Matrix generators, MatpowerModel model) {
...
-            gen.setQc1Min(generators.getDouble(row, 12));
-            gen.setQc1Max(generators.getDouble(row, 13));
-            gen.setQc2Min(generators.getDouble(row, 14));
-            gen.setQc2Max(generators.getDouble(row, 15));
-            gen.setRampAgc(generators.getDouble(row, 16));
-            gen.setRampTenMinutes(generators.getDouble(row, 17));
-            gen.setRampThirtyMinutes(generators.getDouble(row, 18));
-            gen.setRampQ(generators.getDouble(row, 19));
-            gen.setApf(generators.getDouble(row, 20));
+            if (generatorColumnsToRead > 9) {
```
olperr1, line 165: Here, the method will throw an exception in case of a generator table in v1 format (method called with `MATPOWER_V1_GENERATORS_COLUMNS` which equals 10).

But it may be clearer to have this test instead:
```suggestion
            if (generatorColumnsToRead == MATPOWER_V2_GENERATORS_COLUMNS) {
```

because it will be easier to relate it to the possible values for `generatorColumnsToRead`.

olperr1 (changes requested):

matpower/matpower-model/src/main/java/com/powsybl/matpower/model/MatpowerReader.java
geofjamg: I tried another design

olperr1: Thanks, it looks good to me.

olperr1 (approved):

PR #3267, opened 2025-01-10
Replace CGMES.switchType property by CGMES.originalClass property.

rcourtier:

No.

Refactoring.

At CGMES import, if the Switch kind doesn't have a direct map to IIDM, i.e. if its class is: Switch, ProtectedSwitch, GroundDisconnector, Jumper, then its CGMES original class is stored in a CGMES.switchType property. The purpose is to be able to export back to CGMES the object with the correct class.
Similarly, at CGMES import, the original CGMES class of load or generator is stored in a CGMES.originalClass property.

For consistency purposes, drop the use of CGMES.switchType property in favor of CGMES.originalClass.

*Does this PR introduce a breaking change or deprecate an API?**
- [X] Yes
- [ ] No

Any access to the `CGMES.switchType` property should be replaced with `CGMES.originalClass` property. For instance:
```java
String cgmesSwitchType = network.getSwitch("someSwitchUUID").getProperty("CGMES.switchType");
```
should be replaced by:
```java
String cgmesSwitchType = network.getSwitch("someSwitchUUID").getProperty(Conversion.PROPERTY_CGMES_ORIGINAL_CLASS);
```
Similarly, any network saved to a serialized IIDM file should be updated by replacing all occurences of `CGMES.switchType` with `CGMES.originalClass` property, in case the network needs to be exported to CGMES.

**Other information**:

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/SwitchConversion.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/SwitchConversion.java
@@ -125,7 +125,7 @@ public class SwitchConversion extends AbstractConductingEquipmentConversion impl
     }
 
     private void addTypeAsProperty(Switch s) {
-        s.setProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + "switchType", p.getLocal("type"));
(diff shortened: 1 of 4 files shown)


flo-dup (commented):
This is a breaking change... we should warn the users about it.

flo-dup (approved):

PR #3268, opened 2025-01-13
Allow standByAutomaton extension to have inconsistent voltage thresholds if not in stand by

### LINKED ISSUE #3261: Allow StandbyAutomaton to have inconsistent voltage config when its standBy mode is not active

When importing a network containing `StaticVarCompensator` with `StandbyAutomaton` extension, even if the extension is in `standBy=false` (because we don't want to use it in a simulator) the import fails if the low voltage threshold is not strictly less than the high voltage threshold (for instance if both are set to `0` because not really set for the use case).

When importing a network containing `StaticVarCompensator` with `StandbyAutomaton` extension with:
- a `standBy` mode set to false
- the low voltage threshold not strictly less than the high voltage threshold  

Then the import does not fail and I can see a warning message instead about the inconsistence of the voltage config.  

Then, if I activate the `standBy` mode and the voltage config is not right (low voltage threshold >= high voltage threshold) then it must throw an error (Inconsistent low and high voltage thresholds).

### Describe the motivation

Add more flexibility when reading networks containing `StandbyAutomaton` extensions with a `standBy` mode deactivated.  

### Extra Information

Add flexibility in the check on the voltage config:
- in the `StandbyAutomatonImpl::checkVoltageConfig` method add the check on the standBy value to throw exception only in the case standBy=true, and log a warning message if false.

Add control to check voltage config correctness when activating a `StandbyAutomaton`:
- in the `StandbyAutomatonImpl::setStandby` method, before updating the`standBy` attribute to the new value, call the `StandbyAutomatonImpl::checkVoltageConfig` method to check that the voltage values are correct with the new standBy value.

alicecaron:

Fixes #3261

- [X] Yes
- [ ] No

Custom IIDM implementation maintainers should allow having inconsistent low/high voltages in the `StandbyAutomaton` extension (which is an extension on `StaticVarCompensator`) when "standBy" is false. In this case, setting "standBy" to true should then throw an exception.

**Other information**:


iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/extensions/AbstractStandbyAutomatonTest.java
```diff
@@ -68,6 +66,12 @@ public void test() {
             fail();
         } catch (Exception ignored) {
         }
+
+        // when standby is false do not throw error on inconsistent low and high voltage thresholds use case
+        standbyAutomaton.setStandby(false);
+        assertFalse(standbyAutomaton.isStandby());
+        standbyAutomaton.setHighVoltageThreshold(200f);
+        assertEquals(200f, standbyAutomaton.getHighVoltageThreshold(), 0.0);
```
olperr1, line 74: We could add an assertion checking that setting the standBy mode when data are inconsistent leads to an exception:
```suggestion
        assertEquals(200f, standbyAutomaton.getHighVoltageThreshold(), 0.0);
        
        // but an exception is thrown when the standby mode is latter used
        assertThrows(IllegalArgumentException.class, () -> standbyAutomaton.setStandby(true));
```

olperr1 (changes requested):

olperr1 (approved):

PR #3273, opened 2025-01-14
IIDM ValidationLevel: Add a silent option.

### LINKED ISSUE #3185: Network validation: too many reported errors

All validation checks are performed always, independent of the minimum validation level established in the network.
Even if we have set that we accept the network to be in `EQUIPMENT` validation level, any change will emit logs and report errors related to the verification of steady state attributes.

This can be easily reproduced with the following code:

```java
Network network = Network.create("test", "iidm");
VoltageLevel vl = network.newSubstation().setId("s1").add()
        .newVoltageLevel().setId("vl1").setNominalV(100).setTopologyKind(TopologyKind.NODE_BREAKER).add();
network.setMinimumAcceptableValidationLevel(ValidationLevel.EQUIPMENT);
vl.newLoad().setId("l1").setNode(0).add();
```

The logger contents will show the following errors:

```
ERROR c.p.i.n.ValidationUtil | Load 'l1': p0 is invalid
ERROR c.p.i.n.ValidationUtil | Load 'l1': q0 is invalid
```

We expect to obtain errors based in the current minimum accepted validation level. 

If we want to keep the validations for levels higher than the minimum accepted (to compute the actual validation level of the network no matter what is the minimum set), at least we should be able to configure that we do not want to emit logs/report errors for the "higher" level checks.

Currently validation checks can be called with a boolean flag that controls the behaviour of error processing: either raise an exception or simply log/report the error. This flag could be changed to an enum with a third option to run the checks _silently_ (no exception, no log, no report), just to update the validation level of the network.

### Describe the motivation

This suggestion is made in the context of CGMES import refactoring to split the conversion of networks in two stages:

1. Building the network structure from equipment (EQ) files that contain only connectivity and electric characteristics of the devices.
2. Apply information from a steady state hypothesis (SSH) file.

During the first stage we do not want errors to be emitted for missing SSH data.

### Extra Information

When we request a validation report, we miss the identification of the related equipment. In the logger, we have messages  like `Load 'l1': p0 is invalid`. In the output from reports we get only the attribute with error, but not the equipment identifier.

To check it, run the following code:

```java
Network network = Network.create("test", "iidm");
VoltageLevel vl = network.newSubstation().setId("s1").add()
        .newVoltageLevel().setId("vl1").setNominalV(100).setTopologyKind(TopologyKind.NODE_BREAKER).add();
network.setMinimumAcceptableValidationLevel(ValidationLevel.EQUIPMENT);
vl.newLoad().setId("l1").setNode(0).add();

ReportNode reportNode = ReportNode.newRootReportNode().withMessageTemplate("root", "Report").build();
network.runValidationChecks(false, reportNode);

StringWriter sw = new StringWriter();
reportNode.print(sw);
System.out.println(sw.toString());
```
It will produce this output:
```
+ Report
   + Running validation checks on IIDM network test
      p0 is invalid
      p0 is invalid
```

marqueslanauja:

Feature
Closes #3185.

When only an `EQ` file is imported after setting the validation level to `EQUIPMENT`, the validation process generates warnings indicating that some equipment attributes are not defined. These warnings are unnecessary, as we know that only the equipment attributes are imported.

In the validation process, a new option (IGNORE) has been added. The validation level warnings are only generated when the `runValidationChecks` is executed. With this PR, it will be possible to import the `EQ` and `SSH` files separately without generating unnecessary warnings.

- [x] Yes
- [ ] No

The `boolean throwException` parameter of `ValidationUtil.validate(...)` was replaced by an enum value.
If you used this method, you can use one of the following values:
- `ValidationUtil.ActionOnError.THROW_EXCEPTION`: throw an exception if an error is detected - corresponds to older `throwException` = `true`);
- `ValidationUtil.ActionOnError.LOG_ERROR`: report all the errors in the given reportNode and on the console - corresponds to older `throwException` = `false`);
- `ValidationUtil.ActionOnError.IGNORE`: ignore all the errors (no exception, no logs and no reports)

**Other information**:


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -23,6 +23,12 @@
  */
 public final class ValidationUtil {
 
+    public enum LogLevel {
```
zamarrenolm, line 26: Instead of `LogLevel`(that could be confused with the logger behavior), I prefer `ActionOnError` for this enum name.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -63,33 +69,37 @@ private static void logError(Validable validable, String message, ReportNode rep
...
         }
-        logError(validable, message, reportNode);
+        if (throwException.equals(LogLevel.LOG_ERROR)) {
+            logError(validable, message, reportNode);
+        }
     }
 
     public static void throwExceptionOrLogError(Validable validable, String message, ValidationLevel validationLevel, ReportNode reportNode) {
-        throwExceptionOrLogError(validable, message, validationLevel == ValidationLevel.STEADY_STATE_HYPOTHESIS, reportNode);
+        throwExceptionOrLogError(validable, message, checkValidationLogLevel(validationLevel), reportNode);
```
zamarrenolm, line 82: Following the proposed name change, replace `checkValidationLogLevel` by `checkValidationActionOnError`

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/ValidationLevelLogsTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+class ValidationLevelLogsTest {
+
+    @Test
+    void equipmentTest() throws IOException {
+        Network network = networkAssociatedWithReportNode();
+
+        network.setMinimumAcceptableValidationLevel(ValidationLevel.EQUIPMENT);
+        createNetworkWithEquipmentValidationLevel(network);
+
```
zamarrenolm, line 34: Maybe assert at this point that no validation messages have been produced. Something like:

```java
// We check that the SILENT option for action on error works
// While creating only network equipment if we have established validation level to EQUIPMENT,
// We should not see anything reported
assertTrue(checkReportNode("""
      Test validationLevel Logs
      """, network.getReportNodeContext().getReportNode()));
```

marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
marqueslanauja: Done

zamarrenolm (dismissed):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -23,6 +23,12 @@
  */
 public final class ValidationUtil {
 
+    public enum ActionOnError {
+        THROW_EXCEPTION,
+        LOG_ERROR,
+        SILENT,
```
olperr1, line 29: I think it would be better to rename this option to use a verb instead, since the enum describe an "action".
We could use for instance, `IGNORE`.

olperr1 (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -63,33 +69,37 @@ private static void logError(Validable validable, String message, ReportNode rep
         LOGGER.error("{}{}", validable.getMessageHeader(), message);
     }
 
-    private static void throwExceptionOrLogError(Validable validable, String message, boolean throwException, ReportNode reportNode) {
-        if (throwException) {
+    private static void throwExceptionOrLogError(Validable validable, String message, ActionOnError throwException, ReportNode reportNode) {
```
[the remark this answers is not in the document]
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -63,33 +69,37 @@ private static void logError(Validable validable, String message, ReportNode rep
...
+        if (throwException.equals(ActionOnError.THROW_EXCEPTION)) {
```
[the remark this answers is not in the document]
marqueslanauja: Done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -63,33 +69,37 @@ private static void logError(Validable validable, String message, ReportNode rep
...
             throw new ValidationException(validable, message);
         }
-        logError(validable, message, reportNode);
+        if (throwException.equals(ActionOnError.LOG_ERROR)) {
```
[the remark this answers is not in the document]
marqueslanauja: Done

olperr1 (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -63,47 +69,51 @@ private static void logError(Validable validable, String message, ReportNode rep
...
-        logError(validable, message, reportNode);
+        if (actionOnError == ActionOnError.LOG_ERROR) {
+            logError(validable, message, reportNode);
+        }
     }
 
     public static void throwExceptionOrLogError(Validable validable, String message, ValidationLevel validationLevel, ReportNode reportNode) {
-        throwExceptionOrLogError(validable, message, validationLevel == ValidationLevel.STEADY_STATE_HYPOTHESIS, reportNode);
+        throwExceptionOrLogError(validable, message, checkValidationActionOnError(validationLevel), reportNode);
     }
```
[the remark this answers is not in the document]
marqueslanauja: Done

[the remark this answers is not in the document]
olperr1: Thanks!

olperr1 (approved):

Issue #3248, opened 2024-12-09
Force creation of the extension in the network modifications that creates new feeders


colinepiloquet:
### Describe the current behavior

When using the network modifications `CreateFeederBay` or `CreateBranchFeederBays`, if the order position given is correct but not available, then a warning is returned and the connectable is still created.

### Describe the expected behavior

If the order position is incorrect, an optional parameter should indicate whether an error should be returned and in case an error is returned, the connectable should not be created. The optional parameter allows to avoid a change in the behavior of the modifications.

### Describe the motivation

This allows to force users to indicate a correct order position.

PR #3275, opened 2025-01-16
[Network modifications] Add boolean indicating whether position extensions must be created. 

colinepiloquet:

#3248


iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeederBays.java
```diff
@@ -65,6 +68,28 @@ public void apply(Network network, NamingStrategy namingStrategy, boolean throwE
...
+                    if (getForceExtensionCreation(side)) {
+                        return;
+                    } else {
+                        createExtension.put(side, false);
+                    }
+                } else {
+                    createExtension.put(side, true);
+                }
+            }
+        }
```
rolnico: extract to a new private method

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeederBays.java
```diff
@@ -65,6 +68,28 @@ public void apply(Network network, NamingStrategy namingStrategy, boolean throwE
         if (!setAdderConnectivity(network, reportNode, throwException)) {
             return;
         }
+
+        // Check if ConnectablePosition extension can and should be created
+        Map<Integer, Boolean> createExtension = new HashMap<>();
+        for (int side : sides) {
+            String busOrBusbarSectionId = getBusOrBusbarSectionId(side);
+            Identifiable<?> busOrBusbarSection = network.getIdentifiable(busOrBusbarSectionId);
+            if (busOrBusbarSection instanceof BusbarSection bbs) {
```
rolnico: Rename variable to `busbarSection` to be consistent with existing code
```suggestion
            if (busOrBusbarSection instanceof BusbarSection busbarSection) {
```

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeederBays.java
```diff
@@ -119,40 +144,73 @@ private boolean checkOrders(int side, VoltageLevel voltageLevel, ReportNode repo
...
+        if (positionOrder != null && busOrBusbarSection instanceof Bus bus) {
+            VoltageLevel voltageLevel = bus.getVoltageLevel();
             ignoredPositionOrder(reportNode, positionOrder, voltageLevel);
             LOGGER.warn("Voltage level {} is BUS_BREAKER. Position order {} is ignored", voltageLevel.getId(), positionOrder);
         }
         return true;
     }
 
-    private boolean checkOrderValue(int side, BusbarSection busbarSection, Set<Integer> takenFeederPositions, ReportNode reportNode) {
+    private boolean checkOrderValue(int side, BusbarSection busbarSection, Set<Integer> takenFeederPositions, ReportNode reportNode, boolean throwException) {
```
rolnico, line 179: refacto method to reduce complexity

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeederBays.java
```diff
@@ -119,40 +144,73 @@ private boolean checkOrders(int side, VoltageLevel voltageLevel, ReportNode repo
...
+            if (getForceExtensionCreation(side)) {
+                LOGGER.error("PositionOrder {} already taken.", positionOrder);
+                positionOrderAlreadyTakenReport(reportNode, positionOrder, TypedValue.ERROR_SEVERITY);
+                if (throwException) {
+                    throw new PowsyblException("PositionOrder " + positionOrder + " already taken.");
+                }
+            } else {
+                LOGGER.warn("PositionOrder {} already taken.", positionOrder);
+                positionOrderAlreadyTakenReport(reportNode, positionOrder, TypedValue.WARN_SEVERITY);
+            }
```
rolnico: to reduce complexity, you can factorize this part, it appears multiple times with similar parameters

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeederBays.java
```diff
@@ -65,6 +68,28 @@ public void apply(Network network, NamingStrategy namingStrategy, boolean throwE
...
+        Map<Integer, Boolean> createExtension = new HashMap<>();
+        for (int side : sides) {
+            String busOrBusbarSectionId = getBusOrBusbarSectionId(side);
+            Identifiable<?> busOrBusbarSection = network.getIdentifiable(busOrBusbarSectionId);
+            if (busOrBusbarSection instanceof BusbarSection bbs) {
+                VoltageLevel voltageLevel = bbs.getTerminal().getVoltageLevel();
+                Set<Integer> takenFeederPositions = TopologyModificationUtils.getFeederPositions(voltageLevel);
+                boolean checkOrderValue = checkOrderValue(side, bbs, takenFeederPositions, reportNode, throwException);
+                if (!checkOrderValue) {
+                    if (getForceExtensionCreation(side)) {
```
rolnico: I feel like the use of this boolean is inverted compared to the name of the boolean, right?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeederBays.java
```diff
@@ -65,6 +68,28 @@ public void apply(Network network, NamingStrategy namingStrategy, boolean throwE
...
+                        return;
+                    } else {
+                        createExtension.put(side, false);
+                    }
+                } else {
+                    createExtension.put(side, true);
+                }
```
rolnico: I would rewrite this part like this:
```suggestion
                if (!checkOrderValue && !shouldForceExtensionCreation(side)) {
                    return;
                }
                createExtension.put(side, checkOrderValue);
```
(note that I renamed the boolean and used it as I understands its value)

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeederBays.java
```diff
@@ -237,7 +293,7 @@ private void createExtensionAndTopology(Connectable<?> connectable, Network netw
             int positionOrder = getPositionOrder(side);
             if (!takenFeederPositions.isEmpty() || voltageLevel.getConnectableStream().filter(c -> !(c instanceof BusbarSection)).count() == 1) {
                 // check that there are existing position extensions on other connectables or there is only one connectable (that we added)
-                if (checkOrderValue(side, (BusbarSection) busOrBusbarSection, takenFeederPositions, reportNode)) { // BusbarSection as voltage level is NODE_BREAKER
+                if (createExtension.get(side)) { // BusbarSection as voltage level is NODE_BREAKER
```
rolnico:
```suggestion
                if (Boolean.TRUE.equals(createExtension.get(side))) { // BusbarSection as voltage level is NODE_BREAKER
```

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeederBays.java
```diff
@@ -218,14 +276,12 @@ private static boolean checkNetworks(Connectable<?> connectable, Network network
         return true;
     }
 
-    private void createExtensionAndTopology(Connectable<?> connectable, Network network, NamingStrategy namingStrategy, ReportNode reportNode) {
+    private void createExtensionAndTopology(Map<Integer, Boolean> createExtension, Connectable<?> connectable, Network network, NamingStrategy namingStrategy, ReportNode reportNode) {
```
rolnico, line 284: there is a high duplication of the logic between this method and the checks you added at the beginning of the `apply` method. You should rewrite the code to avoid this duplication. 

You could for example do the check for each side, at the same time save in some record the information needed to create the extension and the topology, and then in `createExtensionAndTopology` just loop on the list of records and create what is needed.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeederBays.java
colineplqt: Well in my mind the idea was, if the order position is already taken:
- false -> don't create extension but create the connectable (current behavior)
- true -> throwOrLog

This is why for me the name was coherent because it indicates if the extension has to be created or not. Maybe I could call it throwOrLogWrongOrderPosition or something even more explicit ?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeederBays.java
rolnico: Careful, this modification has to be done in the whole code!

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeederBays.java
colineplqt: You are right! I think most of the logic duplication is because we check again if the extension should be created with the condition that if there are no other extension. I think this check could be done in checkOrderValue and the creation of the extension would only check the `createExtensionMap`.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeederBays.java
```diff
@@ -73,7 +88,49 @@ public void apply(Network network, NamingStrategy namingStrategy, boolean throwE
...
+    private Optional<Map<Integer, Boolean>> getCreateExtensionBySideMap(Network network, boolean throwException, ReportNode reportNode) {
+        Map<Integer, Boolean> createExtensionMap = new HashMap<>();
+        for (int side : sides) {
+            Optional<Boolean> createExtension = checkIfExtensionShouldBeCreated(network, side, throwException, reportNode);
+            if (createExtension.isEmpty()) {
+                return Optional.empty();
+            }
+            createExtensionMap.put(side, createExtension.get());
+        }
+        return Optional.of(createExtensionMap);
```
rolnico: Instead of returning an optional, Maybe you could return a Map in every case, by replacing your empty Optional by an empty Map?
```suggestion
        for (int side : sides) {
            Optional<Boolean> createExtension = checkIfExtensionShouldBeCreated(network, side, throwException, reportNode);
            if (createExtension.isEmpty()) {
                return new HashMap<>();
            }
            createExtensionMap.put(side, createExtension.get());
        }
        return createExtensionMap;
```

rolnico: If you choose to do so, you'll have to modify l77 too

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeederBays.java
```diff
@@ -112,46 +169,68 @@ private boolean checkOrders(int side, VoltageLevel voltageLevel, ReportNode repo
...
         Integer positionOrder = getPositionOrder(side);
+        boolean logOrThrowIfIncorrectPositionOrder = getLogOrThrowIfIncorrectPositionOrder(side);
 
         if (takenFeederPositions.contains(positionOrder)) {
-            LOGGER.warn("PositionOrder {} already taken. No position extension created.", positionOrder);
-            positionOrderAlreadyTakenReport(reportNode, positionOrder);
-            return false;
+            String msg = POSITION_ORDER_STRING + positionOrder + " already taken.";
+            return logAndReport(logOrThrowIfIncorrectPositionOrder, throwException, msg,
+                severity -> positionOrderAlreadyTakenReport(reportNode, positionOrder, severity)
```
rolnico: As `logAndReport` always return false, for clarity I would prefer to keep it `void`  and have `return false;` on the next line

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeederBays.java
```diff
@@ -112,46 +169,68 @@ private boolean checkOrders(int side, VoltageLevel voltageLevel, ReportNode repo
...
+            LOGGER.error(message);
+        } else {
+            LOGGER.warn(message);
+        }
+
+        report.accept(severity);
+
+        if (logOrThrowIfIncorrectPositionOrder && throwException) {
+            throw new PowsyblException(message);
+        }
```
rolnico: Should this message still be logged if the exception is thrown right after?
I'm not really sure how you want to manage the log/report level here.
Would this proposal be correct?
```suggestion
        report.accept(severity);

        if (logOrThrowIfIncorrectPositionOrder) {
            logOrThrow(throwException, message);
        } else {
            LOGGER.warn(message);
        }

        return false;
```

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeederBays.java
```diff
@@ -73,7 +88,49 @@ public void apply(Network network, NamingStrategy namingStrategy, boolean throwE
...
+            Optional<Boolean> createExtension = checkIfExtensionShouldBeCreated(network, side, throwException, reportNode);
+            if (createExtension.isEmpty()) {
+                return Optional.empty();
+            }
+            createExtensionMap.put(side, createExtension.get());
+        }
+        return Optional.of(createExtensionMap);
+    }
+
+    private Optional<Boolean> checkIfExtensionShouldBeCreated(Network network, int side, boolean throwException, ReportNode reportNode) {
```
rolnico, line 105: I'm not a fan of the `Optional<Boolean>` here but I didn't find another way to do it yet

rolnico (changes requested):
Still a few things to fix but we are getting there! 
Good job on dividing the code in multiple smaller methods, it's way easier to read and to understand now

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeederBays.java
colineplqt: I think if throwException is false, then we want to log the error ?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeederBays.java
colineplqt: Yeah, me neither... Because I really need to keep the information to throw the exception. But I agree it looks a bit odd

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/AbstractCreateConnectableFeederBays.java
rolnico: Yes, that's what the `logOrThrow` does. But in your case, you log it with the ERROR level, while it is usually logged on the WARN level.

PR #3278, opened 2025-01-20
Cgmes import update. Import EQ and SSH separately. ShuntCompensators 

marqueslanauja:

Feature

EQ and SSH are imported together during the conversion process.

EQ and SSH are imported separately.
The objective is to ultimately import a single EQ file and multiple SSH files.

All the changes will be split in multiple smaller PRs so the implementation will be done in an incremental way.
Current main workflow will be kept for some equipment. We will move them one by one to the new scheme.
This PR contains only the changes for shuntCompensators.

When performing the update, we traverse all the IIDM `shuntCompensators` and for each `shuntCompensator` we obtain the SSH data and call a specific update method for every conversion class. 

The update method is located in the corresponding convert class and a different update class is defined for each
equipment. The update classes are defined as public static classes.

There are some methods with both implementations:

connect and convertedTerminals for importing EQ + SSH together
connectWithOnlyEq and convertedTerminalsWithOnlyEq for importing EQ + SSH separately
The validation scheme has been modified, initially the validation level is set to EQUIPMENT and only after the update is moved to SSH.

`NormalSections` is now recorded as a property.

- [X] Yes
- [ ] No

The following method of `AbstractConductingEquipmentConversion` was removed:
```java
PowerFlow powerFlowSV()
```

**Other information**:


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForShuntCompensators.java
```diff
@@ -37,99 +33,43 @@ public class RegulatingControlMappingForShuntCompensators {
...
     private void apply(ShuntCompensator shuntCompensator) {
-        CgmesRegulatingControlForShuntCompensator cgmesRc = mapping.get(shuntCompensator.getId());
-        apply(shuntCompensator, cgmesRc);
+        String rcId = mapping.get(shuntCompensator.getId());
+        apply(shuntCompensator, rcId);
     }
 
-    private void apply(ShuntCompensator shuntCompensator, CgmesRegulatingControlForShuntCompensator cgmesRc) {
-        if (cgmesRc == null) {
+    private void apply(ShuntCompensator shuntCompensator, String regulatingControlId) {
```
rcourtier: You could merge the two `apply` methods.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForShuntCompensators.java
```diff
@@ -37,99 +33,43 @@ public class RegulatingControlMappingForShuntCompensators {
...
-        }
-        RegulatingControl rc = parent.cachedRegulatingControls().get(rcId);
+        RegulatingControl rc = parent.cachedRegulatingControls().get(regulatingControlId);
         if (rc == null) {
-            context.missing(String.format("Regulating control %s", rcId));
-            setDefaultRegulatingControl(shuntCompensator);
             return;
         }
         // Finally, equipment participates in it regulating control,
         // and the regulating control information is present in the CGMES model
```
rcourtier:
```suggestion
        // Equipment participates in regulation and the regulating control information is present in the CGMES model
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Update.java
```diff
@@ -0,0 +1,153 @@
...
+        addPropertyBags(cgmes.shuntCompensators(), CgmesNames.SHUNT_COMPENSATOR, equipmentIdPropertyBag);
+        addPropertyBags(cgmes.equivalentShunts(), CgmesNames.EQUIVALENT_SHUNT, equipmentIdPropertyBag);
+
+        network.getShuntCompensators().forEach(shuntCompensator -> updateShuntCompensator(shuntCompensator, getPropertyBag(shuntCompensator.getId(), equipmentIdPropertyBag), context));
+        context.popReportNode();
+    }
+
+    private static void updateShuntCompensator(ShuntCompensator shuntCompensator, PropertyBag cgmesData, Context context) {
+        String isEquivalentShunt = shuntCompensator.getProperty(Conversion.PROPERTY_IS_EQUIVALENT_SHUNT);
+        if (isEquivalentShunt != null && Boolean.parseBoolean(isEquivalentShunt)) {
```
rcourtier:
```suggestion
        if (Boolean.parseBoolean(isEquivalentShunt)) {
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java
```diff
@@ -70,12 +63,94 @@ public void convert() {
...
+        boolean enabled = cgmesRegulatingControl.map(propertyBag -> findRegulatingOn(propertyBag, defaultRegulatingOn, DefaultValueUse.NOT_DEFINED)).orElse(defaultRegulatingOn);
+
+        setRegulation(shuntCompensator, targetV, targetDeadband, controlEnabled && enabled && isValidTargetV(targetV) && isValidTargetDeadband(targetDeadband));
+    }
+
+    private static void setRegulation(ShuntCompensator shuntCompensator, double targetV, double targetDeadband, boolean regulatingOn) {
+        if (regulatingOn) {
+            shuntCompensator.setTargetV(targetV).setTargetDeadband(targetDeadband).setVoltageRegulatorOn(true);
+        } else {
+            shuntCompensator.setVoltageRegulatorOn(false).setTargetV(targetV).setTargetDeadband(targetDeadband);
```
rcourtier: Could you please add line breaks after the set for better readability?

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/ShuntCompensatorUpdateTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+        readCgmesResources(network, DIR, "shuntCompensator_SSH.xml");
+        assertFirstSsh(network);
+
+        readCgmesResources(network, DIR, "shuntCompensator_SSH_1.xml");
+        assertSecondSsh(network);
+
+        assertFlowsBeforeSv(network);
+        readCgmesResources(network, DIR, "shuntCompensator_SV.xml");
+        assertFlowsAfterSv(network);
```
rcourtier, line 60: Could you also please add `SvShuntCompensatorSections` in CGMES SV file, and show that the IIDM `ShuntCompensator.getSolvedSectionCount()` is null before SV import/non-null after SV import?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java
```diff
@@ -70,12 +63,94 @@ public void convert() {
...
+    public static void update(ShuntCompensator shuntCompensator, PropertyBag cgmesData, Context context) {
+        updateTerminals(shuntCompensator, context, shuntCompensator.getTerminal());
+        updateSections(shuntCompensator, cgmesData, context);
+
+        boolean controlEnabled = cgmesData.asBoolean(CgmesNames.CONTROL_ENABLED, false);
+        updateRegulatingControl(shuntCompensator, controlEnabled, context);
+    }
+
+    private static void updateSections(ShuntCompensator shuntCompensator, PropertyBag cgmesData, Context context) {
+        int normalSections = Integer.parseInt(shuntCompensator.getProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.NORMAL_SECTIONS));
```
olperr1: This line will throw an exception if the property is absent. You should use a default value when `getProperty(...)` returns `null`.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java
```diff
@@ -70,12 +63,94 @@ public void convert() {
...
+        int defaultSections = getDefaultSections(shuntCompensator, normalSections, context);
+        int sections = getSections(cgmesData).orElse(defaultSections);
+        shuntCompensator.setSectionCount(Math.min(sections, shuntCompensator.getMaximumSectionCount()));
```
olperr1, line 86: Before this PR, `maximumSectionCount` was set to the max between the declared `maximumSectionCount` and the `sectionCount`. Now, the `sectionCount` is set as the min of these 2 values.
Shouldn't we increase the `maximumSectionCount` here instead of lowering the `sectionCount` (in order to keep the same behavior)?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java
```diff
@@ -70,12 +63,94 @@ public void convert() {
...
+    private static double getDefaultTargetDeadband(ShuntCompensator shuntCompensator, Context context) {
+        return getDefaultValue(null, shuntCompensator.getTargetDeadband(), 0.0, 0.0, context);
+    }
+
+    private static boolean getDefaultRegulatingOn(ShuntCompensator shuntCompensator, Context context) {
+        return getDefaultValue(null, shuntCompensator.isVoltageRegulatorOn(), false, false, context);
+    }
+
+    private static boolean isDefaultRegulatingControl(ShuntCompensator shuntCompensator, boolean controlEnabled) {
+        String regulatingControlId = shuntCompensator.getProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.REGULATING_CONTROL);
```
olperr1:
```suggestion
        String regulatingControlId = shuntCompensator.getProperty(Conversion.PROPERTY_REGULATING_CONTROL);
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java
```diff
@@ -70,12 +63,94 @@ public void convert() {
...
+
+        setRegulation(shuntCompensator, targetV, targetDeadband, controlEnabled && enabled && isValidTargetV(targetV) && isValidTargetDeadband(targetDeadband));
+    }
+
+    private static void setRegulation(ShuntCompensator shuntCompensator, double targetV, double targetDeadband, boolean regulatingOn) {
+        if (regulatingOn) {
+            shuntCompensator.setTargetV(targetV).setTargetDeadband(targetDeadband).setVoltageRegulatorOn(true);
+        } else {
+            shuntCompensator.setVoltageRegulatorOn(false).setTargetV(targetV).setTargetDeadband(targetDeadband);
+        }
```
olperr1, line 142: Can you add a comment to explain that the order of the call is important here?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java
```diff
@@ -70,12 +63,94 @@ public void convert() {
...
+        return Double.isFinite(sections) ? OptionalInt.of(Math.abs(fromContinuous(sections))) : OptionalInt.empty();
+    }
+
+    private static int getDefaultSections(ShuntCompensator shuntCompensator, int normalSections, Context context) {
+        return getDefaultValue(normalSections, shuntCompensator.getSectionCount(), 0, 0, context);
+    }
+
+    private static void updateRegulatingControl(ShuntCompensator shuntCompensator, boolean controlEnabled, Context context) {
+        if (isDefaultRegulatingControl(shuntCompensator, controlEnabled)) {
+            setDefaultRegulatingControl(shuntCompensator);
```
olperr1, line 113: Can you add a comment inspired by the previous comment?

> // The equipment is participating in regulating control, but no regulating control information has been found.
> // We create default regulation data.

olperr1 (changes requested):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java
marqueslanauja: Fixed

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/ShuntCompensatorUpdateTest.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/RegulatingControlMappingForShuntCompensators.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java
marqueslanauja: In my opinion, it is more reasonable to fix the `sectionCount` during the update process when it is outside the limits, rather than increasing the limits every time the `sectionCount` is out of bounds. Additionally, I believe that `maxSectionCount `cannot be modified once the equipment has been created

rcourtier: I agree with @marqueslanauja on this one. It's generally best to import min and max from EQ and adapt SSH/SV values imports in case they are out of bound rather than expanding the limits.

rcourtier (approved):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java
olperr1: OK

rcourtier (approved):

PR #3280, opened 2025-01-21
Fix remove voltage level

flo-dup:

No

Bug fix

Remove voltage level network modification fails for some voltage levels which:
- contain loops
- are connected via a tie line.

It doesn't fail

- [ ] Yes
- [x] No


flo-dup:
The first commit adds a unit test where a loop is introduced in the voltage level during the voltage level removal: before this PR, the busbar sections were removed first when removing the voltage level. But the whole logic of `RemoveFeederBay` relies on busbar sections. Once they are removed the traversing done when removing a feeder bay goes through the whole voltage level, which might lead to some loops. There were no special care for loops in `RemoveFeederBay`, hence we tried to remove twice the same switch because of that.

This didn't occur before last 6.6.0 release because of the bug fixed by #2341: the edges which closed a loop were never traversed before.

flo-dup:
The third commit enriches the unit test to add a loop with the feeder bay (two parallel breakers). In real data, this may occur for a phase-shift transformer in parallel of a breaker, used to shunt it.

Such loops require an extra check within RemoveFeederBay.

olperr1 (approved):

PR #3286, opened 2025-01-22
[UCTE] Refactor generator export and allow export of multiple generators

colineplqt:

No

Feature and refactoring.

If a bus of a network has more than one generator, then the network cannot be exported to UCTE.

The network can be exported to UCTE and the UCTE nodes are created with:
 - Node type code: 2 (PU node) if one generator regulating voltage connected to the bus, 3 if the bus is the slack node and 0 (PQ node) by default.
- Voltage reference: `TargetV` of the generators connected to the bus. We take the closest to `nominalV` is there are more than one generator regulating with different `TargetV`.
- Active power generation: sum of `TargetP`.
- Reactive power generation: sum of `TargetQ`.
- Minimum permissible generation in active power: sum of `minP`.
- Maximum permissible generation in active power: sum of `maxP`.
- Minimum permissible generation in reactive power: sum of `minQ` of the reactive limits of each generator at their `TargetP`.
- Maximum permissible generation in reactive power: sum of `maxQ` of the reactive limits of each generator at their `TargetP`. 
- Power plant type: if all generators have the same type, then this type, otherwise, `F`.

- [ ] Yes
- [X] No


ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/export/elements/GeneratorUcteExport.java
```diff
@@ -0,0 +1,125 @@
...
+                activePowerGeneration += generator.getTargetP();
+            }
+            if (!Double.isNaN(generator.getTargetQ())) {
+                reactivePowerGeneration += generator.getTargetQ();
+            }
+            if (generator.isVoltageRegulatorOn()) {
+                // If one of the generators regulates voltage, then the node is a PU node.
+                nodeType = UcteNodeTypeCode.PU;
+                if (!Double.isNaN(generator.getTargetV())) {
+                    voltageReferences.add(getTargetV(generator));
```
colinepiloquet, line 54: There was no check before that the generator is regulating. But I don't think it makes sense to fill the value if there is no regulation.

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/export/elements/GeneratorUcteExport.java
```diff
@@ -0,0 +1,125 @@
...
+        List<UctePowerPlantType> powerPlantTypes = new ArrayList<>();
+
+        for (Generator generator : bus.getGenerators()) {
+            if (!Double.isNaN(generator.getTargetP())) {
+                activePowerGeneration += generator.getTargetP();
+            }
+            if (!Double.isNaN(generator.getTargetQ())) {
+                reactivePowerGeneration += generator.getTargetQ();
+            }
+            if (!Double.isNaN(generator.getTargetV())) {
```
colinepiloquet, line 50: I am not sure about this. I think there should also be the condition that the generator is regulating, but then it breaks the import that considers only the node type for the regulation.

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/export/elements/GeneratorUcteExport.java
```diff
@@ -0,0 +1,125 @@
...
+        // Default value if all limits are undefined is Double.NaN because these fields are optional in UCTE
+        // TODO: remove the DEFAULT_POWER_LIMIT in UCTEImporter
+        Double minP = minPs.stream().filter(minp -> minp != -Double.MAX_VALUE && minp != -DEFAULT_POWER_LIMIT).reduce(Double::sum).orElse(Double.NaN);
+        Double maxP = maxPs.stream().filter(maxp -> maxp != Double.MAX_VALUE && maxp != DEFAULT_POWER_LIMIT).reduce(Double::sum).orElse(Double.NaN);
+        Double minQ = minQs.stream().filter(minq -> minq != -Double.MAX_VALUE && minq != -DEFAULT_POWER_LIMIT).reduce(Double::sum).orElse(Double.NaN);
+        Double maxQ = maxQs.stream().filter(maxq -> maxq != Double.MAX_VALUE && maxq != DEFAULT_POWER_LIMIT).reduce(Double::sum).orElse(Double.NaN);
+        ucteNode.setMinimumPermissibleActivePowerGeneration(-minP);
+        ucteNode.setMaximumPermissibleActivePowerGeneration(-maxP);
+        ucteNode.setMinimumPermissibleReactivePowerGeneration(-minQ);
+        ucteNode.setMaximumPermissibleReactivePowerGeneration(-maxQ);
```
rcourtier, line 80: I'm not sure about that. On the one hand, you can aggregate generators but on the other hand you discard their limits if they are unknown. You could end up in a situation where you have an aggregated target greater than the limits.

I would suggest to:
- have a calculated (sum) min/max if all the values are known.
- have the default power limit values if at least one is unknown.

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/export/elements/GeneratorUcteExport.java
```diff
@@ -0,0 +1,125 @@
...
+            if (!Double.isNaN(generator.getTargetP())) {
+                activePowerGeneration += generator.getTargetP();
+            }
+            if (!Double.isNaN(generator.getTargetQ())) {
+                reactivePowerGeneration += generator.getTargetQ();
+            }
+            if (!Double.isNaN(generator.getTargetV())) {
+                voltageReferences.add(getTargetV(generator));
+            }
+            if (generator.isVoltageRegulatorOn()) {
```
rcourtier, line 53: I'm wondering if you shouldn't also checked that the generator is regulating locally, otherwise this node is not really PU (it is the remote node):
```suggestion
            if (generator.isVoltageRegulatorOn()
            && generator.getRegulatingTerminal().getConnectable().getId() == generator.getId()) {
```

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/export/elements/GeneratorUcteExport.java
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * Initialize the power generation fields from the generators connected to the specified bus.
+     *
+     * @param ucteNode The UCTE node to fill
+     * @param bus The bus the generators are connected to
+     */
+    public static void convertGenerators(UcteNode ucteNode, Bus bus) {
+        double activePowerGeneration = 0;
+        double reactivePowerGeneration = 0;
+        List<Double> voltageReferences = new ArrayList<>();
```
rcourtier, line 34: You should maybe make the disctinction between generators that regulate remotely and those that regulate locally.
In `ucteNode.setVoltageReference()` you should give precedence to locally regulating generators over those regulating remotely.

rcourtier (changes requested):

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/export/elements/GeneratorUcteExport.java
colinepiloquet: You are right, but then do you agree that I need to set the remote note to PU? Even though there is no Generator?

colinepiloquet: At the same time, as we set the targetV at the bus where the generator is, I don't know if the node should not be PU

rcourtier: All things considered I think it's correct the way you have done. Considering the generators regulate locally is a simplification but it is consistent with UCTE format.

ucte/ucte-converter/src/main/java/com/powsybl/ucte/converter/export/elements/GeneratorUcteExport.java
rcourtier: If we compare with CGMES or IIDM, it's ok to have a targetV even though the regulation is off.
However, you raise a good point as the main challenge here is to select the correct data from the good generator in case of multiple generators connected to the node.
I believe you should proceed similarly as what you did for chosing preferably a local rather than a remote, and chosing preferably a generator regulating rather than not regulating.
So the selection order would be:
- regulating local generator
- regulating remote generator
- not-regulating local generator
- not-regulating remote generator

ucte/ucte-converter/src/test/java/com/powsybl/ucte/converter/UcteExporterGeneratorTest.java
```diff
@@ -0,0 +1,126 @@
...
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+
+class UcteExporterGeneratorTest {
+
+    Network network = EurostagTutorialExample1Factory.create();
+
+    @Test
+    void testGenerator() throws IOException {
+        Properties p = new Properties();
+        p.put(UcteExporter.NAMING_STRATEGY, "Counter");
```
rcourtier, line 21: It seems you are using this property in every test. You could create a setUp() method with the annotation @BeforeAll containing these two lines.
That could also include the network initialization.

rcourtier (changes requested):

So-Fras:
No plan to reuse this PR as such. However, there is still an interest for a full UCTE export so this PR could be used as inspiration for the remaining work to tackle.

PR #3298, opened 2025-01-30
Add backward compatibility for CgmesControlAreas extension deserialization 

olperr1:

No

Feature

- [ ] Yes
- [X] No

#3149 removes the `CgmesControlAreas` in favor of the usage of IIDM `Area`s, but the backward compatibility with the extension is not supported. Therefore when importing an old generated IIDM network containing the extension, the content of this extension is not loaded and no CGMES areas are created.

With this PR, the serialized extension is read and IIDM `Area`s are created for each `CgmesControlArea` of the extension.

Note that when reading a CgmesControlArea from the serialized extension, if an `Area` with the same ID is already present on the network, the CgmesControlArea is skipped.

- [ ] Yes
- [X] No

**Other information**:

For now, only the reading compatibility is supported, because it is not trivial to implement the writing compatibility: there's no extension corresponding to the CGMES areas on the network, so the extension serialization mechanism could not be used directly


rcourtier (approved):

zamarrenolm:
I have tested the backward compatibility using a _real world_ test case and it works perfectly.

zamarrenolm (approved):

cgmes/cgmes-extensions/src/main/java/com/powsybl/cgmes/extensions/compatibility/CgmesControlAreasSerDe.java
```diff
@@ -0,0 +1,187 @@
...
+    @Override
+    public Map<String, String> getArrayNameToSingleNameMap() {
+        return Map.of(CONTROL_AREA_ARRAY_ELEMENT, CONTROL_AREA_ROOT_ELEMENT,
+                TERMINAL_ARRAY_ELEMENT, TERMINAL_ROOT_ELEMENT,
+                BOUNDARY_ARRAY_ELEMENT, BOUNDARY_ROOT_ELEMENT);
+    }
+
+    @Override
+    public void write(DummyExt extension, SerializerContext context) {
+        throw new UnsupportedOperationException("Backward writing compatibility is not yet supported.");
```
flo-dup, line 92: `yet` means we'll do it one day, will we? Besides shouldn't this rather be something like "cannot happen"? We're here with a DummyExt which is never created, and cannot be created.

cgmes/cgmes-extensions/src/main/java/com/powsybl/cgmes/extensions/compatibility/CgmesControlAreasSerDe.java
```diff
@@ -0,0 +1,187 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.OptionalDouble;
+
+/**
+ * @author Olivier Perrin {@literal <olivier.perrin at rte-france.com>}
+ */
+@AutoService(ExtensionSerDe.class)
+public class CgmesControlAreasSerDe extends AbstractVersionableNetworkExtensionSerDe<Network, CgmesControlAreasSerDe.DummyExt> {
```
flo-dup, line 39: why not using `AbstractExtensionSerDe` instead of `AbstractVersionableNetworkExtensionSerDe`. I don't get why we need the versionable part as there's only one the `1.0` version.

flo-dup (changes requested):

cgmes/cgmes-extensions/src/main/java/com/powsybl/cgmes/extensions/compatibility/CgmesControlAreasSerDe.java
olperr1: Yes you're right, there's no need for the versionable management.
I will replace it.

olperr1: Hum no, sorry. I wanted to limit the extension to the IIDM versions <= 1.13.
The extension should not be accepted for the newer versions of IIDM.

flo-dup: Hmm, there's no harm in accepting it for versions > 1.13, even though it will never occur, as there won't be any way to get a 1.14 xiidm with such an extension.

flo-dup (approved):

PR #3301, opened 2025-01-30
ReportNode: change timestamp feature

flo-dup:

No

Remove unused feature

timestamp may be added automatically on all report nodes if specified at root node construction.

timestamp may be added on any report node manually.

- [x] Yes
- [ ] No

Timestamps can not be automatically added on a whole `ReportNode` tree.
If timestamp is needed on a specific `ReportNode`, the user has to add it explicitly with a call to `withTimestamp()` or `withTimestamp(pattern)`, and to refer to it in the message template if it should be displayed to the end user.

The `TimestampPattern` and the `Locale` used are specified at the construction of the root `ReportNode`, but separately. Hence, you need to replace the use of `withTimestampPattern(String pattern, Locale locale)` with `withDefaultTimestampPattern(String pattern).withLocale(Locale locale)`.
Finally, you should replace the use of `withTimestampPattern(String pattern)` with `withDefaultTimestampPattern(String pattern)`.

### DIFF
--- a/commons/src/main/java/com/powsybl/commons/report/AbstractReportNodeAdderOrBuilder.java
+++ b/commons/src/main/java/com/powsybl/commons/report/AbstractReportNodeAdderOrBuilder.java
@@ -9,6 +9,7 @@ package com.powsybl.commons.report;
 
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Objects;
 
 /**
  * @author Florian Dupuy {@literal <florian.dupuy at rte-france.com>}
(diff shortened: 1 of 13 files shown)


flo-dup:
After some discussion with @geofjamg and @alicecaron, it seems handy to keep a reduced timestamp feature, something like a `withTimeStamp()` method inside the adder/builder.

olperr1 (changes requested):

olperr1 (approved):

PR #3302, opened 2025-01-30
Migrate all CGMES reports to the CgmesReports dedicated class

alicecaron:

Clean code to have all the reports in the cgmes-conversion module are moved in the CgmesReports class.

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesReports.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesReports.java
@@ -17,6 +17,11 @@ import com.powsybl.iidm.network.Substation;
  */
 public final class CgmesReports {
 
+    public static final String REPORT_NODE_KEY_EXPORTED_CGMES_ID = "ExportedCgmesId";
+    public static final String REPORT_VALUE_EXPORTED_CGMES_ID = "cgmesId";
+    public static final String REPORT_VALUE_EXPORTED_CGMES_SUBSET = "cgmesSubset";
+    public static final String REPORT_VALUE_EXPORTED_CGMES_NETWORK_ID = "networkId";
+
     private CgmesReports() {
     }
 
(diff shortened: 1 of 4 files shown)


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesReports.java
```diff
@@ -17,6 +17,11 @@
  */
 public final class CgmesReports {
 
+    public static final String REPORT_NODE_KEY_EXPORTED_CGMES_ID = "ExportedCgmesId";
+    public static final String REPORT_VALUE_EXPORTED_CGMES_ID = "cgmesId";
+    public static final String REPORT_VALUE_EXPORTED_CGMES_SUBSET = "cgmesSubset";
+    public static final String REPORT_VALUE_EXPORTED_CGMES_NETWORK_ID = "networkId";
```
flo-dup, line 23: Could you replace these keys in the reports instead of using a once used constant?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesReports.java
```diff
@@ -194,4 +199,25 @@ public static void missingMandatoryAttributeReport(ReportNode reportNode, String
                 .withSeverity(TypedValue.ERROR_SEVERITY)
                 .add();
     }
+
+    public static ReportNode importingCgmesFileReport(ReportNode reportNode) {
+        return reportNode.newReportNode()
+                .withMessageTemplate("CGMESConversion", "Importing CGMES file(s)")
```
flo-dup, line 205: A bit outside the move scope, but while we're at it could you add as values the extension and the basename in the template?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesReports.java
flo-dup: oh, didn't see that. I was suggesting this for i18n matters: if the key is defined through a constant you don't have a direct access to its value, so it's a bit more difficult to understand the report.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesReports.java
flo-dup: in fact, I think that `getDataExtension()` will always be null for CGMES files, so don't add it to the report. Sorry for asking this! :slightly_frowning_face:

commons/src/main/java/com/powsybl/commons/report/TypedValue.java
```diff
@@ -54,9 +54,9 @@ public class TypedValue {
      * @param type a string representing the value type (see the public constants for some generic types)
      */
     public TypedValue(Object value, String type) {
-        this.value = Objects.requireNonNull(value);
+        this.value = null == value ? "" : value;
```
flo-dup, line 57: I think we should keep in mind to make it robust to null String, but in this PR it doesn't make sense anymore... sorry again!

flo-dup (changes requested):

commons/src/main/java/com/powsybl/commons/report/TypedValue.java
alicecaron: No problem! I removed it from the PR

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesReports.java
alicecaron: So I can remove the use of these constants and write the key directly in the tests class to stay homogenous with the other usages of ReportNode keys. What do you think?

flo-dup (approved):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesReports.java
alicecaron: I removed the constants!

flo-dup (changes requested):
didn't see, there's one code smell introduced by your change, could you fix it?

flo-dup (approved):
ok to not introduce the constant, it's helping checking the template correctness

PR #3303, opened 2025-01-30
Add copy to ReportNode API

flo-dup:

No

Feature

A `ReportNode` can only insert a root `ReportNode`

A `ReportNode` can insert any `ReportNode` by calling `reportNode.copy(otherReportNode)`

- [ ] Yes
- [x] No

**Other information**:
The implementation of copy is done with serialization / deserialization. This gives a very simple version of copying, which besides does not need to check for an infinite loop. The drawback is that the inherited values of the report node to copy are lost.

### DIFF
--- a/commons/src/main/java/com/powsybl/commons/report/ReportNodeImpl.java
+++ b/commons/src/main/java/com/powsybl/commons/report/ReportNodeImpl.java
@@ -15,10 +15,14 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.powsybl.commons.PowsyblException;
 import com.powsybl.commons.ref.RefChain;
 import com.powsybl.commons.ref.RefObj;
+import org.apache.commons.io.IOUtils;
 import org.apache.commons.text.StringSubstitutor;
 
 import java.io.IOException;
+import java.io.StringWriter;
+import java.io.UncheckedIOException;
 import java.io.Writer;
+import java.nio.charset.StandardCharsets;
 import java.time.format.DateTimeFormatter;
 import java.util.*;
 import java.util.stream.Stream;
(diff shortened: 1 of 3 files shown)


commons/src/main/java/com/powsybl/commons/report/ReportNodeImpl.java
```diff
@@ -169,6 +173,21 @@ public void include(ReportNode reportNode) {
         reportNodeImpl.treeContext.setRef(treeContext);
     }
 
+    @Override
+    public void copy(ReportNode reportNode) {
```
alicecaron, line 177: Maybe add a unit test?

flo-dup: indeed! that's why I added [WIP] in fact :slightly_smiling_face:

commons/src/main/java/com/powsybl/commons/report/ReportNodeImpl.java
```diff
@@ -169,6 +173,24 @@ public void include(ReportNode reportNode) {
         reportNodeImpl.treeContext.setRef(treeContext);
     }
 
+    @Override
+    public void addCopy(ReportNode reportNode) {
+        var om = new ObjectMapper().registerModule(new ReportNodeJsonModule());
```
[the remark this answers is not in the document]
flo-dup: No, we don't need to check it, there's no specific implementation constraint.
And the copied reportNode can be a root report node, there's no reason to forbid it, is there?

commons/src/test/java/com/powsybl/commons/report/ReportNodeTest.java
```diff
@@ -134,30 +134,34 @@ void testInclude() throws IOException {
...
                 .withMessageTemplate("includedRoot", "Included root message")
                 .build();
-        otherRoot.newReportNode()
+        ReportNode otherRootChild = otherRoot.newReportNode()
                 .withMessageTemplate("includedChild", "Included child message")
                 .add();
 
+        PowsyblException e1 = assertThrows(PowsyblException.class, () -> root.include(ReportNode.NO_OP));
+        PowsyblException e2 = assertThrows(PowsyblException.class, () -> otherRootChild.include(otherRoot));
+        PowsyblException e3 = assertThrows(PowsyblException.class, () -> otherRootChild.include(otherRoot));
```
olperr1, line 146: What is the purpose of testing it twice in a row?
If this test is really needed, can you add a comment explaining why?

olperr1 (changes requested):

commons/src/test/java/com/powsybl/commons/report/ReportNodeTest.java
flo-dup: My mistake! the two lines were different before eb90bdf commit. Fixed

olperr1 (approved):

PR #3304, opened 2025-02-03
Fix throwException support when applying NetworkModifications

### LINKED ISSUE #3204: Fix the handle of not applicable NetworkModification when throwException boolean is present
In the `NetworkModification::apply` methods that take the `throwException` boolean in input parameter the use of `AbstractNetworkModification::logOrThrow` method should be called when the network modification cannot be applied.

This method allows to either log the not applicable reason as a warning (if `throwException=false`) or to throw a `PowsyblException` (if `throwException=true`).

**Problem:**
Some of the `NetworkModification::apply` method implementations do not call this method when the modification can not be applied.

**Solution:**
Fix each of thoses implementations to use the `logOrThrow` method.

**Example:**
In `ConnectGenerator::apply` we throw a `PowsyblException` in some case, but we should `logOrThrow` instead (and in case it is a log we should then return):

Replace
```java
throw new PowsyblException("Generator '" + generatorId + "' not found");
```
With
```java
logOrThrow(throwException, "Generator '" + generatorId + "' not found");
return;
```

cphili:

Fixes #3204

Bug fix

- [ ] Yes
- [X] No

Some `NetworkModification` are throwing an exception when an error is encountered during their application (method `apply(...)`) even with `throwException` set to `false`.

The `throwException` boolean is correctly taken into account by the `NetworkModification.apply(...)` methods.

- [X] Yes
- [ ] No

This is not really a breaking change, but the behavior of some `NetworkModification`s was changed by this PR. They previously threw an Exception when a problem was encountered in their `apply(...)` method, even with `throwException` set to `false`. Now they only throw an Exception when `throwException` is set to `true`. These network modifications are:
- every modification extending `AbstractTripping` (i.e. network modification classes named `XxxTripping`)
- `CloseSwitch`
- `ConnectGenerator`
- `OpenSwitch`
- `PhaseShifterOptimizeTap`
- `PhaseShifterSetAsFixedTap`
- `PhaseShifterShiftTap`
- `ShuntCompensatorModification`

**Other information**:


iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/LoadFlowBasedPhaseShifterOptimizer.java
```diff
@@ -41,16 +42,13 @@ public LoadFlowBasedPhaseShifterOptimizer(ComputationManager computationManager)
...
-        try {
-            String loadFlowName = config.getLoadFlowName().orElse(null);
-            LoadFlowResult result = LoadFlow.find(loadFlowName)
-                                            .run(network, workingStateId, computationManager, LoadFlowParameters.load());
-            if (!result.isOk()) {
-                throw new PowsyblException("Load flow diverged during phase shifter optimization");
-            }
-        } catch (Exception e) {
-            throw new PowsyblException(e);
+    private void runLoadFlow(Network network, String workingStateId, boolean throwException) {
```
olperr1, line 45: I don't think this class should be modified.
You can just add a try-catch in [`PhaseShifterOptimizeTap.apply()`](https://github.com/powsybl/powsybl-core/blob/1e0f8f569cab34a86ab13e868ea7643f3c0e2b1c/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/PhaseShifterOptimizeTap.java#L34) around the usage of `LoadFlowBasedPhaseShifterOptimizer` and then call `logOrThrow(...)` in the catch clause.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/PhaseShifterOptimizer.java
```diff
@@ -14,6 +14,6 @@
  */
 public interface PhaseShifterOptimizer {
 
-    void findMaximalFlowTap(Network network, String phaseShifterId);
+    void findMaximalFlowTap(Network network, String phaseShifterId, boolean throwException);
```
olperr1, line 17: This change should be reverted (see previous comment).

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/ReplaceTeePointByVoltageLevelOnLine.java
```diff
@@ -137,84 +137,82 @@ public void apply(Network network, NamingStrategy namingStrategy, boolean throwE
...
-            } else {
-                return;
-            }
-        }
 
-        // tapped voltage level is the voltage level of tpLineToRemove, at the opposite side of the tee point
-        VoltageLevel tappedVoltageLevel = tpLineToRemove.getTerminal1().getVoltageLevel() == teePoint
-                ? tpLineToRemove.getTerminal2().getVoltageLevel()
-                : tpLineToRemove.getTerminal1().getVoltageLevel();
+        if (teePoint != null) {
```
olperr1, line 141: In this method, it would be more readable with the following test at the beginning:

```java
if (teePoint == null) {
    noTeePointAndOrTappedVoltageLevelReport(reportNode, oldLine1Id, oldLine2Id, lineToRemoveId);
    ModificationLogs.logOrThrow(throwException, String.format("Unable to find the tee point and the tapped voltage level from lines %s, %s and %s", oldLine1Id, oldLine2Id, lineToRemoveId));
    return;
}
```
(Which corresponds to restore the original order of the nullability test of `teePoint`.)

olperr1 (changes requested):
@cphili Some units tests are failing and should be fixed.
We will fix the DCO check later. Please just don't merge `main` in this branch for now.

iidm/iidm-modification/src/test/java/com/powsybl/iidm/modification/VscConverterStationModificationTest.java
```diff
@@ -54,8 +54,7 @@ void testApplyChecks() {
             2.);
         assertThrows(PowsyblException.class, () -> modif1.apply(network, true, ReportNode.NO_OP),
             "An invalid ID should fail to apply.");
-        assertDoesNotThrow(() -> modif1.apply(network, false, ReportNode.NO_OP),
-            "An invalid ID should not throw if throwException is false.");
+        assertDoesNotThrow(() -> modif1.apply(network, false, ReportNode.NO_OP));
```
olperr1, line 57: Can you revert this change?
The message is legit here: it corresponds to the behavior when the assertion fails.

iidm/iidm-modification/src/test/java/com/powsybl/iidm/modification/TestThreeWindingsTransformerModification.java
```diff
@@ -62,8 +62,7 @@ void testApplyChecks() {
         ThreeWindingsTransformerModification t3wtModification = new ThreeWindingsTransformerModification("UNKNOWN_ID", Double.NaN);
         assertThrows(PowsyblException.class, () -> t3wtModification.apply(network, true, ReportNode.NO_OP),
                 "An invalid ID should fail to apply.");
-        assertDoesNotThrow(() -> t3wtModification.apply(network, false, ReportNode.NO_OP),
-                "An invalid ID should not throw if throwException is false.");
+        assertDoesNotThrow(() -> t3wtModification.apply(network, false, ReportNode.NO_OP));
```
olperr1, line 65: Can you revert here to (same thing as the previous comment)?

iidm/iidm-modification/src/test/java/com/powsybl/iidm/modification/topology/CreateVoltageLevelTopologyTest.java
```diff
@@ -90,6 +91,7 @@ void testWithUnsupportedSwitchKind() {
                 .withSectionCount(4)
                 .withSwitchKinds(SwitchKind.BREAKER, SwitchKind.LOAD_BREAK_SWITCH, SwitchKind.DISCONNECTOR)
                 .build();
+        assertDoesNotThrow(() -> modification.apply(network, false, reportNode));
```
olperr1, line 94: You missed this occurrence (reportNode -> ReportNode.NO_OP)

iidm/iidm-modification/src/test/java/com/powsybl/iidm/modification/tripping/BranchTrippingTest.java
```diff
@@ -47,12 +50,14 @@ void lineTrippingTest() {
         assertFalse(line.getTerminal2().isConnected());
 
         LineTripping unknownLineTripping = new LineTripping("NOT_EXISTS");
-        Exception e1 = assertThrows(PowsyblException.class, () -> unknownLineTripping.apply(network));
+        Exception e1 = assertThrows(PowsyblException.class, () -> unknownLineTripping.apply(network, true, ReportNode.NO_OP));
         assertEquals("Line 'NOT_EXISTS' not found", e1.getMessage());
+        assertDoesNotThrow(() -> unknownLineTripping.apply(network));
```
olperr1, line 55: We should add a section in the PR description to list all the NetworkModifications that threw an Exception in all cases and that now only throw one when `throwException` is true.

olperr1 (changes requested):

olperr1:
Can you also take a look at the sonar issues? Some can be easily fixed.

olperr1 (changes requested):

iidm/iidm-modification/src/test/java/com/powsybl/iidm/modification/topology/ReplaceTeePointByVoltageLevelOnLineTest.java
```diff
@@ -125,6 +125,25 @@ void replaceTeePointByVoltageLevelOnLineNbTest() throws IOException {
...
+        modification.apply(network);
+
+        // create additional line to test bad configuration
+        VoltageLevel vl = network.newVoltageLevel().setId("VL3").setNominalV(380).setTopologyKind(TopologyKind.NODE_BREAKER).add();
+        vl.getNodeBreakerView().newBusbarSection().setId("bbs3").setNode(0).add();
+        vl.getNodeBreakerView().newSwitch().setId("breaker3").setName("breaker3").setKind(SwitchKind.BREAKER).setRetained(false).setOpen(true).setFictitious(false).setNode1(0).setNode2(1).add();
+        vl = network.newVoltageLevel().setId("VL4").setNominalV(380).setTopologyKind(TopologyKind.NODE_BREAKER).add();
+        vl.getNodeBreakerView().newBusbarSection().setId("bbs4").setNode(0).add();
+        vl.getNodeBreakerView().newSwitch().setId("breaker4").setName("breaker4").setKind(SwitchKind.BREAKER).setRetained(false).setOpen(true).setFictitious(false).setNode1(0).setNode2(1).add();
+        network.newLine().setId("LINE34").setR(0.1).setX(0.1).setG1(0.0).setB1(0.0).setG2(0.0).setB2(0.0).setNode1(1).setVoltageLevel1("VL3").setNode2(1).setVoltageLevel2("VL4").add();
```
olperr1, line 146: This code is the same as in the previous method. Can you introduce a method to share instead of duplicate it?

olperr1 (changes requested):

olperr1 (approved):

Issue #3259, opened 2025-01-07
CGMES loading from zipped profiles inside a folder


geofjamg:
### Describe the current behavior

We can load a CGMES either given a zipped file containing all profiles or given a folder containing profiles using 
Ǹetwork.read(zipOrFolder)`

### Describe the expected behavior

As CGMES data is often exchanged as individual zipped profiles it would a nice to have to be able to load a CGMES given a folder containing all zipped profiles.

PR #3309, opened 2025-02-05
CGMES loading from zipped profiles inside a folder

quinarygio:

#3259

- [ ] Yes
- [x] No

**Other information**:


cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesOnDataSource.java
```diff
@@ -109,27 +123,59 @@ public Set<String> names() {
...
+                return ns.contains(RDF_NAMESPACE) && ns.stream().anyMatch(CgmesNamespace::isValid);
+            } catch (XMLStreamException e) {
+                return false;
+            } catch (IOException x) {
+                throw new CgmesModelException(String.format(LISTING_CGMES_NAMES_IN_DATA_SOURCE, dataSource), x);
+            }
+        }
         try (InputStream is = dataSource.newInputStream(name)) {
             Set<String> ns = NamespaceReader.namespaces1(is);
             return ns.contains(RDF_NAMESPACE) && ns.stream().anyMatch(CgmesNamespace::isValid);
```
alicecaron, line 157: You can mutualize some code by loading the `InputStream` whether it is from an xml or zip file by creating a private method doing all the work and not closing the input stream.  Keeping the try/catch outside the private method allows to then put the logic as it is. the private method can be called for the 3 times you need the `InputStream` (thank you fot the tip @olperr1 !)  

```suggestion
try (InputStream is = loadInputStream(n)) {
    return NamespaceReader.base(is);
} catch (IOException x) {
    throw new UncheckedIOException(x);
}
```
Example for the private method to load the InputStream:
```
private InputStream loadInputStream(String n) throws IOException {
        InputStream is;
        String fileExtension = n.substring(n.lastIndexOf('.') + 1);
        if (fileExtension.equals(CompressionFormat.ZIP.getExtension())) {
            ZipSecurityHelper.checkIfZipExtractionIsSafe(dataSource, n);
            is = new ZipInputStream(dataSource.newInputStream(n));
            ((ZipInputStream) is).getNextEntry();
        } else {
            is = dataSource.newInputStream(n);
        }
        return is;
    }
```

quinarygio: In fact this was my first implementation (see my first commit), but then Sonar complained because the InputStream was not closed, so I had to duplicate some code...

commons/src/test/java/com/powsybl/commons/compress/ZipSecurityHelperTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+import com.google.common.jimfs.Configuration;
+import com.google.common.jimfs.Jimfs;
+import com.powsybl.commons.datasource.DirectoryDataSource;
+import com.powsybl.commons.datasource.ReadOnlyDataSource;
+
+/**
+ * @author Giovanni Ferrari {@literal <giovanni.ferrari at soft.it>}
+ */
+public class ZipSecurityHelperTest {
```
flo-dup, line 33: Could you remove the public access modifiers in this file to avoid the 7 sonar code smells?

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesOnDataSource.java
```diff
@@ -109,27 +123,59 @@ public Set<String> names() {
...
     }
 
     public Set<String> namespaces() {
         Set<String> ns = new HashSet<>();
         names().forEach(n -> {
+            String fileExtension = n.substring(n.lastIndexOf('.') + 1);
+            if (fileExtension.equals(CompressionFormat.ZIP.getExtension())) {
+                ZipSecurityHelper.checkIfZipExtractionIsSafe(dataSource, n);
+                try (ZipInputStream is = new ZipInputStream(dataSource.newInputStream(n))) {
+                    is.getNextEntry();
```
flo-dup, line 172: we take the first entry only, shouldn't we check there's only one entry in the zip file? and throw a clear exception otherwise.

flo-dup (changes requested):

alicecaron (approved):

commons/src/main/java/com/powsybl/commons/compress/ZipSecurityHelper.java
```diff
@@ -0,0 +1,84 @@
...
+
+            if (totalEntryArchive > thresholdEntries) {
+                // too many entries in this archive, can lead to inodes exhaustion of the system
+                return false;
+            }
+            ZipEntry currentEntry = ze;
+            ze = zipInputStream.getNextEntry();
+            long entrySize = currentEntry.getCompressedSize();
+            double compressionRatio = (double) totalSizeEntry / (double) entrySize;
+            if (compressionRatio > thresholdCompressionRatio) {
```
alicecaron, line 76: I tried with real files, and for a EQ xml zipped the compressed ratio is not sufficient for the current proposed `THRESHOLD_RATIO=10`. I had a `compressionRatio` of around `11`. In consequence a `java.io.UncheckedIOException` error was thrown: `Zip file extraction is not safe`  
Maybe it is necessar to increase this threshold because of the size of CGMES EQ files and poor ratio? @flo-dup @olperr1

alicecaron: It is ok with `THRESHOLD_RATIO = 12` in my case.

olperr1: 12 won't be sufficient.

I checked some of our test files and found some with a bigger ratio when zipped with the maximum compression level:

```shell
path='cgmes/cgmes-conformity/src/main/resources/cgmes3-test-models/Svedala'
cgmes_file="$path/20201202T1843Z_1D_Svedala Area_DL_001.xml"
cgmes_size="$(ls -l "$cgmes_file"  | awk '{print $5}')"
rm /tmp/test.zip
zip -9 /tmp/test.zip "$cgmes_file"
zip_size="$(ls -l /tmp/test.zip  | awk '{print $5}')"
bc -l <<< "$cgmes_size / $zip_size"
```
=> 12.11357674909463524857

We can maybe use 20 to have a security margin.

olperr1 (changes requested):

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesOnDataSource.java
```diff
@@ -109,34 +107,51 @@ public Set<String> names() {
...
-        names().forEach(n -> {
-            try (InputStream is = dataSource.newInputStream(n)) {
-                ns.addAll(NamespaceReader.namespaces(is));
-            } catch (IOException x) {
-                throw new UncheckedIOException(x);
+    private <T> T loadInputStreamAndGetNamespace(String n, Function<InputStream, T> namespaceGetter) {
+        try (InputStream in = dataSource.newInputStream(n)) {
+            String fileExtension = n.substring(n.lastIndexOf('.') + 1);
+            if (fileExtension.equals(CompressionFormat.ZIP.getExtension())) {
+                ZipSecurityHelper.checkIfZipExtractionIsSafe(dataSource, n);
```
flo-dup, line 129: There's a performance issue here. We're reading the whole unzipped file to detect if the zip extraction is safe but we only need the first tag (for namespaces definition or the base attribute). I think we need to do something smarter to get safely the first characters only, not unzipping the complete file.

flo-dup (changes requested):

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesOnDataSource.java
quinarygio: I propose to check just first zip entry safety, checking only the compression ratio

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/LoadZippedProfilesTest.java
```diff
@@ -0,0 +1,80 @@
...
+    void oneFolderAndOneZipPerProfileTest() throws Exception {
+        var testDataSource = CgmesConformity1Catalog.microGridBaseCaseBE().dataSource();
+        Set<String> profiles = testDataSource.listNames(".*");
+        try (FileSystem fileSystem = Jimfs.newFileSystem(Configuration.unix())) {
+            // copy and compress each of the profile to the file system
+            Path workDir = fileSystem.getPath("/work");
+            for (String profile : profiles) {
+                try (var is = testDataSource.newInputStream(profile);
+                        var os = new ZipOutputStream(Files.newOutputStream(workDir.resolve(profile + ".zip")))) {
+                    os.putNextEntry(new ZipEntry(profile));
```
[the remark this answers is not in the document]
quinarygio: The issue is that it is not possible to check the number of entries in a zip file without reading them from the input stream. Additionally, if we attempt to read more than one entry, we cannot go back and read from the first one without reopening the ZipInputStream.
Currently, the exception thrown is a PowsyblException with the message 'Unsupported file format or invalid file', similar to when attempting to read an incompatible file.

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesOnDataSource.java
flo-dup: I think a simpler way is to use a `SafeZipInputStream` class wrapping the `ZipInputStream` instead of the `ZipInputStream` itself:
```java
public class SafeZipInputStream extends ForwardingInputStream<ZipInputStream> {
    @Override
    public int read() throws IOException {
        int byteRead = super.read();
        if (byteRead != -1 && this.bytesRead++ > this.maxBytesToRead) {
          throw new IOException();
        }
        return byteRead;
    }

    @Override
    public int read(byte[] b, int off, int len) throws IOException {
        ... // similar way
    }
}
```
That way you'll only read the first lines, not unzipping the full file, which might be more than 1GB uncompressed. Besides we could restrict maxBytesRead a lot in this PR (a few kB, or 1MB?), as we only read the first tag anyway (we stop at first `START_ELEMENT` in the 3 use cases).

docs/grid_exchange_formats/cgmes/import.md
```diff
@@ -5,6 +5,8 @@ The CIM-CGMES importer reads and converts a CIM-CGMES model to the PowSyBl grid
 - Convert CIM-CGMES data retrieved by SPARQL requests from the created triplestore to PowSyBl grid model
 
 The data in input CIM/XML files uses RDF (Resource Description Framework) syntax. In RDF, data is described making statements about resources using triplet expressions: (subject, predicate, object).
+The CIM-CGMES importer supports also ZIP compressed input files. The importer will decompress the ZIP files to read CIM/XML data.
```
flo-dup, line 8: It's not so clear here, I think we should enumerate the 3 ways (or more?) to import a CGMES file:
- a folder containing all uncompressed profile files
- a folder containing the zipped profile files, each one being in a separate zip file
- a zipped file containing all profile files

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/LoadZippedProfilesTest.java
```diff
@@ -0,0 +1,75 @@
...
+        Set<String> profiles = testDataSource.listNames(".*");
+        try (FileSystem fileSystem = Jimfs.newFileSystem(Configuration.unix())) {
+            // copy and compress each of the profile to the file system
+            Path workDir = fileSystem.getPath("/work");
+            for (String profile : profiles) {
+                try (var is = testDataSource.newInputStream(profile);
+                        var os = new ZipOutputStream(Files.newOutputStream(workDir.resolve(profile + ".zip")))) {
+                    os.closeEntry();
+                }
+            }
```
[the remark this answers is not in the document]
olperr1: In `emptyZipErrorTest()`, there's no need to create an input stream for each profile since they are not read.

quinarygio:
Using the new SafeZipInputStream class to read from the ZipInputStream, the ZipSecurityHelper class is not used anymore. Should we keep it or not ?

commons/src/main/java/com/powsybl/commons/compress/SafeZipInputStream.java
```diff
@@ -0,0 +1,50 @@
...
+import java.io.IOException;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+
+import com.powsybl.commons.io.ForwardingInputStream;
+
+public class SafeZipInputStream extends ForwardingInputStream<ZipInputStream> {
+
+    private int bytesRead;
+    private int maxBytesToRead;
```
flo-dup, line 19: We might want to go further than 2GB, so an int is not enough.
```suggestion
    private long bytesRead;
    private long maxBytesToRead;
```

commons/src/main/java/com/powsybl/commons/compress/SafeZipInputStream.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public SafeZipInputStream(ZipInputStream in, int entryNumber, int maxBytesToRead) throws IOException {
+        super(in);
+        this.maxBytesToRead = maxBytesToRead;
+        for (int i = 0; i < entryNumber; i++) {
+            ZipEntry zipEntry = in.getNextEntry();
```
flo-dup, line 25: We might want to reuse this for other usecases where several ZipInputStream entries have to be read. This is not really easy to use that way, I think you'd need to create a `new SafeZipInputStream(zin, 1, max)` for each entry?

Adding a `SafeZipInputStream::getNextEntry` method would do the trick, at the cost of adding a `protected T getDelegate()` in `ForwardingInputStream`.

flo-dup (changes requested):

commons/src/main/java/com/powsybl/commons/compress/SafeZipInputStream.java
```diff
@@ -15,8 +15,8 @@
 
 public class SafeZipInputStream extends ForwardingInputStream<ZipInputStream> {
 
-    private int bytesRead;
-    private int maxBytesToRead;
+    private long bytesRead;
+    private long maxBytesToRead;
 
     public SafeZipInputStream(ZipInputStream in, int entryNumber, int maxBytesToRead) throws IOException {
```
flo-dup, line 21: You forgot to change maxBytesToRead to a long here. 

Besides it's not clear what entryNumber is here. It's not exactly the number of entries to skip, as when you just created the zipInputStream and give entryNumber = 1 you're not skipping anything. And it's not really something to do with the "safe" responsability of the SafeZipInputStream class. Therefore, could you remove this argument and let the caller do the call(s) to getNextEntry?

flo-dup (approved):

commons/src/main/java/com/powsybl/commons/compress/SafeZipInputStream.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.zip.ZipInputStream;
+
+import com.powsybl.commons.io.ForwardingInputStream;
+
+public class SafeZipInputStream extends ForwardingInputStream<ZipInputStream> {
+
+    private long bytesRead;
+    private long maxBytesToRead;
+
+    public SafeZipInputStream(ZipInputStream in, long maxBytesToRead) throws IOException {
```
flo-dup, line 21: woops didn't see the code smell on sonar! IOException cannot be thrown anymore

commons/src/main/java/com/powsybl/commons/compress/SafeZipInputStream.java
```diff
@@ -0,0 +1,50 @@
...
+import java.io.IOException;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+
+import com.powsybl.commons.io.ForwardingInputStream;
+
+public class SafeZipInputStream extends ForwardingInputStream<ZipInputStream> {
+
+    private long bytesRead;
+    private long maxBytesToRead;
```
flo-dup, line 19: and this field can be marked final

flo-dup (changes requested):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
```diff
@@ -303,8 +306,27 @@ private Set<ReadOnlyDataSource> separateByModelingAuthority() {
 
         private Optional<String> readModelingAuthority(String name) {
             String modellingAuthority = null;
-            try (InputStream is = dataSource.newInputStream(name)) {
-                XMLStreamReader reader = xmlInputFactory.createXMLStreamReader(is);
+            try (InputStream in = dataSource.newInputStream(name)) {
+                String fileExtension = name.substring(name.lastIndexOf('.') + 1);
+                if (fileExtension.equals(CompressionFormat.ZIP.getExtension())) {
+                    try (SafeZipInputStream zis = new SafeZipInputStream(new ZipInputStream(in), 2048)) {
```
flo-dup, line 312: Just saw an xml file with a copyright header! It's probably too drastic to put 2048 as a limit here, and to 1024 in the other call, as you might have some comments in the header (like a copyright). Shouldn't we raise it a bit?

quinarygio: Which value could be reasonable ? A standard copyright should not take more than 1000 bytes, but there could be additional comments...

flo-dup: I agree it should be ok with current parameters, but we should still take some more margin just in case. Plus the values are hardcoded. As the maximum value is only to avoid zip bombing, you could take something like 10kB instead, we're still pretty safe.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
```diff
@@ -309,7 +309,7 @@ private Optional<String> readModelingAuthority(String name) {
             try (InputStream in = dataSource.newInputStream(name)) {
                 String fileExtension = name.substring(name.lastIndexOf('.') + 1);
                 if (fileExtension.equals(CompressionFormat.ZIP.getExtension())) {
-                    try (SafeZipInputStream zis = new SafeZipInputStream(new ZipInputStream(in), 2048)) {
+                    try (SafeZipInputStream zis = new SafeZipInputStream(new ZipInputStream(in), 1024000L)) {
```
alicecaron, line 312: is it ok to pass from `2048` to `1024000L` and not `2048000L`?

flo-dup: yest I think it is :slightly_smiling_face: There's no reason to have a different margin for both use cases.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
flo-dup: The overall limit is fine, and in fact that's safer.

Besides right now it's only used for a single entry. If we read several entries in another usecase, the overall  limit would be raise in that use case. And at that time we would also need to improve the SafeZipInputStream to check the number of entries.

flo-dup (approved):
Approving again! :sweat_smile:

commons/src/main/java/com/powsybl/commons/compress/SafeZipInputStream.java
```diff
@@ -0,0 +1,50 @@
...
+    private long bytesRead;
+    private final long maxBytesToRead;
+
+    public SafeZipInputStream(ZipInputStream in, long maxBytesToRead) {
+        super(in);
+        this.maxBytesToRead = maxBytesToRead;
+    }
+
+    public ZipEntry getNextEntry() throws IOException {
+        return this.getDelegate().getNextEntry();
```
olperr1, line 27: I think we could also limit the number of calls to `getNextEntry()` on the delegate, just in case this class is used later for more complex zips:
- new constructor parameter `int maxEntriesToRead`
- control in `getNextEntry()`: if counter > maxEntriesToRead then throw an IOException.

olperr1 (changes requested):

olperr1 (approved):

PR #3310, opened 2025-02-06
Cgmes import update. Import EQ and SSH separately. DanglingLines.

marqueslanauja:

- [ ] No

- [ ] Yes
- [ ] No

**Other information**:


cgmes/cgmes-model/src/main/resources/CIM16-update.sparql
```diff
@@ -0,0 +1,272 @@
...
+        cim:SvVoltage.angle ?angle .
+}
+
+# query: switches
+SELECT *
+{
+    ?Switch a ?type ;
+        cim:Switch.open ?open .
+    VALUES ?type { cim:Switch cim:Breaker cim:Disconnector cim:LoadBreakSwitch cim:ProtectedSwitch cim:GroundDisconnector }
+}
```
olperr1, line 116: Shouldn't we have it also in `CIM100-update.sparql` to handle the `cim:Jumper` case?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/SwitchConversion.java
```diff
@@ -133,5 +133,16 @@ private void warnDanglingLineCreated() {
...
 
+    public static void update(DanglingLine danglingLine, PropertyBag cgmesData, Context context) {
+        updateTerminals(danglingLine, context, danglingLine.getTerminal());
+        boolean isClosed = !cgmesData.asBoolean("open").orElse(defaultOpen(context));
+        updateTargetsAndRegulationAndOperationalLimits(danglingLine, isBoundaryTerminalConnected(danglingLine, context) && isClosed, context);
+        computeFlowsOnModelSide(danglingLine, context);
+    }
+
+    private static boolean defaultOpen(Context context) {
+        return getDefaultValue(null, false, false, false, context);
```
rcourtier: Don't you want to use `Switch.normalOpen` from CGMES EQ as first argument (equipmentValue) instead of `null`?

cgmes/cgmes-conversion/src/test/resources/update/dangling-line/danglingLine_SSH.xml
```diff
@@ -0,0 +1,86 @@
...
+        <cim:ACDCTerminal.connected>true</cim:ACDCTerminal.connected>
+    </cim:Terminal>
+    <cim:Terminal rdf:about="#_ACLineSegment-EquivalentInjection-T">
+        <cim:ACDCTerminal.connected>true</cim:ACDCTerminal.connected>
+    </cim:Terminal>
+    <cim:EquivalentInjection rdf:about="#_ACLineSegment-EquivalentInjection">
+        <cim:EquivalentInjection.p>284.5</cim:EquivalentInjection.p>
+        <cim:EquivalentInjection.q>70.5</cim:EquivalentInjection.q>
+        <cim:EquivalentInjection.regulationStatus>true</cim:EquivalentInjection.regulationStatus>
+        <cim:EquivalentInjection.regulationTarget>409.0</cim:EquivalentInjection.regulationTarget>
```
rcourtier: You could remove these 2 attributes since they are not mandatory and in the EQ file, this EquivalentInjection's `regulationCapability` flag is set to `false`.

cgmes/cgmes-conversion/src/test/resources/update/dangling-line/danglingLine_SSH_1.xml
```diff
@@ -0,0 +1,86 @@
...
+        <cim:ACDCTerminal.connected>true</cim:ACDCTerminal.connected>
+    </cim:Terminal>
+    <cim:Terminal rdf:about="#_ACLineSegment-EquivalentInjection-T">
+        <cim:ACDCTerminal.connected>true</cim:ACDCTerminal.connected>
+    </cim:Terminal>
+    <cim:EquivalentInjection rdf:about="#_ACLineSegment-EquivalentInjection">
+        <cim:EquivalentInjection.p>280.0</cim:EquivalentInjection.p>
+        <cim:EquivalentInjection.q>70.0</cim:EquivalentInjection.q>
+        <cim:EquivalentInjection.regulationStatus>false</cim:EquivalentInjection.regulationStatus>
+        <cim:EquivalentInjection.regulationTarget>409.0</cim:EquivalentInjection.regulationTarget>
```
rcourtier: You could remove these 2 attributes since they are not mandatory and in the EQ file, this EquivalentInjection's `regulationCapability` flag is set to `false`.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/DanglingLineUpdateTest.java
```diff
@@ -0,0 +1,228 @@
...
+        assertTrue(Double.isNaN(powerTransformer.getTerminal().getQ()));
+
+        DanglingLine breaker = network.getDanglingLine("Breaker");
+        assertEquals(10.0, breaker.getTerminal().getP(), tol);
+        assertEquals(5.0, breaker.getTerminal().getQ(), tol);
+
+        readCgmesResources(network, DIR, "danglingLine_TP.xml", "danglingLine_SV.xml");
+
+        assertEquals(0.0503090159, acLineSegment.getTerminal().getP(), tol);
+        assertEquals(-145.5845194744, acLineSegment.getTerminal().getQ(), tol);
```
rcourtier, line 108: This is the only place that differs between the two imports:
- EQ + SSH + SV all at once
- EQ + SSH, and then SV

If I'm not mistaken, this is due to the retrieval of P0/Q0 at the boundary side:
- In the first case, the P0/Q0 are available in the EquivalentInjection at the boundary side in the SSH. These non-zero values are then used together with the SV information to calculate the P/Q at the network side.
- In the second case, the P0/Q0 are imported during the first import (EQ + SSH), but when importing the SV (second import), they are not available anymore. For that reason, when we calculate the P/Q at the network side, we consider P0/Q0 at the boundary side are 0.0/0.0.

I wonder if we shouldn't use the existing P0/Q0 of the dangling line in case we couldn't find an attached EquivalentInjection? The ones that have already been imported in a previous import. Something like:

In EquivalentInjectionConversion.java:
```java
    private static void update(DanglingLine danglingLine, Context context) {
        Optional<PropertyBag> cgmesEquivalentInjection = getCgmesEquivalentInjection(danglingLine, context);
        PowerFlow updatedPowerFlow = cgmesEquivalentInjection
                .map(propertyBag -> updatedPowerFlow(danglingLine, propertyBag, context))
                .orElse(new PowerFlow(getDefaultTargetP(danglingLine, context), getDefaultTargetQ(danglingLine, context)));
        [...]
```

Please note that if you accept this change, there is no need anymore to have 2 different unit tests `importSvTogetherTest` and `importSvSeparatelyTest`.

rcourtier (changes requested):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/SwitchConversion.java
marqueslanauja: I've added it.

cgmes/cgmes-conversion/src/test/resources/update/dangling-line/danglingLine_SSH.xml
marqueslanauja: Done

cgmes/cgmes-conversion/src/test/resources/update/dangling-line/danglingLine_SSH_1.xml
marqueslanauja: Done

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/DanglingLineUpdateTest.java
marqueslanauja: I haven’t applied your suggestion, as it would cause the update to use previous values without their use being activated in the configuration. The idea is to use previous values only when they are explicitly required by the configuration.

rcourtier: Never mind my proposal, there is no need to change anything to potentially take into account the previous values if the user decides to do so: the default values are already fetched in the `getTargetP`/`getTargetQ` when the power flow is undefined:

```java
    private static double getTargetP(PowerFlow updatedPowerFlow, DanglingLine danglingLine, Context context) {
        return updatedPowerFlow.defined() ? updatedPowerFlow.p() : getDefaultTargetP(danglingLine, context);
    }
```

Yet the previous values are currently never considered in `getDefaultTargetP`, since this method calls `getDefaultValue` with a predefined List: `context.config().updateDefaultValuesPriority()` which doesn't contain `DefaultValue.PREVIOUS`.
Is it planned to offer that possibility to the user through configuration in a subsequent PR?

marqueslanauja: Yes, I'm preparing a final PR to give the user the option to use previous values after selecting them in the configuration.

rcourtier: Great! Then it's all good!

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/EquivalentInjectionImportTest.java
```diff
@@ -45,8 +45,10 @@ void equivalentInjectionMissingRegulationTargetTest() {
                         "ei_regulation_missing_target_SSH.xml",
                         "ei_regulation_EQ_BD.xml")));
         DanglingLine dl = network.getDanglingLine("ACLS1");
-        // No generation data has been created for the dangling line
```
olperr1, line 48: The behavior changes reflected by this unit test and `equivalentInjectionMissingRegulationTargetTest()` should be reported in the PR's description.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Update.java
```diff
@@ -0,0 +1,185 @@
...
+
+        context.popReportNode();
+    }
+
+    private static void updateDanglingLine(DanglingLine danglingLine, Map<String, PropertyBag> equipmentIdPropertyBag, Context context) {
+        String originalClass = danglingLine.getProperty(Conversion.PROPERTY_CGMES_ORIGINAL_CLASS);
+        switch (originalClass) {
+            case CgmesNames.AC_LINE_SEGMENT -> ACLineSegmentConversion.update(danglingLine, context);
+            case CgmesNames.POWER_TRANSFORMER -> TwoWindingsTransformerConversion.update(danglingLine, context);
+            case CgmesNames.EQUIVALENT_BRANCH -> EquivalentBranchConversion.update(danglingLine, context);
```
olperr1, line 150: The code of these 3 `update(...)` static methods is exactly the same. Is there a real interest to have it in 3 different classes?
If yes, we can keep the methods in the 3 classes, but have them call a new shared method.

olperr1:
Can you provide more details about this section of the PR description?

> An alternative from the current version, which calculated the flow at the boundary by accumulating the injection from all the equipment inside the boundary connected to the boundary node, has been removed.

Did you encounter a case were keeping this behavior was problematic?

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/DanglingLineUpdateTest.java
olperr1: I understand that this was done to allow supporting different import use cases.
In the final PR, the documentation should describes them with the configuration to use for each one.

cgmes/cgmes-model/src/main/resources/CIM16-update.sparql
marqueslanauja: Yes, I've added it.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Update.java
marqueslanauja: Done. I've kept the method in the three classes, and they now call a new shared method.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/EquivalentInjectionImportTest.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/AbstractConductingEquipmentConversion.java
```diff
@@ -234,51 +211,83 @@ public DanglingLine convertToDanglingLine(String eqInstance, int boundarySide, d
...
+        dl.addAlias(terminalId(boundarySide), Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.TERMINAL_BOUNDARY);
+        dl.setProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.TERMINAL_BOUNDARY, terminalId(boundarySide)); // TODO: delete when aliases are correctly handled by mergedlines
         dl.addAlias(terminalId(boundarySide == 1 ? 2 : 1), Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.TERMINAL1);
         dl.setProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + "Terminal", terminalId(boundarySide == 1 ? 2 : 1)); // TODO: delete when aliases are correctly handled by mergedlines
         Optional.ofNullable(topologicalNodeId(boundarySide)).ifPresent(tn -> dl.setProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.TOPOLOGICAL_NODE_BOUNDARY, tn));
         Optional.ofNullable(connectivityNodeId(boundarySide)).ifPresent(cn ->
-            dl.setProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.CONNECTIVITY_NODE_BOUNDARY, cn)
+                dl.setProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.CONNECTIVITY_NODE_BOUNDARY, cn)
         );
         setBoundaryNodeInfo(boundaryNode, dl);
```
rcourtier: If the TP file is provided during the import, these properties are added to the dangling line. However, if the TP file is provided during the update, these properties aren't added.
Could you please add them in `updateDanglingLine()`?

rcourtier (changes requested):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/AbstractConductingEquipmentConversion.java
marqueslanauja: Done. 

The idea is not to record properties during the update process. The information recorded during the conversion process should be sufficient to perform the update. At the moment, the current version does not record any properties during the update.

We were recording the TopologicalNode defined at the boundary in cases where it was not necessary.
We’ve modified the code to record it only when strictly required, that is, when the TopologicalNode is an official boundary node.

We considered a TopologicalNode an official boundary node when it is defined in the TP_BD file.
We also treat a TopologicalNode as official when no boundary information is available,
this occurs when reading bus-branch models that lack both EQ_BD and TP_BD files.

In both cases, the TopologicalNode must be recorded as a property to ensure it is preserved during the export process.

olperr1 (approved):

PR #3313, opened 2025-02-07
AmplNetworkWriter: only call getExtension for the ones that have a writer

jonenst:

NO

Feature/bugfix

- [x] No

when exporting a network in ampl format, we call getExtensions and then we filter to do nothing based on the absence/presence of extension specific writes.

when exporting a network in ampl format, we only call getExtension for data that we will actually serialize

- [ ] Yes
- [x] No

**Other information**:
This is very useful for iidm implementations like network-store which don't have everything loaded in memory and do network requests to load the data

### DIFF
--- a/ampl-converter/src/main/java/com/powsybl/ampl/converter/AmplNetworkWriter.java
+++ b/ampl-converter/src/main/java/com/powsybl/ampl/converter/AmplNetworkWriter.java
@@ -247,10 +247,13 @@ public class AmplNetworkWriter {
     }
 
     private <E> void addExtensions(int extendedNum, Extendable<E> extendable) {
-        for (Extension<E> ext : extendable.getExtensions()) {
-            List<AmplExtension> extList = extensionMap.computeIfAbsent(ext.getName(), k -> new ArrayList<>());
-            extList.add(new AmplExtension(extendedNum, extendable, ext));
-            extensionMap.put(ext.getName(), extList);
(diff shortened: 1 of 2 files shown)


ampl-converter/src/main/java/com/powsybl/ampl/converter/AmplNetworkWriter.java
```diff
@@ -247,10 +247,13 @@ private void writeBuses(AmplExportContext context, TableFormatter formatter) thr
...
-        for (Extension<E> ext : extendable.getExtensions()) {
-            List<AmplExtension> extList = extensionMap.computeIfAbsent(ext.getName(), k -> new ArrayList<>());
-            extList.add(new AmplExtension(extendedNum, extendable, ext));
-            extensionMap.put(ext.getName(), extList);
+        for (String amplExtName : AmplExtensionWriters.getWriterNames()) {
+            Extension<E> ext = extendable.getExtensionByName(amplExtName);
+            if (ext != null) {
+                List<AmplExtension> extList = extensionMap.computeIfAbsent(ext.getName(), k -> new ArrayList<>());
+                extList.add(new AmplExtension(extendedNum, extendable, ext));
+                extensionMap.put(ext.getName(), extList);
```
flo-dup, line 255:
```suggestion
                extensionMap.computeIfAbsent(ext.getName(), k -> new ArrayList<>()).add(new AmplExtension(extendedNum, extendable, ext));
```

PR #3314, opened 2025-02-07
Add a check on minQ <= maxQ

So-Fras:

Quality

- [ ] Yes
- [x] No

There is no check to ensure minQ <= maxQ in a ReactiveCapabilityCurve Point.

In `PointImpl`, a constructor is added to allow de/activating a new check is added to ensure `minQ <= maxQ`. An exception is thrown when the check is activated and `minQ > maxQ`.
The existing constructor will now call the new one with the check activated.

The same check has been added to the `PointAdderImpl`, in `ReactiveCapabilityCurveAdderImpl`.
A condition has been added to deactivate the check for IIDM versions preceding the version 1.14.

- [x] Yes
- [ ] No

If you need to create a `PointImpl` with `minQ > maxQ`, you need to change `PointImpl(p, minQ, maxQ)` to `PointImpl(p, minQ, maxQ, false)`

Custom IIDM implementations' maintainers should implement `ReactiveCapabilityCurveAdder.setShouldCheckMinMaxValues(boolean shouldCheckMinMaxValues)`.

**Other information**:
- Is it really a change in IIDM since no new attribute is added? Maybe it is since an IIDM file with a minQ > maxQ will not be processed the same way as before...
- Should we throw an exception or should we "fix" one of the values (take the smallest value both for minQ and maxQ to be conservative?) with a warning to make the model more robust?

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ReactiveCapabilityCurveImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ReactiveCapabilityCurveImpl.java
@@ -46,8 +46,12 @@ class ReactiveCapabilityCurveImpl implements ReactiveCapabilityCurve {
         private double maxQ;
 
         PointImpl(double p, double minQ, double maxQ) {
-            if (minQ > maxQ) {
-                throw new IllegalStateException("minQ should be inferior or equal to maxQ");
+            this(p, minQ, maxQ, true);
+        }
+
+        PointImpl(double p, double minQ, double maxQ, boolean shouldCheckMinMaxValues) {
+            if (shouldCheckMinMaxValues && minQ > maxQ) {
+                throw new IllegalStateException("maximum reactive power is expected to be greater than or equal to minimum reactive power");
             }
             this.p = p;
             this.minQ = minQ;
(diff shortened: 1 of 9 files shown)


geofjamg:
Yes this is a breaking change as you won't be able to load old file without any exception. You can add this kind of change BUT only in a new version of IIDM and keeping old behaviour (do nothing) for other IIDM versions.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ReactiveCapabilityCurveImpl.java
```diff
@@ -30,6 +30,9 @@ static class PointImpl implements Point {
         private double maxQ;
 
         PointImpl(double p, double minQ, double maxQ) {
+            if (minQ > maxQ) {
+                throw new IllegalStateException("minQ should be inferior or equal to maxQ");
+            }
```
flo-dup, line 35: There's a related TODO written in 2017 (to plan it for IIDM v1.1!!) in the corresponding adder, and a `@Disabled` in a unit test in AbstractReactiveCapabilityCurveTest. Could you remove them?

This should be only activated for next IIDM version. There should be a check in the importer based on the IIDM version to flip or not.

olperr1 (changes requested):

geofjamg (changes requested):
I don't think this is a good idea to have a option to enable or not the check. We should just add the check and fix minq and maxq in importers (like just inverting)

So-Fras:
Newer PR opened on this topic #3617 that mentions this PR.
This one can be closed.

PR #3315, opened 2025-02-07
CGMES export: Add a parameter to export as bus branch

rcourtier:

No.

New feature.

- [X] No

CGMES export's topology kind depends on the network's topology kind computation. A network with at least one node/breaker voltage level will be exported as `NODE_BREAKER`, and a full bus/breaker network will always be exported as `BUS_BRANCH`.

The network topology kind computation has been revised. The network has to be a full node/breaker to be exported in CGMES as `NODE_BREAKER`, likewise it has to be a full bus/breaker to be exported in CGMES as `BUS_BRANCH`. In case of a hybrid (mixed topology) network, the computed topology kind depends on the CIM version for export: it is `BUS_BRANCH` in case of a CIM 16 export and `NODE_BREAKER` for a CIM 100 export.

A new parameter has been added to override the computed export topology kind and force a desired one. It allows for example to export a node/breaker network as `BUS_BRANCH`. The allowed values for this parameter are `NODE_BREAKER` and `BUS_BRANCH`.

The differences in the various configurations of CIM version and topology kind are:
* In case of a CIM 16 `BUS_BRANCH` export:
  * `ConnectivityNode` are not exported
  * non-retained `Switch` are not exported

* In case of a CIM 100 `BUS_BRANCH` export,
  * `ConnectivityNode` are exported from buses of the BusBreakerView
  * non-retained `Switch` are not exported

* In case of a `NODE_BREAKER` export:
  * `ConnectivityNode` are always exported, either from nodes if the VoltageLevel is at node/breaker connectivity level, or from buses if the VoltageLevel is at bus/breaker connectivity level.
  * non-retained `Switch` are always exported.

Also, there is a specificity for the CIM 16 `BUS_BRANCH` export where this export intrinsically means not writting the equipment operation profile. This means the following classes won't be written:
* ConnectivityNode
* StationSupply
* GroundDisconnector
* ActivePowerLimit
* ApparentPowerLimit
* LoadArea
* SubLoadArea
* SvStatus

As well as the following attributes:
* LoadGroup.LoadArea
* LoadGroup.SubLoadArea
* ControlArea.EnergyArea

Note: the CGMES EquipmentOperation profile contains more elements, but they don't exist in PowSyBl and are already not exported.

- [X] Yes
- [ ] No

In case of a network with mixed topology, that is with some `VoltageLevel` in `node/breaker` topology kind and some other `VoltageLevel` in `bus/breaker` topology kind, the CGMES export without this feature was always a `NODE_BREAKER` one. The change introduced here is that the export will now be a `BUS_BRANCH` one if the CIM version for export is 16 (default value).
In order to restore a `NODE_BREAKER` export for a mixed-topology network, one has to set the TOPOLOGY_KIND CgmesExport parameter as follows:
``` java
Network network = ...;

// Export to CGMES as node-breaker
Properties exportParameters = new Properties();
exportParameters.put(CgmesExport.TOPOLOGY_KIND, "NODE_BREAKER");
network.write("CGMES", exportParameters, Path.of("cgmes_file")); 
```

**Other information**:

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
@@ -25,7 +25,6 @@ import com.powsybl.commons.parameters.ParameterType;
 import com.powsybl.commons.report.ReportNode;
 import com.powsybl.commons.xml.XmlUtil;
 import com.powsybl.iidm.network.*;
-import com.powsybl.triplestore.api.PropertyBag;
 import com.powsybl.triplestore.api.PropertyBags;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -163,11 +162,8 @@ public class CgmesExport implements Exporter {
         String namingStrategyImpl = Parameter.readString(getFormat(), parameters, NAMING_STRATEGY_PARAMETER, defaultValueConfig);
         UUID uuidNamespace = UUID.fromString(Parameter.readString(getFormat(), parameters, UUID_NAMESPACE_PARAMETER, defaultValueConfig));
         NamingStrategy namingStrategy = NamingStrategyFactory.create(namingStrategyImpl, uuidNamespace);
-        CgmesExportContext context = new CgmesExportContext(network, referenceDataProvider, namingStrategy);
-        addParametersToContext(context, parameters, reportNode, referenceDataProvider);
-        context.addIidmMappings(network);
-
-        return context;
+        ExportParameters exportParameters = readExportParameters(parameters);
+        return new CgmesExportContext(network, referenceDataProvider, namingStrategy, reportNode, exportParameters);
     }
 
     private Set<String> getBoundaryDcNodes(ReferenceDataProvider referenceDataProvider) {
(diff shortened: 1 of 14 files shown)


olperr1 (changes requested):

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CgmesTopologyKindTest.java
```diff
@@ -0,0 +1,256 @@
...
+
+        // Assert the exports are valid
+        assertValidProfileInHeader(eqFile, topologyKind, cim100Export);
+        assertValidCim16EquipmentOperationElements(eqFile, sshFile, svFile, topologyKind, cim100Export);
+        assertValidNodeBreakerElements(eqFile, topologyKind, cim100Export);
+        assertValidTerminalCount(eqFile, topologyKind, cim100Export);
+    }
+
+    private void assertValidProfileInHeader(String eqFile, CgmesTopologyKind topologyKind, boolean cim100Export) {
+        if (topologyKind == CgmesTopologyKind.NODE_BREAKER && !cim100Export) {
```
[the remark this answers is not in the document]
rcourtier: No it shouldn't. But you are right, there is no check asserting CIM_16_EQ_OPERATION_PROFILE isn't present in the header in case of a bus-breaker cim 16 export.
Thanks for pointing that out!

docs/grid_exchange_formats/cgmes/export.md
```diff
@@ -153,6 +153,53 @@ And the file `manualExampleBasename_SV.xml` will contain:
...
 Remember that, in addition to setting the info for metadata models in the IIDM extensions, you could also rely on parameters passed to the export methods.
 
+## Topology kind
+
+The elements written in the exported files depend on the topology kind of the export and on the CIM version.
+By default, the export topology kind is computed from the IIDM network's `VoltageLevel` [connectivity level](../../grid_model/network_subnetwork.md#voltage-level) as follows:
+* If all `VoltageLevel` of the network are at `node/breaker` connectivity level, then the export topology kind is `NODE_BREAKER`
+* If all `VoltageLevel` of the network are at `bus/breaker` connectivity level, then the export topology kind is `BUS_BRANCH`
+* If some `VoltageLevel` of the network are at `node/breaker` and some other at `bus/breaker` connectivity level, then the export's topology kind depends on the CIM version for export: 
+it is `BUS_BRANCH` for CIM 16 and `NODE_BREAKER` for CIM 100
```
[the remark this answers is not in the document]
rcourtier: Actually no. In prior development versions, I have been using a mixed topology kind for the export, but I drop that topology kind since mixed is the same as bus-branch for cim 16, and is the same as node-breaker for cim 100. So there is no use to it.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -694,6 +703,12 @@ public String getFormat() {
             "Profiles to export",
             List.of("EQ", "TP", "SSH", "SV"),
             List.of("EQ", "TP", "SSH", "SV"));
+    private static final Parameter TOPOLOGY_KIND_PARAMETER = new Parameter(
+            TOPOLOGY_KIND,
+            ParameterType.STRING,
+            "The topology kind of the export",
```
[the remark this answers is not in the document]
rcourtier: done in 0fe7cc70df0973bb915a350223e9983ed13fe925

colinepiloquet (approved):
I tested it and from my point of view it works great, thanks for your work.

olperr1 (approved):

zamarrenolm (changes requested):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -164,6 +165,7 @@ private CgmesExportContext createContext(Network network, Properties parameters,
         NamingStrategy namingStrategy = NamingStrategyFactory.create(namingStrategyImpl, uuidNamespace);
         CgmesExportContext context = new CgmesExportContext(network, referenceDataProvider, namingStrategy);
         addParametersToContext(context, parameters, reportNode, referenceDataProvider);
+        context.addIidmMappings(network);
```
zamarrenolm, line 168: call to `addIidmMappings` is already made inside the constructor of the CGMES export context. We need to call it here again because parameters may have changed something?

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CgmesTopologyKindTest.java
```diff
@@ -0,0 +1,257 @@
...
+        // Build the export parameters
+        Properties exportParams = new Properties();
+        exportParams.put(CgmesExport.TOPOLOGY_KIND, topologyKind.name());
+        if (cim100Export) {
+            exportParams.put(CgmesExport.CIM_VERSION, "100");
+        }
+
+        // Export to CGMES
+        String eqFile = ConversionUtil.writeCgmesProfile(network, "EQ", tmpDir, exportParams);
+        String sshFile = ConversionUtil.writeCgmesProfile(network, "SSH", tmpDir, exportParams);
```
zamarrenolm, line 54: I think we need to check that TP files are also built properly. They are required for CIM16 bus/branch exports. Also, we could check that they have been built properly

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CgmesTopologyKindTest.java
```diff
@@ -0,0 +1,257 @@
...
+                .add();
+        VoltageLevel voltageLevel1 = substation1.newVoltageLevel()
+                .setId("VL_1")
+                .setNominalV(400.0)
+                .setTopologyKind(TopologyKind.BUS_BREAKER)
+                .add();
+        voltageLevel1.getBusBreakerView().newBus()
+                .setId("BUS")
+                .add();
+        voltageLevel1.newGenerator()
```
zamarrenolm, line 156: I think it would be nice to also have a breaker in this voltage level. It could be defined between the generator and its bus, something like:
```java
        voltageLevel1.getBusBreakerView().newBus()
                .setId("BUS")
                .add();
        voltageLevel1.getBusBreakerView().newBus()
                .setId("BUS-GEN")
                .add();
        voltageLevel1.getBusBreakerView().newSwitch().setBus1("BUS")
                .setId("BK-GEN")
                .setBus2("BUS-GEN")
                .setOpen(false)
                .add();
        voltageLevel1.newGenerator()
                .setId("GEN")
                .setBus("BUS-GEN")
...
```

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CgmesTopologyKindTest.java
```diff
@@ -0,0 +1,257 @@
...
+                .add();
+        voltageLevel2.getNodeBreakerView().newBusbarSection()
+                .setId("BBS_2B")
+                .setNode(3)
+                .add();
+        voltageLevel2.getNodeBreakerView().newSwitch()
+                .setNode1(0)
+                .setNode2(3)
+                .setId("DIS")
+                .setKind(SwitchKind.DISCONNECTOR)
```
zamarrenolm, line 195: It is not big deal, but it would be more "natural" to see a `BREAKER` between busbars, not a `DISCONNECTOR`; they are the switching equipment that are usually "retained" in the bus/branch model

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CgmesTopologyKindTest.java
```diff
@@ -0,0 +1,257 @@
...
+            assertEquals(0, getElementCount(svFile, "SvStatus"));
+        }
+    }
+
+    private void assertValidConnectivityElements(String eqFile, CgmesTopologyKind topologyKind, boolean cim100Export) {
+        if (topologyKind == CgmesTopologyKind.NODE_BREAKER) {
+            assertEquals(4, getElementCount(eqFile, "ConnectivityNode"));
+            assertEquals(1, getElementCount(eqFile, "Disconnector"));
+        } else {
+            // In case of a CIM100 bus-branch export, the buses from the BusBreakerView aren't exported as ConnectivityNode
```
zamarrenolm, line 110: I think the comment should say "In case of a CIM16 bus-branch export, the buses ... aren't exported as ConnectivityNode", so in this case the count is 0. Count is 3 for CIM100 export, both bus-branch and node-braker

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
```diff
@@ -158,13 +158,26 @@ public ReferenceDataProvider getReferenceDataProvider() {
...
+                .filter(vl -> vl.getTopologyKind() == TopologyKind.BUS_BREAKER)
+                .count();
+
+        if (nodeBreakerVoltageLevelsCount > 0 && busBreakerVoltageLevelsCount == 0) {
+            return CgmesTopologyKind.NODE_BREAKER;
+        } else if (nodeBreakerVoltageLevelsCount == 0 && busBreakerVoltageLevelsCount > 0) {
+            return CgmesTopologyKind.BUS_BRANCH;
+        } else {
+            // For mixed-topology network, the topology kind is node/breaker for CIM 100 and bus/branch for CIM 16
+            if (getCimVersion() < 100) {
```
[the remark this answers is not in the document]
rcourtier: done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
rcourtier: Yes that was the case. I changed the logic to avoid this second call to `addIidmMappings`. The change is to pass the export parameters to the constructor of CgmesExportContext to compute once and only once the information needed for the export.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CgmesTopologyKindTest.java
rcourtier: done

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CgmesTopologyKindTest.java
rcourtier: done

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CgmesTopologyKindTest.java
rcourtier: done

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CgmesTopologyKindTest.java
rcourtier: fixed

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
zamarrenolm: your refactoring is great, the code is much more clear now. Thanks!

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CgmesTopologyKindTest.java
```diff
@@ -0,0 +1,275 @@
...
+                .add();
+        voltageLevel1.getBusBreakerView().newSwitch()
+                .setId("BK1")
+                .setBus1("BUS")
+                .setBus2("GEN-BUS")
+                .setOpen(false)
+                .add();
+        voltageLevel1.newGenerator()
+                .setId("GEN")
+                .setBus("BUS")
```
zamarrenolm, line 176: the idea was to connect the generator to the new "GEN-BUS", I think you missed to change the bus here.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CgmesTopologyKindTest.java
```diff
@@ -0,0 +1,275 @@
...
+                .setId("LD_NC")
+                .setNode(4)
+                .setP0(0.0)
+                .setQ0(1.0)
+                .add()
+                .setProperty(Conversion.PROPERTY_CGMES_ORIGINAL_CLASS, CgmesNames.NONCONFORM_LOAD);
+        voltageLevel2.getNodeBreakerView().newSwitch()
+                .setId("GRDIS")
+                .setNode1(3)
+                .setNode2(5)
```
zamarrenolm, line 234: In CIM16 bus/branch export validation in CIMdesk emits a warning complaining about a topological node without terminals. It is the TP node corresponding to IIDM node 5, the "empty end" of the ground disconnector. Because the ground disconnector is set to retained, there is a TP node there. But in the export, ground disconnector is not exported, so in the output we end up with a TP node defined but not terminals attached. I do not think this is a big issue. I propose to keep it as it is.

zamarrenolm: CIMdesk emits the same warning in CIM16 node/breaker, because the TP node now has only one terminal, 2 at least are expected (_Fewer than 2 Terminals are associated with TopologicalNode via Association Terminal.TopologicalNode, expecting at least 2._). The same proposal: no big issue here, it is just a warning.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CgmesTopologyKindTest.java
rcourtier: You are right, I have added the GEN-BUS but I missed to connect elements to it. Thanks for noticing!
I have connected the generator and the auxiliary load now.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CgmesTopologyKindTest.java
rcourtier: In my initial developments I had added a Ground after the GroundDisconnector, but only to realize the export of the IIDM Ground to CGMES Ground doesn't exist. Exporting the Ground should fix the warning for the node/breaker export (the node would have 2 terminals associated to it), but not the warning for the bus/branch export.

rcourtier: Created https://github.com/powsybl/powsybl-core/issues/3367.

zamarrenolm (approved):
I've verified the exports using validation tools against a real-world network and they work as expected. Great work!

PR #3325, opened 2025-02-17
Cgmes load zipped profiles mutualize stream

alicecaron:

#3259 

- [ ] Yes
- [x] No


flo-dup (commented):
I'm not 100% convinced by the SonarQube rule, but that way it lead to something more concise. The try-with-resources is centralized in one method, the lambda operations are one-liners. At the cost of the Function complexity though. What do you think of it @olperr1 @alicecaron?

olperr1:
This Sonar rule is very uncleared... But your solution to introduce a lambda and to apply it in a try-with-resource block looks good to me. In some other cases, we may encounter the same problem with the rule without being able to apply your solution, but here it fits well.

olperr1 (approved):

alicecaron:
For this case it works at least :) I agree with this solution

PR #3326, opened 2025-02-19
Doc update: New behavior for BalanceType PROPORTIONAL_TO_REMAINING_MARGIN

vmouradian:

Documentation for the changes introduced by https://github.com/powsybl/powsybl-open-loadflow/pull/1189

- [x] Yes
- [ ] No

If you are a load flow model **maintainer**, you should implement the following change of behavior in your model:
- In the previous versions, when using `PROPORTIONAL_TO_REMAINING_MARGIN` balance type, participation factor of a generator was proportional to the difference `maxP - tartgetP`, even if the power distribution needs to reduce `targetP`.
- Starting from this release, the factor depends on the sign of the mismatch that is distributed. If it is positive, behavior stays the same. If it is negative, participation factor is proportional to the difference `targetP - minP`.


jeandemanged (approved):

docs/simulation/loadflow/configuration.md
```diff
@@ -81,7 +81,7 @@ The default value is `true`.
...
 - If using `PROPORTIONAL_TO_GENERATION_P_MAX` then the participating elements are the generators. The participation factor is computed using the maximum active power target $MaxP$ and the active power control droop. The default droop value is `4`. If present, the simulator uses the droop of the generator given by the [active power control extension](../../grid_model/extensions.md#active-power-control).
 - If using `PROPORTIONAL_TO_GENERATION_P` then the participating elements are the generators. The participation factor is computed using the active power set point $TargetP$.
-- If using `PROPORTIONAL_TO_GENERATION_REMAINING_MARGIN` then the participating elements are the generators. The participation factor is computed using the difference between the maximum active power target $MaxP$ with active power set point $TargetP$.
+- If using `PROPORTIONAL_TO_GENERATION_REMAINING_MARGIN` then the participating elements are the generators. The participation factor is computed using the difference between the active power limit and active power set point $TargetP$. The limit used is $MaxP$ if production needs to increase and $MinP$ if it needs to decrease.
```
vidaldid-rte, line 84: I'm OK with this change. But this algorithm has some limitations:
    - Not supported by all loadflow providers (I think it is specific to OpenLoadFlow)
    - In OpenLoadFlow, it is not available for all modes (just loadflow and security analysis but not sensitivity analysis)
Is there a way to document these limitations ?

vmouradian: I agree on the limitations. On the two things you mention:
- For the support in other providers, I do not really know what is the best, maybe just mention "OLF specific" ? But can we be completely sure of that ?
- For the case of sensi analysis, it is not the only distribution type that is not supported (only 3 are) and this is already mentioned in the doc of OLF (btw, thanks to your comment I noticed that one was missing and opened [OLF-!1201](https://github.com/powsybl/powsybl-open-loadflow/pull/1201)). So I would say that there is nothing to add here about that

vidaldid-rte: My concerne is how to tell to the non expert user that this algorithm is not valid in all cases, although defined at powsybl level.
I believe the others do not have these limitations.
One way could be just to tell "this algorithm may not be supported by all loadfow providers  or all simulation types. If you pkan to use it, check the documentation of your LoadFLow Provider. In the case of OpenLoadFlow there are limitations fo sensitivity analysys, see (+ link)."
What do you think ?

vmouradian: updated in [6869e45](https://github.com/powsybl/powsybl-core/pull/3326/commits/6869e456ef83dafde81b32f02a928ed179ff9dfa)

docs/simulation/loadflow/configuration.md
```diff
@@ -78,16 +78,19 @@ The `distributedSlack` property is an optional property that defines if the acti
...
 - If using `PROPORTIONAL_TO_LOAD` then the participating elements are the loads. The participation factor is computed using the active power $P0$.
 - If using `PROPORTIONAL_TO_CONFORM_LOAD` then the participating elements are the loads which have a conform active power part. The participation factor is computed using the [load detail extension](../../grid_model/extensions.md#load-detail), which specifies the variable and the fixed parts of $P0$. The slack is distributed only on loads that have a variable part. If the extension is not available on a load, the whole $P0$ is considered as a variable.
+  
+Some algorithms may not be supported by all LoadFLow providers or all simulation types. If you plan to use it, check the documentation of your LoadFLow provider. In the case of OpenLoadFlow there are limitations for sensitivity analysis, see [OpenLoadFlow sensitivity analysis documentation](https://github.com/powsybl/powsybl-open-loadflow/blob/main/docs/sensitivity/sensitivity.md).
```
vidaldid-rte, line 90: I think there is a better to way for cross-product doc references
I found this syntax
[PowSyBl Core](inv:powsyblcore:*:*#simulation/security/index).

See https://github.com/powsybl/powsybl-open-loadflow/blob/main/docs/security/inputs.md?plain=1
and result: https://powsybl.readthedocs.io/projects/powsybl-open-loadflow/en/latest/security/inputs.html

vidaldid-rte (approved):

PR #3329, opened 2025-02-20
Cgmes import update. Import EQ and SSH separately. VoltageLevels

marqueslanauja:

- [ ] No

- [ ] Yes
- [ ] No

**Other information**:


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
```diff
@@ -386,12 +459,115 @@ private static String className(Identifiable<?> o) {
...
+    private static String getPropertyNameForOperationalLimitSet(String limitType) {
+        return Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.OPERATIONAL_LIMIT_SET + "_"
+                + limitType;
+    }
+
+    private static String getPropertyName(String operationalLimitSetId, String limitType, String tagProperty) {
+        return Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + tagProperty + "_"
+                + operationalLimitSetId + "_"
+                + limitType;
+    }
```
rcourtier: That's gonna be a problem for PyPowsybl. Yes it works, but there is a downside: we are creating a lot of different property names, which will result in serious performances issues in Pypowsybl. See for instance issue https://github.com/powsybl/pypowsybl/issues/894 that we have met before.

I don't think we should store that many properties anyway. See my next comment.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
```diff
@@ -386,12 +459,115 @@ private static String className(Identifiable<?> o) {
...
+        if (operationalLimitSetId != null) {
+            String operationalLimitId = getOperationalLimitId(getPropertyName(operationalLimitSetId, limitType, CgmesNames.OPERATIONAL_LIMIT), voltageLevel);
+            Double normalValue = getNormalValue(getPropertyName(operationalLimitSetId, limitType, CgmesNames.NORMAL_VALUE), voltageLevel);
+            Double voltageLevelNormalValue = getVoltageLevelNormalValue(voltageLevel, limitType);
+            double defaultLimitValue = getDefaultValue(normalValue != null ? normalValue : voltageLevelNormalValue, previousValue, context);
+            return updatedValue(operationalLimitId, context).orElse(defaultLimitValue);
+        } else {
+            return previousValue;
+        }
+    }
```
rcourtier, line 552: I don't agree with the logic here. It is assumed that the OperationalLimitSet that we have stored as a property during EQ import is the same OperationalLimitSet that we should use during SSH update.

The OperationalLimitSet that we have stored during EQ import has been selected on the basis that it is the one that contains the most restrictive normalLimit (the lowest highLimit and the highest lowLimit). There is no guarantee that this same OperationalLimitSet contains the most restrictive limit during SSH update.

The issue can be shown with a little change to the test data created in this PR: if the SSH limit values of the BusbarSection and the EnergyConsumer are swapped while keeping the same EQ normalLimit values, the wrong limits are chosen for the update: the BusbarSection has the most restrictive EQ normalValue limits and for that reason its SSH limits values are considered for SSH update even though the EnergyConsummer (now with the suggested change) has the most restrictive SSH limits value.

My opinion is that we should do the following:
- During EQ import, keep the lowest highLimit normalValue and the highest lowLimit normalValue. They may not be in the same OperationalLimitSet.
- During SSH update, keep the lowest highLimit value and the highest lowLimit value. They may not be in the same set. If no data can be found in the SSH update, apply the logic to find a defaultValue (the one you have defined here is good: depending on the configuration, take the OperationalLimit's EQ normalValue or the VoltageLevel's EQ highLimit/lowLimit value or the previous SSH value).
- If the SSH limits are less restrictive than the EQ limits, we should keep them anyway, since the EQ only contains a "normalValue" while the SSH contains a real "value".

As a consequence, I don't think we should store that many properties.

cgmes/cgmes-conversion/src/test/resources/update/voltage-level/voltageLevel_EQ.xml
```diff
@@ -0,0 +1,120 @@
...
+        <cim:OperationalLimit.OperationalLimitSet rdf:resource="#_EnergyConsumer-T-OperationalLimitSet" />
+        <cim:OperationalLimit.OperationalLimitType rdf:resource="#_EnergyConsumer-T-OperationalLimitSet-VoltageLimit-High-Type" />
+        <cim:VoltageLimit.normalValue>426.0</cim:VoltageLimit.normalValue>
+    </cim:VoltageLimit>
+
+    <cim:OperationalLimitType rdf:ID="_EnergyConsumer-T-OperationalLimitSet-VoltageLimit-High-Type">
+        <cim:IdentifiedObject.name>EnergyConsumer-T-OperationalLimitSet-VoltageLimit-High-Type</cim:IdentifiedObject.name>
+        <cim:OperationalLimitType.acceptableDuration>45000</cim:OperationalLimitType.acceptableDuration>
+        <eu:OperationalLimitType.kind rdf:resource="http://iec.ch/TC57/CIM100-European#LimitTypeKind.highVoltage" />
+    </cim:OperationalLimitType>
```
rcourtier: You don't need to define another highVoltage `OperationalLimitType`, one is enough.

cgmes/cgmes-conversion/src/test/resources/update/voltage-level/voltageLevel_EQ.xml
```diff
@@ -0,0 +1,120 @@
...
+        <cim:OperationalLimit.OperationalLimitSet rdf:resource="#_EnergyConsumer-T-OperationalLimitSet" />
+        <cim:OperationalLimit.OperationalLimitType rdf:resource="#_EnergyConsumer-T-OperationalLimitSet-VoltageLimit-Low-Type" />
+        <cim:VoltageLimit.normalValue>394.0</cim:VoltageLimit.normalValue>
+    </cim:VoltageLimit>
+
+    <cim:OperationalLimitType rdf:ID="_EnergyConsumer-T-OperationalLimitSet-VoltageLimit-Low-Type">
+        <cim:IdentifiedObject.name>EnergyConsumer-T-OperationalLimitSet-VoltageLimit-Low-Type</cim:IdentifiedObject.name>
+        <cim:OperationalLimitType.acceptableDuration>45000</cim:OperationalLimitType.acceptableDuration>
+        <eu:OperationalLimitType.kind rdf:resource="http://iec.ch/TC57/CIM100-European#LimitTypeKind.lowVoltage" />
+    </cim:OperationalLimitType>
```
rcourtier: You don't need to define another lowVoltage `OperationalLimitType`, one is enough.

marqueslanauja: Done

cgmes/cgmes-conversion/src/test/resources/update/voltage-level/voltageLevel_EQ.xml
marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
marqueslanauja: I've minimized the stored properties to a maximum of four per voltage level.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
marqueslanauja: Considering your opinion, I've made the following changes:

I've deleted the recording of the properties `PROPERTY_OPERATIONAL_LIMIT_SET_RDFID` and `PROPERTY_OPERATIONAL_LIMIT_SET_NAME` at the voltage level (only one set was recorded), as they can differ for high and low limits and are not used in the export process. The limits are exported through the EQ attributes `lowVoltageLimit` and `highVoltageLimit`

During the EQ import, I've recorded the lowest `highLimit normalValue` and the highest `lowLimit normalValue`, along with their corresponding `limitSetId`. The `normalValue` and `limitSetId` are only recorded if the` normalValue` is more restrictive than the `lowVoltageLimit` or `highVoltageLimit`.

Only the `limitSetIds` recorded as properties during the EQ import can be used in the update process. The `SSH` file does not contain any information to associate a `limitSetId` with a voltageLevel. Of course, during the update, we cannot guarantee that the most restrictive limits are selected.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
rcourtier: It is actually 6 per VoltageLevel:
- CGMES.highVoltageLimit
- CGMES.lowVoltageLimit
- CGMES.normalValue_highVoltageLimit
- CGMES.normalValue_lowVoltageLimit
- CGMES.OperationalLimit_highVoltageLimit
- CGMES.OperationalLimit_lowVoltageLimit

which is fine. The real issue was that before your fix, the property names where different for each VoltageLevel. Now it's the 6 same property names for all VoltageLevel in the network.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
rcourtier: As you mentionned, the difficulty is to associate during SSH update the `OperationalLimit` with the `VoltageLevel`. There may be a possibility to do that: at EQ import, you could store in a unique property (to avoid performance issues in PyPowsybl) of the VoltageLevel a serialized json that would contain all this VoltageLevel's OperationalLimit ids and type (high or low). Something similar to what has been done in the past to store CGMES id/name of OperationalLimitSet in a unique property of the connectable. See https://github.com/powsybl/powsybl-core/pull/3225/files.

The code to store ids of voltage limit of a VoltageLevel in EQ import to access them easily during SSH update would be something like this:
```java
try {
      ObjectMapper mapper = new ObjectMapper();
      JsonNode node;
      if (voltageLevel.hasProperty(PROPERTY_VOLTAGE_LIMITS)) {
          node = mapper.readTree(identifiable.getProperty(PROPERTY_VOLTAGE_LIMITS));
      } else {
          node = mapper.createObjectNode();
      }
      ((ObjectNode) node).put(voltageLimit, limitType);
      identifiable.setProperty(PROPERTY_VOLTAGE_LIMITS, mapper.writeValueAsString(node));
  } catch (JsonProcessingException e) {
      throw new PowsyblException(e.getMessage(), e);
  }
```

With the voltage limit ids at your disposal during SSH update, you would then be able to select the most restrictive SSH limit values for the VoltageLevel (or get the default one if it's not in the SSH file).

marqueslanauja: Done. I recorded all the `operationalLimitIds` associated with a voltageLevel as a single string, obtained by concatenating them. I chose this approach for simplicity, but let me know if you'd prefer using `ObjectMapper` instead.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
```diff
@@ -240,23 +259,46 @@ public void convert() {
...
-            } else if (value < vl.getHighVoltageLimit() || Double.isNaN(vl.getHighVoltageLimit())) {
+            convertLowVoltageLimit(operationalLimitId, value);
+        } else if (limitTypeName.equalsIgnoreCase("lowvoltage") || "LimitTypeKind.lowVoltage".equals(limitType)) {
+            convertHighVoltageLimit(operationalLimitId, value);
+        } else {
+            notAssigned(vl);
+        }
+    }
+
+    private void convertLowVoltageLimit(String operationalLimitId, double value) {
```
rcourtier:
```suggestion
    private void convertHighVoltageLimit(String operationalLimitId, double value) {
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
```diff
@@ -240,23 +259,46 @@ public void convert() {
...
+
     private void convertVoltageLimit(double value) {
+        String operationalLimitId = p.getId(CgmesNames.OPERATIONAL_LIMIT);
         String limitTypeName = p.getLocal(OPERATIONAL_LIMIT_TYPE_NAME);
         String limitType = p.getLocal(LIMIT_TYPE);
         if (limitTypeName.equalsIgnoreCase("highvoltage") || "LimitTypeKind.highVoltage".equals(limitType)) {
-            if (value < vl.getLowVoltageLimit()) {
-                context.ignored("HighVoltageLimit", "Inconsistent with low voltage limit (" + vl.getLowVoltageLimit() + "kV)");
-            } else if (value < vl.getHighVoltageLimit() || Double.isNaN(vl.getHighVoltageLimit())) {
+            convertLowVoltageLimit(operationalLimitId, value);
```
rcourtier:
```suggestion
            convertHighVoltageLimit(operationalLimitId, value);
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
```diff
@@ -240,23 +259,46 @@ public void convert() {
...
+        } else if (limitTypeName.equalsIgnoreCase("lowvoltage") || "LimitTypeKind.lowVoltage".equals(limitType)) {
+            convertHighVoltageLimit(operationalLimitId, value);
```
rcourtier, line 270:
```suggestion
            convertLowVoltageLimit(operationalLimitId, value);
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
```diff
@@ -240,23 +259,46 @@ public void convert() {
...
+                addVoltageLimitProperty(vl, CgmesNames.HIGH_VOLTAGE_LIMIT, value);
             }
-        } else if (limitTypeName.equalsIgnoreCase("lowvoltage") || "LimitTypeKind.lowVoltage".equals(limitType)) {
-            if (value > vl.getHighVoltageLimit()) {
-                context.ignored("LowVoltageLimit", "Inconsistent with high voltage limit (" + vl.getHighVoltageLimit() + "kV)");
-            } else if (value > vl.getLowVoltageLimit() || Double.isNaN(vl.getLowVoltageLimit())) {
+        }
+    }
+
+    private void convertHighVoltageLimit(String operationalLimitId, double value) {
```
rcourtier:
```suggestion
    private void convertLowVoltageLimit(String operationalLimitId, double value) {
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
```diff
@@ -240,23 +259,46 @@ public void convert() {
...
+        } else {
+            notAssigned(vl);
+        }
+    }
+
+    private void convertLowVoltageLimit(String operationalLimitId, double value) {
+        if (value < vl.getLowVoltageLimit()) {
+            context.ignored("HighVoltageLimit", "Inconsistent with low voltage limit (" + vl.getLowVoltageLimit() + "kV)");
+        } else {
+            addVoltageLimitProperty(vl, CgmesNames.HIGH_VOLTAGE_LIMIT, operationalLimitId);
```
rcourtier: We could move it before the if/else block to always consider the SSH value even if the EQ value isn't good.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
```diff
@@ -240,23 +259,46 @@ public void convert() {
...
-                context.ignored("LowVoltageLimit", "Inconsistent with high voltage limit (" + vl.getHighVoltageLimit() + "kV)");
-            } else if (value > vl.getLowVoltageLimit() || Double.isNaN(vl.getLowVoltageLimit())) {
+        }
+    }
+
+    private void convertHighVoltageLimit(String operationalLimitId, double value) {
+        if (value > vl.getHighVoltageLimit()) {
+            context.ignored("LowVoltageLimit", "Inconsistent with high voltage limit (" + vl.getHighVoltageLimit() + "kV)");
+        } else {
+            addVoltageLimitProperty(vl, CgmesNames.LOW_VOLTAGE_LIMIT, operationalLimitId);
```
rcourtier: Likewise, we could move it before the if/else block to always consider the SSH value even if the EQ value isn't good.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
```diff
@@ -386,12 +470,116 @@ private static String className(Identifiable<?> o) {
...
+            return OptionalDouble.empty();
+        }
+        return Arrays.stream(operationalLimitIds.split(";"))
+                .map(operationalLimitId -> updatedValue(operationalLimitId, context))
+                .filter(OptionalDouble::isPresent)
+                .mapToDouble(OptionalDouble::getAsDouble)
+                .reduce(isMax(limitType) ? Double::max : Double::min);
+    }
+
+    private static boolean isMax(String limitType) {
```
rcourtier:
```suggestion
    private static boolean getUpperBound(String limitType) {
```

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/VoltageLevelUpdateTest.java
```diff
@@ -0,0 +1,74 @@
...
+        assertEq(network.getVoltageLevel("VoltageLevel"), 396.0, 425.0);
+
+        readCgmesResources(network, DIR, "voltageLevel_SSH.xml");
+        assertSsh(network.getVoltageLevel("VoltageLevel"), 405.0, 435.0);
+
+        readCgmesResources(network, DIR, "voltageLevel_SSH_1.xml");
+        assertSsh(network.getVoltageLevel("VoltageLevel"), 399.0, 430.0);
+    }
+
+    private static void assertEq(VoltageLevel voltageLevel, double lowVoltageLimit, double highVoltageLimit) {
```
rcourtier: You can remove the `double lowVoltageLimit` and the `double highVoltageLimit` from this method arguments and inline the values `396.0` and `425.0` instead.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/VoltageLevelUpdateTest.java
```diff
@@ -0,0 +1,74 @@
...
+        assertNotNull(voltageLevel);
+        assertNotNull(voltageLevel.getProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.LOW_VOLTAGE_LIMIT));
+        assertNotNull(voltageLevel.getProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.HIGH_VOLTAGE_LIMIT));
+        assertTrue(voltageLevel.getPropertyNames().stream().anyMatch(propertyName -> propertyName.contains(CgmesNames.OPERATIONAL_LIMIT + "_") && propertyName.contains(CgmesNames.LOW_VOLTAGE_LIMIT)));
+        assertTrue(voltageLevel.getPropertyNames().stream().anyMatch(propertyName -> propertyName.contains(CgmesNames.OPERATIONAL_LIMIT + "_") && propertyName.contains(CgmesNames.HIGH_VOLTAGE_LIMIT)));
+        assertTrue(voltageLevel.getPropertyNames().stream().anyMatch(propertyName -> propertyName.contains(CgmesNames.NORMAL_VALUE + "_") && propertyName.contains(CgmesNames.LOW_VOLTAGE_LIMIT)));
+        assertTrue(voltageLevel.getPropertyNames().stream().anyMatch(propertyName -> propertyName.contains(CgmesNames.NORMAL_VALUE + "_") && propertyName.contains(CgmesNames.HIGH_VOLTAGE_LIMIT)));
```
rcourtier: You can check the presence of the 6 property names in a simpler way:
```java
        assertTrue(voltageLevel.hasProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.LOW_VOLTAGE_LIMIT));
        assertTrue(voltageLevel.hasProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.HIGH_VOLTAGE_LIMIT));
        assertTrue(voltageLevel.hasProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.OPERATIONAL_LIMIT + "_" + CgmesNames.LOW_VOLTAGE_LIMIT));
        assertTrue(voltageLevel.hasProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.OPERATIONAL_LIMIT + "_" + CgmesNames.HIGH_VOLTAGE_LIMIT));
        assertTrue(voltageLevel.hasProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.NORMAL_VALUE + "_" + CgmesNames.LOW_VOLTAGE_LIMIT));
        assertTrue(voltageLevel.hasProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.NORMAL_VALUE + "_" + CgmesNames.HIGH_VOLTAGE_LIMIT));
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/VoltageLevelUpdateTest.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/VoltageLevelUpdateTest.java
marqueslanauja: Done

rcourtier (approved):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
```diff
@@ -240,35 +259,55 @@ public void convert() {
...
         }
     }
 
+    private void convertHighVoltageLimit(String operationalLimitId, double value) {
+        addVoltageLimitProperty(vl, CgmesNames.HIGH_VOLTAGE_LIMIT, operationalLimitId);
+        if (value < vl.getLowVoltageLimit()) {
+            context.ignored("HighVoltageLimit", "Inconsistent with low voltage limit (" + vl.getLowVoltageLimit() + "kV)");
+        } else if (value < vl.getHighVoltageLimit() || Double.isNaN(vl.getHighVoltageLimit())) {
+            vl.setHighVoltageLimit(value);
+            addVoltageLimitProperty(vl, CgmesNames.HIGH_VOLTAGE_LIMIT, value);
```
olperr1: It would be nice to have 2 different names instead of 2 different signatures for `addVoltageLimitProperty(…)`.
Reading this code, some can imagine there's a bug here: the property is defined at line 282 with an id, and sometimes it is replaced by a value at line 287.
I suggest to use respectively `addVoltageLimitIdProperty(…)` and `addVoltageLimitNormalValueProperty(…)` to make it more readable.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
```diff
@@ -386,12 +466,116 @@ private static String className(Identifiable<?> o) {
...
+        double defaultLimitValue = getDefaultValue(getNormalValue(getPropertyName(limitSubclass, isInfiniteDuration, duration, CgmesNames.NORMAL_VALUE), operationalLimitsGroup), previousValue, context);
+        return updatedValue(operationalLimitId, context).orElse(defaultLimitValue);
+    }
+
+    private static OptionalDouble updatedValue(String operationalLimitId, Context context) {
+        PropertyBag p = context.operationalLimit(operationalLimitId);
+        return (p == null) ? OptionalDouble.empty() : p.asOptionalDouble("value");
+    }
+
+    private static String getOperationalLimitIds(String propertyName, VoltageLevel voltageLevel) {
```
olperr1: Can you change the signature of this method to return an `Optional<String>`?
I think you can also change the first parameter to have `String limitType` instead of `String propertyName`:


```java
private static Optional<String> getOperationalLimitIds(String limitType, VoltageLevel voltageLevel) {
    String propertyName = getPropertyName(limitType, CgmesNames.OPERATIONAL_LIMIT)
    String ids = voltageLevel.getProperty(propertyName);
    if (ids == null || ids.isEmpty()) {
            return Optional.empty();
    }
    return Arrays.stream(ids.split(";"));
}

File: cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java [hunk 16]
```diff
@@ -386,12 +466,116 @@ private static String className(Identifiable<?> o) {
...
+                .filter(OptionalDouble::isPresent)
+                .mapToDouble(OptionalDouble::getAsDouble)
+                .reduce(getUpperBound(limitType) ? Double::max : Double::min);
+    }
+
+    private static boolean getUpperBound(String limitType) {
+        return CgmesNames.LOW_VOLTAGE_LIMIT.equals(limitType);
+    }
+
+    private static Double getNormalValue(String propertyName, VoltageLevel voltageLevel) {
```
olperr1: Can you also adapt this method to have `String limitType` as the first parameter?
The `propertyName` would then be computed in this method.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
```diff
@@ -386,12 +466,116 @@ private static String className(Identifiable<?> o) {
...
+    private static double getDefaultValue(Double normalValue, double previousValue, Context context) {
+        return getDefaultValue(normalValue, previousValue, normalValue, normalValue != null ? normalValue : previousValue, context);
+    }
+
+    private static double getValue(VoltageLevel voltageLevel, String limitType, double previousValue, Context context) {
+        String operationalLimitIds = getOperationalLimitIds(getPropertyName(limitType, CgmesNames.OPERATIONAL_LIMIT), voltageLevel);
+        Double normalValue = getNormalValue(getPropertyName(limitType, CgmesNames.NORMAL_VALUE), voltageLevel);
+        Double voltageLevelNormalValue = getVoltageLevelNormalValue(voltageLevel, limitType);
+        double defaultLimitValue = getDefaultValue(normalValue != null ? normalValue : voltageLevelNormalValue, previousValue, context);
+        return updatedValue(limitType, operationalLimitIds, context).orElse(defaultLimitValue);
```
olperr1: Instead of an `orElse(…)` instruction, you could use an `orElseGet(…)` one, and compute `normalValue`, `voltageLevelNormalValue` and `defaultLimitValue` only if needed.
This would highlight the fact that we first try to compute the updated value using the operational limit values. And if it is not possible, we use the normal values as a fallback.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
marqueslanauja: Done. I've used `List<String> getOperationalLimitIds(String limitType, VoltageLevel voltageLevel)`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
marqueslanauja: Done

olperr1 (approved):

PR #3330, opened 2025-02-20
Triple store queries: replace subqueries with LIMIT 1 by proper subqueries using GROUP BY

zamarrenolm:

Bug fix

- [x] No

Some SPARQL queries use subqueries with LIMIT 1 expecting inner variables to be instantiated by the outer scope. For example, query for regions and subregions of a substation. The subquery combined with LIMIT was used to obtain only the first region associated with a given substation.

This way of querying the triple store works in some situations, but is not a guarantee, according to the SPARQL language specification. In theory it should never work: _Subqueries are executed first and then its results are projected on the outer query_ [https://www.w3.org/TR/sparql11-query/#subqueries](https://www.w3.org/TR/sparql11-query/).

As a minimum example, this query returns correct results for looking one of the past projects of a person using our current version of RDF4J:
```sparql
prefix foaf: <http://xmlns.com/foaf/0.1/>
SELECT *
WHERE {
    ?Person a foaf:Person ;
        foaf:name ?name .
    OPTIONAL {
        {
            SELECT ?Project
            WHERE {
                ?Person foaf:pastProject ?Project .
            }
            LIMIT 1
        }
        ?Project foaf:name ?projectName .
    }
}
```
It should return the same Project for all Persons, because in the subquery the ?Person variable should not be instantiated from the outer scope, and just the first project of any Person should be returned up. But instead, this version of the query returns a (proper) Project for every Person.

Just by altering very slightly the query, we obtain (as expected) bad results:
```sparql
prefix foaf: <http://xmlns.com/foaf/0.1/>
SELECT *
WHERE {
    ?Person a foaf:Person ;
        foaf:name ?name .
    OPTIONAL {
        {
            SELECT ?Project ?projectName
            WHERE {
                ?Person foaf:pastProject ?Project .
                ?Project foaf:name ?projectName .
            }
            LIMIT 1
        }
    }
}
```

The proper way of obtaining the intended data is to use the subquery with the variable that we want to match in the outer scope also projected. In the subquery, the additional data for it should be obtained using a GROUP BY clause and  aggregate function. Performance wise, SAMPLE could be used to aggregate, but that could lead to nondeterministic results, so MIN is used.

This is the proposed way of rewriting the subqueries to remove the LIMIT 1 related to the (wrong) assumption that inner variables were already instantiated:
```sparql
prefix foaf: <http://xmlns.com/foaf/0.1/>
SELECT *
WHERE {
    ?Person a foaf:Person ;
        foaf:name ?name .
    OPTIONAL {
        {
            SELECT ?Person (MIN(?Project) AS ?Project)
            WHERE {
                ?Person foaf:pastProject ?Project .
            }
            GROUP BY ?Person
        }
        ?Project foaf:name ?projectName .
    }
}
```

- [ ] Yes
- [X] No

**Other information**:

To evaluate the potential impact on performance the change has been verified importing a huge network (an 1.6 GB EQ XML file of a real world network). The size of the network is summarised in the following table:

Class | # of objects
-- | --:
Substations | 5,000
Voltage levels | 9,000
Connectivity nodes | 60,000
Busbars | 12,000
Lines | 9,000
Transformers | 2,500
Switches | 70,000
Terminals | 180,000

The import of the network takes an average of 72s in a reference machine before the change.
With this change the average time for import is 73s.
The figures have been obtained from 6 sample runs on an Apple MacBook Pro 2.3 GHz 8-core Intel Core i9 with 16 GB main memory.


rcourtier (dismissed):

triple-store/triple-store-test/src/test/java/com/powsybl/triplestore/test/TripleStoreSubqueryLimit1Test.java
```diff
@@ -0,0 +1,60 @@
+package com.powsybl.triplestore.test;
```
olperr1, line 1: The copyright is missing:

```suggestion
/**
 * Copyright (c) 2025, RTE (http://www.rte-france.com)
 * This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/.
 * SPDX-License-Identifier: MPL-2.0
 */
package com.powsybl.triplestore.test;
```

triple-store/triple-store-test/src/test/java/com/powsybl/triplestore/test/TripleStoreSubqueryLimit1Test.java
```diff
@@ -0,0 +1,60 @@
...
+
+import com.powsybl.triplestore.api.QueryCatalog;
+import com.powsybl.triplestore.api.TripleStoreFactory;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class TripleStoreSubqueryLimit1Test {
```
olperr1, line 8: The authorship is missing:

```suggestion
/**
 * @author Luma Zamarreño {@literal <zamarrenolm at aia.es>}
 */
class TripleStoreSubqueryLimit1Test {
```

PR #3338, opened 2025-02-28
Scalable access

Lisrte:

No

Feature

`LoadScalable` is package private

`LoadScalable` is public and setter and getter are added for P0, Q0.

- [ ] Yes
- [x] No

**Other information**:

Needed by the powsybl-dynawo PR [420](https://github.com/powsybl/powsybl-dynawo/pull/420)


geofjamg (commented):
I think the reason for this to be private package is that it is supposed to be accessed through Scalable.onLoad(...). Why do you need to open this class ?

Lisrte:
> I think the reason for this to be private package is that it is supposed to be accessed through Scalable.onLoad(...). Why do you need to open this class ?

For Dynawo margin calculation [PR 420](https://github.com/powsybl/powsybl-dynawo/pull/420) we need  to calculate proportional scalable on loads and take the result as input for an [automation system](https://github.com/powsybl/powsybl-dynawo/pull/420/files#diff-3136ecc44702d5a737bb0a957a82d539f10ca9ecf2dc51b66068821dea4e7846).
Powsybl-core does the computation we need, but update the network with the result.
In order to not duplicate code the preferred solution was to open `LoadScallable` and extends it in [powsybl-dynawo](https://github.com/powsybl/powsybl-dynawo/pull/420/files#diff-324a131fc57c5bd01056c46d605d54ee596e8fb88c4451db0a47e4f211ca0e6b).

olperr1 (approved):

olperr1 (changes requested):
Can you set the constructors of `GeneratorScalable`and `StackScalable` to `protected` too?

Lisrte:
> Can you set the constructors of `GeneratorScalable`and `StackScalable` to `protected` too?

Done

olperr1 (approved):

PR #3342, opened 2025-03-04
Support of all characters in `IdWithWildcardsNetworkElementIdentifier`

olperr1:

No

Feature

- [X] No

- The only allowed characters in the `IdWithWildcardsNetworkElementIdentifier` pattern were from A to Z (capitalized or not), the digits, '_', '.' or '-'.
- The wildcard was '?'.

- All the UTF-16 characters are allowed in the pattern.
- By default, the wildcard is '?', but the user can change this character if needed / wanted.

- [X] Yes
- [ ] No

- The JSON serialization of `IdWithWildcardsNetworkElementIdentifier` now contains a `wildcard` attribute. When reading a JSON without this attribute, the default wildcard (`?`) is used.
- As a consequence, the serializations of the following elements (which can contain a `IdWithWildcardsNetworkElementIdentifier`) are potentially impacted. Their serialization version numbers were increased:

| Category | Element | Previous version | New version |
| --- | --- | --- | --- |
| Action list | `ActionList` | `1.1` | `1.2` |
| Contingency list | `IdentifierContingencyList` | `1.2` | `1.3` |
| Contingency list | `ListOfContingencyLists` | `1.0` | `1.1` |

:warning: *The JSON serialized version of these elements (action or contingency list) may be not readable with older versions of PowSyBl.*

Also, this change of serialization versions can have an impact on your unit tests (if you use a JSON serialized object as a reference).

**Other information**:


alicecaron (dismissed):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/identifiers/IdWithWildcardsNetworkElementIdentifier.java
```diff
@@ -16,64 +16,77 @@
...
         }
         if (separatorNumber == 0) {
             throw new PowsyblException("There is no wildcard in your identifier, please use IdBasedNetworkElementIdentifier instead");
         }
-        identifier = identifier.replace(".", "\\.").replace(WILDCARD, '.');
+        // Escape all non wildcard characters
+        String[] chunks = originalIdentifier.split(Pattern.quote(wildcardCharacter));
+        StringBuilder sb = new StringBuilder();
+        for (String chunk : chunks) {
+            sb.append(Pattern.quote(chunk)).append('.');
```
flo-dup: Isn't there one extra dot at the end?

olperr1: Oh, you are right!
I fixed this and added a unit test.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/identifiers/IdWithWildcardsNetworkElementIdentifier.java
```diff
@@ -69,8 +69,14 @@ private void initialize() {
         // Escape all non wildcard characters
         String[] chunks = originalIdentifier.split(Pattern.quote(wildcardCharacter));
         StringBuilder sb = new StringBuilder();
-        for (String chunk : chunks) {
-            sb.append(Pattern.quote(chunk)).append('.');
+        for (int i = 0; i < chunks.length; i++) {
```
flo-dup: Can't you use `String.join(".", chunks)` instead?

olperr1: Each chunk should be escaped with `Pattern.quote(...)`.
But it is possible to write this code in a more Java8+ way by using a `StringJoiner`.
I changed this.

PR #3347, opened 2025-03-06
Psse: fix the export process

marqueslanauja:

Bug fix

The attributes `XFRRAT` and `NXFRAT` in the case identification data are exported with decimals.
A comment line `"/ BEGIN SUBSTATION NODE DATA"` is written after the substation record data, which is not supported by PSSE.
The attribute `vsref` of the facts is considered a double and must be an integer.

The attributes `XFRRAT` and `NXFRAT` in the case identification data must be exported without decimals.
The comment line `"/ BEGIN SUBSTATION NODE DATA"` has been removed.
The attribute `vsref` of the facts is considered an integer.
Define reasonable buses in the fact records.

- [ ] Yes
- [x] No

**Other information**:


zamarrenolm (approved):

psse/psse-model/src/main/java/com/powsybl/psse/model/pf/io/FactsDeviceData.java
```diff
@@ -23,15 +23,14 @@
  */
 class FactsDeviceData extends AbstractRecordGroup<PsseFacts> {
 
-    private static final String[] FIELD_NAMES_32_33 = {"name", "i", "j", "mode", "pdes", "qdes", "vset", "shmx", "trmx", "vtmn", "vtmx", "vxmx", "imx", "linx",
-        "rmpct", "owner", "set1", "set2", "vsref", "remot", STR_MNAME};
+    private static final String[] FIELD_NAMES_32_33 = {"name", "i", "j", "mode", "pdes", "qdes", "vset", "shmx", "trmx", "vtmn", "vtmx", "vsmx", "imx", "linx", "rmpct", "owner", "set1", "set2", "vsref", "remot", STR_MNAME};
```
olperr1, line 26: Can you add a unit test for the `vxmx` -> `vsmx` change?

olperr1 (changes requested):

psse/psse-model/src/main/java/com/powsybl/psse/model/pf/io/FactsDeviceData.java
marqueslanauja: This change has already been checked by the test `ieee14BusCompletedWriteTest`. A bad reference to a PSSE attribute only affects the export process. A badly defined reference causes the attribute not to be exported. As you can see in the differences of the exported file, before the attribute was not exported, `two commas were together`. After the change, the attribute is exported with the default value (1.0).

image

olperr1: Thanks, I missed this one.

geofjamg (approved):

PR #3353, opened 2025-03-11
Partly clean ReportNode use in ValidationUtil

flo-dup:

No

Bug fix

The ReportNode is used in ValidationUtil with generated keys (Validable::getMessageHeader) for message template

The ReportNode is used in ValidationUtil with fixed keys for a given template

- [ ] Yes
- [x] No

**Other information**:
:warning: The message lacks the correct information which is expected by the user: getMessageHeader usually gives the equipmentType and the equipment id. But this was already the case before, as this information was used as a template key, hence never displayed to the user. A PR will follow later to fix this, but this requires breaking changes.


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/NetworkReports.java
```diff
@@ -142,4 +142,194 @@ public static void testImportPostProcessor(ReportNode reportNode) {
...
+
+    public static void ptcPhaseRegulationFixedMode(ReportNode reportNode) {
+        reportNode.newReportNode()
+                .withMessageTemplate("core.iidm.network.ptcPhaseRegulationFixedMode")
+                .withSeverity(TypedValue.ERROR_SEVERITY)
+                .add();
+    }
+
+    public static void targetDeadbandUndefinedValue(String validableType, ReportNode reportNode) {
+        String key = switch (validableType) {
```
alicecaron, line 324: Instead of manipulating a String can we create en enum to handle all the possibilities? For instance:
```
public enum ValidableType {
    SHUNT_COMPENSATOR,
    PHASE_TAP_CHANGER,
    RATIO_TAP_CHANGER
}
```
But there are more than 3 ValidableType so maybe we need list them all

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -475,23 +507,35 @@ private static ValidationLevel checkRatioTapChangerRegulation(Validable validabl
...
-                throwExceptionOrLogError(validable, "a regulation value has to be set for a regulating ratio tap changer", actionOnError, reportNode);
+                throwExceptionOrLogError(validable, "a regulation value has to be set for a regulating ratio tap changer", actionOnError);
+                if (actionOnError != ActionOnError.IGNORE) {
+                    NetworkReports.regulatingRtcNoRegulationValue(reportNode);
+                }
                 validationLevel = ValidationLevel.min(validationLevel, ValidationLevel.EQUIPMENT);
             }
             if (regulationMode == RatioTapChanger.RegulationMode.VOLTAGE && regulationValue <= 0) {
-                throwExceptionOrLogError(validable, "bad target voltage " + regulationValue, actionOnError, reportNode);
+                throwExceptionOrLogError(validable, "bad target voltage " + regulationValue, actionOnError);
```
olperr1: Can you add a report for this case?

olperr1 (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/NetworkReports.java
flo-dup: Yes, we could differentiate with an enum, but I didn't want to change the already existing String differentiation. In fact I think we'd better un-factorize this call instead... Creating a string which is then "switched case" in a factorized method is not a proper factorization I think, at least for such a basic use case.

commons/src/main/resources/com/powsybl/commons/reports.properties
```diff
@@ -122,19 +122,46 @@ core.iidm.modification.voltageLevelNotFound = Voltage level ${voltageLevelId} is
...
 core.iidm.network.InconsistentAliasValues = Inconsistencies found for alias type '${type}'('${alias1}' for '${danglingLineId1}' and '${alias2}' for '${danglingLineId2}'). Types are respectively renamed as '${type}_1' and '${type}_2'.
 core.iidm.network.InconsistentPropertyValues = Inconsistencies of property ${propertyName} between ${danglingLineId1} (value=${propertyValue1}) and ${danglingLineId2} (value=${propertyValue2}). Property is not added to merged line
+core.iidm.network.invalidP0 = p0 is invalid
+core.iidm.network.invalidQ0 = q0 is invalid
 core.iidm.network.MoveCommonAlias = Alias ${alias} found in dangling lines ${danglingLineId1} and ${danglingLineId2} is moved to their merged line.
 core.iidm.network.PropertyOnlyOnOneSide = Inconsistencies of property ${propertyName} between both sides (${danglingLineId1) on side 1 and ${danglingLineId2} on side2) of merged line. Side ${side} has no value. Value on other side is kept.
+core.iidm.network.ptcPhaseRegulationFixedMode = phase regulation cannot be on if mode is FIXED
```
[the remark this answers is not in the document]
flo-dup: Yes indeed! missed that.

PR #3357, opened 2025-03-12
Handle negative zero case in reactive capability curves

olperr1:

No

Bug fix

When the lowest point of a reactive capability curve has a `p` equals to `0.0`, calling `curve.getMinQ(-0.0);` leads to a `NullPointerException`.

When the lowest point of a reactive capability curve has a `p` equals to `0.0`, calling `curve.getMinQ(-0.0);` return the `minQ` of the lowest point.

- [ ] Yes
- [X] No

**Other information**:

The [JLS](https://docs.oracle.com/javase/specs/jls/se17/html/jls-4.html#jls-4.2.3) states that `0.0 == -0.0`, but [`Double.compareTo(Double)`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Double.html#compareTo(java.lang.Double)) considers that `0.0 > -0.0`.

For a curve whose lowest point has `p == 0`, this inconsistency leads to calling [`points.floorEntry(-0.0).getValue()`](https://github.com/powsybl/powsybl-core/blob/ac44e4462e7e80ad3d6c00b452d4b0a258a2abaf/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ReactiveCapabilityCurveImpl.java#L127) (because `p >= this.getMinP()` — `-0.0 >= 0.0`) but the `TreeMap` considers that there's no entry with a `p` lower or equal to `-0.0` (it uses `Double.compareTo(Double)`).
Therefore, a `NullPointerException` is thrown:

```
java.lang.NullPointerException: Cannot invoke "java.util.Map$Entry.getValue()" because the return value of "java.util.TreeMap.floorEntry(Object)" is null
```

<br/>
This PR sets a comparator to the tree map for it to consider `0.0 == -0.0`.

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ReactiveCapabilityCurveImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ReactiveCapabilityCurveImpl.java
@@ -13,6 +13,7 @@ import com.powsybl.iidm.network.util.ReactiveCapabilityCurveUtil;
 
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.TreeMap;
 import java.util.function.ToDoubleFunction;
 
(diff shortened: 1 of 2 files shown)


vidaldid-rte (approved):

colinepiloquet (approved):
Thanks!

flo-dup (approved):

PR #3358, opened 2025-03-13
[ReportNode] Remove bundle name from withMessageTemplate

flo-dup:

No

Feature

One need to add the bundle name each time a template message key is provided.

The bundle name is provided at root creation through a MessageTemplateProvider, or a child creation if the bundle needs to be changed

- [x] Yes
- [ ] No

TODO


commons/src/main/java/com/powsybl/commons/report/PowsyblCoreReportResourceBundles.java
```diff
@@ -10,11 +10,14 @@
 /**
  * @author Alice Caron {@literal <alice.caron at rte-france.com>}
  */
-public final class ReportBundleBaseName {
+public final class PowsyblCoreReportResourceBundles {
 
-    public static final String BUNDLE_BASE_NAME = "com.powsybl.commons.reports";
-    public static final String BUNDLE_TEST_BASE_NAME = "i18n.reports";
+    public static final String BASE_NAME = "com.powsybl.commons.reports";
+    public static final String TEST_BASE_NAME = "i18n.reports";
```
geofjamg, line 16: You should not mix runtime and test configuration. You should create another PowsyblCoreTestReportResourceBundles which contains the 2 bundles like this one and keep a single bundle  BASE_NAME for this one.

dynamic-simulation/dynamic-simulation-dsl/src/main/java/com/powsybl/dynamicsimulation/groovy/DynamicSimulationReports.java
```diff
@@ -16,26 +16,26 @@
 public final class DynamicSimulationReports {
 
     public static ReportNode buildRootDynamicSimulationTool() {
-        return ReportNode.newRootReportNode()
-                .withLocaleMessageTemplate("core.dynasim.dynamicSimulationTool", ReportBundleBaseName.BUNDLE_BASE_NAME)
+        return ReportNode.newRootReportNode(PowsyblCoreReportResourceBundles.BASE_NAME)
```
geofjamg, line 19: A remark that might be helpful for the design of this API: you almost never use this part of the API in tests but prefer calling  `withMessageTemplateProvider` after

geofjamg:
As a general comment, this change is much much cleaner that previous API so that we don't need to pass explicitly the message bundle at each message call.

commons/src/main/java/com/powsybl/commons/report/PowsyblCoreReportResourceBundles.java
flo-dup: Sliced in two files, the test one going to powsybl-commons-test module.

dynamic-simulation/dynamic-simulation-dsl/src/main/java/com/powsybl/dynamicsimulation/groovy/DynamicSimulationReports.java
flo-dup: I changed it in unit tests, they all use the `newRootReportNode(String... bundleBaseNames)` interface now. The withMessageTemplateProvider is kept for custom/expert usecases.

geofjamg (approved):

PR #3372, opened 2025-03-19
Rework nodes mapping to allow export of isolated nodes

rcourtier:

No.

Fix.

The node of an isolated IIDM equipment (a connectable not linked to any other element either through switch or internal connection) isn't exported to CGMES. This leads to a cardinality issue in the CGMES EQ file where the terminal of the isolated equipment doesn't have a ConnectivityNode association.

The node of an isolated IIDM equipment is exported to CGMES.

- [ ] Yes
- [X] No

**Other information**:

Nodes are now exported in order of their number. This shall ensure that nodes are exported in the same order they have been imported and thus in the case of a round-trip import/export the numbers should be the same.

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
@@ -25,6 +25,7 @@ import com.powsybl.iidm.network.Identifiable;
 import com.powsybl.iidm.network.extensions.RemoteReactivePowerControl;
 import com.powsybl.iidm.network.extensions.VoltagePerReactivePowerControl;
 
+import com.powsybl.math.graph.TraverseResult;
 import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -117,7 +118,7 @@ public final class EquipmentExport {
             for (VoltageLevel vl : network.getVoltageLevels()) {
                 String cgmesVlId = context.getNamingStrategy().getCgmesId(vl);
                 if (vl.getTopologyKind() == TopologyKind.NODE_BREAKER && !context.isBusBranchExport()) {
-                    writeNodes(vl, cgmesVlId, new VoltageLevelAdjacency(vl.getNodeBreakerView(), context), mapNodeKey2NodeId, cimNamespace, writer, context);
+                    writeNodes(vl, cgmesVlId, getVoltageLevelAdjacencies(vl.getNodeBreakerView(), context), mapNodeKey2NodeId, cimNamespace, writer, context);
                 } else {
                     writeBuses(vl, cgmesVlId, mapNodeKey2NodeId, cimNamespace, writer, context);
                 }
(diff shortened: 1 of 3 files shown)


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
```diff
@@ -1537,83 +1499,66 @@ private static String connectivityNodeId(Map<String, String> mapNodeKey2NodeId,
 
     private static class VoltageLevelAdjacency {
 
-        private final List<List<Integer>> voltageLevelNodes;
+        private final List<List<Integer>> adjacencies;
```
flo-dup: VoltageLevelAdjacency is just wrapping the List<List<Integer>>, so you could use that list instead directly.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
```diff
@@ -1537,83 +1499,66 @@ private static String connectivityNodeId(Map<String, String> mapNodeKey2NodeId,
...
-                            return;
-                        }
-                        adjacentNodes.add(adjacent);
-                        visitedNodes.add(adjacent);
-                    });
+            // Build a list of adjacent nodes.
+            for (Integer node : nodeBreakerView.getNodes()) {
+                if (!visitedNodes.contains(node)) {
+                    List<Integer> adjacentNodes = new ArrayList<>(1);
+                    computeAdjacentNodes(nodeBreakerView, List.of(node), adjacentNodes, visitedNodes, context);
```
flo-dup: the recursive method you wrote already exist in a generic way, that's the nodeBreakerView travers method:
```java
nodeBreakerView.traverse(node, (node1, sw, node2) -> {
    if (sw ...) {
       return TraverseResult.CONTINUE;
    }else {
       return TraverseResults.TERMINATE_PATH;
    }
}
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
```diff
@@ -1499,6 +1451,16 @@ private static void writeTerminals(Network network, Map<Terminal, String> mapTer
         }
     }
 
+    private static void fillSwitchNodeKeys(VoltageLevel vl, Switch sw, String[] nodeKeys, CgmesExportContext context) {
```
flo-dup: maybe put it back where you found it? :police_officer: 
Like buildNodeKey. Following the method history is then easier!

But that might make the code methods too scrambled... up to you!

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
rcourtier: done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
rcourtier: updated

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
rcourtier: done

Issue #3385, opened 2025-03-26
PSS/E errors with parsing PSS/E Files (Delimiters, Missing Q Record, Comments)


pjanecek0:
### Describe the current behavior

PowSyBl returns a parser error even for files that PSS/E reads. 
The following problems:

1. **Delimiters** - error occurs if both whitespace and spaces are used, such as: 
  0 100.000000 32, 0, 1, 50.000
  Also, if the first line contains a space before the comma, a parsing error will occur:
   0, 100.000000 , 32, 0, 1, 50.000**

2.   **Missing Q Record**: A file missing a `Q` record causes a parsing error in PowSyBl. However, PSS/E documentation states that the `Q` record is only indicative and not mandatory.

3. **Comments**: Comments starting with `//` are incorrectly removed during file loading, leading to errors.

### Describe the expected behavior

- PowSyBl should support rows with mixed delimiters (spaces and commas), including additional spaces before commas, as per PSS/E standards.
- A missing `Q` record should not trigger a parsing error, as it is optional according to the standards.
- Comments starting with `//` should be correctly removed without interfering with the rest of the file.

geofjamg:
> * PowSyBl should support rows with mixed delimiters (spaces and commas), including additional spaces before commas, as per PSS/E standards.

Do you have real data with such mixed delimiters? The only one I have ever seen is one of the standard IEEE_xxx.raw (I don't remember which one)

marqueslanauja:
My proposal is to implement the following changes to maximize the preservation of the existing code. Theses changes will fix all the issues.

Change the following main branch code

```
public String readRecordLine() throws IOException {
        String line = reader.readLine();
        if (line == null) {
            throw new PsseException("PSSE. Unexpected end of file");
        }
        if (isRecordLineDefiningTheAttributeFields(line)) {
            return ""; // an empty line must be returned
        }
        StringBuilder newLine = new StringBuilder();
        Matcher m = FileFormat.LEGACY_TEXT_QUOTED_OR_WHITESPACE.matcher(removeComment(line));
        while (m.find()) {
            // If current group is quoted, keep it as it is
            if (m.group().indexOf(LEGACY_TEXT.getQuote()) >= 0) {
                m.appendReplacement(newLine, replaceSpecialCharacters(m.group()));
            } else {
                // current group is whitespace, keep a single whitespace
                m.appendReplacement(newLine, " ");
            }
        }
        m.appendTail(newLine);
        return newLine.toString().trim();
[...]

PR #3386, opened 2025-03-26
PSS/E errors with parsing PSS/E Files (Delimiters, Missing Q Record, Comments)

pjanecek0:

#3385 

#3385 

- [ ] Yes
- [x] No

**Other information**:


pjanecek0:
@marqueslanauja Yes, that seems to solve the bug. Converting white spaces to commas is probably a better solution.

We could try replacing 

`1485,'O'Donnell ~1', 115.0000,1, 6, 1, 1,1.03539, 16.4735`

with

`1485, "O'Donnell ~1", 115.0000,1, 6, 1, 1,1.03539, 16.4735`

which I would expect to fix this test case and double quotes are also valid.However, we really don't have to support this...

For the tests, I need to add input
[IEEE_14_bus_delimiter.raw](https://github.com/powsybl/powsybl-core/pull/3386/files#diff-72a63e567944c4ad9b3f21635897242df53685a49b640eb7167b0ebca6fe55e5)
 that represents the data I will be using.

Using comma as an export delimiter is a better choice because it allows you to record the default value, and PSS/E does the same.

pjanecek0:
It was easier to create a new branch and a new pull request #3440 that solves this bug.

PR #3388, opened 2025-03-28
Fix: CreateVoltageLevelTopology should also work in non empty voltage levels

colinepiloquet:

No

Bug fix

The nodes used for the creation of new busbar sections in a voltage level start at 0.

The node 0 will usually be taken if the voltage level is not empty. Instead, the first available node will be used so that it is possible to add busbar sections in voltage levels that already contains busbar sections (from a diagram point of view, below or next to them). 

- [ ] Yes
- [X] No

### DIFF
--- a/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/CreateVoltageLevelTopology.java
+++ b/iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/CreateVoltageLevelTopology.java
@@ -183,7 +183,7 @@ public class CreateVoltageLevelTopology extends AbstractNetworkModification {
     }
 
     private void createBusbarSections(VoltageLevel voltageLevel, NamingStrategy namingStrategy) {
-        int node = 0;
+        int node = voltageLevel.getNodeBreakerView().getMaximumNodeIndex() + 1;
         for (int sectionNum = lowSectionIndex; sectionNum < lowSectionIndex + sectionCount; sectionNum++) {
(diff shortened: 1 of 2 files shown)


colinepiloquet:
I did not add documentation because this behavior is already the one described in the readthedocs...

rolnico (commented):
It works well, however I'm questioning the case where the user provides a position range where some BBS already exist. It still works, as the position extension does not forbid two BBS to have the same position, but should we not, at least, raise a warning? For example by adding such method before the creation of the new BBSs:

```java

    private void checkBusbarSectionWithSamePosition(VoltageLevel voltageLevel) {
        voltageLevel.getNodeBreakerView().getBusbarSectionStream().forEach(bbs -> {
            BusbarSectionPosition bbsPosition = bbs.getExtension(BusbarSectionPosition.class);
            if (bbsPosition != null
                && bbsPosition.getBusbarIndex() >= lowSectionIndex && bbsPosition.getBusbarIndex() < lowSectionIndex + sectionCount
                && bbsPosition.getSectionIndex() >= lowBusOrBusbarIndex && bbsPosition.getSectionIndex() < lowBusOrBusbarIndex + alignedBusesOrBusbarCount) {
                LOG.warn("Busbar section {} in the same voltage level has the same position ({}, {}) as a new busbar section. " +
                        "This can lead to unexpected behavior.", bbs.getId(), bbsPosition.getBusbarIndex(), bbsPosition.getSectionIndex());
            }
        });
    }
```

Or should we throw an exception?

rolnico (approved):
LGTM

rolnico (approved):

PR #3391, opened 2025-03-31
Add handling for possible null return to avoid NullPointerException in `AbstractEquipmentCriterionContingencyListDeserializer`

arthurscchan:
This is a proposed fix to stability issue discovered by OSS-Fuzz when fuzzing the powsybl-core module. The original OSS-Fuzz issue can be found in https://issues.oss-fuzz.com/u/1/issues/406830033.


arthurscchan:
A `NullPointerException` can occur in the `deserializeCommonAttributes` method of `AbstractEquipmentCriterionContingencyListDeserializer` class when deserializing JSON input with a `null` value for the `"type"` field.

```java
case "type" -> {
    if (!parser.nextTextValue().equals(expectedType)) {
        throw new IllegalStateException("type should be: " + expectedType);
    }
    return true;
}
```

The method `parser.nextTextValue()` may return `null` when the input JSON contains:

```json
{
  "type": null,
  ...
}
```

In such a case, the call to `.equals(expectedType)` becomes:

```java
null.equals("HvdcLineCriterionContingencyList")
```

This results in a `NullPointerException` because the `equals(...)` method is invoked on a `null` reference.

arthurscchan:
This is a stability issue due to a lack of validation for null checking from parsing of untrusted JSON. Here is a simple proof of concept to trigger the problem.

```java

import com.fasterxml.jackson.core.JsonFactory;
import com.fasterxml.jackson.core.JsonParser;
import com.fasterxml.jackson.databind.DeserializationContext;
import com.fasterxml.jackson.databind.ObjectMapper;
import com.fasterxml.jackson.databind.deser.DefaultDeserializationContext;
import com.powsybl.contingency.contingency.list.HvdcLineCriterionContingencyList;
import com.powsybl.contingency.json.HvdcLineCriterionContingencyListDeserializer;
import java.io.StringReader;

public class ProofOfConcept {
    public static void main(String[] args) throws Exception {
        String json = "{\"type\": null, \"name\": \"test-list\"}";

        JsonFactory factory = new JsonFactory();
        JsonParser parser = factory.createParser(new StringReader(json));
        ObjectMapper mapper = new ObjectMapper();
        DeserializationContext ctx = new DefaultDeserializationContext.Impl(
            mapper.getDeserializationContext().getFactory()
        );

        HvdcLineCriterionContingencyListDeserializer deserializer = new HvdcLineCriterionCon>
        parser.nextToken();
        deserializer.deserialize(parser, ctx);
    }
}
```

To execute and test the PoC, follow the steps below. It is assumed that OpenJDK 17.0.2 and Maven 3.9.9 is used.
```
# Prepare OpenJDK 17.0.2
export JAVA_HOME=./jdk-17.0.2

# Prepare Maven 3.9.9

# Build Powsybl-core
cd powsybl-core

# Group jar files
mkdir jar

# Build and run PoC
javac -cp "jar/*" ProofOfConcept.java
java -cp "jar/*:./" ProofOfConcept
```

```
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "String.equals(Object)" because the return value of "com.fasterxml.jackson.core.JsonParser.nextTextValue()" is null
        at com.powsybl.contingency.json.AbstractEquipmentCriterionContingencyListDeserializer.deserializeCommonAttributes(AbstractEquipmentCriterionContingencyListDeserializer.java:72)
        at com.powsybl.contingency.json.HvdcLineCriterionContingencyListDeserializer.lambda$deserialize$0(HvdcLineCriterionContingencyListDeserializer.java:32)
        at com.powsybl.commons.json.JsonUtil.parseObject(JsonUtil.java:517)
        at com.powsybl.commons.json.JsonUtil.parsePolymorphicObject(JsonUtil.java:495)
        at com.powsybl.contingency.json.HvdcLineCriterionContingencyListDeserializer.deserialize(HvdcLineCriterionContingencyListDeserializer.java:32)
        at ProofOfConcept.main(ProofOfConcept.java:23)
```

arthurscchan:
The root cause is down at the AbstractEquipmentCriterionContingencyListDeserializer ::deserializeCommonAttributes method. The fix is simply adding a null checking before calling to the `String::equals` method to avoid NPE from direct chain invocation.

contingency/contingency-api/src/main/java/com/powsybl/contingency/json/AbstractEquipmentCriterionContingencyListDeserializer.java
```diff
@@ -69,7 +70,9 @@ protected boolean deserializeCommonAttributes(JsonParser parser, Deserialization
                 return true;
             }
             case "type" -> {
-                if (!parser.nextTextValue().equals(expectedType)) {
+                // parser.nextTextValue() could returns null
+                String typeStr = Objects.requireNonNull(parser.nextTextValue());
```
olperr1, line 74: Thank you for your contribution!

Here, calling `Objects.requireNonNull(...)` will also throw a NPE if the type is null (but the cause will indeed be different). 
It may be better to reverse the test (`!expectedType.equals(typeStr)`) because `expectedType` is not an input. In this case, an `IllegalStateException` with a clear message will be thrown.

olperr1 (changes requested):

olperr1 (approved):

PR #3392, opened 2025-03-31
Fix NPE from Boolean auto-unboxing

arthurscchan:
This is a proposed fix to stability issue discovered by OSS-Fuzz when fuzzing the powsybl-core module. The original OSS-Fuzz issue can be found in https://issues.oss-fuzz.com/u/1/issues/406871272 and https://issues.oss-fuzz.com/u/1/issues/406999127.

### DIFF
--- a/sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityFactor.java
+++ b/sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityFactor.java
@@ -57,12 +57,12 @@ public class SensitivityFactor {
      * @param contingencyContext see {@link com.powsybl.contingency.ContingencyContext}
      */
     public SensitivityFactor(SensitivityFunctionType functionType, String functionId, SensitivityVariableType variableType,
-                              String variableId, boolean variableSet, ContingencyContext contingencyContext) {
+                              String variableId, Boolean variableSet, ContingencyContext contingencyContext) {
         this.functionType = Objects.requireNonNull(functionType);
         this.functionId = Objects.requireNonNull(functionId);
         this.variableType = Objects.requireNonNull(variableType);
         this.variableId = Objects.requireNonNull(variableId);
-        this.variableSet = variableSet;
+        this.variableSet = Objects.requireNonNull(variableSet);
         this.contingencyContext = Objects.requireNonNull(contingencyContext);
     }
 


arthurscchan:
The method `parseJson` uses an internal `ParsingContext` object to accumulate parsed values from a JSON object. One of the fields in `ParsingContext` is declared as a boxed `Boolean` with default value equals to `null`.

```java
static final class ParsingContext {
...
    Boolean variableSet;
...
}
```

The problem is, the same variable `variableSet` in the outer SensitivityFactor is defined as primitive boolean instead of Boolean class object. This create a auto-boxing/unboxing when transferring between these two varaibles.

```java
public class SensitivityFactory {
...
    private final boolean variableSet;
...
}
```

Later, the code **unconditionally unboxes** it when calling the constructor:

```java
return new SensitivityFactor(
    context.functionType,
    context.functionId,
    context.variableType,
    context.variableId,
    context.variableSet,
    new ContingencyContext(context.contingencyId, context.contingencyContextType)
);
```

If the input JSON is **missing the `variableSet` field**, `context.variableSet` will remain `null`, and unboxing this will cause the `NullPointerException`.

arthurscchan:
This is a stability issue caused by insufficient validation during the conversion between primitive and object variables. In many cases, object wrappers for primitive types accept a wider range of values than their primitive counterparts. For example, a `Boolean` object can be `null` in addition to `true` or `false`. This becomes problematic when a `null` `Boolean` object is auto-unboxed to a primitive `boolean` value, as the process internally calls the `booleanValue()` method on the `Boolean` object—resulting in a `NullPointerException` if the object is `null`.

Below is a Proof of Concept (PoC) demonstrating the issue. It calls the `SensitivityFactor::parseJson` method with a crafted JSON string that omits the `variableSet` key, forcing a `null` unboxing scenario.

```java
import com.fasterxml.jackson.core.JsonFactory;
import com.fasterxml.jackson.core.JsonParser;
import com.powsybl.sensitivity.SensitivityFactor;
import java.io.StringReader;

public class ProofOfConcept {
    public static void main(String[] args) throws Exception {
        String json = """
            {
              "functionType": "BUS_VOLTAGE",
              "functionId": "branch1",
              "variableType": "BUS_TARGET_VOLTAGE",
              "variableId": "gen1",
              "contingencyContextType": "NONE"
            }
            """;

        JsonFactory factory = new JsonFactory();
        JsonParser parser = factory.createParser(new StringReader(json));
        parser.nextToken();
        SensitivityFactor.parseJson(parser);
    }
}
```

To execute and test the PoC, follow the steps below. It is assumed that OpenJDK 17.0.2 and Maven 3.9.9 is used.
```
# Prepare OpenJDK 17.0.2
export JAVA_HOME=./jdk-17.0.2

# Prepare Maven 3.9.9

# Build Powsybl-core
cd powsybl-core

# Group jar files
mkdir jar

# Build and run PoC
javac -cp "jar/*" ProofOfConcept.java
java -cp "jar/*:./" ProofOfConcept
```

```
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "java.lang.Boolean.booleanValue()" because "context.variableSet" is null
        at com.powsybl.sensitivity.SensitivityFactor.parseJson(SensitivityFactor.java:160)
        at ProofOfConcept.main(ProofOfConcept.java:21)
```

arthurscchan:
The issue stems from the constructor of the `SensitivityFactor` class during the auto-unboxing process. The appropriate fix depends on whether the absence of `variableSet` (i.e., a `null` value) is considered valid in the JSON input.

Here are two proposed fixes for different situations.

**Fix 1 – Null Values Not Permitted:**  
- Assumes `variableSet` cannot be `null` (i.e., it must be explicitly provided in the JSON string).  
- Modifies the `SensitivityFactor` constructor to accept a `Boolean` object instead of a primitive `boolean`.  
- Enforces non-null values using `Objects.requireNonNull` to validate the parsed JSON.  
- https://github.com/arthurscchan/powsybl-core/commit/46f9bb926252d9be78cf77ccbf4994aff087f706 

**Fix 2 – Null Values Default to `false`:**  
- Treats missing `variableSet` in the JSON string as a valid case, defaulting to `false` when `null`.  
- Similarly updates the constructor to accept a `Boolean` object.  
- Uses a ternary operator instead of a null check.  
- https://github.com/arthurscchan/powsybl-core/commit/e391469d00b6dc9ef7368b208a914873d45881f4

**Fix 1 is used in this PR, we can change the fix to Fix 2 if needed.**

rolnico (changes requested):
Could you add your PoC as a test?

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityFactor.java
```diff
@@ -57,12 +57,12 @@ public class SensitivityFactor {
      * @param contingencyContext see {@link com.powsybl.contingency.ContingencyContext}
      */
     public SensitivityFactor(SensitivityFunctionType functionType, String functionId, SensitivityVariableType variableType,
-                              String variableId, boolean variableSet, ContingencyContext contingencyContext) {
+                              String variableId, Boolean variableSet, ContingencyContext contingencyContext) {
```
olperr1: I think it would be cleaner to keep the parameter's type to `boolean`. It defines a "contract": this parameter should not be null.
The problem here is that this method is called with an unchecked `Boolean` variable that will be unboxed. So it would be better to fix the caller which does not respect this contract.
Therefore, can you do the following?
- Keep the `boolean` type for this variable;
- Change `SensitivityFactor.parseJson(JsonParser parser)` to call the constructor with `Objects.requireNonNull(context.variableSet)` instead of `context.variableSet`.

olperr1 (changes requested):

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityFactor.java
arthurscchan: Yeah, that maybe a better fix.

olperr1 (approved):

rolnico (approved):

PR #3395, opened 2025-04-01
Add checking to avoid unexpected ArrayIndexOutOfBoundException when reading IEEE CDF files

arthurscchan:
This is a proposed fix to stability issue discovered by OSS-Fuzz when fuzzing the powsybl-core module. The original OSS-Fuzz issue can be found in https://issues.oss-fuzz.com/u/1/issues/406332771.


arthurscchan:
The `IeeeCdfReader.read(BufferedReader)` method in the `com.powsybl.ieeecdf.model` package assumes that the first line of the file (the title line) will always be valid and parsable into an `IeeeCdfTitle` bean. However, if the input file is malformed (e.g., empty or with an invalid format), this assumption fails silently.

https://github.com/powsybl/powsybl-core/blob/455fd74f3b9d03754ab8774b50d59e07823a793c/ieee-cdf/ieee-cdf-model/src/main/java/com/powsybl/ieeecdf/model/IeeeCdfReader.java#L34-L36

The problem arises from the behavior of the Univocity `FixedWidthParser`, inherited via `AbstractParser`, which **silently skips invalid or unparsable lines**. As a result:

- The call to `parseLine(null)` or `parseLine(<malformed>)` is skipped by the parser.
- `BeanListProcessor.getBeans()` returns an **empty list**.
- The `.get(0)` call throws an `IndexOutOfBoundsException`:

From the above understanding, this crash can be triggered easily by supplying an empty file or one with an invalid first line (malformed fixed-width format that doesn't match the expected bean schema for `IeeeCdfTitle`). This cause stability issue with unclear exception message.

arthurscchan:
The root cause is down at the `IeeeCdfReader::read method`. The fix is simply adding a empty checking before calling to the get method to avoid malformed input crash the execution with `ArrayIndexOutOfBoundException`.

ieee-cdf/ieee-cdf-model/src/main/java/com/powsybl/ieeecdf/model/IeeeCdfReader.java
```diff
@@ -33,7 +33,13 @@ enum IeeeCdfSection {
...
         String line = reader.readLine();
 
-        IeeeCdfTitle title = parseLines(Collections.singletonList(line), IeeeCdfTitle.class).get(0);
+        // Ensure malformed input does not trigger unexpected ArrayIndexOutOfBoundException
+        List<IeeeCdfTitle> parsedLines = parseLines(Collections.singletonList(line), IeeeCdfTitle.class);
+        if (parsedLines.size() == 0) {
+            throw new IllegalArgumentException("Failed to parse the IeeeCdfModel");
+        }
+
+        IeeeCdfTitle title = parsedLines.get(0);
```
olperr1, line 42: Can you add the following test to `IeeeCdfReaderWriterTest`?

```java
    @Test
    void testInvalidFile() throws IOException {
        try (BufferedReader reader = new BufferedReader(new StringReader(""))) {
            IeeeCdfReader r = new IeeeCdfReader();
            IllegalArgumentException e = assertThrows(IllegalArgumentException.class, () -> r.read(reader));
            assertEquals("Failed to parse the IeeeCdfModel", e.getMessage());
        }
    }
```

olperr1 (changes requested):

olperr1 (changes requested):
Some imports are missing in `IeeeCdfReaderWriterTest`. Can you add them?

olperr1 (approved):

PR #3405, opened 2025-04-14
Bump IIDM to 1.14

olperr1:
**Please check if the PR fulfills these requirements**

- [X] The commit message follows our guidelines
- [X] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] A PR or issue has been opened in all impacted repositories (if any)

**Does this PR already have an issue describing the problem?**

No

**What kind of change does this PR introduce?**

IIDM version update

**What is the current behavior?**

Current IIDM version is 1.13.

**What is the new behavior (if this is a feature change)?**
Current IIDM version is **1.14**.

**Does this PR introduce a breaking change or deprecate an API?**
- [X] Yes
- [ ] No

**If yes, please check if the following requirements are fulfilled**

- [X] The *Breaking Change* or *Deprecated* label has been added
- [X] The migration steps are described in the following section

**What changes might users need to make in their application due to this PR? (migration steps)**

Downstream projects should:
- Update their classes extending `AbstractVersionableNetworkExtensionSerDe` to add an entry for the new version (1.14) in the first map of the call to `super(...)`;
- Upgrade the version in their XIIDM / JIIDM unit test files corresponding to the current version:
  - For ".xiidm" or ".xml" files: replace `xmlns:iidm="http://www.powsybl.org/schema/iidm/1_13"` with `xmlns:iidm="http://www.powsybl.org/schema/iidm/1_14"`;
  - For ".json" or ".jiidm" files: replace `"version" : "1.13"` with `"version" : "1.14"`.

**Other information**:

PR #3423, opened 2025-04-24
Support properties on OperationalLimitsGroup

### LINKED ISSUE #3420: OperationalLimitsGroup can be tagged

`OperationalLimitGroup` is not an `Identifiable` and so cannot have properties attached to it.

`OperationalLimitGroup` can have properties even though it is not an `Identifiable`.

### Describe the motivation

Be able to attach specific data to an `OperationalLimitGroup`, with business properties for instance. These attached properties are optional.

### Extra Information

We can create an abstract class extended by `OperationalLimitGroup` that allows to add properties to it. (Like `AbstractPropertyBag` class in powsybl-open-loadflow). This abstract could be extended in the future by more than only `OperationalLimitGroup` if other IIDM objects need it and are not `Identifiable` (so they do not have the `Properties properties` attribute).

olperr1:
:warning: **Draft status**: should not be merged in `bump_iidm_to_1_14` but in `xiidm_version_1_14` after #3405 is merged.

</br>
</br>

Fixes #3420 

- [X] Yes
- [ ] No

Custom IIDM implementations' maintainers should implement the following methods in their `OperationalLimitsGroup` implementations:
- `public boolean hasProperty()`
- `public boolean hasProperty(String key)`
- `public String getProperty(String key)`
- `public String getProperty(String key, String defaultValue)`
- `public String setProperty(String key, String value)`
- `public boolean removeProperty(String key)`
- `public Set<String> getPropertyNames()`

**Other information**:


iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractOperationalLimitsGroupsTest.java
```diff
@@ -491,4 +493,34 @@ private static void checkBehaveLikeNoLimitsWhenNoDefaultOneOnTWT(ThreeWindingsTr
...
+    public void propertiesTests() {
+        Network network = createNetworkWithOperationalLimitsGroupsOnLine();
+        Line l = network.getLine("L");
+        OperationalLimitsGroup group = l.getOperationalLimitsGroup1("1").orElseThrow();
+        assertFalse(group.hasProperty());
+        String property1 = "property_1";
+        String property2 = "property_2";
+
+        group.setProperty(property1, "A");
+        assertAll(
```
rolnico, line 507: Why using `assertAll` here and l520?

olperr1: With `assertAll`, it displays all the failing assertions instead of stopping at the first failure.
The result is the same, if one assertion fails, the test fails. But in that case, it offers more information.

rolnico (approved):
Code duplication seems impossible to avoid

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/PropertiesBearer.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+package com.powsybl.iidm.network;
+
+import java.util.Set;
+
+/**
+ * @author Olivier Perrin {@literal <olivier.perrin at rte-france.com>}
+ */
+public interface PropertiesBearer {
```
flo-dup: We rather use Holder in similar interfaces, plus it's a key word commonly used in Java for these intefaces, whereas I hadn't seen a Bearer class so far.
```suggestion
public interface PropertiesHolder {
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/PropertiesContainer.java
```diff
@@ -0,0 +1,86 @@
...
+
+    public String setProperty(String key, String value) {
+        String oldValue = (String) properties.put(key, value);
+        if (bearer instanceof Identifiable<?> identifiable) {
+            if (Objects.isNull(oldValue)) {
+                getNetwork().getListeners().notifyPropertyAdded(identifiable, () -> getPropertyStringForNotification(key), value);
+            } else {
+                getNetwork().getListeners().notifyPropertyReplaced(identifiable, () -> getPropertyStringForNotification(key), oldValue, value);
+            }
+        }
```
flo-dup: All this kind of logic is specific to this kind of bearer, hence this should rather be inside the AbstractIdentifiable. Which allows to get rid of the bearer field.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractNonIdentifiablePropertiesBearer.java
```diff
@@ -0,0 +1,60 @@
...
+
+import com.powsybl.iidm.network.PropertiesBearer;
+
+import java.util.Properties;
+import java.util.Set;
+
+/**
+ * @author Olivier Perrin {@literal <olivier.perrin at rte-france.com>}
+ */
+public abstract class AbstractNonIdentifiablePropertiesBearer implements PropertiesBearer {
```
flo-dup: There's no other class than OperationalLimitsGroups so far, maybe we should introduce it later? Or if we want to keep it, we should rather name it simply `AbstractPropertiesHolder`. With such a name it's already expected that it's non identifiable (and non extendable!).

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/PropertiesContainer.java
```diff
@@ -0,0 +1,86 @@
...
+        return oldValue;
+    }
+
+    public boolean removeProperty(String key) {
+        Object oldValue = properties.remove(key);
+        if (oldValue != null) {
+            if (bearer instanceof Identifiable<?> identifiable) {
+                getNetwork().getListeners().notifyPropertyRemoved(identifiable, () -> getPropertyStringForNotification(key), oldValue);
```
flo-dup: Shouldn't we warn the user in the documentation that he won't get notified in case of an OperationalLimitsGroups property change?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiable.java
```diff
@@ -208,17 +205,15 @@ public String setProperty(String key, String value) {
...
-        Object oldValue = properties.remove(key);
-        if (oldValue != null) {
-            getNetwork().getListeners().notifyPropertyRemoved(this, () -> getPropertyStringForNotification(key), oldValue);
-            return true;
-        }
-        return false;
+        String oldValue = getProperty(key);
+        boolean removed = properties.removeProperty(key);
+        getNetwork().getListeners().notifyPropertyRemoved(this, () -> getPropertyStringForNotification(key), oldValue);
+        return removed;
```
rolnico: you now notify even when the property is not present, I'm not sure you really want that

rolnico (changes requested):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/PropertiesContainer.java
```diff
@@ -0,0 +1,55 @@
...
+    public String getProperty(String key, String defaultValue) {
+        Object val = properties.getOrDefault(key, defaultValue);
+        return val != null ? val.toString() : null;
+    }
+
+    public String setProperty(String key, String value) {
+        return (String) properties.put(key, value);
+    }
+
+    public boolean removeProperty(String key) {
```
flo-dup: You could change the signature here like to return the old value and avoid getting it twice (in AbstractIdentifiable):
```suggestion
    public String removeProperty(String key) {
```

rolnico (approved):

PR #3435, opened 2025-05-05
Move test on VariantManager from AbstractSwitchSetRetainedTest to VariantManagerImplTest

antoinebhs:

No

Test improvement

The test AbstractSwitchSetRetainedTest.test() requires a multithread access implementation which is not necessary in this test. 

Also it's an issue in the network-store where multithread access is not yet implemented and this test can't be executed: https://github.com/powsybl/powsybl-network-store/pull/521/files/351111f3c27aa4cc3f0d7f3388e477e78900111c#diff-d65207657a584037066dde551d7a277330a0eca13e6fd8eb639d35b8496eeaaa.

The test AbstractSwitchSetRetainedTest.test() does not require multithread access implementation and we only test the retain part, not the variant manager part. The variantManager part was moved to a new test in variantManager tests without requiring multithread access implementation.

- [ ] Yes
- [x] No

### DIFF
--- a/contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyBuilder.java
+++ b/contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyBuilder.java
@@ -145,7 +145,7 @@ public class ContingencyBuilder {
     }
 
     public ContingencyBuilder addIdentifiable(Identifiable<?> identifiable) {
-        elements.add(ContingencyElement.of(identifiable));
+        elements.add(ContingencyElementFactory.create(identifiable));
         return this;
     }
 
--- a/contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyElement.java
+++ b/contingency/contingency-api/src/main/java/com/powsybl/contingency/ContingencyElement.java
@@ -7,9 +7,7 @@
  */
 package com.powsybl.contingency;
 
-import com.powsybl.commons.PowsyblException;
 import com.powsybl.iidm.modification.tripping.Tripping;
-import com.powsybl.iidm.network.*;
 
 /**
  * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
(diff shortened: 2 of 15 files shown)


rolnico (changes requested):
I would completely remove the tests on `VariantManager` here, but also make sure that this test is still covered somewhere (probably in `VariantManagerTest` or something like this).

antoinebhs:
> I would completely remove the tests on `VariantManager` here, but also make sure that this test is still covered somewhere (probably in `VariantManagerTest` or something like this).

The original test is linked to https://github.com/powsybl/powsybl-core/pull/133. The objective was to test if the retained property of switches is stateful. So I'm not confident in removing the variantManager part as it's part of the regression test.

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/VariantManagerImplTest.java
```diff
@@ -435,4 +435,36 @@ void testVariantIndexSwitch() throws Exception {
...
+        Switch b1 = vl.getNodeBreakerView().newBreaker()
+                .setId("B1")
+                .setNode1(0)
+                .setNode2(1)
+                .setOpen(false)
+                .setRetained(true)
+                .add();
+
+        VariantManager variantManager = network.getVariantManager();
+        variantManager.cloneVariant(VariantManagerConstants.INITIAL_VARIANT_ID, "backup");
```
rolnico, line 460: I would keep the multithread access since the test is related to the implementation

```suggestion
        VariantManager variantManager = network.getVariantManager();
        variantManager.allowVariantMultiThreadAccess(true);
        variantManager.cloneVariant(VariantManagerConstants.INITIAL_VARIANT_ID, "backup");
```

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/VariantManagerImplTest.java
```diff
@@ -435,4 +435,36 @@ void testVariantIndexSwitch() throws Exception {
...
+
+        assertTrue(b1.isRetained());
```
rolnico, line 462: Same thing, you can add again the assertEquals?

rolnico (changes requested):

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/VariantManagerImplTest.java
antoinebhs: I'm not sure we ever wanted to test multiThreadAccess here. I think we only wanted to test that the retained property can be set to a different value on different variant. I added it back too as asked.

rolnico (approved):
LGTM

rolnico (approved):

Issue #3418, opened 2025-04-21
Add new attributes to store solved values for shunt compensators, switch and tap changers


So-Fras:
### Describe the expected behavior

The resulting tap positions, switch positions and section counts should be stored in other variables that would be called:

solvedPosition for tap changers?
solvedSection for compensators?
solvedStatus for switches?

geofjamg:
solvedOpen to keep consistency. Status would require an enum not a boolean.

PR #3441, opened 2025-05-14
iIDM v1.14: Solved values - Add solved section count

colinepiloquet:

Fixes (partially) #3418
 

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
@@ -123,6 +123,19 @@ public interface ShuntCompensator extends Injection<ShuntCompensator> {
     int getSectionCount();
 
     /**
+     * Get the solved count of sections in service.
+     * Please note that sections can only be sequentially in service, i.e., the first sectionCount sections are in service.
+     * <p>
+     * It is expected to be greater than one and lesser than or equal to the
+     * maximum section count.
+     * It can be null if no calculation has been launched on the network.
+     * <p>
+     * Depends on the working variant.
+     * @see VariantManager
+     */
+    Integer getSolvedSectionCount();
+
+    /**
      * Get the count of sections in service if it is defined.
      * Otherwise, get an empty optional.
      * <p>
@@ -157,6 +170,27 @@ public interface ShuntCompensator extends Injection<ShuntCompensator> {
     default ShuntCompensator unsetSectionCount() {
         throw ValidationUtil.createUnsetMethodException();
     }
+
+    /**
+     * Change the solved count of sections in service.
+     * Please note sections can only be sequentially in service, i.e., the first sectionCount sections are in service.
+     * <p>
+     * Depends on the working variant.
+     *
+     * @see VariantManager
+     * @param solvedSectionCount the number of solved sections in service
+     * @return the shunt compensator to chain method calls.
+     */
+    ShuntCompensator setSolvedSectionCount(int solvedSectionCount);
+
+    /**
+     * Unset the solved count of sections in service.
+     * Note: this can be done <b>only</b> in SCADA validation level.
+     */
+    default ShuntCompensator unsetSolvedSectionCount() {
+        throw ValidationUtil.createUnsetMethodException();
+    }
+
 
     /**
      * Get the susceptance (in S) of the shunt in its current state i.e. the sum of the sections' susceptances for all sections in service.
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
@@ -23,6 +23,8 @@ class ShuntCompensatorAdderImpl extends AbstractInjectionAdder<ShuntCompensatorA
 
     private Integer sectionCount;
 
+    private Integer solvedSectionCount;
+
     private double targetV = Double.NaN;
 
     private double targetDeadband = Double.NaN;
@@ -178,6 +180,12 @@ class ShuntCompensatorAdderImpl extends AbstractInjectionAdder<ShuntCompensatorA
     }
 
     @Override
+    public ShuntCompensatorAdder setSolvedSectionCount(int solvedSectionCount) {
+        this.solvedSectionCount = solvedSectionCount;
+        return this;
+    }
+
+    @Override
     public ShuntCompensatorAdder setRegulatingTerminal(Terminal regulatingTerminal) {
         this.regulatingTerminal = (TerminalExt) regulatingTerminal;
         return this;
@@ -220,7 +228,7 @@ class ShuntCompensatorAdderImpl extends AbstractInjectionAdder<ShuntCompensatorA
                 network.getMinValidationLevel(), network.getReportNodeContext().getReportNode()));
 
         ShuntCompensatorImpl shunt = new ShuntCompensatorImpl(getNetworkRef(),
-                id, getName(), isFictitious(), modelBuilder.build(), sectionCount, regulatingTerminal,
+                id, getName(), isFictitious(), modelBuilder.build(), sectionCount, solvedSectionCount, regulatingTerminal,
                 voltageRegulatorOn, targetV, targetDeadband);
 
         shunt.addTerminal(terminal);
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorImpl.java
@@ -32,6 +32,9 @@ class ShuntCompensatorImpl extends AbstractConnectable<ShuntCompensator> impleme
     /* the current number of section switched on */
     private final ArrayList<Integer> sectionCount;
 
+    /* the solved number of section switched on */
+    private final ArrayList<Integer> solvedSectionCount;
+
     /* the target voltage value */
     private final TDoubleArrayList targetV;
 
@@ -40,8 +43,8 @@ class ShuntCompensatorImpl extends AbstractConnectable<ShuntCompensator> impleme
 
     ShuntCompensatorImpl(Ref<NetworkImpl> network,
                          String id, String name, boolean fictitious, ShuntCompensatorModelExt model,
-                         Integer sectionCount, TerminalExt regulatingTerminal, Boolean voltageRegulatorOn,
-                         double targetV, double targetDeadband) {
+                         Integer sectionCount, Integer solvedSectionCount, TerminalExt regulatingTerminal,
+                         Boolean voltageRegulatorOn, double targetV, double targetDeadband) {
         super(network, id, name, fictitious);
         this.network = network;
         int variantArraySize = this.network.get().getVariantManager().getVariantArraySize();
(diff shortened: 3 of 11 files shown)


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorAdderImpl.java
```diff
@@ -220,7 +228,7 @@ public ShuntCompensatorImpl add() {
                 network.getMinValidationLevel(), network.getReportNodeContext().getReportNode()));
 
         ShuntCompensatorImpl shunt = new ShuntCompensatorImpl(getNetworkRef(),
-                id, getName(), isFictitious(), modelBuilder.build(), sectionCount, regulatingTerminal,
+                id, getName(), isFictitious(), modelBuilder.build(), sectionCount, solvedSectionCount, regulatingTerminal,
```
colinepiloquet, line 231: I don't know if there should be a check of the `solvedSectionCount` here to be sure it is not higher than the maximum? Or somewhere else?

rolnico: Since there is nothing changing the network in the present method before this line, I would put the check in the constructor `ShuntCompensatorImpl`. Doing so, the check could be a private method also used in the new `setSolvedSectionCount(int solvedSectionCount)` method.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorImpl.java
```diff
@@ -111,6 +121,25 @@ public ShuntCompensator unsetSectionCount() {
...
+        int variantIndex = n.getVariantIndex();
+        Integer oldValue = this.solvedSectionCount.set(variantIndex, checkSolvedSectionCount(solvedSectionCount, this.model.getMaximumSectionCount()));
+        String variantId = n.getVariantManager().getVariantId(variantIndex);
+        notifyUpdate("solvedSectionCount", variantId, oldValue, solvedSectionCount);
+        return this;
+    }
+
+    @Override
+    public ShuntCompensator unsetSolvedSectionCount() {
+        int variantIndex = network.get().getVariantIndex();
```
olperr1: You should use `NetworkImpl n = getNetwork();` to get the network.
This checks that the equipment was not removed.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorImpl.java
```diff
@@ -50,10 +53,12 @@ class ShuntCompensatorImpl extends AbstractConnectable<ShuntCompensator> impleme
         this.sectionCount = new ArrayList<>(variantArraySize);
         this.targetV = new TDoubleArrayList(variantArraySize);
         this.targetDeadband = new TDoubleArrayList(variantArraySize);
+        this.solvedSectionCount = new ArrayList<>(variantArraySize);
```
olperr1: This is obviously not critical, but can you move this line just after the "sectionCount`'s one?
This will improve readability.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorImpl.java
```diff
@@ -50,10 +53,12 @@ class ShuntCompensatorImpl extends AbstractConnectable<ShuntCompensator> impleme
...
         for (int i = 0; i < variantArraySize; i++) {
             this.sectionCount.add(sectionCount);
             this.targetV.add(targetV);
             this.targetDeadband.add(targetDeadband);
+            this.solvedSectionCount.add(checkSolvedSectionCount(solvedSectionCount, model.getMaximumSectionCount()));
```
olperr1: Same with this line.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/ShuntCompensatorXmlTest.java
```diff
@@ -87,6 +87,17 @@ void nullBPerSection() {
...
 
+    @Test
+    void solvedSectionCount() throws IOException {
+        Network network = ShuntTestCaseFactory.create();
+        network.getShuntCompensator("SHUNT").setSolvedSectionCount(1);
+
+        allFormatsRoundTripTest(network, "shuntWithSolvedSectionCountRoundTripRef.xml", CURRENT_IIDM_VERSION);
+
+        // backward compatibility
+        allFormatsRoundTripFromVersionedXmlFromMinToCurrentVersionTest("shuntRoundTripRef.xml", IidmVersion.V_1_14);
```
olperr1:
```suggestion
        allFormatsRoundTripFromVersionedXmlFromMinToCurrentVersionTest("shuntWithSolvedSectionCountRoundTripRef.xml", IidmVersion.V_1_14);
```

With `"shuntRoundTripRef.xml"`, the test was already done in `linearShuntTest()`.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorImpl.java
```diff
@@ -275,4 +310,11 @@ public void allocateVariantArrayElement(int[] indexes, final int sourceIndex) {
     protected String getTypeDescription() {
         return "Shunt compensator";
     }
+
+    private Integer checkSolvedSectionCount(Integer solvedSectionCount, int maximumSectionCount) {
+        if (solvedSectionCount != null && (solvedSectionCount < 0 || solvedSectionCount > maximumSectionCount)) {
```
olperr1, line 316: There is an inconsistency: the javadoc of `getSolvedSectionCount()` says that `solvedSectionCount` should be greater than 1.
Note that it is also the case for `sectionCount` which can be set to 0 (see `setSectionCount()`) although `getSectionCount()'s javadoc says it should be greater than 1 too.

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractShuntCompensatorTest.java
```diff
@@ -521,6 +523,43 @@ public void testSetTerminalP() {
...
+        adder.setSolvedSectionCount(5);
+        ShuntCompensator shuntCompensator = adder.add();
+        assertEquals(5, network.getShuntCompensator(SHUNT).getSolvedSectionCount());
+        shuntCompensator.setSolvedSectionCount(6);
+        assertEquals(6, shuntCompensator.getSolvedSectionCount());
+        assertEquals(Optional.of(6), shuntCompensator.findSolvedSectionCount());
+
+        // Check exception if solved section count negative or too high
+        assertThrows(ValidationException.class, () -> shuntCompensator.setSolvedSectionCount(-1));
+        assertThrows(ValidationException.class, () -> shuntCompensator.setSolvedSectionCount(50));
```
olperr1: Can you add assertions with the limit values?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
```diff
@@ -133,6 +135,28 @@ default OptionalInt findSectionCount() {
...
+    Integer getSolvedSectionCount();
+
+    /**
+     * Get the solved count of sections in service if it is defined.
+     * Otherwise, get an empty optional.
+     * <p>
+     * Depends on the working variant.
+     * @see VariantManager
+     */
+    default Optional<Integer> findSolvedSectionCount() {
```
olperr1: What is the reason to use `Optional<Integer>` instead of `OptionalInt` as for `findSectionCount()`?
I know `sectionCount` is an int and `solvedSectionCount` is an Integer. But in both cases, when `isEmpty()` is false on the result, you have a non-null value.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/ShuntCompensatorXmlTest.java
olperr1: With this suggestion, there's no need to keep the line 95:
```java
 allFormatsRoundTripTest(network, "shuntWithSolvedSectionCountRoundTripRef.xml", CURRENT_IIDM_VERSION);
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorImpl.java
colineplqt: I guess the javadoc is incorrect but I am not sure

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ShuntCompensator.java
colineplqt: I will change it

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ShuntCompensatorImpl.java
olperr1: It's not clear...
[The documentation](https://powsybl.readthedocs.io/projects/powsybl-core/en/stable/grid_model/network_subnetwork.html#section) states that:

> `B` and `G` attributes can be equal zero, but the disconnected status of the non-linear shunt compensator can be modeled by setting the `sectionCount` attribute to zero. The section which `sectionCount` equal to is the first effective section, and it would be more efficient to affect it a non-zero susceptance.

This could explain why there are `maximumSectionCount + 1` possible values (with values outside `[0 ; maxSectionCount]`, `setSectionCount(...)` throws an exception).

But the `setSectionCount(...)` javadoc tells that this method sets "the count of sections **in service**".
So maybe `0` value should be rejected too?

jeandemanged: jumping in...

the javadoc is indeed wrong. For maximumSectionCount=2 valid values are 0, 1, 2.

olperr1: Thanks @jeandemanged!
The Javadoc should be updated, then.

olperr1: @colinepiloquet: can you also update the Javadoc of `getSectionCount()`?

colineplqt: Done! Thanks

olperr1 (approved):

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ShuntSerDe.java
```diff
@@ -71,6 +71,7 @@ protected void writeRootElementAttributes(ShuntCompensator sc, VoltageLevel vl,
             OptionalInt sectionCount = sc.findSectionCount();
             context.getWriter().writeOptionalIntAttribute("sectionCount", sectionCount.isPresent() ? sectionCount.getAsInt() : null);
```
olperr1: We missed something here:
For IIDM version < 1.14, if the **solved** section count is present, it should be exported in the `"sectionCount"` field (instead of the section count), for backward compatibility.

colineplqt: Do we have the same understanding that before IIDM 1.3, this field was currentSectionCount?

olperr1: Yes, that's it!

colineplqt: Thank you

olperr1 (approved):

olperr1 (approved):

PR #3442, opened 2025-05-14
iIDM v1.14: Solved values - Add solved tap position

colinepiloquet:

Fixes (partially) #3418 

- [X] Yes
- [ ] No

Custom IIDM implementation maintainers should implement the following methods:
- in their `TapChanger`'s implementations:
  - `Integer getSolvedTapPosition()`
  - `C setSolvedTapPosition(int solvedTapPosition)`
  - `C unsetSolvedTapPosition()`
  - `S getSolvedCurrentStep();`
- in their `TapChangerAdder`'s implementations:
  - `S setSolvedTapPosition(Integer solvedTapPosition)`

**Other information**:

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChanger.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChanger.java
@@ -172,7 +172,6 @@ abstract class AbstractTapChanger<H extends TapChangerParent, C extends Abstract
 
     public C unsetSolvedTapPosition() {
         NetworkImpl n = getNetwork();
-        ValidationUtil.throwExceptionOrLogError(parent, "solved tap position has been unset", n.getMinValidationLevel(), n.getReportNodeContext().getReportNode());
         int variantIndex = network.get().getVariantIndex();
         Integer oldValue = this.solvedTapPosition.set(variantIndex, null);
         String variantId = network.get().getVariantManager().getVariantId(variantIndex);
@@ -199,8 +198,9 @@ abstract class AbstractTapChanger<H extends TapChangerParent, C extends Abstract
             throwIncorrectTapPosition(getTapPosition(), newHighTapPosition);
         }
         // We check if the solved tap position is still correct
-        if (getSolvedTapPosition() != null && getSolvedTapPosition() > newHighTapPosition) {
-            throwIncorrectSolvedTapPosition(getTapPosition(), newHighTapPosition);
+        OptionalInt solvedTapPosition = findSolvedTapPosition();
+        if (solvedTapPosition.isPresent() && solvedTapPosition.getAsInt() > newHighTapPosition) {
+            throwIncorrectSolvedTapPosition(solvedTapPosition.getAsInt(), newHighTapPosition);
         }
 
         this.steps = steps;
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChanger.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChanger.java
@@ -45,10 +45,12 @@ abstract class AbstractTapChanger<H extends TapChangerParent, C extends Abstract
 
     protected final TDoubleArrayList targetDeadband;
 
+    protected final ArrayList<Integer> solvedTapPosition;
+
     protected AbstractTapChanger(H parent,
                                  int lowTapPosition, List<S> steps, TerminalExt regulationTerminal,
                                  boolean loadTapChangingCapabilities,
-                                 Integer tapPosition, boolean regulating, double targetDeadband, String type) {
+                                 Integer tapPosition, boolean regulating, double targetDeadband, String type, Integer solvedTapPosition) {
         // The Ref object should be the one corresponding to the subnetwork of the tap changer holder
         // (to avoid errors when the subnetwork is detached)
         this.network = parent.getParentNetwork().getRootNetworkRef();
@@ -62,9 +64,11 @@ abstract class AbstractTapChanger<H extends TapChangerParent, C extends Abstract
         regulatingPoint.setRegulatingTerminal(regulationTerminal);
         this.tapPosition = new ArrayList<>(variantArraySize);
         this.targetDeadband = new TDoubleArrayList(variantArraySize);
+        this.solvedTapPosition = new ArrayList<>(variantArraySize);
         for (int i = 0; i < variantArraySize; i++) {
             this.tapPosition.add(tapPosition);
             this.targetDeadband.add(targetDeadband);
+            this.solvedTapPosition.add(solvedTapPosition);
         }
         this.type = Objects.requireNonNull(type);
         relativeNeutralPosition = getRelativeNeutralPosition();
@@ -113,6 +117,15 @@ abstract class AbstractTapChanger<H extends TapChangerParent, C extends Abstract
         return position == null ? OptionalInt.empty() : OptionalInt.of(position);
     }
 
+    public Integer getSolvedTapPosition() {
+        return solvedTapPosition.get(network.get().getVariantIndex());
+    }
+
+    public OptionalInt findSolvedTapPosition() {
+        Integer solvedPosition = solvedTapPosition.get(network.get().getVariantIndex());
+        return solvedPosition == null ? OptionalInt.empty() : OptionalInt.of(solvedPosition);
+    }
+
     public OptionalInt getNeutralPosition() {
         return relativeNeutralPosition != null ? OptionalInt.of(lowTapPosition + relativeNeutralPosition) : OptionalInt.empty();
     }
@@ -144,6 +157,29 @@ abstract class AbstractTapChanger<H extends TapChangerParent, C extends Abstract
         return (C) this;
     }
 
+    public C setSolvedTapPosition(int solvedTapPosition) {
+        NetworkImpl n = getNetwork();
+        if (solvedTapPosition < lowTapPosition
+            || solvedTapPosition > getHighTapPosition()) {
+            throwIncorrectSolvedTapPosition(solvedTapPosition, getHighTapPosition());
+        }
+        int variantIndex = n.getVariantIndex();
+        Integer oldValue = this.solvedTapPosition.set(variantIndex, solvedTapPosition);
+        String variantId = n.getVariantManager().getVariantId(variantIndex);
+        parent.getNetwork().getListeners().notifyUpdate(parent.getTransformer(), () -> getTapChangerAttribute() + ".solvedTapPosition", variantId, oldValue, solvedTapPosition);
+        return (C) this;
+    }
+
+    public C unsetSolvedTapPosition() {
+        NetworkImpl n = getNetwork();
+        ValidationUtil.throwExceptionOrLogError(parent, "solved tap position has been unset", n.getMinValidationLevel(), n.getReportNodeContext().getReportNode());
+        int variantIndex = network.get().getVariantIndex();
+        Integer oldValue = this.solvedTapPosition.set(variantIndex, null);
+        String variantId = network.get().getVariantManager().getVariantId(variantIndex);
+        n.getListeners().notifyUpdate(parent.getTransformer(), () -> getTapChangerAttribute() + ".solvedTapPosition", variantId, oldValue, null);
+        return (C) this;
+    }
+
     public S getStep(int tapPosition) {
         if (tapPosition < lowTapPosition || tapPosition > getHighTapPosition()) {
             throwIncorrectTapPosition(tapPosition, getHighTapPosition());
@@ -161,6 +197,10 @@ abstract class AbstractTapChanger<H extends TapChangerParent, C extends Abstract
         int newHighTapPosition = lowTapPosition + steps.size() - 1;
         if (getTapPosition() > newHighTapPosition) {
             throwIncorrectTapPosition(getTapPosition(), newHighTapPosition);
+        }
+        // We check if the solved tap position is still correct
+        if (getSolvedTapPosition() != null && getSolvedTapPosition() > newHighTapPosition) {
+            throwIncorrectSolvedTapPosition(getTapPosition(), newHighTapPosition);
         }
 
         this.steps = steps;
@@ -225,9 +265,11 @@ abstract class AbstractTapChanger<H extends TapChangerParent, C extends Abstract
     public void extendVariantArraySize(int initVariantArraySize, int number, int sourceIndex) {
         targetDeadband.ensureCapacity(targetDeadband.size() + number);
         tapPosition.ensureCapacity(tapPosition.size() + number);
+        solvedTapPosition.ensureCapacity(solvedTapPosition.size() + number);
         for (int i = 0; i < number; i++) {
             tapPosition.add(tapPosition.get(sourceIndex));
             targetDeadband.add(targetDeadband.get(sourceIndex));
+            solvedTapPosition.add(solvedTapPosition.get(sourceIndex));
         }
         regulatingPoint.extendVariantArraySize(initVariantArraySize, number, sourceIndex);
     }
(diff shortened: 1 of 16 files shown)


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChangerAdderImpl.java
```diff
@@ -95,12 +101,18 @@ public T add() {
                 network.setValidationLevelIfGreaterThan(ValidationLevel.EQUIPMENT);
             }
         }
+        if (solvedTapPosition != null) {
+            int highTapPosition = lowTapPosition + steps.size() - 1;
+            if (solvedTapPosition < lowTapPosition || solvedTapPosition > highTapPosition) {
+                LOGGER.warn("{} has incorrect solved tap position {}. It is not within bounds [{}, {}].", parent, solvedTapPosition, lowTapPosition, highTapPosition);
```
colinepiloquet, line 107: I am not sure about this behaviour: 
if the solved tap position is wrong when we create the tap changer, what should happen? The validation level of the network should stay the same but do we want to allow the creation of the solved tap position? And to what should it be set? null?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChangerAdderImpl.java
```diff
@@ -86,21 +92,23 @@ public T add() {
...
+        int highTapPosition = lowTapPosition + steps.size() - 1;
+        if (tapPosition != null && (tapPosition < lowTapPosition || tapPosition > highTapPosition)) {
+            ValidationUtil.throwExceptionOrLogError(parent, "incorrect tap position "
+                + tapPosition + " [" + lowTapPosition + ", "
+                + highTapPosition + "]", network.getMinValidationLevel(), network.getReportNodeContext().getReportNode());
+            network.setValidationLevelIfGreaterThan(ValidationLevel.EQUIPMENT);
+        }
+
+        if (solvedTapPosition != null && (solvedTapPosition < lowTapPosition || solvedTapPosition > highTapPosition)) {
+            LOGGER.warn("{} has incorrect solved tap position {}. It is not within bounds [{}, {}].", parent, solvedTapPosition, lowTapPosition, highTapPosition);
```
rolnico, line 104: Maybe add the info in the warning that the solvedTapPosition is set to null (or something else if you choose another value)?

rolnico, line 104: Regarding throwing an exception or not, I would do the same as for the `tapPosition`. Also, it should probably be done the same way for all solved values in order to be consistent

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractTwoWindingsTransformerTest.java
```diff
@@ -367,16 +402,23 @@ private RatioTapChanger createRatioTapChanger(TwoWindingsTransformer transformer
...
-                .add();
+                .endStep();
+        if (solvedTapPosition != null) {
+            adder.setSolvedTapPosition(solvedTapPosition);
+        }
+        return adder.add();
     }
 
     private PhaseTapChanger createPhaseTapChanger(TwoWindingsTransformer transformer, Terminal terminal) {
         return createPhaseTapChanger(transformer, terminal, false);
```
rolnico, line 413:
```suggestion
        return createPhaseTapChanger(transformer, terminal, false, null);
```
To be consistent

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractTwoWindingsTransformerTest.java
```diff
@@ -335,11 +366,15 @@ private void createTwoWindingTransformer(String id, String name, double r, doubl
...
 
     private RatioTapChanger createRatioTapChanger(TwoWindingsTransformer transformer, Terminal terminal) {
-        return createRatioTapChanger(transformer, terminal, false);
+        return createRatioTapChanger(transformer, terminal, false, null);
     }
 
     private RatioTapChanger createRatioTapChanger(TwoWindingsTransformer transformer, Terminal terminal, boolean regulating) {
-        return transformer.newRatioTapChanger()
+        return createRatioTapChanger(transformer, terminal, regulating, null);
+    }
```
rolnico, line 374: this method does not seem to be used anymore

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractTwoWindingsTransformerTest.java
```diff
@@ -367,16 +402,23 @@ private RatioTapChanger createRatioTapChanger(TwoWindingsTransformer transformer
...
     }
 
     private PhaseTapChanger createPhaseTapChanger(TwoWindingsTransformer transformer, Terminal terminal) {
         return createPhaseTapChanger(transformer, terminal, false);
     }
 
     private PhaseTapChanger createPhaseTapChanger(TwoWindingsTransformer transformer, Terminal terminal, boolean regulating) {
-        return transformer.newPhaseTapChanger()
+        return createPhaseTapChanger(transformer, terminal, regulating, null);
+    }
```
rolnico, line 418: this method does not seem to be used anymore

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTransformerSerDe.java
```diff
@@ -68,6 +69,7 @@ private static void writeTapChanger(TapChanger<?, ?, ?, ?> tc, NetworkSerializer
         context.getWriter().writeIntAttribute(ATTR_LOW_TAP_POSITION, tc.getLowTapPosition());
         var tp = tc.findTapPosition();
         context.getWriter().writeOptionalIntAttribute(ATTR_TAP_POSITION, tp.isPresent() ? tp.getAsInt() : null);
+        IidmSerDeUtil.runFromMinimumVersion(IidmVersion.V_1_14, context, () -> context.getWriter().writeOptionalIntAttribute(ATTR_SOLVED_TAP_POSITION, tc.getSolvedTapPosition()));
```
rolnico, line 72: You could test this in `RatioTapChangerSerDeTest` with and without values

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTransformerSerDe.java
```diff
@@ -214,6 +216,10 @@ private static void readTapChangerAttributes(TapChangerAdder<?, ?, ?, ?, ?, ?> a
         boolean regulating = context.getReader().readOptionalBooleanAttribute(ATTR_REGULATING).orElse(false);
         int lowTapPosition = context.getReader().readIntAttribute(ATTR_LOW_TAP_POSITION);
         OptionalInt tapPosition = context.getReader().readOptionalIntAttribute(ATTR_TAP_POSITION);
+        IidmSerDeUtil.runFromMinimumVersion(IidmVersion.V_1_14, context, () -> {
+            OptionalInt solvedTapPosition = context.getReader().readOptionalIntAttribute(ATTR_SOLVED_TAP_POSITION);
+            solvedTapPosition.ifPresent(adder::setSolvedTapPosition);
+        });
```
rolnico, line 222: You could test this in `RatioTapChangerSerDeTest` with and without values

rolnico (changes requested):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChanger.java
```diff
@@ -43,9 +43,11 @@ abstract class AbstractTapChanger<H extends TapChangerParent, C extends Abstract
 
     protected final TDoubleArrayList targetDeadband;
 
+    protected final ArrayList<Integer> solvedTapPosition;
+
     protected AbstractTapChanger(H parent,
                                  int lowTapPosition, List<S> steps, TerminalExt regulationTerminal,
-                                 Integer tapPosition, boolean regulating, double targetDeadband, String type) {
+                                 Integer tapPosition, boolean regulating, double targetDeadband, String type, Integer solvedTapPosition) {
```
olperr1: Can you change the order and put `solvedTapPosition` just after `tapPosition`?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChanger.java
```diff
@@ -58,9 +60,11 @@ protected AbstractTapChanger(H parent,
         regulatingPoint.setRegulatingTerminal(regulationTerminal);
         this.tapPosition = new ArrayList<>(variantArraySize);
         this.targetDeadband = new TDoubleArrayList(variantArraySize);
+        this.solvedTapPosition = new ArrayList<>(variantArraySize);
```
olperr1, line 66: Can you place this line just after the one for `tapPosition`?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChanger.java
```diff
@@ -58,9 +60,11 @@ protected AbstractTapChanger(H parent,
...
         for (int i = 0; i < variantArraySize; i++) {
             this.tapPosition.add(tapPosition);
             this.targetDeadband.add(targetDeadband);
+            this.solvedTapPosition.add(solvedTapPosition);
```
olperr1, line 70: Same here.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChangerAdderImpl.java
olperr1: There's an inconsistency with the tap position:
- in the adder, an invalid tap position is allowed for the EQUIPMENT level;
- in `AbstractTapChanger`, calling `setTapPosition(...)` with an invalid value throws an exception regardless the level;
- `checkRtc(...)` and `checkPtc(...)` of `ValidationUtil` doesn't check that the position is valid, it only checks that it can't be null in SSH level.

This is strange, because if the tap position is set with an invalid value via the adder, there won't be future validation of this value and it will be possible to set the network in SSH level without fixing it.

As a direct consequence, if this mechanism is not changed and we only warn for invalid values of solvedTapPosition in the adder, copying solved values into "regular" values will throw an exception.

What can be done:
- in another PR:
  - check in `checkPtc` and `checkRtc` that the tap position is among the defined taps 
  - in the tapPosition setter, downgrade the validation level to EQUIPMENT when the new value is invalid and the min validation level allows it, throw an exception if it does not
- in this PR:
  - decide if the solved values should impact the validation level or not.
    - I think it should not and the validation level should be adjusted (or exceptions should be raised if not possible) when copying the values from `solvedTapPosition` to `tapPosition`. But this will be done automatically via the setter.

@So-Fras , @flo-dup: for information.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/RatioTapChangerSerDeTest.java
```diff
@@ -32,7 +32,16 @@ void readFaultyVersionRatioTapChangerFile() {
         });
     }
 
+    @Test
+    void roundTripWithSolvedTapPosition() throws IOException {
+        allFormatsRoundTripTest(createTestNetwork(0), "ratioTapChangerReactivePowerControlRefWithSolvedTapPosition.xml", CURRENT_IIDM_VERSION);
```
olperr1, line 37: Can you also add the following test, in prevision of the future versions?
```java
allFormatsRoundTripFromVersionedXmlFromMinToCurrentVersionTest("ratioTapChangerReactivePowerControlRefWithSolvedTapPosition.xml", IidmVersion.V_1_14);
```

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractTwoWindingsTransformerTest.java
```diff
@@ -367,16 +398,19 @@ private RatioTapChanger createRatioTapChanger(TwoWindingsTransformer transformer
                 .setG(0.0)
                 .setB(0.0)
                 .setRho(1.0)
-                .endStep()
-                .add();
+                .endStep();
+        if (solvedTapPosition != null) {
```
olperr1: This test is not needed: the adder supports `setSolvedTapPosition(null);`

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractTwoWindingsTransformerTest.java
```diff
@@ -407,8 +441,11 @@ private PhaseTapChanger createPhaseTapChanger(TwoWindingsTransformer transformer
                 .setB(0.0)
                 .setRho(1.0)
                 .setAlpha(10.0)
-                .endStep()
-                .add();
+                .endStep();
+        if (solvedTapPosition != null) {
```
olperr1: Same here: this test is not needed.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChangerAdderImpl.java
colineplqt: For reference: #3478

olperr1: Thanks @colinepiloquet!

colineplqt:
Based on one comment in #3476, I am wondering what `getCurrentTapPosition` should return if the `solvedTapPosition` is not null.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTransformerSerDe.java
```diff
@@ -68,6 +69,7 @@ private static void writeTapChanger(TapChanger<?, ?, ?, ?> tc, NetworkSerializer
         context.getWriter().writeIntAttribute(ATTR_LOW_TAP_POSITION, tc.getLowTapPosition());
         var tp = tc.findTapPosition();
         context.getWriter().writeOptionalIntAttribute(ATTR_TAP_POSITION, tp.isPresent() ? tp.getAsInt() : null);
```
olperr1: For backward compatibility, if the **solved** tap position is present, it should be exported in the `"tapPosition"` field for IIDM version < 1.14.

olperr1 (changes requested):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChangerAdderImpl.java
olperr1: After checking this point with @geofjamg, we are convinced that the operations described in my long comment are not ideal.
See https://github.com/powsybl/powsybl-core/pull/3478#issuecomment-2980822289.

colineplqt: Ok so I am guessing that for solved tap position, anything outside bounds is not allowed in any case and null is also allowed in any case because it can be undefined in SSH validation level.

olperr1: Yes I think that we should do that.

olperr1 (dismissed):

rolnico (dismissed):

rcourtier (approved):

PR #3444, opened 2025-05-14
Hvdc adjustments

rcourtier:

No.

Bug fix.

There are a few things at CGMES import of a DC model that give non-uniform or erroneous result:
* `HvdcLine.mode` can't be determined when importing an EQ only because there isn't enough information in the EQ to determine which converter (side) is rectifier and which is inverter. A default value has to be given. This default value depends on the order the converters appear in the EQ file, and on the type of the hvdc line: it is `SIDE_1_RECTIFIER_SIDE_2_INVERTER` for a LCC and `SIDE_1_INVERTER_SIDE_2_RECTIFIER` for a VSC.
* In the case the active power at **p**oint of **c**ommon **c**oupling (pcc) is defined at the inverter side (and the rectifier side sets the rated dc voltage), the inverter pcc active power is brought back to the rectifier side by adding losses on the converters, and that makes the `HvdcLine.activePowerSetpoint`. However, the `HvdcLine.maxP` is still calculated as 120% of the inverter pcc active power. This can be a problem and lead to situations where the `activePowerSetpoint` is greater than the `maxP`.
* `HvdcLine` resistive losses are not considered. This has an impact when the active power is defined at the inverter pcc, and is brought back to the rectifier side by adding the losses. 
* In the case of a DC configuration with 2 CGMES `DCLineSegment` per `ACDCConverter` pair (typically a monopole with metallic return), only 1 `HvdcLine` is created, with the following equivalent resistance: $`\frac{1}{req} = \frac{1}{r1} + \frac{1}{r2}`$
* In the case of a DC configuration with 2 CGMES `ACDCConverter` pair per `DCLineSegment` (typicall a monopole with 2*6-pulses bridge per converter unit), 2 `HvdcLine` are created, each having the following resistance: $`r1 = r2 = 2 \times r`$

The CGMES import of a DC model gives now more uniform and precise results:
* The default `HvdcLine` mode in case of an EQ only import is always `SIDE_1_RECTIFIER_SIDE_2_INVERTER` for all hvdc line types, where side 1 is the side with the lowest converter id.
* `HvdcLine.maxP` is always equal to 120% of `HvdcLine.activePowerSetpoint`.
* `HvdcLine` resistive losses are now considered. They are calculated as: $`r \times idc²`$, where:
  * If CGMES defines active power at rectifier pcc:
$`pDcRecifier = rectifierPccActivePower - rectifierPoleLosses`$
$`idc = r \times \frac{pDcRectifier^2}{nominalV^2}`$
  * If CGMES defines active power at inverter pcc:
$`pDcInverter = -(abs(inverterPccActivePower) + inverterPoleLosses)`$
$`idc = \frac{nominalV - \sqrt{nominalV^2 - 4 \times r \times abs(pDcInverter)}}{2 \times r}`$;
* In the case of a DC configuration with 2 CGMES `DCLineSegment` per `ACDCConverter` pair (typically a monopole with metallic return), only 1 `HvdcLine` is created, with the following equivalent resistance: $`req = r1 + r2`$
* In the case of a DC configuration with 2 CGMES `ACDCConverter` pair per `DCLineSegment` (typicall a monopole with 2*6-pulses bridge per converter unit), 2 `HvdcLine` are created, each having the following resistance: $`r1 = r2 = \frac {r}{2}`$

- [ ] Yes
- [X] No

**Other information**:

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/hvdc/CgmesDcConversion.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/hvdc/CgmesDcConversion.java
@@ -384,12 +384,12 @@ public class CgmesDcConversion {
             double pDcRectifier = pAC2 - poleLossP2;
             return HvdcUtils.getHvdcLineLosses(pDcRectifier, ratedUdc, r);
         } else if (HvdcLine.ConvertersMode.SIDE_1_RECTIFIER_SIDE_2_INVERTER.equals(operatingMode) && pAC2 != 0.0) {
-            double pDcInverter = -1 * pAC2 + poleLossP2;
-            double idc = (ratedUdc - Math.sqrt(ratedUdc * ratedUdc - 4 * r * pDcInverter)) / (2 * r);
+            double pDcInverter = -1 * (Math.abs(pAC2) + poleLossP2);
+            double idc = (ratedUdc - Math.sqrt(ratedUdc * ratedUdc + 4 * r * pDcInverter)) / (2 * r);
             return r * idc * idc;
         } else if (HvdcLine.ConvertersMode.SIDE_1_INVERTER_SIDE_2_RECTIFIER.equals(operatingMode) && pAC1 != 0.0) {
-            double pDcInverter = -1 * pAC1 + poleLossP1;
-            double idc = (ratedUdc - Math.sqrt(ratedUdc * ratedUdc - 4 * r * pDcInverter)) / (2 * r);
+            double pDcInverter = -1 * (Math.abs(pAC1) + poleLossP1);
+            double idc = (ratedUdc - Math.sqrt(ratedUdc * ratedUdc + 4 * r * pDcInverter)) / (2 * r);
             return r * idc * idc;
         } else {
             return 0.0;
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/hvdc/CgmesDcConversion.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/hvdc/CgmesDcConversion.java
@@ -21,6 +21,7 @@ import com.powsybl.iidm.network.HvdcConverterStation;
 import com.powsybl.iidm.network.HvdcConverterStation.HvdcType;
 import com.powsybl.iidm.network.HvdcLine;
 import com.powsybl.iidm.network.LccConverterStation;
+import com.powsybl.iidm.network.util.HvdcUtils;
 import com.powsybl.triplestore.api.PropertyBag;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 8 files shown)


marqueslanauja:
Two observations in the pull request description regarding the calculation of idc:

- When the power is defined on the rectifier side, you have defined the current as the losses.
- When the power is defined on the inverter side, I prefer that the formulas include the flow orientation (the sign convention). (+ inside the square root)

`HvdcLine` resistive losses are now considered. They are calculated as: $`r \times idc²`$, where:

  * If CGMES defines active power at rectifier pcc:
$`pDcRecifier = rectifierPccActivePower - rectifierPoleLosses`$
$`idc = \frac{pDcRectifier}{nominalV}`$

  * If CGMES defines active power at inverter pcc:
$`pDcInverter = inverterPccActivePower + inverterPoleLosses`$
$`idc = \frac{nominalV - \sqrt{nominalV^2 + 4 \times r \times pDcInverter}}{2 \times r}`$;

image

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/hvdc/CgmesDcConversion.java
```diff
@@ -376,6 +376,26 @@ private static double getPAc(PropertyBag p) {
...
+            return r * idc * idc;
+        } else if (HvdcLine.ConvertersMode.SIDE_1_INVERTER_SIDE_2_RECTIFIER.equals(operatingMode) && pAC1 != 0.0) {
+            double pDcInverter = -1 * pAC1 + poleLossP1;
+            double idc = (ratedUdc - Math.sqrt(ratedUdc * ratedUdc - 4 * r * pDcInverter)) / (2 * r);
+            return r * idc * idc;
+        } else {
+            return 0.0;
+        }
+    }
+
```
marqueslanauja, line 398: My proposal to adjust to the formulas including the flow orientation:
```
} else if (HvdcLine.ConvertersMode.SIDE_1_RECTIFIER_SIDE_2_INVERTER.equals(operatingMode) && pAC2 != 0.0) {
            double pDcInverter = -(Math.abs(pAC2) + poleLossP2);
            double idcRectifier = (ratedUdc - Math.sqrt(ratedUdc * ratedUdc + 4 * r * pDcInverter)) / (2 * r);
            return r * idcRectifier * idcRectifier;
        } else if (HvdcLine.ConvertersMode.SIDE_1_INVERTER_SIDE_2_RECTIFIER.equals(operatingMode) && pAC1 != 0.0) {
            double pDcInverter = -(Math.abs(pAC1) + poleLossP1);
            double idcRectifier = (ratedUdc - Math.sqrt(ratedUdc * ratedUdc + 4 * r * pDcInverter)) / (2 * r);
            return r * idcRectifier * idcRectifier;
        } else {
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
```diff
@@ -677,19 +678,21 @@ private static void writeConverters(Network network, String cimNamespace, XMLStr
...
             if (Double.isNaN(p)) {
-                p = converterStation.getHvdcLine().getActivePowerSetpoint();
+                p = hvdcLine.getActivePowerSetpoint();
             }
             double otherConverterStationLossFactor = converterStation.getOtherConverterStation().map(HvdcConverterStation::getLossFactor).orElse(0.0f);
-            double pDCInverter = Math.abs(p) * (1 - otherConverterStationLossFactor / 100);
+            double pDCRectifier = Math.abs(p) * (1 - otherConverterStationLossFactor / 100);
+            double pDCInverter = pDCRectifier - HvdcUtils.getHvdcLineLosses(pDCRectifier, hvdcLine.getNominalV(), hvdcLine.getR());
             poleLoss = pDCInverter * converterStation.getLossFactor() / 100;
         }
```
marqueslanauja, line 697: I believe we cannot manage it the same way if we take the power from the terminal or the setpoint. My proposal:

```
} else {
            double p = converterStation.getTerminal().getP();
            double pDCInverter;
            if (Double.isNaN(p)) {
                p = hvdcLine.getActivePowerSetpoint();
                double otherConverterStationLossFactor = converterStation.getOtherConverterStation().map(HvdcConverterStation::getLossFactor).orElse(0.0f);
                double pDCRectifier = Math.abs(p) * (1 - otherConverterStationLossFactor / 100);
                pDCInverter = pDCRectifier - HvdcUtils.getHvdcLineLosses(pDCRectifier, hvdcLine.getNominalV(), hvdcLine.getR());
            } else {
                pDCInverter = Math.abs(p) * 100.0 / (1 - converterStation.getLossFactor());
            }
            poleLoss = pDCInverter * converterStation.getLossFactor() / 100;
        }
```

marqueslanauja (approved):

olperr1 (changes requested):

olperr1 (dismissed):

olperr1:
@marqueslanauja: changing the base branch has dismissed your approval.
Can you check quickly that it is still OK for you (except the rebasing, no changes should have been made) and re-approve it?

olperr1:
> Reviewed. It is fine.

Thanks!

PR #3465, opened 2025-06-03
Prevent potential backtracking on PSS/E import

olperr1:

No

Bux fix / security

Potential backtracking on PSS/E import.

The regex used is no more vulnerable to backtracking.

- [ ] Yes
- [X] No

**Other information**:

### DIFF
--- a/psse/psse-model/src/main/java/com/powsybl/psse/model/io/LegacyTextReader.java
+++ b/psse/psse-model/src/main/java/com/powsybl/psse/model/io/LegacyTextReader.java
@@ -132,7 +132,7 @@ public class LegacyTextReader {
     }
 
     // Compact spaces, remove spaces before the comma, and replace space with comma outside quoted text
-    private static String processText(String line) {
+    protected static String processText(String line) {
         if (line == null || line.isEmpty()) {
             return line;
         }
--- a/psse/psse-model/src/main/java/com/powsybl/psse/model/io/LegacyTextReader.java
+++ b/psse/psse-model/src/main/java/com/powsybl/psse/model/io/LegacyTextReader.java
@@ -147,7 +147,8 @@ public class LegacyTextReader {
                 result.append(part);
             } else {
                 // Outside quotes: process the txt
-                result.append(part.replaceAll("\\s+,", ",")
+                // On the next line: "(?<=\S|^)" = backtracking protection. The previous char should be a non-white char or a line start.
+                result.append(part.replaceAll("(?<=\\S|^)\\s+,", ",") // see comment above
                         .replaceAll(",\\s+", ",")
                         .replaceAll("\\s+", ","));
             }
--- a/psse/psse-model/src/test/java/com/powsybl/psse/model/io/LegacyTextReaderTest.java
+++ b/psse/psse-model/src/test/java/com/powsybl/psse/model/io/LegacyTextReaderTest.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2025, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+package com.powsybl.psse.model.io;
+
+import org.apache.commons.lang3.StringUtils;
+import org.assertj.core.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+/**
+ * @author Olivier Perrin {@literal <olivier.perrin at rte-france.com>}
+ */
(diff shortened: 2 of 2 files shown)


olperr1:
The security hotspot is always present, but the unit test asserts that the backtracking problem is fixed:
without the regex change, the unit test fails after more than 30 sec. With the new regex, it finishes almost instantaneously.

psse/psse-model/src/test/java/com/powsybl/psse/model/io/LegacyTextReaderTest.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @Test
+    void noBacktrackingOnProcessTextTest() {
+        String whiteChars = StringUtils.repeat(" ", 100_000);
+        String line = "a" + whiteChars + "b" + whiteChars;
+        long start = System.currentTimeMillis();
+        LegacyTextReader.processText(line);
+        long end = System.currentTimeMillis();
+        if (end - start > 500) {
+            Assertions.fail("Timeout");
```
rolnico: directly import the `fail` static method

psse/psse-model/src/test/java/com/powsybl/psse/model/io/LegacyTextReaderTest.java
```diff
@@ -0,0 +1,31 @@
...
+import org.assertj.core.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+/**
+ * @author Olivier Perrin {@literal <olivier.perrin at rte-france.com>}
+ */
+class LegacyTextReaderTest {
+
+    @Test
+    void noBacktrackingOnProcessTextTest() {
```
rolnico, line 22: add a test with `String line = whiteChars + "a" + whiteChars + "b" + whiteChars` for the case with a string starting with spaces (the other possibility in your regex)

psse/psse-model/src/main/java/com/powsybl/psse/model/io/LegacyTextReader.java
```diff
@@ -147,7 +147,8 @@ private static String processText(String line) {
                 result.append(part);
             } else {
                 // Outside quotes: process the txt
-                result.append(part.replaceAll("\\s+,", ",")
+                // On the next line: "(?<=\S|^)" = backtracking protection. The previous char should be a non-white char or a line start.
+                result.append(part.replaceAll("(?<=\\S|^)\\s+,", ",") // see comment above
```
rolnico, line 151: The `|^` might not be necessary, to be checked
```suggestion
                result.append(part.replaceAll("(?<=\\S)\\s+,", ",") // see comment above
```

psse/psse-model/src/test/java/com/powsybl/psse/model/io/LegacyTextReaderTest.java
```diff
@@ -0,0 +1,31 @@
...
+        String whiteChars = StringUtils.repeat(" ", 100_000);
+        String line = "a" + whiteChars + "b" + whiteChars;
+        long start = System.currentTimeMillis();
+        LegacyTextReader.processText(line);
```
rolnico: Maybe add an `assertEquals`

rolnico (changes requested):

psse/psse-model/src/test/java/com/powsybl/psse/model/io/LegacyTextReaderTest.java
olperr1: I changed `line` in the test to have the leading white spaces, but they are trimmed before using the modified regex (`part` has not these spaces). So it really has no impact.

psse/psse-model/src/main/java/com/powsybl/psse/model/io/LegacyTextReader.java
olperr1: I don't think it is really needed because leading white spaces are trimmed from `part` in the current state of the code, but it's maybe safer to keep it.

psse/psse-model/src/test/java/com/powsybl/psse/model/io/LegacyTextReaderTest.java
olperr1: Done

psse/psse-model/src/test/java/com/powsybl/psse/model/io/LegacyTextReaderTest.java
olperr1: Done

psse/psse-model/src/main/java/com/powsybl/psse/model/io/LegacyTextReader.java
rolnico: I though that the `^` case was already managed through the `\\S` case but that does not seem to be the case so keeping it is better indeed

PR #3469, opened 2025-06-06
Execute each deserialization end task's step only once

olperr1:

No

Bug fix / performance fix

When importing an XIIDM network file, `executeEndTasks` is executed for the `BEFORE_EXTENSIONS` step a lot of times (before each `<extension>` tag) instead of once.
Since only `OverloadManagementSystem` related tasks are defined for this step, only network file with this type of equipment may encountered import problems.
But since every end task is checked for every `<extension>` tag of the file, this slows the import unnecessarily.

Each step is only processed once, as expected. `executeEndTasks` is always called before each `<extension>` tag, but it immediately exits except for the first one.

- [ ] Yes
- [X] No

**Other information**:

### DIFF
--- a/iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkDeserializerContext.java
+++ b/iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkDeserializerContext.java
@@ -35,6 +35,7 @@ public class NetworkDeserializerContext extends AbstractNetworkSerDeContext<Impo
     private ValidationLevel networkValidationLevel;
 
     private Set<String> ignoredEquipments = new HashSet<>();
+    private EnumSet<DeserializationEndTask.Step> alreadyProcessedEndTasksSteps = EnumSet.noneOf(DeserializationEndTask.Step.class);
 
     public NetworkDeserializerContext(Anonymizer anonymizer, TreeDataReader reader) {
         this(anonymizer, reader, new ImportOptions(), CURRENT_IIDM_VERSION, Collections.emptyMap());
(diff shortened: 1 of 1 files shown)


olperr1:
`EnumSet` is not thread-safe, but I don't think this is needed here.
@flo-dup: If you think it is a problem, I can change the code to have it thread-safe. Let me know.

antoinebhs (dismissed):
I confirm that this PR addresses the performance regression on network read or copy.
On my computer, execution time for Network.read() on a real network is 1.7s with this PR instead of 7s on main branch.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkDeserializerContext.java
```diff
@@ -35,6 +35,7 @@ public class NetworkDeserializerContext extends AbstractNetworkSerDeContext<Impo
     private ValidationLevel networkValidationLevel;
 
     private Set<String> ignoredEquipments = new HashSet<>();
+    private EnumSet<DeserializationEndTask.Step> alreadyProcessedEndTasksSteps = EnumSet.noneOf(DeserializationEndTask.Step.class);
```
flo-dup: can be final, and renamed? as "already" and "processed" are carrying the same information twice
```suggestion
    private final EnumSet<DeserializationEndTask.Step> processedEndTasksSteps = EnumSet.noneOf(DeserializationEndTask.Step.class);
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkDeserializerContext.java
```diff
@@ -58,6 +59,11 @@ public void addEndTask(DeserializationEndTask.Step step, Runnable task) {
     }
 
     public void executeEndTasks(Network network, DeserializationEndTask.Step step, ReportNode reportNode) {
+        if (alreadyProcessedEndTasksSteps.contains(step)) {
```
flo-dup: you could maybe directly test with add instead of contains

flo-dup:
> `EnumSet` is not thread-safe, but I don't think this is needed here.

Indeed, no need of thread-safe set, the deserialization is not thread-safe

flo-dup (approved):

Issue #3239, opened 2024-12-03
When disconnecting an equipment choose to open all real switches after fictitious ones


alicecaron:
### Describe the current behavior

When disconnecting an equipement (`Line`), the graph traversal stops when it encounters a fictitious switch and opens it.

### Describe the expected behavior

Thanks to a new boolean option, when disconnecting an equipement we can choose to not stop but to traverse fictitious switches (without openning them) and continue until encountering a real switch to open too.

### Describe the motivation

Add the possibility to propagate the opening of real switches aftter fictitious ones when disconnecting an equipement.

### Extra Information

![image](https://github.com/user-attachments/assets/f47e6a79-88c3-41f3-8b86-298055504bf9)

In the situation above:
If we want to disconnect the line L1:
- With the propagation option enabled we ignore the fictitious switches so we expect: S1f, S2f and S3f to stay closed, and S11, S22 and S33 to be opened.
- With the propagation option disabled we stop to the fictitious switches so we expect only S11, S1f, S2f and S3f to be opened.

flo-dup:
After some discussion, it seems interesting to traverse the voltage level only if it is marked as fictitious. This corresponds to the given usecase.

PR #3274, opened 2025-01-16
Add option to propagate line connection and disconnection through fictitious voltage levels

rolnico:
**Please check if the PR fulfills these requirements**

- [ ] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**Does this PR already have an issue describing the problem?**
Fixes #3239

**What kind of change does this PR introduce?**
Feature

**Does this PR introduce a breaking change or deprecate an API?**
- [x] Yes
- [ ] No

**If yes, please check if the following requirements are fulfilled**

- [ ] The *Breaking Change* or *Deprecated* label has been added
- [ ] The migration steps are described in the following section

**What changes might users need to make in their application due to this PR? (migration steps)**

**Other information**:

PR #3471, opened 2025-06-11
Add option to propagate line connection and disconnection through fictitious voltage levels

rolnico:
This is another proposition rather than #3274.

Fixes https://github.com/powsybl/powsybl-core/issues/3239

Feature

- [x] Yes
- [ ] No

**Other information**:

### DIFF
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Line.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Line.java
@@ -7,6 +7,9 @@
  */
 package com.powsybl.iidm.network;
 
+import com.powsybl.iidm.network.util.ConnectionElementsContainer;
+
+import java.util.HashSet;
 import java.util.function.Predicate;
 
 /**
@@ -38,13 +41,31 @@ public interface Line extends Branch<Line>, Connectable<Line>, MutableLineCharac
 
     boolean connect(boolean propagateDisconnectionIfNeeded);
 
-    boolean connect(Predicate<Switch> isTypeSwitchToOperate, boolean propagateDisconnectionIfNeeded);
+    default boolean connect(Predicate<Switch> isTypeSwitchToOperate, boolean propagateDisconnectionIfNeeded) {
+        return connect(isTypeSwitchToOperate, null, propagateDisconnectionIfNeeded);
+    }
 
-    boolean connect(Predicate<Switch> isTypeSwitchToOperate, ThreeSides side, boolean propagateDisconnectionIfNeeded);
+    default boolean connect(Predicate<Switch> isTypeSwitchToOperate, ThreeSides side, boolean propagateDisconnectionIfNeeded) {
+        return connect(isTypeSwitchToOperate, side, propagateDisconnectionIfNeeded, true,
+            new ConnectionElementsContainer(new HashSet<>(), new HashSet<>()));
+    }
+
+    boolean connect(Predicate<Switch> isTypeSwitchToOperate, ThreeSides side,
+                    boolean propagateDisconnectionIfNeeded, boolean connectFromHere,
+                    ConnectionElementsContainer connectionElementsContainer);
 
     boolean disconnect(boolean propagateDisconnectionIfNeeded);
 
-    boolean disconnect(Predicate<Switch> isSwitchOpenable, boolean propagateDisconnectionIfNeeded);
+    default boolean disconnect(Predicate<Switch> isSwitchOpenable, boolean propagateDisconnectionIfNeeded) {
+        return disconnect(isSwitchOpenable, null, propagateDisconnectionIfNeeded);
+    }
 
-    boolean disconnect(Predicate<Switch> isSwitchOpenable, ThreeSides side, boolean propagateDisconnectionIfNeeded);
+    default boolean disconnect(Predicate<Switch> isSwitchOpenable, ThreeSides side, boolean propagateDisconnectionIfNeeded) {
+        return disconnect(isSwitchOpenable, side, propagateDisconnectionIfNeeded, true,
+            new ConnectionElementsContainer(new HashSet<>(), new HashSet<>()));
+    }
+
+    boolean disconnect(Predicate<Switch> isSwitchOpenable, ThreeSides side,
+                       boolean propagateDisconnectionIfNeeded, boolean disconnectFromHere,
+                       ConnectionElementsContainer connectionElementsContainer);
 }
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Line.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Line.java
@@ -7,6 +7,8 @@
  */
 package com.powsybl.iidm.network;
 
+import java.util.function.Predicate;
+
 /**
  * An AC line.
  * <p>
@@ -33,4 +35,16 @@ public interface Line extends Branch<Line>, Connectable<Line>, MutableLineCharac
     default IdentifiableType getType() {
         return IdentifiableType.LINE;
     }
+
+    boolean connect(boolean propagateDisconnectionIfNeeded);
+
+    boolean connect(Predicate<Switch> isTypeSwitchToOperate, boolean propagateDisconnectionIfNeeded);
+
+    boolean connect(Predicate<Switch> isTypeSwitchToOperate, ThreeSides side, boolean propagateDisconnectionIfNeeded);
+
+    boolean disconnect(boolean propagateDisconnectionIfNeeded);
+
+    boolean disconnect(Predicate<Switch> isSwitchOpenable, boolean propagateDisconnectionIfNeeded);
+
+    boolean disconnect(Predicate<Switch> isSwitchOpenable, ThreeSides side, boolean propagateDisconnectionIfNeeded);
 }
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
@@ -11,11 +11,11 @@ import com.powsybl.commons.report.ReportNode;
 import com.powsybl.iidm.network.Identifiable;
 import com.powsybl.iidm.network.Switch;
 import com.powsybl.iidm.network.Terminal;
+import com.powsybl.iidm.network.util.ConnectionElementsContainer;
 import com.powsybl.iidm.network.util.NetworkReports;
 
 import java.util.HashSet;
 import java.util.List;
-import java.util.Set;
 import java.util.function.Predicate;
 
 import static com.powsybl.iidm.network.TopologyKind.BUS_BREAKER;
@@ -41,20 +41,22 @@ public final class ConnectDisconnectUtil {
      * @return {@code true} if all the specified terminals have been connected, else {@code false}.
      */
     static boolean connectAllTerminals(Identifiable<?> identifiable, List<? extends Terminal> terminals,
-                                       Predicate<Switch> isTypeSwitchToOperate, boolean propagateDisconnectionIfNeeded,
+                                       Predicate<Switch> isTypeSwitchToOperate,
+                                       ConnectionElementsContainer connectionElementsContainer,
+                                       boolean operateSwitchesFromHere, boolean propagateConnectionIfNeeded,
                                        ReportNode reportNode) {
 
         // Booleans
         boolean isAlreadyConnected = true;
         boolean isNowConnected = true;
 
-        // Initialisation of a list to open in case some terminals are in node-breaker view
-        Set<SwitchImpl> switchForDisconnection = new HashSet<>();
+        // Initialize a list of switches to close and of bus-breaker terminals to connect
+        ConnectionElementsContainer localConnectionElementsContainer = new ConnectionElementsContainer(new HashSet<>(), new HashSet<>());
 
         // We try to connect each terminal
         for (Terminal terminal : terminals) {
             // Check if the terminal is already connected
-            if (terminal.isConnected()) {
+            if (terminal.isConnected() && (!propagateConnectionIfNeeded || !terminal.getVoltageLevel().isFictitious())) {
                 NetworkReports.alreadyConnectedIdentifiableTerminal(reportNode, identifiable.getId());
                 continue;
             } else {
@@ -64,9 +66,12 @@ public final class ConnectDisconnectUtil {
             // If it's a node-breaker terminal, the switches to connect are added to a set
             if (terminal.getVoltageLevel().getTopologyKind() == NODE_BREAKER) {
                 NodeBreakerTopologyModel topologyModel = (NodeBreakerTopologyModel) ((VoltageLevelImpl) terminal.getVoltageLevel()).getTopologyModel();
-                isNowConnected = topologyModel.getConnectingSwitches(terminal, isTypeSwitchToOperate, switchForDisconnection);
+                isNowConnected = topologyModel.canTheTerminalBeConnected(terminal, isTypeSwitchToOperate, propagateConnectionIfNeeded,
+                    localConnectionElementsContainer);
+            } else if (terminal.getVoltageLevel().getTopologyKind() == BUS_BREAKER) {
+                // If it's a bus-breaker terminal, the terminal is just added to the set
+                localConnectionElementsContainer.busBreakerTerminalsToOperate().add(terminal);
             }
-            // If it's a bus-breaker terminal, there is nothing to do
 
             // Exit if the terminal cannot be connected
             if (!isNowConnected) {
@@ -79,17 +84,17 @@ public final class ConnectDisconnectUtil {
             return false;
         }
 
-        // Connect all bus-breaker terminals
-        for (Terminal terminal : terminals) {
-            if (!terminal.isConnected()
-                && terminal.getVoltageLevel().getTopologyKind() == BUS_BREAKER) {
-                // At this point, isNowConnected should always stay true but let's be careful
-                isNowConnected = isNowConnected && terminal.connect(isTypeSwitchToOperate);
-            }
+        if (operateSwitchesFromHere) {
+            // Close all switches on node-breaker terminals to connect the network element
+            localConnectionElementsContainer.switchesToOperate().forEach(sw -> sw.setOpen(false));
+
+            // Connect all bus-breaker terminals
+            localConnectionElementsContainer.busBreakerTerminalsToOperate().forEach(t -> t.connect(isTypeSwitchToOperate));
+        } else {
+            // Add the switches and the bus-breaker terminals to the sets
+            connectionElementsContainer.addAll(localConnectionElementsContainer);
         }
 
-        // Disconnect all switches on node-breaker terminals
-        switchForDisconnection.forEach(sw -> sw.setOpen(false));
         return isNowConnected;
     }
 
@@ -105,14 +110,15 @@ public final class ConnectDisconnectUtil {
      * @return {@code true} if all the specified terminals have been disconnected, else {@code false}.
      */
     static boolean disconnectAllTerminals(Identifiable<?> identifiable, List<? extends Terminal> terminals,
-                                          Predicate<Switch> isSwitchOpenable, boolean propagateDisconnectionIfNeeded,
+                                          Predicate<Switch> isSwitchOpenable,
+                                          ConnectionElementsContainer connectionElementsContainer,
+                                          boolean operateSwitchesFromHere, boolean propagateDisconnectionIfNeeded,
                                           ReportNode reportNode) {
         // Booleans
         boolean isAlreadyDisconnected = true;
-        boolean isNowDisconnected = true;
 
-        // Initialisation of a list to open in case some terminals are in node-breaker view
-        Set<SwitchImpl> switchForDisconnection = new HashSet<>();
+        // Initialize a list of switches to open and of bus-breaker terminals to disconnect
+        ConnectionElementsContainer localConnectionElementsContainer = new ConnectionElementsContainer(new HashSet<>(), new HashSet<>());
 
         // We try to disconnect each terminal
         for (Terminal terminal : terminals) {
@@ -127,12 +133,14 @@ public final class ConnectDisconnectUtil {
             // If it's a node-breaker terminal, the switches to disconnect are added to a set
             if (terminal.getVoltageLevel().getTopologyKind() == NODE_BREAKER) {
                 NodeBreakerTopologyModel topologyModel = (NodeBreakerTopologyModel) ((VoltageLevelImpl) terminal.getVoltageLevel()).getTopologyModel();
-                if (!topologyModel.getDisconnectingSwitches(terminal, isSwitchOpenable, switchForDisconnection)) {
+                if (!topologyModel.canTheTerminalBeDisconnected(terminal, isSwitchOpenable, propagateDisconnectionIfNeeded, localConnectionElementsContainer)) {
                     // Exit if the terminal cannot be disconnected
                     return false;
                 }
+            } else if (terminal.getVoltageLevel().getTopologyKind() == BUS_BREAKER) {
+                // If it's a bus-breaker terminal, the terminal is just added to the set
+                localConnectionElementsContainer.busBreakerTerminalsToOperate().add(terminal);
             }
-            // If it's a bus-breaker terminal, there is nothing to do
         }
 
         // Exit if the connectable is already fully disconnected
@@ -140,16 +148,17 @@ public final class ConnectDisconnectUtil {
             return false;
         }
 
-        // Disconnect all bus-breaker terminals
-        for (Terminal terminal : terminals) {
-            if (terminal.isConnected()
-                && terminal.getVoltageLevel().getTopologyKind() == BUS_BREAKER) {
-                // At this point, isNowDisconnected should always stay true but let's be careful
-                isNowDisconnected = isNowDisconnected && terminal.disconnect(isSwitchOpenable);
-            }
+        if (operateSwitchesFromHere) {
+            // Open all switches on node-breaker terminals to disconnect the network element
+            localConnectionElementsContainer.switchesToOperate().forEach(sw -> sw.setOpen(true));
+
+            // Connect all bus-breaker terminals
+            localConnectionElementsContainer.busBreakerTerminalsToOperate().forEach(t -> t.disconnect(isSwitchOpenable));
+        } else {
+            // Add the switches and the bus-breaker terminals to the sets
+            connectionElementsContainer.addAll(localConnectionElementsContainer);
         }
-        // Disconnect all switches on node-breaker terminals
-        switchForDisconnection.forEach(sw -> sw.setOpen(true));
-        return isNowDisconnected;
+
+        return true;
     }
 }
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
@@ -40,7 +40,9 @@ public final class ConnectDisconnectUtil {
      * @param reportNode report node
      * @return {@code true} if all the specified terminals have been connected, else {@code false}.
      */
-    static boolean connectAllTerminals(Identifiable<?> identifiable, List<? extends Terminal> terminals, Predicate<Switch> isTypeSwitchToOperate, ReportNode reportNode) {
+    static boolean connectAllTerminals(Identifiable<?> identifiable, List<? extends Terminal> terminals,
+                                       Predicate<Switch> isTypeSwitchToOperate, boolean propagateDisconnectionIfNeeded,
+                                       ReportNode reportNode) {
 
         // Booleans
         boolean isAlreadyConnected = true;
@@ -102,7 +104,9 @@ public final class ConnectDisconnectUtil {
      * @param reportNode report node
      * @return {@code true} if all the specified terminals have been disconnected, else {@code false}.
      */
-    static boolean disconnectAllTerminals(Identifiable<?> identifiable, List<? extends Terminal> terminals, Predicate<Switch> isSwitchOpenable, ReportNode reportNode) {
+    static boolean disconnectAllTerminals(Identifiable<?> identifiable, List<? extends Terminal> terminals,
+                                          Predicate<Switch> isSwitchOpenable, boolean propagateDisconnectionIfNeeded,
+                                          ReportNode reportNode) {
         // Booleans
         boolean isAlreadyDisconnected = true;
         boolean isNowDisconnected = true;
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java
@@ -16,6 +16,7 @@ import com.powsybl.commons.util.Colors;
 import com.powsybl.iidm.network.*;
 import com.powsybl.iidm.network.VoltageLevel.NodeBreakerView.InternalConnectionAdder;
 import com.powsybl.iidm.network.VoltageLevel.NodeBreakerView.SwitchAdder;
+import com.powsybl.iidm.network.util.ConnectionElementsContainer;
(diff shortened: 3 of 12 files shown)


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java
```diff
@@ -71,10 +72,10 @@ class NodeBreakerTopologyModel extends AbstractTopologyModel {
     private final class VariantImpl implements Variant {
 
         final CalculatedBusTopology calculatedBusTopology
-                = new CalculatedBusTopology();
+            = new CalculatedBusTopology();
```
flo-dup: Could you remove the formatting changes? We should not include these in that PR

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/TopologyModel.java
```diff
@@ -22,7 +24,7 @@ interface TopologyModel {
 
     void detach(TerminalExt terminal);
 
-    boolean connect(TerminalExt terminal, Predicate<? super SwitchImpl> isTypeSwitchToOperate);
+    boolean connect(TerminalExt terminal, Predicate<Switch> isTypeSwitchToOperate);
```
flo-dup, line 27: I don't see why we did need this and why we don't anymore, do you?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Line.java
```diff
@@ -33,4 +38,34 @@ default boolean isTieLine() {
...
+    default boolean disconnect(Predicate<Switch> isSwitchOpenable, boolean propagateDisconnectionIfNeeded) {
+        return disconnect(isSwitchOpenable, null, propagateDisconnectionIfNeeded);
+    }
+
+    default boolean disconnect(Predicate<Switch> isSwitchOpenable, ThreeSides side, boolean propagateDisconnectionIfNeeded) {
+        return disconnect(isSwitchOpenable, side, propagateDisconnectionIfNeeded, true,
+            new ConnectionElementsContainer(new HashSet<>(), new HashSet<>()));
+    }
+
+    boolean disconnect(Predicate<Switch> isSwitchOpenable, ThreeSides side,
```
flo-dup: I think this should rather be in the LineImpl, and the default contents above removed. The ConnectionElementsContainer could remain in the implementation then, exposing it in the API brings confusion I think.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Line.java
```diff
@@ -33,4 +38,34 @@ default boolean isTieLine() {
     default IdentifiableType getType() {
         return IdentifiableType.LINE;
     }
+
+    boolean connect(boolean propagateDisconnectionIfNeeded);
```
flo-dup, line 31: Does the propagateDisconnectionIfNeeded corresponds to the need of the usecase or should it be a propagateDisconnection? That is, should we propagate even if not needed?

We should clarify the usecases with the user, I'm thinking of 3:
- when connecting, if a closeable switch is open on a tee point fictitious voltage level, should we close it and stop or should we still propagate to other voltage levels?
- when disconnecting, if a bus breaker fictitious voltage level is encountered, should we disconnect the terminal and stop or should we propagate to other voltage levels?
- when disconnecting, if an openable switch is closed on a tee point fictitious voltage level, should we open it and stop or should we still propagate to other voltage levels?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java
```diff
@@ -1208,55 +1213,191 @@ public boolean connect(TerminalExt terminal, Predicate<? super SwitchImpl> isSwi
...
             return true;
         } else {
             return false;
         }
     }
 
-    boolean getConnectingSwitches(Terminal terminal, Predicate<? super SwitchImpl> isSwitchOperable, Set<SwitchImpl> switchForConnection) {
+    /**
+     * Find all paths starting from the current terminal to a busbar section that does not contain an open switch,
+     * and that is not of the type of switch the user wants to operate.
```
flo-dup: Shouldn't we say "that does not contain an open switch which  is not of the type of switch the user wants to operate."

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java
```diff
@@ -1208,55 +1213,191 @@ public boolean connect(TerminalExt terminal, Predicate<? super SwitchImpl> isSwi
...
+    }
+
+    private static boolean isLine(Terminal t) {
+        return t != null && t.getConnectable().getType() == IdentifiableType.LINE;
+    }
+
+    private boolean canPropagateConnection(int node, Predicate<Switch> isSwitchOperable,
+                                           ConnectionElementsContainer connectionElementsContainer) {
+        // Find the paths from the line to other lines
+        List<TIntArrayList> pathsFromLine = graph.findAllPaths(node, NodeBreakerTopologyModel::isLine, SwitchPredicates.IS_OPEN);
```
flo-dup: We should propagate to transformers also, as there's the usecase of the star node voltage level with 3 two-winding transformers used to replace three-winding transformers in some networks.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java
```diff
@@ -1208,55 +1213,191 @@ public boolean connect(TerminalExt terminal, Predicate<? super SwitchImpl> isSwi
...
+        if (!pathsFromLine.isEmpty()) {
+            // Initialize a list of lines connected to the BusbarSection
+            Map<Line, ThreeSides> linesAndSidesConnectedToIncomingLine = addOppositeLines(pathsFromLine);
+
+            // We check if the opposite side of each line can be connected
+            for (Map.Entry<Line, ThreeSides> lineAndSide : linesAndSidesConnectedToIncomingLine.entrySet()) {
+                // If any line cannot be connected, we return false
+                if (!lineAndSide.getKey().connect(isSwitchOperable, lineAndSide.getValue(), true, false, connectionElementsContainer)) {
```
flo-dup: shouldn't we secure this call to avoid a misguided user to get an infinite loop resulting in a stack overflow? especially with a `s -> false` predicate...

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java
rolnico: mb, I didn't even see them

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/TopologyModel.java
rolnico: I would say we did not need them... Maybe I previously didn't find a proper way to do without? I don't remember and I didn't find any discussion about that in the previous PR on the subject

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java
rolnico: Should we do this in the current PR or in another one? The initial issue was related to lines (and the release is coming fast :open_mouth: )

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Line.java
rolnico: yeah, that would change the logic quite a bit... I'll add tests so that we can discuss more easily with the users on each usecase

rolnico: IMO, the predicate is here to tell us which switch can be operated. If operating on those switches allows to connect/disconnect the line, then we should not propagate. If it is not enough, we should try to propagate.
But I'm not the user, so...

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java
```diff
@@ -1208,55 +1211,191 @@ public boolean connect(TerminalExt terminal, Predicate<? super SwitchImpl> isSwi
...
+    private Terminal getEndOfPathTerminal(TIntArrayList path) {
+        return graph.getVertexObject(getEndOfPathVertex(path));
+    }
+
+    private int getEndOfPathVertex(TIntArrayList path) {
+        // Get the terminal at the end of the path
+        int e = path.get(path.size() - 1);
+        int v1 = graph.getEdgeVertex1(e);
+        int v2 = graph.getEdgeVertex2(e);
+        return graph.getVertexObject(v1) != null ? v1 : v2;
```
flo-dup: This is a dangerous conclusion. The two last vertices might both have terminals, and in that case there's only a 50% probability that this is giving the end of path vertex

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java
```diff
@@ -1208,55 +1211,191 @@ public boolean connect(TerminalExt terminal, Predicate<? super SwitchImpl> isSwi
...
+        // Initialize the map
+        Map<LineImpl, ThreeSides> linesAndSidesConnectedToIncomingLine = new HashMap<>();
+
+        // Each path is visited, and we add to the list the line at the end
+        for (TIntArrayList pathFromLine : pathsFromLine) {
+            // Get the terminal at the end of the path
+            Terminal terminalEndOfPathFromBbs = getEndOfPathTerminal(pathFromLine);
+
+            // Connectables should only be lines by now, but in case...
+            if (terminalEndOfPathFromBbs.getConnectable() instanceof LineImpl line) {
```
flo-dup: I got with a busbar section here due to that, and it was silently ignoring it. In these kinds of cases, you'd better fail for the unexpected class (the class cast exception would be enough, but it might annoy sonar)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java
```diff
@@ -1208,55 +1211,191 @@ public boolean connect(TerminalExt terminal, Predicate<? super SwitchImpl> isSwi
...
+                linesAndSidesConnectedToIncomingLine.putIfAbsent(line, line.getTerminal1() == terminalEndOfPathFromBbs ? ThreeSides.TWO : ThreeSides.ONE);
```
flo-dup:
```suggestion
                linesAndSidesConnectedToIncomingLine.putIfAbsent(line, line.getSide(terminalEndOfPathFromBbs));
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java
```diff
@@ -1208,55 +1211,191 @@ public boolean connect(TerminalExt terminal, Predicate<? super SwitchImpl> isSwi
...
     }
 
+    private Map<LineImpl, ThreeSides> addOppositeLines(List<TIntArrayList> pathsFromLine) {
+        // Initialize the map
+        Map<LineImpl, ThreeSides> linesAndSidesConnectedToIncomingLine = new HashMap<>();
+
+        // Each path is visited, and we add to the list the line at the end
+        for (TIntArrayList pathFromLine : pathsFromLine) {
+            // Get the terminal at the end of the path
+            Terminal terminalEndOfPathFromBbs = getEndOfPathTerminal(pathFromLine);
```
flo-dup: We do not start the path finding from bbs but from line
```suggestion
            Terminal terminalEndOfPathFromLine = getEndOfPathTerminal(pathFromLine);
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java
```diff
@@ -1208,55 +1211,191 @@ public boolean connect(TerminalExt terminal, Predicate<? super SwitchImpl> isSwi
...
-                SwitchImpl sw = graph.getEdgeObject(e);
-                if (SwitchPredicates.IS_OPEN.test(sw)) {
-                    // Since the paths were constructed using the method checkNonClosableSwitches, only operable switches can be open
-                    switchForConnection.add(sw);
+
+            // Get the shortest paths by busbar sections
+            List<TIntArrayList> pathsByBusbarSection = propagateConnectionIfNeeded ? getShortestPathByBusbarSection(paths) : List.of(paths.get(0));
+
+            // Loop on the paths until a connectable path is found
+            for (TIntArrayList path : pathsByBusbarSection) {
```
flo-dup: we don't need to think by busbar section, do we?
- without propagation it seems it was forgotten by the previous code, which only took the shortest path (`paths.get(0)`) even if it was a all closed-switches one - but it's outside the scope of this PR
- with propagation it's not needed as we don't reason by busbar section but by incoming line, so in fact we might be calling the same methods in each element of the loop

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Line.java
```diff
@@ -33,4 +35,20 @@ default boolean isTieLine() {
...
+
+    default boolean connect(Predicate<Switch> isTypeSwitchToOperate, boolean propagateDisconnectionIfNeeded) {
+        return connect(isTypeSwitchToOperate, null, propagateDisconnectionIfNeeded);
+    }
+
+    boolean connect(Predicate<Switch> isTypeSwitchToOperate, ThreeSides side, boolean propagateDisconnectionIfNeeded);
+
+    boolean disconnect(boolean propagateDisconnectionIfNeeded);
```
flo-dup, line 39: It's strange to call these methods disconnect as calling line.getTerminal(side).isConnected() might return true even after a successful disconnect with propagation. We should at least add a warning in the javadoc.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Line.java
flo-dup: After discussion with the user, the current logic is (almost) ok. Here's the detail for the 3 usecases mentioned
- close it and stop like it's done now
- for bus breaker voltage levels in a mixed topology network, it might be interesting to change the logic, but let's wait for such networks and the corresponding need to come
- open it and stop like it's done now

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java
flo-dup: After discussion with @rolnico, amending the first bullet point:
- without propagation a singleton list is used, and indeed it's not needed to reason by busbar section (if we're there in the code that means the line's disconnected, that is, the bus for the line in the bus view is null; if the line is connected to a busbar section that results in a bus != null: one feeder + one busbar section is enough)

This leads to following simplification (canUse and addSwitches can now easily be merged to avoid duplicating the path loop):
```java
        if (!paths.isEmpty()) {
            // Get the shortest path
            TIntArrayList path = paths.get(0);
            if (addSwitchesToCloseForConnectToSet(path, isSwitchOperable, propagateConnectionIfNeeded, node, connectionElementsContainer)) {
                return true;
            }
        } else if (propagateConnectionIfNeeded) {
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java
flo-dup: I would say in the current PR as it's quite easy to add propagation to two-winding transformers:
- changing here the NodeBreakerTopologyModel::isLine
- duplicate the methods added in Line interface to TwoWindingsTransformer interface
- move the aded code of LineImpl in AbstractConnectableBranch

Besides, for the unplanned outage usecase, if the propagation only goes through lines, the resulting network state is wrong

flo-dup:
After some discussion with @benoitflo which requested this feature for disconnection:
- two usecases: planned outage / unplanned outage
- planned outage: disconnecting with openable switches = any non-fictitious switch, propagating if needed through voltage levels, fictitious or not: some corner-case situations might need propagation through non-fictitious voltage levels
- unplanned outage: disconnecting with openable switches = any non-fictitious switch, propagating if needed through voltage levels (fictitious or not)
- we *could* consider that in a fictitious voltage level we propagate without trying to connect first, but there's no identified usecase where this will be useful: in all known usecases this would lead to same result, so let's forget it

For connection, it would be better to keep the same kind of logic

So the logic implemented in this PR is the good one! if we except the `terminal.getVoltageLevel().isFictitious()` test in connectAllTerminals.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java
flo-dup: An idea to avoid this: couldn't we add the voltage levels traversed in the ConnectionElementsContainer? and then check if we have already traversed it?

rolnico: Maybe the lines more than the voltage levels?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java
rolnico: No, I want exactly the opposite: if the terminal is on side1, I want to put side 2 in the map

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Line.java
```diff
@@ -33,4 +35,20 @@ default boolean isTieLine() {
...
         return IdentifiableType.LINE;
     }
+
+    boolean connect(boolean propagateDisconnectionIfNeeded);
+
+    default boolean connect(Predicate<Switch> isTypeSwitchToOperate, boolean propagateDisconnectionIfNeeded) {
+        return connect(isTypeSwitchToOperate, null, propagateDisconnectionIfNeeded);
+    }
+
+    boolean connect(Predicate<Switch> isTypeSwitchToOperate, ThreeSides side, boolean propagateDisconnectionIfNeeded);
```
colineplqt:
```suggestion
    default boolean connect(Predicate<Switch> isTypeSwitchToOperate, boolean propagateConnectionIfNeeded) {
        return connect(isTypeSwitchToOperate, null, propagateConnectionIfNeeded);
    }

    boolean connect(Predicate<Switch> isTypeSwitchToOperate, ThreeSides side, boolean propagateConnectionIfNeeded);
```

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/TwoWindingsTransformer.java
```diff
@@ -247,4 +248,20 @@ default void setRatioTapPositionToSolvedTapPosition() {
     default void setPhaseTapPositionToSolvedTapPosition() {
         this.getOptionalPhaseTapChanger().ifPresent(TapChanger::applySolvedValues);
     }
+
+    boolean connect(boolean propagateDisconnectionIfNeeded);
+
+    default boolean connect(Predicate<Switch> isTypeSwitchToOperate, boolean propagateDisconnectionIfNeeded) {
```
colineplqt:
```suggestion
    default boolean connect(Predicate<Switch> isTypeSwitchToOperate, boolean propagateConnectionIfNeeded) {
        return connect(isTypeSwitchToOperate, null, propagateConnectionIfNeeded);
    }

    boolean connect(Predicate<Switch> isTypeSwitchToOperate, ThreeSides side, boolean propagateConnectionIfNeeded);
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractConnectableBranch.java
```diff
@@ -268,4 +271,58 @@ public Overload checkTemporaryLimits2(LimitType type) {
     public double getValueForLimit(Terminal t, LimitType type) {
         return BranchUtil.getValueForLimit(t, type);
     }
+
+    public boolean connect(boolean propagateDisconnectionIfNeeded) {
+        return connect(SwitchPredicates.IS_NONFICTIONAL_BREAKER, null, propagateDisconnectionIfNeeded);
+    }
```
colineplqt:
```suggestion
    public boolean connect(boolean propagateConnectionIfNeeded) {
        return connect(SwitchPredicates.IS_NONFICTIONAL_BREAKER, null, propagateConnectionIfNeeded);
    }
```

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractConnectionDisconnectionPropagationTest.java
```diff
@@ -0,0 +1,1995 @@
...
+        Line line2 = network.getLine("L1_2");
+        List<Line> lines = List.of(line1, line2);
+
+        // Check that the 2 lines are connected on both sides
+        lines.forEach(line -> line.getTerminals().forEach(terminal -> assertTrue(terminal.isConnected())));
+
+        // Try a disconnection without propagation - it should fail
+        assertFalse(line1.disconnect(SwitchPredicates.IS_NONFICTIONAL, false));
+
+        // Check that the 3 lines are still connected on both sides
```
colineplqt, line 134:
```suggestion
        // Check that the 2 lines are still connected on both sides
```

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractConnectionDisconnectionPropagationTest.java
```diff
@@ -0,0 +1,1995 @@
...
+
+        // Lines
+        Line line1 = network.getLine("L1_1");
+        Line line2 = network.getLine("L1_2");
+        List<Line> lines = List.of(line1, line2);
+
+        // Check that the 2 lines are connected on both sides
+        lines.forEach(line -> line.getTerminals().forEach(terminal -> assertTrue(terminal.isConnected())));
+
+        // Check the switches that will be opened are close for now
```
colineplqt, line 181: Typo
```suggestion
        // Check the switches that will be opened are closed for now
```

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractConnectionDisconnectionPropagationTest.java
```diff
@@ -0,0 +1,1995 @@
...
+        switchList.forEach(s -> assertFalse(network.getSwitch(s).isOpen()));
+
+        // Disconnect the line works with the propagation
+        assertTrue(line1.disconnect(SwitchPredicates.IS_NONFICTIONAL, true));
+
+        // Check that the 3 lines are disconnected on one side only
```
colineplqt, line 187:
```suggestion
        // Check that the 2 lines are disconnected on one side only
```

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractConnectionDisconnectionPropagationTest.java
```diff
@@ -0,0 +1,1995 @@
...
+        // Lines
+        Line line1 = network.getLine("L1_1");
+        Line line2 = network.getLine("L1_2");
+        Line line4 = network.getLine("L1_4");
+        List<Line> lines = List.of(line1, line2, line4);
+
+        // Check that the 3 lines are connected on both sides
+        lines.forEach(line -> line.getTerminals().forEach(terminal -> assertTrue(terminal.isConnected())));
+
+        // Check the switches that will be opened are close for now
```
colineplqt, line 213:
```suggestion
        // Check the switches that will be opened are closed for now
```

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractConnectionDisconnectionPropagationTest.java
```diff
@@ -0,0 +1,1995 @@
...
```
colineplqt, line 246:
```suggestion
        // Check the switches that will be opened are closed for now
```

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractConnectionDisconnectionPropagationTest.java
```diff
@@ -0,0 +1,1995 @@
...
+
+        // Lines
+        Line line1 = network.getLine("L1_1");
+        Line line2 = network.getLine("L1_2");
+        List<Line> lines = List.of(line1, line2);
+
+        // Check that the 2 lines are connected on both sides
+        lines.forEach(line -> line.getTerminals().forEach(terminal -> assertTrue(terminal.isConnected())));
+
+        // Check the switches that will be opened are close for now
```
colineplqt, line 313:
```suggestion
        // Check the switches that will be opened are closed for now
```

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractConnectionDisconnectionPropagationTest.java
```diff
@@ -0,0 +1,1995 @@
...
+        Line line2 = network.getLine("L1_2");
+        Line line3 = network.getLine("L1_3");
+        Line line4 = network.getLine("L1_4");
+        Line line5 = network.getLine("L1_mid");
+        List<Line> lines = List.of(line1, line2, line3, line4, line5);
+
+        // Check that all the lines are connected on both sides
+        lines.forEach(line -> line.getTerminals().forEach(terminal -> assertTrue(terminal.isConnected())));
+
+        // Check the switches that will be opened are close for now
```
colineplqt, line 347:
```suggestion
        // Check the switches that will be opened are closed for now
```

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractConnectionDisconnectionPropagationTest.java
```diff
@@ -0,0 +1,1995 @@
...
+        switchList.forEach(s -> assertFalse(network.getSwitch(s).isOpen()));
+
+        // Disconnect the line works with the propagation
+        assertTrue(line1.disconnect(SwitchPredicates.IS_NONFICTIONAL, true));
+
+        // Check that the 3 lines are disconnected on one side only
```
colineplqt, line 353:
```suggestion
        // Check that the 5 lines are disconnected
```

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractConnectionDisconnectionPropagationTest.java
```diff
@@ -0,0 +1,1995 @@
...
+     *         |                          |                          |
+     *         2                          3                          1
+     *         |                          |                          |
+     * L1_1_DISCONNECTOR_2_0     L1_2_DISCONNECTOR_3_0           D_L1_BBS4
+     *         |                          |                          |
+     *         -------------0--------------                      ----0-----
+     *                  BBS3 (VL3)                               BBS4 (VL4)
+     * </pre>
+     */
+    private Network createNetworkWithThreeLinesConnectedOnTwoFictitiousVoltageLevel() {
```
colineplqt, line 1267: I think this network is not used actually. Should it be removed or did you intend to write a test with it?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractConnectableBranch.java
```diff
@@ -268,4 +271,58 @@ public Overload checkTemporaryLimits2(LimitType type) {
...
+
+    public boolean connect(Predicate<Switch> isTypeSwitchToOperate, ThreeSides side, boolean propagateConnectionIfNeeded) {
```
colineplqt: Would it not be better to use TwoSides as this is for Branches and then call `toThreeSides` when you get into something more general? Same for disconnect then.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
colineplqt:
```suggestion
     * @param terminals list of the terminals that should be disconnected. For a connectable, it should be its own
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
```diff
@@ -123,29 +132,38 @@ static boolean disconnectAllTerminals(Identifiable<?> identifiable, List<? exten
...
-            if (terminal.isConnected()
-                && terminal.getVoltageLevel().getTopologyKind() == BUS_BREAKER) {
-                // At this point, isNowDisconnected should always stay true but let's be careful
-                isNowDisconnected = isNowDisconnected && terminal.disconnect(isSwitchOpenable);
-            }
+        if (operateSwitchesFromHere) {
+            // Open all switches on node-breaker terminals to disconnect the network element
+            localConnectionElementsContainer.switchesToOperate().forEach(sw -> sw.setOpen(true));
+
+            // Connect all bus-breaker terminals
```
colineplqt:
```suggestion
            // Disconnect all bus-breaker terminals
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
colineplqt, line 32: I think it would be nice to have in the javadoc the new parameters of this function, to understand better how it works. Also in disconnectAllTerminals.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
```diff
@@ -40,19 +39,23 @@ private ConnectDisconnectUtil() {
...
+        ConnectionElementsContainer localConnectionElementsContainer = new ConnectionElementsContainer(new HashSet<>(), new HashSet<>(), new HashSet<>(connectionElementsContainer.connectables()));
 
         // We try to connect each terminal
         for (Terminal terminal : terminals) {
             // Check if the terminal is already connected
-            if (terminal.isConnected()) {
+            if (terminal.isConnected() && !propagateConnectionIfNeeded) {
                 NetworkReports.alreadyConnectedIdentifiableTerminal(reportNode, identifiable.getId());
                 continue;
             } else {
```
colineplqt, line 64: I am not sure that this condition should be added here. If we call connect on something that is already connected with propagation true, then the return of this method will be `FAILURE` even though we would expect `NO_CHANGE_NEEDED` because isNowConnected will return false.

I think you want to return `NO_CHANGE_NEEDED` if the terminal is already connected and the propagation did not connect or disconnect anything, no?

rolnico: If every terminal is already connected and we don't want to try to propagate (`isAlreadyConnected` stays `true`), then we return `NO_CHANGE_NEEDED`.

However, in some cases were a terminal is already connected (for example in the case of a tee-point), we might want to propagate the connection to connect further down the network (on the other sides of the tee-point, for example).

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractConnectionDisconnectionPropagationTest.java
rolnico: ... maybe? :smile:

Issue #3459, opened 2025-05-29
PSS/E transformer and AC line names are not imported


pjanecek0:
### Describe the current behavior

When importing data from PSS/E format, names/identifiers for transformers and AC lines are not being imported. These names are present in the original PSS/E file but missing in the imported network model.

### Describe the expected behavior

Names from PSS/E should be preserved during import for both transformers and AC lines. - If the name is empty, it should be automatically generated using bus names and CTK.

PR #3474, opened 2025-06-12
Psse import names

pjanecek0:

#3459 

#3459 

Transformer names and lines (non-transformer branches) names are imported from PSS/E data. For versions 32 or 33, line names are generated by combining bus names and CKT fields.

- [ ] Yes
- [x] No

**Other information**:


psse/psse-converter/src/main/java/com/powsybl/psse/converter/LineConverter.java
```diff
@@ -148,6 +148,32 @@ private static int getStatus(Line line) {
...
+                int busId = bus.getI();
+                if (busId == psseLine.getI()) {
+                    busI = bus;
+                } else if (busId == psseLine.getJ()) {
+                    busJ = bus;
+                }
+                if (busI != null && busJ != null) {
+                    break;
+                }
+            }
```
olperr1: In `PsseImporter.convert()`, a `Map<Integer, PsseBus>` called `busNumToPsseBus` is created to easily have access to a `PsseBus` from its number:

https://github.com/powsybl/powsybl-core/blob/4c014da6a34f8c1a090f2bba2598fde2a6791641/psse/psse-converter/src/main/java/com/powsybl/psse/converter/PsseImporter.java#L185

It can be used as a new `LineConverter` constructor's argument to avoid this costly loop on every bus of the network. See for instance how it is used in `TransformerConverter`:

https://github.com/powsybl/powsybl-core/blob/4c014da6a34f8c1a090f2bba2598fde2a6791641/psse/psse-converter/src/main/java/com/powsybl/psse/converter/TransformerConverter.java#L68

psse/psse-converter/src/main/java/com/powsybl/psse/converter/LineConverter.java
```diff
@@ -148,6 +148,32 @@ private static int getStatus(Line line) {
...
+            if (busI != null && busJ != null) {
+                name = String.format("%s_%s_%s", busI.getName().trim(), busJ.getName().trim(), psseLine.getCkt());
+            }
+        }
+        if (version.major() == V35) {
```
olperr1:
```suggestion
        } else if (version.major() == V35) {
```

psse/psse-converter/src/test/resources/ExampleVersion32.xiidm
```diff
@@ -76,16 +76,16 @@
...
+    <iidm:line id="L-1-2-1 " name="Bus 1_Bus 2_1 " r="3.6907272" x="11.2683348" g1="0.0" b1="1.3862633900441084E-4" g2="0.0" b2="1.3862633900441084E-4" voltageLevelId1="VL1" bus1="B1" connectableBus1="B1" voltageLevelId2="VL2" bus2="B2" connectableBus2="B2"/>
+    <iidm:line id="L-1-4-1 " name="Bus 1_Bus 4_1 " r="10.2894732" x="42.475737599999995" g1="0.0" b1="1.2917454316320101E-4" g2="0.0" b2="1.2917454316320101E-4" voltageLevelId1="VL1" bus1="B1" connectableBus1="B1" voltageLevelId2="VL4" bus2="B4" connectableBus2="B4"/>
+    <iidm:line id="L-2-3-1 " name="Bus 2_Bus 3_1 " r="8.9487756" x="37.7014068" g1="0.0" b1="1.1499684940138626E-4" g2="0.0" b2="1.1499684940138626E-4" voltageLevelId1="VL2" bus1="B2" connectableBus1="B2" voltageLevelId2="VL3" bus2="B3" connectableBus2="B3"/>
+    <iidm:line id="L-3-3-1 " name="" r="11.066468400000002" x="33.578380800000005" g1="0.0" b1="8.926696072253729E-5" g2="0.0" b2="8.926696072253729E-5" voltageLevelId1="VL3" bus1="B3" connectableBus1="B3" voltageLevelId2="VL3" bus2="B3" connectableBus2="B3"/>
```
olperr1: Here, `"Bus 3_Bus 3_ 1"` is expected.
Using the map in `LineConverter.convert()` should fix this.

psse/psse-converter/src/test/resources/twoSubstations_rev35.xiidm
```diff
@@ -101,7 +101,7 @@
             </iidm:nodeBreakerTopology>
             <iidm:load id="B5-L1 " loadType="UNDEFINED" p0="6.428" q0="-2.763" node="26"/>
         </iidm:voltageLevel>
-        <iidm:twoWindingsTransformer id="T-2-3-3 " r="0.01602" x="0.63918" g="0.0" b="0.0" ratedU1="220.0" ratedU2="15.0" voltageLevelId1="VL2" node1="3" voltageLevelId2="VL3" node2="11" selectedOperationalLimitsGroupId1="DEFAULT" selectedOperationalLimitsGroupId2="DEFAULT">
+        <iidm:twoWindingsTransformer id="T-2-3-3 " name="T2D-S2-220-15                           " r="0.01602" x="0.63918" g="0.0" b="0.0" ratedU1="220.0" ratedU2="15.0" voltageLevelId1="VL2" node1="3" voltageLevelId2="VL3" node2="11" selectedOperationalLimitsGroupId1="DEFAULT" selectedOperationalLimitsGroupId2="DEFAULT">
```
olperr1, line 109: Note for reviewers: this file "proves" that the 2 and 3-winding transformer names, as well as the line names in rev35, are indeed imported.

psse/psse-model/src/main/java/com/powsybl/psse/model/PsseVersioned.java
```diff
@@ -28,6 +29,11 @@ public void setModel(PssePowerFlowModel model) {
         this.model = Objects.requireNonNull(model);
     }
 
+    @JsonIgnore
+    public PssePowerFlowModel getModel() {
```
olperr1: This file's changes can be reverted if the map is used in `LineConverter.convert()`.

psse/psse-converter/src/test/resources/IEEE_14_buses_duplicate_ids_rev35.xiidm
```diff
@@ -152,25 +152,25 @@
...
-    <iidm:line id="L-9-14-1 " r="24.2068284" x="51.4911672" g1="0.0" b1="0.0" g2="0.0" b2="0.0" voltageLevelId1="VL4-7-9" bus1="B9" connectableBus1="B9" voltageLevelId2="VL14" bus2="B14" connectableBus2="B14"/>
-    <iidm:line id="L-10-11-1 " r="15.625601999999999" x="36.577810799999995" g1="0.0" b1="0.0" g2="0.0" b2="0.0" voltageLevelId1="VL10" bus1="B10" connectableBus1="B10" voltageLevelId2="VL11" bus2="B11" connectableBus2="B11"/>
-    <iidm:line id="L-12-13-1 " r="42.0720048" x="38.0651472" g1="0.0" b1="0.0" g2="0.0" b2="0.0" voltageLevelId1="VL12" bus1="B12" connectableBus1="B12" voltageLevelId2="VL13" bus2="B13" connectableBus2="B13"/>
-    <iidm:line id="L-13-14-1 " r="32.5519092" x="66.2769288" g1="0.0" b1="0.0" g2="0.0" b2="0.0" voltageLevelId1="VL13" bus1="B13" connectableBus1="B13" voltageLevelId2="VL14" bus2="B14" connectableBus2="B14"/>
+    <iidm:line id="L-1-2-1 " name=" " r="3.6907272" x="11.2683348" g1="0.0" b1="1.3862633900441084E-4" g2="0.0" b2="1.3862633900441084E-4" voltageLevelId1="VL1" bus1="B1" connectableBus1="B1" voltageLevelId2="VL2" bus2="B2" connectableBus2="B2"/>
```
olperr1: To be consistent with the PSS/E file, the line names in this file should be `""` and not `" "`.
It is not detected by the unit test (see `.ignoreWhitespace()` in [`ComparisonUtils.assertXmlEquals`](https://github.com/powsybl/powsybl-core/blob/4c014da6a34f8c1a090f2bba2598fde2a6791641/commons-test/src/main/java/com/powsybl/commons/test/ComparisonUtils.java#L35)), but fixing this could prevent further questions about this inconsistency.

zamarrenolm:
Conflicts in xiidm files are minor and can be solved by:
- adding the new flag added for tap changers: `loadTapChangingCapabilities=”<value>”`.
- adding new hvdc lines in the example IEEE_14_buses_badly_defined_controlled_buses.xiidm

psse/psse-converter/src/test/resources/IEEE_14_buses_duplicate_ids_rev35.xiidm
marqueslanauja: To be consistent it is also necessary to modify the default value assigned to `name` when it is not provided in V35 cases. To do this, the following code inside the `PsseNonTransformerBranch` class

```
   @Parsed(defaultNullRead = " ")
    @Revision(since = 35)
    private String name = " ";

```

must be changed to 

```
   @Parsed(defaultNullRead = "")
    @Revision(since = 35)
    private String name = "";

```

olperr1:
I wonder if for transformers, it is not better to set the name to `null` when the PSS/E name is empty (white chars only).
This way, `transformer.getNameOrId()` will return the id instead of a white String...
(Same for lines in rev. 35)

What do you think about it?

marqueslanauja:
> I wonder if for transformers, it is not better to set the name to `null` when the PSS/E name is empty (white chars only). This way, `transformer.getNameOrId()` will return the id instead of a white String... (Same for lines in rev. 35)
> 
> What do you think about it?

It's ok. If we don't set the name when it's an empty string, it will be null, and `getNameOrId()` will return the `id`, which is almost always more informative than an empty name, which isn't useful.

psse/psse-converter/src/main/java/com/powsybl/psse/converter/LineConverter.java
```diff
@@ -190,10 +200,25 @@ static void update(Network network, PssePowerFlowModel psseModel) {
...
+    private String getName() {
+        String name = "";
+        if (version.major() == V32 || version.major() == V33) {
+            PsseBus busI = busNumToPsseBus.get(psseLine.getI());
+            PsseBus busJ = busNumToPsseBus.get(psseLine.getJ());
+            if (busI != null && busJ != null) {
+                name = String.format("%s_%s_%s", busI.getName().trim(), busJ.getName().trim(), psseLine.getCkt());
+            }
+        } else if (version.major() == V35) {
+            name = psseLine.getName();
```
olperr1: When the name contains only white chars, it would be better to return `null` (see comments on this PR).

psse/psse-converter/src/main/java/com/powsybl/psse/converter/TransformerConverter.java
```diff
@@ -100,6 +100,8 @@ private void createTwoWindingsTransformer() {
         // As vn2 is used to convert to eu, only the ratio remains to be applied
         TapChanger tapChangerAdjustedYsh = tapChangerAdjustmentAfterMovingShuntAdmittanceBetweenRatioAndTransmissionImpedance(tapChangerAdjustedRatio);
 
+        String name = psseTransformer.getName();
```
olperr1: When the name contains only white chars, it would be better to use `null` (see comments on this PR).

psse/psse-converter/src/main/java/com/powsybl/psse/converter/TransformerConverter.java
```diff
@@ -200,12 +203,15 @@ private void createThreeWindingsTransformer() {
         // move ysh between w1 and z
         TapChanger tapChanger1AdjustedYsh = tapChangerAdjustmentAfterMovingShuntAdmittanceBetweenRatioAndTransmissionImpedance(tapChanger1);
 
+        String name = psseTransformer.getName();
```
olperr1: When the name contains only white chars, it would be better to return `null` (see comments on this PR).

olperr1 (changes requested):

psse/psse-converter/src/main/java/com/powsybl/psse/converter/TransformerConverter.java
```diff
@@ -247,6 +249,14 @@ private void createThreeWindingsTransformer() {
         defineOperationalLimits(twt, voltageLevel1.getNominalV(), voltageLevel2.getNominalV(), voltageLevel3.getNominalV());
     }
 
+    private String getNameOrNull() {
+        String name = psseTransformer.getName();
+        if (name.trim().isEmpty()) {
```
olperr1: There's no check in `PsseTransformer.setName(String)` to ensure that the name is not `null`.
I think it would be safer to consider that the name can be `null` here:

```suggestion
        if (name != null && name.trim().isEmpty()) {
```

olperr1 (changes requested):

psse/psse-converter/src/main/java/com/powsybl/psse/converter/LineConverter.java
```diff
@@ -190,10 +200,28 @@ static void update(Network network, PssePowerFlowModel psseModel) {
...
+            if (name.trim().isEmpty()) {
+                name = null;
+            }
+        }
+        return name;
+    }
+
```
marqueslanauja, line 221: My proposal:

 Standardize the method name
Set name to null initially
 Build the name only if the bus names are not empty or blank

  ```
private String getNameOrNull() {
        String name = null;
        if (version.major() == V32 || version.major() == V33) {
            PsseBus busI = busNumToPsseBus.get(psseLine.getI());
            PsseBus busJ = busNumToPsseBus.get(psseLine.getJ());
            if (busI != null && busJ != null && !busI.getName().trim().isEmpty() && !busJ.getName().trim().isEmpty()) {
                name = String.format("%s_%s_%s", busI.getName().trim(), busJ.getName().trim(), psseLine.getCkt());
            }
        } else if (version.major() == V35) {
            name = psseLine.getName();
            if (name.trim().isEmpty()) {
                name = null;
            }
        }
        return name;
    }
```

File: psse/psse-converter/src/main/java/com/powsybl/psse/converter/LineConverter.java [hunk 12]
```diff
@@ -190,10 +200,30 @@ static void update(Network network, PssePowerFlowModel psseModel) {
...
+            }
+        } else if (version.major() == V35) {
+            if (isNotEmeptyName(psseLine.getName())) {
+                return psseLine.getName();
+            }
+        }
+        return null;
+    }
+
+    private static boolean isNotEmeptyName(String name) {
```
olperr1: There is a typo in this method's name:
```patch
- isNotEmeptyName
+ isNotEmptyName
```

olperr1 (changes requested):

marqueslanauja (approved):

olperr1:
@pjanecek0: Thank you for your contribution!

PR #3477, opened 2025-06-16
Add countryTo support for CGMES boundary lines

### LINKED ISSUE #3255: Information on both countries for unpaired dangling lines

Currently, an unpaired dangling line is associated with a pairing_key and **one** voltage level id, for which we can identify the substation and the country. However, in the ENTSOE EQ BD xml file, the pairing key is associated with a connectivity node that has a country from and country to. I don't think it is possible to access this information in pypowsybl right now, so we get only one of both country whereas it would be useful to know where the unpaired dangling line is supposed to go. 

It could be useful to get the information on the country from and to for the unpaired dangling lines. For instance, we could have a dataframe with for each pairing key of the dangling lines of the network, the country from and country to associated with it.

### Describe the motivation

This feature is needed for the CGMES dataset validation step in METIS 4, to identify for each unpaired dangling lines, whether it is flow that goes outside the merging zone or some mistake of modelisation.

### Extra Information

The feature will need a development in powsybl-core as well as in pypowsybl (handling the migration to the new IIDM version).
**A change in IIDM is needed for this feature.**
- An attribute should be added in the IIDM Boundary object to store the "countryTo" piece of information, optional field.
- If the CGMES EQ BD file is provided, then the fields `toEndIsoCode` and `fromEndIsoCode` are present on the Xnode and should be read and put in the new IIDM `Boundary.countryTo` attribute. Depending to the node from which the `DanglingLine` is built, we need to select the good end iso code that represents the "countryTo" to take: it can be either the `toEndIsoCode` or `fromEndIsoCode`.

cphili:

https://github.com/powsybl/powsybl-core/issues/3255

New feature

When importing CGMES EQ Boundary data, PowSyBl creates DanglingLine/BoundaryLine objects from XNode information. The local country can be inferred from the IIDM network topology (VoltageLevel / Substation), but the remote country information available in CGMES (fromEndIsoCode / toEndIsoCode) is lost during the conversion process.

This PR introduces support for remote boundary country information through a new optional attribute:
Boundary.getCountryTo()
BoundaryLine.getCountryTo()

CGMES conversion : fromEndIsoCode and toEndIsoCode from CGMES XNode objects are now available during boundary conversion. The remote country is extracted and assigned to the created BoundaryLine.  
IIDM model : a new optional attribute has been added (Country countryTo)

- [ ] Yes
- [x] No

The feature is backward compatible, countryTo is optional and defaults to null for existing networks and files (older IIDM versions remain readable without modification). For IIDM XML exports, version >=1.19 should be used if persistence of countryTo is required.


geofjamg (changes requested):
Wy do we need to country from and to? We just need to country on internal point, the one on network side is already accessible from substation.

cphili:
> Wy do we need to country from and to? We just need to country on internal point, the one on network side is already accessible from substation.

Hi @geofjamg, I've resumed the work on this PR and pushed the latest updates.
Following your remark, the implementation now only stores the remote country information, as the local/network-side country is already available through the substation.

PR #3478, opened 2025-06-16
Add checks about tap position depending on validation level.

colineplqt:

No

Inconsistency fix

The behavior when setting the tap position of a tap changer is inconsistent.
- in the adder, an invalid tap position is allowed for the EQUIPMENT level;
- in AbstractTapChanger, calling setTapPosition(...) with an invalid value throws an exception regardless the level;
- checkRtc(...) and checkPtc(...) of ValidationUtil don't check that the position is valid, it only checks that it can't be null in SSH level.

Now
- in the adder, an invalid tap position is allowed for the EQUIPMENT level;
- in AbstractTapChanger, calling setTapPosition(...) with an invalid value throws an exception only if the validation level is SSH; Otherwise, if allowed by the minimum validation level, it sets the tap position and level to EQ.
- checkRtc(...) and checkPtc(...) of ValidationUtil check that the position is valid 

- [ ] Yes
- [X] No

### DIFF
--- a/iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractTapChangerTest.java
+++ b/iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractTapChangerTest.java
@@ -785,7 +785,7 @@ public abstract class AbstractTapChangerTest {
     }
 
     @Test
-    public void createInvalidTapChangerEquipmentLevel() {
+    public void createInvalidRatioTapChangerEquipmentLevel() {
         network.setMinimumAcceptableValidationLevel(ValidationLevel.EQUIPMENT);
         assertEquals(ValidationLevel.STEADY_STATE_HYPOTHESIS, network.getValidationLevel());
         createRatioTapChangerWith3Steps(0, 1, true, false, 10.0, 1.0, terminal);
@@ -793,6 +793,19 @@ public abstract class AbstractTapChangerTest {
         rtc.setTapPosition(4);
         assertEquals(ValidationLevel.EQUIPMENT, network.getValidationLevel());
         assertEquals(4, rtc.getTapPosition());
+    }
+
+    @Test
+    public void createInvalidPhaseTapChangerEquipmentLevel() {
+        createPhaseTapChangerWith2Steps(1, 0, false,
+            PhaseTapChanger.RegulationMode.FIXED_TAP, 1.0, 1.0, terminal);
+        PhaseTapChanger ptc = network.getTwoWindingsTransformer("twt").getPhaseTapChanger();
+        assertThrows(ValidationException.class, () -> ptc.setTapPosition(4));
+        network.setMinimumAcceptableValidationLevel(ValidationLevel.EQUIPMENT);
+        assertEquals(ValidationLevel.STEADY_STATE_HYPOTHESIS, network.getValidationLevel());
+        ptc.setTapPosition(4);
+        assertEquals(ValidationLevel.EQUIPMENT, network.getValidationLevel());
+        assertEquals(4, network.getTwoWindingsTransformer("twt").getPhaseTapChanger().getTapPosition());
     }
 
     private void createRatioTapChangerWith3Steps(int low, int tap, boolean load, boolean regulating,


olperr1:
The following code, using powsybl-open-loadflow, shows the problem before this PR:
```java
        Network network = FourSubstationsNodeBreakerFactory.create();
        network.setMinimumAcceptableValidationLevel(ValidationLevel.EQUIPMENT);

        // Affect a wrong tap position (outside the bounds)
        TwoWindingsTransformer twt = network.getTwoWindingsTransformer("TWT");
        twt.newPhaseTapChanger(twt.getPhaseTapChanger())
                        .setTapPosition(1000) // Invalid tap position
                        .add();

        // This invalidates the validation level => it will be recomputed.
        network.getGenerator("GH1").setTargetP(85.3570);

        List<Contingency> contingencies = List.of(
                Contingency.line("LINE_S2S3")
        );
        SecurityAnalysisResult result = SecurityAnalysis.find("OpenLoadFlow").run(network, contingencies).getResult();
        result.getPostContingencyResults();
```
<br/>


- Without changing the generator's targetP, the validation level is EQUIPMENT and the security analysis cannot be run:
```
java.util.concurrent.CompletionException: com.powsybl.commons.PowsyblException: Only STEADY STATE HYPOTHESIS validation level of the network is supported
(...)
Caused by: com.powsybl.commons.PowsyblException: Only STEADY STATE HYPOTHESIS validation level of the network is supported
	at com.powsybl.openloadflow.network.impl.LfNetworkLoaderImpl.load(LfNetworkLoaderImpl.java:1267)
(...)
```

- When the targetP is changed, the check on the validation level succeeds, but the simulation crashes when the tap position is retrieved by open-loadflow:
```
java.util.concurrent.CompletionException: com.powsybl.iidm.network.ValidationException: 2 windings transformer 'TWT': incorrect tap position 1000 [0, 32]
(...)
Caused by: com.powsybl.iidm.network.ValidationException: 2 windings transformer 'TWT': incorrect tap position 1000 [0, 32]
	at com.powsybl.iidm.network.impl.AbstractTapChanger.throwIncorrectTapPosition(AbstractTapChanger.java:255)
	at com.powsybl.iidm.network.impl.AbstractTapChanger.getStep(AbstractTapChanger.java:145)
(...)
```

olperr1:
We discussed this PR with @geofjamg: in fact, out-of-bounds values should not be allowed, even in EQUIPMENT level.
This level is meant to allow having network with missing values, not inconsistencies.
=>
- `NaN` is allowed in EQUIPMENT level and should be checked in ValidationUtil (it's indeed the case);
- out-of-bounds values should always be rejected. So there's no need to check if the value is inside the bounds in ValidationUtil.

~~If think that for clarity, we could also remove the check that is performed by `getStep()`.~~

@colinepiloquet: Sorry for the wrong analysis of the problem (https://github.com/powsybl/powsybl-core/pull/3442#discussion_r2139371703) that is at the origin of this PR!

colineplqt:
> We discussed this PR with @geofjamg: in fact, out-of-bounds values should not be allowed, even in EQUIPMENT level. This level is meant to allow having network with missing values, not inconsistencies. =>
> 
>     * `NaN` is allowed in EQUIPMENT level and should be checked in ValidationUtil (it's indeed the case);
> 
>     * out-of-bounds values should always be rejected. So there's no need to check if the value is inside the bounds in ValidationUtil.
> 
> 
> If think that for clarity, we could also remove the check that is performed by `getStep()`.
> 
> @colinepiloquet: Sorry for the wrong analysis of the problem ([#3442 (comment)](https://github.com/powsybl/powsybl-core/pull/3442#discussion_r2139371703)) that is at the origin of this PR!


But actually in the adder, out of bound values can be allowed if the minimum level is EQUIPMENT no?

olperr1:
> > We discussed this PR with @geofjamg: in fact, out-of-bounds values should not be allowed, even in EQUIPMENT level. This level is meant to allow having network with missing values, not inconsistencies. =>
> > ```
> > * `NaN` is allowed in EQUIPMENT level and should be checked in ValidationUtil (it's indeed the case);
> > 
> > * out-of-bounds values should always be rejected. So there's no need to check if the value is inside the bounds in ValidationUtil.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If think that for clarity, we could also remove the check that is performed by `getStep()`.
> > @colinepiloquet: Sorry for the wrong analysis of the problem ([#3442 (comment)](https://github.com/powsybl/powsybl-core/pull/3442#discussion_r2139371703)) that is at the origin of this PR!
> 
> But actually in the adder, out of bound values can be allowed if the minimum level is EQUIPMENT no?

Currently, it is allowed to import out-of-bounds values for this field in EQUIPMENT level. But this level was originally introduced to allow importing network without some values (mainly the ones that will be filled in by a state estimator).
In the actual state, if you import an out-of-bounds tap position, you will have an exception when you call `getCurrentStep()`, even in EQUIPMENT level.

PR #3494, opened 2025-07-02
Add OperationalLimitDirectionKind.high property to OperationalLimitsGroups

cphili:

Currently CGMES import only handles high limits, this is now reflected as a property OperationalLimitDirectionKind = "high" in the OperationalLimitsGroup.

We only accept high limits and this is not reflected anywhere.

Add OperationalLimitDirectionKind.high property to OperationalLimitsGroups.

- [ ] Yes
- [x] No


olperr1:
After a second though, we don't think this PR is really useful. We should first focus on how TC and TCT will be stored (which will have an impact on the way this data should be stored: in a property or using another way).

PR #3498, opened 2025-07-02
Migrate to Java 21

rolnico:

No

Dependency update

Java 17 is used

Java 21 is used

- [x] Yes
- [ ] No

PowSyBl-Core now only supports Java 21 and higher. Please check that your installed SDK is still compatible. If you are using Ubuntu 20.04 LTS and the preinstalled Maven version, you will need to upgrade your Maven version too at least up to a version 3.9.x.

**Other information**:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -15,7 +15,7 @@
     <parent>
         <groupId>com.powsybl</groupId>
         <artifactId>powsybl-parent</artifactId>
-        <version>20.2</version>
+        <version>23</version>
         <relativePath/>
     </parent>
 
@@ -89,7 +89,7 @@
     </modules>
 
     <properties>
-        <java.version>17</java.version>
+        <java.version>21</java.version>
 
         <argLine/> <!--  This is required for later correct replacement of argline -->
         <sonar.coverage.jacoco.xmlReportPaths>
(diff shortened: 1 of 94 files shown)


pom.xml
```diff
@@ -150,9 +150,22 @@
...
+        <profile>
+            <id>java21+</id>
+            <activation>
+                <jdk>[21,)</jdk>
+            </activation>
+            <properties>
+                <!-- In Java 21+, we have to add a parameter to allow dynamic agent loading (cf. powsybl-parent) -->
+                <enable-dynamic-agent-loading.argLine>-XX:+EnableDynamicAgentLoading</enable-dynamic-agent-loading.argLine>
+            </properties>
+        </profile>
```
olperr1, line 168: Can't we directly set the property, instead of setting it in a profile? We will always be in java21+ (it won't compile with lower versions).

olperr1 (changes requested):
The `README.md` file should also be updated: it indicates [the minimum JDK version to use](https://github.com/powsybl/powsybl-core/blob/4457e080ab3ad75e862e4712230bd00c6ba6c51a/README.md?plain=1#L48).

olperr1 (dismissed):

olperr1 (approved):

olperr1 (approved):

PR #3513, opened 2025-07-14
Psse: Support substations with buses at the same nominal voltage in different areas

marqueslanauja:

Bug fix

All the buses inside a substation at the same nominal voltage are included in the same voltage level.

Only the buses inside a substation that have the same nominal voltage and are in the same area are included in the same voltage level.

- [x] Yes
- [ ] No

The `ContainersMapping#create(...)` method has an additional parameter: an integer function to assign an area to the bus: `ToIntFunction<Integer> busToControlArea`

**Other information**:


zamarrenolm (approved):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/ContainersMapping.java
```diff
@@ -57,21 +57,22 @@ public String getSubstationId(String voltageLevelId) {
...
-                                                  Predicate<B> branchToIsTransformer, Function<Set<Integer>, String> busesToVoltageLevelId,
-                                                  IntFunction<String> substationNumToId) {
+    @Deprecated(since = "6.9.0")
+    public static <N, B> ContainersMapping create(List<N> buses, List<B> branches,
+                                                 ToIntFunction<N> busToNum, ToIntFunction<B> branchToNum1, ToIntFunction<B> branchToNum2,
+                                                 Predicate<B> branchToIsZeroImpedance, Predicate<B> branchToIsTransformer,
+                                                 ToDoubleFunction<Integer> busToNominalVoltage, Function<Set<Integer>, String> busesToVoltageLevelId,
+                                                 Function<Set<Integer>, String> busesToSubstationId) {
         throw new PowsyblException("Deprecated. Not used anymore");
     }
```
olperr1: Since the deprecated method doesn't work anymore, I think it would be better to remove it. This way, developers will have a compilation failure instead of a problem when executing their code if they directly call this method.

marqueslanauja: Done

zamarrenolm (approved):

PR #3517, opened 2025-07-17
Fix standby automaton adder checks

geofjamg:

No

Bug fix

When creating a standby automaton extension with incomplete attibutes, we can a strange error message. This is because setpoint and threshold are intiliazed to 0 in the adder instead of nan (undefined)

We can a correct error message and also a ValidationException which is more consistent with the actual design.

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/extensions/AbstractStandbyAutomatonTest.java
+++ b/iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/extensions/AbstractStandbyAutomatonTest.java
@@ -10,6 +10,7 @@ package com.powsybl.iidm.network.tck.extensions;
 import com.powsybl.commons.PowsyblException;
 import com.powsybl.iidm.network.Network;
 import com.powsybl.iidm.network.StaticVarCompensator;
+import com.powsybl.iidm.network.ValidationException;
 import com.powsybl.iidm.network.VariantManager;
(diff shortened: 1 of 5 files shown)


iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/extensions/AbstractStandbyAutomatonTest.java
```diff
@@ -74,7 +75,20 @@ public void test() {
...
+    @Test
+    public void testIncompleteAdderBadException() {
+        Network network = SvcTestCaseFactory.create();
+        StaticVarCompensator svc = network.getStaticVarCompensator("SVC2");
+        assertNotNull(svc);
+
+        StandbyAutomatonAdder standbyAutomatonAdder = svc.newExtension(StandbyAutomatonAdder.class)
+                .withStandbyStatus(true);
+        var e = assertThrows(ValidationException.class, standbyAutomatonAdder::add);
+        // instead of "Static var compensator 'SVC2': Inconsistent low (0.0) and high (0.0) voltage thresholds"
```
rolnico:
```suggestion
```

rolnico: I think we should not keep comments about differences with previous versions, right?

rolnico (changes requested):
A small change but else lgtm

Issue #3487, opened 2025-06-23
DC detailed model topology processing


jeandemanged:
### Describe the current behavior

DC Detailed model is introduced in #3439 in beta, but there is no computation of DC buses, DC islands/components. Cannot store DC bus voltages solution.

### Describe the expected behavior

Computation of DC buses, DC islands/components.
Ability to store DC bus voltages.

PR #3520, opened 2025-07-18
DC detailed model topology processing

jeandemanged:

#3487 

- [X] Yes

Custom IIDM implementations' maintainers should:
- write their own implementations of `DcBus` and `AbstractDcComponent`;
- implement the following methods:
  - in `DcNode` implementations:
    - `double getV();`
    - `DcNode setV(double v);`
    - `DcBus getDcBus();`
    - `int getDcTerminalCount();`
    - `List<DcTerminal> getDcTerminals();`
    - `Stream<DcTerminal> getDcTerminalStream();`
    - `int getConnectedDcTerminalCount();`
    - `List<DcTerminal> getConnectedDcTerminals();`
    - `Stream<DcTerminal> getConnectedDcTerminalStream();`
    - `Iterable<DcGround> getDcGrounds();`
    - `Stream<DcGround> getDcGroundStream();`
    - `Iterable<DcLine> getDcLines();`
    - `Stream<DcLine> getDcLineStream();`
    - `Iterable<LineCommutatedConverter> getLineCommutatedConverters();`
    - `Stream<LineCommutatedConverter> getLineCommutatedConverterStream();`
    - `Iterable<VoltageSourceConverter> getVoltageSourceConverters();`
    - `Stream<VoltageSourceConverter> getVoltageSourceConverterStream();`
    - `void visitConnectedEquipments(DcTopologyVisitor visitor);`
    - `void visitConnectedOrConnectableEquipments(DcTopologyVisitor visitor);`
  - in `DcTerminal` implementations:
    - `DcBus getDcBus();`
  - in `Network` implementations:
    - `Collection<Component> getDcComponents();`
    - `DcBus getDcBus(String id);`
    - `Iterable<DcBus> getDcBuses();`
    - `Stream<DcBus> getDcBusStream();`
    - `int getDcBusCount();`

**Other information**:


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/components/AbstractComponentsManager.java
```diff
@@ -43,24 +50,34 @@ public void update() {
...
+        Map<String, Integer> acBusId2num = new HashMap<>();
+        Map<String, Integer> dcBusId2num = new HashMap<>();
+        List<Bus> num2AcBus = new ArrayList<>();
+        List<DcBus> num2DcBus = new ArrayList<>();
+        if (ac) {
+            for (Bus bus : getNetwork().getBusView().getBuses()) {
+                num2AcBus.add(bus);
+                acBusId2num.put(bus.getId(), num);
+                num++;
+                numAc++;
```
olperr1: It would be clearer to rename this variable `nbAcBuses` and to compute it after the `if (ac)` block:
```java
int nbAcBuses = num2AcBus.size();
```

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/components/AbstractComponentsManager.java
```diff
@@ -89,38 +111,89 @@ public C getComponent(int num) {
...
+                Bus bus1 = acDcConverter.getTerminal1().getBusView().getBus();
+                DcBus dcBus1 = acDcConverter.getDcTerminal1().getBus();
+                DcBus dcBus2 = acDcConverter.getDcTerminal2().getBus();
+                addToAdjacencyList(bus1, dcBus1, acBusId2num, dcBusId2num, adjacencyList);
+                addToAdjacencyList(bus1, dcBus2, acBusId2num, dcBusId2num, adjacencyList);
+                acDcConverter.getTerminal2().ifPresent(t2 -> {
+                    Bus bus2 = t2.getBusView().getBus();
+                    addToAdjacencyList(bus2, dcBus1, acBusId2num, dcBusId2num, adjacencyList);
+                    addToAdjacencyList(bus2, dcBus2, acBusId2num, dcBusId2num, adjacencyList);
+                });
```
olperr1, line 192: Shouldn't we also add `addToAdjacencyList(bus1, bus2, acBusId2num, adjacencyList);`?
In the `if (ac)` condition, I think.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/components/AbstractComponentsManager.java
```diff
@@ -43,24 +50,34 @@ public void update() {
...
         int num = 0;
-        Map<String, Integer> id2num = new HashMap<>();
-        List<Bus> num2bus = new ArrayList<>();
-        for (Bus bus : getNetwork().getBusView().getBuses()) {
-            num2bus.add(bus);
-            id2num.put(bus.getId(), num);
-            num++;
+        int numAc = 0;
+        Map<String, Integer> acBusId2num = new HashMap<>();
+        Map<String, Integer> dcBusId2num = new HashMap<>();
```
olperr1: There's no need to use 2 maps: the IIDM model ensures that there is no collision between AC and DC bus ids.
Having a single map makes the code more readable (and less error-prone).

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcNodeImpl.java
```diff
@@ -67,6 +84,54 @@ public DcNode setNominalV(double nominalV) {
...
+        ValidationUtil.checkModifyOfRemovedEquipment(this.id, this.removed, "v");
+        int variantIndex = getNetwork().getVariantIndex();
+        double oldValue = this.v.set(variantIndex, v);
+        String variantId = getNetwork().getVariantManager().getVariantId(variantIndex);
+        getNetwork().getListeners().notifyUpdate(this, "v", variantId, oldValue, v);
+        return this;
+    }
+
+    public void setConnectedComponentNumber(int connectedComponentNumber) {
+        int variantIndex = networkRef.get().getVariantIndex();
```
olperr1, line 125: Is this on purpose that no check is performed on `removed` in this method and in `setDcComponentNumber(...)`?

In `setV(...)`, `checkModifyOfRemovedEquipement(...)` is called. Also `getNetwork()` (which performs the control too) is used instead of `networkRef.getNetwork()`.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
```diff
@@ -82,6 +85,32 @@ void removeFromSubnetworks(String subnetworkId) {
...
+
+    @Override
+    protected void attachDcTopologyModel(DcTopologyModel dcTopologyModel) {
+        this.dcTopologyModel = dcTopologyModel;
+        dcTopologyModel.updateRef(ref, subnetworkRef);
+    }
+
+    protected void flattenDcTopologyModel(DcTopologyModel otherDcTopologyModel) {
+        if (otherDcTopologyModel.getDcBusStream().findFirst().isPresent()) {
+            // TODO merge elements
```
olperr1: Do not forget this TODO.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
```diff
@@ -1391,6 +1473,7 @@ public void flatten() {
             // Those which are already present in the current network are not transferred.
             transferExtensions(subnetwork, this, true);
             transferProperties(subnetwork, this, true);
+            flattenDcTopologyModel(subnetwork.detachDcTopologyModel());
```
olperr1: Maybe `mergeDcTopologyModel` will be a more intuitive name?

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/DcTopologyTest.java
```diff
@@ -0,0 +1,103 @@
...
+        assertSame(n11.getDcBus(), net1.getDcBus("n11_dcBus"));
+        assertSame(n12.getDcBus(), net1.getDcBus("n12_dcBus"));
+        n11.remove();
+        assertDcBusesAre(net1, List.of("n12_dcBus"));
+    }
+
+    private void assertDcBusesAre(Network network, List<String> expected) {
+        assertEquals(
+                expected.stream().sorted(Comparator.naturalOrder()).toList(),
+                network.getDcBusStream().map(Identifiable::getId).sorted(Comparator.naturalOrder()).toList()
```
olperr1: The comparator can be removed. `.sorted()` sorts a Stream according to natural order.

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/DcTopologyTest.java
```diff
@@ -0,0 +1,103 @@
...
+        assertDcBusesAre(net1, List.of("n11_dcBus"));
+        assertDcBusesAre(net2, List.of("n21_dcBus", "n22_dcBus"));
+        net1 = net1.detach();
+        assertDcBusesAre(merged, List.of());
+        assertDcBusesAre(net1, List.of("n11_dcBus"));
+        assertDcBusesAre(net2, List.of("n21_dcBus", "n22_dcBus"));
+
+        // flatten
+        merged = Network.merge(net1, net2);
+        // TODO merged.flatten();
```
olperr1: Don't forget the TODO here.

olperr1 (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/components/AbstractComponentsManager.java
jeandemanged: indeed, thanks, fixed

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/components/AbstractComponentsManager.java
jeandemanged: indeed, thanks, fixed

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/components/AbstractComponentsManager.java
jeandemanged: not in the "ac only" condition (AC synchronous component), but rather on the ac && dc condition (connected component) where it was indeed missing, thank you.
Sounds counterintuitive, so I added the following explanation on why it is not added in the ac-only adjacency but rather the ac+dc adjacency:
```
            // Note that AC/DC converters with two AC terminals are not included here (AC synchronous component):
            // The converter does not synchronize the 2 AC terminals together,
            // the converter does not impose phase or frequency alignment.
            // The adjacency of the two AC terminals is however added in the case of ac && dc (connected component).
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
jeandemanged: I hesitated long time on this ...
From a code standpoint it is indeed clearly merging, but I was afraid of possible confusion with network merging vs. network flattening... then we would have to call mergeDcTopologyModel in flatten() but not in merge()...

I cannot come up with a good alternative name avoiding "merge", so I guess I will rename as you suggest and document/comment to avoid any confusion for future generations.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/components/AbstractComponentsManager.java
olperr1: Thanks for the explanation and the comment :+1:

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
olperr1: You are right, there can be confusion with either of these 2 names.
What about `addDcTopologyModelContent(DcTopologyModel)` or `addContent(DcTopologyModel)`?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcTopologyModel.java
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+    void addAndDetachSubnetworkDcTopologyModel(SubnetworkImpl subnetwork) {
+        // called by network flatten, the subnetwork DC topology model is merged into this DC topology model.
+        subnetwork.getDcNodeStream().forEach(dcNode -> addDcNodeToTopology((DcNodeImpl) dcNode));
+        subnetwork.getDcSwitchStream().forEach(dcSwitch -> addDcSwitchToTopology((DcSwitchImpl) dcSwitch, dcSwitch.getDcNode1().getId(), dcSwitch.getDcNode2().getId()));
+
+        subnetwork.detachDcTopologyModel(); // to remove refs
+        // purge cache for all variants
+        variants = new VariantArray<>(networkRef, DcTopologyModel.VariantImpl::new);
```
olperr1: This is not a critical problem, but `variants` will be reinitialized several times in "flatten" (one for every subnetworks).

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcTopologyModel.java
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+    /**
+     * DCBus topology cache
+     */
+    private static final class DcBusCache {
```
olperr1: This class can be defined as a record.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcTopologyModel.java
```diff
@@ -0,0 +1,349 @@
...
+            }
+            var node = dcNodeSet.stream().min(Comparator.comparing(DcNodeImpl::getId)).orElseThrow();
+            String dcBusId = Identifiables.getUniqueId(node.getId() + "_dcBus", getNetwork().getIndex()::contains);
+            String dcBusName = node.getOptionalName().orElse(null);
+            return new DcBusImpl(networkRef, subnetworkRef, dcBusId, dcBusName, dcNodeSet);
+        }
+
+        private void updateCache() {
+            if (variants.get().cache != null) {
+                return;
```
olperr1: The method is called `updateCache()` but nothing is done when the cache is not null.
I think it would be better to change its signature to:
```java
private DcBusCache getCache() {
```
(or `getInitializedCache()`, or `getValidCache()`, if you think that someone could be tempted to call a `getCache()` method in `invalidatedCache()`.)

... And then to change all its usage.
For instance:
```patch
private Collection<DcBusImpl> getDcBuses() {
-    updateCache();
-    return variants.get().cache.getDcBuses();
+    return getCache().getDcBuses();
}
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcTopologyModel.java
```diff
@@ -0,0 +1,349 @@
...
+    public Stream<DcBus> getDcBusStream() {
+        return calculatedDcBusTopology.getDcBuses().stream().map(Function.identity());
+    }
+
+    public DcBusImpl getDcBus(String id) {
+        return calculatedDcBusTopology.getDcBus(id);
+    }
+
+    public DcBusImpl getDcBusOfDcNode(String dcNodeId) {
+        DcNodeImpl dcNode = (DcNodeImpl) getNetwork().getDcNode(dcNodeId);
```
olperr1: Do we really need to use the `DcNodeImpl` as a key in `DcBusCache.mapping`?
We can use the node ids. This will avoid having to retrieve the `DcNodeImpl` here... unless you want to ensure that the given id is actually a valid DC node id.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcTopologyModel.java
```diff
@@ -0,0 +1,349 @@
...
+        @Override
+        public DcTopologyModel.VariantImpl copy() {
+            return new DcTopologyModel.VariantImpl();
+        }
+    }
+
+    public void invalidateCache() {
+        calculatedDcBusTopology.invalidateCache();
+        getNetwork().getConnectedComponentsManager().invalidate();
+        getNetwork().getDcComponentsManager().invalidate();
```
olperr1, line 321: Can you add a comment in this method to indicate that not calling `getNetwork().getSynchronousComponentsManager().invalidate();` is not an oversight?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/SubnetworkImpl.java
```diff
@@ -1020,7 +999,7 @@ public Network detach() {
         NetworkImpl detachedNetwork = new NetworkImpl(getId(), getNameOrId(), getSourceFormat());
         transferExtensions(this, detachedNetwork);
         transferProperties(this, detachedNetwork);
-        detachedNetwork.attachDcTopologyModel(this.detachDcTopologyModel());
+        detachedNetwork.attachDcTopologyModel(this.detachDcTopologyModel(), rootNetworkRef, ref);
```
olperr1, line 1005:
```suggestion
        detachedNetwork.attachDcTopologyModel(this.detachDcTopologyModel(), rootNetworkRef, ref); // refs are updated below
```

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/DcTopologyTest.java
```diff
@@ -0,0 +1,358 @@
...
+        PowsyblException e = assertThrows(PowsyblException.class, () -> action.accept(dcBus));
+        assertEquals("DcBus has been invalidated", e.getMessage());
+    }
+
+    private static void assertDcBusesAre(Network network, List<String> expected) {
+        assertEquals(
+                expected.stream().sorted().toList(),
+                network.getDcBusStream().map(Identifiable::getId).sorted().toList()
+        );
+        assertEquals(network.getDcBusCount(), expected.size());
```
olperr1: Expected value comes usually first:
```suggestion
        assertEquals(expected.size(), network.getDcBusCount());
```

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/DcTopologyTest.java
```diff
@@ -0,0 +1,358 @@
...
+    private static void assertComponent(Component component, int expectedNum, List<String> expectedAcBuses, List<String> expectedDcBuses) {
+        Objects.requireNonNull(component);
+        Objects.requireNonNull(expectedAcBuses);
+        Objects.requireNonNull(expectedDcBuses);
+        assertEquals(expectedNum, component.getNum());
+        assertEquals(expectedAcBuses.size() + expectedDcBuses.size(), component.getSize());
+        assertEquals(expectedAcBuses.stream().sorted().toList(), component.getBusStream().map(Identifiable::getId).sorted().toList());
+        assertEquals(expectedAcBuses.stream().sorted().toList(), StreamSupport.stream(component.getBuses().spliterator(), false).map(Identifiable::getId).sorted().toList());
+        assertEquals(expectedDcBuses.stream().sorted().toList(), component.getDcBusStream().map(Identifiable::getId).sorted().toList());
+        assertEquals(expectedDcBuses.stream().sorted().toList(), StreamSupport.stream(component.getDcBuses().spliterator(), false).map(Identifiable::getId).sorted().toList());
```
olperr1: Can you introduce local variables for `expectedAcBuses.stream().sorted().toList()` and `expectedDcBuses.stream().sorted().toList()`? There's no need to recompute them.

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/DcTopologyTest.java
```diff
@@ -0,0 +1,358 @@
...
+        DcBus n12DcBus = net1.getDcBus("n12_dcBus");
+        assertDcBusContainsDcNodes(n11DcBus, List.of(n11));
+        assertDcBusContainsDcNodes(n12DcBus, List.of(n12));
+        assertTrue(Double.isNaN(n11DcBus.getV()));
+        assertTrue(Double.isNaN(n11.getV()));
+        assertTrue(Double.isNaN(n12DcBus.getV()));
+        assertTrue(Double.isNaN(n12.getV()));
+        n11DcBus.setV(501.);
+        n12DcBus.setV(-502.);
+        assertEquals(501., n11DcBus.getV(), 1e-3);
```
olperr1: Can you define an `EPSILON` constant (=`1e-3`)?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcNodeImpl.java
jeandemanged: yes on purpose, like in ConfiguredBusImpl which is a close cousin of DcNode. Controls can be skipped since no user would be able to reach this, this is not available by API, only available to topology processor or via DcBus. 

setXxxxComponentNumber : Unless a bug, the topology processor should not call on removed DcNode-s.
getXxxxComponent : Unless a bug, would never by called by DcBus-es (DcBus would have been invalidated, and checkValidity would throw before).

or am I missing something ?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcTopologyModel.java
jeandemanged: indeed.
Could be changed to have a dedicated separated method doing it only once at the end of NetworkImpl#flatten(), but most probably not worth it given that this re-init is cheap.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/components/DcComponent.java
jeandemanged, line 1: was added just to be in line with their cousin ConnectedComponent and SynchronousComponent in same package. But not used at all within powsybl-core.
I do not know whether these files:
* are dead code to be removed
* or are needed by other iIDM implementations

https://sonarcloud.io/code?id=com.powsybl%3Apowsybl-core&selected=com.powsybl%3Apowsybl-core%3Aiidm%2Fiidm-api%2Fsrc%2Fmain%2Fjava%2Fcom%2Fpowsybl%2Fiidm%2Fnetwork%2Fcomponents

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcTopologyModel.java
jeandemanged: meanwhile changed to purge the cache for all variants at the end of NetworkImpl#flatten()

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractDcConnectable.java
```diff
@@ -61,9 +61,12 @@ public void remove() {
 
         network.getIndex().remove(this);
 
+        dcTerminals.forEach(dcTerminal -> {
+            ((AbstractNetwork) getParentNetwork()).getDcTopologyModel().detach(dcTerminal);
+            dcTerminal.remove();
+        });
```
olperr1, line 67: This should also be added in the `remove()` method of `AbstractAcDcConverter`.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcNodeImpl.java
olperr1: OK, then we can keep it like this for consistency.

olperr1 (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/components/DcComponent.java
olperr1: I have no idea about this...

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractDcConnectable.java
jeandemanged: many thanks for catching this

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcTopologyVisitor.java
```diff
@@ -0,0 +1,28 @@
...
+
+    default void visitDcGround(DcGround dcGround) {
+        // empty default implementation
+    }
+
+    default void visitDcLine(DcLine dcLine, TwoSides side) {
+        // empty default implementation
+    }
+
+    // TODO update following TerminalNumber PR https://github.com/powsybl/powsybl-core/pull/3533
```
olperr1: A comment just to remember this TODO…

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractDcTopologyTest.java
```diff
@@ -0,0 +1,630 @@
...
+        assertEquals(List.of("dcGroundFr"), asSortedIds(dcNodeFrMid.getDcGrounds()));
+        assertEquals(List.of("dcGroundFr"), asSortedIds(dcNodeFrMid.getDcGroundStream()));
+        assertEquals(List.of(), asSortedIds(dcNodeGbMid.getDcGrounds())); // because DC ground not connected
+        assertEquals(List.of(), asSortedIds(dcNodeGbMid.getDcGroundStream())); // because DC ground not connected
+
+        // DC Lines
+        assertEquals(List.of("dcLineSegmentFrPosA", "dcLineSegmentPosA"), asSortedIds(dcNodePosA1.getDcLines()));
+        assertEquals(List.of("dcLineSegmentFrPosA", "dcLineSegmentPosA"), asSortedIds(dcNodePosA1.getDcLineStream()));
+        assertEquals(List.of(), asSortedIds(dcNodeFrMid.getDcLines()));
+        assertEquals(List.of(), asSortedIds(dcNodeFrMid.getDcLines()));
```
olperr1:
```suggestion
        assertEquals(List.of(), asSortedIds(dcNodeFrMid.getDcLineStream()));
```

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractDcTopologyTest.java
```diff
@@ -0,0 +1,630 @@
...
+        assertEquals(List.of(), asSortedIds(dcBusGbMid.getDcGrounds())); // because DC ground not connected
+        assertEquals(List.of(), asSortedIds(dcBusGbMid.getDcGroundStream())); // because DC ground not connected
+
+        // DC Lines
+        assertEquals(List.of("dcLineSegmentFrPosA", "dcLineSegmentPosA"), asSortedIds(dcBusPosA1.getDcLines()));
+        assertEquals(List.of("dcLineSegmentFrPosA", "dcLineSegmentPosA"), asSortedIds(dcBusPosA1.getDcLineStream()));
+        assertEquals(List.of("dcLineSegmentFrPosA", "dcLineSegmentFrPosB"), asSortedIds(dcBusFrPos.getDcLines()));
+        assertEquals(List.of("dcLineSegmentFrPosA", "dcLineSegmentFrPosB"), asSortedIds(dcBusFrPos.getDcLineStream()));
+        assertEquals(List.of(), asSortedIds(dcBusFrMid.getDcLines()));
+        assertEquals(List.of(), asSortedIds(dcBusFrMid.getDcLines()));
```
olperr1:
```suggestion
        assertEquals(List.of(), asSortedIds(dcBusFrMid.getDcLineStream()));
```

olperr1 (changes requested):

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractDcTopologyTest.java
jeandemanged: damn me ... thanks

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcTopologyVisitor.java
jeandemanged: done

PR #3531, opened 2025-07-29
Remove deprecaed withMessageTemplae and allow to add a message templa…

### LINKED ISSUE #3369: [ReportNode] Remove deprecated withMessageTemplate(key, template)

Presence of the deprecated withMessageTemplate(key, template)

Deprecated method removed. Keep the possibiliy to still add a message for a root repor node. The new method `withName` on root report node builder allows it.

### Describe the motivation

Cleaning

### Extra Information

A project using `ReportNode` API will no longer be able to create reports with the `withMessageTemplate(key, messageTemplate)`. The project must replace the use of this method by the new one `withMessageTemplate(key)`.

A `RootReportNode` can be build with a "name' that allows to have a specific message for it. It will not be translated in a locale though.

To make this work:
- create a new `ResourceBundle`, implementing the Powsybl `ReportResourceBundle` interface
- create the `reports.properties` bundle file to ensure the key/message template correspondance, in the resource folder of a common module of the project for instance. Some scripts have been created [here](https://github.com/powsybl/powsybl-dev-tools/tree/main/scripts/reportNode) to help to generate the `reports.properties` file with the existing keys in the project.
- then, to ensure the no collision between the reports keys among all the repositories using the `ReportNode` API, the keys in the `reports.properties` must now be prefixed with a unique project related prefix (for instance for powsybl-open-loadflow the prefix for each key is `olf.`).
- replace all the old use of `withMessageTemplate(key, messageTemplate)` by the new method `withMessageTemplate(key)` with the new key.

alicecaron:
…te on root report nodes with prefix message internationalization

Fixes #3369 

- [ ] Yes
- [ ] No

**Other information**:


commons/src/main/java/com/powsybl/commons/report/ReportNodeRootBuilder.java
```diff
@@ -0,0 +1,23 @@
+/**
+ * Copyright (c) 2024, RTE (http://www.rte-france.com)
```
olperr1:
```suggestion
 * Copyright (c) 2025, RTE (http://www.rte-france.com)
```

commons/src/main/java/com/powsybl/commons/report/ReportNodeRootBuilderImpl.java
```diff
@@ -42,7 +42,14 @@ public ReportNode build() {
...
     @Override
-    public ReportNodeBuilder self() {
+    public ReportNodeRootBuilder self() {
         return this;
     }
+
+    @Override
+    public ReportNodeRootBuilder withName(String name) {
+        this.key = "core.report.name";
+        values.put("name", TypedValue.untyped(name));
```
olperr1: All the `"${name}"` occurrences should be replaced in `name` to avoid an infinite recursion:

> java.lang.IllegalStateException: Infinite loop in property interpolation of Report ${name}: name

You can for instance use `"$${name}"` (`$` is the escape character of `org.apache.commons.text.StringSubstitutor` which is used to generated the report's message).

Here is a unit test to check that this problem is fixed:
```java
    @Test
    void testRootReportNodeWithMessageTemplateNoInfiniteRecursion() {
        ReportNode root = ReportNode.newRootReportNode()
                .withName("My ${name} report")
                .withResourceBundles(TEST_BASE_NAME, PowsyblCoreReportResourceBundle.BASE_NAME)
                .build();
        assertDoesNotThrow(() -> root.getMessage());
    }
```

olperr1: It may be better to use a less generic name for the value (for instance `name` -> `reportName`), since it will be inherited (and therefore usable) by the child nodes.

olperr1 (changes requested):

olperr1:
> With the withName method, the ReportNode is losing the functional logging. A user should be able to understand the message. What's a report name? If it is meant to provide a process id for instance, the corresponding template should rather be "Process id 394u593srm9" than "Report: 394u593srm9".

Thanks for your advice!
Then we can replace the `withName(…)` method by `withProcessId(…)`, and change the template (name and message).
@flo-dup, @alicecaron, @Lisrte:  Is it OK for you?

flo-dup:
> Thanks for your advice! Then we can replace the `withName(…)` method by `withProcessId(…)`, and change the template (name and message). @flo-dup, @alicecaron, @Lisrte: Is it OK for you?

This isn't much different from the previous proposal. Indeed processId is more precise than name and therefore makes more sense. But this is still not a very functional logging. Plus this is feasible with current API, just add (or let the user add) a new entry in the bundle files and then the user uses current API to use this template with a value. There's no need to enrich the API a new method which is bypassing it.

commons/src/main/java/com/powsybl/commons/report/AbstractReportNodeAdderOrBuilder.java
```diff
@@ -22,21 +24,15 @@ public abstract class AbstractReportNodeAdderOrBuilder<T extends ReportNodeAdder
     protected String messageTemplate;
     protected boolean withTimestamp = false;
     protected String timestampPattern;
-    protected MessageTemplateProvider messageTemplateProvider = MessageTemplateProvider.EMPTY;
+    protected boolean strictMode = true;
+    protected MessageTemplateProvider messageTemplateProvider = MessageTemplateProvider.EMPTY_STRICT;
 
     @Override
     public T withMessageTemplateProvider(MessageTemplateProvider messageTemplateProvider) {
         this.messageTemplateProvider = Objects.requireNonNull(messageTemplateProvider);
```
flo-dup, line 32: The strict mode might have been defined before the call to withMessageTemplateProvider.
```suggestion
        this.messageTemplateProvider = Objects.requireNonNull(messageTemplateProvider);
        this.messageTemplateProvider.setStrictMode(strictMode);
```

commons/src/main/java/com/powsybl/commons/report/MessageTemplateProvider.java
```diff
@@ -14,18 +14,29 @@
...
+    MessageTemplateProvider EMPTY_STRICT = new EmptyMessageTemplateProvider(true);
+    MessageTemplateProvider EMPTY_LOOSE = new EmptyMessageTemplateProvider(false);
 
     ResourceBundle.Control NO_FALLBACK_CONTROL = ResourceBundle.Control.getNoFallbackControl(ResourceBundle.Control.FORMAT_DEFAULT);
 
+    void setStrictMode(boolean strictMode);
+
+    boolean isStrictMode();
+
     String getTemplate(String key, Locale locale);
```
flo-dup, line 55: Could you specify in the interface what does it mean if null is returned?

olperr1: I added some Javadoc about it.

PR #3439, opened 2025-05-12
iIDM v1.14: Add detailed DC model

jeandemanged:
**Please check if the PR fulfills these requirements**

- [ ] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] A PR or issue has been opened in all impacted repositories (if any)

**Does this PR already have an issue describing the problem?**

#3381

**What kind of change does this PR introduce?**

Feature

**What is the current behavior?**

We have simple HVDC iIDM model

**What is the new behavior (if this is a feature change)?**
Detailed DC model, including DC line segments, DC grounds, DC switches, metallic return, bipoles, multi-terminal, meshed etc ...

**Does this PR introduce a breaking change or deprecate an API?**
- [x] Yes

**If yes, please check if the following requirements are fulfilled**

- [x] The *Breaking Change* or *Deprecated* label has been added
- [x] The migration steps are described in the following section

**What changes might users need to make in their application due to this PR? (migration steps)**

Custom iIDM implementations must ... TBD

**Other information**:

PR #3533, opened 2025-07-30
AC/DC Converters have terminal number instead of side

jeandemanged:

No

Between bugfix and feature

Relates to the DC detailed model introduced as beta feature in iIDM v1.14 June 2025 release (#3439).

There were criticisms about usage of Side.ONE and Side.TWO for the AC/DC Converters terminals, because these aren't really sides as e.g. the sides of a line or a transformer.

A new enum TerminalNumber is introduced specifically for AC/DC converters, and is used for both the AC and the DC terminals of the AC/DC converter.

- [x] Yes

TODO

**Other information**:

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractAcDcConverter.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractAcDcConverter.java
@@ -76,27 +76,27 @@ abstract class AbstractAcDcConverter<I extends AcDcConverter<I>> extends Abstrac
     }
 
     @Override
-    public TwoSides getSide(Terminal terminal) {
+    public TerminalNumber getTerminalNumber(Terminal terminal) {
         Objects.requireNonNull(terminal);
         if (getTerminal1() == terminal) {
-            return TwoSides.ONE;
+            return TerminalNumber.ONE;
         } else if (getTerminal2().orElse(null) == terminal) {
-            return TwoSides.TWO;
+            return TerminalNumber.TWO;
         } else {
             throw new IllegalStateException("The terminal is not connected to this AC/DC converter");
         }
     }
 
     @Override
-    public Terminal getTerminal(TwoSides side) {
-        Objects.requireNonNull(side);
+    public Terminal getTerminal(TerminalNumber terminalNumber) {
+        Objects.requireNonNull(terminalNumber);
         ValidationUtil.checkAccessOfRemovedEquipment(this.id, this.removed, "terminal");
-        if (side == TwoSides.ONE) {
+        if (terminalNumber == TerminalNumber.ONE) {
             return getTerminal1();
-        } else if (side == TwoSides.TWO) {
+        } else if (terminalNumber == TerminalNumber.TWO) {
             return getTerminal2().orElseThrow(() -> new IllegalStateException("This AC/DC converter does not have a second AC Terminal"));
         }
-        throw new IllegalStateException("Unexpected side: " + side);
+        throw new IllegalStateException("Unexpected side: " + terminalNumber);
     }
 
     @Override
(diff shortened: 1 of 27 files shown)


olperr1:
In the current state (commit b495025), it is OK for me.
It just needs one or two additional unit tests (for instance, testing on a line terminal the `Terminal`'s methods dealing with `TerminalNumber`s) to pass the quality gate.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcTerminalImpl.java
jeandemanged, line 1: Note that I changed the listener update notification attributes as follows:
- `pX` -> `p_dcX`
- `iX` -> `i_dcX`
- `connectedX` -> `connected_dcX`

Otherwise in the case of AC/DC converters it is not possible to know whether the change notification relates to the AC or DC terminal.

Let me know if this should be a separate PR.

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractDcTerminalTest.java
```diff
@@ -114,4 +115,89 @@ private static void checkDcTerminalInMultiVariant(Network network, DcTerminal dc
...
+        converterDcTerminal1.setP(33.);
+        Mockito.verify(mockedListener, Mockito.times(1))
+                .onUpdate(converter, "p_dc1", VariantManagerConstants.INITIAL_VARIANT_ID, 3., 33.);
+
+        // ... AC/DC converter I1 // DC
+        converterDcTerminal1.setI(44.);
+        Mockito.verify(mockedListener, Mockito.times(1))
+                .onUpdate(converter, "i_dc1", VariantManagerConstants.INITIAL_VARIANT_ID, 4., 44.);
+
+        // ... AC/DC converter P1 // DC
```
olperr1:
```suggestion
        // ... AC/DC converter P1 // AC
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcTerminalImpl.java
olperr1: Having it in a dedicated PR will offer more visibility on this fix... but the DC modeling feature is still a beta one and I think developers already using it will have to look attentively at this PR in any case.
So I guess we can keep it here. I will just report this change in the migration guide.

jeandemanged: thanks for feedback, and thanks for the migration guide update too!

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractDcTerminalTest.java
jeandemanged: thanks, fixed

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractAcDcConverter.java
```diff
@@ -76,27 +76,27 @@ public Optional<Terminal> getTerminal2() {
...
+            return TerminalNumber.TWO;
         } else {
             throw new IllegalStateException("The terminal is not connected to this AC/DC converter");
         }
     }
 
     @Override
-    public Terminal getTerminal(TwoSides side) {
-        Objects.requireNonNull(side);
+    public Terminal getTerminal(TerminalNumber terminalNumber) {
```
zamarrenolm: Throwing an exception if terminal 2 does not exist implies that the user will always know in advance if the object has one or two terminals. Maybe this method should return an optional, to make explicit that not always both terminals exist ?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTerminal.java
```diff
@@ -42,8 +44,12 @@ abstract class AbstractTerminal implements TerminalExt {
 
     protected boolean removed = false;
 
-    AbstractTerminal(Ref<? extends VariantManagerHolder> network, ThreeSides side) {
+    AbstractTerminal(Ref<? extends VariantManagerHolder> network, ThreeSides side, TerminalNumber terminalNumber) {
+        if (side != null && terminalNumber != null) {
```
zamarrenolm, line 48: We are keeping in a single object two alternative and exclusive references (side and terminalNumber). I think this will force users of this class to introduce a lot of logic for handling both ways of expressing references ... Maybe this approach for solving the original issue may end-up with an over-kill implementation?

zamarrenolm:
Just for re-consideration: instead of having an optional (or null) second terminal for some ac/dc converters, that introduces generic complexity in user's code, shouldn't be easier to have explicit classes for ac/dc converters with one pole (ac terminal) and ac/dc converters with exactly two poles (ac terminals)? Maybe the potential code duplication (I'm assuming not too much) compensates for the complexity of handling invalid/optional returns when querying terminals

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTerminal.java
zamarrenolm: And, at the end, bot references (side and terminalNumber) are just numbers, 1, 2, 3  :-(

jeandemanged: > We are keeping in a single object two alternative and exclusive references (side and terminalNumber).

actually, three alternative and exclusive references, since there is also the (pre-existing) case of null side for e.g. all injections

jeandemanged: to summarize, Terminals may:

- Injections: without side, without terminal number
- Branches & 3WT: with side, without terminal number
- AC/DC converters: without side, with terminal number

If I have a magic wand I would change so that everybody has always and only TerminalNumber, but I would rather propose this for future discussion with the community, eventually for future iIDM v1.15 or later.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractAcDcConverter.java
jeandemanged: Changed to return an optional, let me know

PR #3542, opened 2025-08-08
Move R,X characteristics from LineCharaceristics to Branch interface

### LINKED ISSUE #3521: Incoherence between documentation and code on IIDM Branch
The interface `com.powsybl.iidm.network.Branch` javadoc don't match the code of Branch (and all 3 implementations _Line_, _TieLine_ and _TwoWindingsTransformer_):
* it say that `x` and `r` are properties of a branch, but the getters aren't declared in it but in `com.powsybl.iidm.network.LineCharacteristics` for _Line_+_TieLine_ and directly in _TwoWindingsTransformer_
* it say that `b1`, `b2`, `g1` and `g2` are also commons properties but it's only declared in _LineCharacteristics_ but not in _TwoWindingsTransformer_ where we have instead `b` and `g`.

So
1. is `x` & `r` really common properties of branches? If yes can it be declared in _Branch_ interface instead because we have otherwise to cast/instanceof on subtypes to access the getters...
2. the documentation need to be updated for other properties it seems.

alicecaron:

Fixes #3521

Refacto, no functionnal change. TieLine are still not able to update R and X, an error is thrown.

- [x] Yes
- [ ] No

- Implementations of Branch interface should now implement getR/setR/getX/setX methods
- Implementations of LineCharacteristics should no longer implement getR/getX methods
- Implementations of MutableLineCharacteristics should no longer implement setR/setX methods

**Other information**:

### DIFF
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/MutableLineCharacteristics.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/MutableLineCharacteristics.java
@@ -8,19 +8,51 @@
 package com.powsybl.iidm.network;
 
 /**
+ * A line with mutable characteristics.
+ *
+ * <p>
+ *  Characteristics
+ * </p>
+ *
+ * <table style="border: 1px solid black; border-collapse: collapse">
+ *     <thead>
+ *            <tr>
+ *                <td style="border: 1px solid black">G1</td>
+ *                <td style="border: 1px solid black">double</td>
+ *                <td style="border: 1px solid black">S</td>
+ *                <td style="border: 1px solid black">yes</td>
+ *                <td style="border: 1px solid black"> - </td>
+ *                <td style="border: 1px solid black">The first side shunt conductance</td>
+ *            </tr>
+ *            <tr>
+ *                <td style="border: 1px solid black">B1</td>
+ *                <td style="border: 1px solid black">double</td>
+ *                <td style="border: 1px solid black">S</td>
+ *                <td style="border: 1px solid black">yes</td>
+ *                <td style="border: 1px solid black"> - </td>
+ *                <td style="border: 1px solid black">The first side shunt susceptance</td>
+ *            </tr>
+ *            <tr>
+ *                <td style="border: 1px solid black">G2</td>
+ *                <td style="border: 1px solid black">double</td>
+ *                <td style="border: 1px solid black">S</td>
+ *                <td style="border: 1px solid black">yes</td>
+ *                <td style="border: 1px solid black"> - </td>
+ *                <td style="border: 1px solid black">The second side shunt conductance</td>
+ *            </tr>
+ *            <tr>
+ *                <td style="border: 1px solid black">B2</td>
+ *                <td style="border: 1px solid black">double</td>
+ *                <td style="border: 1px solid black">S</td>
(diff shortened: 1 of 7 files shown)


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/MutableLineCharacteristics.java
```diff
@@ -8,20 +8,52 @@
...
+ *            <tr>
+ *                <td style="border: 1px solid black">B2</td>
+ *                <td style="border: 1px solid black">double</td>
+ *                <td style="border: 1px solid black">S</td>
+ *                <td style="border: 1px solid black">yes</td>
+ *                <td style="border: 1px solid black"> - </td>
+ *                <td style="border: 1px solid black">The second side shunt susceptance</td>
+ *            </tr>
+ *     </tbody>
+ * </table>
```
NathanDissoubray: Why is this in `MutableLineCharacteristics` and not in `LineCharacteristics` ? SInce mutable extends from `LineCharacteristics`, maybe we could put the table in there and use `@InheritDoc` in `MutableLineCharacteristics`

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/TieLineImpl.java
```diff
@@ -115,6 +115,16 @@ public double getX() {
...
+    @Override
+    public TieLine setR(double r) {
+        throw new PowsyblException("Cannot update TieLine resistance");
+    }
+
+    @Override
+    public TieLine setX(double x) {
+        throw new PowsyblException("Cannot update TieLine reactance");
+    }
+
```
NathanDissoubray, line 127: I think explaining why it cannot be updated might be better:
"Cannot update TieLine resistance, it is the sum of the resistance of its two dangling lines, update the dangling lines instead".
That way people won't even have to go look at the tie line model in read the docs to understand why.

NathanDissoubray (commented):
Just a few questions, nothing directly related to the code.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/MutableLineCharacteristics.java
alicecaron: done

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/TieLineImpl.java
alicecaron: I am not sure about giving an absolute rule of how it works in this exception as it is very related to business rules.

NathanDissoubray: If that was in an interface I would agree, but since this in an implementation, can't we be more specific ?

olperr1: A `UnsupportedOperationException` is more suitable in this case.
And we can provide details:
`throw new PowsyblException("Cannot update TieLine reactance: it is computed from its boundary lines.");`

(Same thing for `setR(...)`, of course)

Issue #3427, opened 2025-04-28
Add an extensions blacklist for IIDM import


alicecaron:
### Describe the current behavior

The `iidm.import.xml.extensions` property is an optional property that defines the list of extensions that will be imported by the XIIDM importer. By default, all extensions are imported.

### Describe the expected behavior

Another `iidm.import.xml.without-extensions` property is an optional property that defines the list of extensions that will not be imported by the XIIDM importer. By default, all extensions are imported except if it is limited by the use of the `iidm.import.xml.extensions` property.

### Describe the motivation

Sometimes we want to import all the extensions except some of them, for that it is easier to propose a blacklist of extensions than having to list all the extensions to import except some of them.

PR #3559, opened 2025-09-01
Add filter list for import and export of extension. Implementation of feature https://github.com/powsybl/powsybl-core/issues/3427

snoopfab:
Describe the motivation
Sometimes we want to import all the extensions except some of them, for that it is easier to propose a blacklist of extensions than having to list all the extensions to import except some of them.
New Parameters :
For Import : iidm.import.xml.filtered-extensions (string list of extensions that will not be imported)
For export iidm.export.xml.filtered-extensions (string list of extensions that will not be exported
To avoid incoherence, the rule can be that :
if both iidm.import.xml.extensions and iidm.import.xml.filtered-extensions are used,
then ignore the iidm.import.xml.filtered-extensions has priority hence the extension will not be imported
Same behavior for export filter list.
Signed-off-by: Fabrice Buscaylet <fabrice.buscaylet@artelys.com>

### DIFF
--- a/iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractOptions.java
+++ b/iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractOptions.java
@@ -7,6 +7,7 @@
  */
 package com.powsybl.iidm.serde;
 
+import com.google.common.collect.Sets;
 import com.powsybl.commons.io.TreeDataFormat;
 
 import java.util.Objects;
@@ -19,15 +20,27 @@ import java.util.Set;
 public abstract class AbstractOptions<T> {
 
     protected Set<String> extensions;
+    /**
+     * Extensions to be loaded must be in the extensions set but must not belong to the filteredExtension Set
+     */
+    protected Set<String> filteredExtension = Sets.newHashSet();
 
     protected TreeDataFormat format = TreeDataFormat.XML;
 
     public abstract T setExtensions(Set<String> extensions);
 
+    public abstract T setFilteredExtensions(Set<String> extensions);
+
     public abstract T addExtension(String extension);
+
+    public abstract T addFilteredExtension(String extensionToBeFiltered);
 
     public Optional<Set<String>> getExtensions() {
         return Optional.ofNullable(extensions);
+    }
+
+    public Optional<Set<String>> getFilteredExtensions() {
+        return Optional.ofNullable(filteredExtension);
     }
 
     public boolean withNoExtension() {
@@ -54,6 +67,10 @@ public abstract class AbstractOptions<T> {
         return withAllExtensions() || extensions.contains(extensionName);
     }
 
+    public boolean withFilteredExtension(String extensionName) {
+        return filteredExtension != null && filteredExtension.contains(extensionName);
+    }
+
     public abstract boolean isThrowExceptionIfExtensionNotFound();
 
     public TreeDataFormat getFormat() {
@@ -64,4 +81,5 @@ public abstract class AbstractOptions<T> {
         this.format = Objects.requireNonNull(format);
         return (T) this;
     }
+
 }
--- a/iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTreeDataExporter.java
+++ b/iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTreeDataExporter.java
@@ -111,6 +111,7 @@ public abstract class AbstractTreeDataExporter implements Exporter {
     public static final String TOPOLOGY_LEVEL = "iidm.export.xml.topology-level";
     public static final String THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND = "iidm.export.xml.throw-exception-if-extension-not-found";
     public static final String EXTENSIONS_LIST = "iidm.export.xml.extensions";
+    public static final String EXTENSIONS_FILTERED_LIST = "iidm.export.xml.filtered-extensions";
     public static final String SORTED = "iidm.export.xml.sorted";
     public static final String VERSION = "iidm.export.xml.version";
     public static final String WITH_AUTOMATION_SYSTEMS = "iidm.export.xml.with-automation-systems";
@@ -126,6 +127,9 @@ public abstract class AbstractTreeDataExporter implements Exporter {
             Arrays.stream(TopologyLevel.values()).map(Enum::name).collect(Collectors.toList()));
     private static final Parameter THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND_PARAMETER = new Parameter(THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND, ParameterType.BOOLEAN, "Throw exception if extension not found", Boolean.FALSE);
     private static final Parameter EXTENSIONS_LIST_PARAMETER = new Parameter(EXTENSIONS_LIST, ParameterType.STRING_LIST,
+            "The list of extensions that will be ignored during export", null,
+            EXTENSIONS_SUPPLIER.get().getProviders().stream().map(ExtensionProvider::getExtensionName).collect(Collectors.toList()));
+    private static final Parameter EXTENSIONS_FILTERED_LIST_PARAMETER = new Parameter(EXTENSIONS_FILTERED_LIST, ParameterType.STRING_LIST,
             "The list of exported extensions", null,
             EXTENSIONS_SUPPLIER.get().getProviders().stream().map(ExtensionProvider::getExtensionName).collect(Collectors.toList()));
     private static final Parameter SORTED_PARAMETER = new Parameter(SORTED, ParameterType.BOOLEAN, "Sort export output file", Boolean.FALSE);
@@ -136,7 +140,7 @@ public abstract class AbstractTreeDataExporter implements Exporter {
     private static final List<Parameter> STATIC_PARAMETERS = List.of(INDENT_PARAMETER, WITH_BRANCH_STATE_VARIABLES_PARAMETER,
             ONLY_MAIN_CC_PARAMETER, ANONYMISED_PARAMETER, IIDM_VERSION_INCOMPATIBILITY_BEHAVIOR_PARAMETER,
             TOPOLOGY_LEVEL_PARAMETER, THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND_PARAMETER, EXTENSIONS_LIST_PARAMETER,
-            SORTED_PARAMETER, VERSION_PARAMETER, WITH_AUTOMATION_SYSTEMS_PARAMETER);
+            EXTENSIONS_FILTERED_LIST_PARAMETER, SORTED_PARAMETER, VERSION_PARAMETER, WITH_AUTOMATION_SYSTEMS_PARAMETER);
 
     private final ParameterDefaultValueConfig defaultValueConfig;
 
@@ -195,6 +199,7 @@ public abstract class AbstractTreeDataExporter implements Exporter {
                 .setTopologyLevel(TopologyLevel.valueOf(Parameter.readString(getFormat(), parameters, TOPOLOGY_LEVEL_PARAMETER, defaultValueConfig)))
                 .setThrowExceptionIfExtensionNotFound(Parameter.readBoolean(getFormat(), parameters, THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND_PARAMETER, defaultValueConfig))
                 .setExtensions(Parameter.readStringList(getFormat(), parameters, EXTENSIONS_LIST_PARAMETER, defaultValueConfig) != null ? new HashSet<>(Parameter.readStringList(getFormat(), parameters, EXTENSIONS_LIST_PARAMETER, defaultValueConfig)) : null)
+                .setFilteredExtensions(Parameter.readStringList(getFormat(), parameters, EXTENSIONS_FILTERED_LIST_PARAMETER, defaultValueConfig) != null ? new HashSet<>(Parameter.readStringList(getFormat(), parameters, EXTENSIONS_FILTERED_LIST_PARAMETER, defaultValueConfig)) : null)
                 .setSorted(Parameter.readBoolean(getFormat(), parameters, SORTED_PARAMETER, defaultValueConfig))
                 .setVersion(Parameter.readString(getFormat(), parameters, VERSION_PARAMETER, defaultValueConfig))
                 .setFormat(getTreeDataFormat())
--- a/iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTreeDataImporter.java
+++ b/iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTreeDataImporter.java
@@ -49,6 +49,8 @@ public abstract class AbstractTreeDataImporter implements Importer {
 
     public static final String EXTENSIONS_LIST = "iidm.import.xml.extensions";
 
+    public static final String EXTENSIONS_FILTERED_LIST = "iidm.import.xml.filtered-extensions";
+
     public static final String WITH_AUTOMATION_SYSTEMS = "iidm.import.xml.with-automation-systems";
 
     public static final String MISSING_PERMANENT_LIMIT_PERCENTAGE = "iidm.import.xml.missing-permanent-limit-percentage";
@@ -61,6 +63,10 @@ public abstract class AbstractTreeDataImporter implements Importer {
 
     private static final Parameter EXTENSIONS_LIST_PARAMETER
             = new Parameter(EXTENSIONS_LIST, ParameterType.STRING_LIST, "The list of extension files ", null,
+            EXTENSIONS_SUPPLIER.get().getProviders().stream().map(ExtensionProvider::getExtensionName).collect(Collectors.toList()));
+
+    private static final Parameter EXTENSIONS_FILTERED_LIST_PARAMETER
+            = new Parameter(EXTENSIONS_FILTERED_LIST, ParameterType.STRING_LIST, "The list of extension files that will be ignored and not imported ", null,
             EXTENSIONS_SUPPLIER.get().getProviders().stream().map(ExtensionProvider::getExtensionName).collect(Collectors.toList()));
 
     private static final Parameter WITH_AUTOMATION_SYSTEMS_PARAMETER = new Parameter(WITH_AUTOMATION_SYSTEMS, ParameterType.BOOLEAN,
@@ -89,6 +95,7 @@ public abstract class AbstractTreeDataImporter implements Importer {
     @Override
     public List<Parameter> getParameters() {
         List<Parameter> parameters = List.of(THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND_PARAMETER, EXTENSIONS_LIST_PARAMETER,
+                EXTENSIONS_FILTERED_LIST_PARAMETER,
                 WITH_AUTOMATION_SYSTEMS_PARAMETER, MISSING_PERMANENT_LIMIT_PERCENTAGE_PARAMETER,
                 MINIMAL_VALIDATION_LEVEL_PARAMETER);
         return ConfiguredParameter.load(parameters, getFormat(), defaultValueConfig);
@@ -175,6 +182,7 @@ public abstract class AbstractTreeDataImporter implements Importer {
         return new ImportOptions()
                 .setThrowExceptionIfExtensionNotFound(Parameter.readBoolean(getFormat(), parameters, THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND_PARAMETER, defaultValueConfig))
                 .setExtensions(Parameter.readStringList(getFormat(), parameters, EXTENSIONS_LIST_PARAMETER, defaultValueConfig) != null ? new HashSet<>(Parameter.readStringList(getFormat(), parameters, EXTENSIONS_LIST_PARAMETER, defaultValueConfig)) : null)
+                .setFilteredExtensions(Parameter.readStringList(getFormat(), parameters, EXTENSIONS_FILTERED_LIST_PARAMETER, defaultValueConfig) != null ? new HashSet<>(Parameter.readStringList(getFormat(), parameters, EXTENSIONS_FILTERED_LIST_PARAMETER, defaultValueConfig)) : null)
                 .setWithAutomationSystems(Parameter.readBoolean(getFormat(), parameters, WITH_AUTOMATION_SYSTEMS_PARAMETER, defaultValueConfig))
                 .setMissingPermanentLimitPercentage(Parameter.readDouble(getFormat(), parameters, MISSING_PERMANENT_LIMIT_PERCENTAGE_PARAMETER, defaultValueConfig))
                 .setMinimalValidationLevel(Parameter.readString(getFormat(), parameters, MINIMAL_VALIDATION_LEVEL_PARAMETER, defaultValueConfig));
--- a/iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ExportOptions.java
+++ b/iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ExportOptions.java
@@ -106,6 +106,16 @@ public class ExportOptions extends AbstractOptions<ExportOptions> {
         return this;
     }
 
+    @Override
+    public ExportOptions addFilteredExtension(String extensionToBeFiltered) {
+        if (filteredExtension != null) {
+            filteredExtension.add(extensionToBeFiltered);
+        } else {
+            this.filteredExtension = Sets.newHashSet(extensionToBeFiltered);
+        }
+        return this;
+    }
+
     public boolean isWithBranchSV() {
         return withBranchSV;
     }
@@ -160,6 +170,18 @@ public class ExportOptions extends AbstractOptions<ExportOptions> {
             }
         });
         this.extensions = extensions;
+        return this;
+    }
+
+    @Override
+    public ExportOptions setFilteredExtensions(Set<String> filteredExtension) {
+        // this warning is to prevent people to use setSkipExtensions and setExtensions at the same time
+        Optional.ofNullable(this.filteredExtension).ifPresent(e -> {
+            if (e.isEmpty()) {
+                LOGGER.warn("Extensions have already been set as empty. This call will override it.");
+            }
+        });
+        this.filteredExtension = filteredExtension;
         return this;
     }
 
--- a/iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
+++ b/iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
@@ -211,7 +211,9 @@ public final class NetworkSerDe {
                 continue;
             }
             Collection<? extends Extension<? extends Identifiable<?>>> extensions = identifiable.getExtensions().stream()
-                    .filter(e -> canTheExtensionBeWritten(getExtensionSerializer(context.getOptions(), e, extensionsSupplier), context.getVersion(), context.getOptions()))
+                    .filter(e ->
+                            !isExtensionFiltered(getExtensionSerializer(context.getOptions(), e, extensionsSupplier), context.getOptions()) &&
+                            canTheExtensionBeWritten(getExtensionSerializer(context.getOptions(), e, extensionsSupplier), context.getVersion(), context.getOptions()))
                     .toList();
 
             if (!extensions.isEmpty()) {
@@ -229,6 +231,16 @@ public final class NetworkSerDe {
     private static boolean ignoreEquipmentAtExport(Identifiable<?> identifiable, NetworkSerializerContext context) {
         return !context.isExportedEquipment(identifiable)
                 || identifiable instanceof OverloadManagementSystem && !context.getOptions().isWithAutomationSystems();
+    }
+
+    private static boolean isExtensionFiltered(ExtensionSerDe extensionSerDe, ExportOptions options) {
+        if (extensionSerDe == null) {
+            return false;
+        }
+        if (options.getFilteredExtensions().isPresent()) {
+            return options.getFilteredExtensions().get().contains(extensionSerDe.getExtensionName());
+        }
+        return false;
     }
 
     private static boolean canTheExtensionBeWritten(ExtensionSerDe extensionSerDe, IidmVersion version, ExportOptions options) {
@@ -860,8 +872,8 @@ public final class NetworkSerDe {
             // missing extension serializer, we must not check for an extension in sub elements.
             ExtensionSerDe extensionSerde = extensionsSupplier.get().findProvider(extensionSerializationName);
             String extensionName = extensionSerde != null ? extensionSerde.getExtensionName() : extensionSerializationName;
-            if (!context.isIgnoredEquipment(id)
-                    && (context.getOptions().withExtension(extensionName) || context.getOptions().withExtension(extensionSerializationName))) {
+            if (!context.isIgnoredEquipment(id) && ((context.getOptions().withExtension(extensionName) || context.getOptions().withExtension(extensionSerializationName))) &&
+                    !(context.getOptions().withFilteredExtension(extensionName) || context.getOptions().withFilteredExtension(extensionSerializationName))) {
                 if (extensionSerde != null) {
                     Identifiable identifiable = getIdentifiable(network, id);
                     extensionSerde.read(identifiable, context);
--- a/iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/ExportOptionsTest.java
+++ b/iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/ExportOptionsTest.java
@@ -90,4 +90,18 @@ class ExportOptionsTest {
         assertEquals(StandardCharsets.UTF_8, options.getCharset());
         assertEquals(Boolean.TRUE, options.isWithAutomationSystems());
     }
+
+    @Test
+    void exportOptionsTestFilteredExtensions() {
+        ExportOptions options = new ExportOptions();
+        options.setExtensions(Sets.newHashSet("loadFoo", "loadBar"));
+        options.setFilteredExtensions(Sets.newHashSet("loadBar"));
+        options.addFilteredExtension("test");
+        assertEquals(Boolean.FALSE, options.withNoExtension());
+        assertEquals(Boolean.TRUE, options.withExtension("loadFoo"));
+        assertEquals(Boolean.FALSE, options.withFilteredExtension("loadFoo"));
+        assertEquals(Boolean.TRUE, options.withFilteredExtension("loadBar"));
+        assertEquals(2, (int) options.getExtensions().map(Set::size).orElse(-1));
+        assertEquals(2, (int) options.getFilteredExtensions().map(Set::size).orElse(-1));
+    }
 }
--- a/iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/ExtensionDeserIssueTest.java
+++ b/iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/ExtensionDeserIssueTest.java
@@ -9,6 +9,8 @@ package com.powsybl.iidm.serde;
 
 import com.google.common.jimfs.Configuration;
 import com.google.common.jimfs.Jimfs;
+import com.powsybl.computation.local.LocalComputationManager;
+import com.powsybl.iidm.network.ImportConfig;
 import com.powsybl.iidm.network.Network;
 import com.powsybl.iidm.network.extensions.SlackTerminal;
 import com.powsybl.iidm.network.extensions.SlackTerminalAdder;
@@ -16,8 +18,11 @@ import com.powsybl.iidm.network.test.EurostagTutorialExample1Factory;
 import org.junit.jupiter.api.Test;
 
 import java.io.IOException;
+import java.util.Properties;
 
+import static com.powsybl.iidm.serde.AbstractTreeDataExporter.EXTENSIONS_FILTERED_LIST;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
 
 /**
  * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
@@ -39,6 +44,68 @@ class ExtensionDeserIssueTest {
             var vlload2 = network2.getVoltageLevel("VLLOAD");
             var slackTerminal = vlload2.getExtension(SlackTerminal.class);
             assertNotNull(slackTerminal.getTerminal().getBusView().getBus());
+        }
+    }
+
+    @Test
+    void testIgnoreExtensionExport() throws IOException {
+        Network network = EurostagTutorialExample1Factory.create();
+        var load = network.getLoad("LOAD");
+        var vlload = network.getVoltageLevel("VLLOAD");
+        vlload.newExtension(SlackTerminalAdder.class)
+                .withTerminal(load.getTerminal())
+                .add();
+        try (var fs = Jimfs.newFileSystem(Configuration.unix())) {
+            var file = fs.getPath("/work/test.xiidm");
+            Properties exportParams = new Properties();
+            exportParams.put(EXTENSIONS_FILTERED_LIST, "slackTerminal");
+            network.write("XIIDM", exportParams, file);
+            Network network2 = Network.read(file);
+            var vlload2 = network2.getVoltageLevel("VLLOAD");
+            var slackTerminal = vlload2.getExtension(SlackTerminal.class);
+            assertNull(slackTerminal);
+        }
+    }
+
+    @Test
+    void testIgnoreExtensionImport() throws IOException {
+        Network network = EurostagTutorialExample1Factory.create();
+        var load = network.getLoad("LOAD");
+        var vlload = network.getVoltageLevel("VLLOAD");
+        vlload.newExtension(SlackTerminalAdder.class)
+                .withTerminal(load.getTerminal())
+                .add();
+        try (var fs = Jimfs.newFileSystem(Configuration.unix())) {
+            var file = fs.getPath("/work/test.xiidm");
+            network.write("XIIDM", null, file);
+            Properties importParams = new Properties();
+            importParams.put(AbstractTreeDataImporter.EXTENSIONS_FILTERED_LIST, "slackTerminal");
+            Network network2 = Network.read(file, LocalComputationManager.getDefault(), ImportConfig.CACHE.get(), importParams);
+            var vlload2 = network2.getVoltageLevel("VLLOAD");
+            var slackTerminal = vlload2.getExtension(SlackTerminal.class);
+            assertNull(slackTerminal);
+        }
+    }
+
+    @Test
+    void testIgnoreExtensionImportExtensionAndFiltered() throws IOException {
+        Network network = EurostagTutorialExample1Factory.create();
+        var load = network.getLoad("LOAD");
+        var vlload = network.getVoltageLevel("VLLOAD");
+        vlload.newExtension(SlackTerminalAdder.class)
+                .withTerminal(load.getTerminal())
+                .add();
+        try (var fs = Jimfs.newFileSystem(Configuration.unix())) {
+            var file = fs.getPath("/work/test.xiidm");
+            network.write("XIIDM", null, file);
+            Properties importParams = new Properties();
+            importParams.put(AbstractTreeDataImporter.EXTENSIONS_LIST, "identifiableShortCircuit,slackTerminal");
+            // Filter has priority over positive import list
+            importParams.put(AbstractTreeDataImporter.EXTENSIONS_FILTERED_LIST, "slackTerminal");
+            Network network2 = Network.read(file, LocalComputationManager.getDefault(), ImportConfig.CACHE.get(), importParams);
+            var vlload2 = network2.getVoltageLevel("VLLOAD");
+            var slackTerminal = vlload2.getExtension(SlackTerminal.class);
+            assertNull(slackTerminal);
         }
     }
 }
--- a/iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/ImportOptionsTest.java
+++ b/iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/ImportOptionsTest.java
@@ -46,6 +46,19 @@ class ImportOptionsTest {
     }
 
     @Test
+    void importOptionsTestFilteredExtensions() {
+        ImportOptions options = new ImportOptions(Boolean.FALSE);
+        options.setExtensions(Sets.newHashSet("loadFoo", "loadBar"));
+        options.setFilteredExtensions(Sets.newHashSet("loadBar"));
+        assertEquals(Boolean.FALSE, options.withNoExtension());
+        assertEquals(Boolean.TRUE, options.withExtension("loadFoo"));
+        assertEquals(Boolean.FALSE, options.withFilteredExtension("loadFoo"));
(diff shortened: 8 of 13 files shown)


docs/grid_exchange_formats/iidm/export.md
```diff
@@ -32,6 +32,11 @@ The `iidm.export.xml.throw-exception-if-extension-not-found` property is an opti
 **iidm.export.xml.extensions**  
 The `iidm.export.xml.extensions` property is an optional property that defines the list of extensions that we will be exported by the XIIDM exporter. By default, all extensions will be exported.
 
+**iidm.export.xml.filtered-extensions**  
+The `iidm.export.xml.filtered-extensions` property is an optional property that defines the list of extensions that will not be exported by the XIIDM exporter. 
+When both `iidm.export.xml.extensions` and `iidm.export.xml.filtered-extensions` contains the same extension, `iidm.export.xml.filtered-extensions` takes the priority hence the extension will not be imported.
```
rolnico: This seems to be the opposite of the rule described in the issue. It's not a problem in itself but it would be nice to have an explanation in the PR description (or in comment) on why you chose to do it this way.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractOptions.java
```diff
@@ -19,17 +20,29 @@
 public abstract class AbstractOptions<T> {
 
     protected Set<String> extensions;
+    /**
+     * Extensions to be loaded must be in the extensions set but must not belong to the filteredExtension Set
+     */
+    protected Set<String> filteredExtension = Sets.newHashSet();
```
rolnico: No need to initialize the set here
```suggestion
    protected Set<String> filteredExtensions;
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractOptions.java
```diff
@@ -19,17 +20,29 @@
 public abstract class AbstractOptions<T> {
 
     protected Set<String> extensions;
+    /**
+     * Extensions to be loaded must be in the extensions set but must not belong to the filteredExtension Set
+     */
```
rolnico: You might want to move this comment just over the previous variable since it concerns both `extensions` and `filteredExtension`

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ExportOptions.java
```diff
@@ -163,6 +173,18 @@ public ExportOptions setExtensions(Set<String> extensions) {
...
     }
 
+    @Override
+    public ExportOptions setFilteredExtensions(Set<String> filteredExtension) {
+        // this warning is to prevent people to use setSkipExtensions and setExtensions at the same time
+        Optional.ofNullable(this.filteredExtension).ifPresent(e -> {
+            if (e.isEmpty()) {
+                LOGGER.warn("Extensions have already been set as empty. This call will override it.");
+            }
+        });
```
rolnico: I think this can be removed, as well as in the method `setExtensions`, since the method `setSkipExtensions` does not exist anymore

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTreeDataExporter.java
```diff
@@ -126,6 +127,9 @@ public abstract class AbstractTreeDataExporter implements Exporter {
             Arrays.stream(TopologyLevel.values()).map(Enum::name).collect(Collectors.toList()));
     private static final Parameter THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND_PARAMETER = new Parameter(THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND, ParameterType.BOOLEAN, "Throw exception if extension not found", Boolean.FALSE);
     private static final Parameter EXTENSIONS_LIST_PARAMETER = new Parameter(EXTENSIONS_LIST, ParameterType.STRING_LIST,
+            "The list of extensions that will be ignored during export", null,
```
rolnico: This description is for `EXTENSIONS_FILTERED_LIST_PARAMETER` and not for `EXTENSIONS_LIST_PARAMETER`

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTreeDataImporter.java
```diff
@@ -63,6 +65,10 @@ public abstract class AbstractTreeDataImporter implements Importer {
             = new Parameter(EXTENSIONS_LIST, ParameterType.STRING_LIST, "The list of extension files ", null,
             EXTENSIONS_SUPPLIER.get().getProviders().stream().map(ExtensionProvider::getExtensionName).collect(Collectors.toList()));
 
+    private static final Parameter EXTENSIONS_FILTERED_LIST_PARAMETER
+            = new Parameter(EXTENSIONS_FILTERED_LIST, ParameterType.STRING_LIST, "The list of extension files that will be ignored and not imported ", null,
```
rolnico:
```suggestion
            = new Parameter(EXTENSIONS_FILTERED_LIST, ParameterType.STRING_LIST, "The list of extensions that will be ignored and not imported", null,
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractOptions.java
```diff
@@ -54,6 +67,10 @@ public boolean withExtension(String extensionName) {
         return withAllExtensions() || extensions.contains(extensionName);
     }
 
+    public boolean withFilteredExtension(String extensionName) {
+        return filteredExtension != null && filteredExtension.contains(extensionName);
+    }
```
rolnico: You can avoid adding this method by changing the logic of `withExtension`:

```java
    public boolean withExtension(String extensionName) {
        return (filteredExtension == null || !filteredExtension.contains(extensionName))
            && (withAllExtensions() || extensions.contains(extensionName));
    }
```
It would simplify your checks in `NetworkSerde`

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -860,8 +871,8 @@ private static void readExtensions(Network network, String id, NetworkDeserializ
             // missing extension serializer, we must not check for an extension in sub elements.
             ExtensionSerDe extensionSerde = extensionsSupplier.get().findProvider(extensionSerializationName);
             String extensionName = extensionSerde != null ? extensionSerde.getExtensionName() : extensionSerializationName;
-            if (!context.isIgnoredEquipment(id)
-                    && (context.getOptions().withExtension(extensionName) || context.getOptions().withExtension(extensionSerializationName))) {
+            if (!context.isIgnoredEquipment(id) && ((context.getOptions().withExtension(extensionName) || context.getOptions().withExtension(extensionSerializationName))) &&
+                    !context.getOptions().withFilteredExtension(extensionName) && !context.getOptions().withFilteredExtension(extensionSerializationName)) {
```
rolnico: With the new logic, the code might stay the same as it was

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -211,7 +211,9 @@ private static void writeExtensions(Network n, NetworkSerializerContext context,
                 continue;
             }
             Collection<? extends Extension<? extends Identifiable<?>>> extensions = identifiable.getExtensions().stream()
-                    .filter(e -> canTheExtensionBeWritten(getExtensionSerializer(context.getOptions(), e, extensionsSupplier), context.getVersion(), context.getOptions()))
+                    .filter(e ->
+                            !isExtensionFiltered(getExtensionSerializer(context.getOptions(), e, extensionsSupplier), context.getOptions()) &&
+                            canTheExtensionBeWritten(getExtensionSerializer(context.getOptions(), e, extensionsSupplier), context.getVersion(), context.getOptions()))
```
rolnico, line 217: same here

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ExportOptions.java
```diff
@@ -106,6 +106,16 @@ public ExportOptions addExtension(String extension) {
         return this;
     }
 
+    @Override
+    public ExportOptions addFilteredExtension(String extensionToBeFiltered) {
```
rolnico: If this method is the same in ExportOptions and in ImportOptions, it could be implemented directly in AbstractOptions instead

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTreeDataImporter.java
snoopfab: do you want me to change also EXTENSIONS_LIST comment ? "The list of extension files " to "The list of extensions" ?

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
snoopfab: canTheExtensionBeWritten doesnt check withExtensions directly (it just checks versions) But maybe versions can only be added if withExtensions test pass ?

docs/grid_exchange_formats/iidm/export.md
snoopfab: I was  thinking of giving **filter (ignore) list** higher priority over the inclusive one.
because IMO this is the most intuitive choice — it ensures you can *always* force-disable something, even if it’s normally included.

Example:

Case 1:

* Inclusive = `[A, B, C, D]`
* Filter = `[C]`
  =>️ Result = `[A, B, D]`

Case 2:

If inclusive is `null` (all):

* Inclusive = `null`
* Filter = `[B]`
  => Result = `[all except B]`


This the filter list acts like a *safety override*; It Works the same whether you’re whitelisting or using "all"and lets users say: "Load these, except this one".

On the over hand, the suggested initial rule provided in the PR would give for example :

All available plugins: `[A, B, C, D]`

Case 1:

* Inclusive = `null`
* Filter = `[B]`
  => Result = `[A, B, C, D]` (*filter ignored*)

Case 2:

* Inclusive = `[A, B, C]`
* Filter = `[B]`
  =>️ Result = `[A, B, C]` (*filter ignored*)


This is simple as the inclusive list is always the "final say", it prevents unexpected exclusions 
and is useful if filters are set globally but inclusives are very intentional.

But you lose the ability to "force-disable" a plugin when using inclusives.
and IMO this might surprise users who expect the filter list to always work as an override.

At the end it boils down to philosophy:

* **Filter wins = safe mode (force-disable anything)**
* **Inclusive wins = strict control (only what I say gets loaded, no hidden exclusions)**

alicecaron: As updated in the issue for this feature, we do not want to have both inclusive and exclusive filters. So if both parameters are present, then we prefer throwing an exception. This way if a user is no explicit in which extensions to filter or with which one only he wans to load it will throw an error and force the user to reconfigure explicitly wha he wants.

alicecaron: As updated in the issue for this feature, we do not want to have both inclusive and exclusive filters. So if both parameters are present, then we prefer throwing an exception. This way if a user is no explicit in which extensions to filter or with which one only he wans to load it will throw an error and force the user to reconfigure explicitly wha he wants.

snoopfab: @alicecaron : 
I'm okay when both options are **set** directly and both the `extension` and `filteredExtensions` sets have an intersection, to throw an Exception. However, consider the following test case:

```
@Test
void importOptionsTestFilteredExtensions3() {
    try {
        ImportOptions options = new ImportOptions(Boolean.FALSE);
        options.addFilteredExtension("loadBar");
    } catch (Exception e) {
        assertInstanceOf(PowsyblException.class, e);
    }
}
```

In this case, no extensions are defined **explicitly**; hence, Powsybl should load all extensions. But one extension is filtered, so by your definition, the code should throw an exception.

This would mean that one cannot simply add an ignored extension without explicitly defining the list of extensions to load, which defeats the initial purpose of having a filter list of extensions to ignore.

In my opinion, PowSybl should allow this specific case without throwing any exception (only when the load list is not explicitly set by the user). Could you specify the intended behavior?

rolnico: I would say, the different cases are:
- whitelist and blacklist both explicitly defined in the configuration -> exception
- only a whitelist explicitly defined in the configuration:
	- no manual addition -> whitelist
	- manual addition via addExtension -> whitelist + addition
	- manual addition via addFilteredExtension -> exception
- only a blacklist explicitly defined in the configuration:
	- no manual addition -> all except the blacklist
	- manual addition via addExtension -> exception
	- manual addition via addFilteredExtension -> all except the { blacklist + addition }
- No whitelist nor blacklist explicitly defined in the configuration:
	- no manual addition -> all extensions
	- manual addition via addExtension -> only the extension added
	- manual addition via addFilteredExtension -> all extensions except the extension added

Do you think it's ok and it covers every case?

rolnico: More thoughts on this... 
We have to separate two use-cases: when we define the lists from the config and when we define the lists manually (there is no place where we do both at the same time, I think).

- Lists based on the configuration: the list can be defined in the configuration file (default case) or in the Parameters. In that case, the method `setExtensions` is used. Both the whitelist and the blacklist _cannot_ be defined at the same time. If both are defined (even one in the configuration file and one in the Parameters), an exception should be thrown
- Lists created manually: users can define ImportOptions manually. In that case, we have to follow some logic to manage the lists.

What we propose would be to follow a main rule: you cannot have at the same time a blacklist and a whitelist defined. Therefore:
- rename `setExtensions` (resp. `setFilteredExtensions`) to `setIncludedExtensions` (resp. `setExcludedExtensions`)
- rename `addExtension` (resp. `addFilteredExtension`) to `addIncludedExtension` (resp. `addExcludedExtension`)
- when calling `setIncludedExtensions` (resp. `setExcludedExtensions`), set the blacklist (resp. whitelist) as null in addition to setting the provided list (ie. reset the other list)
- when calling `addIncludedExtension` (resp. `addExcludedExtension`):
    - if a whitelist (resp. blacklist) is defined, add the extension to the list
    - if a blacklist (resp. whitelist) is defined, remove the extension from the list if it is already present. If it's not present, then do nothing
    - if no list is already defined, then initialize the whitelist (resp. blacklist) with the provided extension

The check and exception on both a whitelist and a blacklist being defined should be managed in the `AbstractTreeDataImporter`/`AbstractTreeDataExporter` classes.

snoopfab:
renamed setExtensions (resp. setFilteredExtensions) to setIncludedExtensions (resp. setExcludedExtensions)
renamed addExtension (resp. addFilteredExtension) to addIncludedExtension (resp. addExcludedExtension)
when calling setIncludedExtensions (resp. setExcludedExtensions), set the blacklist (resp. whitelist) as null in addition to setting the provided list (ie. reset the other list)
when calling addIncludedExtension (resp. addExcludedExtension):
if a whitelist (resp. blacklist) is defined, add the extension to the list
if a blacklist (resp. whitelist) is defined, remove the extension from the list if it is already present. If it's not present, then do nothing
if no list is already defined, then initialize the whitelist (resp. blacklist) with the provided extension

I Added the following warning  logs when specific conditions are met (for the user to be aware of the previous settings overriding):

for setters methods (setFilteredExtensions,setIncludedExtensions ) :
- Previously excluded extensions {} will be ignored ! (setIncludedExtensions )
- All Extensions will be disabled (when passing en empty set to setIncludedExtensions to allow disabling all extensions
- Previously included extensions {} will be ignored !

for adder methods (addIncludedExtension,addExcludedExtension) :
- Previously excluded extensions {} will be included"
- Extension {} is not already explicitly included {} ! It will still be added to the exclusion list as it may be included after (this one is when you add an exclusion when a list of included extension was defined but doesn't contain the extension to be filtered , in this case the programmer still has the possibility to add the extension in the included list )

- Adapted and added tests to reflect the new expected behavior
- Adapted the documentation

for ImportOptions / ExportOptions Parameters:
now throw an ConfigurationException when both include and exclude extensions list are defined at the same time

docs/grid_exchange_formats/iidm/import.md
```diff
@@ -9,8 +9,14 @@ These properties can be defined in the configuration file in the [import-export-
...
 The `iidm.import.xml.throw-exception-if-extension-not-found` property is an optional property that defines if the XIIDM importer throws an exception while trying to import an unknown or undeserializable extension or if it just ignores it. Its default value is `false`.
 
-**iidm.import.xml.extensions**  
-The `iidm.import.xml.extensions` property is an optional property that defines the list of extensions that will be imported by the XIIDM importer. By default, all extensions will be imported.
+**iidm.import.xml.included.extensions**  
+The `iidm.import.xml.included.extensions` property is an optional property that defines the list of extensions that will be imported by the XIIDM importer. By default, all extensions will be imported.
+When set to an empty string, all extensions will be ignored during the import.
+
+**iidm.import.xml.excluded-extensions**
+The `iidm.import.xml.excluded-extensions` property is an optional property that defines the list of extensions that will not be imported by the XIIDM importer.
```
rolnico: We should be consistent in the naming
```suggestion
The `iidm.import.xml.excluded.extensions` property is an optional property that defines the list of extensions that will not be imported by the XIIDM importer.
```

docs/grid_exchange_formats/iidm/import.md
```diff
@@ -9,8 +9,14 @@ These properties can be defined in the configuration file in the [import-export-
...
+When both `iidm.import.xml.included.extensions` and `iidm.import.xml.excluded-extensions` are defined, a configuration exception is thrown.
```
rolnico:
```suggestion
When both `iidm.import.xml.included.extensions` and `iidm.import.xml.excluded.extensions` are defined, a configuration exception is thrown. 
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractOptions.java
```diff
@@ -18,40 +22,96 @@
...
+    protected Set<String> includedExtensions;
+
+    protected Set<String> excludedExtensions;
 
     protected TreeDataFormat format = TreeDataFormat.XML;
 
-    public abstract T setExtensions(Set<String> extensions);
+    public T setIncludedExtensions(Set<String> extensions) {
+        if (excludedExtensions != null) {
+            LOGGER.warn("Previously excluded extensions {} will be ignored !", excludedExtensions);
```
rolnico:
```suggestion
            LOGGER.warn("Previously excluded extensions list will now be ignored: {}", excludedExtensions);
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractOptions.java
```diff
@@ -18,40 +22,96 @@
...
+            LOGGER.warn("All Extensions will be disabled");
+        }
+        this.excludedExtensions = null;
+        this.includedExtensions = extensions;
+        return castThis();
+    }
+
+    public T setExcludedExtensions(Set<String> extensions) {
+        if (includedExtensions != null) {
+            LOGGER.warn("Previously included extensions {} will be ignored !", includedExtensions);
```
rolnico:
```suggestion
            LOGGER.warn("Previously included extensions list will now be ignored: {}", includedExtensions);
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractOptions.java
```diff
@@ -18,40 +22,96 @@
...
 
     protected TreeDataFormat format = TreeDataFormat.XML;
 
-    public abstract T setExtensions(Set<String> extensions);
+    public T setIncludedExtensions(Set<String> extensions) {
+        if (excludedExtensions != null) {
+            LOGGER.warn("Previously excluded extensions {} will be ignored !", excludedExtensions);
+        }
+        if (extensions != null && extensions.isEmpty()) {
+            LOGGER.warn("All Extensions will be disabled");
```
rolnico:
```suggestion
            LOGGER.warn("All extensions will be excluded");
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractOptions.java
```diff
@@ -18,40 +22,96 @@
...
+            LOGGER.warn("Previously excluded extensions {} will be included", extension);
+        }
+        if (excludedExtensions != null) {
+            excludedExtensions.remove(extension);
+        }
+        if (includedExtensions != null) {
+            includedExtensions.add(extension);
+        } else {
+            this.includedExtensions = Sets.newHashSet(extension);
+        }
```
rolnico: If the excluded extensions list is not null, we do not want to add the extension to the other list since we do not want to manage both lists at the same time

```suggestion
        if (excludedExtensions != null) {
            if (excludedExtensions.remove(extension)) {
                LOGGER.warn("Previously excluded extensions {} will now be included", extension);
            }
        } else {
            if (includedExtensions == null) {
                includedExtensions = new HashSet<>();
            }
            includedExtensions.add(extension);
        }
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractOptions.java
```diff
@@ -18,40 +22,96 @@
...
+    public T setExcludedExtensions(Set<String> extensions) {
+        if (includedExtensions != null) {
+            LOGGER.warn("Previously included extensions {} will be ignored !", includedExtensions);
+        }
+        this.includedExtensions = null;
+        this.excludedExtensions = extensions;
+        return castThis();
+    }
+
+    public T addIncludedExtension(String extension) {
```
rolnico, line 69: Add javadoc

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractOptions.java
```diff
@@ -18,40 +22,96 @@
...
+        }
+        if (includedExtensions != null) {
+            includedExtensions.add(extension);
+        } else {
+            this.includedExtensions = Sets.newHashSet(extension);
+        }
+        return castThis();
+    }
+
+    public T addExcludedExtension(String extension) throws PowsyblException {
```
rolnico, line 88: add javadoc

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractOptions.java
```diff
@@ -18,40 +22,96 @@
...
+            LOGGER.warn("Extension {} is not already explicitly included {} ! It will still be added to the exclusion list as it may be included after", extension, includedExtensions);
+        }
+        if (includedExtensions != null) {
+            includedExtensions.remove(extension);
+        }
+        if (excludedExtensions != null) {
+            excludedExtensions.add(extension);
+        } else {
+            this.excludedExtensions = Sets.newHashSet(extension);
+        }
```
rolnico:
```suggestion
        if (includedExtensions != null) {
            if (includedExtensions.remove(extension)) {
                LOGGER.warn("Previously included extensions {} will now be excluded", extension);
            }
        } else {
            if (excludedExtensions == null) {
                excludedExtensions = new HashSet<>();
            }
            excludedExtensions.add(extension);
        }
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTreeDataExporter.java
```diff
@@ -169,20 +177,26 @@ public List<Parameter> getParameters() {
...
-            Parameter parameter = new Parameter("iidm.export.xml." + extensionName + ".version",
-                    ParameterType.STRING, "Version of " + extensionName, null);
-            String extensionVersion = Parameter.readString(getFormat(), parameters, parameter, defaultValueConfig);
-            if (extensionVersion != null) {
-                if (options.getExtensions().map(extensions -> extensions.contains(extensionName)).orElse(true)) {
+        if (parameters != null && parameters.containsKey(EXTENSIONS_INCLUDED_LIST)) {
+            String extensionToIncludeInExport = parameters.getProperty(EXTENSIONS_INCLUDED_LIST);
+            if (extensionToIncludeInExport != null && extensionToIncludeInExport.isEmpty()) {
+                // ignore all extensions case
+                LOGGER.warn("All extensions are ignored.");
```
rolnico:
```suggestion
                LOGGER.info("All extensions are excluded.");
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTreeDataExporter.java
```diff
@@ -169,20 +177,26 @@ public List<Parameter> getParameters() {
...
+            }
+        } else {
+            EXTENSIONS_SUPPLIER.get().getProviders().forEach(extensionSerializer -> {
+                String extensionName = extensionSerializer.getExtensionName();
+                Parameter parameter = new Parameter("iidm.export.xml." + extensionName + ".version",
+                        ParameterType.STRING, "Version of " + extensionName, null);
+                String extensionVersion = Parameter.readString(getFormat(), parameters, parameter, defaultValueConfig);
+                if (options.getIncludedExtensions().map(extensions -> extensions.contains(extensionName)).orElse(true)) {
```
rolnico: You need to have a check on the excluded extensions too

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTreeDataExporter.java
```diff
@@ -169,20 +177,26 @@ public List<Parameter> getParameters() {
...
 
     private void addExtensionsVersions(Properties parameters, ExportOptions options) {
-        EXTENSIONS_SUPPLIER.get().getProviders().forEach(extensionSerializer -> {
-            String extensionName = extensionSerializer.getExtensionName();
-            Parameter parameter = new Parameter("iidm.export.xml." + extensionName + ".version",
-                    ParameterType.STRING, "Version of " + extensionName, null);
-            String extensionVersion = Parameter.readString(getFormat(), parameters, parameter, defaultValueConfig);
-            if (extensionVersion != null) {
-                if (options.getExtensions().map(extensions -> extensions.contains(extensionName)).orElse(true)) {
+        if (parameters != null && parameters.containsKey(EXTENSIONS_INCLUDED_LIST)) {
```
rolnico: Instead of rechecking if the parameter is defined in the properties, you can provide as arguments of this method the two List<String> computed in createExportOptions

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTreeDataExporter.java
```diff
@@ -194,11 +208,21 @@ private ExportOptions createExportOptions(Properties parameters) {
...
+        List<String> excludedExtensions = Parameter.readStringList(getFormat(), parameters, EXTENSIONS_EXCLUDED_LIST_PARAMETER, defaultValueConfig);
+        if (includedExtensions != null && excludedExtensions != null) {
+            throw new ConfigurationException("You can't define both included and excluded extensions in parameters.");
+        }
+        if (includedExtensions != null) {
+            options.setIncludedExtensions(new HashSet<>(includedExtensions));
+        }
+        if (excludedExtensions != null) {
+            options.setExcludedExtensions(new HashSet<>(excludedExtensions));
+        }
```
rolnico: If possible, try to avoid code duplication with the importer

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ExportOptions.java
```diff
@@ -215,7 +191,7 @@ public ExportOptions setCharset(Charset charset) {
      * serialization name is used for the <code>extensionName</code> parameter, the method will have no effect.</p>
      */
     public ExportOptions addExtensionVersion(String extensionName, String extensionVersion) {
-        if (extensions != null && !extensions.contains(extensionName)) {
+        if (includedExtensions != null && !includedExtensions.contains(extensionName)) {
```
rolnico, line 196: what if `excludedExtensions` is defined instead of `includedExtensions`?

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -231,6 +234,15 @@ private static boolean ignoreEquipmentAtExport(Identifiable<?> identifiable, Net
                 || identifiable instanceof OverloadManagementSystem && !context.getOptions().isWithAutomationSystems();
     }
 
+    private static boolean isExtensionExcluded(ExtensionSerDe extensionSerDe, ExportOptions options) {
+        if (extensionSerDe == null) {
+            return false;
+        }
+        return options.getIncludedExtensions().filter(Set::isEmpty).isPresent() || options.getExcludedExtensions()
```
rolnico: If the list of included extensions is defined, are you really checking that the extension is not contained in the list?

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -860,8 +872,7 @@ private static void readExtensions(Network network, String id, NetworkDeserializ
             // missing extension serializer, we must not check for an extension in sub elements.
             ExtensionSerDe extensionSerde = extensionsSupplier.get().findProvider(extensionSerializationName);
             String extensionName = extensionSerde != null ? extensionSerde.getExtensionName() : extensionSerializationName;
-            if (!context.isIgnoredEquipment(id)
-                    && (context.getOptions().withExtension(extensionName) || context.getOptions().withExtension(extensionSerializationName))) {
+            if (!context.isIgnoredEquipment(id) && (context.getOptions().withExtension(extensionName) || context.getOptions().withExtension(extensionSerializationName))) {
```
rolnico: revert, it can stay on 2 lines

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/ExtensionDeserIssueTest.java
```diff
@@ -9,15 +9,22 @@
...
 import org.junit.jupiter.api.Test;
 
 import java.io.IOException;
+import java.nio.file.Path;
+import java.util.Properties;
 
-import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static com.powsybl.iidm.serde.AbstractTreeDataExporter.EXTENSIONS_EXCLUDED_LIST;
+import static com.powsybl.iidm.serde.AbstractTreeDataExporter.EXTENSIONS_INCLUDED_LIST;
+import static org.junit.jupiter.api.Assertions.*;
```
rolnico: avoid using *

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/ExtensionDeserIssueTest.java
```diff
@@ -41,4 +48,93 @@ void test() throws IOException {
...
+
+    @Test
+    void testIgnoreExtensionExport() throws IOException {
+        Network network = EurostagTutorialExample1Factory.create();
+        var load = network.getLoad("LOAD");
+        var vlload = network.getVoltageLevel("VLLOAD");
+        vlload.newExtension(SlackTerminalAdder.class)
+                .withTerminal(load.getTerminal())
+                .add();
+        try (var fs = Jimfs.newFileSystem(Configuration.unix())) {
```
rolnico: avoid these `try` by implementing methods annotated with `@BeforeEach` and `@AfterEach`.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/ExtensionDeserIssueTest.java
```diff
@@ -41,4 +48,93 @@ void test() throws IOException {
...
+            network.write("XIIDM", null, file);
+            Properties importParams = new Properties();
+            importParams.put(AbstractTreeDataImporter.EXTENSIONS_INCLUDED_LIST, "identifiableShortCircuit");
+            importParams.put(AbstractTreeDataImporter.EXTENSIONS_EXCLUDED_LIST, "slackTerminal");
+            try {
+                readNetwork(file, importParams);
+                fail();
+            } catch (ConfigurationException e) {
+                assertInstanceOf(ConfigurationException.class, e);
+            }
```
rolnico: rewrite these tests by using `assertThrows` and checking the exception message afterwards

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/ExtensionDeserIssueTest.java
```diff
@@ -41,4 +48,93 @@ void test() throws IOException {
...
+            var file = fs.getPath("/work/test.xiidm");
+            Properties exportParams = new Properties();
+            exportParams.put(EXTENSIONS_EXCLUDED_LIST, "slackTerminal");
+            exportParams.put(EXTENSIONS_INCLUDED_LIST, "loadBar");
+            try {
+                network.write("XIIDM", exportParams, file);
+                fail();
+            } catch (ConfigurationException e) {
+                assertInstanceOf(ConfigurationException.class, e);
+            }
```
rolnico: same here

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/ExportOptionsTest.java
```diff
@@ -90,4 +90,61 @@ private void testDefaultExportOptions(ExportOptions options) {
         assertEquals(StandardCharsets.UTF_8, options.getCharset());
         assertEquals(Boolean.TRUE, options.isWithAutomationSystems());
     }
+
+    @Test
+    void exportOptionsTestFilteredExtensions() {
```
rolnico: This test will have to be fixed since the logic will have been fixed

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/ImportOptionsTest.java
```diff
@@ -23,35 +25,137 @@ class ImportOptionsTest {
...
 
         assertEquals(Boolean.FALSE, options.isThrowExceptionIfExtensionNotFound());
         assertEquals(Boolean.FALSE, options.withNoExtension());
-        assertEquals(2, (int) options.getExtensions().map(Set::size).orElse(-1));
+        assertEquals(2, (int) options.getIncludedExtensions().map(Set::size).orElse(-1));
         assertEquals(95., options.getMissingPermanentLimitPercentage());
     }
 
+    @Test
+    void importOptionsTestExcludedExtensions() {
```
rolnico: probably same here

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractOptions.java
snoopfab: With the proposed changes in addExcludedExtension if an includedExtension was defined before, the extensions could never be added to excludedExtensions and without any warning after so i propose the following change to cover this issue :
```
if (includedExtensions != null && includedExtensions.remove(extension)) {
    LOGGER.warn("Previously included extensions {} will now be excluded", extension);
} 
if (excludedExtensions == null) {
    excludedExtensions = new HashSet<>();
}
excludedExtensions.add(extension);

```
the same problem for addIncludedExtension if an excludedExtensions was defineds before. with a similar fix

rolnico: If `includedExtensions` was previously defined, it already means that only the extensions from that list will be included. In other word: all the extensions not in this list are excluded.
Therefore, if you add a new excluded extension via `addExcludedExtension` while `includedExtensions` is defined, you just have to remove the extension from `includedExtensions` and it will not be included (and therefore it will be excluded).

snoopfab: ImportOptions options2 = new ImportOptions(Boolean.FALSE);
options2.addIncludedExtension("loadFoo");
options2.addIncludedExtension("loadBar");
options2.addExcludedExtension("loadBar");


With the behavior proposed loadBar would not be excluded (ie loadFoo and loadBar would be imported).
for me the logical thing is to exclude it and add a warning that it was previously included

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/ImportOptionsTest.java
```diff
@@ -51,9 +57,23 @@ void importOptionsDefaultValues() {
...
+        ImportOptions options = getOptions();
+        options.addIncludedExtension("loadFoo").addIncludedExtension("loadBar"); // only "loadFoo" and "loadBar" are included
+        options.addExcludedExtension("loadBar").addExcludedExtension("loadTest"); // "loadBar" is excluded so only "loadFoo" is included
+
+        Set<String> includedExtensions = options.getIncludedExtensions().orElse(null);
+        assertNotNull(includedExtensions);
+        assertEquals(1, includedExtensions.size());
+        assertTrue(includedExtensions.contains("loadFoo"));
+        assertTrue(options.getExcludedExtensions().isEmpty()); // The second list is null (i.e., the optional is empty)
+    }
```
rolnico: This test can probably move to AbstractOptionsTest since it's duplicated in the two classes?

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/ExtensionDeserIssueTest.java
```diff
@@ -9,36 +9,127 @@
...
+    private Network network;
+    private FileSystem fs;
+    private Path file;
+
+    /**
+     * Sets up the common test environment:
+     * 1. Creates the base Network.
+     * 2. Adds the SlackTerminal extension.
+     * 3. Creates the Jimfs filesystem and the test file path.
+     * 4. Writes the network to the temporary XIIDM file.
```
rolnico: step 4 seems to have been moved to the different tests

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractOptionExtensionChecker.java
```diff
@@ -0,0 +1,59 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Properties;
+
+/**
+ * This delegate class avoid code duplication in AbstractTreeDataImporter and AbstractTreeDataExporter classes
+ */
+public abstract class AbstractOptionExtensionChecker {
```
rolnico: Since it's an utilitary class, I would suggest to make it `final` and not `abstract`

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractOptionExtensionChecker.java
```diff
@@ -0,0 +1,59 @@
...
+            throw new ConfigurationException("You can't define both included and excluded extensions in parameters.");
+        }
+        if (excludedExtensions != null) {
+            options.setExcludedExtensions(new HashSet<>(excludedExtensions));
+        }
+        if (includedExtensions != null) {
+            options.setIncludedExtensions(new HashSet<>(includedExtensions));
+            if (shouldWarnOnInclusionEmptiness && includedExtensions.isEmpty()) {
+                LOGGER.info("All extensions are excluded.");
+                return false;
```
rolnico: if `shouldWarnOnInclusionEmptiness` is false, you still returns true, even if `includedExtensions.isEmpty()` is true.
Should you not do something like this:
```suggestion
            if (includedExtensions.isEmpty()) {
                if (shouldWarnOnInclusionEmptiness) {
                    LOGGER.warn("No extensions are included.");
                }
                return false;
```
Maybe there is a way to have a logic a bit more clean with less ifs

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTreeDataExporter.java
```diff
@@ -194,12 +198,14 @@ private ExportOptions createExportOptions(Properties parameters) {
...
                 .setThrowExceptionIfExtensionNotFound(Parameter.readBoolean(getFormat(), parameters, THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND_PARAMETER, defaultValueConfig))
-                .setExtensions(Parameter.readStringList(getFormat(), parameters, EXTENSIONS_LIST_PARAMETER, defaultValueConfig) != null ? new HashSet<>(Parameter.readStringList(getFormat(), parameters, EXTENSIONS_LIST_PARAMETER, defaultValueConfig)) : null)
                 .setSorted(Parameter.readBoolean(getFormat(), parameters, SORTED_PARAMETER, defaultValueConfig))
                 .setVersion(Parameter.readString(getFormat(), parameters, VERSION_PARAMETER, defaultValueConfig))
                 .setFormat(getTreeDataFormat())
                 .setWithAutomationSystems(Parameter.readBoolean(getFormat(), parameters, WITH_AUTOMATION_SYSTEMS_PARAMETER, defaultValueConfig));
-        addExtensionsVersions(parameters, options);
+        boolean someExtensionsShouldBeIncluded = AbstractOptionExtensionChecker.getAndCheckExtensionsToInclude(parameters, options, getFormat(), defaultValueConfig, EXTENSIONS_INCLUDED_LIST_PARAMETER, EXTENSIONS_EXCLUDED_LIST_PARAMETER, true);
```
rolnico: Import directly the static method

```suggestion
        boolean someExtensionsShouldBeIncluded = getAndCheckExtensionsToInclude(parameters, options, getFormat(), defaultValueConfig, EXTENSIONS_INCLUDED_LIST_PARAMETER, EXTENSIONS_EXCLUDED_LIST_PARAMETER, true);
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTreeDataImporter.java
```diff
@@ -172,12 +179,15 @@ public Network importData(ReadOnlyDataSource dataSource, NetworkFactory networkF
...
+                .setExcludedExtensions(Parameter.readStringList(getFormat(), parameters, EXTENSIONS_EXCLUDED_LIST_PARAMETER, defaultValueConfig) != null ? new HashSet<>(Parameter.readStringList(getFormat(), parameters, EXTENSIONS_EXCLUDED_LIST_PARAMETER, defaultValueConfig)) : null)
                 .setWithAutomationSystems(Parameter.readBoolean(getFormat(), parameters, WITH_AUTOMATION_SYSTEMS_PARAMETER, defaultValueConfig))
                 .setMissingPermanentLimitPercentage(Parameter.readDouble(getFormat(), parameters, MISSING_PERMANENT_LIMIT_PERCENTAGE_PARAMETER, defaultValueConfig))
                 .setMinimalValidationLevel(Parameter.readString(getFormat(), parameters, MINIMAL_VALIDATION_LEVEL_PARAMETER, defaultValueConfig));
+        AbstractOptionExtensionChecker.getAndCheckExtensionsToInclude(parameters, importOptions, getFormat(), defaultValueConfig, EXTENSIONS_INCLUDED_LIST_PARAMETER, EXTENSIONS_EXCLUDED_LIST_PARAMETER, false);
```
rolnico: same here

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTreeDataExporter.java
```diff
@@ -174,13 +180,11 @@ private void addExtensionsVersions(Properties parameters, ExportOptions options)
             Parameter parameter = new Parameter("iidm.export.xml." + extensionName + ".version",
                     ParameterType.STRING, "Version of " + extensionName, null);
             String extensionVersion = Parameter.readString(getFormat(), parameters, parameter, defaultValueConfig);
-            if (extensionVersion != null) {
```
rolnico, line 186: Why are you removing the check on `extensionVersion`?

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -231,6 +234,19 @@ private static boolean ignoreEquipmentAtExport(Identifiable<?> identifiable, Net
                 || identifiable instanceof OverloadManagementSystem && !context.getOptions().isWithAutomationSystems();
     }
 
+    private static boolean isExtensionExcluded(ExtensionSerDe extensionSerDe, ExportOptions options) {
+        if (extensionSerDe == null) {
+            return false;
```
rolnico, line 239: should not the extension be excluded if the `extensionSerDe` is null?
The line seems to be covered but the tests works both when I return true and false so there might be an issue. You may want to have a dedicated test for this

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -231,6 +234,19 @@ private static boolean ignoreEquipmentAtExport(Identifiable<?> identifiable, Net
...
+        }
+        return options.getIncludedExtensions().filter(Set::isEmpty).isPresent() ||
+                options.getExcludedExtensions()
+                        .map(extensions -> extensions.contains(extensionSerDe.getExtensionName()))
+                        .orElse(false) &&
+                        options.getIncludedExtensions()
+                                .map(extensions -> !extensions.contains(extensionSerDe.getExtensionName()))
+                                .orElse(false);
+    }
```
rolnico: You could use the method `withExtension` here, and inverse the logic:
```suggestion
    private static boolean isExtensionIncluded(ExtensionSerDe extensionSerDe, ExportOptions options) {
        if (extensionSerDe == null) {
            return false;
        }
        return options.withExtension(extensionSerDe.getExtensionName());
    }
```

In that case you have to remove a "!" where this method is used

rolnico (changes requested):

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTreeDataExporter.java
snoopfab: mistakely removed

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractOptionExtensionChecker.java
snoopfab: the linter impose it for a reason that i don't clearly understand : for the maven build to pass , the checks must pass and i had to comply with it

rolnico: Just rename the class to something that does not start with `Abstract`

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTreeDataImporter.java
```diff
@@ -47,7 +48,9 @@ public abstract class AbstractTreeDataImporter implements Importer {
 
     public static final String THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND = "iidm.import.xml.throw-exception-if-extension-not-found";
 
-    public static final String EXTENSIONS_LIST = "iidm.import.xml.extensions";
+    public static final String EXTENSIONS_INCLUDED_LIST = "iidm.import.xml.extensions";
+
+    public static final String EXTENSIONS_EXCLUDED_LIST = "iidm.import.xml.filtered-extensions";
```
rolnico: Should be modified to be consistent with `AbstractTreeDataExporter`

PR #3569, opened 2025-09-15
Enable AutoService Discovery for CGMES Naming Strategy

colineplqt:

No

Feature

Naming strategy for CGMES import and export must be defined in powsybl-core.

This PR allows automatic discovery of Naming Strategies, allowing thus to define one outside powsybl-core.

- [ ] Yes
- [X] No

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/naming/AbstractCgmesAliasNamingStrategy.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/naming/AbstractCgmesAliasNamingStrategy.java
@@ -32,9 +32,9 @@ import static com.powsybl.cgmes.conversion.naming.CgmesObjectReference.ref;
  */
 public abstract class AbstractCgmesAliasNamingStrategy implements NamingStrategy {
 
-    public final BiMap<String, String> idByUuid = HashBiMap.create();
-    public final Map<String, String> uuidSeed = new HashMap<>();
-    public final NameBasedGenerator nameBasedGenerator;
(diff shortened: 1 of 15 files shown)


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/naming/AbstractCgmesAliasNamingStrategy.java
```diff
@@ -32,9 +32,9 @@
  */
 public abstract class AbstractCgmesAliasNamingStrategy implements NamingStrategy {
 
-    private final BiMap<String, String> idByUuid = HashBiMap.create();
-    private final Map<String, String> uuidSeed = new HashMap<>();
-    private final NameBasedGenerator nameBasedGenerator;
+    public final BiMap<String, String> idByUuid = HashBiMap.create();
```
zamarrenolm: I think it would be preferable to keep these attributes `private`. If you need to use them from subclasses, you can change the access to `protected` instead of `public`. If you need to obtain them from other classes, keep them `private` and increase this class adding a function returning them or (better) crate new methods that extend this class with the required functionality that use the private data without exposing it to the outside world.

colineplqt: I need to access it from subclasses so I put it to protected, thanks!

zamarrenolm (approved):

olperr1 (approved):

zamarrenolm (approved):

PR #3327, opened 2025-02-19
Cgmes import update. Import EQ and SSH separately. TieLines

marqueslanauja:
**Please check if the PR fulfills these requirements**

- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**Does this PR already have an issue describing the problem?**

**What kind of change does this PR introduce?**

**Does this PR introduce a new Powsybl Action implying to be implemented in simulators or pypowsybl?**
- [ ] Yes, the corresponding issue is [here](link)
- [x] No

**What is the current behavior?**

**What is the new behavior (if this is a feature change)?**

**Does this PR introduce a breaking change or deprecate an API?**
- [ ] Yes
- [x] No

**If yes, please check if the following requirements are fulfilled**

- [ ] The *Breaking Change* or *Deprecated* label has been added
- [ ] The migration steps are described in the following section

**What changes might users need to make in their application due to this PR? (migration steps)**

**Other information**:

PR #3570, opened 2025-09-15
Cgmes import update. Import EQ and SSH separately: When data is missing, use previous values or not based on configuration (default is not to use)

marqueslanauja:

Feature

EQ and SSH are imported together during the conversion process.

EQ and SSH are imported separately. 
This PR contains the changes to use the previous update values as default during the next update.

To use the previous values as default, the property `iidm.import.cgmes.use-previous-values-during-update`  must be set to `true` 
Some adjustments have been made to the default values of` generators`, `staticVarCompensators` and `hvdcLines`.

- [x] Yes
- [ ] No

**All these changes concern you only if you override or directly call specific inner methods of the CGMES import.**

The following methodof `DCLinkUpdate` takes an additional parameter:
```java
public DCLinkUpdate(HvdcLine hvdcLine, PropertyBag converter1, PropertyBag converter2, DefaultData defaultData);
# Before, it was:
# public DCLinkUpdate(HvdcLine hvdcLine, PropertyBag converter1, PropertyBag converter2);
```

**Other information**:

### DIFF
--- a/cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/TieLineUpdateTest.java
+++ b/cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/TieLineUpdateTest.java
@@ -94,14 +94,14 @@ class TieLineUpdateTest {
     @Test
     void usePreviousValuesTest() {
         Network network = readCgmesResources(DIR, "tieLine_EQ.xml", "tieLine_EQ_BD.xml", "tieLine_SSH.xml", "tieLine_TP.xml", "tieLine_SV.xml");
-        assertEqCount(network, 2, 5, 4);
+        assertEqCount(network, 2, 4);
         assertFirstSsh(network);
         assertSv(network.getTieLine("ACLineSegment-1 + ACLineSegment-2"));
 
         Properties properties = new Properties();
         properties.put("iidm.import.cgmes.use-previous-values-during-update", "true");
         readCgmesResources(network, properties, DIR, "../empty_SSH.xml", "../empty_SV.xml");
-        assertEqCount(network, 2, 5, 4);
+        assertEqCount(network, 2, 4);
         assertFirstSsh(network);
         assertSv(network.getTieLine("ACLineSegment-1 + ACLineSegment-2"));
     }


cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/DanglingLineUpdateTest.java
```diff
@@ -0,0 +1,242 @@
...
+    void usePreviousValuesTest() {
+        Network network = readCgmesResources(DIR, "danglingLine_EQ.xml", "danglingLine_EQ_BD.xml", "danglingLine_SSH.xml");
+        assertEquals(4, network.getDanglingLineCount());
+        assertFirstSsh(network);
+
+        Properties properties = new Properties();
+        properties.put("iidm.import.cgmes.use-previous-values-during-update", "true");
+        readCgmesResources(network, properties, DIR, "../empty_SSH.xml", "../empty_SV.xml");
+        assertFirstSsh(network);
+    }
```
rcourtier, line 120: I think it would be interesting to import the SV and TP file (in addition to EQ, EQ_BD and SSH), then assert the SV (in addition to assert SSH), and check that after the update with the empty SV, the same assert SV is OK, meaning the update also uses the previous SV values.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/GeneratorUpdateTest.java
```diff
@@ -0,0 +1,142 @@
...
+    void usePreviousValuesTest() {
+        Network network = readCgmesResources(DIR, "generator_EQ.xml", "generator_SSH.xml");
+        assertEquals(3, network.getGeneratorCount());
+        assertFirstSsh(network);
+
+        Properties properties = new Properties();
+        properties.put("iidm.import.cgmes.use-previous-values-during-update", "true");
+        readCgmesResources(network, properties, DIR, "../empty_SSH.xml", "../empty_SV.xml");
+        assertFirstSsh(network);
+    }
```
rcourtier, line 78: Same here, could you please import the SV file and show that the assertFlowsAfterSv is OK before and after the empty SV update?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/dc/DCLinkUpdate.java
```diff
@@ -0,0 +1,201 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(DCLinkUpdate.class);
+
+    public DCLinkUpdate(HvdcLine hvdcLine, PropertyBag converter1, PropertyBag converter2, DefaultData defaultData) {
+        this.hvdcLine = hvdcLine;
+        this.converter1 = converter1;
+        this.converter2 = converter2;
+
+        computeMode(defaultData.mode);
+        boolean isDefaultTargetP = computeActivePowers(defaultData.targetP, defaultData.pDcInverter);
+        computeLossFactors(isDefaultTargetP, defaultData.lossFactor1, defaultData.lossFactor2);
```
rcourtier: With above change proposal, default data become a class attribute and is accessible from within all methods without the need to pass each default data as an argument.
```suggestion
        computeMode();
        computeActivePowers();
        computeLossFactors();
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/dc/DCLinkUpdate.java
```diff
@@ -0,0 +1,201 @@
...
+        if (isRectifier(mode1) && isInverter(mode2)) {
+            mode = SIDE_1_RECTIFIER_SIDE_2_INVERTER;
+        } else if (isInverter(mode1) && isRectifier(mode2)) {
+            mode = SIDE_1_INVERTER_SIDE_2_RECTIFIER;
+        } else if (targetPpcc1() > 0.0 || targetPpcc2() < 0.0) {
+            mode = SIDE_1_RECTIFIER_SIDE_2_INVERTER;
+        } else if (targetPpcc1() < 0.0 || targetPpcc2() > 0.0) {
+            mode = SIDE_1_INVERTER_SIDE_2_RECTIFIER;
+        } else {
+            mode = defaultMode;
```
rcourtier:
```suggestion
            mode = defaultData.mode();
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/dc/DCLinkUpdate.java
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    private double pole2Losses() {
+        double poleLosses = converter2.asDouble(POLE_LOSS_P);
+        return Double.isNaN(poleLosses) ? 0.0 : poleLosses;
+    }
+
+    private boolean computeActivePowers(double defaultTargetP, double defaultPdcInverter) {
+        boolean isDefaultTargetP = false;
+
```
rcourtier:
```suggestion
    private void computeActivePowers(double defaultTargetP, double defaultPdcInverter) {
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/dc/DCLinkUpdate.java
```diff
@@ -0,0 +1,201 @@
...
+            pDcInverter = -1 * (Math.abs(getTargetPpccInverter()) + getPoleLossesInverter());
+            pDcRectifier = Math.abs(pDcInverter) + resistiveLossesFromPdcInverter(pDcInverter);
+            targetP = pDcRectifier + getPoleLossesRectifier();
+        } else {
+            isDefaultTargetP = true;
+            targetP = defaultTargetP;
+            pDcInverter = defaultPdcInverter;
+        }
+        return isDefaultTargetP;
+    }
```
rcourtier, line 112:
```suggestion
        } else {
            targetP = defaultData.targetP();
            // No need to calculate DC powers when using default data.
        }
    }
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/dc/DCLinkUpdate.java
```diff
@@ -0,0 +1,201 @@
...
+            return pole2Losses();
+        }
+        return pole1Losses();
+    }
+
+    private double resistiveLossesFromPdcRectifier(double pDcRectifier) {
+        return resistiveLossesFromPdcRectifier(pDcRectifier, hvdcLine.getR(), hvdcLine.getNominalV());
+    }
+
+    private static double resistiveLossesFromPdcRectifier(double pDcRectifier, double r, double ratedUdc) {
```
rcourtier: This was necessary when `pDcInverter` was part of the default data and needed to be calculated from the static update method. Removing `pDcInverter` from the default data (see below) means we can simplify that.
```suggestion
    private double resistiveLossesFromPdcRectifier(double pDcRectifier) {
        double r = hvdcLine.getR();
        double ratedUdc = hvdcLine.getNominalV();
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/dc/DCLinkUpdate.java
```diff
@@ -0,0 +1,201 @@
...
+        double ratedUdc = hvdcLine.getNominalV();
+        double idc = r != 0.0 ? (ratedUdc - Math.sqrt(ratedUdc * ratedUdc - 4 * r * Math.abs(pDcInverter))) / (2 * r) : 0.0;
+        return r * idc * idc;
+    }
+
+    private void computeLossFactors(boolean isDefaultTargetP, double defaultLossFactor1, double defaultLossFactor2) {
+        // Loss factor is pole losses divided by incoming power.
+        if (isDefaultTargetP || targetP == 0.0) {
+            lossFactor1 = defaultLossFactor1;
+            lossFactor2 = defaultLossFactor2;
```
rcourtier:
```suggestion
        if (targetP == 0.0 || targetP == defaultData.targetP()) {
            lossFactor1 = defaultData.lossFactor1();
            lossFactor2 = defaultData.lossFactor2();
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/dc/DCLinkUpdate.java
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    static double getPdcInverter(HvdcLine hvdcLine) {
+        double poleLossesRectifier = hvdcLine.getConvertersMode() == SIDE_1_RECTIFIER_SIDE_2_INVERTER
+                ? hvdcLine.getConverterStation1().getLossFactor() * hvdcLine.getActivePowerSetpoint() / 100.0
+                : hvdcLine.getConverterStation2().getLossFactor() * hvdcLine.getActivePowerSetpoint() / 100.0;
+        double pDcRectifier = Math.abs(hvdcLine.getActivePowerSetpoint()) - poleLossesRectifier;
+        double resistiveLosses = resistiveLossesFromPdcRectifier(pDcRectifier, hvdcLine.getR(), hvdcLine.getNominalV());
+        return -1 * (pDcRectifier - resistiveLosses);
+    }
```
rcourtier: You don't need that when `pDcInverter` isn't part of the default data.
```suggestion
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/dc/DCLinkUpdate.java
```diff
@@ -0,0 +1,201 @@
...
+
+    record DefaultData(HvdcLine.ConvertersMode mode, double targetP, double pDcInverter, double lossFactor1, double lossFactor2) {
```
rcourtier: `pDcInverter` doesn't need to be part of the default data. The only things we need to calculate everything is:
structural data coming from EQ:
- voltage at rectifier side
- line's resistance
state data coming from SSH/SV:
- active power (at rectifier or at inverter side, one is enough)
- loss or loss factor for both converters

If the active power at rectifier side is provided, we can always calculate active power at inverter side with these other inputs (voltage, resistance, losses).

Also make it public to remove the warning that this record is used outside of this class.

```suggestion
    public record DefaultData(HvdcLine.ConvertersMode mode, double targetP, double lossFactor1, double lossFactor2) {
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/dc/HvdcLineConversion.java
```diff
@@ -11,10 +11,14 @@
 import com.powsybl.cgmes.conversion.Context;
 import com.powsybl.cgmes.conversion.Conversion;
 import com.powsybl.cgmes.conversion.elements.AbstractIdentifiedObjectConversion;
-import com.powsybl.iidm.network.HvdcLine;
-import com.powsybl.iidm.network.HvdcLineAdder;
+import com.powsybl.iidm.network.*;
+import com.powsybl.triplestore.api.PropertyBag;
 
+import static com.powsybl.cgmes.conversion.elements.AbstractConductingEquipmentConversion.updateTerminals;
+import static com.powsybl.cgmes.conversion.elements.dc.DCLinkUpdate.getPdcInverter;
```
rcourtier: This is not needed. To be removed.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/dc/HvdcLineConversion.java
```diff
@@ -58,4 +68,43 @@ public void convert() {
...
+                .setActivePowerSetpoint(dcLinkUpdate.getTargetP())
+                .setMaxP(DEFAULT_MAXP_FACTOR * dcLinkUpdate.getTargetP());
+        return dcLinkUpdate;
+    }
+
+    private static DCLinkUpdate.DefaultData getDefaultData(HvdcLine hvdcLine, Context context) {
+        boolean defaultRectifierOnSide1 = getDefaultValue(null, isRectifierOnSide1(hvdcLine.getConvertersMode()), true, true, context);
+        HvdcLine.ConvertersMode defaultMode = defaultRectifierOnSide1 ? SIDE_1_RECTIFIER_SIDE_2_INVERTER : SIDE_1_INVERTER_SIDE_2_RECTIFIER;
+        double defaultTargetP = getDefaultValue(null, hvdcLine.getActivePowerSetpoint(), 0.0, 0.0, context);
+        double defaultPdcInverter = getDefaultValue(null, getPdcInverter(hvdcLine), 0.0, 0.0, context);
```
rcourtier: This is not needed. To be removed.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/dc/HvdcLineConversion.java
```diff
@@ -58,4 +68,43 @@ public void convert() {
...
+        double lossFactor1 = getDefaultValue(null, (double) hvdcLine.getConverterStation1().getLossFactor(), 0.0, 0.0, context);
+        double lossFactor2 = getDefaultValue(null, (double) hvdcLine.getConverterStation2().getLossFactor(), 0.0, 0.0, context);
+
+        return new DCLinkUpdate.DefaultData(defaultMode, defaultTargetP, defaultPdcInverter, lossFactor1, lossFactor2);
```
rcourtier:
```suggestion
        return new DCLinkUpdate.DefaultData(defaultMode, defaultTargetP, lossFactor1, lossFactor2);
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/dc/DCLinkUpdate.java
```diff
@@ -0,0 +1,201 @@
...
+/**
+ * @author Romain Courtier {@literal <romain.courtier at rte-france.com>}
+ */
+public class DCLinkUpdate {
+
+    private ConvertersMode mode;
+    private double targetP;
+    private double pDcInverter;
+    private double lossFactor1;
+    private double lossFactor2;
```
rcourtier, line 28: You can create a class attribute to hold default data. That way, you won't need to pass each default data to the compute methods.
```suggestion
    private double lossFactor2;
    private final DefaultData defaultData;
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/dc/DCLinkUpdate.java
```diff
@@ -0,0 +1,201 @@
...
+
+    private static final String OPERATING_MODE = "operatingMode";
+    private static final String TARGET_PPCC = "targetPpcc";
+    private static final String POLE_LOSS_P = "poleLossP";
+    private static final Logger LOG = LoggerFactory.getLogger(DCLinkUpdate.class);
+
+    public DCLinkUpdate(HvdcLine hvdcLine, PropertyBag converter1, PropertyBag converter2, DefaultData defaultData) {
+        this.hvdcLine = hvdcLine;
+        this.converter1 = converter1;
+        this.converter2 = converter2;
```
rcourtier, line 43:
```suggestion
        this.converter2 = converter2;
        this.defaultData = defaultData;
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/dc/HvdcLineConversion.java
```diff
@@ -39,12 +43,18 @@ public boolean valid() {
 
     @Override
     public void convert() {
+
+        // arbitrary value because there is no maxP attribute in CGMES
+        double maxP = 0.0;
+        HvdcLine.ConvertersMode convertersMode = HvdcLine.ConvertersMode.SIDE_1_RECTIFIER_SIDE_2_INVERTER;
+        double activePowerSetpoint = 0.0;
```
rcourtier, line 48: Could you please inline these values in the adder instead? That would remove a warning in my IDE (IntelliJ).
And since you have added with this PR an static import of ConvertersMode enum, you don't need to add the `HvdcLine.ConvertersMode.` prefix anymore.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/HvdcUpdateTest.java
```diff
@@ -0,0 +1,218 @@
...
+    void usePreviousValuesTest() {
+        Network network = readCgmesResources(DIR, "hvdc_EQ.xml", "hvdc_SSH.xml");
+        assertEquals(2, network.getHvdcLineCount());
+        assertFirstSsh(network);
+
+        Properties properties = new Properties();
+        properties.put("iidm.import.cgmes.use-previous-values-during-update", "true");
+        readCgmesResources(network, properties, DIR, "../empty_SSH.xml", "../empty_SV.xml");
+        assertFirstSsh(network);
+    }
```
rcourtier, line 85: Could you please import the SSH_1 and SV file and show that the assertLossFactorAfterSshSv is OK before and after the empty SV update?

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/LineUpdateTest.java
```diff
@@ -0,0 +1,180 @@
...
+    void usePreviousValuesTest() {
+        Network network = readCgmesResources(DIR, "line_EQ.xml", "line_SSH.xml");
+        assertEquals(3, network.getLineCount());
+        assertFirstSsh(network);
+
+        Properties properties = new Properties();
+        properties.put("iidm.import.cgmes.use-previous-values-during-update", "true");
+        readCgmesResources(network, properties, DIR, "../empty_SSH.xml", "../empty_SV.xml");
+        assertFirstSsh(network);
+    }
```
rcourtier, line 73: Same here, could you please also import the SV file and show that the assertFlowsAfterSv is OK before and after the empty SV update?

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/LoadUpdateTest.java
```diff
@@ -0,0 +1,142 @@
...
+    void usePreviousValuesTest() {
+        Network network = readCgmesResources(DIR, "load_EQ.xml", "load_SSH.xml");
+        assertEquals(6, network.getLoadCount());
+        assertFirstSsh(network);
+
+        Properties properties = new Properties();
+        properties.put("iidm.import.cgmes.use-previous-values-during-update", "true");
+        readCgmesResources(network, properties, DIR, "../empty_SSH.xml", "../empty_SV.xml");
+        assertFirstSsh(network);
+    }
```
rcourtier, line 75: Same here, could you please also import the SV file and show that the assertFlowsAfterSv is OK before and after the empty SV update?

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/ShuntCompensatorUpdateTest.java
```diff
@@ -0,0 +1,148 @@
...
+    void usePreviousValuesTest() {
+        Network network = readCgmesResources(DIR, "shuntCompensator_EQ.xml", "shuntCompensator_SSH.xml");
+        assertEquals(3, network.getShuntCompensatorCount());
+        assertFirstSsh(network);
+
+        Properties properties = new Properties();
+        properties.put("iidm.import.cgmes.use-previous-values-during-update", "true");
+        readCgmesResources(network, properties, DIR, "../empty_SSH.xml", "../empty_SV.xml");
+        assertFirstSsh(network);
+    }
```
rcourtier, line 80: Same here, could you please also import the SV file and show that the assertFlowsAfterSv and assertSectionsAfterSv is OK before and after the empty SV update?

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/StaticVarCompensatorUpdateTest.java
```diff
@@ -0,0 +1,128 @@
...
+    void usePreviousValuesTest() {
+        Network network = readCgmesResources(DIR, "staticVarCompensator_EQ.xml", "staticVarCompensator_SSH.xml");
+        assertEquals(2, network.getStaticVarCompensatorCount());
+        assertFirstSsh(network);
+
+        Properties properties = new Properties();
+        properties.put("iidm.import.cgmes.use-previous-values-during-update", "true");
+        readCgmesResources(network, properties, DIR, "../empty_SSH.xml", "../empty_SV.xml");
+        assertFirstSsh(network);
+    }
```
rcourtier, line 76: Same here, could you please also import the SV file and show that the assertFlowsAfterSv is OK before and after the empty SV update?

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/TieLineUpdateTest.java
```diff
@@ -0,0 +1,178 @@
...
+    void usePreviousValuesTest() {
+        Network network = readCgmesResources(DIR, "tieLine_EQ.xml", "tieLine_EQ_BD.xml", "tieLine_SSH.xml");
+        assertEquals(1, network.getTieLineCount());
+        assertFirstSsh(network);
+
+        Properties properties = new Properties();
+        properties.put("iidm.import.cgmes.use-previous-values-during-update", "true");
+        readCgmesResources(network, properties, DIR, "../empty_SSH.xml", "../empty_SV.xml");
+        assertFirstSsh(network);
+    }
```
rcourtier, line 107: Same here, could you please also import the TP and SV file and show that the assertSv(tieLine) is OK before and after the empty SV update?

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/TransformerUpdateTest.java
```diff
@@ -0,0 +1,323 @@
...
+        Network network = readCgmesResources(DIR, "transformer_EQ.xml", "transformer_SSH.xml");
+        assertEquals(1, network.getTwoWindingsTransformerCount());
+        assertEquals(1, network.getThreeWindingsTransformerCount());
+        assertFirstSsh(network);
+
+        Properties properties = new Properties();
+        properties.put("iidm.import.cgmes.use-previous-values-during-update", "true");
+        readCgmesResources(network, properties, DIR, "../empty_SSH.xml", "../empty_SV.xml");
+        assertFirstSsh(network);
+    }
```
rcourtier, line 113: Same here, could you please also import the SV file and show that the assertFlowsAfterSv and assertTapChangerStepsAfterSv are OK before and after the empty SV update?

rcourtier (changes requested):
Thank you for your work!
I have what it seems to be a lot of changes requested, but in the end it's all about 2 things:
- Simplify the use of default data for Hvdc during conversion.
- Assert in the unit tests that the previous SV values can also be used during update.
Thanks again!

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/EnergyConsumerConversion.java
```diff
@@ -115,33 +116,61 @@ private void addZipModel(LoadAdder adder) {
...
-        return powerFlow().defined() ? powerFlow().p() : p.asDouble("pFixed", 0.0);
+    public static void update(Load load, PropertyBag cgmesData, Context context) {
+        updateTerminals(load, context, load.getTerminal());
+
+        double pFixed = Double.parseDouble(load.getProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.P_FIXED, "0.0"));
+        double qFixed = Double.parseDouble(load.getProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.Q_FIXED, "0.0"));
+
+        PowerFlow updatedPowerFlow = updatedPowerFlow(load, cgmesData, context);
+        load.setP0(updatedPowerFlow.defined() ? updatedPowerFlow.p() : getDefaultP0(load, pFixed, context));
+        load.setQ0(updatedPowerFlow.defined() ? updatedPowerFlow.q() : getDefaultQ0(load, qFixed, context));
```
rcourtier, line 146: When comparing the following use cases:
- Import EQ+SSH+SV
- Import EQ, then update SSH, then update SV with the use of previous values (from SSH)
We end up in different values for P0/Q0.

In the first case, the final P0/Q0 values are the ones from SSH.
In the second case, the final P0/Q0 values are the ones from SV, even though SSH P0/S0 values were already there and the option to use previous values was activated.

I understand P0/Q0 are needed to have a SSH ValidationLevel and it makes sense to initialize them with (by descending priority order):
- Values coming from SSH
- Values coming from SV
- Default value (0.0, 0.0)

But I don't think it's legitimate to prefer SV values over SSH values for "target" fields when the option to use previous values has been raised.

I have made this comment for Loads but it also applies for every object with P0/Q0 fields.

rcourtier (changes requested):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/AbstractIdentifiedObjectConversion.java
```diff
@@ -93,6 +98,33 @@ protected String complete(String what) {
         }
     }
 
+    protected static double getDefaultValue(Double equipmentValue, Double previousValue, Double defaultValue, double emptyValue, Context context) {
+        return getDefaultValue(equipmentValue, previousValue, defaultValue, emptyValue, context.config().updateDefaultValuesPriority());
+    }
```
rcourtier, line 106: When considering the following use case:
- Import EQ, then update SSH with the use of previous values

There is a blocking error during the update.

The issue happens when we have an object present in EQ but not in SSH, and this object has a field of type double that has been initialized to Double.NaN. For example, let's consider a Load present in the EQ file but not present in the SSH file. After the EQ import, its P0/Q0 values are initialized to Double.NaN. And when performing the SSH update, this Load P0/Q0 are "updated" with the previous values which are Double.NaN/Double.NaN. This works until the final step of the update process which is calling 
```java
network.setMinimumAcceptableValidationLevel(ValidationLevel.STEADY_STATE_HYPOTHESIS);
```
This throws an error because the validation level of that Load is still `EQUIPMENT` because P0/Q0 are still equal to Double.NaN.

To fix this, the solution would be to consider that a `previousValue` equals to `Double.NaN` is improper and force it to `null`. That way it's not chosen and the equipment or default value (0.0 in our case instead of null) is selected.

```java
    protected static double getDefaultValue(Double equipmentValue, Double previousValue, Double defaultValue, double emptyValue, Context context) {
        Double properPreviousValue = Double.isNaN(previousValue) ? null : previousValue;
        return getDefaultValue(equipmentValue, properPreviousValue, defaultValue, emptyValue, context.config().updateDefaultValuesPriority());
    }
```

You might also want to generalize it to equipment and/or default value. In that case, you could define a predicate to check if the value is proper or not:
```java
    protected static double getDefaultValue(Double equipmentValue, Double previousValue, Double defaultValue, double emptyValue, Context context) {
        return getDefaultValue(equipmentValue, previousValue, defaultValue, emptyValue, context.config().updateDefaultValuesPriority(),
                val -> val == null || Double.isNaN(val));
    }

    protected static int getDefaultValue(Integer equipmentValue, Integer previousValue, Integer defaultValue, int emptyValue, Context context) {
        return getDefaultValue(equipmentValue, previousValue, defaultValue, emptyValue, context.config().updateDefaultValuesPriority(), Objects::isNull);
    }

    protected static boolean getDefaultValue(Boolean equipmentValue, Boolean previousValue, Boolean defaultValue, boolean emptyValue, Context context) {
        return getDefaultValue(equipmentValue, previousValue, defaultValue, emptyValue, context.config().updateDefaultValuesPriority(), Objects::isNull);
    }

    private static <T> T getDefaultValue(T equipmentValue, T previousValue, T defaultValue, T emptyValue, List<Conversion.Config.DefaultValue> selectors,
                                         Predicate<T> isEmpty) {
        for (Conversion.Config.DefaultValue selector : selectors) {
            if (selector == EQ && !isEmpty.test(equipmentValue)) {
                return equipmentValue;
            } else if (selector == PREVIOUS && !isEmpty.test(previousValue)) {
                return previousValue;
            } else if (selector == DEFAULT && !isEmpty.test(defaultValue)) {
                return defaultValue;
            } else if (selector == EMPTY) {
                return emptyValue;
            }
        }
        return emptyValue;
    }
```

rcourtier (changes requested):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/dc/HvdcLineConversion.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/dc/HvdcLineConversion.java
marqueslanauja: Done. I've removed the unnecessary attributes

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/dc/HvdcLineConversion.java
marqueslanauja: Done. I've removed the unnecessary imports.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java
```diff
@@ -70,12 +63,108 @@ public void convert() {
...
+        convertedTerminalsWithOnlyEq(shunt.getTerminal());
         context.regulatingControlMapping().forShuntCompensators().add(shunt.getId(), p);
+        shunt.setProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.NORMAL_SECTIONS, String.valueOf(normalSections));
+    }
+
+    public static void update(ShuntCompensator shuntCompensator, PropertyBag cgmesData, Context context) {
+        updateTerminals(shuntCompensator, context, shuntCompensator.getTerminal());
+        updateSections(shuntCompensator, cgmesData, context);
+
+        boolean controlEnabled = cgmesData.asBoolean(CgmesNames.CONTROL_ENABLED, false);
```
rcourtier: Could you please handle the case where control enabled isn't part of the CGMES input data? Right now it's always defaulted to `false` if it's not present, but the previous value should be used instead (if the corresponding property has been set). Just like it has been done for static var compensators and others.

rcourtier (changes requested):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
```diff
@@ -624,6 +645,7 @@ private void copyStream(ReadOnlyDataSource from, DataSource to, String fromName,
     public static final String IMPORT_CGM_WITH_SUBNETWORKS = "iidm.import.cgmes.cgm-with-subnetworks";
     public static final String IMPORT_CGM_WITH_SUBNETWORKS_DEFINED_BY = "iidm.import.cgmes.cgm-with-subnetworks-defined-by";
     public static final String CREATE_FICTITIOUS_VOLTAGE_LEVEL_FOR_EVERY_NODE = "iidm.import.cgmes.create-fictitious-voltage-level-for-every-node";
+    public static final String USE_PREVIOUS_VALUES_DURING_UPDATE = "iidm.import.cgmes.use-previous-values-during-update";
```
olperr1, line 648: This new option should be defined in the documentation (in `docs/grid_exchange_formats/cgmes/import.md`).

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
```diff
@@ -737,6 +759,12 @@ private void copyStream(ReadOnlyDataSource from, DataSource to, String fromName,
             Boolean.TRUE)
             .addAdditionalNames("createFictitiousVoltageLevelForEveryNode");
 
+    private static final Parameter USE_PREVIOUS_VALUES_DURING_UPDATE_PARAMETER = new Parameter(
+            USE_PREVIOUS_VALUES_DURING_UPDATE,
+            ParameterType.BOOLEAN,
+            "Use previous values (from a previous update) during the current update",
+            Boolean.FALSE);
```
olperr1, line 766: Shouldn't we set the default value to `TRUE`?
When someone load an EQ file then update their network with an SSH one, the update is incremental. It may be more intuitive to keep this "incremental update" behavior by default.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/DanglingLineUpdateTest.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/GeneratorUpdateTest.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/HvdcUpdateTest.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/LineUpdateTest.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/LoadUpdateTest.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/ShuntCompensatorUpdateTest.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/StaticVarCompensatorUpdateTest.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/TieLineUpdateTest.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/TransformerUpdateTest.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/AbstractIdentifiedObjectConversion.java
marqueslanauja: Done. Implemented only for the previous value.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/ShuntConversion.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/EnergyConsumerConversion.java
marqueslanauja: The `previousValue` is a default option and only applies when no value is received during the update process and the use of previous values has been enabled. In this case, the value recorded in the equipment is used without analyzing how it was assigned during the previous update.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/TieLineUpdateTest.java
rcourtier: Could you please add calls to `assertSv` before and after the update with empty SSH/SV files?

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/ShuntCompensatorUpdateTest.java
rcourtier: Could you please add calls to `assertFlowsAfterSv` before and after the update with empty SSH/SV files?

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/HvdcUpdateTest.java
```diff
@@ -0,0 +1,221 @@
...
+
+        readCgmesResources(network, DIR, "hvdc_SSH_1.xml", "hvdc_SV.xml");
+        assertSecondSsh(network);
+        assertLossFactorAfterSshSv(network);
+
+        Properties properties = new Properties();
+        properties.put("iidm.import.cgmes.use-previous-values-during-update", "true");
+        readCgmesResources(network, properties, DIR, "../empty_SSH.xml", "../empty_SV.xml");
+        assertLossFactorAfterSshSv(network);
+    }
```
rcourtier, line 85:
```suggestion
    @Test
    void usePreviousValuesTest() {
        Network network = readCgmesResources(DIR, "hvdc_EQ.xml", "hvdc_SSH_1.xml", "hvdc_SV.xml");
        assertEquals(2, network.getHvdcLineCount());
        assertSecondSsh(network);
        assertLossFactorAfterSshSv(network);

        Properties properties = new Properties();
        properties.put("iidm.import.cgmes.use-previous-values-during-update", "true");
        readCgmesResources(network, properties, DIR, "../empty_SSH.xml", "../empty_SV.xml");
        assertLossFactorAfterSshSv(network);
    }
```

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/DanglingLineUpdateTest.java
rcourtier: Could you please create a method `assertSvTogether` that checks the SV content (basically what's done in `importSvTogetherTest()`), and call that here before and after the update with empty SSH/SV files?

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/HvdcUpdateTest.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/TieLineUpdateTest.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/ShuntCompensatorUpdateTest.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/DanglingLineUpdateTest.java
marqueslanauja: Done

olperr1 (changes requested):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
marqueslanauja: We decided to set the default to `false` because, when the use of previous values is enabled, the results obtained may depend on the update sequence of the SSH files. By requiring the user to enable this option explicitly, we ensure they are aware of this dependency.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
marqueslanauja: Done

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/HvdcUpdateTest.java
```diff
@@ -0,0 +1,217 @@
...
+        assertLossFactorBeforeSshSv(network);
+        readCgmesResources(network, DIR, "hvdc_SSH_1.xml", "hvdc_SV.xml");
+        assertLossFactorAfterSshSv(network);
+    }
+
+    @Test
+    void usePreviousValuesTest() {
+        Network network = readCgmesResources(DIR, "hvdc_EQ.xml", "hvdc_SSH_1.xml", "hvdc_SV.xml");
+        assertEquals(2, network.getHvdcLineCount());
+        assertSecondSsh(network);
```
rcourtier, line 75:
```suggestion
        assertSecondSsh(network);
        assertFlowsAfterSv(network);
```

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/HvdcUpdateTest.java
```diff
@@ -0,0 +1,217 @@
...
+        assertLossFactorAfterSshSv(network);
+
+        Properties properties = new Properties();
+        properties.put("iidm.import.cgmes.use-previous-values-during-update", "true");
+        readCgmesResources(network, properties, DIR, "../empty_SSH.xml", "../empty_SV.xml");
+        assertLossFactorAfterSshSv(network);
```
rcourtier, line 84:
```suggestion
        assertSecondSsh(network);
        assertFlowsAfterSv(network);
        assertLossFactorAfterSshSv(network);
```

rcourtier (changes requested):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
olperr1: Thanks

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/TieLineUpdateTest.java
```diff
@@ -89,6 +91,21 @@ void importSvSeparatelyTest() {
         assertSv(tieLine);
     }
 
+    @Test
+    void usePreviousValuesTest() {
+        Network network = readCgmesResources(DIR, "tieLine_EQ.xml", "tieLine_EQ_BD.xml", "tieLine_SSH.xml", "tieLine_TP.xml", "tieLine_SV.xml");
+        assertEqCount(network, 2, 5, 4);
```
olperr1: In the TieLine "1 EQ / 24 SSH" PR (#3327), the second int parameter of this method was removed since it was always valued to 5.
Can you fix this PR, now that all the other PRs were merged in main?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
olperr1: OK.
I think it would be very useful to have the different use cases documented, with the configuration to use for each of them.
But this can be done in another PR (even after the RC1 is published).

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/TieLineUpdateTest.java
marqueslanauja: Done

olperr1: Thanks!

PR #3572, opened 2025-09-17
Fix bad overload detection on last temporary limit

FranckLecuyer:

Bug fix

Overload on the last temporary limit (the temporary limit with highest value) is not detected

Overload on the last temporary limit (the temporary limit with highest value) is now properly detected

- [ ] Yes
- [x] No

**Other information**:


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/LimitViolationUtils.java
```diff
@@ -69,12 +90,17 @@ private static OverloadImpl getOverload(LoadingLimits limits, double i, double l
...
         for (LoadingLimits.TemporaryLimit tl : temporaryLimits) { // iterate in ascending order
             if (i >= previousLimit * limitReductionValue && i < tl.getValue() * limitReductionValue) {
-                return new OverloadImpl(tl, previousLimitName, previousLimit);
+                return createOverload(tl, previousLimit, previousLimitName, null, false, -1);
             }
             previousLimitName = tl.getName();
             previousLimit = tl.getValue();
+            lastTemporaryLimit = tl;
+        }
+        if (lastTemporaryLimit != null && i >= lastTemporaryLimit.getValue()) {
```
colineplqt: Why is the limit reduction not taken into account in this case?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/LimitViolationUtils.java
```diff
@@ -69,12 +90,17 @@ private static OverloadImpl getOverload(LoadingLimits limits, double i, double l
         Collection<LoadingLimits.TemporaryLimit> temporaryLimits = limits.getTemporaryLimits();
         String previousLimitName = PERMANENT_LIMIT_NAME;
         double previousLimit = permanentLimit;
+        LoadingLimits.TemporaryLimit lastTemporaryLimit = null;
         for (LoadingLimits.TemporaryLimit tl : temporaryLimits) { // iterate in ascending order
             if (i >= previousLimit * limitReductionValue && i < tl.getValue() * limitReductionValue) {
-                return new OverloadImpl(tl, previousLimitName, previousLimit);
+                return createOverload(tl, previousLimit, previousLimitName, null, false, -1);
```
colineplqt: Here I also think it's odd that the returned overload does not have the limit reduction (it will be 1 in createOverload even though it could be different in this method. Maybe it would be good to have it also as a parameter of the method createOverload?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/LimitViolationUtils.java
FranckLecuyer: Fixed

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/LimitViolationUtils.java
FranckLecuyer: Changed

Issue #3527, opened 2025-07-23
IIDM shouldn't accept negative r values for HVDCLines


rcourtier:
### Describe the current behavior

- IIDM accepts negative r values for HvdcLine.
- Negative r values are rejected for DCLine (detailed DC model).

### Describe the expected behavior
Negative r values should also be rejected for HvdcLine (simplified DC model).

=> The add() method of HvdcLineAdder should not only verify that r isn't Double.NaN, but also verify that r is a positive or null double.

### Extra Information

- After consultation, we should accept negative r values for lines and transformers. See comments on #3576.
- There's already a check to prevent negative r values on DCLine (detailed DC model).

PR #3576, opened 2025-09-25
Check if r is a positive or null double (for lines and 2w-transfos)

cphili:

https://github.com/powsybl/powsybl-core/issues/3527

In LineAdderImpl, HvdcLineAdder and TwoWindingsTransformerAdderImpl classes : the method add() checks if r values are invalid (NaN).

In these cases, the method add() now also checks if r values are negative.

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
@@ -299,6 +299,15 @@ public final class ValidationUtil {
         }
     }
 
+    public static void checkPositiveR(Validable validable, double r) {
+        if (Double.isNaN(r)) {
+            throw new ValidationException(validable, "r is invalid");
+        }
+        if (r < 0) {
+            throw new ValidationException(validable, "r must be positive");
+        }
+    }
+
     public static void checkX(Validable validable, double x) {
         if (Double.isNaN(x)) {
             throw new ValidationException(validable, "x is invalid");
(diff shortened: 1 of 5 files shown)


jeandemanged (changes requested):
Hello,

I would be against this change unless:
1. there is an opt-out mecanism for the negative R check (e.g. must be flagged as fictitious or other mecanism)
2. this change is flagged as breaking change

and also probably 3/ need backward compatibility for existing iIDMs serialized with negative R.

In CGMES today:
- it is perfectly valid for a cim:EquivalentBranch to have a negative R, and some TSOs are using this
- few TSOs are sending negativeR for 2WT, while this is indeed not-so-good, today QoCDC does not block publication (info level only)

geofjamg:
> Hello,
> 
> I would be against this change unless:
> 
> 1. there is an opt-out mecanism for the negative R check (e.g. must be flagged as fictitious or other mecanism)
> 2. this change is flagged as breaking change
> 
> and also probably 3/ need backward compatibility for existing iIDMs serialized with negative R.
> 
> In CGMES today:
> 
> * it is perfectly valid for a cim:EquivalentBranch to have a negative R, and some TSOs are using this
> * few TSOs are sending negativeR for 2WT, while this is indeed not-so-good, today QoCDC does not block publication (info level only)

This is a good question. Does classic reduction algorithm (like ward injection) create negative r ? If it does, we indeed should remove this check.

JB-H:
I agree with geofjamg and jeandemanged, such modification is questionable,  as grid equivalent might provide negative R on some lines while fully respecting the load flow on the non-transformed part. Secondly, it is possible that the physical model of a transformer with positive R might be equivalenced by a PI-L-T equivalent model containing a negative R. Again, Equivalent model is not the same as a physical model using phyical parameters. Here we are putting the test on a non-physical model, which can lead to block something that is physically existing

olperr1:
After consultation, negative r values should be accepted for lines and transformers.

PR #3582, opened 2025-09-30
Improve sensitivity analysis run parameters

### LINKED ISSUE #3409: Improve sensitivity analysis run parameters

To run a sensitivity analysis on a `Network`, a lot of parameters must be set.

To run a sensitivity analysis on a `Network`, we can precise the network, sensitivity factors and only the wanted parameters to override (none if need be).

### Describe the motivation

For some use cases the use of `SensitivityAnalysis::run` does not need a lot of parameters. The mandatory parameters should always be the `Network` on which the analysis is wanted and a list of `SensitivityFactor`.  
The other parameters could be set by default and overridden only if needed:
- variant: `VariantManagerConstants.INITIAL_VARIANT_ID`
- List of `SensitivityVariableSet` : empty list
- List of `Contingency`: empty list
- parameters of the `SensitivityAnalysis`: `SensitivityAnalysisParameters.load()`
- computation manager: `LocalComputationManager.getDefault()`
- report node: `ReportNode.NO_OP`

For that, we can create methods in the `SensitivityAnalysis` `Runner` inner class that sets all the parameters by default and allows them to be overridden, like this following example:
```
SensitivityAnalysis.find(network, sensitivityFactors)
    .setComputationManager(manager)
    .setReportNode(report)
    .run();
```

### Extra Information

We keep all the current `SensitivityAnalysis::run` methods for simple cases.

quinarygio:

Fixes #3409 

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysis.java
+++ b/sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysis.java
@@ -41,7 +41,15 @@ public final class SensitivityAnalysis {
      */
     public static final class Runner implements Versionable {
 
+        private Network network;
         private final SensitivityAnalysisProvider provider;
+        private List<SensitivityFactor> sensitivityFactors;
+        private String variant = VariantManagerConstants.INITIAL_VARIANT_ID;
+        private List<SensitivityVariableSet> variableSets = Collections.emptyList();
+        private List<Contingency> contingencies = Collections.emptyList();
+        private SensitivityAnalysisParameters parameters = SensitivityAnalysisParameters.load();
+        private ComputationManager computationManager = LocalComputationManager.getDefault();
+        private ReportNode reportNode = ReportNode.NO_OP;
 
         public Runner(SensitivityAnalysisProvider provider) {
             this.provider = Objects.requireNonNull(provider);
@@ -93,6 +101,12 @@ public final class SensitivityAnalysis {
                     .thenApply(unused -> new SensitivityAnalysisResult(factors, resultWriter.getContingencyStatuses(), resultWriter.getValues()));
         }
 
+        public CompletableFuture<SensitivityAnalysisResult> runAsync(Network network,
+                                                                     List<SensitivityFactor> factors) {
+
+            return runAsync(network, this.variant, factors, this.contingencies, this.variableSets, this.parameters, this.computationManager, this.reportNode);
+        }
+
         public void run(Network network,
                         String workingVariantId,
                         SensitivityFactorReader factorReader,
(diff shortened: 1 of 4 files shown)


alicecaron (approved):

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysis.java
```diff
@@ -203,6 +262,11 @@ public static CompletableFuture<Void> runAsync(Network network,
         return find().runAsync(network, workingVariantId, factorReader, resultWriter, contingencies, variableSets, parameters, computationManager, reportNode);
     }
 
+    public static CompletableFuture<SensitivityAnalysisResult> runAsync(Network network,
+                                                                        List<SensitivityFactor> factors) {
+        return find().runAsync(network, factors);
+    }
```
rolnico: What is this method needed for?

geofjamg (changes requested):
This is a very different design than the model one that is SecurityAnalysis.
In SecurityAnalysis the Runner Object is a stateless one, everything is carried by a bean object RunParameter. Here you are  making Runner the one with has the state.
For me this is a worth design and moreover this is inconsistent with the SecurityAnalysis One.
I can see that LoadFlow API has also be done like this and should be fixed before next release.

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysis.java
```diff
@@ -113,48 +159,104 @@ public SensitivityAnalysisResult run(Network network,
...
+         * @deprecated use {@link #run(Network, List, SensitivityAnalysisRunParameters)} instead
+         */
+        @Deprecated(since = "6.9.0", forRemoval = true)
         public SensitivityAnalysisResult run(Network network,
                                              List<SensitivityFactor> factors,
                                              List<Contingency> contingencies,
                                              List<SensitivityVariableSet> variableSets,
                                              SensitivityAnalysisParameters parameters) {
-            return run(network, VariantManagerConstants.INITIAL_VARIANT_ID, factors, contingencies, variableSets, parameters);
+            return run(network, network.getVariantManager().getWorkingVariantId(), factors,
```
olperr1: We previously had `VariantManagerConstants.INITIAL_VARIANT_ID` here... which is weird because it is not really intuitive and it is not documented.

I think that this new value makes more sense, but that it should be changed in another PR to highlight this modification.

=>
- In this PR: use `VariantManagerConstants.INITIAL_VARIANT_ID` as default variant ID.
- New PR to create to change this value to network.getVariantManager().getWorkingVariantId()

olperr1 (changes requested):

olperr1 (approved):

geofjamg (approved):

Issue #3577, opened 2025-09-26
Add new parameter option to run a power flow only on the main synchronous component


alicecaron:
### Describe the current behavior

Currently there is only 2 possible modes to run power flow by defining the `LoadFlowParameters` `ConnectedComponentMode`. The values are:
- `MAIN` (default): to run on the main connected component
- `ALL` to run on all connected components

### Describe the expected behavior

It is possible to run the calculation on only the main synchronous component by using a new `ConnectedComponentMode` mode when setting the `LoadFlowParameters`. The new mode is: `MAIN_SYNCHRONOUS`. 

### Describe the motivation

This allows to be more selective when interested in only the main component results simulation.

### Extra Information

- create a new mode: `ConnectedComponentMode::MAIN_SYNCHRONOUS`
- Update the documentation (https://powsybl.readthedocs.io/projects/powsybl-core/en/stable/simulation/loadflow/configuration.html#parameters)

jeandemanged:
Hi

I think it would be cleaner to rename the enum to `ConnectedComponentMode` to just `ComponentMode` with enum constants:
- MAIN_CONNECTED
- ALL_CONNECTED
- MAIN_SYNCHRONOUS

also don't forget that:
- this requires to bump load flow parameters api version from 1.9 to 1.10, as well as adding backward compatibility support for JSONs
- pypowsybl is also impacted

vidaldid-rte:
@jeandemanged  @cphili 
  If we change the enum values, how do we manage compatibility
     - For existing json files with parameters that users have
     - For pypowsyb users and their scripts ?

PR #3583, opened 2025-10-01
Add new parameter option to run a power flow only on the main synchronous component

cphili:

https://github.com/powsybl/powsybl-core/issues/3577

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowParameters.java
+++ b/loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowParameters.java
@@ -71,6 +71,7 @@ public class LoadFlowParameters extends AbstractExtendable<LoadFlowParameters> {
     public enum ConnectedComponentMode {
         MAIN,
         ALL,
+        MAIN_SYNCHRONOUS
     }
 
     public static final Logger LOGGER = LoggerFactory.getLogger(LoadFlowParameters.class);


loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowParameters.java
```diff
@@ -153,7 +168,8 @@ protected static void load(LoadFlowParameters parameters, PlatformConfig platfor
                     config.getOptionalEnumProperty("balanceType", BalanceType.class).ifPresent(parameters::setBalanceType);
                     config.getOptionalBooleanProperty("dcUseTransformerRatio").ifPresent(parameters::setDcUseTransformerRatio);
                     config.getOptionalEnumSetProperty("countriesToBalance", Country.class).ifPresent(parameters::setCountriesToBalance);
-                    config.getOptionalEnumProperty("connectedComponentMode", ConnectedComponentMode.class).ifPresent(parameters::setConnectedComponentMode);
+                    config.getOptionalEnumProperty("componentMode", ComponentMode.class).ifPresentOrElse(parameters::setComponentMode,
+                            () -> config.getOptionalEnumProperty("connectedComponentMode", ComponentMode.class).ifPresent(parameters::setComponentMode));
```
vidaldid-rte, line 172: Does it break if someone has the old property in its paltform file and not the new one ?
Is it on purpose ? (I don't see how you convert the old values to the new one).

If it is on purpose to break is it possible to display a better error message explaining the new property name and possible values ?

vidaldid-rte: Other question; 
  Is there a mecanism to avoid breaking pypowsybl scripts that used the connectedComponentMode parameter ?

vidaldid-rte: Handled by pypowsybl.

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowParameters.java
```diff
@@ -85,7 +99,8 @@ public enum ConnectedComponentMode {
     // VERSION = 1.7 hvdcAcEmulation
     // VERSION = 1.8 noGeneratorReactiveLimits -> useReactiveLimits
     // VERSION = 1.9 dcPowerFactor
-    public static final String VERSION = "1.9";
+    // VERSION = 2.0 componentMode instead of connectedComponentMode
```
colineplqt: Why do you change the version to 2.0? I think 1.10 would be better for this change.

loadflow/loadflow-api/src/test/java/com/powsybl/loadflow/json/JsonLoadFlowParametersTest.java
```diff
@@ -145,6 +145,13 @@ void readJsonVersion19() {
         assertEquals(0.8d, parameters.getDcPowerFactor(), 0d);
     }
 
+    @Test
+    void readJsonVersion20() {
```
colineplqt: If you change the version, the name of this test and of the resource file should also be changed.

sensitivity-analysis-api/src/test/resources/SensitivityAnalysisParametersV1.0Invalid.json
colineplqt, line 4: Should this version not be updated too.

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowParameters.java
colineplqt: It would be better to fix the method `JsonUtil.assertGreaterOrEqualThanReferenceVersion` and put the version to 1.10.

commons/src/main/java/com/powsybl/commons/json/JsonUtil.java
```diff
@@ -479,6 +479,19 @@ public static void assertLessThanReferenceVersion(String contextName, String ele
         }
     }
 
+    public static int compareVersions(String v1, String v2) {
+        String[] parts1 = v1.split("\\.");
```
olperr1, line 486: Since previously the comparison was made on Strings, which can accept any value, I think we should ensure that the method will not throw an exception when the version are not real version numbers.

So I think we could add the following securities:
- add a try/catch around the block checking the major versions
  - if a `NumberFormatException` is caught, return `v1.compareTo(v2)` (similar to the existing)
- add a try/catch around the block checking the minor versions
  - if a `NumberFormatException` is caught, return `minor1.compareTo(minor2)`

<br/>

Can you also add the following checks at the beginning of the method?
```java
Objects.requireNonNull(v1);
Objects.requireNonNull(v2);
```

commons/src/main/java/com/powsybl/commons/json/JsonUtil.java
```diff
@@ -479,6 +479,19 @@ public static void assertLessThanReferenceVersion(String contextName, String ele
         }
     }
 
+    public static int compareVersions(String v1, String v2) {
```
olperr1, line 482: Can you add unit tests on this method?

commons/src/main/java/com/powsybl/commons/json/JsonUtil.java
colineplqt: Also a question, could it happen that the version has more than 2 digits? Like 1.2.3

commons/src/main/java/com/powsybl/commons/json/JsonUtil.java
```diff
@@ -471,14 +471,27 @@ public static void assertGreaterOrEqualThanReferenceVersion(String contextName,
...
+        String[] parts1 = v1.split("\\.");
+        String[] parts2 = v2.split("\\.");
+        int major1 = Integer.parseInt(parts1[0]);
+        int major2 = Integer.parseInt(parts2[0]);
+        if (major1 != major2) {
+            return Integer.compare(major1, major2);
+        }
```
colineplqt:
```suggestion
    int majorCompare = Integer.compare(major1, major2);
    if (majorCompare != 0) {
        return majorCompare;
    }
```

commons/src/main/java/com/powsybl/commons/json/JsonUtil.java
olperr1: Thanks!

commons/src/main/java/com/powsybl/commons/json/JsonUtil.java
```diff
@@ -471,14 +471,46 @@ public static void assertGreaterOrEqualThanReferenceVersion(String contextName,
...
+            return v1.compareTo(v2);
+        }
+        try {
+            int minor1 = parts1.length > 1 ? Integer.parseInt(parts1[1]) : 0;
+            int minor2 = parts2.length > 1 ? Integer.parseInt(parts2[1]) : 0;
+            if (minor1 != minor2) {
+                return Integer.compare(minor1, minor2);
+            }
+        } catch (NumberFormatException e) {
+            return parts1[1].compareTo(parts2[1]);
```
olperr1: This line may throw an `OutOfBoundException` if one of the versions have no second part.

commons/src/main/java/com/powsybl/commons/json/JsonUtil.java
```diff
@@ -471,14 +471,46 @@ public static void assertGreaterOrEqualThanReferenceVersion(String contextName,
...
+        }
+        try {
+            int patch1 = parts1.length > 2 ? Integer.parseInt(parts1[2]) : 0;
+            int patch2 = parts2.length > 2 ? Integer.parseInt(parts2[2]) : 0;
+            return Integer.compare(patch1, patch2);
+        } catch (NumberFormatException e) {
+            return parts1[2].compareTo(parts2[2]);
```
olperr1: Same thing here with the third part.

commons/src/main/java/com/powsybl/commons/json/JsonUtil.java
cphili: This method can compare versions in such a format (by checking each segment in order: first major, then minor, then patch)

commons/src/test/java/com/powsybl/commons/json/JsonUtilTest.java
```diff
@@ -79,8 +80,27 @@ void testCompareVersions() {
...
+        assertTrue(JsonUtil.compareVersions("2.1.1", "2.") > 0);
+        assertTrue(JsonUtil.compareVersions(".1", "2.1") < 0);
+        assertTrue(JsonUtil.compareVersions("2.1", ".1") > 0);
+        assertTrue(JsonUtil.compareVersions(".", "1.1") < 0);
+        assertTrue(JsonUtil.compareVersions("1.1", ".") > 0);
         assertEquals(0, JsonUtil.compareVersions("1.10.1", "1.10.1"));
+        assertTrue(JsonUtil.compareVersions("1.10.", "1.10.1") < 0);
+        assertTrue(JsonUtil.compareVersions("1.10.1", "1.10") > 0);
         assertTrue(JsonUtil.compareVersions("1.10.1", "1.10.2") < 0);
         assertTrue(JsonUtil.compareVersions("A.2", "B.2") < 0);
```
olperr1, line 88: This list can be shorter and have a better coverage.
I suggest the following instead:

```suggestion
        assertTrue(JsonUtil.compareVersions("1.9", "1.10") < 0);
        assertTrue(JsonUtil.compareVersions("2.0", "1.9") > 0);

        assertTrue(JsonUtil.compareVersions("1.10", "1.10.1") < 0);
        assertTrue(JsonUtil.compareVersions("1.10.2", "1.10") > 0);

        assertEquals(0, JsonUtil.compareVersions("1.10.1", "1.10.1"));

        assertTrue(JsonUtil.compareVersions("A.2", "B.2") < 0);
        assertTrue(JsonUtil.compareVersions("B.2", "A.2") > 0);

        assertTrue(JsonUtil.compareVersions("3.2", "3.C") < 0);
        assertTrue(JsonUtil.compareVersions("3.C.2", "3.C") > 0);
```

PR #3587, opened 2025-10-06
Cgmes import: handle duplicate region name

rcourtier:

No.

Bug fix.

When CGMES input data have duplicate GeographicalRegion.rdf:Id but with different name, the CGMES import fails and throws an exception: "Substation should be null". This is due to the substations SPARQL query finding 2 Substation instead of one (1 PropertyBag with the first region name, and another PropertyBag with the second region name).

Only 1 Substation is returned by the SPARQL query in case of duplicate GeographicalRegion with different name. The lowest region name is kept.

- [X] Yes
- [ ] No

The method microGridBaseCaseDuplicateRegion() of CgmesConformity1ModifiedCatalog has been removed. It was used for unit testing, but it relied on big a CGMES file ; this slowed the tests execution. If you still need to use it you can retrieve the methods at [this URL](https://github.com/powsybl/powsybl-core/blob/v6.7.0/cgmes/cgmes-conformity/src/main/java/com/powsybl/cgmes/conformity/CgmesConformity1ModifiedCatalog.java) and the corresponding CGMES files in [this directory](https://github.com/powsybl/powsybl-core/tree/v6.8.1/cgmes/cgmes-conformity/src/main/resources/conformity-modified/cas-1.1.3-data-4.0.3).

**Other information**:

Original PR to fix duplicate in sub-region ids and region ids: https://github.com/powsybl/powsybl-core/pull/2507. The same logic is now applied to region name.


cgmes/cgmes-model/src/main/resources/CIM16.sparql
```diff
@@ -195,14 +195,15 @@ WHERE {
...
+                    ?Region
+                        a cim:GeographicalRegion ;
+                        cim:IdentifiedObject.name ?regionName .
                 }
                 GROUP BY ?SubRegion
             }
-            ?Region cim:IdentifiedObject.name ?regionName .
         }
     }
     # ... But we do not want results where the regions are missing
```
marqueslanauja: Maybe we can improve the query by removing unnecessary OPTIONAL blocks. Since we’re using FILTER, they’re not truly optional. Here's my proposal:

```
SELECT ?Substation ?name
       ?SubRegion (MIN(?subRegionName) AS ?subRegionName)
       ?Region (MIN(?regionName) AS ?regionName)
WHERE {
  ?Substation a cim:Substation ;
              cim:IdentifiedObject.name ?name .

  {
    SELECT ?Substation (MIN(?sr) AS ?SubRegion)
    WHERE {
      ?Substation cim:Substation.Region ?sr .
      ?sr a cim:SubGeographicalRegion .
    }
    GROUP BY ?Substation
  }
  ?SubRegion cim:IdentifiedObject.name ?subRegionName .

  {
    SELECT ?SubRegion (MIN(?r) AS ?Region)
    WHERE {
      ?SubRegion cim:SubGeographicalRegion.Region ?r .
      ?r a cim:GeographicalRegion .
    }
    GROUP BY ?SubRegion
  }
  ?Region cim:IdentifiedObject.name ?regionName .

}
GROUP BY ?Substation ?name ?SubRegion ?Region
ORDER BY ?Substation

```

I've sorted the query results by substation because the order of substations varies depending on the resource file order, which causes some unit tests to fail

rcourtier: Change applied!

marqueslanauja (approved):

PR #3601, opened 2025-10-09
fix doc default value for writeSlackBus and readSlackBus

vidaldid-rte:

No

The documented default values for `readSlackBus` and `writeSlackBus` are incorrect.
It seems they were not updated in these PRs: #2922 and #1908

Incorrect doc

Correct doc default value regarding these parameters.

- [ ] Yes
- [X] No


docs/simulation/loadflow/configuration.md
olperr1: The PR description talks about the `read...` parameter, but the PR change the `write...` one...
... But both must be changed: the doc states that their default values are `false`; in the code, they are in fact `true`:
https://github.com/powsybl/powsybl-core/blob/main/loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowParameters.java#L96

vidaldid-rte: Good catch. I updated the PR title and content.

olperr1: Thanks!

PR #3603, opened 2025-10-10
Fix the acceptable duration on the last temporary limits' violations

olperr1:

No

Bug fix

When the limits are defined like in [the 1st example of the limits' documentation](https://powsybl.readthedocs.io/projects/powsybl-core/en/stable/grid_model/additional.html#first-example) and the current value is over the last temporary limit, the acceptable duration of the `LimitViolation` is wrong: it indicates the last temporary limit's duration (in the example 600s) instead of a `0` duration.

When the last temporary limit is violated, the acceptable duration is `0`.

- [ ] Yes
- [X] No

**Other information**:

The violation was not detected before #3572, but the `temporaryLimit` used to construct the `Overload` object in this case is the last temporary limit, where it should be:

> The temporary limit under which the current is.
([source](https://javadoc.io/doc/com.powsybl/powsybl-core/7.0.0-RC1/com/powsybl/iidm/network/Overload.html))

A fictitious limit with a "0s" acceptable duration is used instead.

The cases covered by `AbstractLimitViolationDetectionTest.detectCurrentLimit()` were build and tested with powsybl-open-loadflow. Therefore, with this fix the results of powsybl-core's `LimitViolationDetector` are consistent with the open-loadflow's results.

### DIFF
--- a/security-analysis/security-analysis-api/src/test/java/com/powsybl/security/detectors/AbstractLimitViolationDetectionTest.java
+++ b/security-analysis/security-analysis-api/src/test/java/com/powsybl/security/detectors/AbstractLimitViolationDetectionTest.java
@@ -41,7 +41,6 @@ public abstract class AbstractLimitViolationDetectionTest {
     private static Network networkWithFixedCurrentLimits;
     private static Network networkWithFixedCurrentLimitsOnDanglingLines;
     private static Network networkWithVoltageAngleLimit;
-    protected static Network networkMultiSets;
     protected List<LimitViolation> violationsCollector;
 
     @BeforeAll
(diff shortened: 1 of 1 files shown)


security-analysis/security-analysis-api/src/test/java/com/powsybl/security/detectors/AbstractLimitViolationDetectionTest.java
```diff
@@ -49,6 +54,50 @@ static void setUpClass() {
         networkWithCurrentLimitsOn3WT = ThreeWindingsTransformerNetworkFactory.createWithCurrentLimits();
         networkWithApparentLimitsOn3WT = ThreeWindingsTransformerNetworkFactory.createWithApparentPowerLimits();
         networkWithActiveLimitsOn3WT = ThreeWindingsTransformerNetworkFactory.createWithActivePowerLimits();
+
+        networkMultiSets = EurostagTutorialExample1Factory.create();
```
rolnico: Since this network is used only in one test, should you not define it only in the test?

PR #3615, opened 2025-10-20
Allow creating AC/DC Converters without a PCC Terminal

jeandemanged:

No

In between bugfix and feature / change made in the context of the beta status of the DC detailed model.

1. AC/DC converters with 2 AC Terminals can only be created with a PCC Terminal defined as a branch terminal. The converter terminal cannot be used
2. when the branch used as PCC Terminal is removed from the model, the PCC Terminal becomes `null`

The point 1. above is an excessive limitation, which is especially annoying when importing models since typically importing a model is a two step process: first importing equipment and second importing regulations. By imposing the PCC Terminal to be defined at equipment creation requires complex model import processing. However, the expected behavior for an AC/DC Converter with two AC Terminals and with one of its AC Terminal designated as PCC Terminal must be clarified.

The point 2. is a bug because it leaves a model with a converter with an undefined PCC Terminal.

- AC/DC Converters with 2 AC Terminals do not require anymore a PCC Terminal to be defined at creation time. PCC Terminal in this case defaults to the first AC Terminal. Expected processing by simulators is clarified in documentation.
- If an equipment used as PCC Terminal is removed, the AC/DC converter PCC Terminal is automatically changed to be converter's first AC Terminal

- [x] Yes
- [ ] No

Custom IIDM implementations should adapt to the updated AC/DC Converters validation rules tested in IIDM TCK:
- `com.powsybl.iidm.network.tck.AbstractAcDcConverterTest#testCreationError`
- `com.powsybl.iidm.network.tck.AbstractAcDcConverterTest#testTwoAcTerminals`

**Other information**:

Needed for both CGMES import and IIDM import.

### DIFF
--- a/iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractAcDcConverterTest.java
+++ b/iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractAcDcConverterTest.java
@@ -801,16 +801,16 @@ public abstract class AbstractAcDcConverterTest {
                 .setReactivePowerSetpoint(0.0)
                 .add();
         assertTrue(acDcConverterA.getTerminal2().isPresent());
-        assertSame(acDcConverterA.getPccTerminal(), acDcConverterA.getTerminal1()); // defaults to AC Terminal 1
+        assertSame(acDcConverterA.getTerminal1(), acDcConverterA.getPccTerminal()); // defaults to AC Terminal 1
         assertTrue(acDcConverterA.getTerminal(TerminalNumber.TWO).isPresent());
 
         // change PCC Terminal to line terminal
         acDcConverterA.setPccTerminal(lineax.getTerminal1());
-        assertSame(acDcConverterA.getPccTerminal(), lineax.getTerminal1());
+        assertSame(lineax.getTerminal1(), acDcConverterA.getPccTerminal());
 
         // remove the line, check PCC Terminal automatically fixed to be the converter terminal 1
         lineax.remove();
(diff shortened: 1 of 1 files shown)


rcourtier (approved):

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractAcDcConverterTest.java
```diff
@@ -789,6 +786,33 @@ public void testSingleAcTerminal() {
...
+                .setDcNode1(dcNode1a.getId())
+                .setDcNode2(dcNode2a.getId())
+                .setControlMode(AcDcConverter.ControlMode.P_PCC)
+                .setTargetP(100.)
+                .setTargetVdc(500.)
+                .setVoltageRegulatorOn(false)
+                .setReactivePowerSetpoint(0.0)
+                .add();
+        assertTrue(acDcConverterA.getTerminal2().isPresent());
+        assertSame(acDcConverterA.getPccTerminal(), acDcConverterA.getTerminal1()); // defaults to AC Terminal 1
```
olperr1: Can you change the order of the parameters in all the `assertSame` calls of this unit test? The second parameter is the value we want to check and the first one the expected value. Having the PCC terminal as the second parameter will thus be more logical.

olperr1 (approved):

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractAcDcConverterTest.java
olperr1: Thanks!

PR #3625, opened 2025-10-23
Security analysis detailed result handler

obrix:

A proposal for a detailed result handler interface in the security analysis API

- [ ] Yes
- [ ] No

**Other information**:
A PR is necessary in OLF too.

### DIFF
--- a/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/AbstractSecurityAnalysisRunParameters.java
+++ b/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/AbstractSecurityAnalysisRunParameters.java
@@ -13,6 +13,7 @@ import com.powsybl.computation.ComputationManager;
 import com.powsybl.computation.local.LocalComputationManager;
 import com.powsybl.security.interceptors.SecurityAnalysisInterceptor;
 import com.powsybl.security.monitor.StateMonitor;
+import com.powsybl.security.results.SecurityAnalysisResultHandler;
 import com.powsybl.security.strategy.OperatorStrategy;
 
 import java.util.ArrayList;
@@ -34,6 +35,7 @@ public abstract class AbstractSecurityAnalysisRunParameters<T extends AbstractSe
     private List<OperatorStrategy> operatorStrategies = new ArrayList<>();
     private List<Action> actions = new ArrayList<>();
     private List<StateMonitor> monitors = new ArrayList<>();
+    private List<SecurityAnalysisResultHandler> handlers = new ArrayList<>();
     private ReportNode reportNode = ReportNode.NO_OP;
 
     /**
@@ -132,6 +134,15 @@ public abstract class AbstractSecurityAnalysisRunParameters<T extends AbstractSe
     }
 
     /**
+     * Sets the list of handlers referenced in {@link SecurityAnalysisResultHandler}
+     */
+    public T setResultHandlers(List<SecurityAnalysisResultHandler> handlers) {
+        Objects.requireNonNull(handlers, "Handler list should not be null");
+        this.handlers = handlers;
+        return self();
+    }
+
+    /**
      * Sets the reportNode used for functional logs, see {@link ReportNode}
      */
     public T setReportNode(ReportNode reportNode) {
@@ -161,6 +172,12 @@ public abstract class AbstractSecurityAnalysisRunParameters<T extends AbstractSe
     public T addAction(Action action) {
         Objects.requireNonNull(action, "Action should not be null");
         actions.add(action);
+        return self();
+    }
+
+    public T addResultHandler(SecurityAnalysisResultHandler handler) {
+        Objects.requireNonNull(handlers, "Handler should not be null");
+        handlers.add(handler);
         return self();
     }
 
--- a/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/results/SecurityAnalysisResultHandler.java
+++ b/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/results/SecurityAnalysisResultHandler.java
@@ -1,0 +1,14 @@
+package com.powsybl.security.results;
+
+import com.powsybl.contingency.Contingency;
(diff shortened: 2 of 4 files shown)


obrix:
@geofjamg @vidaldid-rte 

In the proposed approach a new parameter for the security analysis is available : a list of SecurityAnalysisResultHandlers. Theses handlers will be notified in the security analysis implementation when data is available.

A basic example implementation is provided, InMemoryResultHandler, that just store the data in memory as they come.

The user can provide its own implementation. The contingency and operator strategy is provided in the API to let the user deduce the context of the data provided (see in memory implementation).

What is missing here to a least cover our use case : a GeneratorResult, providing target and real flow for each state. But to add this the StateMonitor API should probably be updated to be able to monitor generators.

Maybe we could add some more notification API in the handler interface such as "base case done", "contingency done", "operator strategy done". It could be handy when doing large security analysis and some progression notification system is available in the user architecture.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/results/SecurityAnalysisResultHandler.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+public interface SecurityAnalysisResultHandler {
+
+    void writeBranchResult(Contingency contingency, OperatorStrategy operatorStrategy, BranchResult branchResult);
+
+    void writeThreeWindingsTransformerResult(Contingency contingency, OperatorStrategy operatorStrategy, ThreeWindingsTransformerResult branchResult);
+
+    void writeBusResult(Contingency contingency, OperatorStrategy operatorStrategy, BusResult busResult);
+
+}
```
vidaldid-rte, line 24: To align with the suggestion in https://github.com/powsybl/powsybl-core/issues/3626 (that extends the sensitivity writer for multi-component, better failure diagnosis and interrupt) you could add

// numCC and numCs are used for contingencies that have equipments in multiple components
void writeContingencyStatus contingency(int contingencyIndex, LoadFlowResult.ComponentResult.Status, String statusText, int numCC, int numCs);

void writeSynchronousComponentStatus(int numCC, int numCS, LoadFlowResult.ComponentResult.Status, String statusText,);

// Called at the end of the computation if the computation has not been interrupted
void computationComplete();

vidaldid-rte: To turn this in to a way to get result in streaming, you could also add methods for listening to violations.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/AbstractSecurityAnalysisRunParameters.java
```diff
@@ -131,6 +133,15 @@ public T setActions(List<Action> actions) {
         return self();
     }
 
+    /**
+     * Sets the list of handlers referenced in {@link SecurityAnalysisResultHandler}
+     */
+    public T setResultHandlers(List<SecurityAnalysisResultHandler> handlers) {
```
vidaldid-rte, line 139: Why a list and not only one ?
  If there was only one, we coud mimic the sensi API: no result handler provided means you return an in memory result.
  Handler provided means you return void and send the data to the writer.
  This view implies that the writer supports also the violation results, not only the monitored objects.
  
  Then if it is useful to have serveral handlers, we would provided the implementation of a SecurityAnalysisResultMultiplexer that would dispatch the calls to a list of handlers.

obrix: Yes I agree, if some kind of multiplexing is needed the user has everything available to do it himself anyway. So ok for a single handler API.

obrix: An handler provided would mean we do not fill the result objects then ? NetworkResult in PreContingencyResult / PostContingencyResult would remain empty ?

vidaldid-rte: Here is the Sensitivity API. We could use the same idea for security + writer

No writer -> returns a Result. Internally it creates a writer that creates the result object. The provider only knows the writer.
        public SensitivityAnalysisResult run(Network network,
                                             String workingVariantId,
                                             List<SensitivityFactor> factors,
                                             SensitivityAnalysisRunParameters runParameters) 

A writer -> returns void. The user code handles that data as needed, without memory overhead, for example if the result is directly streamed into a database or a python dataframe.
        public void run(Network network,
                        String workingVariantId,
                        SensitivityFactorReader factorReader,
                        SensitivityResultWriter resultWriter,
                        SensitivityAnalysisRunParameters runParameters) {
            runAsync(network, workingVariantId, factorReader, resultWriter, runParameters).join();
        }

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/results/SecurityAnalysisResultHandler.java
geofjamg: @obrix yes, here you only streamed only a part of the SecurityAnalysisResult API. We should fully convert it even for violations.

jonenst:
hi @obrix , I'm interested in this to implement a progress bar in GridSuite, so I want to help move this forward. Can you make a quick recap of the current status, and what was the plan to move forward ? Thanks !

obrix:
Hello @jonenst 

At the moment API for base case / contingency / operator strategy status notification is missing. I think this is what you need here to be notified in the handler when the computation is moving forward.

Notification API for violations is also missing.

I will not have time to continue working on that before the beginning of next year. If you want to continue iterating to add what is missing, no problem on my side !

So-Fras:
See https://github.com/gautierbureau/powsybl-core/pull/35 to check if it fully covers the span of this PR.

PR #3627, opened 2025-10-23
Replace univocity parsers with FastCSV for CSV handling except for PSSE

rolnico:

No

Change of dependency

PowSyBl uses univocity-parsers for handling CSV files (parsing and writing), but [this lib is abandoned since 2021](https://github.com/uniVocity/univocity-parsers/issues/534).

PowSyBl now uses [FastCSV](https://fastcsv.org/) for handling CSV files.

- [ ] Yes
- [x] No

**Other information**:

PSS-E handling with be managed in a specific PR


rolnico:
I did some tests to compare the performances of this version vs main. Each tested operation was performed 50 times consecutively, and the results are based on the last 30 draws.

The operations tested where:

- For the Anonymizer, I tested `writeCsv` and `readCsv` for 5000000 values anonymized.
- For the Time Series, I tested the time series generation, `TimeSeriesTable.writeCsv`, and `TimeSeries.parseCsv`. The dataset consisted of 1000 double time series of 3000 time steps in 5 versions, which represents a CSV file of 290 Mo once exported.
- For the IEEE CDF, I tested the `IeeeCdfNetworkFactory.create33` and `IeeeCdfNetworkFactory.create69` methods.

The results are as follows:

Operation | main (ms) | FastCSV (ms) | Difference (ms) | Difference (%)
-- | -- | -- | -- | --
Anonymizer - Write | 311,393 | 181,985 | -129,408 | -41,56 %
Anonymizer - Read | 2006,383 | 1971,176 | -35,207 | -1,75 %
Time Series - Generation | 198,604 | 181,077 | -17,527 | -8,82 %
Time Series - Write | 1818,053 | 1800,404 | -17,649 | -0,97 %
Time Series - Read | 2614,167 | 2729,676 | +115,509 | +4,42 %
IEEE CDF - 33 | 7,210 | 0,849 | -6,361 | -88,22 %
IEEE CDF - 69 | 5,048 | 1,020 | -4,028 | -79,80 %

The results here are very encouraging: it seems that the change of dependency can optimize the performances in some cases.
However, since there is not much change in the Time series performances, it would be interesting so see if the time is spent in the library or in the code around it.

commons/src/main/java/com/powsybl/commons/util/StringAnonymizer.java
```diff
@@ -89,18 +89,14 @@ public void writeCsv(BufferedWriter writer) {
...
-            for (Map.Entry<String, String> e : mapping.entrySet()) {
-                nextLine[0] = e.getKey();
-                nextLine[1] = e.getValue();
-                csvWriter.writeRow(nextLine);
-            }
-        } finally {
-            csvWriter.close();
+        try (CsvWriter csvWriter1 = CsvWriter.builder()
+            .fieldSeparator(separator)
+            .quoteCharacter('"')
```
NathanDissoubray, line 94: According to the javadoc of FastCSV, " is already the default
https://javadoc.io/doc/de.siegmar/fastcsv/latest/de.siegmar.fastcsv/de/siegmar/fastcsv/writer/CsvWriter.CsvWriterBuilder.html#quoteCharacter(char)
Is there a need to specify this ?

commons/src/main/java/com/powsybl/commons/util/StringAnonymizer.java
```diff
@@ -66,21 +68,19 @@ public void readCsv(BufferedReader reader) {
...
-        setFormat(settings.getFormat(), separator);
-        CsvParser csvParser = new CsvParser(settings);
-        for (String[] nextLine : csvParser.iterate(reader)) {
-            if (nextLine.length != 2) {
-                throw new PowsyblException("Invalid line '" + Arrays.toString(nextLine) + "'");
-            }
-            mapping.put(nextLine[0], nextLine[1]);
+        try (CsvReader<CsvRecord> csvReader = CsvReader.builder()
+            .fieldSeparator(separator)
+            .quoteCharacter('"')
```
NathanDissoubray, line 74: same question regarding the default quote character https://javadoc.io/doc/de.siegmar/fastcsv/latest/de.siegmar.fastcsv/de/siegmar/fastcsv/reader/CsvReader.CsvReaderBuilder.html#quoteCharacter(char)

iidm/iidm-test/src/main/java/com/powsybl/iidm/network/test/EuropeanLvTestFeederFactory.java
```diff
@@ -91,145 +112,84 @@ private static void createSource(Transformer transformer, Network network) {
...
+            rec.getField("Conn_pri"),
+            rec.getField("Conn_sec"),
+            Double.parseDouble(rec.getField("%XHL")),
+            Double.parseDouble(rec.getField("% resistance"))
+        );
     }
 
+    // Casting is safe here
+    @SuppressWarnings("unchecked")
     private static <T> List<T> parseCsv(String resourceName, Class<T> clazz) {
```
NathanDissoubray, line 182: because the construction of MAPPERS guarantees that if it doesn't exist, we get a null ?

time-series/time-series-api/src/main/java/com/powsybl/timeseries/TimeSeries.java
```diff
@@ -580,16 +587,18 @@ static Map<Integer, List<TimeSeries>> parseCsv(BufferedReader reader, TimeSeries
...
-        settings.getFormat().setQuoteEscape('"');
-        settings.getFormat().setLineSeparator(System.lineSeparator());
-        settings.setMaxColumns(timeSeriesCsvConfig.getMaxColumns());
-        CsvParser csvParser = new CsvParser(settings);
-        ResultIterator<String[], ParsingContext> iterator = csvParser.iterate(reader).iterator();
-        CsvParsingContext context = readCsvHeader(iterator, timeSeriesCsvConfig);
-        readCsvValues(iterator, context, timeSeriesPerVersion, reportNode);
+        try (CsvReader<CsvRecord> csvReader = CsvReader.builder()
+            .fieldSeparator(timeSeriesCsvConfig.separator())
+            .quoteCharacter('"')
```
NathanDissoubray, line 596: same question here regarding the default

commons/src/main/java/com/powsybl/commons/util/StringAnonymizer.java
rolnico: I prefer to have it explicitely specified, as it was before.

commons/src/main/java/com/powsybl/commons/util/StringAnonymizer.java
rolnico: same answer

time-series/time-series-api/src/main/java/com/powsybl/timeseries/TimeSeries.java
rolnico: same answer

iidm/iidm-test/src/main/java/com/powsybl/iidm/network/test/EuropeanLvTestFeederFactory.java
rolnico: yes, and we know exactly what to expect from a call to MAPPERS since it's static

olperr1:
Since univocity is no more imported, it is also no more available for downstream projects.
Adaptations are needed for powsybl-dynawo: powsybl/powsybl-dynawo#663

ieee-cdf/ieee-cdf-converter/src/main/java/com/powsybl/ieeecdf/converter/IeeeCdfNetworkFactory.java
```diff
@@ -183,13 +191,12 @@ private static Network createFromCsv(String name, NetworkFactory networkFactory,
...
-        settings.getFormat().setDelimiter(" ");
-        parseBuses(model, settings, name + "-bus.csv", baseKv);
-        parseLines(model, settings, name + "-line.csv");
+        CsvReader.CsvReaderBuilder builder = CsvReader.builder()
+            .fieldSeparator(" ");
+        parseBuses(model, builder, name + "-bus.csv", baseKv);
+        parseLines(model, builder, name + "-line.csv");
         if (meshed) {
-            parseLines(model, settings, name + "-mesh.csv");
+            parseLines(model, builder, name + "-mesh.csv");
```
olperr1: Instead of reusing the same builder, it would be safer to create a new method to call in `parseBuses` and `parseLines`. This new method can either create a builder or directly the `CsvReader`.
In the current state, `parseBuses` or `parseLines` can set options in the builder, and they will apply to the subsequent usages of the builder.

time-series/time-series-api/src/main/java/com/powsybl/timeseries/TimeSeries.java
```diff
@@ -580,16 +587,18 @@ static Map<Integer, List<TimeSeries>> parseCsv(BufferedReader reader, TimeSeries
 
         Map<Integer, List<TimeSeries>> timeSeriesPerVersion = new HashMap<>();
 
-        CsvParserSettings settings = new CsvParserSettings();
-        settings.getFormat().setDelimiter(timeSeriesCsvConfig.separator());
-        settings.getFormat().setQuoteEscape('"');
-        settings.getFormat().setLineSeparator(System.lineSeparator());
-        settings.setMaxColumns(timeSeriesCsvConfig.getMaxColumns());
```
olperr1, line 587: This option is ignored now...

olperr1 (changes requested):

olperr1 (approved):

PR #3630, opened 2025-10-24
CGMES: fix SynchronousMachine kind (EQ) and operatingMode (SSH) and use isCondenser

colineplqt:

No

Bug fix

1. At CGMES import, the flag `isCondenser` is set to `true` only if the `SynchronousMachineKind` is `condenser`.
2. We save the `SynchronousMachine.OperatingMode` as a property so it can be reexported.
3. At EQ export, the reactive limits are used to determine the `SynchronousMachineKind` and the flag `isCondenser` is ignored.
- if minP is strictly positive, then the kind is `generator`
- if maxP is strictly negative, then the kind is `motor`
- if minP and maxP are zero, then the kind is `condenser`
- if minP is zero, then the kind is `generatorOrCondenser`
- if maxP is zero, then the kind is `motorOrCondenser`
- otherwise, we use the original kind or return `generatorOrCondenserOrMotor`
4. At SSH export, only the value of targetP is checked:
- if targetP is strictly negative, the operating mode is `motor`
- if targetP is strictly positive, the operating mode is `generator`
- otherwise, the operating mode is `condenser`

1. At CGMES import, the flag `isCondenser` of a `Generator` is set to `true` if its `SynchronousMachineKind` contains the word `condenser`. Indeed, the documentation says: "True if the generator **may** behave as a condenser".
2. At CGMES, the  `SynchronousMachine.OperatingMode` is not saved as it will be in any case recomputed.
3. At EQ export, the `SynchronousMachineKind` is calculated using the reactive limits of the synchronous machine (min/max or the reactive capabibility curve if it is a generator to obtain minP and maxP). Then, according to the value of minP and maxP and the flag `isCondenser` (always true if it is a battery), the kind is obtained:
- if `isCondenser` is `true`: 
if minP and maxP are positive, then the kind is `generatorOrCondenser`
if minP and maxP are negative, then the kind is `motorOrCondenser`
if minP and maxP are zero, then the kind is `condenser`
otherwise, the kind is `generatorOrCondenserOrMotor`
- if `isCondenser` is `false`:
if minP is positive, then the kind is `generator`
if minP is negative, then the kind is `motor`
otherwise, the kind is `generatorOrMotor`
6. At SSH export, the `SynchronousMachineOperatingMode` is calculated using the targetP of the generator:
- if targetP is positive, then the operating mode is `generator`
- if targetP is negative, then the operating mode is `motor`
- if targetP is zero, then it depends if the generator is regulating or not. If it is regulating, then the operating mode is `condenser`, otherwise, it is `generator`, `motor` or `condenser` depending on what the EQ allows.

**Other information**:

Originally, the issue comes from the fact that if a generator is disconnected, then its OperatingMode would be `condenser` because its `targetP` is zero but its kind in EQ does not have to allow this value. This violates the QoCDC rule SMOperatingModeConsistency. As a consequence, we thought of an alternative way to decide the kind and operating mode of a synchronous machine that also includes its ability to regulate.


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
```diff
@@ -358,20 +392,48 @@ private static void writeBatteries(Network network, String cimNamespace, XMLStre
     }
 
     private static <I extends ReactiveLimitsHolder & Injection<I>> String obtainOperatingMode(I i, double minP, double maxP, double targetP) {
-        String kind = i.getProperty(Conversion.PROPERTY_CGMES_SYNCHRONOUS_MACHINE_TYPE);
         String operatingMode = i.getProperty(Conversion.PROPERTY_CGMES_SYNCHRONOUS_MACHINE_OPERATING_MODE);
-        String calculatedKind = CgmesExportUtil.obtainCalculatedSynchronousMachineKind(minP, maxP, CgmesExportUtil.obtainCurve(i), kind);
-        String calculatedOperatingMode = obtainOperatingMode(targetP);
+        String calculatedKind = obtainCalculatedSynchronousMachineKind(minP, maxP, obtainCurve(i), i instanceof Generator gen && gen.isCondenser());
+        String calculatedOperatingMode = obtainOperatingMode(targetP, i, calculatedKind);
         return operatingMode != null && calculatedKind.contains(operatingMode) ? operatingMode : calculatedOperatingMode;
```
rcourtier: Shouldn't we lower case calculatedKind before checking if it contains the operatingMode?
Like if calculatedKind is `generatorOrMotor` and operatingMode is `motor`, the condition will be false whereas we want it to be true.

```suggestion
        return operatingMode != null && calculatedKind.toLowerCase().contains(operatingMode) ? operatingMode : calculatedOperatingMode;
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
```diff
@@ -358,20 +392,48 @@ private static void writeBatteries(Network network, String cimNamespace, XMLStre
...
+            return OPERATING_MODE_MOTOR;
         } else if (targetP > 0) {
-            return "generator";
+            return OPERATING_MODE_GENERATOR;
         } else {
-            return "condenser";
+            if (isOperatingAsACondenser(injection)) {
+                return OPERATING_MODE_CONDENSER;
+            } else {
+                if (calculatedKind.contains(OPERATING_MODE_GENERATOR)) {
```
rcourtier:
```suggestion
                if (calculatedKind.toLowerCase().contains(OPERATING_MODE_GENERATOR)) {
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
```diff
@@ -358,20 +392,48 @@ private static void writeBatteries(Network network, String cimNamespace, XMLStre
...
+                    return OPERATING_MODE_GENERATOR;
+                } else if (calculatedKind.contains(OPERATING_MODE_MOTOR)) {
```
rcourtier:
```suggestion
                } else if (calculatedKind.toLowerCase().contains(OPERATING_MODE_MOTOR)) {
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
rcourtier: In the end the `operatingMode` is taking precedence over the `calculatedOperatingMode`. Are we sure that's what we want? Especially if the user has updated the targetP of the Generators, the initial modes are exported rather than the new calculated modes. I would have given priority to the calculated ones.
What do you think?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
```diff
@@ -358,20 +392,48 @@ private static void writeBatteries(Network network, String cimNamespace, XMLStre
...
+                    double targetV = generator.getTargetV();
+                    return !Double.isNaN(targetV) && targetV != 0;
+                } else {
+                    double targetQ = generator.getTargetQ();
+                    return !Double.isNaN(targetQ) && targetQ != 0;
+                }
+            }
+            case Battery battery -> {
+                double targetQ = battery.getTargetQ();
+                return !Double.isNaN(targetQ) && targetQ != 0;
```
rcourtier: You could also check if the battery has a `VoltageRegulation` extension with voltageRegulatorOn and a non-Nan non-null targetV.

docs/grid_exchange_formats/cgmes/export.md
```diff
@@ -245,6 +245,29 @@ generator has the extension [`RemoteReactivePowerControl`](../../grid_model/exte
...
+The `SynchronousMachine.SynchronousMachineKind` is exported in the EQ profile depending on the [reactive limits](../../grid_model/additional.md#reactive-limits) of the 
+generator or battery and its capacity to behave like a condenser (a battery cannot behave like a condenser so the flag `isCondenser` is considered as `false`):
+- if the flag `isCondenser` is `true`: 
+  - if the minimum and the maximum active power limit are positive, then the generator or battery will be exported as `generatorOrCondenser`,
+  - if the minimum and the maximum active power limit are negative, then the generator or battery will be exported as `motorOrCondenser`,
+  - if the minimum and the maximum active power limit are both equal to zero, then the generator or battery will be exported as `condenser`,
+  - otherwise, the generator or battery will be exported as `generatorOrCondenserOrMotor`.
+- if the flag `isCondenser` is `false`:
+  - if the minimum active power limit is positive, then the generator or battery will be exported as `generator`,
+  - if the minimum active power limit is negative, then the generator or battery will be exported as `motor`,
```
rcourtier:
```suggestion
  - if the maximum active power limit is negative, then the generator or battery will be exported as `motor`,
```

docs/grid_exchange_formats/cgmes/export.md
```diff
@@ -245,6 +245,29 @@ generator has the extension [`RemoteReactivePowerControl`](../../grid_model/exte
...
+  - otherwise, the generator will be exported as `generatorOrMotor`.
+
+The `SynchronousMachine.SynchronousMachineOperatingMode` is exported in the SSH profile depending on the target active 
+power of the generator or battery and on fact that it is regulating or not:
+- if the target active power is positive, then the generator or battery will be exported as `generator`,
+- if the target active power is negative, then the generator or battery will be exported as `motor`,
+- if the target active power is zero and the generator or battery is regulating, then the operating mode will be `condenser`.
+- otherwise, the generator or battery will be exported as `generator` if is allowed by its `SynchronousMachine.SynchronousMachineKind`,
+otherwise `motor` and otherwise `condenser`. 
+To know if the generator is behaving as a condenser, its `targetV`, `targetQ` and `voltageRegulatorOn` attributes are used.
```
rcourtier:
```suggestion
To know if the generator or battery is behaving as a condenser, its `targetV`, `targetQ` and `voltageRegulatorOn` attributes are used.
```

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/GeneratorConversionTest.java
```diff
@@ -60,4 +61,137 @@ void reactiveLimitsExportTest() throws IOException {
...
+        assertTrue(network.getGenerator("sm_condenser").isCondenser());
+    }
+
+    @Test
+    void synchronousMachineOperatingModeAndKindConversion() throws IOException {
+        Network network = createNetwork();
+        Generator generator1 = network.getGenerator("GEN");
+        Generator generator2 = network.getGenerator("GEN2");
+        Generator generator3 = network.getGenerator("GEN3");
+        Battery battery = network.getBattery("BAT");
```
rcourtier: You don't need to retrieve these connectables. You can directly use their ids in the `getElement` methods below.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/GeneratorConversionTest.java
```diff
@@ -60,4 +61,137 @@ void reactiveLimitsExportTest() throws IOException {
...
+            .add();
+        generator2.newMinMaxReactiveLimits().setMinQ(-50.0).setMaxQ(50.0).add();
+        voltageLevel1.getNodeBreakerView().newInternalConnection().setNode1(0).setNode2(2).add();
+
+        // Will be exported as motor and operating as a motor
+        Generator generator3 = voltageLevel1.newGenerator()
+            .setId("GEN3")
+            .setNode(3)
+            .setMinP(0.0)
+            .setMaxP(100.0)
```
rcourtier: The generator minP/maxP are not in phase with the reactiveCapabilityCurve minP/maxP. Is that done on purpose to show that the curve limits take precedence over the generator limits?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
```diff
@@ -358,20 +392,48 @@ private static void writeBatteries(Network network, String cimNamespace, XMLStre
...
 
     private static <I extends ReactiveLimitsHolder & Injection<I>> String obtainOperatingMode(I i, double minP, double maxP, double targetP) {
-        String kind = i.getProperty(Conversion.PROPERTY_CGMES_SYNCHRONOUS_MACHINE_TYPE);
         String operatingMode = i.getProperty(Conversion.PROPERTY_CGMES_SYNCHRONOUS_MACHINE_OPERATING_MODE);
-        String calculatedKind = CgmesExportUtil.obtainCalculatedSynchronousMachineKind(minP, maxP, CgmesExportUtil.obtainCurve(i), kind);
-        String calculatedOperatingMode = obtainOperatingMode(targetP);
+        String calculatedKind = obtainCalculatedSynchronousMachineKind(minP, maxP, obtainCurve(i), i instanceof Generator gen && gen.isCondenser());
+        String calculatedOperatingMode = obtainOperatingMode(targetP, i, calculatedKind);
         return operatingMode != null && calculatedKind.contains(operatingMode) ? operatingMode : calculatedOperatingMode;
     }
```
marqueslanauja, line 369: We think it doesn’t make sense to record the property         `Conversion.PROPERTY_CGMES_SYNCHRONOUS_MACHINE_OPERATING_MODE` during the update process. It is also recorded during the conversion process, but that situation will never occur because this data is not available in the EQ file. The SSH export must be based on the information available in the IIDM model, as it is the only data used in the analysis. We cannot export something different from what is used.

We propose removing all code related to this property. Additionally, one unit test needs to be updated. These changes should have been implemented in 1EQ_24SSH, but I forgot to include them. Sorry.

Below is our proposal for the `obtainOperatingMode` method.

```
private static <I extends ReactiveLimitsHolder & Injection<I>> String obtainOperatingMode(I i, double minP, double maxP, double targetP) {
        String calculatedKind = obtainCalculatedSynchronousMachineKind(minP, maxP, obtainCurve(i), i instanceof Generator gen && gen.isCondenser());
        return obtainOperatingMode(targetP, i, calculatedKind);
    }
```

```

File: cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java [hunk 10]
```diff
@@ -358,20 +392,48 @@ private static void writeBatteries(Network network, String cimNamespace, XMLStre
...
+                    return OPERATING_MODE_GENERATOR;
+                } else if (calculatedKind.contains(OPERATING_MODE_MOTOR)) {
+                    return OPERATING_MODE_MOTOR;
+                } else {
+                    return OPERATING_MODE_CONDENSER;
+                }
+            }
+        }
+    }
+
```
marqueslanauja, line 390: If I'm correct, the method `isOperatingAsACondenser` will always return `true` if `targetQ` is not forced to be != 0.0. 

First, I think that targetQ = 0.0 is a perfectly valid target for generators and batteries operating as condensers and when the IIDM model is valid at the `SSH` level, either` targetV` or `targetQ` must be defined for generators and batteries. The check methods do not allow a generator or battery to have both `targetV` and `targetQ` undefined. 

Reviewing the voltageRegulation extension, I realized that only the regulating terminal is checked. I think targetV should also be checked. Maybe, it’s better to do it in a separate PR.

Below is our proposal for the `obtainOperatingMode` method.

```
private static String obtainOperatingMode(double targetP, Injection<?> injection, String calculatedKind) {
        if (targetP < 0) {
            return OPERATING_MODE_MOTOR;
        } else if (targetP > 0) {
            return OPERATING_MODE_GENERATOR;
        } else {
            if (calculatedKind.contains(OPERATING_MODE_CONDENSER)) {
                return OPERATING_MODE_CONDENSER;
            } else if (calculatedKind.contains(OPERATING_MODE_GENERATOR)) {
                return OPERATING_MODE_GENERATOR;
            } else {
                return OPERATING_MODE_MOTOR;
            }
        }
    }
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
colineplqt: That's right!

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/GeneratorConversionTest.java
colineplqt: I just forgot to change them. Thanks!

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
colineplqt: I wanted to write a test for that part and I just realized that there is a problem. In case of a battery, we don't have the isCondenser flag. So in the EQ should it always have `condenser` in the kind?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
colineplqt: Yes you are right! Actually I don't see a use for the old `operatingMode` anymore because `calculatedOperatingMode` should not be null.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
colineplqt: I have modified the `obtainOperatingMode`accordingly. Thanks. Should we remove the code related to the property in another PR? What do you think is best?

marqueslanauja: If we decide to remove the property, I would prefer to do it here in this PR.
However, I reviewed the` SteadyStateHypothesisExport.obtainOperatingMode(I i, double minP, double maxP, double targetP)` method, and the property is still being used. Before removing the property, we need to modify this method to stop using the property and instead rely only on `calculatedOperatingMode`.

colineplqt: Yes I have done the changes accordingly. I was also wondering about the `Conversion.PROPERTY_CGMES_SYNCHRONOUS_MACHINE_TYPE`. It is use in the EQ export but mostly to log if the type has changed. I don't know if it is still useful or if we could also remove it.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
colineplqt: Now all the batteries can act as a condenser.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
colineplqt: Yes, it's true that the `targetV`of the extension should also be checked, in another PR.

About the `obtainOperatingMode`, the idea that I coded was that if `targetP` is 0, we first check if the generator/battery is actually operating as a `condenser`. If it is not, then it means that it is either disconnected or connected and waiting for instructions to producte active or reactive power power. Then, first I think the operating mode is not really important because all the possibilities are wrong: the generator or battery is not operating as a generator, a motor or a condenser. However, I would prioritize the operating mode `generator` if the synchronous machine can, as it would be probably connected to inject active power on the network. Then if it cannot, it would be `motor` and if it cannot it would be `condenser`.

marqueslanauja: Understood. That’s fine.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
marqueslanauja: I think we should preserve it. It’s an attribute recorded from the EQ file and used during the export process to export the same initial type whenever possible. Then I’d prefer not to delete the complete test `EquipmentExportTest.synchronousMachineKindExportAndImportTest`. I think It’s better to remove the operating-mode check and keep the rest. Here is my proposal:

```
@Test
    void synchronousMachineKindExportAndImportTest() throws IOException {
        Network network = createOneGeneratorNetwork();

        // Define the synchronous machine kind property
        Generator expectedGenerator = network.getGenerator("generator1");
        expectedGenerator.setMinP(-50.0).setMaxP(0.0).setTargetP(-10.0);
        String expectedSynchronousMachineKind = "motorOrCondenser";
        expectedGenerator.setProperty(Conversion.PROPERTY_CGMES_SYNCHRONOUS_MACHINE_TYPE, expectedSynchronousMachineKind);

        // Export as cgmes
        Path outputPath = tmpDir.resolve("temp.cgmesExport");
        Files.createDirectories(outputPath);
        String baseName = "oneGeneratorSynchronousMachineKind";
        new CgmesExport().export(network, new Properties(), new DirectoryDataSource(outputPath, baseName));

        // re-import
        Network actual = new CgmesImport().importData(new DirectoryDataSource(outputPath, baseName), NetworkFactory.findDefault(), new Properties());
        Generator actualGenerator = actual.getGenerator("generator1");

        // check the synchronous machine kind
        String actualSynchronousMachineKind = actualGenerator.getProperty(Conversion.PROPERTY_CGMES_SYNCHRONOUS_MACHINE_TYPE);
        assertEquals(expectedSynchronousMachineKind, actualSynchronousMachineKind);
    }
```

marqueslanauja: ok

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
```diff
@@ -358,20 +393,53 @@ private static void writeBatteries(Network network, String cimNamespace, XMLStre
...
+    private static boolean isOperatingAsACondenser(Injection<?> injection) {
+        switch (injection) {
+            case Generator generator -> {
+                if (generator.isVoltageRegulatorOn()) {
+                    double targetV = generator.getTargetV();
+                    return !Double.isNaN(targetV) && targetV != 0;
+                } else {
+                    double targetQ = generator.getTargetQ();
+                    return !Double.isNaN(targetQ) && targetQ != 0;
+                }
```
rcourtier: I suggest a simplification: it isn't necessary to check targetV value since it can't be negative or null because of the checks in the adder and setter.

```suggestion
                return generator.isVoltageRegulatorOn() && !Double.isNaN(generator.getTargetV())
                        || !Double.isNaN(generator.getTargetQ()) && generator.getTargetQ() != 0;
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
```diff
@@ -358,20 +393,53 @@ private static void writeBatteries(Network network, String cimNamespace, XMLStre
...
+            case Battery battery -> {
+                double targetQ = battery.getTargetQ();
+                VoltageRegulation voltageRegulationExtension = battery.getExtension(VoltageRegulation.class);
+                if (voltageRegulationExtension != null && voltageRegulationExtension.isVoltageRegulatorOn()) {
+                    double targetV = voltageRegulationExtension.getTargetV();
+                    if (!Double.isNaN(targetV) && targetV != 0) {
+                        return true;
+                    }
+                }
+                return !Double.isNaN(targetQ) && targetQ != 0;
```
rcourtier:
```suggestion
                VoltageRegulation voltageRegulation = battery.getExtension(VoltageRegulation.class);
                return voltageRegulation != null && voltageRegulation.isVoltageRegulatorOn() && !Double.isNaN(voltageRegulation.getTargetV())
                        || !Double.isNaN(battery.getTargetQ()) && battery.getTargetQ() != 0;
```

docs/grid_exchange_formats/cgmes/export.md
```diff
@@ -298,6 +298,29 @@ generator has the extension [`RemoteReactivePowerControl`](../../grid_model/exte
...
+  - if the minimum and the maximum active power limit are positive, then the generator or battery will be exported as `generatorOrCondenser`,
+  - if the minimum and the maximum active power limit are negative, then the generator or battery will be exported as `motorOrCondenser`,
+  - if the minimum and the maximum active power limit are both equal to zero, then the generator or battery will be exported as `condenser`,
+  - otherwise, the generator or battery will be exported as `generatorOrCondenserOrMotor`.
+- if the flag `isCondenser` is `false`:
+  - if the minimum active power limit is positive, then the generator or battery will be exported as `generator`,
+  - if the maximum active power limit is negative, then the generator or battery will be exported as `motor`,
+  - otherwise, the generator will be exported as `generatorOrMotor`.
+
+The `SynchronousMachine.SynchronousMachineOperatingMode` is exported in the SSH profile depending on the target active
```
rcourtier:
```suggestion
The `SynchronousMachine.operatingMode` is exported in the SSH profile depending on the target active 
```

docs/grid_exchange_formats/cgmes/export.md
```diff
@@ -298,6 +298,29 @@ generator has the extension [`RemoteReactivePowerControl`](../../grid_model/exte
...
+  - if the minimum active power limit is positive, then the generator or battery will be exported as `generator`,
+  - if the maximum active power limit is negative, then the generator or battery will be exported as `motor`,
+  - otherwise, the generator will be exported as `generatorOrMotor`.
+
+The `SynchronousMachine.SynchronousMachineOperatingMode` is exported in the SSH profile depending on the target active 
+power of the generator or battery and on fact that it is regulating or not:
+- if the target active power is positive, then the generator or battery will be exported as `generator`,
+- if the target active power is negative, then the generator or battery will be exported as `motor`,
+- if the target active power is zero and the generator or battery is regulating, then the operating mode will be `condenser`.
+- otherwise, the generator or battery will be exported as `generator` if is allowed by its `SynchronousMachine.SynchronousMachineKind`,
```
rcourtier:
```suggestion
- otherwise, the generator or battery will be exported as `generator` if is allowed by its `SynchronousMachine.type`,
```

docs/grid_exchange_formats/cgmes/export.md
```diff
@@ -298,6 +298,29 @@ generator has the extension [`RemoteReactivePowerControl`](../../grid_model/exte
 with the `enabled` activated and the generator attribute `voltageRegulatorOn` set to `false`. In all other cases, a
 `RegulatingControl` is exported with `RegulatingControl.mode` set to `RegulatingControlModeKind.voltage`.
 
+#### Synchronous machine kind (EQ) and operating mode (SSH)
```
rcourtier:
```suggestion
#### SynchronousMachine type (EQ) and operatingMode (SSH)
```

docs/grid_exchange_formats/cgmes/export.md
```diff
@@ -298,6 +298,29 @@ generator has the extension [`RemoteReactivePowerControl`](../../grid_model/exte
...
+
+The `SynchronousMachine.SynchronousMachineKind` is exported in the EQ profile depending on the [reactive limits](../../grid_model/additional.md#reactive-limits) of the
```
rcourtier:
```suggestion
The `SynchronousMachine.type` is exported in the EQ profile depending on the [reactive limits](../../grid_model/additional.md#reactive-limits) of the 
```

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/GeneratorConversionTest.java
```diff
@@ -60,4 +62,155 @@ void reactiveLimitsExportTest() throws IOException {
...
+        assertTrue(network.getGenerator("sm_generator_condenser").isCondenser());
+        assertTrue(network.getGenerator("sm_condenser").isCondenser());
+    }
+
+    @Test
+    void synchronousMachineOperatingModeAndKindConversion() throws IOException {
+        Network network = createNetwork();
+
+        String eqXml = ConversionUtil.writeCgmesProfile(network, "EQ", tmpDir, new Properties());
+        String sshXml = ConversionUtil.writeCgmesProfile(network, "SSH", tmpDir, new Properties());
```
rcourtier, line 81: You can use this prefix
```suggestion
        String sshXml = ConversionUtil.writeCgmesProfile(network, "SSH", tmpDir, new Properties());
        String type = "http://iec.ch/TC57/2013/CIM-schema-cim16#SynchronousMachineKind.";
```

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/GeneratorConversionTest.java
```diff
@@ -60,4 +62,155 @@ void reactiveLimitsExportTest() throws IOException {
...
+
+        String generator1Eq = getElement(eqXml, "SynchronousMachine", "GEN");
+        assertNotNull(generator1Eq);
+        assertTrue(generator1Eq.contains("SynchronousMachineKind.generatorOrCondenser"));
```
rcourtier: And check that the requested CGMES attribute of that rdf element has the expected value:
```suggestion
        assertEquals(type + "generatorOrCondenser", getResource(generator1Eq, "SynchronousMachine.type"));
```

rcourtier (approved):

marqueslanauja (approved):

PR #3605, opened 2025-10-13
CimCharacteristics cleanup

rcourtier:
**Please check if the PR fulfills these requirements**

- [X] The commit message follows our guidelines
- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been added / updated (for bug fixes / features)
- [ ] A PR or issue has been opened in all impacted repositories (if any)

**Does this PR already have an issue describing the problem?**

No.

**What kind of change does this PR introduce?**

Refactoring and documentation update.

**What is the current behavior?**

Topology kind of imported CGMES dataset is stored in a network property as well as in the dedicated extension CimCharacteristics.

**What is the new behavior (if this is a feature change)?**

Topology kind of imported CGMES dataset is only stored in the dedicated extension CimCharacteristics.

**Does this PR introduce a breaking change or deprecate an API?**
- [X] Yes
- [ ] No

**If yes, please check if the following requirements are fulfilled**

- [x] The *Breaking Change* or *Deprecated* label has been added
- [x] The migration steps are described in the following section

**What changes might users need to make in their application due to this PR? (migration steps)**

The topology kind of the imported CGMES dataset is no more stored as properties of the network at the import.
It was previously stored twice: as a property and in the CimCharacteristics extension.

If you relied on this property, you can replace your usages:

```java
String topologyKind = network.getProperty(Conversion.NETWORK_PS_CGMES_MODEL_DETAIL);

// Possible values:
// - Conversion.NETWORK_PS_CGMES_MODEL_DETAIL_BUS_BRANCH (="bus-branch")
// - Conversion.NETWORK_PS_CGMES_MODEL_DETAIL_NODE_BREAKER (="node-breaker")
```

by

```java
CimCharacteristics cimCharacteristics = network.getExtension(CimCharacteristics.class);
CgmesTopologyKind topologyKind = cimCharacteristics.getTopologyKind();

// Possible values:
// - CgmesTopologyKind.NODE_BREAKER
// - CgmesTopologyKind.BUS_BRANCH
```

As a consequence, the following constants (which are now useless) have been removed:
- `Conversion.NETWORK_PS_CGMES_MODEL_DETAIL`;
- `Conversion.NETWORK_PS_CGMES_MODEL_DETAIL_BUS_BRANCH`;
- `Conversion.NETWORK_PS_CGMES_MODEL_DETAIL_NODE_BREAKER`.

**Other information**:

Documentation about CimCharacteristics extension was missing and has been added.

PR #3604, opened 2025-10-10
CgmesMetadataModels cleanup

rcourtier:
**Please check if the PR fulfills these requirements**

- [X] The commit message follows our guidelines
- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been added / updated (for bug fixes / features)
- [ ] A PR or issue has been opened in all impacted repositories (if any)

**Does this PR already have an issue describing the problem?**

No.

**What kind of change does this PR introduce?**

Refactoring and documentation update.

**What is the current behavior?**

At CGMES import, model IDs are imported into the CgmesMetadataModels extension, but also as network properties.

**What is the new behavior (if this is a feature change)?**

Model IDs and other metadata information are imported only into the dedicated CgmesMetadataModels extension.

**Does this PR introduce a breaking change or deprecate an API?**
- [X] Yes
- [ ] No

**If yes, please check if the following requirements are fulfilled**

- [x] The *Breaking Change* or *Deprecated* label has been added
- [x] The migration steps are described in the following section

**What changes might users need to make in their application due to this PR? (migration steps)**

The CGMES profile IDs are no more stored as properties of the network at the import.
They were previously stored twice: as properties and in the CgmesMetadataModels extension.

If you relied on these properties, you can replace your usages:

```java
String id = network.getProperty(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + profile + "_ID");
```

by

```java
CgmesMetadataModels models = network.getExtension(CgmesMetadataModels.class);
CgmesMetadataModel profileModel = models.getModelForSubset(CgmesSubset.valueOf(profile)).orElseThrow();
String id = profileModel.getId();
```

**Other information**:

Documentation about CgmesMetadataModels extension was missing and has been added.

PR #3633, opened 2025-10-29
cgmes import: config option to remove properties and aliases after import

### LINKED ISSUE #3613: Add config option to not create properties and aliases at CGMES import

At CGMES import, a lot of properties and aliases are created. They are used to store data that don't have a native placeholder in IIDM format. Their purpose can be to keep track of some data (identifiers, original class, ...) to be able to export them identically at CGMES export for the sake of persistency, or to be able to call an update of the network (update already imported network with SSH data).

In the case the user doesn't need to keep CGMES data for the sake of persistency, or in the case the user doesn't need to perform an update of the network, these properties and aliases created at CGMES don't need to be created. A configuration option should give that possibility to the user.

### Describe the motivation

Not storing CGMES import related properties and aliases would reduce the size of the IIDM network at serialization, and thus improve performances.

### Extra Information

_No response_

marqueslanauja:

Fixes #3613  

Properties and aliases are generated during the EQ import process and are used both in the initial and subsequent network updates. 

To remove all the properties and aliases after import, the property `iidm.import.cgmes.remove-properties-and-aliases-after-import` must be set to `true`.

When this option is set to true, all generated properties and aliases are removed after the first import. During this initial import, both EQ and SSH files must be provided to obtain a valid network at the steady state hypothesis level. 
Removing properties and aliases invalidates all subsequent updates. 

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
@@ -176,6 +176,9 @@ public class Conversion {
         cgmes.baseVoltages().forEach(bv -> bvAdder.addBaseVoltage(bv.getId("BaseVoltage"), bv.asDouble("nominalVoltage"), isBoundaryBaseVoltage(bv.getLocal("graph"))));
         bvAdder.add();
         cgmes.computedTerminals().forEach(t -> context.terminalMapping().buildTopologicalNodeCgmesTerminalsMapping(t));
+        cgmes.computedTerminals().stream()
+                .filter(t -> !CgmesNames.SWITCH_TYPES.contains(t.conductingEquipmentType()))
+                .forEach(t -> context.terminalMapping().buildConnectivityNodeCgmesTerminalsMapping(t));
         cgmes.regulatingControls().forEach(p -> context.regulatingControlMapping().cacheRegulatingControls(p));
         context.popReportNode();
 
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Context.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Context.java
@@ -55,7 +55,7 @@ public class Context {
         terminalMapping = new TerminalMapping();
         loadingLimitsMapping = new LoadingLimitsMapping(this);
         regulatingControlMapping = new RegulatingControlMapping(this);
-        nodeMapping = new NodeMapping();
+        nodeMapping = new NodeMapping(this);
 
         cachedGroupedTransformerEnds = new HashMap<>();
         cachedGroupedRatioTapChangers = new HashMap<>();
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/NodeMapping.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/NodeMapping.java
@@ -9,9 +9,11 @@
 package com.powsybl.cgmes.conversion;
 
 import com.powsybl.cgmes.model.CgmesTerminal;
+import com.powsybl.commons.PowsyblException;
 import com.powsybl.iidm.network.VoltageLevel;
 
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 /**
@@ -19,9 +21,12 @@ import java.util.Map;
  */
 public class NodeMapping {
 
-    public NodeMapping() {
-        cgmes2iidm = new HashMap<>(100);
-        voltageLevelNumNodes = new HashMap<>(100);
+    private final Map<String, Integer> cgmes2iidm = HashMap.newHashMap(100);
+    private final Map<VoltageLevel, Integer> voltageLevelNumNodes = HashMap.newHashMap(100);
+    private final Context context;
+
+    public NodeMapping(Context context) {
+        this.context = context;
     }
 
     public int iidmNodeForTopologicalNode(String id, int associatedNode, VoltageLevel vl) {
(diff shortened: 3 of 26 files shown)


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -263,6 +263,11 @@ public Network convert(ReportNode reportNode) {
 
         updateWithAllInputs(network, reportNode);
 
+        // Remove all properties and aliases, this will invalidate all subsequent updates
+        if (context.config().getRemovePropertiesAndAliasesAfterImport()) {
+            removeAllAliasesAndProperties(network);
+        }
+
```
rcourtier: The `removeAllAliasesAndProperties` is called here at the end of the `convert`, which is good. But I think the same should be done at the end of the `update`.

Right now the method lacks a `Properties properties` argument to pass the import parameters:
```java
public void update(Network network, ReportNode reportNode)
```

But you can add it and pass it from where it is called in `CgmesImport.update`.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -281,6 +286,34 @@ private void updateWithAllInputs(Network network, ReportNode reportNode) {
         update(network, updateContext, reportNode);
     }
 
+    private static void removeAllAliasesAndProperties(Network network) {
+        network.getIdentifiables().forEach(identifiable -> {
+            identifiable.getAliases().stream().toList().forEach(identifiable::removeAlias);
+            identifiable.getPropertyNames().stream().toList().forEach(identifiable::removeProperty);
```
rcourtier: You don't need to `stream().toList()`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -281,6 +286,34 @@ private void updateWithAllInputs(Network network, ReportNode reportNode) {
...
+            removeProperties(t3w.getLeg3().getOperationalLimitsGroups());
+        });
+
+        network.getDanglingLines().forEach(danglingLine ->
+                removeProperties(danglingLine.getOperationalLimitsGroups()));
+    }
+
+    private static void removeProperties(Collection<OperationalLimitsGroup> operationalLimitsGroupCollection) {
+        operationalLimitsGroupCollection.forEach(operationalLimitsGroup ->
+                operationalLimitsGroup.getPropertyNames().stream().toList().forEach(operationalLimitsGroup::removeProperty));
```
rcourtier: You don't need to `stream().toList()`

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/BusbarSectionUpdateTest.java
```diff
@@ -60,6 +60,27 @@ void usePreviousValuesTest() {
         assertEq(network);
     }
 
+    @Test
+    void removeAllPropertiesAndAliasesTest() {
```
rcourtier, line 64: I wonder if we should have only one test where we import a network as complete as possible (possibly a microgrid), serialize it to XIIDM and verify that the output file doesn't contain a iidm:property or iidm:alias tag? That way we can be sure that no other property or alias get imported, and so no more implementation of `removeAllPropertiesAndAliasesTest` test regarding the future changes in IIDM model/CGMES import is needed.
What do you think?

rcourtier:
Tested with RTE network:
- Without the import option: 3800 properties, 6052 aliases, 34,27 MB for the XIIDM file.
- With the import option: 0 property, 0 alias, 15,87 MB for the XIIDM file.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
marqueslanauja: Done. I've removed it

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
marqueslanauja: Done. I've removed it

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
marqueslanauja: The `removeAllAliasesAndProperties` method is called at the end of the conversion, after the first update. It is not necessary to call it again at the end of the `update` method since subsequent updates are not possible. Without properties, the update process will crash. I’ve added a specific exception to handle cases where a subsequent update is attempted without properties.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/BusbarSectionUpdateTest.java
marqueslanauja: We try to minimize resource usage and execution time in the unit tests. This is the main reason I decided to add a very fast unit test to all the update cases. It also makes sense to add a new update case test for each new piece of equipment that will be added to the model in the future.

Adding a test that imports a complete model would increase the execution time, and we’d prefer not to do that. However, if you want it, I can add it. Furthermore, it is difficult to have always a complete model, and we have to consider that missing to remove some properties in the future will not cause a malfunction in the code, only an increase in memory usage.

marqueslanauja:
While testing the PR with real cases, I noticed that the CGMES export crashes in certain scenarios when the model is exported without properties or aliases. The cgmes export must always work, with or without properties. My proposal is to fix this issue in a separated pull request.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
rcourtier: That is correct, but what I meant is a user could import an EQ without this option, proceed with several SSH updates, and finally perform a last update with this option set. If that use case makes sense, that would be useful to be able to call this option during the update.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/BusbarSectionUpdateTest.java
rcourtier: It's fine, let's keep it that way.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
marqueslanauja: You're right, understood. It's enough to do it only at the end of the update process. Done.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -289,6 +317,11 @@ private static boolean sshOrSvIsIncludedInCgmesModel(CgmesModel cgmes) {
 
     public void update(Network network, ReportNode reportNode) {
         Objects.requireNonNull(network);
+
+        if (network.getPropertyNames().isEmpty()) {
+            throw new ConversionException("The network has no properties and aliases, they have been removed. Update is not allowed.");
+        }
```
rcourtier: The idea is good, but the check of the presence of properties shouldn't be made on the network. The reason is that the properties on network will be removed with #3604 and #3605. I recommend you check instead if any of the network's identifiable has a property:

```suggestion
        if (network.getIdentifiables().stream().allMatch(i -> i.getPropertyNames().isEmpty())) {
            throw new ConversionException("The network has no properties and aliases, they have been removed. Update is not allowed.");
        }
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -334,6 +367,11 @@ private void update(Network network, Context updateContext, ReportNode reportNod
 
         network.runValidationChecks(false, reportNode);
         network.setMinimumAcceptableValidationLevel(ValidationLevel.STEADY_STATE_HYPOTHESIS);
+
+        // Remove all properties and aliases, this will invalidate all subsequent updates
+        if (updateContext.config().getRemovePropertiesAndAliasesAfterImport()) {
+            removeAllAliasesAndProperties(network);
+        }
```
olperr1: The documentation states that:

> When this option is set to true, all generated properties and aliases are removed after the first import.

Due to its location, this operation is also performed after each update. So you can import the EQ profile alone, and then update the network with another profiles, and remove the properties and aliases by providing the import parameter only for the update operation.
I think that this behavior can be useful. But if we keep it, the documentation should be changed.

docs/grid_exchange_formats/cgmes/import.md
```diff
@@ -677,3 +677,12 @@ Optional property that defines whether the CGMES importer should use previous va
 When EQ and one or more SSH files are imported separately, and this property is set to `true`, the importer will use values from previously imported SSH files to complete missing attributes in the SSH file currently being imported.
 If set to `false`, missing SSH attributes will be filled using default values.
 `false` by default.
+
+**iidm.import.cgmes.remove-properties-and-aliases-after-import**  
+Properties and aliases are generated during the EQ import process and are used both in the initial and subsequent network updates.
+When this option is set to `true`, all generated properties and aliases are removed after the first import.
```
olperr1: Due to its location, this operation is also performed after each update. So you can import the EQ profile alone, and then update the network with another profiles, and remove the properties and aliases by providing the import parameter only for the update operation.
(This behavior is tested in `LoadUpdateTest.removeAllPropertiesAndAliasesAfterSomeUpdatesTest()`.)
I think that the documentation should be more precise on this point.

marqueslanauja: Done. I've updated the documentation, and I’ve also removed the properties during the initial import when no `SSH` file is provided and the configuration flag is set to `true`. The code did not previously handle this case.

rcourtier (approved):

docs/grid_exchange_formats/cgmes/import.md
olperr1: You are right, we missed this case.
Thanks for the changes!

Issue #3537, opened 2025-08-05
Store the locale targetV of a generator


alicecaron:
### Describe the current behavior

The targetV of a generator is the target voltage by a local or remote regulating terminal. No local `targetV` can be defined for the generator.

### Describe the expected behavior

A generator has a `localBackupTargetV` available, that can be null or set by he user for a local regulating control.

### Describe the motivation

Be able to get a set local targetV of a generator besides the targetV that can be set for remote regulation.

### Extra Information

The generator has a new attribute `localBackupTargetV` in IIDM that can be null or set. If set, when passing from remote to local regulation, the targetV should be updated with the localBackupTargetV if present.  
When creating he generator it is possible to:
- set the target V like before: setTargetV(v) => set the `targetV` and set `localBackupTargetV` to `NaN`
- set both argets: setTargetV(v, vLocaleBackup) => set the `targetV` and `localBackupTargetV`

It is not wanted to be able to set only the `localBackupTargetV`: no method `setLocalBackupTargetV()`

Update the method when referenced terminal is removed from regulation, follows #3437

snoopfab:
I think we need to bump the iidm version to for this ? ( latest 1.14 doesn't contain any localBackupTargetV  in xsd)

MatthieuSAUR:
I used `equivalentLocalTargetV` instead of `localBackupTargetV`
In the near future, we are going to refactor the voltage control (see #3618 ) so in this issue we use variables and methods that can be easily deprecated

PR #3635, opened 2025-10-30
Store the locale targetV of a generator

MatthieuSAUR:

issue #3537 

feature

issue #3537 

issue #3537 

- [x] Yes
- [ ] No

- [x] The Breaking Change or Deprecated label has been added

Projects having their own IIDM implementation should now implement the following new methods:
- `Generator.getEquivalentLocalTargetV()`
- `Generator.setTargetV(double targetV, double equivalentLocalTargetV)`
- `GeneratorAdder.setTargetV(double targetV, double equivalentLocalTargetV)`

**Other information**:


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Generator.java
```diff
@@ -195,13 +195,31 @@ public interface Generator extends Injection<Generator>, ReactiveLimitsHolder {
...
      */
     Generator setTargetV(double targetV);
 
+    /**
+     * Get the local TargetV target in kV.
+     * <p>
+     * Depends on the working variant.
+     * @see VariantManager
+     */
+    double getLocalTargetV();
```
vidaldid-rte: @rcourtier @MatthieuSAUR 
This name can be confusing. Users can believe it is the method to call to get the voltage target when a generator is in local voltage control.
In reality, this is meant to be the local value to use to reach the remote targetV, when simulating in local votlage control only.

Maybe rename getEquivalentLocalTargetV() ?  (same for getter)

MatthieuSAUR: This name is a bit confusing. So we are adding precise Javadoc.
The "Equivalent" name can be misleading too, we can think that the method returns a value that has been calculated from the TargetV.
@vidaldid-rte What do you think with the new JavaDoc ?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Generator.java
```diff
@@ -195,13 +195,44 @@ public interface Generator extends Injection<Generator>, ReactiveLimitsHolder {
     double getTargetV();
 
     /**
-     * Set the voltage target in kV.
      * <p>
-     * Depends on the working variant.
+     *     Set the voltage target in kV AND set the local voltage target to {@link Double#NaN}
```
vidaldid-rte: I think this is still ambiguous. A try, using the equivalentLocalTargetV term

 Set the voltage target in kV, for the regulated terminal which can be remote or local AND without setting the equivalentLocalTargetV which takes the value {@link Double#NaN}

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Generator.java
```diff
@@ -195,13 +195,44 @@ public interface Generator extends Injection<Generator>, ReactiveLimitsHolder {
...
+     * <p/>
+     * <p>Depends on the working variant.</p>
      * @see VariantManager
      */
     Generator setTargetV(double targetV);
 
+    /**
+     * <p>
+     *     Get the local terminal voltage target in kV.
+     * </p>
```
vidaldid-rte, line 215: If set, returns a local target voltage that is expected to be consistent with the remote target voltage. When defined, this value can be used by silulators that deactivate the remote voltage algorithmes, or by dynamic simulators that use this voltage as a starting value.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Generator.java
vidaldid-rte: @MatthieuSAUR  comments added to the javadoc to try to correct the ambiguities that may remain.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/GeneratorImpl.java
```diff
@@ -191,17 +195,31 @@ public double getTargetV() {
...
         ValidationUtil.checkVoltageControl(this, regulatingPoint.isRegulating(variantIndex),
-                targetV, targetQ.get(variantIndex), n.getMinValidationLevel(), n.getReportNodeContext().getReportNode());
-        double oldValue = this.targetV.set(variantIndex, targetV);
+            targetV, targetQ.get(variantIndex), n.getMinValidationLevel(), n.getReportNodeContext().getReportNode());
+        double oldValueTargetV = this.targetV.set(variantIndex, targetV);
         String variantId = network.get().getVariantManager().getVariantId(variantIndex);
         n.invalidateValidationLevel();
-        notifyUpdate("targetV", variantId, oldValue, targetV);
+        notifyUpdate("targetV", variantId, oldValueTargetV, targetV);
+        //
```
colineplqt:
```suggestion

```

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractGeneratorTest.java
```diff
@@ -355,19 +359,74 @@ public void testSetterGetterInMultiVariants() {
...
-                .setEnergySource(source)
-                .setMaxP(maxP)
-                .setMinP(minP)
-                .setRatedS(ratedS)
-                .setTargetP(activePowerSetpoint)
-                .setTargetQ(reactivePowerSetpoint)
-                .setNode(1)
-                .setTargetV(voltageSetpoint)
-                .add();
+            .setId(id)
```
colineplqt, line 404: I guess you have left two methods here to call in one case `setTargetV(double)` and in the other `setTargetV(double, double)`?

MatthieuSAUR: Yes, that's right

PR #3640, opened 2025-11-05
Add droop control mode for AC/DC converters

2nisb:

No

Feature

AC DC converters cannot be set in Droop control mode

Is it possible to set AC DC converters in Droop control mode, with multiple segments droop. 
The droop control mode is use in post contingency loadflow, droop controlling converters introduce a linear relation between the steady-state deviation of DC voltage and the steady-state deviation of power setpoint. With multiple segments droop, the linear coefficient changes depending of the actual voltage of the converter.

- [x] Yes

If you defined your own IIDM implementation, you should:
- Write your own implementations of `DroopCurve` and `DroopCurveAdder`
- Implement the following methods in your `AcDcConverter` implementations:
  - `DroopCurveAdder newDroopCurve()`
  - `DroopCurve getDroopCurve()`

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractAcDcConverter.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractAcDcConverter.java
@@ -200,7 +200,7 @@ abstract class AbstractAcDcConverter<I extends AcDcConverter<I>> extends Abstrac
     public I setPccTerminal(Terminal pccTerminal) {
         Objects.requireNonNull(pccTerminal);
         ValidationUtil.checkModifyOfRemovedEquipment(this.id, this.removed, PCC_TERMINAL);
-        ValidationUtil.checkAcDcConverterPccTerminal(this, getTerminal2().isPresent(), pccTerminal, getTerminal1().getVoltageLevel());
+        ValidationUtil.checkAcDcConverterPccTerminal(this, pccTerminal, getTerminal1().getVoltageLevel());
         Terminal oldValue = pccRegulatingPoint.getRegulatingTerminal();
         pccRegulatingPoint.setRegulatingTerminal((TerminalExt) pccTerminal);
         notifyUpdate(PCC_TERMINAL, oldValue, pccRegulatingPoint.getRegulatingTerminal());
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
@@ -608,6 +608,12 @@ public class CgmesImport implements Importer {
                                 getFormat(),
                                 p,
                                 USE_PREVIOUS_VALUES_DURING_UPDATE_PARAMETER,
+                                defaultValueConfig))
+                .setRemovePropertiesAndAliasesAfterImport(
+                        Parameter.readBoolean(
+                                getFormat(),
+                                p,
+                                REMOVE_PROPERTIES_AND_ALIASES_AFTER_IMPORT_PARAMETER,
                                 defaultValueConfig));
 
         String namingStrategy = Parameter.readString(getFormat(), p, NAMING_STRATEGY_PARAMETER, defaultValueConfig);
@@ -684,6 +690,7 @@ public class CgmesImport implements Importer {
     public static final String IMPORT_CGM_WITH_SUBNETWORKS_DEFINED_BY = "iidm.import.cgmes.cgm-with-subnetworks-defined-by";
     public static final String CREATE_FICTITIOUS_VOLTAGE_LEVEL_FOR_EVERY_NODE = "iidm.import.cgmes.create-fictitious-voltage-level-for-every-node";
     public static final String USE_PREVIOUS_VALUES_DURING_UPDATE = "iidm.import.cgmes.use-previous-values-during-update";
+    public static final String REMOVE_PROPERTIES_AND_ALIASES_AFTER_IMPORT = "iidm.import.cgmes.remove-properties-and-aliases-after-import";
 
     public static final String SOURCE_FOR_IIDM_ID_MRID = "mRID";
     public static final String SOURCE_FOR_IIDM_ID_RDFID = "rdfID";
@@ -803,6 +810,12 @@ public class CgmesImport implements Importer {
             "Use previous values (from a previous update) during the current update",
             Boolean.FALSE);
 
+    private static final Parameter REMOVE_PROPERTIES_AND_ALIASES_AFTER_IMPORT_PARAMETER = new Parameter(
+            REMOVE_PROPERTIES_AND_ALIASES_AFTER_IMPORT,
+            ParameterType.BOOLEAN,
+            "Remove all properties and aliases after CGMES import",
+            Boolean.FALSE);
+
     private static final List<Parameter> STATIC_PARAMETERS = List.of(
             CONVERT_BOUNDARY_PARAMETER,
             CONVERT_SV_INJECTIONS_PARAMETER,
@@ -823,7 +836,8 @@ public class CgmesImport implements Importer {
             IMPORT_CGM_WITH_SUBNETWORKS_DEFINED_BY_PARAMETER,
             MISSING_PERMANENT_LIMIT_PERCENTAGE_PARAMETER,
             CREATE_FICTITIOUS_VOLTAGE_LEVEL_FOR_EVERY_NODE_PARAMETER,
-            USE_PREVIOUS_VALUES_DURING_UPDATE_PARAMETER);
+            USE_PREVIOUS_VALUES_DURING_UPDATE_PARAMETER,
+            REMOVE_PROPERTIES_AND_ALIASES_AFTER_IMPORT_PARAMETER);
 
     private final Parameter boundaryLocationParameter;
     private final Parameter preProcessorsParameter;
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Context.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Context.java
@@ -55,7 +55,7 @@ public class Context {
         terminalMapping = new TerminalMapping();
         loadingLimitsMapping = new LoadingLimitsMapping(this);
         regulatingControlMapping = new RegulatingControlMapping(this);
-        nodeMapping = new NodeMapping();
+        nodeMapping = new NodeMapping(this);
 
         cachedGroupedTransformerEnds = new HashMap<>();
         cachedGroupedRatioTapChangers = new HashMap<>();
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
@@ -176,6 +176,9 @@ public class Conversion {
         cgmes.baseVoltages().forEach(bv -> bvAdder.addBaseVoltage(bv.getId("BaseVoltage"), bv.asDouble("nominalVoltage"), isBoundaryBaseVoltage(bv.getLocal("graph"))));
         bvAdder.add();
         cgmes.computedTerminals().forEach(t -> context.terminalMapping().buildTopologicalNodeCgmesTerminalsMapping(t));
+        cgmes.computedTerminals().stream()
+                .filter(t -> !CgmesNames.SWITCH_TYPES.contains(t.conductingEquipmentType()))
+                .forEach(t -> context.terminalMapping().buildConnectivityNodeCgmesTerminalsMapping(t));
         cgmes.regulatingControls().forEach(p -> context.regulatingControlMapping().cacheRegulatingControls(p));
         context.popReportNode();
 
@@ -261,13 +264,17 @@ public class Conversion {
 
         CgmesReports.importedCgmesNetworkReport(reportNode, network.getId());
 
-        updateWithAllInputs(network, reportNode);
+        updateWithAllInputs(network, reportNode, context);
 
         return network;
     }
 
-    private void updateWithAllInputs(Network network, ReportNode reportNode) {
+    private void updateWithAllInputs(Network network, ReportNode reportNode, Context importContext) {
         if (!sshOrSvIsIncludedInCgmesModel(this.cgmes)) {
+            // Remove all properties and aliases, this will invalidate all subsequent updates
+            if (importContext.config().getRemovePropertiesAndAliasesAfterImport()) {
+                removeAllAliasesAndProperties(network);
+            }
             return;
         }
         this.cgmes.setQueryCatalog(QUERY_CATALOG_NAME_UPDATE);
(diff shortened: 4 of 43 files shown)


docs/grid_model/network_subnetwork.md
```diff
@@ -975,6 +976,17 @@ between the converter DC Node 1 and the DC Node 2 to be equal to `TargetVdc`
   - `+TargetVdc / 2` at the converter DC Node 1
   - `-TargetVdc / 2` at the converter DC Node 2
 
+When the `ControlMode` of the converter is set to `P_PCC_DROOP`, the converter controls active power as in the `P_PCC` control mode
+for normal load flow, but when a security analysis in run, the converter controls the relation between DC Voltage and DC Power:
+$P_{DC} - P_{REF} = -k * (V_{DC} - V_{REF})$
+Where $k$ is the droop coefficient of the actual droop segment.
```
jeandemanged: I am not sure to fully understand:
- can you define more precisely what are Pdc, Pref, Vdc, Vref ?
- do some of them relate to or equal to targetP / targetVdc ?
- you mention distinction between load flow and security analysis, are some of these values the solved values from the base case load flow? (e.g. Pref is solved Pdc after load flow ?)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AcDcConverterDroopImpl.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * @author Denis Bonnand {@literal <denis.bonnand at supergrid-institute.com>}
+ */
+public class AcDcConverterDroopImpl extends AbstractAcDcConverterDroop {
+
+    AcDcConverterDroopImpl(double uMin, double uMax, double k, AcDcConverter<?> converter) {
+        super(uMin, uMax, k, converter);
+    }
+}
```
jeandemanged: AbstractAcDcConverterDroop seems superfluous since AcDcConverterDroopImpl is the same.
Delete AbstractAcDcConverterDroop and put everything in AcDcConverterDroopImpl.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractAcDcConverter.java
```diff
@@ -334,5 +336,20 @@ public void remove() {
...
+
+    @Override
+    public List<AcDcConverterDroop> getDroopList() {
+        return droopList;
+    }
+
+    @Override
+    public AcDcConverterDroopAdder newDroop() {
+        return new AcDcConverterDroopAdderImpl(this);
+    }
```
jeandemanged, line 347: API users are able to mutate the content of the droop list obtained via getDroopList() which should be forbidden.

I would suggest a different approach where IIDM API defines explicitly the DroopCurve type. A DroopCurve is always defined "at once" and can never be mutated once created/added (but redefining the curve completely is possible). This would allow also checking for overlapping Umin-Umax ranges and raising exceptions as appropriate. The existing ReactiveCapabilityCurve api/impl may serve as an inspiration for this approach.

jeandemanged:
Hello @2nisb ,
Thank you for your contribution, I do have some questions and first remarks above.
I took the liberty to adjust the PR description. Modifying the IIDM API is a breaking change, therefore a migration guide must be written.

docs/grid_model/network_subnetwork.md
2nisb: Sorry for the confusion, I try to make this more understandable:

When we run a security analysis, this equation is activated and :
- Pref is the power which was calculated during the base loadflow, at DC side, so it is not equal to targetP which is the AC setpoint. It represents the operating point before the security analysis starts.
- Vref is the DC voltage which was calculated during the base loadflow. The droop control is only used for P controlled converters, so they should not have a targetVdc.
- Pdc is the actual power at DC side during the security analysis, which is determined by Newton Raphson
- Vdc is the actual DC voltage during the security analysis, which is determined by Newton Raphson

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractAcDcConverter.java
2nisb: Ok I will work on that.

2nisb:
> Hello @2nisb , Thank you for your contribution, I do have some questions and first remarks above. I took the liberty to adjust the PR description. Modifying the IIDM API is a breaking change, therefore a migration guide must be written.

Hi @jeandemanged, thank you for your review, I made changes is the implementation of the droop curve as you mentioned, is it better like this ? 

I am not sure to understand which functions I need to precise in the migration guide ? It is only those which are in an IIDM interface ?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DroopCurve.java
```diff
@@ -0,0 +1,59 @@
...
+ * The following graph shows a 3 segments droop curve. On the X-axis
+ * is the <code>AcDcConverter<code> DC voltage, on the Y-axis the droop coefficient.
+ *<pre>
+ * k
+ * ^
+ * |
+ * |
+ * |------
+ * |      --------
+ * +---------------------> V (MW)
```
jeandemanged: kV ?

```suggestion
 * +---------------------> V (kV)
```

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DroopCurve.java
```diff
@@ -0,0 +1,59 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+package com.powsybl.iidm.network;
+
+import java.util.Collection;
+
+/**
+ * A droop curve to define droop function of a <code>AcDcConverter<code>
+ * in <code>P_PCC_Droop<code> mode.
```
jeandemanged: - fix code closing tag missing '/'
- uppercase P_PCC_DROOP

```suggestion
 * A droop curve to define droop function of a <code>AcDcConverter</code>
 * in <code>P_PCC_DROOP</code> mode.
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DroopCurveImpl.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Override
+    public double getK(double v) {
+        for (Segment segment : segments) {
+            if (v >= segment.getMinV() && v < segment.getMaxV()) {
+                return segment.getK();
+            }
+        }
+        throw new PowsyblException("Droop coefficient is not defined for Vdc = " + v);
```
jeandemanged: Instead of throwing, wouldn't it make sense instead to
- first sort the segments (similarly as in DroopCurveAdder) and then:
  - if requested v is below defined range , then use first K
  - if requested v is above defined range, then use last K
?

And corner case of empty curve: just return zero ?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DroopCurveImpl.java
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        @Override
+        public double getMaxV() {
+            return maxV;
+        }
+
+    }
+
+    private final HashSet<Segment> segments;
```
geofjamg: I don't think you need to HashSet because, equals and hashcode are not implemented for SegmentImpl. As segments have already been sorted and checked in the adder, you just need to simple list.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DroopCurveAdderImpl.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.*;
+
+/**
+ * @author Denis Bonnand {@literal <denis.bonnand at supergrid-institute.com>}
+ */
+class DroopCurveAdderImpl<O extends Validable & AcDcConverter<?>> implements DroopCurveAdder {
+
+    private final O owner;
+
+    private final HashSet<DroopCurve.Segment> segments = new HashSet<>();
```
geofjamg: Same here, the hashset is useless, as it is required by the adder to create segment in the right order and with continuity (why not!), you can just use a List

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/AcDcConverter.java
```diff
@@ -229,4 +241,19 @@ enum ControlMode {
...
+
+    /**
+     * Create a Droop curve;
+     */
+    DroopCurveAdder newDroopCurve();
+
+    /**
+     * Set the Droop curve;
+     */
+    void setDroopCurve(DroopCurve droopCurve);
```
geofjamg: This method should not be part of the API, just keep it on the implementation side (and no problem in the impl side if you need to cast)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractAcDcConverter.java
```diff
@@ -34,6 +34,8 @@ abstract class AbstractAcDcConverter<I extends AcDcConverter<I>> extends Abstrac
     private double switchingLoss;
     private double resistiveLoss;
 
+    private DroopCurve droopCurve;
```
jeandemanged: as it is null at creation, API users might face null pointer exceptions etc..

See proposal https://github.com/powsybl/powsybl-core/pull/3640/files#r2537027367

And here in AbstractAcDcConverter, set the droopCurve to be initially an Empty DroopCurve

```suggestion
    private DroopCurve droopCurve = DroopCurve.EMPTY;
```

jeandemanged (changes requested):

docs/grid_model/network_subnetwork.md
jeandemanged: Thanks for clarifying!
Minor comment: please check also https://powsybl--3640.org.readthedocs.build/projects/powsybl-core/en/3640/grid_model/network_subnetwork.html#ac-dc-converter there might be some line breaks to be added for readability of the added doc

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DroopCurve.java
```diff
@@ -0,0 +1,59 @@
...
+ * <p>
+ * This curve is made of <code>Segment</code> and each segment is defined by its minimum
+ * and maximum voltage associated with a droop coefficient.
+ * <p>
+ * The following graph shows a 3 segments droop curve. On the X-axis
+ * is the <code>AcDcConverter<code> DC voltage, on the Y-axis the droop coefficient.
```
jeandemanged:
```suggestion
 * is the <code>AcDcConverter</code> DC voltage, on the Y-axis the droop coefficient.
```

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DroopCurve.java
```diff
@@ -0,0 +1,59 @@
...
+ * +---------------------> V (MW)
+ * |              --------
+ * |
+ * |
+ * |
+ *</pre>
+ * @author Denis Bonnand {@literal <denis.bonnand at supergrid-institute.com>}
+ */
+public interface DroopCurve {
+
```
jeandemanged, line 38: Suggestion to define the empty curve here, to be used as initial curve for any AC/DC Converter

```suggestion

    DroopCurve EMPTY = new DroopCurve() {

        @Override
        public Collection<Segment> getSegments() {
            return List.of();
        }

        @Override
        public double getK(double v) {
            return 0;
        }
    };

```

jeandemanged:
> Hi @jeandemanged, thank you for your review, I made changes is the implementation of the droop curve as you mentioned, is it better like this ?
> 
> I am not sure to understand which functions I need to precise in the migration guide ? It is only those which are in an IIDM interface ?

Hi @2nisb , yes much better thank you. Yes changes in iidm-api must be summarized, you did good in the migration guide

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DroopCurveImpl.java
2nisb: For the below and above cases it makes sense, I agree. 
But for the empty curve isn't it more relevant to throw an error ?

jeandemanged: Throwing should remain exceptional once the IIDM is built. It is very okay to throw when adding via adders, or when modifying.
But it should be very hard to build an IIDM that throws systematically on simple "getting" operations. As it is now, a newly added converter would always throw when requesting droop value on an empty curve, so this needs to be fixed.
If you do not to return return zero, you would have to make sure that P_PCC_DROOP mode cannot be set on a converter with an empty curve, you would have to make sure that it is not possible to set an empty curve on a converter on P_PCC_DROOP mode, you will have to make sure that it is not possible to get droop value of a converter not in P_PCC_DROOP mode, and perhaps other scenarios I did not think of, and also considering both when creating (via adders) and when modifying converters. More complex to do....
You could go the complex route, but my take would be to just return 0 in the empty curve case. From you description of droop I think that functionally it makes sense ? with K=0 then Pdc=Pref, no droop ?

geofjamg: Or maybe just return an Optional ?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractAcDcConverter.java
```diff
@@ -334,5 +336,19 @@ public void remove() {
         super.remove();
     }
 
+    @Override
+    public DroopCurveAdderImpl newDroopCurve() {
+        return new DroopCurveAdderImpl(this);
+    }
```
jeandemanged, line 342: to silence sonar warning:

```suggestion
    @Override
    public DroopCurveAdder newDroopCurve() {
        return new DroopCurveAdderImpl<>(this);
    }
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DroopCurveImpl.java
jeandemanged: indeed, returning an optional would be fine for me

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractDroopCurveTest.java
```diff
@@ -0,0 +1,121 @@
...
+    public void invalidK() {
+        ValidationException e = assertThrows(ValidationException.class, () -> converter.newDroopCurve()
+                    .beginSegment()
+                        .setK(Double.NaN)
+                        .setMaxV(500.0)
+                        .setMinV(100.0)
+                    .endSegment()
+                .add());
+        assertTrue(e.getMessage().contains("k is not set"));
+    }
```
jeandemanged, line 61: to silence the sonar warnings, use this pattern here and all tests afterwards: first create an adder, and in the assertion only call the throwing endSegment() or add() method as relevant

```suggestion    @Test
    public void invalidK() {
        DroopCurveAdder.SegmentAdder segmentAdder = converter.newDroopCurve()
                .beginSegment()
                .setK(Double.NaN)
                .setMaxV(500.0)
                .setMinV(100.0);
        ValidationException e = assertThrows(ValidationException.class, segmentAdder::endSegment);
        assertTrue(e.getMessage().contains("k is not set"));
    }
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DroopCurveImpl.java
2nisb: Ok I understand, I will let 0 is empty curve case. I was thinking that it would be useful to alert the user that there is an error in the curve construction but returning k=0 is coherent with an empty droop.

jeandemanged (approved):

Issue #3607, opened 2025-10-14
PSS/E Importing duplicate elements


pjanecek0:
### Describe the current behavior

If the RAW file contains elements with the same ID, an error will occur (for example, with a bus) or their ID will be changed. 

### Describe the expected behavior

PowSyBl will behave like PSS/E, which updates the attributes of duplicate elements. In practice, it will use the last record.

PR #3643, opened 2025-11-06
Refactor PSS/E duplicated ID handling

pjanecek0:

#3607

Refactor PSS/E duplicated ID handling.

In the previous implementation, when a data record with the same ID was found, its ID was modified and a new element was created. This does not match PSS/E behavior, which in such cases updates the data of the existing element.

The new implementation uses a "last record wins" strategy - when a duplicate ID is encountered in the data, the last occurrence replaces all previous ones. 

- [ ] Yes
- [x] No

**Other information**:


psse/psse-converter/src/test/java/com/powsybl/psse/converter/PsseImporterIllinoisTest.java
marqueslanauja, line 115: I think this method is unused, you can remove it.

psse/psse-model/src/main/java/com/powsybl/psse/model/pf/PsseFixes.java
```diff
@@ -84,105 +86,22 @@ private String sortedBuses(int... buses) {
...
+            var id = idBuilder.apply(psseObject);
+            if (indexes.containsKey(id)) {
+                LOGGER.warn("Duplicated {} Id: {}", elementTypeName, id);
+                unique.set(indexes.get(id), psseObject);
+            } else {
+                indexes.put(id, unique.size());
+                unique.add(psseObject);
             }
-            foundIds.add(id);
         });
```
marqueslanauja, line 102: Your method is perfectly valid, but since the map values are never null, we can simplify it a bit to be faster. Here is my proposal:


```
private <T> List<T> fixDuplicatedIds(List<T> psseObjects, Function<T, String> idBuilder, String elementTypeName) {
        List<T> unique = new ArrayList<>();
        Map<String, Integer> indexes = new HashMap<>();

        psseObjects.forEach(psseObject -> {
            var id = idBuilder.apply(psseObject);
            Integer index = indexes.get(id);
            if (index != null) {
                LOGGER.warn("Duplicated {} Id: {}", elementTypeName, id);
                unique.set(index, psseObject);
            } else {
                indexes.put(id, unique.size());
                unique.add(psseObject);
            }
        });

        return unique;
    }

```

psse/psse-model/src/main/java/com/powsybl/psse/model/pf/PsseFixes.java
marqueslanauja, line 192: I thinl this method is not used, we can remove it

marqueslanauja:
I checked the performance by importing a large case, and I didn’t see any impact.

olperr1 (approved):

olperr1:
The remaining comments will be handled in an upcoming PR.

olperr1:
The PR dealing with the remaining comments is #3672.

PR #3646, opened 2025-11-07
Handle unsupported IIDM file versions with a better error message

### LINKED ISSUE #3553: Improve the error when importing files with right format but unsupported version

For example, if a file in IIDM version 1.14 is loaded using a version of powsybl that does not support it (< 6.8.0), the error obtained will be "Unsupported file format or invalid file."

We should, if possible, clarify the errors in these kinds of cases, to have something like: "Unsupported version XXX for format YYY, maximum supported version : ZZZ.", to separate the "real" format errors from the "version" errors.

### Describe the motivation

_No response_

### Extra Information

_No response_

cphili:

https://github.com/powsybl/powsybl-core/issues/3553

Make the error clearer when importing an IIDM file from an unsupported version.

Currently, the error message  "Unsupported file format or invalid file." is unclear.

The error message now states that version <IIDM file version> is not supported (maximum supported version: <IIDM_CURRENT_VERSION>).

- [ ] Yes
- [ ] No

**Other information**:


iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/IidmVersion.java
```diff
@@ -83,25 +81,34 @@ public String getXsd(boolean valid) {
...
-        String namespaceUriV = v.getNamespaceURI();
-        if (!namespaceURI.equals(namespaceUriV)) {
-            if (v.compareTo(V_1_7) >= 0 && namespaceURI.equals(v.getNamespaceURI(false))) {
-                return v;
-            }
-            throw new PowsyblException("Namespace " + namespaceURI + " is not supported. " +
-                    "The namespace for IIDM XML version " + v.toString(".") + " is: " + namespaceUriV + ".");
-        }
-        return v;
-    }
```
olperr1, line 99: Why did you remove this method?
It is a public static one, so this is a breaking change. We can of course remove it, but only if needed (for instance, if the method does no longer make sense).

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/IidmVersion.java
```diff
@@ -83,25 +81,34 @@ public String getXsd(boolean valid) {
...
-
     public static IidmVersion of(String version, String separator) {
         Objects.requireNonNull(version);
         return Stream.of(IidmVersion.values())
                 .filter(v -> version.equals(v.toString(separator)))
                 .findFirst() // there can only be 0 or exactly 1 match
                 .orElseThrow(() -> new PowsyblException("Version " + version + " is not supported."));
     }
+
+    public static int compareVersions(String v1, String v2) {
```
olperr1: Usually, we use `v1.compareTo(v2)` to compare versions. This method may not be needed (and therefore neither `parseVersion(...)`).

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/XMLImporter.java
```diff
@@ -58,32 +59,43 @@ public String getComment() {
...
+
+    private boolean isValidNetworkRoot(XMLStreamReader xmlsr) throws XMLStreamException {
+        while (xmlsr.hasNext()) {
+            if (xmlsr.next() == XMLStreamConstants.START_ELEMENT) {
+                String name = xmlsr.getLocalName();
+                String ns = xmlsr.getNamespaceURI();
+
+                if (!ns.isEmpty()) {
+                    String xmlIidmVersion = ns.substring(ns.lastIndexOf('/') + 1);
+                    if (IidmVersion.compareVersions(xmlIidmVersion, CURRENT_IIDM_VERSION.toString()) > 0) {
```
olperr1: Instead of comparing the strings of the 2 versions, it would be safer to try to load the version corresponding to the namespace, and if it fails to throw the exception with your message (in which you can also add the version that was encountered).

olperr1: You can load the IidmVersion corresponding to a string with `IidmVersion.of(String version, String separator)`.
It throws an exception containing `version` if no `IidmVersion` value is found (so the version is not supported). You can just catch this exception and create your exception (with your message untouched) using the caught exception as cause.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/XMLImporter.java
```diff
@@ -58,32 +59,43 @@ public String getComment() {
...
+                        throw new PowsyblException("Unsupported IIDM Version (maximum supported version: "
+                                + CURRENT_IIDM_VERSION.toString(".") + ")");
                     }
                 }
+
+                return NetworkSerDe.NETWORK_ROOT_ELEMENT_NAME.equals(name)
```
olperr1: I think you should return `false` if `!NetworkSerDe.NETWORK_ROOT_ELEMENT_NAME.equals(name)` before the `if (!ns.isEmpty())` block.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/XMLImporterTest.java
```diff
@@ -90,6 +92,19 @@ private void writeNetworkWithComment(String fileName) throws IOException {
...
+    private void writeNetworkUnsupportedIidmVersion(String fileName) throws IOException {
+        writeNetwork(fileName, unsupportedIidmVersionNamespaceURI(), false);
+    }
+
+    private String unsupportedIidmVersionNamespaceURI() {
+        String currentIidmVersion = CURRENT_IIDM_VERSION.toString();
+        String currentIidmNamespaceURI = CURRENT_IIDM_VERSION.getNamespaceURI();
+        int currentVersionMajor = Integer.parseInt(currentIidmVersion.split("_")[1]);
+        int currentVersionMinor = Integer.parseInt(currentIidmVersion.split("_")[2]);
+        String unsupportedIidmVersion = String.format("%d_%d", currentVersionMajor, ++currentVersionMinor);
```
olperr1: There's no need to increment `currentVersionMinor` here:
```suggestion
        String unsupportedIidmVersion = String.format("%d_%d", currentVersionMajor, currentVersionMinor + 1);
```

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/XMLImporterTest.java
```diff
@@ -90,6 +92,19 @@ private void writeNetworkWithComment(String fileName) throws IOException {
...
 
+    private void writeNetworkUnsupportedIidmVersion(String fileName) throws IOException {
+        writeNetwork(fileName, unsupportedIidmVersionNamespaceURI(), false);
+    }
+
+    private String unsupportedIidmVersionNamespaceURI() {
+        String currentIidmVersion = CURRENT_IIDM_VERSION.toString();
+        String currentIidmNamespaceURI = CURRENT_IIDM_VERSION.getNamespaceURI();
+        int currentVersionMajor = Integer.parseInt(currentIidmVersion.split("_")[1]);
+        int currentVersionMinor = Integer.parseInt(currentIidmVersion.split("_")[2]);
```
olperr1: `currentIidmVersion.split("_")` can be stored in a local variable to prevent computing it twice.

olperr1 (changes requested):

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/XMLImporterTest.java
```diff
@@ -154,6 +171,19 @@ void exists() {
         assertFalse(importer.exists(new DirectoryDataSource(fileSystem.getPath("/"), "testDummy"))); // namespace URI is not defined
     }
 
+    @Test
+    void testUnsupportedIidmVersion() {
+        String[] currentIidmVersionSplit = CURRENT_IIDM_VERSION.toString().split("_");
+        int unsupportedIidmVersionMajor = Integer.parseInt(currentIidmVersionSplit[1]);
+        int unsupportedIidmVersionMinor = Integer.parseInt(currentIidmVersionSplit[2]) + 1;
+        String unsupportedIidmVersion = String.format("%d.%d", unsupportedIidmVersionMajor, unsupportedIidmVersionMinor);
```
olperr1: This code could be extracted in a private static method to be reused by `unsupportedIidmVersionNamespaceURI()`.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/XMLImporter.java
```diff
@@ -58,32 +58,44 @@ public String getComment() {
...
+            } finally {
+                cleanClose(xmlsr);
+            }
+        } catch (XMLStreamException e) {
+            return false; // not a valid XML file
+        }
+    }
+
+    private boolean isValidNetworkRoot(XMLStreamReader xmlsr) throws XMLStreamException {
+        while (xmlsr.hasNext()) {
```
olperr1, line 88: The comment that previously existed was useful:

```suggestion
        // check the first root element is network and namespace is IIDM
        while (xmlsr.hasNext()) {
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/IidmVersion.java
```diff
@@ -83,6 +81,15 @@ public String getXsd(boolean valid) {
         return "iidm_equipment_V" + toString("_") + ".xsd";
     }
 
+    public static IidmVersion of(String version, String separator) {
+        Objects.requireNonNull(version);
+        return Stream.of(IidmVersion.values())
+                .filter(v -> version.equals(v.toString(separator)))
+                .findFirst() // there can only be 0 or exactly 1 match
+                .orElseThrow(() -> new PowsyblException("IIDM Version " + version + " is not supported. Max supported version: "
+                        + CURRENT_IIDM_VERSION.toString(".")));
```
olperr1:
```suggestion
                        + CURRENT_IIDM_VERSION.toString(separator)));
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/XMLImporter.java
```diff
@@ -58,32 +58,44 @@ public String getComment() {
...
+                String ns = xmlsr.getNamespaceURI();
+
+                if (!NetworkSerDe.NETWORK_ROOT_ELEMENT_NAME.equals(name)) {
+                    return false;
+                }
+
+                if (!ns.isEmpty()) {
+                    String version = ns.substring(ns.lastIndexOf('/') + 1);
+                    if (!version.isEmpty()) {
+                        IidmVersion.fromNamespaceURI(ns);
```
olperr1: - It would be better to use `IidmVersion.of(...)`.
- Comments should be added to explain the purpose of this line.

```suggestion
                        // Try to load an IidmVersion from the version number.
                        // If it is not supported, this will throw an exception giving details on the encountered version
                        // and the maximum supported version.
                        IidmVersion.of(version, "_");
```

olperr1: Another problem here. You could not throw a exception here without a minimum of checks.

When a file is imported, in `Importer.find(...)`, all the importer present in the class loader will be tested to see if one of them support the file format. I think that you can reach this line each time you import an XML file whose extension is `.xml` and whose first root element is `network`.
If you have a custom network file format with these characteristics, you may encountered the unsupported IIDM version exception depending on the order the importers are tested.

One control that can be added is on the known namespace prefixes. I suggest that you loop on `IidmVersion.values()` to retrieve their namespace prefixes (by this, I mean the part before the last`'/'`) and to check that the one of the current namespace is equal to one of them. If it is indeed, you can check the version and raise an exception if it is unknown. Else, you'll have to ignore the check.

olperr1 (changes requested):

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/XMLImporterTest.java
```diff
@@ -154,6 +173,15 @@ void exists() {
         assertFalse(importer.exists(new DirectoryDataSource(fileSystem.getPath("/"), "testDummy"))); // namespace URI is not defined
     }
 
+    @Test
+    void testUnsupportedIidmVersion() {
+        Path dir = fileSystem.getPath("/");
+        ReadOnlyDataSource dataSource = new DirectoryDataSource(dir, "test9");
+        String expectedError = "IIDM Version " + unsupportedIidmVersion() + "is not supported. Max supported version: "
+                + CURRENT_IIDM_VERSION.toString(".");
+        assertThrows(PowsyblException.class, () -> importer.exists(dataSource), expectedError);
```
samirromdhani: The 3rd argument of assertThrows is the failure message, it does not assert the exception message


```suggestion
        PowsyblException e = assertThrows(PowsyblException.class, () -> importer.exists(dataSource), "should throw exception");
        assertEquals(expectedError, e.getMessage());
```

or
```suggestion
        assertThatCode(() -> importer.exists(dataSource))
                .isInstanceOf(PowsyblException.class)
                .hasMessage(expectedError);
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/XMLImporter.java
```diff
@@ -58,32 +58,73 @@ public String getComment() {
...
+        int lastSlash = namespace.lastIndexOf('/');
+        return (lastSlash >= 0 && lastSlash < namespace.length() - 1) ? namespace.substring(lastSlash + 1) : "";
+    }
+
+    private boolean isValidPrefix(String prefix) {
+        return Stream.of(IidmVersion.values())
+                .map(v -> extractPrefix(v.getNamespaceURI(true)))
+                .anyMatch(prefix::equals)
+                || Stream.of(IidmVersion.values())
+                .filter(v -> v.compareTo(IidmVersion.V_1_7) >= 0)
```
samirromdhani: We can use `supportEquipmentValidationLevel` in both prefix and namespace check

```suggestion
                .filter(IidmVersion::supportEquipmentValidationLevel)
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/XMLImporter.java
```diff
@@ -58,32 +58,73 @@ public String getComment() {
...
+        }
+    }
+
+    private boolean isValidNetworkRoot(XMLStreamReader xmlsr) throws XMLStreamException {
+        while (xmlsr.hasNext()) {
+            if (xmlsr.next() == XMLStreamConstants.START_ELEMENT) {
+                String name = xmlsr.getLocalName();
+                String ns = xmlsr.getNamespaceURI();
+
+                if (!NetworkSerDe.NETWORK_ROOT_ELEMENT_NAME.equals(name) || ns.isEmpty()) {
```
samirromdhani: check NPE for ns

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/XMLImporter.java
```diff
@@ -58,32 +58,73 @@ public String getComment() {
...
+                    return false;
+                }
+
+                String currentPrefix = extractPrefix(ns);
+                if (isValidPrefix(currentPrefix)) {
+                    String version = extractVersion(ns);
+                    if (!version.isEmpty()) {
+                        // If it is not supported, this will throw an exception giving details on the encountered version
+                        // and the maximum supported version.
+                        IidmVersion.of(version, "_");
```
samirromdhani: I think exists() must stay boolean and not throw exception, as example, in this method all the XML errors are caught and mapped to false.
```java
catch (XMLStreamException e) {
       // not a valid xml file
     return false;
 }
```

samirromdhani: I think we should also check the version If it is not supported in `JsonImporter` when working with JIIDM format.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/JsonImporterTest.java
```diff
@@ -7,17 +7,66 @@
...
+        importer = new JsonImporter();
+        DataSource dataSource = new DirectoryDataSource(fileSystem.getPath("/"), "test-json-network.json");
+
+        String expectedError = "IIDM Version " + unsupportedIidmVersion()
+                + " is not supported. Max supported version: "
+                + CURRENT_IIDM_VERSION.toString(".");
+
+        PowsyblException exception = assertThrows(PowsyblException.class, () -> importer.exists(dataSource));
+        assertEquals(expectedError, exception.getMessage());
+    }
```
olperr1, line 110: To fix the unit tests:
```java
    private void createTempJson(String content, String basename) throws IOException {
        Path temp = tmpDir.resolve(basename + ".json");
        Files.writeString(temp, content);
    }

    @Test
    void testValidJsonVersionSupported() throws Exception {
        String json = "{ \"version\": \"1.15\" }";
        String basename = "supportedVersion";
        createTempJson(json, basename);
        importer = new JsonImporter();
        DataSource dataSource = new DirectoryDataSource(tmpDir, basename, "json", null);
        assertDoesNotThrow(() -> importer.exists(dataSource));
    }

    @Test
    void testInvalidVersion() throws Exception {
        String json = "{ \"version\": \"99.0\" }";
        String basename = "unsupportedVersion";
        createTempJson(json, basename);

        importer = new JsonImporter();
        DataSource dataSource = new DirectoryDataSource(tmpDir, basename, "json", null);

        String expectedError = "IIDM Version 99.0 is not supported. Max supported version: "
                + CURRENT_IIDM_VERSION.toString(".");

        PowsyblException exception = assertThrows(PowsyblException.class, () -> importer.exists(dataSource));
        assertEquals(expectedError, exception.getMessage());
    }
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/JsonImporter.java
```diff
@@ -45,20 +45,32 @@ public String getComment() {
...
+                String fieldName = parser.nextFieldName();
+                if (!JsonReader.VERSION_NAME.equals(fieldName)) {
+                    return false;
+                }
+                parser.nextToken();
+                String version = parser.getValueAsString();
+                if (version == null || version.isEmpty()) {
+                    return false;
                 }
+                IidmVersion.of(version, "_");
```
olperr1: In JSON, the separator for the version numbers is `.`:

```suggestion
                IidmVersion.of(version, ".");
```

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/JsonImporterTest.java
olperr1: Note that the separator must be changed in JsonImporter (see my other comment).

olperr1 (changes requested):

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/XMLImporter.java
samirromdhani: In Network **read** method there is two main actions: find an importer (1) and then import data (2)

- The `exists` method is called during the `find` action (1). It is important to keep this responsibility there (no exception should be thrown during the find action).
- `importData` (2) it can be the place to check the version is supported or not.

image

Here is an example  `importData` https://github.com/powsybl/powsybl-core/commit/ba9441c2bb749a3aa2a899348b7e5040e4e277f0

cphili: Thanks for the suggestions. I’ve applied your recommendations so that exists() remains focused solely on importer detection, while the version validation now takes place in importData and raises an exception if the version is not supported.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/JsonImporter.java
```diff
@@ -45,20 +51,60 @@ public String getComment() {
...
+            return false;
+        }
+    }
+
+    @Override
+    public Network importData(ReadOnlyDataSource dataSource, NetworkFactory networkFactory,
+                              Properties parameters, ReportNode reportNode) {
+        try {
+            String ext = findExtension(dataSource);
+            String version = readRootVersion(dataSource, ext);
```
samirromdhani: `dataSource` called twice by `findExtension` and `readRootVersion`, we can keep it one and `readRootVersion` can find itself the extension ;) 


```suggestion
            IidmVersion.of(readRootVersion(dataSource));
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/XMLImporter.java
```diff
@@ -58,32 +67,107 @@ public String getComment() {
...
+
+    @Override
+    public Network importData(ReadOnlyDataSource dataSource, NetworkFactory networkFactory,
+                              Properties parameters, ReportNode reportNode) {
+        try {
+            String ext = findExtension(dataSource);
+            String ns = readNamespaceURI(dataSource, ext);
+            if (ns != null) {
+                String version = extractVersion(ns);
+                IidmVersion.of(version);
```
samirromdhani: `dataSource` called twice by `findExtension` and `readNamespaceURI`, we can keep it one and `readNamespaceURI` can find itself the extension ;)

```suggestion
            Optional.ofNullable(readNamespaceURI(dataSource))
                    .map(ns -> IidmVersion.of(extractVersion(ns)));
```

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/XMLImporterTest.java
```diff
@@ -154,6 +175,31 @@ void exists() {
...
+        ReadOnlyDataSource dataSource = new DirectoryDataSource(dir, "test10");
+        assertDoesNotThrow(() -> importer.exists(dataSource));
+        assertTrue(importer.exists(dataSource));
+    }
+
+    @Test
+    void testImportDataUnsupportedIidmVersion() {
+        ReadOnlyDataSource dataSource = new DirectoryDataSource(fileSystem.getPath("/"), "test9");
+        NetworkFactory networkFactory = NetworkFactory.findDefault();
+        String expectedError = String.format("IIDM Version %s is not supported. Max supported version: %s", unsupportedIidmVersion(), CURRENT_IIDM_VERSION.toString("."));
```
samirromdhani, line 198: It’s usually better to keep the expected value as plain text in the test for readability.

```suggestion
String expectedError = "IIDM Version 1.17 is not supported. Max supported version: 1.16";
```

cphili: In this case, I think it’s clearer and safer to use unsupportedIidmVersion() so we compare directly against CURRENT_IIDM_VERSION without hard‑coding expected value and update when the version will change.

PR #3647, opened 2025-11-10
Optimize GeoJson parsing in iidm-geodata

rolnico:

No

Optimization

Parsing of GeoJson files can take a lot of memory.

Parsing is optimized and takes less memory.

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/docs/grid_exchange_formats/iidm/export.md
+++ b/docs/grid_exchange_formats/iidm/export.md
@@ -1,6 +1,49 @@
 # Export
 
-<span style="color: red">TODO</span>
+Networks can be exported in three IIDM formats:
+- XML: "XIIDM"
+- JSON: "JIIDM"
+- binary: "BIIDM"
+
+### Exporting with default values
+
+To export a network using the default parameters, use:
+```java
+Network n = ...;
+String FORMAT = "XIIDM"; // or "BIIDM" or "JIIDM"
+n.write(FORMAT, new Properties(), Path.of("/path/to/output.format"));
+```
+
+### Exporting with custom properties
+You can configure your export either with `ExportOptions` or with `Properties`.
+
+#### From ExportOptions
+
+```java
+Network n = ...;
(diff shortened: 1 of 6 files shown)


rolnico:
Here is the improvement, for a usual french network:
| Metric              | Current version | New version | Difference | Relative difference |
|---------------------|-----------------|-------------|------------|---------------------|
| CPU time (ms)       | 900             | 420         | −480       | −53.3 %             |
| Memory (MB)         | 1290            | 138.1       | −1151.9    | −89.3 %             |

rolnico:
> To be sure that I understood : was the excessive memory use coming from the fact that we used to create a complete map of the positions before using them to create the extensions, and now it's better because we do not store them and only use them directly for the extensions ?

Yes, we used to create the whole structure from the JSON file, which can be very memory-consuming, before using the data. Now it's stream-lined and directly used.

HugoKulesza (approved):

iidm/iidm-geodata/src/main/java/com/powsybl/iidm/geodata/geojson/GeoJsonDataAdder.java
```diff
@@ -1,28 +1,56 @@
...
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(GeoJsonDataAdder.class);
+
     protected GeoJsonDataAdder() {
     }
 
     public static void fillNetworkSubstationsGeoDataFromFile(Network network, Path path, boolean forceGeoDataComputation) throws IOException {
-        fillNetworkSubstationsGeoData(network, getSubstationsCoordinates(path), forceGeoDataComputation);
+        AtomicInteger substationsWithNewData = new AtomicInteger();
+        AtomicInteger substationsWithOldData = new AtomicInteger();
```
colineplqt: I wonder if it would not be better to call it something like substationsWithUpdatedData?

PR #3650, opened 2025-11-12
Fix CGMES taps import

olperr1:

No

Bug fix

At CGMES import, when the normal step (`cim:TapChanger.normalStep`) is outside the acceptable tap interval:
- if only the EQ file is imported: the step is defined as outside the interval (erroneous value);
- if the EQ and another file are imported: an exception is thrown: `2 windings transformer 'T2W': incorrect tap position 0 [1, 3]`

With 6.x PowSyBl versions, the tap was set to the neutral position and no exception was encountered.

When the normal step is defined outside the acceptable tap interval, it is set to the neutral position.

- [ ] Yes
- [X] No

**Other information**:

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/dc/DCConversion.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/dc/DCConversion.java
@@ -61,8 +61,13 @@ public class DCConversion {
 
     private void computeDcEquipments() {
         // Store the CGMES terminal to CGMES node association.
-        String node = context.nodeBreaker() ? DC_NODE : DC_TOPOLOGICAL_NODE;
-        cgmesDcTerminalNodes.forEach(t -> dcTerminalNodes.put(t.getId(DC_TERMINAL), t.getId(node)));
+        String nodeTag = context.nodeBreaker() ? DC_NODE : DC_TOPOLOGICAL_NODE;
+        cgmesDcTerminalNodes.forEach(t -> dcTerminalNodes.put(t.getId(DC_TERMINAL), t.getId(nodeTag)));
+        dcTerminalNodes.forEach((terminal, node) -> {
+            if (node == null) {
+                context.missing(String.format("Association to a DCNode or DCTerminalNode for ACDCConverterDCTerminal/DCTerminal %s", terminal));
+            }
+        });
 
         // Store the CGMES DCEquipment base data: id, type, node1, node2
         cgmesAcDcConverters.forEach(c -> addDcEquipment(c, ACDC_CONVERTER));
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/dc/DCEquipment.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/dc/DCEquipment.java
@@ -17,7 +17,7 @@ public record DCEquipment(String id, String type, String node1, String node2) {
 
     public boolean isAdjacentTo(DCEquipment otherDcEquipment) {
         // Two DCEquipment are adjacent if they share a node.
-        return node1.equals(otherDcEquipment.node1) || node1.equals(otherDcEquipment.node2)
+        return node1 != null && (node1.equals(otherDcEquipment.node1) || node1.equals(otherDcEquipment.node2))
                 || node2 != null && (node2.equals(otherDcEquipment.node1) || node2.equals(otherDcEquipment.node2));
     }
 
--- a/cgmes/cgmes-conversion/src/test/resources/issues/hvdc/invalid_DCConfiguration.xml
+++ b/cgmes/cgmes-conversion/src/test/resources/issues/hvdc/invalid_DCConfiguration.xml
@@ -204,7 +204,6 @@
     <cim:ACDCTerminal.sequenceNumber>2</cim:ACDCTerminal.sequenceNumber>
     <cim:ACDCConverterDCTerminal.polarity rdf:resource="http://iec.ch/TC57/2013/CIM-schema-cim16#DCPolarityKind.middle"/>
 <!--<cim:DCBaseTerminal.DCNode rdf:resource="#_DCN_2_2"/>-->
-    <cim:DCBaseTerminal.DCNode rdf:resource="#_INVALID_REF_1"/>
     <cim:ACDCConverterDCTerminal.DCConductingEquipment rdf:resource="#_VSC_2N"/>
 </cim:ACDCConverterDCTerminal>
 <cim:ACDCConverterDCTerminal rdf:ID="_T_VSC_2N_3">
@@ -268,7 +267,6 @@
     <cim:IdentifiedObject.name>Terminal DC line 12N 1</cim:IdentifiedObject.name>
     <cim:ACDCTerminal.sequenceNumber>1</cim:ACDCTerminal.sequenceNumber>
 <!--<cim:DCBaseTerminal.DCNode rdf:resource="#_DCN_1_3"/>-->
-    <cim:DCBaseTerminal.DCNode rdf:resource="#_INVALID_REF_2"/>
     <cim:DCTerminal.DCConductingEquipment rdf:resource="#_DCL_12N"/>
 </cim:DCTerminal>
 <cim:DCTerminal rdf:ID="_T_DCL_12N_2">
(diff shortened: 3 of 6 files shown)


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractTransformerConversion.java
```diff
@@ -65,7 +65,8 @@ protected static void setToIidmRatioTapChanger(TapChanger rtc, RatioTapChangerAd
                     .setG(g1)
                     .endStep();
         });
-        rtca.add();
+        RatioTapChanger tapChanger = rtca.add();
+        fixTapPosition(tapChanger);
```
rcourtier: That works, but I believe this isn't the _right_ place to do it.
I think this should be done in the `build()` method of `AbstractCgmesTapChangerBuilder` class instead.

Currently, this method is doing:
tap position = normal step if it exists, else neutral step

What the method should do is:
tap position = normal step if it exists **and is valid**, else neutral step if it exists **and is valid**, **else low step**

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractTransformerConversion.java
```diff
@@ -344,9 +346,16 @@ private static int findTapPosition(PropertyBag p, int defaultTapPosition) {
...
+        OptionalInt step = getNormalStep(tw, tapChangerId);
+        if (step.isPresent()) {
+            normalStep = step.getAsInt();
+            if (!isValidTapPosition(tapChanger, normalStep)) {
+                normalStep = null;
+            }
+        }
         OptionalInt neutralPosition = tapChanger.getNeutralPosition();
-        return getDefaultValue(normalStep.isPresent() ? normalStep.getAsInt() : null,
+        return getDefaultValue(normalStep,
```
rcourtier: I would rather have `normalStep` and `validNormalStep` as variable names. `step` is confusing, it could be misinterpretated as the current step position.
Also, I find it easier to read when the isPresent and isValid checks are made in the same `if` statement:

```suggestion
        Integer validNormalStep = null;
        OptionalInt normalStep = getNormalStep(tw, tapChangerId);
        if (normalStep.isPresent() && isValidTapPosition(tapChanger, normalStep.getAsInt())) {
            validNormalStep = normalStep.getAsInt();
        }
        OptionalInt neutralPosition = tapChanger.getNeutralPosition();
        return getDefaultValue(validNormalStep,
```

cgmes/cgmes-conversion/src/test/resources/update/transformer/transformer_erroneous_tap_EQ.xml
rcourtier, line 1: This is not a big deal at all, but we usually put the `EQ` or `SSH` profile as a suffix.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/TransformerUpdateTest.java
```diff
@@ -45,6 +45,25 @@ void importEqAndSshTogetherTest() {
...
     }
 
+    @Test
+    void importErroneousTapTest() {
+        Network network = readCgmesResources(DIR, "transformer_EQ_erroneous_tap.xml");
+        TwoWindingsTransformer twt = network.getTwoWindingsTransformerStream().findAny().orElseThrow();
+        assertEquals(1, twt.getPhaseTapChanger().getLowTapPosition());
+        assertEquals(3, twt.getPhaseTapChanger().getHighTapPosition());
+        assertEquals(2, twt.getPhaseTapChanger().getNeutralPosition().orElseThrow());
+        assertEquals(2, twt.getPhaseTapChanger().getTapPosition()); // The EQ tap position is outside the bounds, it is set to the neutral position
```
rcourtier: EQ tap position is a bit unclear, I suggest we use the actual EQ fields. 
```suggestion
        assertEquals(2, twt.getPhaseTapChanger().getTapPosition()); // The EQ normalStep is outside the bounds, it is set to the neutralStep
```

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/TransformerUpdateTest.java
```diff
@@ -45,6 +45,25 @@ void importEqAndSshTogetherTest() {
...
+
+        ThreeWindingsTransformer t3wt = network.getThreeWindingsTransformerStream().findAny().orElseThrow();
+        assertEquals(1, t3wt.getLeg2().getRatioTapChanger().getLowTapPosition());
+        assertEquals(15, t3wt.getLeg2().getRatioTapChanger().getHighTapPosition());
+        assertTrue(t3wt.getLeg2().getRatioTapChanger().getNeutralPosition().isEmpty()); // Neutral step is invalid => empty
```
rcourtier:
```suggestion
        assertTrue(t3wt.getLeg2().getRatioTapChanger().getNeutralPosition().isEmpty()); // The EQ neutralStep is invalid => empty
```

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/TransformerUpdateTest.java
```diff
@@ -45,6 +45,25 @@ void importEqAndSshTogetherTest() {
...
+        assertEquals(1, t3wt.getLeg2().getRatioTapChanger().getTapPosition()); // The EQ tap position and the neutral position are outside the bounds => set to the lowest tap position
```
rcourtier:
```suggestion
        assertEquals(1, t3wt.getLeg2().getRatioTapChanger().getTapPosition()); // The EQ normalStep and neutralStep are outside the bounds => set to the lowStep
```

rcourtier (changes requested):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractTransformerConversion.java
olperr1: I manually applied your suggestion (GitHub was failing to apply it from its site).

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractCgmesTapChangerBuilder.java
```diff
@@ -49,9 +49,13 @@ protected TapChanger build() {
         lowStep = p.asInt(CgmesNames.LOW_STEP);
         highStep = p.asInt(CgmesNames.HIGH_STEP);
         addSteps();
+        tapChanger.setLowTapPosition(lowStep);
+
         int neutralStep = p.asInt(CgmesNames.NEUTRAL_STEP);
-        int normalStep = p.asInt(CgmesNames.NORMAL_STEP, neutralStep);
-        tapChanger.setLowTapPosition(lowStep).setTapPosition(normalStep);
+        int validNeutralStep = isValidTapPosition(neutralStep) ? neutralStep : lowStep;
```
rcourtier: To be consistent with the SSH update, could you actually set the validNeutralStep in case it is invalid to the nearest valid step?

Something like:

```suggestion
        int validNeutralStep = isValidTapPosition(neutralStep) ? neutralStep : getClosestNeutralStep(neutralStep);
```

and:

```java
    private int getClosestNeutralStep(int neutralStep) {
        if (neutralStep < lowStep) {
            return lowStep;
        } else if (neutralStep > highStep) {
            return highStep;
        }
        return neutralStep;
    }
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractTransformerConversion.java
rcourtier: I actually changed the requirements, see below. The new requirement is:

What the method should do is:
tap position = normal step if it exists **and is valid**, else neutral step if it exists **and is valid**, **else closest valid neutral step**.

so that the EQ import is consistent with the SSH update.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractCgmesTapChangerBuilder.java
marqueslanauja: What about finding the step closest to ratio 1.0 and angle 0.0?
Here is my proposal:

```
private int getClosestNeutralStep() {
        if (tapChanger.getSteps().isEmpty()) {
            throw new PowsyblException("Unexpected tapChanger without steps: " + tapChanger.getId());
        }

        Comparator<TapChanger.Step> comparator =
                Comparator.comparingDouble((TapChanger.Step step) -> Math.abs(step.getRatio() - 1.0)) // closest ratio to 1.0
                        .thenComparingDouble(step -> Math.abs(step.getAngle())); // closest angle to 0.0

        TapChanger.Step bestStep = tapChanger.getSteps().stream()
                .min(comparator)
                .orElse(tapChanger.getSteps().getFirst());

        return tapChanger.getSteps().indexOf(bestStep);
    }


```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractTransformerConversion.java
```diff
@@ -344,9 +344,13 @@ private static int findTapPosition(PropertyBag p, int defaultTapPosition) {
     }
 
     private static <C extends Connectable<C>> int getDefaultTapPosition(Connectable<C> tw, com.powsybl.iidm.network.TapChanger<?, ?, ?, ?> tapChanger, String tapChangerId, int closestNeutralTapPosition, Context context) {
+        Integer validNormalStep = null;
         OptionalInt normalStep = getNormalStep(tw, tapChangerId);
+        if (normalStep.isPresent() && isValidTapPosition(tapChanger, normalStep.getAsInt())) {
+            validNormalStep = normalStep.getAsInt();
```
marqueslanauja, line 350: Maybe it isn’t necessary to validate the `normalStep` here. We are getting a default tap position during the update process using a validated `normalStep` previously recorded in the conversion process. An incorrect value only happens if the extension (where the `normalStep` is recorded) is deleted and created again with an invalid value. It doesn’t make much sense to do it.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractCgmesTapChangerBuilder.java
olperr1: I applied you suggestion @marqueslanauja: it is the closest to what is done during the SSH update.
I just fixed a little problem on the last line:

```diff
-return tapChanger.getSteps().indexOf(bestStep);
+return tapChanger.getSteps().indexOf(bestStep) + lowStep;
```

The steps are numbered from `lowStep`, which is not always `0`.

marqueslanauja: ok, thanks.

PR #3652, opened 2025-11-13
Fixing post-processors list in ImportConfig load

HugoKulesza:

No

Bug fix

When using `ImportConfig.load()`, the field `postProcessors` of `ImportConfig` is set to the `DEFAULT_POST_PROCESSORS` list without a copy. If `addPostProcessors` is then called on the `ImportConfig` instance, it will add the postprocessor to `DEFAULT_POST_PROCESSORS` which is not the desired  behaviour.

- [ ] Yes
- [x] No

### DIFF
--- a/cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/UpdateWithPostProcessorTest.java
+++ b/cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/UpdateWithPostProcessorTest.java
@@ -1,0 +1,57 @@
+/**
+ * Copyright (c) 2025, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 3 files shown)


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ImportConfig.java
```diff
@@ -37,7 +37,7 @@ public static ImportConfig load(PlatformConfig platformConfig) {
         List<String> postProcessors = platformConfig.getOptionalModuleConfig("import")
                 .flatMap(config -> config.getOptionalStringListProperty("postProcessors"))
                 .orElse(DEFAULT_POST_PROCESSORS);
-        return new ImportConfig(postProcessors);
+        return new ImportConfig(new ArrayList<>(postProcessors));
```
geofjamg: This is not the correct place for the copy. This should be in the ImportConfig constructor as a defensive copy. It is a bad practice to pass List to an object constructor that is part of an API without any copy.

geofjamg (approved):

PR #3653, opened 2025-11-13
Fix switch flow computation for SV export

rolnico:

No

Bug fix

During switch flow computation, the graph is transformed to a tree, which induce a loss of information: some switches and/or internal connections are missing from the results. Also, the results depends on how the tree is built, which can be dependent on the order of the edges in the subgraph.

Switch flow computation is done on the graph directly.

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SwitchesFlow.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SwitchesFlow.java
@@ -7,15 +7,27 @@
  */
 package com.powsybl.iidm.network.util;
 
-import com.powsybl.iidm.network.*;
+import com.powsybl.iidm.network.Bus;
+import com.powsybl.iidm.network.Switch;
+import com.powsybl.iidm.network.Terminal;
+import com.powsybl.iidm.network.TopologyKind;
+import com.powsybl.iidm.network.VoltageLevel;
 import org.jgrapht.alg.connectivity.ConnectivityInspector;
-import org.jgrapht.alg.interfaces.SpanningTreeAlgorithm.SpanningTree;
-import org.jgrapht.alg.spanning.KruskalMinimumSpanningTree;
 import org.jgrapht.graph.AsSubgraph;
 import org.jgrapht.graph.DefaultWeightedEdge;
 import org.jgrapht.graph.SimpleWeightedGraph;
 
-import java.util.*;
+import java.io.Serial;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
 
 /**
  * Utility class to compute the flow of the switches associated to a voltageLevel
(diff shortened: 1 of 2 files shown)


rolnico:
@zamarrenolm @marqueslanauja I think I found a bug here, or at least I don't really understand why the switches flow was computed as it was (by building a tree from the graph).
Since you worked on it previously (a few years ago), do you remember why you did it like this?
Could you have a look at my proposal here?

marqueslanauja:
> @zamarrenolm @marqueslanauja I think I found a bug here, or at least I don't really understand why the switches flow was computed as it was (by building a tree from the graph). Since you worked on it previously (a few years ago), do you remember why you did it like this? Could you have a look at my proposal here?

We did it intentionally because we believe this is the best way to solve the problem. When you have a meshed network with zero impedance in all the links (switches, in our case), you cannot determine the flow through the links that form a closed loop.

This is possible if the network is a tree. In that case, you can start from the lowest level and move upward until you reach the root, accumulating at each node the injection from its descendant nodes. By following this procedure, you can determine the flow in all links, and all nodes will be balanced.

So, if we have a meshed network, a general way to resolve the indeterminacy of link flows is to remove all loops by constructing the tree. Then, the flow in the links that belong to the tree can be calculated, while the flow in the links outside the tree must be zero. Of course, there isn’t a unique tree, but in my opinion, that’s not relevant , we just want to resolve the indeterminacy, and this is one way to do it. If we have the flow limits of the switches, we can minimize the flow limit of the tree by trying to include the more restrictive switches within the tree.

I've reviewed your proposal, and I believe that, in general, the results you obtain will not balance each node, which is mandatory.

zamarrenolm:
The problem is indeterminate, additional conditions must be imposed to solve it. Current selection is to assign maximum flow to some of the switches in the non-impedant graph and zero to the rest, while preserving injections/consumptions and balance at interior nodes. Be aware that if a different way of resolving the indeterminacy is selected, it should be applied, for consistency, to other points in the code:
- In Open Loadflow, in class `ZeroImpedanceFlows`.
- In core loadflow-results-completion, in class `Z0FlowsCompletion`.

marqueslanauja:
> > @zamarrenolm @marqueslanauja I think I found a bug here, or at least I don't really understand why the switches flow was computed as it was (by building a tree from the graph). Since you worked on it previously (a few years ago), do you remember why you did it like this? Could you have a look at my proposal here?
> 
> We did it intentionally because we believe this is the best way to solve the problem. When you have a meshed network with zero impedance in all the links (switches, in our case), you cannot determine the flow through the links that form a closed loop.
> 
> This is possible if the network is a tree. In that case, you can start from the lowest level and move upward until you reach the root, accumulating at each node the injection from its descendant nodes. By following this procedure, you can determine the flow in all links, and all nodes will be balanced.
> 
> So, if we have a meshed network, a general way to resolve the indeterminacy of link flows is to remove all loops by constructing the tree. Then, the flow in the links that belong to the tree can be calculated, while the flow in the links outside the tree must be zero. Of course, there isn’t a unique tree, but in my opinion, that’s not relevant , we just want to resolve the indeterminacy, and this is one way to do it. If we have the flow limits of the switches, we can minimize the flow limit of the tree by trying to include the more restrictive switches within the tree.
> 
> I've reviewed your proposal, and I believe that, in general, the results you obtain will not balance each node, which is mandatory.

To validate the balance of your proposal, I implemented a method, but it did not work properly because I forgot to include internal connections.  Sorry about that.
I re-evaluated your proposal using five small networks containing only loads and switches, without internal connections, and in all cases the balance was correct.

marqueslanauja:
Regarding the `CgmesExportTest.testFromIidm()` test, I reviewed it and believe the case is not a solved case. Some data are missing, so it is expected that `switchesFlow` will not balance at some nodes..

rolnico:
Let's not change how this is computed, since it's a usual method

PR #3443, opened 2025-05-14
Add Third-party licenses

rolnico:
**Please check if the PR fulfills these requirements**

- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] A PR or issue has been opened in all impacted repositories (if any)

**Does this PR already have an issue describing the problem?**
No

**What kind of change does this PR introduce?**
Feature

**What is the current behavior?**
Licenses are not provided in itools packaging

**What is the new behavior (if this is a feature change)?**
A first version of the `THIRD-PARTY.txt` file is introduced, according to the PR https://github.com/powsybl/powsybl-parent/pull/76
Licenses are now provided in itools packaging (`LICENSE.txt` + `THIRD-PARTY.txt`)

**Does this PR introduce a breaking change or deprecate an API?**
- [ ] Yes
- [x] No

**If yes, please check if the following requirements are fulfilled**

- [ ] The *Breaking Change* or *Deprecated* label has been added
- [ ] The migration steps are described in the following section

**What changes might users need to make in their application due to this PR? (migration steps)**

**Other information**:

PR #3654, opened 2025-11-13
iTools packager: license files and misc files

jeandemanged:

No

Bugfix & Feature

Rework of #3443.

- #3443 has issue of looking in parent folder, hence copying licences and 3rd party files does not work for example for powsybl-distribution where we encounter:
```
[INFO] Copy file /home/runner/work/powsybl-distribution/powsybl-distribution/resources/XIIDM/eurostag-tutorial-example1.xml to /home/runner/work/powsybl-distribution/powsybl-distribution/target/powsybl-distribution-2025.2.1/etc
Warning:  License file /home/runner/work/powsybl-distribution/LICENSE.txt not found
Warning:  License file /home/runner/work/powsybl-distribution/THIRD-PARTY.txt not found
[INFO] Zip package
```
- We use maven-plugin-testing-harness v3.3.0

- User can configure the itools packager with explicit file locations for license and third party files. licenseFile and thirdPartyFile are optional. If not provided, the packager will look for LICENSE.txt, LICENSE, THIRD-PARTY and THIRD-PARTY.txt in the project directory as well as its parent directory.
```xml
                <configuration>
                    <licenseFile>LICENSE.txt</licenseFile>
                    <thirdPartyFile>THIRD-PARTY.txt</thirdPartyFile>
                </configuration>
```
In case no file could be found, a log indicates all attempted locations for license and third party.

- User can configure the itools packager to copy any desired file via plugin configuration at the package root:
```xml
                <configuration>
                    <copyToPackageRoot>
                        <files>
                            <file>${project.basedir}/path/to/some/file.txt</file>
                        </files>
                    </copyToPackageRoot>
                </configuration>
```

- We use maven-plugin-testing-harness v3.4.0 making tests much cleaner and easier with junit-jupiter instead of junit-vintage-engine

- [ ] Yes

**Other information**:

### DIFF
--- a/itools-packager/src/main/java/com/powsybl/itools/ItoolsPackagerMojo.java
+++ b/itools-packager/src/main/java/com/powsybl/itools/ItoolsPackagerMojo.java
@@ -177,16 +177,29 @@ public class ItoolsPackagerMojo extends AbstractMojo {
     }
 
     private void addLicenseFiles(Path packageDir) {
-        // List of the license files to copy
-        Path projectRoot = Path.of(project.getBasedir().getPath());
+        Path projectRoot = project.getBasedir().toPath();
 
-        List<Path> candidateLicenseFiles = Stream.of(licenseFile, "LICENSE.txt", "../LICENSE.txt")
-                .filter(Objects::nonNull).map(projectRoot::resolve).toList();
+        List<Path> candidateLicenseFiles = getFilePathList(licenseFile, "LICENSE", projectRoot);
         addLicenseFile(packageDir, candidateLicenseFiles);
 
-        List<Path> candidateThirdPartyFiles = Stream.of(thirdPartyFile, "THIRD-PARTY.txt", "../THIRD-PARTY.txt")
-                .filter(Objects::nonNull).map(projectRoot::resolve).toList();
+        List<Path> candidateThirdPartyFiles = getFilePathList(thirdPartyFile, "THIRD-PARTY", projectRoot);
         addLicenseFile(packageDir, candidateThirdPartyFiles);
+    }
+
+    private List<Path> getFilePathList(String fileName, String defaultFileNameBase, Path projectRoot) {
+        if (fileName != null) {
+            return List.of(
+                    projectRoot.resolve(fileName),
+                    projectRoot.resolve("..").resolve(fileName)
+            );
+        }
+        // Search for default names
+        return Stream.of("", ".txt")
+                .flatMap(ext -> Stream.of(
+                        projectRoot.resolve(defaultFileNameBase + ext),
+                        projectRoot.resolve("..").resolve(defaultFileNameBase + ext)
+                ))
+                .toList();
     }
 
     private void addLicenseFile(Path packageDir, List<Path> candidateLicenseFiles) {
--- a/itools-packager/src/test/java/com/powsybl/itools/ItoolsPackagerMojoTest.java
+++ b/itools-packager/src/test/java/com/powsybl/itools/ItoolsPackagerMojoTest.java
@@ -129,7 +129,7 @@ class ItoolsPackagerMojoTest {
         mojo.execute();
         String target = project.getBuild().getDirectory();
         assertTrue(new File(target, DEFAULT_PACKAGE_NAME + ".zip").exists());
-        assertTrue(new File(target, DEFAULT_PACKAGE_NAME + "/LICENSE.txt").exists());
+        assertFalse(new File(target, DEFAULT_PACKAGE_NAME + "/LICENSE.txt").exists()); // wrong file specified in pom.xml
         assertTrue(new File(target, DEFAULT_PACKAGE_NAME + "/THIRD-PARTY.txt").exists());
     }
 
--- a/itools-packager/src/test/resources/test-maven-project-lic-parent-dir/submodule/pom.xml
+++ b/itools-packager/src/test/resources/test-maven-project-lic-parent-dir/submodule/pom.xml
@@ -31,7 +31,6 @@
                 </executions>
                 <configuration>
                     <licenseFile>myLic.txt</licenseFile>
-                    <thirdPartyFile>my3rdParties.txt</thirdPartyFile>
                 </configuration>
             </plugin>
         </plugins>
--- a/itools-packager/README.md
+++ b/itools-packager/README.md
@@ -82,5 +82,5 @@ Here is how to configure itools package Maven plugin in your project
 - packageType is optional, "zip" is the default value. It can be either zip or tgz.
 - javaXmx, mpiTasks and mpiHosts are used to generate itools.conf, are all optional and default values are respectively 8G, 2 and localhost.
 - additional binaries, libraries, configurations, package files can be added to the package using optional copyToBin, copyToLib, copyToEtc, copyToPackageRoot tags.
-- licenseFile and thirdPartyFile are optional. If not provided or not found, the packager will look for LICENSE.txt and THIRD-PARTY.txt in the project directory as well as its parent directory.
+- licenseFile and thirdPartyFile are optional. If not provided, the packager will look for LICENSE, LICENSE.txt, THIRD-PARTY, THIRD-PARTY.txt in the project directory as well as its parent directory.
 - all the jars with compile and runtime scope will be included in the package
--- a/itools-packager/src/test/resources/test-maven-project-lic-not-found/pom.xml
+++ b/itools-packager/src/test/resources/test-maven-project-lic-not-found/pom.xml
@@ -31,7 +31,6 @@
                 </executions>
                 <configuration>
                     <licenseFile>myLic.txt</licenseFile>
-                    <thirdPartyFile>my3rdParties.txt</thirdPartyFile>
                 </configuration>
             </plugin>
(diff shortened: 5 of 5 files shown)


rolnico:
I think that the issue here is more that the "child project" (powsybl-distribution in your example) does not have a `LICENSE.txt` and a `THIRD-PARTY.txt`. According to some licenses, we should provide those files as soon as we make an executable (for example), that's why I hardcoded the two files during the itools generation.
However: 

- adding the possibility to add more files is nice to have
- Fixing the paths of the files is also of course interesting (looking into the parent folder should be done only if the files are not found in the project directory level)

jeandemanged:
> I think that the issue here is more that the "child project" (powsybl-distribution in your example) does not have a `LICENSE.txt` and a `THIRD-PARTY.txt`. According to some licenses, we should provide those files as soon as we make an executable (for example), that's why I hardcoded the two files during the itools generation. However:
> 
> * adding the possibility to add more files is nice to have
> * Fixing the paths of the files is also of course interesting (looking into the parent folder should be done only if the files are not found in the project directory level)

Hi @rolnico 

The issue for powsybl-distribution is that current implementation looks in parent folder always, so even if these files were added, they would not be found at all.

The warning in GitLab CI is
```
Warning:  License file /home/runner/work/powsybl-distribution/LICENSE.txt not found
Warning:  License file /home/runner/work/powsybl-distribution/THIRD-PARTY.txt not found
```

But what can be added to powsybl-distribution is:
/home/runner/work/powsybl-distribution/powsybl-distribution/LICENSE.txt
not
/home/runner/work/powsybl-distribution/LICENSE.txt

rolnico:
> > I think that the issue here is more that the "child project" (powsybl-distribution in your example) does not have a `LICENSE.txt` and a `THIRD-PARTY.txt`. According to some licenses, we should provide those files as soon as we make an executable (for example), that's why I hardcoded the two files during the itools generation. However:
> > 
> > * adding the possibility to add more files is nice to have
> > * Fixing the paths of the files is also of course interesting (looking into the parent folder should be done only if the files are not found in the project directory level)
> 
> Hi @rolnico
> 
> The issue for powsybl-distribution is that current implementation looks in parent folder always, so even if these files were added, they would not be found at all.
> 
> The warning in GitLab CI is
> 
> ```
> Warning:  License file /home/runner/work/powsybl-distribution/LICENSE.txt not found
> Warning:  License file /home/runner/work/powsybl-distribution/THIRD-PARTY.txt not found
> ```
> 
> But what can be added to powsybl-distribution is: /home/runner/work/powsybl-distribution/powsybl-distribution/LICENSE.txt not /home/runner/work/powsybl-distribution/LICENSE.txt

Exactly, that's why I said that fixing the paths could be interesting.
What I'm wondering is: can we make adding these files a requirement while not providing fixed paths? Maybe having a specific parameter for each and, if it's not specified by the user, looking for the files in the standard directories (project + parent)?

itools-packager/src/test/resources/test-maven-project-lic-parent-dir/submodule/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+                    <execution>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>package-zip</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <licenseFile>myLic.txt</licenseFile>
+                    <thirdPartyFile>my3rdParties.txt</thirdPartyFile>
```
rolnico: In this case, the files do not exist, but the standard files exist in the parent folder and we'll use those files. Is that the expected behavior? Or should we throw an exception when files with the expected names are not found?

jeandemanged:
> ...
> ... Maybe having a specific parameter for each and, if it's not specified by the user, looking for the files in the standard directories (project + parent)?

Hi @rolnico , done as per your suggestion above, let me know

itools-packager/src/main/java/com/powsybl/itools/ItoolsPackagerMojo.java
```diff
@@ -171,28 +178,29 @@ private void writeItoolsConf(BufferedWriter writer) throws IOException {
 
     private void addLicenseFiles(Path packageDir) {
         // List of the license files to copy
-        List<String> licenseFiles = Arrays.asList("LICENSE.txt", "THIRD-PARTY.txt");
+        Path projectRoot = Path.of(project.getBasedir().getPath());
```
rolnico:
```suggestion
        Path projectRoot = project.getBasedir().toPath();
```

itools-packager/src/main/java/com/powsybl/itools/ItoolsPackagerMojo.java
```diff
@@ -171,28 +178,29 @@ private void writeItoolsConf(BufferedWriter writer) throws IOException {
...
 
-        // Get the root directory of the project
-        File projectRoot = project.getBasedir();
-        if (project.getParent() != null) {
-            projectRoot = projectRoot.getParentFile();
-        }
+        List<Path> candidateLicenseFiles = Stream.of(licenseFile, "LICENSE.txt", "../LICENSE.txt")
+                .filter(Objects::nonNull).map(projectRoot::resolve).toList();
```
rolnico:
```suggestion
        List<Path> candidateLicenseFiles = getFilePathList(licenseFile, "LICENSE", projectRoot);
```

itools-packager/src/main/java/com/powsybl/itools/ItoolsPackagerMojo.java
```diff
@@ -171,28 +178,29 @@ private void writeItoolsConf(BufferedWriter writer) throws IOException {
...
-                    Files.copy(sourcePath, packageDir.resolve(licenseFile), StandardCopyOption.REPLACE_EXISTING);
-                } catch (IOException e) {
-                    getLog().warn("Failed to copy license file " + sourcePath + ": " + e.getMessage());
-                }
-            } else {
-                getLog().warn("License file " + sourcePath + " not found");
+        List<Path> candidateThirdPartyFiles = Stream.of(thirdPartyFile, "THIRD-PARTY.txt", "../THIRD-PARTY.txt")
+                .filter(Objects::nonNull).map(projectRoot::resolve).toList();
+        addLicenseFile(packageDir, candidateThirdPartyFiles);
+    }
```
rolnico: I would suggest a change, to only look for default files if no license/3rd party file was given by the user:

```suggestion
        List<Path> candidateThirdPartyFiles = getFilePathList(thirdPartyFile, "THIRD-PARTY", projectRoot);
        addLicenseFile(packageDir, candidateThirdPartyFiles);
    }

    private List<Path> getFilePathList(String fileName, String defaultFileNameBase, Path projectRoot) {
        if (fileName != null) {
            return List.of(
                projectRoot.resolve(fileName),
                projectRoot.resolve("..").resolve(fileName)
            );
        }
        // Search for default names
        return Stream.of("", ".txt")
            .flatMap(ext -> Stream.of(
                projectRoot.resolve(defaultFileNameBase + ext),
                projectRoot.resolve("..").resolve(defaultFileNameBase + ext)
            ))
            .toList();
    }
```

itools-packager/src/test/java/com/powsybl/itools/ItoolsPackagerMojoTest.java
```diff
@@ -7,110 +7,130 @@
...
+    @InjectMojo(goal = "package-zip")
+    void testLicenseFilesInParent(ItoolsPackagerMojo mojo) {
         mojo.execute();
-        File target2 = new File("src/test/resources/test-maven-project-2/target/");
-        assertTrue(new File(target2, DEFAULT_PACKAGE_NAME + "/LICENSE.txt").exists());
-        assertFalse(new File(target2, DEFAULT_PACKAGE_NAME + "/THIRD-PARTY.txt").exists());
-        FileUtils.deleteDirectory(target2); // cleanup
+        String target = project.getBuild().getDirectory();
+        assertTrue(new File(target, DEFAULT_PACKAGE_NAME + ".zip").exists());
+        assertTrue(new File(target, DEFAULT_PACKAGE_NAME + "/LICENSE.txt").exists());
```
rolnico:
```suggestion
        assertFalse(new File(target, DEFAULT_PACKAGE_NAME + "/LICENSE.txt").exists());
```
if you change also in the test pom.xml

itools-packager/src/test/resources/test-maven-project-lic-parent-dir/submodule/pom.xml
rolnico:
```suggestion
```
remove this to test the case where only one file is specified (and since it does not exist, the test is modified too)

rolnico (changes requested):
I would make some changes: if the user specifies a license or 3rd party file, I think we should not look for a default file if the users file does not exist.
What do you think of that?

itools-packager/src/test/resources/test-maven-project-lic-parent-dir/submodule/pom.xml
jeandemanged: was intended, changed as you wanted.
you may change further if needed.

PR #3547, opened 2025-08-20
Choose topology at serialization at the voltage level scale

cphili:
**Please check if the PR fulfills these requirements**

- [X] The commit message follows our guidelines
- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been added / updated (for bug fixes / features)
- [ ] A PR or issue has been opened in all impacted repositories (if any)

**Does this PR already have an issue describing the problem?**
https://github.com/powsybl/powsybl-core/issues/3526

**What kind of change does this PR introduce?**

**What is the current behavior?**

**What is the new behavior (if this is a feature change)?**

**Does this PR introduce a breaking change or deprecate an API?**
- [ ] Yes
- [X] No

**If yes, please check if the following requirements are fulfilled**

- [ ] The *Breaking Change* or *Deprecated* label has been added
- [ ] The migration steps are described in the following section

**What changes might users need to make in their application due to this PR? (migration steps)**

**Other information**:

You can export a voltage level in a topology level  different from the general topology level defined by the `iidm.export.xml.topology-level` export property by using one of the following properties:

- `iidm.export.xml.topology-level.voltage-levels.node-breaker`
  - This property is an optional property that defines a list of voltage level IDs to be exported using the Node/Breaker view by the IIDM exporter. Note: if the voltage level is described using a lower detailed topology (Bus/Breaker or Bus/Branch), it will be exported using the Bus/Breaker (or Bus/Branch) view (Node/Breaker is not possible since the voltage level has not enough details).  
- `iidm.export.xml.topology-level.voltage-levels.bus-breaker`
  - This property is an optional property that defines a list of voltage level IDs to be exported using the Bus/Breaker view by the IIDM exporter.  
- `iidm.export.xml.topology-level.voltage-levels.bus-branch`
  - This property is an optional property that defines a list of voltage level IDs to be exported using the Bus/Branch view by the IIDM exporter.

PR #3655, opened 2025-11-14
Busbranch voltagelevel incompatibility

### LINKED ISSUE #3623: Previously exported BUS_BRANCH network with all switches open is not readable

When exporting a network with voltage level missing bus because all switches are open in BUS_BRANCH then the network is not readable afterwards. Indeed, a Generator::connectableBusId must not be null.

The user should be warned and can chose what to do when exporting a BUS_BRANCH network with voltage levels having all switches open that are consequently missing some buses because they do not have any Connectable connected on it.

An export option allows to chose whether export it anyway or not during BUS_BRANCH export.

If the network has all switches open then log with a warning and in a report that "Network with all switches open". 

If the export option value is to export anyway: then export, else do not export and throw an error with the log error message.

### Describe the steps

- import this NODE_BREAKER topology network 
- export it in BUS_BRANCH
- import it
- it throws an exception "connectable bus is not set"

[busBranchExportError.xml](https://github.com/user-attachments/files/23049771/busBranchExportError.xml)

### Environment

_No response_

### Relevant Log Output

_No response_

### Extra Information

If possible find a way to export the network with a possibility to read the network afterwards for a network with all switches open.

quinarygio:

Fixes #3623 

- [ ] Yes
- [ ] No

**Other information**:

The `iidm.export.xml.bus-branch.voltage-level.incompatibility-behavior` property is an optional property that defines the behavior of the XIIDM exporter when exporting a network in BUS_BRANCH topology with a voltage level missing bus because all switches are open.
There are two possible behaviors:
- `KEEP_ORIGINAL_TOPOLOGY`: the problematic voltage level is exported keeping the original topology and a warning is logged
- `THROW_EXCEPTION`: an exception is thrown when exporting a network in BUS_BRANCH topology with a voltage level missing bus because all switches are open
  By default, this behavior is set as `THROW_EXCEPTION`


iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/VoltageLevelSerDe.java
```diff
@@ -49,13 +50,22 @@ protected void writeRootElementAttributes(VoltageLevel vl, Container<? extends I
         context.getWriter().writeDoubleAttribute("highVoltageLimit", vl.getHighVoltageLimit());
 
         TopologyLevel topologyLevel = TopologyLevel.min(vl.getTopologyKind(), context.getOptions().getTopologyLevel());
-        context.getWriter().writeEnumAttribute("topologyKind", topologyLevel.getTopologyKind());
+        if (shouldExportwithOriginalTopology(vl, context)) {
+            context.getWriter().writeEnumAttribute("topologyKind", vl.getTopologyKind());
+        } else {
+            context.getWriter().writeEnumAttribute("topologyKind", topologyLevel.getTopologyKind());
+        }
```
alicecaron:
```suggestion
context.getWriter().writeEnumAttribute("topologyKind", shouldExportwithOriginalTopology(vl, context) ? vl.getTopologyKind() : topologyLevel.getTopologyKind());
```

olperr1:
Thanks @quinarygio!

While reviewing your PR with @alicecaron, we spotted an existing problem with #3547. For the connectables, nothing was done and the resulting export was inconsistent with "node" attributes on equipment connected to busses. Your solution to change the global topology level in the context works for generators and loads, but may not work with lines and transformers which are exported outside of the `<iidm:voltageLevel>` element.
I reworked an idea of @alicecaron to prepare a fix for #3547. It's in #3659.

I think that it would be better to solve the problem first, before resuming this PR.
If the idea of the fix is kept, the `TopologyLevelUtil.determineTopologyLevel(...)` method could be easily changed to take into account this PR's additions:
- the new tests can be added to this method;
- a cache can be introduced in `NetworkSerializerContext` to avoid recomputing the topology level for each voltage level every time the method is called.

@quinarygio: Can you make a review of #3659? Your opinion on it would be interesting.

quinarygio:
> Thanks @quinarygio!
> 
> While reviewing your PR with @alicecaron, we spotted an existing problem with #3547. For the connectables, nothing was done and the resulting export was inconsistent with "node" attributes on equipment connected to busses. Your solution to change the global topology level in the context works for generators and loads, but may not work with lines and transformers which are exported outside of the `<iidm:voltageLevel>` element. I reworked an idea of @alicecaron to prepare a fix for #3547. It's in #3659.
> 
> I think that it would be better to solve the problem first, before resuming this PR. If the idea of the fix is kept, the `TopologyLevelUtil.determineTopologyLevel(...)` method could be easily changed to take into account this PR's additions:
> 
> * the new tests can be added to this method;
> * a cache can be introduced in `NetworkSerializerContext` to avoid recomputing the topology level for each voltage level every time the method is called.
> 
> @quinarygio: Can you make a review of #3659? Your opinion on it would be interesting.

VoltageLevel topology level is already cached in ExportOptions (voltageLevelTopologyLevel HashMap)

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/util/TopologyLevelUtil.java
```diff
@@ -23,7 +27,23 @@ private TopologyLevelUtil() {
 
     public static TopologyLevel determineTopologyLevel(VoltageLevel vl, NetworkSerializerContext context) {
         TopologyLevel configTopologyLevel = Objects.requireNonNullElse(context.getOptions().getVoltageLevelTopologyLevel(vl.getId()), context.getOptions().getTopologyLevel());
-        return TopologyLevel.min(vl.getTopologyKind(), configTopologyLevel);
+        return shouldExportWithOriginalTopology(vl, context) ? TopologyLevel.valueOf(vl.getTopologyKind().name()) : TopologyLevel.min(vl.getTopologyKind(), configTopologyLevel);
```
olperr1: Here, we want to export the voltage level with in the `TopologyLevel.min(vl.getTopologyKind(), configTopologyLevel)` topology level. But it is not always possible.
I think the test should be reversed and `shouldExportWithOriginalTopology` may be changed to better reflect the meaning of this test. It is more:
```
if vl can be exported at topologyLevel
  - then: return topologyLevel
  - else: return originalTopologyLevel
```

**But more importantly,** your `shouldExportWithOriginalTopology` should test the topologyLevel that will be returned, not `TopologyLevel topologyLevel = TopologyLevel.min(vl.getTopologyKind(), context.getOptions().getTopologyLevel());`. So I think that the topologyLevel to test should be given as a parameter.

olperr1:
> VoltageLevel topology level is already cached in ExportOptions (voltageLevelTopologyLevel HashMap)

`voltageLevelTopologyLevel` in `ExportOptions` is not the result of your test, but the configuration defined by the user.
If he wants to export to export vl1 in bus/branch, `voltageLevelTopologyLevel` will have the entry `vl1 -> bus/branch`.

At the export, PowSyBl will test if the resulting bus/branch export will be valid. If not, it will export vl1 using its original topology level (for instance node/breaker).

To avoid recomputing the real topology level to use for the voltage level each time it will be referenced in the IIDM, we can cache `vl1 -> node/breaker` in the context.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/util/TopologyLevelUtil.java
```diff
@@ -23,7 +27,23 @@ private TopologyLevelUtil() {
...
+    }
+
+    private static boolean shouldExportWithOriginalTopology(VoltageLevel vl, NetworkSerializerContext context) {
```
alicecaron: Add javadoc

PR #3658, opened 2025-11-18
Add a ReportNode parameter to some JSON data extension deserialization methods

vbochetRTE:

No

Feature

The methods `ExtensionConfigLoader.load`, `ExtensionJsonSerializer.deserialize` and `ExtensionJsonSerializer.deserializeAndUpdate` do not take a `ReportNode` object as a parameter.
This prevents classes implementing these interfaces from using `ReportNode` in their processing.

The three methods listed above now have the possibility to take a `ReportNode` as a parameter if needed.

- [ ] Yes
- [x] No

**Other information**:


commons/src/main/java/com/powsybl/commons/json/JsonUtil.java
```diff
@@ -418,7 +454,7 @@ public static <T extends Extendable> Extension<T> readExtension(JsonParser parse
         Objects.requireNonNull(parser);
         Objects.requireNonNull(context);
         Objects.requireNonNull(supplier);
-        return updateExtension(parser, context, supplier::findProvider, extensionsNotFound, null);
+        return updateExtension(parser, context, supplier::findProvider, extensionsNotFound, null, ReportNode.NO_OP);
```
olperr1: I think that it would be more consistent to also have 2 `readExtension` methods with a `ReportNode` as a parameter (corresponding to this method and the previous one).
Methods with the existing signatures, i.e. without `ReportNode`, should also be available to avoid a breaking change.

Issue #3518, opened 2025-07-17
Make all IIDM objects properties bearers


alicecaron:
### Describe the current behavior

Not all the objects in IIDM can have properties.

### Describe the expected behavior

All the objects in IIDM can have properties, they all implements the `PropertiesHolder` interface that make it able to create/update/remove properties to the object.

### Describe the motivation

Be able to attach any property to any object in IIDM

olperr1:
I extracted from the 1.14 IIDM version's XSD all the complex types that do not yet support the properties.

# XSD complex types currently without property support
## Those for which there is nothing to do
For some of them, nothing should be done:

- Not an IIDM object => no extension support
```xsd
<xs:complexType name="Alias">
<xs:complexType name="FictitiousInjection">
```

- Not a real IIDM object - recomputed from the graph each time they are queried => no extension support
```xsd
<xs:complexType name="InternalConnection">
```

- References to existing IIDM objects => no extension support:
```xsd
<xs:complexType name="TerminalRef">
<xs:complexType name="VoltageLevelRef">
```


# Those for which the properties support should be added

Here are the one for which the properties support should be added. 
For each one, I indicated both the interface and the implementation(s) that should be modified.

- `Extension` / `AbstractExtension`:
```xsd
<xs:element name="extension" minOccurs="0" maxOccurs="unbounded">
```

- `TopologyModel` / `AbstractTopologyModel`:
```xsd
<xs:complexType name="NodeBreakerTopology">
<xs:complexType name="BusBreakerTopology">
```

- `ReactiveLimits` / `ReactiveCapabilityCurveImpl` & `MinMaxReactiveLimitsImpl`:
```xsd
<xs:complexType name="ReactiveCapabilityCurve">
<xs:complexType name="MinMaxReactiveLimits">
```

- `ReactiveCapabilityCurve.Point` / `ReactiveCapabilityCurveImpl.PointImpl`:
```xsd
<xs:complexType name="Point">
```

- `LoadModel` / `AbstractLoadModel`:
```xsd
<xs:complexType name="ZipLoadModel">
<xs:complexType name="ExponentialLoadModel">
```

- `ShuntCompensatorModel` / `ShuntCompensatorLinearModelImpl` & `ShuntCompensatorNonLinearModelImpl`:
```xsd
<xs:complexType name="ShuntLinearModel">
<xs:complexType name="ShuntNonLinearModel">
```

[...]

PR #3669, opened 2025-11-24
All iidm objects properties bearers

snoopfab:

Implements #3518 

- [ ] Yes
- [x] No

Custom IIDM implementation maintainers should change the implementations of some interfaces.
- The implementations of the following interfaces should also implement `PropertiesHolder`:
  - `AreaBoundary`
  - `LoadModel`
  - `OperationalLimits`
  - `OverloadManagementSystem.Tripping`
  - `ReactiveCapabilityCurve.Point`
  - `ReactiveLimits`
  - `ShuntCompensatorModel`
  - `ShuntCompensatorNonLinearModel.Section`
  - `TapChanger`
  - `TapChangerStep`
  - `TemporaryLimit`

- The implementations of the following interfaces should implement `BasePropertiesHolder`. If you don't already inherit from another class, you can just inherit from `AbstractBasePropertiesHolder` which already implement all the methods of the interface:
  - `ExponentialLoadModelAdder`
  - `OperationalLimitsAdder`
  - `OverloadManagementSystemAdder.TrippingAdder`
  - `ReactiveCapabilityCurveAdder`
  - `ReactiveCapabilityCurveAdder.PointAdder`
  - `ShuntCompensatorLinearModelAdder`
  - `ShuntCompensatorNonLinearModelAdder`
  - `ShuntCompensatorNonLinearModelAdder.SectionAdder`
  - `TapChangerAdder`
  - `TapChangerStepAdder`
  - `TemporaryLimitAdder`
  - `ZipLoadModelAdder`

TemporaryLimitAdder
> TemporaryLimitAdder<A> addProperty(String testProperty, String testTemporaryLimit);

For your adders' implementations, you can use `this.copyPropertiesTo(objectCreatedByTheAdder);` to copy the limits from the adder to the object created by the adder.

**Other information**:


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/OverloadImpl.java
```diff
@@ -42,6 +45,44 @@ public OverloadImpl(LoadingLimits.TemporaryLimit temporaryLimit, String previous
     }
 
     private static final LoadingLimits.TemporaryLimit UNACCEPTABLE_LIMIT = new LoadingLimits.TemporaryLimit() {
```
olperr1: This limit is a fictive one used only to report a constraint violation in a very specific case.
I think it would be better to not support properties on it. Therefore, you can throw an `UnsupportedOperationException` for the `setProperty` and `removeProperty` methods and return `null`/`false`/`Set.empty()` for the other methods.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/IdentifiableAdder.java
```diff
@@ -11,7 +11,7 @@
  *
  * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
  */
-public interface IdentifiableAdder<T extends Identifiable<T>, A extends IdentifiableAdder> {
+public interface IdentifiableAdder<T extends Identifiable<T>, A extends IdentifiableAdder> extends PropertiesHolder {
```
olperr1: Why is it needed for the `Identifiable` adder? Properties were already handled for `Identifiable`s.
Also, it doesn't seem to be necessary (compilation and tests work without having this class extending `PropertiesHolder`). Can you revert this change if it is not needed?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiableAdder.java
```diff
@@ -21,7 +21,7 @@
  *
  * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
  */
-abstract class AbstractIdentifiableAdder<T extends AbstractIdentifiableAdder<T>> implements Validable {
+abstract class AbstractIdentifiableAdder<T extends AbstractIdentifiableAdder<T>> extends AbstractPropertiesHolder implements Validable {
```
olperr1: This change doesn't seem to be necessary. Can you remove it?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractInjectionAdder.java
```diff
@@ -39,7 +39,7 @@ public T setConnectableBus(String connectableBus) {
     }
 
     @Override
-    protected NetworkImpl getNetwork() {
+    public NetworkImpl getNetwork() {
```
olperr1: Can you revert this change? It is not needed anymore.

```suggestion
    protected NetworkImpl getNetwork() {
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractLoadingLimitsAdder.java
```diff
@@ -42,6 +42,9 @@ public class TemporaryLimitAdderImpl<B extends LoadingLimitsAdder<L, B>> impleme
 
         private boolean ensureNameUnicity = false;
 
+        public TemporaryLimitAdderImpl() {
```
olperr1: This constructor is not needed. Can you remove it?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChanger.java
```diff
@@ -76,7 +76,7 @@ protected AbstractTapChanger(H parent,
 
     protected abstract RegulatingPoint createRegulatingPoint(int variantArraySize, boolean regulating);
 
-    protected NetworkImpl getNetwork() {
+    public NetworkImpl getNetwork() {
```
olperr1: This change is not needed anymore. Can you revert it?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTopologyModel.java
```diff
@@ -44,7 +44,7 @@ protected static int loadNodeIndexLimit(PlatformConfig platformConfig) {
                 .orElse(DEFAULT_NODE_INDEX_LIMIT);
     }
 
-    protected NetworkImpl getNetwork() {
+    public NetworkImpl getNetwork() {
```
olperr1: This change is not needed anymore. Can you revert it?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChangerAdderImpl.java
```diff
@@ -85,7 +82,7 @@ public A setLoadTapChangingCapabilities(boolean loadTapChangingCapabilities) {
         return self();
     }
 
-    NetworkImpl getNetwork() {
+    public NetworkImpl getNetwork() {
```
olperr1: This change is not needed anymore. Can you revert it?

```suggestion
    NetworkImpl getNetwork() {
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ActivePowerLimitsImpl.java
```diff
@@ -24,4 +25,9 @@ class ActivePowerLimitsImpl extends AbstractLoadingLimits<ActivePowerLimitsImpl>
     public void remove() {
         group.removeActivePowerLimits();
     }
+
+    @Override
+    public LimitType getLimitType() {
+        return ActivePowerLimits.super.getLimitType();
+    }
```
olperr1: This change may have been necessary at a certain point of your development, but it is not needed anymore.
Can you revert it?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/MinMaxReactiveLimitsAdderImpl.java
```diff
@@ -7,10 +7,7 @@
  */
 package com.powsybl.iidm.network.impl;
 
-import com.powsybl.iidm.network.MinMaxReactiveLimits;
-import com.powsybl.iidm.network.MinMaxReactiveLimitsAdder;
-import com.powsybl.iidm.network.Validable;
-import com.powsybl.iidm.network.ValidationException;
+import com.powsybl.iidm.network.*;
```
olperr1: Can you revert this change?
It is not necessary.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/OverloadManagementSystemImpl.java
```diff
@@ -108,11 +113,13 @@ public SwitchTripping setSwitchToOperateId(String switchToOperateId) {
             this.switchToOperateId = Objects.requireNonNull(switchToOperateId);
             return this;
         }
+
     }
 
     static class BranchTrippingImpl extends AbstractTrippingImpl implements OverloadManagementSystem.BranchTripping {
         private String branchToOperateId;
         private TwoSides side;
+        private Network network;
```
olperr1: Can you remove this attribute?
```suggestion
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/OverloadManagementSystemImpl.java
```diff
@@ -122,6 +129,11 @@ protected BranchTrippingImpl(String overloadManagementSystemId, String key, Stri
             setSideToOperate(side);
         }
 
+        @Override
+        public Type getType() {
+            return BranchTripping.super.getType();
+        }
+
```
olperr1: This change is not needed anymore.

```suggestion
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/OverloadManagementSystemImpl.java
```diff
@@ -98,6 +98,11 @@ public SwitchTrippingImpl(String overloadManagementSystemId, String key, String
             setSwitchToOperateId(switchToOperateId);
         }
 
+        @Override
+        public Type getType() {
+            return SwitchTripping.super.getType();
+        }
+
```
olperr1: This change is not needed anymore.

```suggestion
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/OverloadManagementSystemImpl.java
```diff
@@ -157,6 +169,12 @@ protected ThreeWindingsTransformerTrippingImpl(String overloadManagementSystemId
             super(overloadManagementSystemId, key, name, currentLimit, openAction);
             setThreeWindingsTransformerToOperateId(threeWindingsTransformerId);
             setSideToOperate(side);
+
+        }
+
+        @Override
+        public Type getType() {
+            return ThreeWindingsTransformerTripping.super.getType();
```
olperr1: Idem

```suggestion
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/PhaseTapChangerStepsReplacerImpl.java
```diff
@@ -29,6 +31,9 @@ class StepAdderImpl implements PhaseTapChangerStepsReplacer.StepAdder {
 
         private double b = 0.0;
 
+        public StepAdderImpl() {
+        }
+
```
olperr1: This is not needed

```suggestion
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/RatioTapChangerStepsReplacerImpl.java
```diff
@@ -27,6 +29,9 @@ class StepAdderImpl implements RatioTapChangerStepsReplacer.StepAdder {
 
         private double b = 0.0;
 
+        public StepAdderImpl() {
+        }
+
```
olperr1: This is not needed

```suggestion
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VoltageLevelImpl.java
```diff
@@ -26,7 +26,7 @@
  *
  * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
  */
-class VoltageLevelImpl extends AbstractIdentifiable<VoltageLevel> implements VoltageLevelExt {
+class VoltageLevelImpl extends AbstractIdentifiable<VoltageLevel> implements VoltageLevelExt, PropertiesHolder {
```
olperr1: `VoltageLevelImpl` already implements Identifiable (via `VoltageLevelExt`):

```suggestion
class VoltageLevelImpl extends AbstractIdentifiable<VoltageLevel> implements VoltageLevelExt {
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/LoadSerDe.java
```diff
@@ -46,6 +46,7 @@ protected void writeRootElementAttributes(Load l, VoltageLevel vl, NetworkSerial
 
     @Override
     protected void writeSubElements(Load load, VoltageLevel parent, NetworkSerializerContext context) {
+        //PropertiesSerDe.write(load, context);
```
olperr1:
```suggestion
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTapChangerAdderImpl.java
```diff
@@ -117,7 +114,9 @@ public T add() {
                 targetDeadband, network.getMinValidationLevel(), network.getReportNodeContext().getReportNode()));
 
         T tapChanger = createTapChanger(parent, lowTapPosition, steps, regulationTerminal, tapPosition, solvedTapPosition, regulating, loadTapChangingCapabilities, regulationValue, targetDeadband);
-
+        for (Map.Entry<Object, Object> z : getProperties().entrySet()) {
+            tapChanger.setProperty((String) z.getKey(), (String) z.getValue());
+        }
```
olperr1: Instead of repeating this code several times, you can add the following method to `PropertyHolder`:
```java

    /**
     * Copy the properties to another properties holder
     * @param propertiesHolder the destination properties holder
     */
    default void copyPropertiesTo(PropertiesHolder propertiesHolder) {
        getPropertyNames().forEach(name -> propertiesHolder.setProperty(name, getProperty(name)));
    }
```

and call this method:
```java
copyPropertiesTo(tapChanger);
```

security-analysis/security-analysis-api/pom.xml
```diff
@@ -59,6 +59,11 @@
             <artifactId>powsybl-iidm-api</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>powsybl-iidm-impl</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
olperr1: This dependency **should not** be added without the test scope here.
Otherwise, security analysis won't work anymore with custom IIDM implementations, such as the `powsybl-network-store` one.

olperr1: Reverting the scope change will lead to errors in `AbstractReducedLoadingLimits`.
I think that for this class and for its `ReducedTemporaryLimit` subclass, it is acceptable to not support properties. So I suggest to apply for them the same strategy as for `OverloadImpl.UNACCEPTABLE_LIMIT` (return null, empty set or false, or throw an UsupportedOperationException).

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/limitreduction/DefaultLimitReductionsApplier.java
```diff
@@ -9,10 +9,7 @@
 
 import com.powsybl.iidm.criteria.translation.DefaultNetworkElementAdapter;
 import com.powsybl.iidm.criteria.translation.NetworkElement;
-import com.powsybl.iidm.network.Identifiable;
-import com.powsybl.iidm.network.LimitType;
-import com.powsybl.iidm.network.LoadingLimits;
-import com.powsybl.iidm.network.ThreeSides;
+import com.powsybl.iidm.network.*;
```
olperr1: Can you revert this change?

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/limitreduction/result/AbstractReducedLoadingLimits.java
```diff
@@ -19,40 +20,59 @@
  * reduced limits without altering the real limits of the network element.</p>
  * @author Olivier Perrin {@literal <olivier.perrin at rte-france.com>}
  */
-public abstract class AbstractReducedLoadingLimits implements LoadingLimits {
+public abstract class AbstractReducedLoadingLimits extends AbstractPropertiesHolder implements LoadingLimits {
```
olperr1: See the comment on `pom.xml`:
No property support is needed for `AbstractReducedLoadingLimits` and `ReducedTemporaryLimit` subclass => same strategy as for OverloadImpl.UNACCEPTABLE_LIMIT

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/limitreduction/result/AbstractReducedLoadingLimits.java
```diff
@@ -19,40 +20,59 @@
...
     private final double permanentLimit;
     private final double originalPermanentLimit;
     private final double permanentLimitReduction;
     private final TreeMap<Integer, TemporaryLimit> temporaryLimits = new TreeMap<>(LoadingLimitsUtil.ACCEPTABLE_DURATION_COMPARATOR);
 
-    public record ReducedTemporaryLimit(String name, double value, int acceptableDuration, boolean fictitious,
-                                 double originalValue, double limitReduction) implements TemporaryLimit {
+    public static final class ReducedTemporaryLimit extends AbstractPropertiesHolder implements TemporaryLimit {
```
olperr1: You can then reuse a record here.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/NetworkSerDeTest.java
```diff
@@ -358,4 +361,303 @@ void emptySourceFormatTest() {
             assertEquals("", readNetwork.getSourceFormat());
         });
     }
+
```
olperr1, line 380: Can you move all these tests in `PropertiesSerDeTest`, which is maybe more suited?
You will have to change the visibility of `NetworkSerDeTest.writeAndRead` to package-protected at least.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/NetworkSerDeTest.java
```diff
@@ -358,4 +361,303 @@ void emptySourceFormatTest() {
...
+     *   [X]  LoadModel / AbstractLoadModel:
+     *   [X]  ShuntCompensatorModel / ShuntCompensatorLinearModelImpl & ShuntCompensatorNonLinearModelImpl:
+     *   [X]  ShuntCompensatorNonLinearModel.Section / ShuntCompensatorNonLinearModelImpl.SectionImpl:
+     *   [X]  OperationalLimits / AbstractLoadingLimits & AbstractReducedLoadingLimits:  (pas trouvé de serialisation iidm pour AbstractReducedLoadingLimits)
+     *   [X]  LoadingLimits / AbstractLoadingLimits.TemporaryLimitImpl & AbstractReducedLoadingLimits.ReducedTemporaryLimit:
+     *   [X]  OverloadManagementSystem.Tripping / OverloadManagementSystemImpl.AbstractTrippingImpl:
+     *   [X]  TapChangerStep / TapChangerStepImpl:
+     *   [X]  TapChanger / AbstractTapChanger:
+     *   [X]  AreaBoundary / AreaBoundaryImpl:
+     */
```
olperr1: This comment should be removed.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/NetworkSerDeTest.java
```diff
@@ -358,4 +361,303 @@ void emptySourceFormatTest() {
...
+            s.setProperty(TEST_PROPERTY, "valueNonLinearShuntCompensatorModelSection");
+        }
+
+        Area defaultControlArea = network.newArea().setId("defaultControlArea").setAreaType(TEST_PROPERTY).add();
+        defaultControlArea.setProperty(TEST_PROPERTY, "testValue");
+
+        BusbarSection bb = network.getBusbarSection("voltageLevel1BusbarSection1");
+        bb.setProperty(TEST_PROPERTY, "testBusbarSectionValue");
+
+        ExportOptions options = new ExportOptions().setVersion(IidmVersion.V_1_15.toString("."));
```
olperr1: In these tests, it would be better to not define the export IIDM version. By default, it will use the last one (for now, the 1.15), so the tests will always be performed on the last version, when newer ones will be introduced.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/NetworkSerDeTest.java
```diff
@@ -358,4 +361,303 @@ void emptySourceFormatTest() {
...
+                .beginTemporaryLimit()
+                .setValue(380)
+                .setAcceptableDuration(10 * 60)
+                .setName("10'")
+                .endTemporaryLimit();
+        adder.setProperty(TEST_PROPERTY, "testLoadingLimit");
+        return adder.add();
+    }
+
+    private static <L extends LoadingLimits, A extends LoadingLimitsAdder<L, A>> L createReducedLoadingLimits(Supplier<A> limitsAdderSupplier) {
```
olperr1: Can you remove this useless method?

olperr1 (changes requested):
Here are a few comments to begin with.
I still have files to review, but it could allow you to begin taking them into account.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/UnsupportedPropertiesHolder.java
```diff
@@ -0,0 +1,53 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+package com.powsybl.iidm.network.util;
+
+import com.powsybl.iidm.network.PropertiesHolder;
+
+import java.util.Set;
+
+/**
+ * Class used to pass Sonar Quality Gates Issues over code duplication in AbstractReducedLoadingLimits ...
```
olperr1: I think you can omit this comment:

```suggestion
```

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/UnsupportedPropertiesHolder.java
```diff
@@ -0,0 +1,53 @@
+/**
+ * Copyright (c) 2024, RTE (http://www.rte-france.com)
```
olperr1:
```suggestion
 * Copyright (c) 2025, RTE (http://www.rte-france.com)
```

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/UnsupportedPropertiesHolder.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+public class UnsupportedPropertiesHolder implements PropertiesHolder {
+    @Override
+    public boolean hasProperty() {
+        return false;
+    }
+
+    @Override
+    public boolean hasProperty(String key) {
+        throw new UnsupportedOperationException();
```
olperr1:
```suggestion
        return false;
```

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/UnsupportedPropertiesHolder.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    public String getProperty(String key) {
+        throw new UnsupportedOperationException();
```
olperr1:
```suggestion
        return null;
```

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/UnsupportedPropertiesHolder.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    public String getProperty(String key, String defaultValue) {
+        throw new UnsupportedOperationException();
```
olperr1:
```suggestion
        return null;
```

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/OverloadImpl.java
```diff
@@ -41,7 +41,7 @@ public OverloadImpl(LoadingLimits.TemporaryLimit temporaryLimit, String previous
         this.limitReductionCoefficient = limitReductionCoefficient;
     }
 
-    private static final LoadingLimits.TemporaryLimit UNACCEPTABLE_LIMIT = new LoadingLimits.TemporaryLimit() {
+    private static final class UnsupportedPropertiesHolderUnacceptableTemporaryLimit extends UnsupportedPropertiesHolder implements LoadingLimits.TemporaryLimit {
```
olperr1: Can you rename it into `UnacceptableTemporaryLimit`?
It would be shorter and there's no need to have this information in its name.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/VoltageLevelSerDe.java
```diff
@@ -320,6 +320,7 @@ protected void readSubElements(VoltageLevel vl, NetworkDeserializerContext conte
                 case VscConverterStationSerDe.ROOT_ELEMENT_NAME -> VscConverterStationSerDe.INSTANCE.read(vl, context);
                 case LccConverterStationSerDe.ROOT_ELEMENT_NAME -> LccConverterStationSerDe.INSTANCE.read(vl, context);
                 case GroundSerDe.ROOT_ELEMENT_NAME -> GroundSerDe.INSTANCE.read(vl, context);
+                case PropertiesSerDe.ROOT_ELEMENT_NAME -> PropertiesSerDe.read(vl, context);
```
olperr1: Properties are already handled by the default case:

```suggestion
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VoltageAngleLimitImpl.java
```diff
@@ -18,7 +18,7 @@
  * @author José Antonio Marqués {@literal <marquesja at aia.es>}
  * @author Bertrand Rix {@literal <bertrand.rix at artelys.com>}
  */
-class VoltageAngleLimitImpl implements VoltageAngleLimit {
+class VoltageAngleLimitImpl extends AbstractPropertiesHolder implements VoltageAngleLimit {
```
olperr1, line 21: Properties can be added on `VoltageAngleLimit`s, but they are not (de)serialized. Their support is missing in:
- `VoltageAngleLimitSerDe`
- the IIDM XSDs

iidm/iidm-serde/src/main/resources/xsd/iidm_equipment_V1_15.xsd
```diff
@@ -208,10 +208,14 @@
     </xs:complexType>
     <xs:complexType name="ReactiveCapabilityCurve">
         <xs:sequence>
+            <xs:element name="property" type="iidm:Property" minOccurs="0" maxOccurs="unbounded" />
             <xs:element name="point" minOccurs="2" maxOccurs="unbounded" type="iidm:Point"/>
```
olperr1, line 217: The properties support on `Point`s is handled in the SerDe, but not in the XSDs.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTransformerSerDe.java
```diff
@@ -97,11 +112,13 @@ protected static void writeRatioTapChanger(String name, RatioTapChanger rtc, Net
         });
         TerminalRefSerDe.writeTerminalRef(rtc.getRegulationTerminal(), context, ELEM_TERMINAL_REF);
 
+        IidmSerDeUtil.runFromMinimumVersion(IidmVersion.V_1_15, context, () -> PropertiesSerDe.write(rtc, context));
```
olperr1: Can you write the properties before the terminalRef?
XSDs should also be updated accordingly.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTransformerSerDe.java
```diff
@@ -208,12 +231,13 @@ protected static void writePhaseTapChanger(String name, PhaseTapChanger ptc, Net
         context.getWriter().writeDoubleAttribute(ATTR_REGULATION_VALUE, ptc.getRegulationValue());
         TerminalRefSerDe.writeTerminalRef(ptc.getRegulationTerminal(), context, ELEM_TERMINAL_REF);
 
+        IidmSerDeUtil.runFromMinimumVersion(IidmVersion.V_1_15, context, () -> PropertiesSerDe.write(ptc, context));
```
olperr1: Same here, can you write the properties before the terminalRef?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ExponentialLoadModelAdder.java
```diff
@@ -10,7 +10,7 @@
 /**
  * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
  */
-public interface ExponentialLoadModelAdder {
+public interface ExponentialLoadModelAdder extends PropertiesHolder {
```
olperr1: Having the PropertiesHolder adders is convenient for the deserialization, and even at the object creation via the API, but it may generate a lot of work to support for the custom IIDM implementations maintainers.
Since the deserialization relies on the fact that the adders can have properties, all custom IIDM implementations would have to implement properties support on the adders in order to read v1.15 IIDM networks.

To handle inner elements at deserialization, two mechanisms were introduced for Identifiables:
- in the case the identifiable can be created before reading its inner elements:
  - the SerDe extends `AbstractSimpleIdentifiableSerDe`
  - the inner elements are read/added to the identifiable after its creation
- in the other case
  - the SerDe extends `AbstractComplexIdentifiableSerDe`
  - when reading the inner elements, a task is memorized in a `List<Consumer<T>>`, where `T` is compatible with the type of the main Identifier to create. Evey task of this list will be later applied on the Identifier once it is created.

These interfaces are only applicable for Identifiers so they could not be used for the new PropertiesHolder objects, but having these 2 mechanisms in mind, it should be possible to have less modifications in the API.

Also note that with the current IIDM implementation, properties are created in the adder only to be copied after in the real object. This have a memory and time overhead.

olperr1 (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ExponentialLoadModelAdder.java
olperr1: Maybe we can think of a mechanism on the adder that allows to memorize tasks that will be automatically added once the object is created.
For instance, an interface for some adders that will have a method to add post-additional tasks that will be automatically applied on the object by the add() method.
If this interface/abstract class can be shared, the change for custom IIDM maintainers will be lighter (and it could be reused later for similar purposes).

olperr1:
Extensions and TopologyModel are not handled by this PR.
But this can maybe be done in another PR?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ExponentialLoadModelAdder.java
snoopfab: I agree on the performances overload (time and memory) of storing and setting back properties on adders . 
Your idea of a post creation hook that can be called is interresting but i won't have enough time on the subject this week to implement it.

snoopfab: I had the time to complete all the other enhancements above

olperr1: Thanks @snoopfab!

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTransformerSerDe.java
olperr1: The XSDs should be updated to have the properties before the terminalRef.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTransformerSerDe.java
olperr1: Same, the XSDs should be updated to have the properties before the terminalRef.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ExponentialLoadModelAdder.java
olperr1: I created a PR draft (#3692) with a possible solution in order to test it and see if it is easier for custom IIDM implementations maintainers to use it.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ConnectableSerDeUtil.java
```diff
@@ -235,38 +237,34 @@ static void readLoadingLimitsGroups(FlowsLimitsHolder h, String groupElementName
         readAllLoadingLimits(groupElementName, group, context);
     }
 
-    static void writeActivePowerLimits(Integer index, ActivePowerLimits limits, TreeDataWriter writer, IidmVersion version,
-                                              boolean valid, ExportOptions exportOptions) {
-        writeLoadingLimits(index, limits, writer, version.getNamespaceURI(valid), version, valid, exportOptions, ACTIVE_POWER_LIMITS);
+    static void writeActivePowerLimits(Integer index, ActivePowerLimits limits, NetworkSerializerContext context) {
+        writeLoadingLimits(index, limits, context, ACTIVE_POWER_LIMITS); // writer, version.getNamespaceURI(valid), version, valid, exportOptions
```
olperr1:
```suggestion
        writeLoadingLimits(index, limits, context, ACTIVE_POWER_LIMITS);
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/LoadSerDe.java
```diff
@@ -59,6 +59,7 @@ private void writeModel(LoadModel model, NetworkSerializerContext context) {
                 context.getWriter().writeStartNode(context.getVersion().getNamespaceURI(context.isValid()), EXPONENTIAL_MODEL);
                 context.getWriter().writeDoubleAttribute("np", expModel.getNp());
                 context.getWriter().writeDoubleAttribute("nq", expModel.getNq());
+                IidmSerDeUtil.runFromMinimumVersion(IidmVersion.V_1_15, context, () -> PropertiesSerDe.write(expModel, context));
```
olperr1: Reminder: all the `IidmSerDeUtil.runFromMinimumVersion(IidmVersion.V_1_15` instances introduced by this PR should be updated to `V_1_16` once the PR is rebased.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/PropertiesSerDe.java
```diff
@@ -57,4 +57,19 @@ private static <T extends PropertiesHolder> Consumer<T> read(NetworkDeserializer
 
     private PropertiesSerDe() {
     }
+
+    public static void readProperties(NetworkDeserializerContext context, PropertiesHolder holder) {
+        if (context.getVersion().compareTo(IidmVersion.V_1_15) <= 0) {
```
olperr1: Same here, this version number should be updated once the PR is rebased.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/PropertiesSerDe.java
```diff
@@ -57,4 +57,19 @@ private static <T extends PropertiesHolder> Consumer<T> read(NetworkDeserializer
...
+                if (elementName.equals(PropertiesSerDe.ROOT_ELEMENT_NAME)) {
+                    String name = context.getReader().readStringAttribute(NAME);
+                    String value = context.getReader().readStringAttribute(VALUE);
+                    context.getReader().readEndNode();
+                    holder.setProperty(name, value);
+                } else {
+                    throw new PowsyblException(String.format("Unknown element name '%s' in '%s'", elementName, holder.getClass().getSimpleName()));
+                }
+            });
+        }
```
olperr1, line 80: It is expected that this method reads the end node (see its usages).
It is the case for the latest IIDM version (via `readChildNodes`), but not for the older versions:

```suggestion
        } else {
            context.getReader().readEndNode();
        }
```

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/limitreduction/result/AbstractReducedLoadingLimits.java
```diff
@@ -19,40 +20,59 @@
...
+public abstract class AbstractReducedLoadingLimits extends UnsupportedPropertiesHolder implements LoadingLimits {
     private final double permanentLimit;
     private final double originalPermanentLimit;
     private final double permanentLimitReduction;
     private final TreeMap<Integer, TemporaryLimit> temporaryLimits = new TreeMap<>(LoadingLimitsUtil.ACCEPTABLE_DURATION_COMPARATOR);
 
-    public record ReducedTemporaryLimit(String name, double value, int acceptableDuration, boolean fictitious,
-                                 double originalValue, double limitReduction) implements TemporaryLimit {
+    public static final class ReducedTemporaryLimit extends UnsupportedPropertiesHolder implements TemporaryLimit {
+        // 1. Fields (equivalent to the record components)
```
olperr1: Since the record doesn't exist anymore, I think it would be easier to only have:
```suggestion
        // 1. Fields
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ConnectableSerDeUtil.java
```diff
@@ -235,38 +237,34 @@ static void readLoadingLimitsGroups(FlowsLimitsHolder h, String groupElementName
...
-                                          boolean valid, ExportOptions exportOptions) {
-        writeCurrentLimits(index, limits, writer, version.getNamespaceURI(valid), version, valid, exportOptions);
+    public static void writeCurrentLimits(Integer index, CurrentLimits limits, NetworkSerializerContext context) {
+        writeLoadingLimits(index, limits, context, CURRENT_LIMITS);
     }
 
-    public static void writeCurrentLimits(Integer index, CurrentLimits limits, TreeDataWriter writer, String nsUri, IidmVersion version,
-                                          boolean valid, ExportOptions exportOptions) {
-        writeLoadingLimits(index, limits, writer, nsUri, version, valid, exportOptions, CURRENT_LIMITS);
-    }
```
olperr1, line 321: This signature is not used in powsybl-core, but it may be used by custom extensions (where the namespace is not the one used for the IIDM version).
To avoid potentially problematic breaking changes, i think we should restore the previous signatures in this class (for the public methods, at least).

In `PropertiesSerDe`, the `write` method could be replaced by:
```java
    public static void write(PropertiesHolder propertiesHolder, NetworkSerializerContext context) {
        write(propertiesHolder, context.getWriter(), context.getVersion().getNamespaceURI(context.isValid()), context.getOptions());
    }

    public static void write(PropertiesHolder propertiesHolder, TreeDataWriter writer, String nsUri, ExportOptions exportOptions) {
        if (propertiesHolder.hasProperty()) {
            writer.writeStartNodes();
            for (String name : IidmSerDeUtil.sortedNames(propertiesHolder.getPropertyNames(), exportOptions)) {
                String value = propertiesHolder.getProperty(name);
                writer.writeStartNode(nsUri, ROOT_ELEMENT_NAME);
                writer.writeStringAttribute(NAME, name);
                writer.writeStringAttribute(VALUE, value);
                writer.writeEndNode();
            }
            writer.writeEndNodes();
        }
    }
```

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ExponentialLoadModelAdder.java
olperr1: I took a look at powsybl-network-store, and I am not convinced my PR is easier to manage for custom IIDM implementation maintainers than what you suggested, @snoopfab. In addition, your solution is simpler, so I think we can keep it like this. There's a little overload, but it should not be critical.
Sorry the delay on this point.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/PropertiesSerDe.java
```diff
@@ -57,4 +62,21 @@ private static <T extends PropertiesHolder> Consumer<T> read(NetworkDeserializer
 
     private PropertiesSerDe() {
     }
+
+    public static void readProperties(NetworkDeserializerContext context, PropertiesHolder holder) {
+        if (context.getVersion().compareTo(IidmVersion.V_1_16) <= 0) {
```
olperr1: There's an error in the sign:
```suggestion
        if (context.getVersion().compareTo(IidmVersion.V_1_16) >= 0) {
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ConnectableSerDeUtil.java
```diff
@@ -327,12 +329,14 @@ private static <L extends LoadingLimits> void writeLoadingLimits(Integer index,
...
             writer.writeDoubleAttribute("permanentLimit", limits.getPermanentLimit());
             writer.writeStartNodes();
+            IidmSerDeUtil.runFromMinimumVersion(IidmVersion.V_1_16, version, () -> PropertiesSerDe.write(limits, writer, nsUri, exportOptions));
             for (LoadingLimits.TemporaryLimit tl : IidmSerDeUtil.sortedTemporaryLimits(limits.getTemporaryLimits(), exportOptions)) {
                 writer.writeStartNode(version.getNamespaceURI(valid), TEMPORARY_LIMITS_ROOT_ELEMENT_NAME);
                 writer.writeStringAttribute("name", tl.getName());
                 writer.writeIntAttribute("acceptableDuration", tl.getAcceptableDuration(), Integer.MAX_VALUE);
                 writer.writeDoubleAttribute("value", tl.getValue(), Double.MAX_VALUE);
                 writer.writeBooleanAttribute("fictitious", tl.isFictitious(), false);
+                IidmSerDeUtil.runFromMinimumVersion(IidmVersion.V_1_15, version, () -> PropertiesSerDe.write(tl, writer, nsUri, exportOptions));
```
olperr1:
```suggestion
                IidmSerDeUtil.runFromMinimumVersion(IidmVersion.V_1_16, version, () -> PropertiesSerDe.write(tl, writer, nsUri, exportOptions));
```

iidm/iidm-serde/src/main/resources/xsd/iidm_equipment_V1_15.xsd
olperr1, line 1: These changes must be moved to the 1.16 XSDs. The 1.15 ones must not be modified.

iidm/iidm-serde/src/main/resources/xsd/iidm_V1_15.xsd
olperr1, line 1: Same with this file.

security-analysis/security-analysis-api/pom.xml
olperr1, line 1: Can you revert the changes of this file?

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/VoltageLevelSerDe.java
olperr1: There's no need to do anything for `VoltageLevel`s. They are `Identifiable`, and properties are already handled for all of them in `AbstractIdentifiableSerDe.write(...)`

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/VoltageLevelSerDe.java
```diff
@@ -363,6 +364,7 @@ private void readNodeBreakerTopology(VoltageLevel vl, NetworkDeserializerContext
         });
         context.getReader().readChildNodes(elementName -> {
             switch (elementName) {
+                case PropertiesSerDe.ROOT_ELEMENT_NAME -> PropertiesSerDe.read(vl, context);
```
olperr1: Can you remove this line?
In the current state of this PR, properties on TopologyModels are only partially supported. I think it is wiser to try to finish this PR without the support for the topology models, then to implement it in another PR once this PR is merged.

Note that with this line, the properties are added in the voltage level and not in its topology model.

Also, on a second thought, I wonder if properties on the topology model is really something we want...

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/TopologyModel.java
```diff
@@ -7,12 +7,14 @@
...
 
+import com.powsybl.iidm.network.PropertiesHolder;
+
 import java.util.function.Predicate;
 
 /**
  * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
  */
-interface TopologyModel {
+interface TopologyModel extends PropertiesHolder {
```
olperr1: Regarding my previous comment on TopologyModels, this change must be changed.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiableAdder.java
olperr1: With this change, the adders of all the identifiables should be changed to copy the properties from the adder to the identifiable in the `add()` method.
This would be really nice to have this feature, but it can be added in another PR (with the corresponding unit tests).
For this PR, i think this should be reverted.

olperr1 (changes requested):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/VoltageLevelAdderImpl.java
```diff
@@ -19,7 +19,7 @@
  *
  * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
  */
-class VoltageLevelAdderImpl extends AbstractIdentifiableAdder<VoltageLevelAdderImpl> implements VoltageLevelAdder {
+class VoltageLevelAdderImpl extends AbstractIdentifiableWithPropertiesAdder<VoltageLevelAdderImpl> implements VoltageLevelAdder {
```
olperr1: Why do you need this particular adder to have properties?
Is it because of the changes you've done in `VoltageLevelSerDe`? If I am not mistaken (but I may be wrong), there's no need to do anything in this SerDe to handle properties since they are already supported for VL.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/VoltageLevelSerDe.java
olperr1: @snoopfab: can you check if this `case PropertiesSerDe.ROOT_ELEMENT_NAME` is really needed ?
I am sure this is not the case.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/PropertiesSerDeTest.java
```diff
@@ -7,17 +7,298 @@
...
+        Network network2 = NetworkSerDeTest.writeAndRead(network, options);
+        DanglingLine dl2 = network2.getDanglingLine("DL");
+        assertEquals("valueActivePowerLimits", dl2.getActivePowerLimits().orElseThrow().getProperty(TEST_PROPERTY));
+        assertEquals("valueApparentPowerLimits", dl2.getApparentPowerLimits().orElseThrow().getProperty(TEST_PROPERTY));
+        assertEquals("valueCurrentLimits", dl2.getCurrentLimits().orElseThrow().getProperty(TEST_PROPERTY));
+        assertEquals("valueVoltageAngleLimit", network2.getVoltageAngleLimit("voltageAngleLimit").getProperty(TEST_PROPERTY));
+    }
+
+    private static <L extends LoadingLimits, A extends LoadingLimitsAdder<L, A>> L createLoadingLimits(Supplier<A> limitsAdderSupplier) {
+        A adder = limitsAdderSupplier.get()
```
olperr1, line 290: Temporary limits can also have properties. Can you add a tests on it?
I think something is missing in the temporary limits adder to handle them easily.

olperr1:
@snoopfab: Thanks for all this huge work!

Can you merge main in this branch and solve the conflict?
The read/write methods were moved to `AbstractShuntCompensatorSerDe`, so your changes in `ShuntSerDe` must be moved there.

After this, I will make some manual tests with the binary format (which have a special mechanism for optional values) and with the `powsybl-network-store` IIDM implementation.
But I think we are really close to finalize this PR now!

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/LoadingLimitsAdder.java
```diff
@@ -29,7 +29,10 @@ interface TemporaryLimitAdder<A> {
 
         TemporaryLimitAdder<A> ensureNameUnicity();
 
+        TemporaryLimitAdder<A> addProperty(String testProperty, String testTemporaryLimit);
```
olperr1:
```suggestion
        TemporaryLimitAdder<A> addProperty(String property, String value);
```

snoopfab:
@olperr1 : I've merged the origin/main branch and resolved the conflicts in ShuntSerDe following refactor.
You can process with your manual tests

olperr1:
@snoopfab: I think that having the adders implementing `PropertiesHolder` (for the interfaces) or extending `AbstractPropertyHolder` (for the classes) may be a little to heavy. There's a lot of methods that are not useful for the adders (such as `removeProperty(...)`).

What can be done is to use a new interface and a new abstract class for the adders. They should have only the necessary methods for the SerDes to work and could store the properties in a `Map<String, String>`. I think this would be easier to implement for custom IIDM implementations.

Therefore I suggest the following modifications:
- Split `PropertyHolder` in 2 interfaces:
  - `BasePropertyHolder`, with the methods necessary for the deserializers
    - `boolean hasProperty();`
    - `String getProperty(String key);`
    - `String setProperty(String key, String value);`
    - `Set<String> getPropertyNames();`
    - `void copyPropertiesTo(BasePropertiesHolder basePropertiesHolder)`, with its default implementation
  - `PropertyHolder`, inheriting from `BasePropertyHolder`, with the other methods
- Change the Adder interfaces extending from `PropertiesHolder` to extends `BasePropertyHolder` instead
- Create an abstract class implementing `BasePropertyHolder`, and implementing all its methods using an internal `Map<String, String>` to store the properties.
  - Since it is not used to store data in the IIDM model, **this class can be available in `iidm-api`**. This way, custom IIDM implementations could reuse it instead of defining their own.
- Change the Adder implementation classes to inherit from the new abstract class instead of `AbstractPropertiesHolder`
- Change the `PropertiesSerDe` methods to deal with `BasePropertyHolder` objects instead of `PropertyHolder` ones

snoopfab:
@olperr1 :
 
The requested changes have been implemented.
I've merged the branch with remote/main (caused a conflict on DanglingLine renamed to BoundaryLine in origin/main)
I've fixed the merge conflict. All tests are passing.

You now have the Ball.

Best Regards.

olperr1:
@snoopfab : Thanks a lot! I check this now.
Best regards

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/UnsupportedPropertiesHolder.java
```diff
@@ -0,0 +1,58 @@
+/**
+ * Copyright (c) 2025, RTE (http://www.rte-france.com)
```
olperr1:
```suggestion
 * Copyright (c) 2026, RTE (http://www.rte-france.com)
```

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/BasePropertiesHolder.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Get property associated to specified key.
+     */
+    String getProperty(String key);
+
+    /**
+     * Get property associated to specified key, with default value.
+     */
+    String getProperty(String key, String defaultValue);
```
olperr1: This method can be defined only in `PropertiesHolder`. It is not needed for adders.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/PropertiesHolder.java
```diff
@@ -13,7 +13,7 @@
 /**
  * @author Olivier Perrin {@literal <olivier.perrin at rte-france.com>}
  */
-public interface PropertiesHolder {
+public interface PropertiesHolder extends BasePropertiesHolder {
```
olperr1, line 14: There's no need to have in this interface the methods that are also defined in `BasePropertiesHolder`. Can you remove them?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/AbstractBasePropertiesHolder.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.Set;
+
+/**
+ * @author Fabrice Buscaylet {@literal <fabrice.buscaylet at artelys.com>}
+ */
+public abstract class AbstractBasePropertiesHolder implements BasePropertiesHolder {
+
+    protected final HashMap<String, String> properties = new LinkedHashMap<>();
```
olperr1: The type can be just `Map<String, String>:

```java
    protected final Map<String, String> properties = new LinkedHashMap<>();
```

(The imports should also be updated.)

olperr1:
@snoopfab:
It's approved! Thanks a lot for this long-term development!

Issue #3626, opened 2025-10-23
Evolution of the sensitivity analysis ResultWriter and Result to support multi-component computation, improve diagnostic and support cancel


vidaldid-rte:
### Describe the current behavior

The sensitivity analysis has currently some limitations.
     - No support of multi-component computation (no way to report if a base simulation succeeds on one component but fails on another)
     - No indication of the cause of a failure to compute a contingency, compared to what is reported for a loadflow run. It makes it hard to troubleshoot issues when results are missing for a contingency.
     - No indication whether the result is complete or only partial, if the operation has been canceled and the thread interrupted.

### Describe the expected behavior

To address these limiations, we propose to extend the ResultWriter as follow:

```
public interface SensitivityResultWriter {

   // No API change - meaning below in the case of multi-component computation
    // A factor has a variable that is linked to en equipement that is in a single component
    // Called once per factor and contingency in the factor context that belongs to the same component as the factor equipment
    // Nan for invalid factors whose equipement does not belongs to a component that has been run
    void writeSensitivityValue(int factorIndex, int contingencyIndex, double value, double functionReference);


    // 3 new parameters
    // Called for every contigency and every numCC/numCS where the contingency has an impact
    // For contingencies that are never run, called once in the end with no IMPACT and numCC and numCs set to -1
    void writeContingencyStatus(int contingencyIndex, SensitivityAnalysisResult.Status status, SensitivityAnalysisResult.LoadFlowStatus loadFlowStatus, int numCC, int numCs);


   // new method
    void writeSynchronousComponentStatus(int numCC, int numCS, SensitivityAnalysisResult.LoadFlowStatus loadFlowStatus);

    // Called at the end of the computation if the computation has not been interrupted
    void computationComplete();

    // Record to be added in the SensitivityAnalysisResult class
    record LoadFlowStatus (LoadFlowResult.ComponentResult.Status status, String statusText) {
    }
}
```

### Describe the motivation

Enable to add multi-component support for sensitivity analysis inOpenLoadFlow
Enable to return partial results in case of interruption
Enable a better diagnosis of contingency failure

snoopfab:
// For contingencies that are never run, called once in the end with no IMPACT and numCC and numCs set to -1

Why setting numCC and numCs to -1  instead of  existing  numCC and numCS ?

Moreover i'm not sure to understand when the third 'new method ' :

[...]

PR #3670, opened 2025-11-24
Add operator strategies to sensitivity analysis API

geofjamg:

No

Feature

In addition to N, post contingency states, we cant to be able to calculate sensitivity values on post operator strategy state.

- [x] Yes
- [ ] No

-----------------------------

## Purpose
With the previous PowSyBl versions, sensitivity analyses could be ran on N and post-contingency states.
Now, these analyses can also be ran to calculate sensitivity values on post operator strategy states.

### Generalities
In order to have access to the operator strategies related objects in the sensitivity analysis API, these objects were moved from the security analysis API to the contingency API.

### Package changes
The following classes were moved:

| Class | Old package | New package |
|---|---|---|
| `BusBreakerViolationLocation` | `com.powsybl.security` | `com.powsybl.contingency.violations` |
| `LimitViolation` | `com.powsybl.security` | `com.powsybl.contingency.violations` |
| `LimitViolationBuilder` | `com.powsybl.security` | `com.powsybl.contingency.violations` |
| `LimitViolationFilter` | `com.powsybl.security` | `com.powsybl.contingency.violations` |
| `LimitViolationHelper` | `com.powsybl.security` | `com.powsybl.contingency.violations` |
| `LimitViolationType` | `com.powsybl.security` | `com.powsybl.contingency.violations` |
| `LimitViolations` | `com.powsybl.security` | `com.powsybl.contingency.violations` |
| `LoadingLimitType` | `com.powsybl.security.detectors` | `com.powsybl.contingency.violations` |
| `NodeBreakerViolationLocation` | `com.powsybl.security` | `com.powsybl.contingency.violations` |
| `ViolationLocation` | `com.powsybl.security` | `com.powsybl.contingency.violations` |

The following packages were renamed:

| Old name | New name |
|---|---|
| `com.powsybl.security.condition` | `com.powsybl.contingency.strategy.condition` |
| `com.powsybl.security.json` | `com.powsybl.contingency.json` |
| `com.powsybl.security.strategy` | `com.powsybl.contingency.strategy` |

### Maven artifacts
All the classes whose package changed due to this refactoring were moved from the `powsybl-security-analysis-api` maven artifact to the `powsybl-contingency-api` one.
Since this latter depends on the first one, this change is transparent for all users. But if you only use classes from the contingency API, you may consider replacing the security analysis API dependency by the contingency API one.

Note also the following dependency changes:
- `powsybl-shortcircuit-api` no longer depends on `powsybl-security-analysis-api`, but now depends on `powsybl-contingency-api`.
- `powsybl-contingency-api` now depends on `powsybl-action-api`.

### Jackson databind modules
The (de)serialization of the following classes is now handled by the `ContingencyJsonModule` Jackson module instead of the `SecurityAnalysisJsonModule` and `DynamicSecurityAnalysisJsonModule` ones.
- `LimitViolation`
- `ViolationLocation`
- `OperatorStrategy`
- `OperatorStrategyList`
- `ConditionalActions`
- `Condition`

Since `SecurityAnalysisJsonModule` and `DynamicSecurityAnalysisJsonModule` both extend from `ContingencyJsonModule`, this change is transparent for all users. But if you only use classes from the contingency module, you may consider registering the contingency module instead of the security analysis one.

### `SensitivityAnalysisParameters` changes
There's no breaking change on this class since the additional data needed to run a sensitivity analysis on operator strategies are optional. By default, without any change in the parameters, the sensitivity analysis will operate as previously: on N and N-k states.
But since the parameters can have more data than before, their serialization format changes.

### New version for sensitivity analysis parameters files

For sensitivity analysis parameters files, a new version, the 1.2, was introduced. Differences with the previous version are:
- a new field `"operator-strategies-calculation-mode"` was added:
  - It represents the calculation mode used for operator strategies sensitivities.
  - Its possible values are:
    - "NONE" (default value): deactivate calculation of operator strategies sensitivities and only calculate N and N-K sensitivities;
    - "ALL_CONTINGENCIES": calculate operator strategies for all contingencies in addition to N, N-K sensitivities;
    - "ONLY_OPERATOR_STRATEGIES": only calculate operator strategies sensitivities and skip N and N-K sensitivities.

### Generalities
Instead of having results by contingency id, the results are now associated with a `SensitivityStateStatus`, which describe the state (pre-contingency, post-contingency _n_, operatorStrategy _m_). This have impacts on the creation / consultation of this object, but also on its (de)serialization.

#### Description
In `SensitivityAnalysisResult`:
- The `List<SensitivityContingencyStatus> contingencyStatuses` argument of the constructor was replaced by 3 arguments:
  - `List<SensitivityStateStatus> stateStatuses`
  - `List<String> contingencyIds`
  - `List<String> operatorStrategyIds`
- `getContingencyStatuses() ` was replaced by `getStateStatuses()`
- The `String contingencyId` argument was replaced by a `SensitivityState state` argument in the following methods:
  - `getSensitivityValue(...)`
  - `getValues(...)`
  - `getContingencyStatus(...)`
  - `getFunctionReferenceValue(...)`

#### Migration examples

`SensitivityAnalysisResult` object creation:
```patch
- SensitivityValue value1 = new SensitivityValue(0, 0, 1d, 2d);
- SensitivityValue value2 = new SensitivityValue(1, -1, 3d, 4d);
+ SensitivityValue value1 = new SensitivityValue(0, 0, -1, 1d, 2d);  // -1: no operator strategy
+ SensitivityValue value2 = new SensitivityValue(1, -1, -1, 3d, 4d);  // -1: no operator strategy


- List<SensitivityAnalysisResult.SensitivityContingencyStatus> contingencyStatus = new ArrayList<>();
- contingencyStatus.add(new SensitivityAnalysisResult.SensitivityContingencyStatus("NHV1_NHV2_2", SensitivityAnalysisResult.Status.SUCCESS));
- contingencyStatus.add(new SensitivityAnalysisResult.SensitivityContingencyStatus("NHV2_NHV3", SensitivityAnalysisResult.Status.NO_IMPACT));

+ List<SensitivityAnalysisResult.SensitivityStateStatus> stateStatuses = new ArrayList<>();
+ stateStatuses.add(new SensitivityAnalysisResult.SensitivityStateStatus(SensitivityState.postContingency("NHV1_NHV2_2"), SensitivityAnalysisResult.Status.SUCCESS));
+ stateStatuses.add(new SensitivityAnalysisResult.SensitivityStateStatus(SensitivityState.postContingency("NHV2_NHV3"), SensitivityAnalysisResult.Status.NO_IMPACT));
+ List<String> contingencyIds = List.of("NHV1_NHV2_2", "NHV2_NHV3");
+ List<String> operatorStrategyIds = Collections.emptyList();


List<SensitivityValue> values = List.of(value1, value2);
- SensitivityAnalysisResult result = new SensitivityAnalysisResult(factors, contingencyStatus, values);
+ SensitivityAnalysisResult result = new SensitivityAnalysisResult(factors, stateStatuses, contingencyIds, operatorStrategyIds, values);
```

`SensitivityAnalysisResult` object usage:
```patch
- result.getValues("NHV1_NHV2_2")
+ result.getValues(SensitivityState.postContingency("NHV1_NHV2_2"))

- result.getFunctionReferenceValue("NHV1_NHV2_2", "llll", SensitivityFunctionType.BRANCH_ACTIVE_POWER_1)
+ result.getFunctionReferenceValue(SensitivityState.postContingency("NHV1_NHV2_2"), "llll", SensitivityFunctionType.BRANCH_ACTIVE_POWER_1);

- result.getSensitivityValue("NHV1_NHV2_2", "g", "l1", SensitivityFunctionType.BRANCH_ACTIVE_POWER_1, SensitivityVariableType.INJECTION_ACTIVE_POWER);
+ result.getSensitivityValue(SensitivityState.postContingency("NHV1_NHV2_2"), "g", "l1", SensitivityFunctionType.BRANCH_ACTIVE_POWER_1, SensitivityVariableType.INJECTION_ACTIVE_POWER);

- result.getSensitivityValue(null, "g2", "l2", SensitivityFunctionType.BRANCH_ACTIVE_POWER_1, SensitivityVariableType.INJECTION_ACTIVE_POWER);
+ result.getSensitivityValue(SensitivityState.PRE_CONTINGENCY, "g2", "l2", SensitivityFunctionType.BRANCH_ACTIVE_POWER_1, SensitivityVariableType.INJECTION_ACTIVE_POWER)

- if (result.getContingencyStatus("NHV1_NHV2_2") == SensitivityAnalysisResult.Status.SUCCESS)
+ if (result.getStateStatus(new SensitivityState("NHV1_NHV2_2", null)) == SensitivityAnalysisResult.Status.SUCCESS)
```

### New version for sensitivity analysis result files
A new version was introduced for the JSON format:
- The version number is now "1.1"
- It contains a `stateStatus` instead of the `contingencyStatus` one.
- It contains the lists of the contingency ids (`contingencyIds`) and the operator strategy ids (`operatorStrategyIds`).

#### Impact on the code
The `SensitivityResultWriter` interface was slightly changed:
- All its methods now take an additional `int operatorStrategyIndex` argument.
  - Use `-1` when you don't have any operator strategy
- The `writeContingencyStatus` method was replaced by a `writeStateStatus` one

The following classes were renamed:

| Old Name | New name |
|---|---|
| `SensitivityAnalysisResult.SensitivityContingencyStatus` | `SensitivityAnalysisResult.SensitivityStateStatus` |
| `SensitivityContingencyStatusJsonSerializer` | `SensitivityStateStatusJsonSerializer` |
| `SensitivityContingencyStatusJsonDeserializer` | `SensitivityStateStatusJsonDeserializer` |

The following methods take additional parameters:
- `SensitivityResultModelWriter`'s contructor now takes a new `List<OperatorStrategy>)` argument;
- `SensitivityValue`'s contructor now takes a new `int operatorStrategyIndex` argument (= -1 when the value isn't related to an operator strategy);

## "sensitivity-analysis" iTools command line

- The `--output-contingency-file` option was replaced by a new `--output-state-status-file` option:
  - This reflect the sensitivity analysis results content's change.
- The command can take 2 new options:
  - `--operator-strategies-file`: the operator strategies input file path
  - `--actions-file`: the actions input file path

- The result file formats changed:
  - JSON:
    - It is now a structured object containing the following fields:
      - `sensitivityValues`: the list of the `SensitivityValue`s
      - `stateStatus`: the list of the `SensitivityStateStatus`es
      - `contingencyIds`: the list of the contigency ids
      - `operatorStrategyIds`: the list op the operator strategy ids
  - CSV:
    - The `Sensitivity analysis result` table has a new `Operator strategy ID` column.
    - The `Sensitivity analysis contingency status result` table:
      - was renamed `Sensitivity analysis status result`;
      - has a new `Status` column;
      - and its `Contingency Status` column was renamed `Operator strategy ID`.

<br>
<br>

**Other information**:

### DIFF
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/LoadingLimitsUtil.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/LoadingLimitsUtil.java
@@ -143,6 +143,7 @@ public final class LoadingLimitsUtil {
             groupToCopy.getCurrentLimits().ifPresent(limit -> copy.newCurrentLimits(limit).add());
             groupToCopy.getActivePowerLimits().ifPresent(limit -> copy.newActivePowerLimits(limit).add());
             groupToCopy.getApparentPowerLimits().ifPresent(limit -> copy.newApparentPowerLimits(limit).add());
+            groupToCopy.getPropertyNames().forEach(propertyName -> copy.setProperty(propertyName, groupToCopy.getProperty(propertyName)));
         });
     }
 }
--- a/iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractOperationalLimitsGroupsTest.java
+++ b/iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractOperationalLimitsGroupsTest.java
@@ -8,6 +8,7 @@
 package com.powsybl.iidm.network.tck;
 
 import com.powsybl.iidm.network.*;
+import com.powsybl.iidm.network.util.LoadingLimitsUtil;
 import org.junit.jupiter.api.Test;
 
 import java.util.Set;
@@ -19,7 +20,7 @@ import static org.junit.jupiter.api.Assertions.*;
  */
 public abstract class AbstractOperationalLimitsGroupsTest {
 
-    private static Network createNetworkWithLine() {
+    private static Network createNetworkWithLines() {
         Network network = Network.create("test", "test");
         Substation s1 = network.newSubstation().setId("S1").setCountry(Country.FR).add();
         VoltageLevel vl1 = s1.newVoltageLevel()
@@ -50,11 +51,27 @@ public abstract class AbstractOperationalLimitsGroupsTest {
                 .setB1(0.0)
                 .setB2(0.0)
                 .add();
+
+        network.newLine()
+                .setId("L2")
+                .setVoltageLevel1("VL1")
+                .setConnectableBus1("B1")
+                .setBus1("B1")
+                .setVoltageLevel2("VL2")
+                .setConnectableBus2("B2")
+                .setBus2("B2")
+                .setR(1.0)
+                .setX(1.0)
+                .setG1(0.0)
+                .setG2(0.0)
+                .setB1(0.0)
+                .setB2(0.0)
+                .add();
         return network;
     }
 
     private static Network createNetworkWithOperationalLimitsGroupsOnLine() {
-        Network network = createNetworkWithLine();
+        Network network = createNetworkWithLines();
         Line l = network.getLine("L");
         l.newOperationalLimitsGroup1("1")
                 .newCurrentLimits()
@@ -134,6 +151,8 @@ public abstract class AbstractOperationalLimitsGroupsTest {
         l.getOperationalLimitsGroup2("1")
                 .flatMap(OperationalLimitsGroup::getApparentPowerLimits)
                 .ifPresent(apl -> apl.setPermanentLimit(900.0));
+        l.getOperationalLimitsGroup1("3").orElseThrow().setProperty("propName1", "propValue1");
+        l.getOperationalLimitsGroup2("1").orElseThrow().setProperty("propName2", "propValue2");
         return network;
     }
 
@@ -522,5 +541,17 @@ public abstract class AbstractOperationalLimitsGroupsTest {
                 () -> assertFalse(group.hasProperty(property1)),
                 () -> assertEquals(Set.of(), group.getPropertyNames()),
                 () -> assertNull(group.getProperty(property1)));
+    }
+
+    @Test
+    public void testCopy() {
+        Network network = createNetworkWithOperationalLimitsGroupsOnLine();
+        Line line = network.getLine("L");
+        Line line2 = network.getLine("L2");
+        LoadingLimitsUtil.copyOperationalLimits(line, line2);
+        OperationalLimitsGroup olg3Side1 = line2.getOperationalLimitsGroup1("3").orElseThrow();
+        assertEquals("propValue1", olg3Side1.getProperty("propName1"));
+        OperationalLimitsGroup olg1Side2 = line2.getOperationalLimitsGroup2("1").orElseThrow();
+        assertEquals("propValue2", olg1Side2.getProperty("propName2"));
     }
 }
--- a/time-series/time-series-api/src/main/java/com/powsybl/timeseries/TimeSeriesTable.java
+++ b/time-series/time-series-api/src/main/java/com/powsybl/timeseries/TimeSeriesTable.java
@@ -54,7 +54,6 @@ import static com.powsybl.timeseries.TimeSeries.writeInstantToNanoString;
  * Some design considerations and limitations:
  * <ul>
  *     <li>Number of version loadable in the table has to be specified at creation</li>
- *     <li>Versions have to contiguous</li>
  *     <li>Once first batch of time series has been loaded, new time series cannot be added but data of existing one can be updated</li>
  *     <li>Concurrent load (i.e multi-thread) of data is supported (using same time series list)</li>
  *     <li>Concurrency between data loading and other operations (CSV writing, statistics computation) is NOT supported</li>
@@ -130,9 +129,7 @@ public class TimeSeriesTable {
         }
     }
 
-    private final int fromVersion;
-
-    private final int toVersion;
+    private final List<Integer> versions;
 
     private List<TimeSeriesMetadata> timeSeriesMetadata;
 
@@ -164,13 +161,18 @@ public class TimeSeriesTable {
     }
 
     public TimeSeriesTable(int fromVersion, int toVersion, TimeSeriesIndex tableIndex, IntFunction<ByteBuffer> byteBufferAllocator) {
-        TimeSeriesVersions.check(fromVersion);
-        TimeSeriesVersions.check(toVersion);
+        this(IntStream.range(fromVersion, toVersion + 1).boxed().toList(), tableIndex, byteBufferAllocator);
         if (toVersion < fromVersion) {
             throw new TimeSeriesException("toVersion (" + toVersion + ") is expected to be greater than fromVersion (" + fromVersion + ")");
         }
-        this.fromVersion = fromVersion;
-        this.toVersion = toVersion;
+    }
+
+    public TimeSeriesTable(List<Integer> versions, TimeSeriesIndex tableIndex, IntFunction<ByteBuffer> byteBufferAllocator) {
+        for (Integer version : versions) {
+            TimeSeriesVersions.check(version);
+        }
+
+        this.versions = versions.stream().sorted().toList();
         this.tableIndex = Objects.requireNonNull(tableIndex);
         this.byteBufferAllocator = Objects.requireNonNull(byteBufferAllocator);
     }
@@ -211,7 +213,7 @@ public class TimeSeriesTable {
                 throw new TimeSeriesException("None of the time series have a finite index");
             }
 
-            int versionCount = toVersion - fromVersion + 1;
+            int versionCount = versions.size();
 
             // allocate double buffer
             long doubleBufferSize = (long) versionCount * doubleTimeSeriesNames.size() * tableIndex.getPointCount();
@@ -265,16 +267,20 @@ public class TimeSeriesTable {
     }
 
     private long getTimeSeriesOffset(int version, int timeSeriesNum) {
-        return (long) timeSeriesNum * tableIndex.getPointCount() * (toVersion - fromVersion + 1) + (long) (version - fromVersion) * tableIndex.getPointCount();
+        return (long) timeSeriesNum * tableIndex.getPointCount() * versions.size() + (long) getVersionIndex(version) * tableIndex.getPointCount();
     }
 
     private int getStatisticsIndex(int version, int timeSeriesNum) {
-        return (version - fromVersion) * doubleTimeSeriesNames.size() + timeSeriesNum;
+        return getVersionIndex(version) * doubleTimeSeriesNames.size() + timeSeriesNum;
     }
 
-    private void checkVersionIsInRange(int version) {
-        if (version < fromVersion || version > toVersion) {
-            throw new IllegalArgumentException("Version is out of range [" + fromVersion + ", " + toVersion + "]");
+    private int getVersionIndex(int version) {
+        return versions.indexOf(version);
+    }
+
+    private void checkVersionIsInList(int version) {
+        if (!versions.contains(version)) {
+            throw new IllegalArgumentException("Version is out of the list " + versions);
         }
     }
 
@@ -318,7 +324,7 @@ public class TimeSeriesTable {
     }
 
     public void load(int version, List<TimeSeries> timeSeriesList) {
-        checkVersionIsInRange(version);
+        checkVersionIsInList(version);
         Objects.requireNonNull(timeSeriesList);
 
         if (timeSeriesList.isEmpty()) {
@@ -360,7 +366,7 @@ public class TimeSeriesTable {
     }
 
     public double getDoubleValue(int version, int timeSeriesNum, int point) {
-        checkVersionIsInRange(version);
+        checkVersionIsInList(version);
         int doubleTimeSeriesNum = checkTimeSeriesNum(timeSeriesNum);
         checkPoint(point);
         long timeSeriesOffset = getTimeSeriesOffset(version, doubleTimeSeriesNum);
@@ -368,7 +374,7 @@ public class TimeSeriesTable {
     }
 
     public String getStringValue(int version, int timeSeriesNum, int point) {
-        checkVersionIsInRange(version);
+        checkVersionIsInList(version);
         int stringTimeSeriesNum = checkTimeSeriesNum(timeSeriesNum);
         checkPoint(point);
         long timeSeriesOffset = getTimeSeriesOffset(version, stringTimeSeriesNum);
@@ -432,7 +438,7 @@ public class TimeSeriesTable {
     }
 
     private double getStatistics(int version, int timeSeriesNum, double[] stats) {
-        checkVersionIsInRange(version);
+        checkVersionIsInList(version);
         int doubleTimeSeriesNum = checkTimeSeriesNum(timeSeriesNum);
         int statisticsIndex = getStatisticsIndex(version, doubleTimeSeriesNum);
 
@@ -509,7 +515,7 @@ public class TimeSeriesTable {
 
     public double[] computePpmcc(String timeSeriesName, int version) {
         int timeSeriesNum1 = doubleTimeSeriesNames.getIndex(timeSeriesName);
-        checkVersionIsInRange(version);
+        checkVersionIsInList(version);
 
         Stopwatch stopWatch = Stopwatch.createStarted();
 
@@ -685,9 +691,8 @@ public class TimeSeriesTable {
                 CsvCache cache = new CsvCache();
 
                 // write data
-                for (int version = fromVersion; version <= toVersion; version++) {
+                for (int version : versions) {
                     for (int point = 0; point < tableIndex.getPointCount(); point += CsvCache.CACHE_SIZE) {
-
                         int cachedPoints = Math.min(CsvCache.CACHE_SIZE, tableIndex.getPointCount() - point);
 
                         // copy from doubleBuffer to cache
--- a/time-series/time-series-api/src/test/java/com/powsybl/timeseries/TimeSeriesTableTest.java
+++ b/time-series/time-series-api/src/test/java/com/powsybl/timeseries/TimeSeriesTableTest.java
@@ -8,10 +8,12 @@
 package com.powsybl.timeseries;
 
 import com.powsybl.timeseries.TimeSeries.TimeFormat;
+import org.assertj.core.api.Assertions;
 import org.junit.jupiter.api.Test;
 
 import java.io.BufferedReader;
 import java.io.StringReader;
+import java.nio.ByteBuffer;
 import java.time.Duration;
 import java.time.Instant;
 import java.time.ZoneId;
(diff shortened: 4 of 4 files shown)


sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityOperatorStrategiesCalculationMode.java
```diff
@@ -0,0 +1,17 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+package com.powsybl.sensitivity;
+
+/**
+ * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
+ */
+public enum SensitivityOperatorStrategiesCalculationMode {
```
SylvestreSakti, line 13: Just a suggestion as a user, to have a better understanding. What about making this a little bit more generic by having **Sensitivity~OperatorStrategies~CalculationMode** that can be :
- `ONLY_CONTINGENCIES` (default value)
- `CONTINGENCIES_AND_OPERATOR_STRATEGIES` (instead of current ALL_CONTINGENCIES)
- `ONLY_OPERATOR_STRATEGIES`

I find it more clear

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
```diff
@@ -137,24 +163,31 @@ private static void parseJson(JsonParser parser, SensitivityContingencyStatus.Pa
     /**
      * Sensitivity analysis result
      * @param factors the list of sensitivity factors that have been computed.
-     * @param contingencyStatuses the list of contingencies and their associated computation status.
+     * @param stateStatuses the list of contingencies and their associated computation status.
```
SylvestreSakti:
```suggestion
     * @param stateStatuses the list of states and their associated computation status.
```

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
```diff
@@ -188,11 +239,11 @@ public List<SensitivityValue> getValues() {
     /**
      * Get a list of sensitivity value associated to a given contingency id
      *
-     * @param contingencyId the ID of the considered contingency. Use null to get pre-contingency sensitivity values.
+     * @param state the considered state. Use null to get pre-contingency sensitivity values.
```
SylvestreSakti: In the comment: I am not sure we can use null to get pre-contingency as mentioned in (we should use SensitivityState.PRE_CONTINGENCY)

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
```diff
@@ -201,24 +252,29 @@ public List<SensitivityValue> getValues(String contingencyId) {
...
     public List<SensitivityValue> getPreContingencyValues() {
-        return valuesByContingencyId.getOrDefault(null, Collections.emptyList());
+        return valuesByState.getOrDefault(SensitivityState.PRE_CONTINGENCY, Collections.emptyList());
     }
 
     /**
      * Get the sensitivity value associated to a given function id and type and a given variable and for a specific contingency.
      *
-     * @param contingencyId the id of the considered contingency. Use null to get a pre-contingency sensitivity value.
+     * @param state the considered state. Use null to get a pre-contingency sensitivity value.
```
SylvestreSakti: Same: In the comment: I am not sure we can use null to get pre-contingency as mentioned in (we should use SensitivityState.PRE_CONTINGENCY)

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
```diff
@@ -231,7 +287,7 @@ public double getSensitivityValue(String contingencyId, String variableId, Strin
      * @return the sensitivity value associated with a given function and a given variable for a given contingency.
      */
     public double getBranchFlow1SensitivityValue(String contingencyId, String variableId, String functionId, SensitivityVariableType variableType) {
```
SylvestreSakti, line 294: Should all these methods getBranchFlow...() be implemented also with sensitivityStates instead of contingencyId ?

SylvestreSakti (commented):
Nice update ! Just for information, there may be some work (regarding which PR is merged first) to make it compatible with https://github.com/powsybl/powsybl-core/issues/3626 (having ComponentResults for the status)

contingency/contingency-api/src/test/java/com/powsybl/contingency/json/JsonActionAndOperatorStrategyTest.java
```diff
@@ -143,11 +145,10 @@ public Action build() {
 
     @Test
     void testJsonPlugins() throws JsonProcessingException {
-        Module jsonModule = new SimpleModule()
-            .registerSubtypes(DummyAction.class, DummyActionBuilder.class);
-        SecurityAnalysisJsonPlugin plugin = () -> List.of(jsonModule);
```
olperr1, line 148: This test's purpose was to check that `SecurityAnalysisJsonPlugin`s were supported.
I don't think it should be changed, but restored in its previous state and moved in `com.powsybl.security.json`.

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
```diff
@@ -392,22 +448,25 @@ public double getBranchCurrent3SensitivityValue(String variableId, String functi
...
     public double getBusVoltageSensitivityValue(String variableId, String functionId, SensitivityVariableType variableType) {
-        return getSensitivityValue(null, variableId, functionId, SensitivityFunctionType.BUS_VOLTAGE, variableType);
+        return getSensitivityValue(SensitivityState.PRE_CONTINGENCY, variableId, functionId, SensitivityFunctionType.BUS_VOLTAGE, variableType);
     }
 
     /**
      * Get the function reference associated to a given contingency Id and a given function id and type.
      *
-     * @param contingencyId the id of the considered contingency. Use null to get a pre-contingency function reference value.
+     * @param state the considered state. Use null to get a pre-contingency function reference value.
```
olperr1: Same here, "state" cannot be null (`requireNonNull(state)`) => use `SensitivityState.PRE_CONTINGENCY`

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
```diff
@@ -172,8 +205,26 @@ public List<SensitivityFactor> getFactors() {
      *
      * @return a list of all the contingency statuses.
```
olperr1:
```suggestion
     * @return a list of all the state statuses.
```

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityState.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
+ */
+public record SensitivityState(String contingencyId, String operatorStrategyId) {
+
+    public static final SensitivityState PRE_CONTINGENCY = new SensitivityState(null, null);
+
+    public static SensitivityState postContingency(String contingencyId) {
+        return new SensitivityState(contingencyId, null);
```
olperr1, line 18: Can you check that `contingencyId` is not null here?

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/json/SensitivityAnalysisResultDeserializer.java
```diff
@@ -52,18 +58,43 @@ public SensitivityAnalysisResult deserialize(JsonParser parser, DeserializationC
...
-        if (version == null || !version.equals("1.0")) {
-            //Only 1.0 version is supported for now
-            throw new IllegalStateException("Version different than 1.0 not supported.");
+        if (version == null || !version.equals("1.0") && !version.equals("1.1")) {
+            throw new IllegalStateException("Only version 1.0 and 1.1 are supported.");
+        }
+        if ("1.0".equals(version)) {
+            // in 1.0 the contingency IDs and the mapping contingency index -> ID was taken from 'contingencyStatus' list
+            // not possible anymore when >= 1.1 because the new 'stateStatus' can contain also some post operator strategy
+            // statuses and are not indexed in contingencies
```
olperr1:
```suggestion
            // In 1.0 the contingency IDs and the mapping contingency index -> ID were directly taken from 'contingencyStatus' list.
            // Therefore for this version, the "contingencyIds" tag was not encountered.
            // Note that these elements cannot be computed from the state statuses (for versions >= 1.1) because 
            // they also contain some post operator strategy statuses and are not indexed by contingency
```

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityResultJsonWriter.java
```diff
@@ -56,8 +63,8 @@ public void close() {
             jsonGenerator.writeEndArray();
             //Write buffered contingency status at the end
             jsonGenerator.writeStartArray();
-            for (SensitivityAnalysisResult.SensitivityContingencyStatus status : contingencyStatusBuffer) {
-                SensitivityAnalysisResult.SensitivityContingencyStatus.writeJson(jsonGenerator, status);
+            for (SensitivityAnalysisResult.SensitivityStateStatus status : contingencyStatusBuffer) {
+                SensitivityAnalysisResult.SensitivityStateStatus.writeJson(jsonGenerator, status);
             }
             jsonGenerator.writeEndArray();
             jsonGenerator.writeEndArray();
```
olperr1, line 82: Should the contingencies and the operator strategies also be generated here?
The result is different from the one generated by `SensitivityAnalysisResultSerializer` (which also includes the contingencies and the operators strategies).

If these elements were excluded on purpose, a Javadoc comment would be useful to explain why.

geofjamg: this is a bug, it has to be consistent

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityOperatorStrategiesCalculationMode.java
geofjamg: Ok I renamed ALL_CONTINGENCIES to CONTINGENCIES_AND_OPERATOR_STRATEGIES but I think we should keep the enum name SensitivityOperatorStrategiesCalculationMode because otherwise it would be confusing with the contingency context attached to the factor

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
geofjamg: Fixed

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
geofjamg: Fixed

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
geofjamg: It is tempting... but as it is widely used everywhere it would be a nasty breaking change. Should we do it anyway?

geofjamg: Or we clutter the API by adding them keeping existing

SylvestreSakti: I was thinking of adding new methods, but that would indeed make the API much bigger

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
geofjamg: Fixed

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
geofjamg: Fixed

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityState.java
geofjamg: Done

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/json/SensitivityAnalysisResultDeserializer.java
geofjamg: Ok

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
olperr1: The Javadoc comment should be updated:

```suggestion
     * @param state the considered state
```

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
```diff
@@ -231,7 +288,7 @@ public double getSensitivityValue(String contingencyId, String variableId, Strin
      * @return the sensitivity value associated with a given function and a given variable for a given contingency.
      */
     public double getBranchFlow1SensitivityValue(String contingencyId, String variableId, String functionId, SensitivityVariableType variableType) {
```
olperr1, line 294: Previously, all these methods could have been called with a `null` contingency ID.
As the Javadoc states:
```java
 * @param contingencyId the id of the considered contingency. Use null to get a pre-contingency sensitivity value.
```

Now, it will throw an exception (`SensitivityState.postContingency(null)` throws an NPE").


To prevent problems, one quick way to solve this could be to replace the `postContingency(String contingencyId)` method of `SensitivityState` by:

```java
    public static SensitivityState forContingency(String contingencyId) {
        return contingencyId == null ? PRE_CONTINGENCY : new SensitivityState(Objects.requireNonNull(contingencyId), null);
    }
```

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityState.java
olperr1: Sorry, this was probably not the best idea (see my previous comment).
Reverting this change could also solve the problem I described.

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
geofjamg: reverted

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityResultJsonWriter.java
geofjamg: not easy to become consistent, I updated to be as closed as possible to SensitivityAnalysisResult serialization, but there is still the factor that are missing and this has been done for performances purposes. On the other hand the SensitivityAnalysisResult need to have the factor to deserialilize. 
I propose that we try to fully solve the inconsistency later because not easy and not related to this work.

olperr1: I think you are right, this can be done later.

PR #3680, opened 2025-12-02
Add flatten in export options

antoinebhs:

No

Feature

There is no option to flatten the network on export.

There is an option to flatten the network on export.
```java
        ExportOptions options = new ExportOptions().setFlatten(true);
        NetworkSerDe.write(merged, options, xmlFile);
```

- [ ] Yes
- [x] No

None

**Other information**:

### DIFF
--- a/docs/user/itools/security-analysis.md
+++ b/docs/user/itools/security-analysis.md
@@ -9,7 +9,7 @@ usage: itools [OPTIONS] security-analysis --case-file <FILE>
        [--contingencies-file <FILE>] [--external] [--help] [-I <property=value>]
        [--import-parameters <IMPORT_PARAMETERS>] [--limit-types <LIMIT-TYPES>]
        [--log-file <FILE>] [--output-file <FILE>] [--output-format <FORMAT>]
-       [--parameters-file <FILE>] [--skip-postproc] [--with-extensions
+       [--parameters-file <FILE>] [--with-extensions
        <EXTENSIONS>]
 
 Available options are:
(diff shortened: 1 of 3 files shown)


iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -527,6 +527,9 @@ private static TreeDataWriter createTreeDataWriter(Network n, ExportOptions opti
     }
 
     private static void write(Network network, NetworkSerializerContext context, ExtensionsSupplier extensionsSupplier) {
+        if (context.getOptions().isFlatten()) {
+            network.flatten();
+        }
```
rolnico: I think there is an issue with this: here your original network is flatten, while exporting a network should not change the network itself (only the exported version of it).
Maybe consider copying the network before flattening it?

rolnico (changes requested):

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
rolnico: You can test this issue by adding `assertEquals(2, merged.getSubnetworks().size());` in your test before and after exporting the network.

geofjamg: agree, this is the basic contract of an exporter, the network should not be modified.

antoinebhs: Yes, I also completely agree. I pushed a correction in https://github.com/powsybl/powsybl-core/pull/3680/commits/7302fe71251953d2e5d92dfdd768769fabfbe078. Sorry for that!

olperr1: Yes, we already discussed this point with @antoinebhs. He is currently changing its code to keep the network unchanged.
It is easy to do it in the SerDe since the subnetworks structure is ignored for IIDM versions prior to 1.11. The only thing to do is slightly change the condition to take the new parameter into account.

rolnico (dismissed):
LGTM

olperr1 (dismissed):

rolnico (approved):

MatthieuSAUR (approved):

PR #3688, opened 2025-12-11
Use reactance seen from tap changer's end to calculate reactance deviations for linear and non-linear PhaseTapChanger

rcourtier:

Fixes https://github.com/powsybl/pypowsybl/issues/1131.

Bug fix.

PhaseTapChanger reactance deviation is calculated from the reactance seen from the primary side.

PhaseTapChanger reactance deviation is calculated from the reactance seen from the tap changer's side.

- [ ] Yes
- [X] No

**Other information**:

### DIFF
--- a/computation-local/src/main/java/com/powsybl/computation/local/ProcessHelper.java
+++ b/computation-local/src/main/java/com/powsybl/computation/local/ProcessHelper.java
@@ -17,46 +17,28 @@ import java.util.concurrent.TimeUnit;
  */
 public final class ProcessHelper {
 
-    private static final long DEFAULT_FAST_POLLING = 500_000_000; // 0.5 seconds
-    private static final long BEGINING_NANO = 3_000_000_000L; // 3 seconds
+    public static final int TIMEOUT_EXIT_CODE = 124;
 
     /**
-     * It requests for exit code by {@link java.lang.Process#exitValue()} every half second at the begining of 3 seconds.
-     * Then it will no more request for exit code until 80% of {@code timeoutInSeconds}.
+     * This method blocks the current thread until either:
+     * <ul>
+     *     <li>The process terminates, in which case it returns the exit code from {@link Process#exitValue()}</li>
+     *     <li>The timeout is reached, in which case the process is destroyed and the exit code {@value #TIMEOUT_EXIT_CODE} is returned</li>
+     * </ul>
      * @param process The process to run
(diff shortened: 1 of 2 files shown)


BenjPetr:
Thank you. Works for me. I build the PR and tested the import in pypowsybl.

rcourtier:
I have added another commit following a suggestion from @marqueslanauja to make the conversion more generic, handling in a better way the location of the structural ratio. The PR is now ready for review.

marqueslanauja (dismissed):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/CgmesT2xModel.java
```diff
@@ -61,12 +72,12 @@ static class CgmesPartialEnd {
         final double ratedU;
         final String terminal;
 
-        CgmesPartialEnd(PropertyBag bagEnd, double x, Context context) {
+        CgmesPartialEnd(PropertyBag bagEnd, double x, double ratedU, Context context) {
```
olperr1: Is there a real interest to change this signature to add the `ratedU` as a parameter?

- It is always called with `bagEnd.asDouble(CgmesNeames.RATEDU)`.
- `g` and `b` are also retrieved from `bagEnd`, but not passed as parameters.
- It introduces an unnecessary breaking change (preventing this PR to be included in a potential corrective version).

If there's a good reason for this change, I think we should have another constructor with the previous signature to "remove" the breaking change:
```java
CgmesPartialEnd(PropertyBag bagEnd, double x, Context context) {
    this(bagEnd, x, bagEnd.asDouble(CgmesNames.RATEDU), context);
}
```

zamarrenolm: I think the change should not be considered as breaking, as the `CgmesPartialEnd` is a static nested class, with no explicit outside visibility. It should not be used from outside the transformer model conversion.

olperr1: This is obviously a **bad practice**, but since the method is package-protected someone can technically use this class from a class defined in the same package. If we want to have this fix in a corrective version, I think we should avoid this kind of change when possible.
@rcourtier: thanks for restoring the original signature!

PR #3696, opened 2025-12-22
Fix runWithTimeout not detecting process terminate before 80% of timeout

antoinebhs:

No

Bug fix

The existing implementation of `runWithTimeout` sleep for a large portion of the timeout (~80%), delaying detection of a completed process.  
Some computation providers force the use of a timeout and these same computation providers can run computations with very different computation times (from few seconds to >15 minutes).   
For example, if we take some margins and set a timeout of 30 minutes, the existing implementation will poll the first 3 seconds then sleep for (timeout*0.8-running) ~= 24 minutes (!). If the process terminates in 10 seconds with this timeout, it will really exit after 24 minutes...!

I propose to use `Process.waitFor(timeout, TimeUnit)` instead of the custom polling implementation.
On modern JVMs, waitFor is reimplemented internally and delegates to OS-level process waiting. This avoid busy polling which might be why we implemented this `smartWait` at first (?).
If we want to keep the custom polling implementation, I can also change the `smartWait` method to check for termination more often.

- [ ] Yes
- [x] No

### DIFF
--- a/computation-local/src/test/java/com/powsybl/computation/local/ProcessHelperTest.java
+++ b/computation-local/src/test/java/com/powsybl/computation/local/ProcessHelperTest.java
@@ -15,6 +15,7 @@ import java.io.OutputStream;
 import java.util.concurrent.TimeUnit;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.fail;
 import static org.mockito.Mockito.*;
 
@@ -41,12 +42,8 @@ class ProcessHelperTest {
 
     @Test
     void testIllegalTimeout() {
-        try {
-            ProcessHelper.runWithTimeout(-1, process);
-            fail();
-        } catch (Exception e) {
-            assertEquals("negative timeout: -1", e.getMessage());
-        }
+        IllegalArgumentException exception = assertThrows(IllegalArgumentException.class, () -> ProcessHelper.runWithTimeout(-1, process));
+        assertEquals("negative timeout: -1", exception.getMessage());
     }
 
     @Test


computation-local/src/test/java/com/powsybl/computation/local/ProcessHelperTest.java
```diff
@@ -51,22 +51,33 @@ void testIllegalTimeout() {
...
+            verify(process, times(1)).waitFor(10, TimeUnit.SECONDS);
+            verify(process, times(1)).exitValue();
             verify(process, never()).destroy();
         } catch (Exception e) {
             fail();
         }
     }
 
+    @Test
+    void testWithLongTimeout() throws Exception {
```
antoinebhs: This adds a ~3s non-regression test, let me know if I should remove it given that other tests in the class use mocking

sBouzols: Maybe add the inverse test ? :
```
Process sleepProcess = new ProcessBuilder("sleep", "10").start();
...
int exitCode = ProcessHelper.runWithTimeout(5, sleepProcess);
assertEquals(124, exitCode);
assertTrue(duration < 5500); // security margin
```

sBouzols: `testTimeouted` doesn't really test this case...

computation-local/src/main/java/com/powsybl/computation/local/ProcessHelper.java
```diff
@@ -17,46 +17,27 @@
...
         closeStream(process);
-        return exitCode;
-    }
-
-    private static long smartWait(long running, long startTime, long timeout) {
-        if (running < BEGINING_NANO || running > 0.8 * TimeUnit.SECONDS.toNanos(timeout)) {
-            return DEFAULT_FAST_POLLING;
-        }
-        return (long) (TimeUnit.SECONDS.toNanos(timeout) * 0.8) - running;
+        return 124;
```
rolnico: You can add `private static final int TIMEOUT_EXIT_CODE = 124;` at the beginning of the class and then replace here with

```suggestion
        return TIMEOUT_EXIT_CODE;
```

Replace also in the javadoc with `{@value #TIMEOUT_EXIT_CODE}`

computation-local/src/main/java/com/powsybl/computation/local/ProcessHelper.java
```diff
@@ -17,46 +17,27 @@
...
-     * It requests for exit code by {@link java.lang.Process#exitValue()} every half second at the begining of 3 seconds.
-     * Then it will no more request for exit code until 80% of {@code timeoutInSeconds}.
+     * Waits for the given process to finish up to the specified timeout.
+     * This method blocks efficiently until either:
+     * <ul>
+     *     <li>The process terminates, in which case it returns the exit code from {@link Process#exitValue()}</li>
+     *     <li>The timeout expires, in which case the process is destroyed and 124 is returned</li>
+     * </ul>
      * @param process The process to run
      * @return Returns exit code for the process. Returns 124 if process timeout.
```
rolnico:
```suggestion
     * This method blocks the current thread until either:
     * <ul>
     *     <li>The process terminates, in which case it returns the exit code from {@link Process#exitValue()}</li>
     *     <li>The timeout is reached, in which case the process is destroyed and the exit code 124 is returned</li>
     * </ul>
     * @param process The process to run
     * @return Returns exit code for the process, or 124 if the process timeouts.
```

computation-local/src/test/java/com/powsybl/computation/local/ProcessHelperTest.java
rolnico: I agree with @sBouzols , the timeout test should be rewritten properly (all the tests, while we are at it)

computation-local/src/test/java/com/powsybl/computation/local/ProcessHelperTest.java
```diff
@@ -12,9 +12,9 @@
 
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.util.concurrent.TimeUnit;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.fail;
+import static org.junit.jupiter.api.Assertions.*;
```
rolnico: revert this

antoinebhs: Done in https://github.com/powsybl/powsybl-core/pull/3696/changes/dcde9ddefad4ebd61ab24fe27fa54347b003dd66

computation-local/src/main/java/com/powsybl/computation/local/ProcessHelper.java
antoinebhs: Done in https://github.com/powsybl/powsybl-core/pull/3696/changes/dcde9ddefad4ebd61ab24fe27fa54347b003dd66

computation-local/src/main/java/com/powsybl/computation/local/ProcessHelper.java
antoinebhs: Done in https://github.com/powsybl/powsybl-core/pull/3696/changes/dcde9ddefad4ebd61ab24fe27fa54347b003dd66

computation-local/src/test/java/com/powsybl/computation/local/ProcessHelperTest.java
rolnico: Here is a suggestion for the tests:

```java
    @Test
    void testProcessFinishedBeforeTimeout() {
        try {
            when(process.waitFor(10, TimeUnit.SECONDS)).thenReturn(true);
            when(process.exitValue()).thenReturn(0);

            int exitCode = ProcessHelper.runWithTimeout(10, process);

            assertEquals(0, exitCode);
            verify(process, times(1)).waitFor(10, TimeUnit.SECONDS);
            verify(process, times(1)).exitValue();
            verify(in, times(1)).close();
            verify(out, times(1)).close();
            verify(err, times(1)).close();
            verify(process, never()).destroy();
        } catch (Exception e) {
            fail();
        }
    }

    @Test
    void testTimeoutBeforeProcessFinishes() {
        try {
            // process never finish
            when(process.waitFor(2, TimeUnit.SECONDS)).thenReturn(false);

            int exitCode = ProcessHelper.runWithTimeout(2, process);

            assertEquals(124, exitCode);
            verify(process, times(1)).waitFor(2, TimeUnit.SECONDS);
            verify(process, never()).exitValue();
            verify(in, times(1)).close();
            verify(out, times(1)).close();
            verify(err, times(1)).close();
            verify(process, times(1)).destroy();
        } catch (Exception e) {
            fail();
        }
    }
```

With the changes you made, I don't think there is a need to add your test with long timeout anymore.

rolnico (changes requested):

computation-local/src/test/java/com/powsybl/computation/local/ProcessHelperTest.java
antoinebhs: Done in https://github.com/powsybl/powsybl-core/pull/3696/commits/c970c22d97a1b8c90cdaf1592219ac2857b9f864

PR #3700, opened 2025-12-30
Fix CGMES update "Query not found in catalog" warnings

jeandemanged:

No

Bug fix

Misleading warnings during CGMES update:
```
WARN  c.p.c.m.t.CgmesModelTripleStore | Query [boundaryNodes] not found in catalog
WARN  c.p.c.m.t.CgmesModelTripleStore | Query [transformerEnds] not found in catalog
WARN  c.p.c.m.t.CgmesModelTripleStore | Query [ratioTapChangerTablePoints] not found in catalog
WARN  c.p.c.m.t.CgmesModelTripleStore | Query [phaseTapChangerTablePoints] not found in catalog
WARN  c.p.c.m.t.CgmesModelTripleStore | Query [nonlinearShuntCompensatorPoints] not found in catalog
WARN  c.p.c.m.t.CgmesModelTripleStore | Query [reactiveCapabilityCurveData] not found in catalog
WARN  c.p.c.m.t.CgmesModelTripleStore | Query [equivalentShunts] not found in catalog
```

The warnings are suppressed

For simplicity, the warnings are removed by adding dummy queries returning nothing to CIM16-update.sparql).

- [x] No

**Other information**:

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesReports.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesReports.java
@@ -100,6 +100,13 @@ public final class CgmesReports {
                 .add();
     }
 
+    public static void noInputsForUpdateReport(ReportNode reportNode) {
+        reportNode.newReportNode()
+            .withMessageTemplate("core.cgmes.conversion.noInputsForUpdate")
+            .withSeverity(TypedValue.INFO_SEVERITY)
+            .add();
+    }
+
     // WARN
     public static void badVoltageTargetValueRegulatingControlReport(ReportNode reportNode, String eqId, double targetValue) {
         reportNode.newReportNode()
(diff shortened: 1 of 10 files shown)


rcourtier:
I understand the need to suppress these warnings. However, instead of creating dummy queries returning nothing, I would:
- Either plain and simple remove the `LOG.warn("Query [{}] not found in catalog", name);` in `CgmesModelTripleStore.namedQuery`, since not having a query text can be a valid behaviour since the split of the EQ import and SSH/SV update.
- Or make the `queryCatalogName` in `CgmesModelTripleStore` a class attribute and have a simple logic to not call `namedQuery` if the query is not expected with that queryCatalogName (in which case return a new PropertyBags).

jeandemanged:
> I understand the need to suppress these warnings. However, instead of creating dummy queries returning nothing, I would:
> 
> * Either plain and simple remove the `LOG.warn("Query [{}] not found in catalog", name);` in `CgmesModelTripleStore.namedQuery`, since not having a query text can be a valid behaviour since the split of the EQ import and SSH/SV update.
> * Or make the `queryCatalogName` in `CgmesModelTripleStore` a class attribute and have a simple logic to not call `namedQuery` if the query is not expected with that queryCatalogName (in which case return a new PropertyBags).

Hi @rcourtier ,
Thank you, I went for first option, changing log from warning to debug level.

PR #3704, opened 2025-12-31
Avoid logging temporary loading limit being set to the previous value

jeandemanged:

No

"inconvenience" fix

Setting a loading temporary limit to its previous value generates an info log.
In the context of CGMES importer update this causes thousands of useless logs being issued such as:

`AC line 'abcd': Temporary limit value changed from 567.0 to 567.0`

Changes below 1e-6 are filtered out.
Also added the temporary limit acceptable duration to the log.

- [x] No

**Other information**:


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractLoadingLimits.java
```diff
@@ -99,23 +99,25 @@ public L setTemporaryLimitValue(int acceptableDuration, double temporaryLimitVal
...
-        Map.Entry<Integer, TemporaryLimit> smallerDurationEntry = temporaryLimitTreeMap.higherEntry(acceptableDuration);
-
         double oldValue = identifiedLimit.getValue();
 
-        if (isTemporaryLimitValueValid(biggerDurationEntry, smallerDurationEntry, acceptableDuration, temporaryLimitValue)) {
-            LOGGER.info("{}Temporary limit value changed from {} to {}", group.getValidable().getMessageHeader(), oldValue, temporaryLimitValue);
-        } else {
-            LOGGER.warn("{}Temporary limit value changed from {} to {}, but it is not valid", group.getValidable().getMessageHeader(), oldValue, temporaryLimitValue);
-        }
+        if (Math.abs(temporaryLimitValue - oldValue) > 1e-6) { // do not apply negligible changes
```
jeandemanged: or strict equality ?

rolnico: I would keep the epsilon, but define it in a variable of the class, for better visibility.

PR #3705, opened 2025-12-31
Improve and optionally silence some CGMES importer warning logs

jeandemanged:

No

"inconvenience" fixes

Too much warning logs on real cases:
* Mostly about Operational Limits ending in practice in the following form:
  ```
   Pending: Operational limit. Reason: Not assigned for ShuntCompensator 123. Limit id, type, typeName, subClass, terminal : 456, LimitTypeKind.tatl, someName, CurrentLimit, ab123
   Missing: Terminal 456 or Equipment null at OperationalLimit TATL20 (def). Reason:
  ```

    Caused by OperationalLimits on equipement types not supporting OperationalLimits in IIDM, such as switches, loads etc...

* When import is Bus/Breaker e.g. by option, but the CGMES is Node/Breaker, many Switches get this warning:
  ```
   Ignored: Switch xxx (xxx). Reason: end buses are the same bus xxx 
  ```
* Generators without minQ/maxQ limits:
  ```
   Missing: minQ/maxQ are missing, default to unbounded reactive limits at generator xxx (xxx)
  ```

* logs with an empty reason end with `<some issue> . Reason:`
* superfluous warning about missing name on cim:TieFlow-s
* superfluous warnings cim:Switch connected to boundary point (perfectly valid in CGMES, and supported by the importer)

* added option to silence these frequent warnings. Default is to not silence them.
* suppressed warning of missing cim:TieFlow name
* warnings about the details of switch connected to boundary point changed to debug level, remains only one single warning.
* messages with empty Reason do not end with `... Reason:` anymore.

- [x] No

**Other information**:


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/SwitchConversion.java
```diff
@@ -119,7 +119,7 @@ private SwitchKind kind() {
     }
 
     private void warnDanglingLineCreated() {
-        fixed("Dangling line with low impedance", "Connected to a boundary node");
+        fixed("Converted as a dangling line with zero impedance", "Connected to a boundary node");
```
jeandemanged: I would change this to just an info log, or even debug, or even remove completely, this behavior is described in the CGMES import documentation [here](https://powsybl.readthedocs.io/projects/powsybl-core/en/stable/grid_exchange_formats/cgmes/import.html#switch-switch-breaker-disconnector-loadbreakswitch-protectedswitch-grounddisconnector-jumper).

jeandemanged:
still some uncovered cases about OperationalLimits ... returning to draft

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/SwitchConversion.java
rcourtier: This generates a warn severity level log, which is usually used when an inconsistent or incomplete data is fixed. Here there is no inconsistency, this is the standard conversion process and it is documented as you pointed out, so I'm ok to change this to an info severity level log.
I think however it is important to keep this as an info and not a debug log, as users often tend to hide the debug logs and they might struggle to understand why their CGMES boundary switch is not visible in the IIDM switches.

jeandemanged: proposing to leave as it is here as a warning - and just turn to debug the other details, so we have only one warning for one boundary switch

PR #3712, opened 2026-01-07
Convert reports_fr from latin1 to utf8

jonenst:
Since java9 jep 226, resource bundle are read in utf8 and if that fails reread as latin1 so it's actually very slightly better for performance and more aligned with the platform to distribute them in utf8 now and everything works the same

Switching to utf8 also prevents that people by accident rewrite the whole file and replace all latin1 characters by the unicode error character � (some editors do this, I saw it with vscode)

no

feature

properties in latin1

properties in utf8

- [ ] Yes
- [x] No

**Other information**:


rolnico (commented):
Shouldn't we switch all the `.properties` files to UTF-8, and not only the reports_fr ones?
There are also multiple `reports.properties` and some reports_fr_FR.properties too that seem to be using another encoding at the moment

PR #3717, opened 2026-01-14
Add new GeneratorEntsoeCategory extension version that allows code value of 0

rcourtier:

No.

Bug fix + doc update.

Generator's extension GeneratorEntsoeCategory with code = 0 is not allowed (min value is 1).

GeneratorEntsoeCategory extension has a new version (v1.1) that allows code = 0 (min value is 0).
Backward compatibililty is ensured, although when trying to serialize a code = 0 in the legacy version (v1.0), the extension is not written and a warning log is generated.

- [ ] Yes
- [X] No

**Other information**:

Generators with code equal to 0 represent synchronous condenser.


iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/extensions/GeneratorEntsoeCategorySerDe.java
```diff
@@ -8,23 +8,55 @@
...
+    }
+
+    @Override
+    public boolean isValid(GeneratorEntsoeCategory entsoeCategory, SerializerContext context) {
+        if (entsoeCategory.getCode() == 0 && getExtensionVersionToExport(context) == Version.V_1_0) {
+            LOGGER.warn("Extension entsoeCategory not valid for Generator: {}. Reason: code = 0 not allowed in version 1.0.",
+                entsoeCategory.getExtendable().getId());
+            return false;
+        }
+        return true;
```
rolnico, line 59: Instead of adding a new `isValid` method, you could override the existing `isSerializable` method.
See for example the override in `MeasurementsSerDe`:

```java
    @Override
    public boolean isSerializable(Measurements<C> extension) {
        return !extension.getMeasurements().isEmpty();
    }
```

rcourtier: Right now this is not possible.
I need a method where I can pass the `SerializerContext` as argument, in order to return a different result based upon the extension version to export.

rcourtier: Created an overload of `isSerializable` instead of a new `isValid` method.
On interface level, the default implementation of `isSerializable(E extension, SerializerContext context)` calls `isSerializable(E extension)`, so no breaking change is introduced.

PR #3718, opened 2026-01-16
Add network case in get identifiable stream, and fix typo in error message

NathanDissoubray:

Feature:
- Change an error message missing a negation in it
- Add a missing subcase for IdentifiableType.NETWORK

- Typo
- Case IdentifiableType.Network in Network.getIdentifiableStream is considered as an exception, when we could get the Subnetworks of that given network instead (if they exist)

The NETWORK case now returns a stream of subnetworks instead of a PowsyblException
This stream starts with the parent network, followed by the subnetworks

- [ ] Yes
- [x] No

Custom IIDM implementations maintainers should handle the network case in `Network.getIdentifiableStream()`.
In this case, it should return a stream with the current network and its subnetworks.

**Other information**:


NathanDissoubray:
I'm just surprised there is no GROUND (only DC_GROUND) in the `AbstractSubnetworksExplorationTest`, did I miss something ?

NathanDissoubray:
Should I add two tests for BUS and DC_BUS to check that they throw ? Because here we'll get 50% coverage on new code

olperr1:
> Should I add two tests for BUS and DC_BUS to check that they throw ? Because here we'll get 50% coverage on new code

I think you can add a `testIdentifiableStreamUnsupportedType()` test method, which assert that when a type is not supported (for instance `BUS`, there's no need to also have `DC_BUS`), an exception is thrown.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Network.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2016, All partners of the iTesla project (http://www.itesla-project.eu/consortium)
+ * Copyright (c) 2016-2026, All partners of the iTesla project (http://www.itesla-project.eu/consortium)
```
olperr1: I think it should be:

```suggestion
 * Copyright (c) 2016, All partners of the iTesla project (http://www.itesla-project.eu/consortium)
 * Copyright (c) 2016-2026, RTE (http://www.rte-france.com)
```

olperr1 (approved):

PR #3737, opened 2026-01-26
Add magnetizing reactance support to TwoWindingsTransformerFortescue

colineplqt:

No

Feature

`TwoWindingsTransformerFortescue` stores the asymmetrical data on `TwoWindingsTransformer` for calculations.

This PR adds the magnetizing reactance as a field of the extension. This can be used for asymmetrical short circuit calculations.

- [X] Yes, for custom IIDM implementations' maintainers only
- [ ] No

**Migration steps**
Custom IIDM implementations' maintainers should implement the following new methods:

- In their `TwoWindingsTransformerFortescue` implementations:
  - `double getXm();`
  - `void setXm(double xm);`
  
- In their `TwoWindingsTransformerFortescueAdder` implementations:
  - `TwoWindingsTransformerFortescueAdder withXm(double xm);`


colineplqt:
What about 3-winding transformers?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/TwoWindingsTransformerFortescueImpl.java
```diff
@@ -22,21 +22,32 @@ public class TwoWindingsTransformerFortescueImpl extends AbstractExtension<TwoWi
...
+    private double xmz;
     private WindingConnectionType connectionType1;
     private WindingConnectionType connectionType2;
     private double groundingR1;
     private double groundingX1;
     private double groundingR2;
     private double groundingX2;
 
-    public TwoWindingsTransformerFortescueImpl(TwoWindingsTransformer twt, double rz, double xz, boolean freeFluxes,
+    public TwoWindingsTransformerFortescueImpl(TwoWindingsTransformer twt, double rz, double xz, double rz1, double xz1,
```
rolnico: I would prefer to keep the old constructor and to add a new one with only the splitted values to avoid any risk of inconsistency

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/TwoWindingsTransformerFortescueImpl.java
```diff
@@ -22,21 +22,32 @@ public class TwoWindingsTransformerFortescueImpl extends AbstractExtension<TwoWi
 
     private double rz;
     private double xz;
+    private double rz1;
+    private double xz1;
+    private double rz2;
+    private double xz2;
```
rolnico: If possible, it would be better to only save the splitted values (if you can compute them from `rz` and `xz`), to avoid redundancy and avoid any risk of inconsistency

rolnico (changes requested):

colineplqt:
@rolnico, I have removed the split impedances. I will rework it and n the meantime, the PR does what its title says :)

rolnico (approved):

PR #3744, opened 2026-01-29
Compute also connectables in CalculatedBus getTerminals()

### LINKED ISSUE #3507: Calculated bus connectable terminals

When calling `getTerminals()` on a calculated bus of the bus view from the node breaker topology, the terminals are the connected terminals and do not include the connectable terminals. Indeed the `CalculatedBusImpl` buses does not hold the corresponding connectable terminals.

This leads to an unexpected behaviour when using `visitConnectedOrConnectableEquipments(visitor)` of the `Bus` interface.

- Either `getTerminals()` should throw an `UnsupportedOperationException("Not implemented");`
- Or we should compute the connectable buses when calculating the topology

### Describe the steps

_No response_

### Environment

_No response_

### Relevant Log Output

_No response_

### Extra Information

_No response_

quinarygio:

Fixes #3507 

- [ ] Yes
- [ ] No

**Other information**:


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/CalculatedBusImpl.java
```diff
@@ -98,7 +98,22 @@ public Stream<TerminalExt> getConnectedTerminalStream() {
...
+                    Bus connectableBus = te.getBusView().getConnectableBus();
+                    if (this.equals(connectableBus) && !all.contains(te)) {
+                        all.add(te);
+                    }
+                } catch (PowsyblException e) {
+                    // ignore removed equipments or inaccessible buses
+                }
+            }
+        }
+        return Collections.unmodifiableCollection(all);
```
samirromdhani: Actually, there is a method called `getConnectedTerminals` that calls `getTerminals`

```java
public Collection<TerminalExt> getConnectedTerminals() {
    return getTerminals();
}
```
If `getTerminals` return both connected or connectable terminals, the method `getConnectedTerminals`, will incorrectly include connectable terminals.

I suggest keeping `getTerminals` as a simple getter of terminals, and updating `getConnectedTerminals` to filter only connected one.

```java
public Collection<TerminalExt> getConnectedTerminals() {
    return getTerminals().stream().filter(TerminalExt::isConnected).toList();
}
```
For this issue fix, I think it is better to define a private method named `getConnectedOrConnectableTerminals`, and used in the existing  `visitConnectedOrConnectableEquipments(TopologyVisitor visitor)`

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/CalculatedBusImpl.java
rolnico:
```suggestion
        return Collections.unmodifiableCollection(terminals);
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/CalculatedBusImpl.java
rolnico: Or just return `Collections.unmodifiableCollection(terminals);` in `getConnectedTerminals()` instead of calling `getTerminals()`

rolnico (dismissed):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/CalculatedBusImpl.java
```diff
@@ -87,7 +87,7 @@ public int getConnectedTerminalCount() {
 
     @Override
     public Collection<TerminalExt> getConnectedTerminals() {
-        return getTerminals();
+        return Collections.unmodifiableCollection(terminals);
```
olperr1, line 93: To be consistent with `getConnectedTerminalStream()`, `getConnectedTerminalCount()`, …
```suggestion
        checkValidity();
        return Collections.unmodifiableCollection(terminals);
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/CalculatedBusImpl.java
```diff
@@ -99,7 +99,22 @@ public Stream<TerminalExt> getConnectedTerminalStream() {
     @Override
     public Collection<TerminalExt> getTerminals() {
         checkValidity();
-        return Collections.unmodifiableCollection(terminals);
+        List<TerminalExt> all = new ArrayList<>(terminals);
+        int[] nodes = voltageLevel.getNodeBreakerView().getNodes();
+        for (int n : nodes) {
```
olperr1:
```suggestion
        int[] vlNodes = voltageLevel.getNodeBreakerView().getNodes();
        for (int n : vlNodes) {
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/CalculatedBusImpl.java
```diff
@@ -99,7 +99,22 @@ public Stream<TerminalExt> getConnectedTerminalStream() {
...
+            Terminal t = voltageLevel.getNodeBreakerView().getTerminal(n);
+            if (t instanceof TerminalExt te) {
+                try {
+                    Bus connectableBus = te.getBusView().getConnectableBus();
```
olperr1: I think this instruction could be expensive. Shouldn't we keep a cache of all the `TerminalExt` linked via a connectableBus, in order to speed-up this method?
In that case, we should consider build the cache with all the `TerminalExt te` such as `this.equals(te.getBusView().getConnectableBus())`. The part `!all.contains(te)` should only be considered when choosing the elements of the cache to add to `all`. This way, there's no need to invalidate the cache each time an element is (dis)connected to its connectable bus.

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractNodeBreakerTest.java
```diff
@@ -597,4 +600,201 @@ public void testCalculatedBusTopologyWithLoop() {
...
+            }
+        });
+        assertEquals(2, bus.getConnectedTerminalCount());
+        assertEquals(2, visitedConnected.size(), "Should visit 2 equipments: BBS and L1");
+        assertTrue(visitedConnected.stream().anyMatch(c -> "BBS".equals(c.getId())));
+        assertTrue(visitedConnected.stream().anyMatch(c -> "L1".equals(c.getId())));
+        assertFalse(visitedConnected.stream().anyMatch(c -> "G1".equals(c.getId())));
+
+        // Verify visitConnectedOrConnectableEquipments includes all equipments
+        List<Connectable<?>> visitedConnectables = new ArrayList<>();
```
olperr1, line 653: There's no need to create a second visitor: if the same list is used twice (after clearing its entries), the visitor will be the same.

olperr1 (changes requested):

olperr1 (approved):

PR #3753, opened 2026-02-03
Refactor graph traversal methods to use new `computeTraversalPartitions` API

clementleclercRTE:

- [  ] A PR or issue has been opened in all impacted repositories (if any)

 Refactoring / New API 

Currently, BusBreakerTopologyModel, NodeBreakerTopologyModel, and DcTopologyModel all contain duplicated logic to traverse the graph and identify connected components (bus construction).

  A new method `computeTraversalPartitions` has been introduced in the `UndirectedGraph` interface. It centralizes the graph traversal and connected component building logic, using standard
  functional interfaces (`Supplier`, `ObjIntConsumer`) for flexibility. The three topology models have been refactored to use this shared method.

  The previously public method `traverse(int, TraversalType, Traverser, boolean[])` is now deprecated  in favor of `traverse(int, TraversalType, Traverser)` and
  `computeTraversalPartitions(...)`.

- [X] Yes
- [ ] No

  **1. Deprecated: `traverse(int, TraversalType, Traverser, boolean[])`** 

  If you were using this overload to manually track visited vertices, switch to the standard `traverse(int, TraversalType, Traverser)` overload. Visited vertices can be captured directly inside your
   `Traverser` lambda.

  **2. Manual connected component computation**

  If you were manually implementing the traversal-and-partition pattern (iterating over vertices, tracking visited state, grouping results), you can now replace this logic with a call to
  `computeTraversalPartitions(Traverser)` or its generic overload `computeTraversalPartitions(Traverser, Supplier<C>, ObjIntConsumer<C>)`.

**Other information**:


math/src/main/java/com/powsybl/math/graph/UndirectedGraph.java
```diff
@@ -331,6 +331,18 @@ public interface UndirectedGraph<V, E> {
...
     boolean traverse(int v, TraversalType traversalType, Traverser traverser, boolean[] verticesEncountered);
 
+    <C> List<C> getConnectedComponents(Predicate<E> isTraversable, ConnectedComponentCollector<C, V> collector);
+
+    interface ConnectedComponentCollector<C, V> {
+        C createComponent();
+
+        void addVertex(C component, int vertexIndex, V vertexObject);
+
+        default boolean isComponentValid(C component) {
```
flo-dup: This method does not make much sense in the graph world. So keep it to iidm side, that is, 
- always add the connected component in the graph world,
- and in iidm world get all the connected components and then dismiss the ones which are not valid for iidm.

math/src/main/java/com/powsybl/math/graph/UndirectedGraphImpl.java
```diff
@@ -579,6 +579,43 @@ public boolean traverse(int v, TraversalType traversalType, Traverser traverser,
...
+        for (int v = 0; v < vertices.size(); v++) {
+            Vertex<V> vertex = vertices.get(v);
+            if (vertex != null && !encountered[v]) {
+                C component = collector.createComponent();
+
+                collector.addVertex(component, v, vertex.getObject());
+
+                traverse(v, TraversalType.BREADTH_FIRST, (v1, e, v2) -> {
+                    E edgeObject = edges.get(e).getObject();
+                    if (edgeObject == null || isTraversable.test(edgeObject)) {
```
flo-dup: we might want to stop even if edgeObject == null. So we should reconsider the traverse signature. We probably need to replace the `Predicate<E> isTraversable` with a `Traverser`

math/src/main/java/com/powsybl/math/graph/UndirectedGraphImpl.java
```diff
@@ -579,6 +579,43 @@ public boolean traverse(int v, TraversalType traversalType, Traverser traverser,
...
+                        if (!encountered[v2]) {
+                            Vertex<V> destVertex = vertices.get(v2);
+                            if (destVertex != null) {
```
flo-dup: Just call directly the collector with the integer, no need to fetch the vertex object

flo-dup (changes requested):

math/src/main/java/com/powsybl/math/graph/UndirectedGraphImpl.java
```diff
@@ -579,6 +579,43 @@ public boolean traverse(int v, TraversalType traversalType, Traverser traverser,
...
+        boolean[] encountered = new boolean[vertices.size()];
+
+        for (int v = 0; v < vertices.size(); v++) {
+            Vertex<V> vertex = vertices.get(v);
+            if (vertex != null && !encountered[v]) {
+                C component = collector.createComponent();
+
+                collector.addVertex(component, v, vertex.getObject());
+
+                traverse(v, TraversalType.BREADTH_FIRST, (v1, e, v2) -> {
```
flo-dup: - this traverse method of the UndirectedGraph interface should be deprecated
- this traverse method in UndirectedGraphImpl should be private and a new parameter boolean[] encounteredEdges added

flo-dup (changes requested):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusBreakerTopologyModel.java
```diff
@@ -230,30 +228,35 @@ private void updateCache() {
...
-                    final Set<ConfiguredBus> busSet = new LinkedHashSet<>(1);
-                    busSet.add(graph.getVertexObject(v));
-                    graph.traverse(v, TraversalType.DEPTH_FIRST, (v1, e, v2) -> {
-                        SwitchImpl aSwitch = graph.getEdgeObject(e);
-                        if (aSwitch.isOpen()) {
-                            return TraverseResult.TERMINATE_PATH;
-                        } else {
-                            busSet.add(graph.getVertexObject(v2));
-                            return TraverseResult.CONTINUE;
+            graph.getConnectedComponents(
```
flo-dup: introduce a variable which is the result to improve readability

math/src/main/java/com/powsybl/math/graph/UndirectedGraphImpl.java
```diff
@@ -369,8 +369,8 @@ public int getVertexCapacity() {
     @Override
     public Iterable<V> getVerticesObj() {
         return FluentIterable.from(vertices)
-                .filter(Predicates.notNull())
```
flo-dup, line 372: Could you remove the formatting changes?

math/src/main/java/com/powsybl/math/graph/UndirectedGraphImpl.java
```diff
@@ -579,19 +585,48 @@ public boolean traverse(int v, TraversalType traversalType, Traverser traverser,
...
+    public <C> List<C> getConnectedComponents(Traverser traverser, UndirectedGraph.ConnectedComponentCollector<C> collector) {
+        Objects.requireNonNull(traverser);
+        Objects.requireNonNull(collector);
+
+        boolean[] encounteredVertices = new boolean[vertices.size()];
+        boolean[] encounteredEdges = new boolean[edges.size()];
+        List<C> components = new ArrayList<>(vertices.size());
+
+        for (int v = 0; v < vertices.size(); v++) {
+            Vertex<V> vertex = vertices.get(v);
```
flo-dup: why do you check that vertex != null?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusBreakerTopologyModel.java
```diff
@@ -230,30 +228,35 @@ private void updateCache() {
...
+                new UndirectedGraph.ConnectedComponentCollector<Set<ConfiguredBus>>() {
+                    @Override
+                    public Set<ConfiguredBus> createComponent() {
+                        return new LinkedHashSet<>(1);
+                    }
+
+                    @Override
+                    public void addVertex(Set<ConfiguredBus> component, int vertexIndex) {
+                        ConfiguredBus bus = graph.getVertexObject(vertexIndex);
+                        if (bus != null) {
```
flo-dup: getVertexObject() cannot return null in bus/breaker topology

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusBreakerTopologyModel.java
```diff
@@ -230,30 +228,35 @@ private void updateCache() {
...
-                            return TraverseResult.CONTINUE;
+            graph.getConnectedComponents(
+                (v1, e, v2) -> {
+                    SwitchImpl sw = graph.getEdgeObject(e);
+                    if (sw != null && sw.isOpen()) {
+                        return TraverseResult.TERMINATE_PATH;
+                    }
+                    return TraverseResult.CONTINUE;
+                },
+                new UndirectedGraph.ConnectedComponentCollector<Set<ConfiguredBus>>() {
```
flo-dup: There's no need to have a set anymore, a list would be enough

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusBreakerTopologyModel.java
```diff
@@ -293,7 +296,7 @@ MergedBus getMergedBus(ConfiguredBus bus) {
     }
 
     final CalculatedBusTopology calculatedBusTopology
-            = new CalculatedBusTopology();
+        = new CalculatedBusTopology();
```
flo-dup: remove formatting changes

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java
```diff
@@ -312,10 +277,30 @@ protected void updateCache(final Predicate<SwitchImpl> terminate) {
...
-            }
+
+            graph.getConnectedComponents((v1, e, v2) -> {
+                SwitchImpl sw = graph.getEdgeObject(e);
+                if (sw != null && terminate.test(sw)) {
+                    return TraverseResult.TERMINATE_PATH;
+                }
+                return TraverseResult.CONTINUE;
+            }).forEach(nodes -> {
+                CopyOnWriteArrayList<NodeTerminal> terminals = new CopyOnWriteArrayList<>();
```
flo-dup: I'm wondering if we really need the CopyOnWriteArrayList for buses

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java
```diff
@@ -312,10 +277,30 @@ protected void updateCache(final Predicate<SwitchImpl> terminate) {
...
+                for (int i = 0; i < nodes.size(); i++) {
+                    NodeTerminal terminal = graph.getVertexObject(nodes.getQuick(i));
+                    if (terminal != null) {
+                        terminals.add(terminal);
+                    }
+                }
+                if (getBusChecker().isValid(graph, nodes, terminals)) {
```
flo-dup: the terminals argument is never used. Should it be removed or rather used?

math/src/main/java/com/powsybl/math/graph/UndirectedGraphImpl.java
```diff
@@ -579,19 +585,48 @@ public boolean traverse(int v, TraversalType traversalType, Traverser traverser,
...
     }
 
+    @Override
+    public <C> List<C> getConnectedComponents(Traverser traverser, UndirectedGraph.ConnectedComponentCollector<C> collector) {
+        Objects.requireNonNull(traverser);
+        Objects.requireNonNull(collector);
+
+        boolean[] encounteredVertices = new boolean[vertices.size()];
+        boolean[] encounteredEdges = new boolean[edges.size()];
+        List<C> components = new ArrayList<>(vertices.size());
```
flo-dup: you initialize your array with a really big size 😲

flo-dup (changes requested):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusBreakerTopologyModel.java
clementleclercRTE: Done. I used a simple for loop instead of streams

math/src/main/java/com/powsybl/math/graph/UndirectedGraphImpl.java
clementleclercRTE: done

math/src/main/java/com/powsybl/math/graph/UndirectedGraphImpl.java
clementleclercRTE: Indeed, I've removed that

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusBreakerTopologyModel.java
clementleclercRTE: Indeed, I've removed that check

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusBreakerTopologyModel.java
clementleclercRTE: Switched to a List and updated relevant private function signatures accordingly.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BusBreakerTopologyModel.java
clementleclercRTE: done

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java
clementleclercRTE: Removed it and moved the logic directly into the addBus method

math/src/main/java/com/powsybl/math/graph/UndirectedGraph.java
```diff
@@ -328,9 +328,33 @@ public interface UndirectedGraph<V, E> {
...
+            }
+
+            @Override
+            public void addVertex(TIntArrayList component, int vertexIndex) {
+                component.add(vertexIndex);
+            }
+        });
+    }
+
+    <C> List<C> getConnectedComponents(Traverser traverser, ConnectedComponentCollector<C> collector);
```
flo-dup: To have something more concise we could split the collector in two: `Supplier<C> supplier, ComponentAdder<C> adder`

math/src/main/java/com/powsybl/math/graph/UndirectedGraph.java
```diff
@@ -328,9 +328,33 @@ public interface UndirectedGraph<V, E> {
      *                           </ul>
      * @return false if the whole traversing has to stop, meaning that a {@link TraverseResult#TERMINATE_TRAVERSER}
      * has been returned from the traverser, true otherwise
+     * @deprecated Use {@link #getConnectedComponents(Traverser)} or {@link #traverse(int, TraversalType, Traverser)} instead.
      */
+    @Deprecated(since = "7.2.0")
     boolean traverse(int v, TraversalType traversalType, Traverser traverser, boolean[] verticesEncountered);
 
+    default List<TIntArrayList> getConnectedComponents(Traverser traverser) {
+        return getConnectedComponents(traverser, new ConnectedComponentCollector<TIntArrayList>() {
```
flo-dup: Then this would become `return getConnectedComponents(traverser, TIntArrayList::new, TIntArrayList::add);`

math/src/main/java/com/powsybl/math/graph/UndirectedGraphImpl.java
```diff
@@ -579,19 +585,48 @@ public boolean traverse(int v, TraversalType traversalType, Traverser traverser,
...
+
+        for (int v = 0; v < vertices.size(); v++) {
+            Vertex<V> vertex = vertices.get(v);
+
+            if (vertex != null && !encounteredVertices[v]) {
+                C component = collector.createComponent();
+                collector.addVertex(component, v);
+
+                traverse(v, TraversalType.BREADTH_FIRST, (v1, e, v2) -> {
+                    TraverseResult result = traverser.traverse(v1, e, v2);
```
flo-dup: here we intercept the edge traversal each time, whereas there's a specific method which is called when a vertex is traversed the first time. So we could pass a nullable `IntConsumer vertexVisitor` to handle this.

math/src/main/java/com/powsybl/math/graph/UndirectedGraphImpl.java
```diff
@@ -579,19 +585,48 @@ public boolean traverse(int v, TraversalType traversalType, Traverser traverser,
...
+        List<C> components = new ArrayList<>();
+
+        for (int v = 0; v < vertices.size(); v++) {
+            Vertex<V> vertex = vertices.get(v);
+
+            if (vertex != null && !encounteredVertices[v]) {
+                C component = collector.createComponent();
+                collector.addVertex(component, v);
+
+                traverse(v, TraversalType.BREADTH_FIRST, (v1, e, v2) -> {
```
flo-dup: If traverse returns false, shouldn't we break?

flo-dup (changes requested):
The naming is questionable, as we're not computing connected component if we rely on graph theory definition. We should rather call it something like `computeTraversalPartitions`, as we're partitioning the graph (all the vertices are traversed if there's no `TERMINATE_TRAVERSER` return) based on a given traversal.

math/src/main/java/com/powsybl/math/graph/UndirectedGraphImpl.java
clementleclercRTE: Fixed. Note that TERMINATE_TRAVERSER should not normally be used here.
Using TERMINATE_TRAVERSER would result in an incomplete partition. The break is just a safety net.

math/src/main/java/com/powsybl/math/graph/UndirectedGraphImpl.java
```diff
@@ -586,29 +594,23 @@ private boolean traverse(int v, TraversalType traversalType, Traverser traverser
...
-                collector.addVertex(component, v);
-
-                traverse(v, TraversalType.BREADTH_FIRST, (v1, e, v2) -> {
-                    TraverseResult result = traverser.traverse(v1, e, v2);
-                    if (result == TraverseResult.CONTINUE && !encounteredVertices[v2]) {
-                        collector.addVertex(component, v2);
-                    }
-                    return result;
-                }, encounteredVertices, encounteredEdges);
+            if (vertices.get(v) != null && !encounteredVertices[v]) {
```
flo-dup: Why did you (still!) keep vertices.get(v) null check?

math/src/main/java/com/powsybl/math/graph/UndirectedGraphImpl.java
```diff
@@ -508,11 +508,15 @@ private void invalidateAdjacencyList() {
     }
 
     private void traverseVertex(int vToTraverse, boolean[] vEncountered, Deque<DirectedEdge> edgesToTraverse,
-                                TIntArrayList[] adjacencyList, TraversalType traversalType) {
+                                TIntArrayList[] adjacencyList, TraversalType traversalType, @Nullable IntConsumer vertexVisitor) {
```
flo-dup: we usually don't need nullability annotations on private methods. Besides, a nullable functional interface is creating unnecessary complexity (my bad, I suggested you to do so in a previous comment!), you'd rather use a no-op IntConsumer instead of null

math/src/test/java/com/powsybl/math/graph/UndirectedGraphImplTest.java
flo-dup, line 1: Remove the use of the deprecated `traverse` in tests

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcTopologyModel.java
```diff
@@ -239,27 +236,23 @@ private DcBusCache getCache() {
...
-                        DcBusImpl dcBus = createDcBus(dcNodeSet);
-                        dcBuses.put(dcBus.getId(), dcBus);
-                        dcNodeSet.forEach(dcNode -> dcNodeIdToDcBus.put(dcNode.getId(), dcBus));
+            List<List<DcNodeImpl>> components = graph.computeTraversalPartitions(
+                (v1, e, v2) -> {
+                    DcSwitchImpl sw = graph.getEdgeObject(e);
+                    if (sw != null && sw.isOpen()) {
+                        return TraverseResult.TERMINATE_PATH;
                     }
+                    return TraverseResult.CONTINUE;
```
flo-dup: An edge cannot be null in DcTopologyModel
```suggestion
                    return sw.isOpen() ? TraverseResult.TERMINATE_PATH : TraverseResult.CONTINUE;
```

math/src/main/java/com/powsybl/math/graph/UndirectedGraphImpl.java
```diff
@@ -548,11 +552,21 @@ public boolean traverse(int v, TraversalType traversalType, Traverser traverser,
         }
 
         boolean[] encounteredEdges = new boolean[edges.size()];
+        return traverse(v, traversalType, traverser, encounteredVertices, encounteredEdges);
+    }
+
+    private boolean traverse(int v, TraversalType traversalType, Traverser traverser, boolean[] encounteredVertices, boolean[] encounteredEdges) {
+        return traverse(v, traversalType, traverser, encounteredVertices, encounteredEdges, null);
```
flo-dup:
```suggestion
        return traverse(v, traversalType, traverser, encounteredVertices, encounteredEdges, vIndex -> {});
```

math/src/main/java/com/powsybl/math/graph/UndirectedGraph.java
```diff
@@ -328,32 +330,16 @@ public interface UndirectedGraph<V, E> {
...
     }
 
-    <C> List<C> getConnectedComponents(Traverser traverser, ConnectedComponentCollector<C> collector);
-
-    interface ConnectedComponentCollector<C> {
-        C createComponent();
-
-        void addVertex(C component, int vertexIndex);
-    }
+    <C> List<C> computeTraversalPartitions(Traverser traverser, Supplier<C> partitionFactory, ObjIntConsumer<C> vertexConsumer);
```
flo-dup: You need to add javadoc to explain clearly the method and the parameters.
Besides, rename vertexConsumer into something more explicit (vertexCollector?)

flo-dup (changes requested):

math/src/main/java/com/powsybl/math/graph/UndirectedGraphImpl.java
flo-dup: Your note should be included in the javadoc of computeTraversalPartitions

math/src/main/java/com/powsybl/math/graph/UndirectedGraphImpl.java
flo-dup: Just saw you kept it, but you were right, vertices array might contain some null values corresponding to removed vertices

flo-dup:
Could you also complete the PR description, regarding the migration guide / breaking changes part?

flo-dup (dismissed):

flo-dup (approved):

PR #3760, opened 2026-02-05
Cgmes Export: Fix multiple islands after exporting an imported UCTE case to CGMES and reloading it

### LINKED ISSUE #3755: Cgmes Export: Multiple islands after exporting an imported UCTE case to CGMES and reloading it

- A UCTE case is first imported into the system and successfully loaded.
- The network is then exported to CGMES format without errors.
- When the exported CGMES files are loaded again, the network appears split into multiple electrical islands

After the three steps, the network was expected to remain a single connected component.

### Environment

_No response_

### Relevant Log Output

_No response_

### Extra Information

Issue reported in Slack. 
The issue can be reproduced using the following case:
https://github.com/powsybl/pypowsybl/blob/main/data/19700101_0000_FO4_UX1.uct

marqueslanauja:

Fixes #3755

The network is split into multiple components after exporting an imported UCTE case to CGMES and reloading it:

- A UCTE case is first imported into the system and successfully loaded.
- The network is then exported to CGMES format without errors.
- When the exported CGMES files are loaded again, the network appears split into multiple electrical islands

The boundary terminals of the paired danglingLines were associated with different topologicalNodes.  

During the export process to CGMES, if no information about the boundary is available, the boundary terminals of the paired danglingLines are associated with the same fictitious topologicalNode. If boundary information is provided, the corresponding boundary connectivityNodeId and topologicalNodeId are selected. 

The boundary connectivityNodeId and topologicalNodeId were selected by matching the `pairingKey` of the paired danglingLines with the corresponding connectivityNode or topologicalNode `name`.

- [ ] Yes
- [x] No

**Other information**:


rcourtier (dismissed):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/TopologyExport.java
```diff
@@ -307,26 +309,75 @@ private static void writeTopologicalNodes(Network network, String cimNamespace,
...
+            // As a last resort, we create the TN in the same container of the boundary line
+            LOG.error("Boundary line {}{} is not connected to a topology node in boundaries files: EQ profile must be exported for consistent results." +
+                            " Boundary line {} is considered entirely inside voltage level {}",
+                    dl.getId(), dl.getPairingKey() != null ? " linked to X-node " + dl.getPairingKey() : "", dl.getId(), dl.getTerminal().getVoltageLevel().getId());
+            containerId = context.getNamingStrategy().getCgmesId(dl.getTerminal().getVoltageLevel());
+        }
+        String fictTopologicalNodeId = context.getNamingStrategy().getCgmesId(refTyped(dl), TOPOLOGICAL_NODE);
+        writeTopologicalNode(fictTopologicalNodeId, dl.getNameOrId() + "_NODE", containerId, baseVoltage, cimNamespace, writer, context);
+
+        return fictTopologicalNodeId;
```
colineplqt: Here I suggest using `bl` instead of `dl` now that we have `BoundaryLine`.
```suggestion
    private static String createNewTopologicalNode(BoundaryLine bl, String cimNamespace, XMLStreamWriter writer, CgmesExportContext context) throws XMLStreamException {

        String baseVoltage = context.getBaseVoltageByNominalVoltage(bl.getTerminal().getVoltageLevel().getNominalV()).getId();
        // If the EQ has also been exported, a fictitious container should have been created
        String containerId = context.getFictitiousContainerFor(b
        l);
        if (containerId == null) {
            // As a last resort, we create the TN in the same container of the boundary line
            LOG.error("Boundary line {}{} is not connected to a topology node in boundaries files: EQ profile must be exported for consistent results." +
                            " Boundary line {} is considered entirely inside voltage level {}",
                    bl.getId(), bl.getPairingKey() != null ? " linked to X-node " + bl.getPairingKey() : "", bl.getId(), bl.getTerminal().getVoltageLevel().getId());
            containerId = context.getNamingStrategy().getCgmesId(bl.getTerminal().getVoltageLevel());
        }
        String fictTopologicalNodeId = context.getNamingStrategy().getCgmesId(refTyped(bl), TOPOLOGICAL_NODE);
        writeTopologicalNode(fictTopologicalNodeId, bl.getNameOrId() + "_NODE", containerId, baseVoltage, cimNamespace, writer, context);

        return fictTopologicalNodeId;
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/TopologyExport.java
```diff
@@ -307,26 +309,75 @@ private static void writeTopologicalNodes(Network network, String cimNamespace,
...
-            }
+        for (BoundaryLine boundaryLine : network.getBoundaryLines(BoundaryLineFilter.UNPAIRED)) {
+            writeBoundaryLineTopologicalNodes(Collections.singletonList(boundaryLine), cimNamespace, writer, context);
+        }
+
+        Set<String> pairingKeys = network.getBoundaryLineStream(BoundaryLineFilter.PAIRED).map(BoundaryLine::getPairingKey).collect(Collectors.toSet());
+        for (String pairingKey : pairingKeys) {
+            List<BoundaryLine> boundaryLineList = network.getBoundaryLineStream(BoundaryLineFilter.PAIRED).filter(boundaryLine -> pairingKey.equals(boundaryLine.getPairingKey())).toList();
+            writeBoundaryLineTopologicalNodes(boundaryLineList, cimNamespace, writer, context);
+        }
```
colineplqt: Here is a simplification that allows not going twice through the stream, but it is not necessary. 

```suggestion
        Map<String, List<BoundaryLine>> boundaryLinesByPairingKey = network.getBoundaryLineStream(BoundaryLineFilter.PAIRED).collect(Collectors.groupingBy(BoundaryLine::getPairingKey));
        for (List<BoundaryLine> boundaryLineList : boundaryLinesByPairingKey.values()) {
            writeBoundaryLineTopologicalNodes(boundaryLineList, cimNamespace, writer, context);
        }
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/TopologyExport.java
colineplqt: This suggestion ends up not being very clear but I only changed the "dl" into "bl".

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/TopologyExport.java
marqueslanauja: Done. I also took the opportunity to sort the export output to make it deterministic and applied the changes to the `EquipmentExport` class.

colineplqt: Thank you

marqueslanauja:
> Thank you for your work! I left two suggestions but feel free to discard them.
> 
> I also wonder, but it is a bit out of the scope of this PR, what we should do in case a pairing key is null, in general. If a tie line has a null pairing key, then the code will fail. I wonder if it is the behaviour that we want and if not what else would be better.

You are right, the pairingKey can be null. I tried to solve this by using an `effectivePairingKey`, which is the `pairingKey` if it is not null and the `tieLine id` if it is null.

colineplqt (dismissed):

olperr1:
@marqueslanauja : could you resolve the conflicts on this PR?

marqueslanauja:
> @marqueslanauja : could you resolve the conflicts on this PR?
Done. I had to adjust the code a bit.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
```diff
@@ -1060,17 +1059,41 @@ private static String writeBoundaryLinesConnectivityNode(List<BoundaryLine> boun
...
-            connectivityNodeId = context.getNamingStrategy().getCgmesId(refTyped(boundaryLine), CONNECTIVITY_NODE);
+            Optional<String> optionalConnectivityNodeId = findConnectivityNodeId(boundaryLineList.getFirst().getPairingKey(), context);
+            connectivityNodeId = optionalConnectivityNodeId.isPresent() ? optionalConnectivityNodeId.get() : createNewConnectivityNode(boundaryLineList, baseVoltageId, cimNamespace, writer, context);
+        }
+        return connectivityNodeId;
+    }
+
+    private static Optional<String> findConnectivityNodeId(String pairingKey, CgmesExportContext context) {
+        ReferenceDataProvider referenceDataProvider = context.getReferenceDataProvider();
+        if (referenceDataProvider == null) {
```
colineplqt: I think here you should have the same as in TP export: ignore if pairingKey is null.

```suggestion
        if (pairingKey == null || referenceDataProvider == null) {
```

PR #3766, opened 2026-02-11
Avoid graph traversals in CalculatedBus get/setFictitious for NodeBreakerTopology bus views

jonenst:

No

Bug fix/Feature

getFictitious/setFictitious on calculated bus is "slow" : 50ms to call getFictitious on every bus of a country wide network in nodebreaker topology.
it does a new graph traversal for every nodebreaker node that doesn't have an equipment (terminal) on it.

The graph traversal has already been done and is already available in the busCache of the view, just reuse it.
The performance is expected to go from 50ms to 5ms.

- [ ] Yes
- [x] No

**Other information**:

Maybe  this can even refactored even further to avoid defensive copies (the voltageLevel.getNodeBreakerView().getNodes() does
```java
    @Override
    public int[] getVertices() {
        TIntArrayList t = new TIntArrayList(vertices.size());
        for (int i = 0; i < vertices.size(); i++) {
            if (vertices.get(i) != null) {
                t.add(i);
            }
        }
        return t.toArray();
    }
```
Another idea is to avoid to filter all the nodes of the voltage level at every getFictitious call to find only the ones of the bus, and instead save in the bus the list of nodes (already passed in the constructor) so that it's available immediately.


rolnico:
Covered by #3834

PR #3773, opened 2026-02-11
Evolution of the sensitivity analysis ResultWriter and Result to support multi-component computation, improve diagnostic and support cancel

### LINKED ISSUE #3626: Evolution of the sensitivity analysis ResultWriter and Result to support multi-component computation, improve diagnostic and support cancel

snoopfab:

Fixes #3626

Feature:
Enable to add multi-component support for sensitivity analysis inOpenLoadFlow
Enable to return partial results in case of interruption
Enable a better diagnosis of contingency failure

See #3626

Implemented changes described in #3626

- [x] Yes
- [ ] No

TODO

**Other information**:

### DIFF
--- a/sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
+++ b/sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
@@ -12,6 +12,7 @@ import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonToken;
 import com.powsybl.commons.PowsyblException;
 import com.powsybl.commons.json.JsonUtil;
+import com.powsybl.loadflow.LoadFlowResult;
 import org.jgrapht.alg.util.Triple;
 
 import java.io.IOException;
@@ -31,7 +32,7 @@ import java.util.*;
  * offers the possibility to calculate the sensitivities on a set of contingencies besides the pre-contingency state.
  * The full set of results consists of:
  *  - the list of factors
- *  - the list of contingencies and their associated computation status
+ *  - the list of states (contingency + optional operator strategy) and their associated computation status
  *  - the list of sensitivity values in pre-contingency and post-contingency states
  *  - the list of function reference values in pre-contingency and post-contingency states.
  *  A sensitivity analysis result offers a set of methods to retrieve sensitivity values or function reference values.
@@ -39,6 +40,7 @@ import java.util.*;
  *  and the ID of a function.
  *
  * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
+ * @author Fabrice Buscaylet {@literal <fabrice.buscaylet at artelys.com>}
  * @see SensitivityValue
  */
 public class SensitivityAnalysisResult {
@@ -51,9 +53,9 @@ public class SensitivityAnalysisResult {
 
     private final List<SensitivityStateStatus> stateStatuses;
 
-    private final List<String> contingencyIds; // to have mapping index -> ID
+    private final List<String> contingencyIds; // mapping index -> ID
 
-    private final List<String> operatorStrategyIds; // to have mapping index -> ID
+    private final List<String> operatorStrategyIds; // mapping index -> ID
 
     private final List<SensitivityValue> values;
 
@@ -65,6 +67,16 @@ public class SensitivityAnalysisResult {
 
     private final Map<SensitivityState, SensitivityStateStatus> statusByState = new HashMap<>();
 
+    private final boolean computationComplete;
+
+    /**
+     * The load flow status reported for a given component.
+     * @param status the load flow component status
+     * @param statusText the human-readable description of the status
+     */
+    public record LoadFlowStatus(LoadFlowResult.ComponentResult.Status status, String statusText) {
+    }
+
     public enum Status {
         SUCCESS,
         FAILURE,
@@ -73,28 +85,93 @@ public class SensitivityAnalysisResult {
 
     public static class SensitivityStateStatus {
 
+        public record ComponentStatus(LoadFlowStatus status, int numCC, int numSC) { }
+
+        static final String COMPONENTS_LOADFLOW_STATUSES = "componentsLoadFlowStatuses";
+        static final String LOAD_FLOW_STATUS = "loadFlowStatus";
+        static final String LOAD_FLOW_STATUS_DESCRIPTION = "loadFlowStatusDescription";
+        static final String NUM_CC = "numCC";
+        static final String NUM_CS = "numCS";
+
         private final SensitivityState state;
 
-        private final Status status;
+        /**
+         * Per-component load flow status (one entry per (numCC, numCS) for which a load flow has run).
+         */
+        private final List<ComponentStatus> componentsLoadFlowStatusList;
 
         public SensitivityState getState() {
             return state;
         }
 
+        @Deprecated(since = "7.4.0")
         public Status getStatus() {
-            return status;
+            if (!getComponentsLoadFlowStatusList().isEmpty()) {
+                switch (getComponentsLoadFlowStatusList().get(0).status().status) {
+                    case CONVERGED -> {
+                        return Status.SUCCESS;
+                    }
+                    case NO_CALCULATION -> {
+                        return Status.NO_IMPACT;
+                    }
+                    default -> {
+                        return Status.FAILURE;
+                    }
+                }
+            } else {
+                return Status.FAILURE;
+            }
         }
 
-        public SensitivityStateStatus(SensitivityState state, Status status) {
+        public List<ComponentStatus> getComponentsLoadFlowStatusList() {
+            return componentsLoadFlowStatusList;
+        }
+
+        public SensitivityStateStatus(SensitivityState state, List<ComponentStatus> statusList) {
             this.state = Objects.requireNonNull(state);
-            this.status = Objects.requireNonNull(status);
+            this.componentsLoadFlowStatusList = new ArrayList<>(statusList);
+        }
+
+        public SensitivityStateStatus(SensitivityState state) {
+            this(state, Collections.emptyList());
+        }
+
+        @Deprecated(since = "7.4.0")
+        public SensitivityStateStatus(SensitivityState state, Status status) {
+            this(state, List.of(new ComponentStatus(
+                    new LoadFlowStatus(toLoadFlowStatus(status), ""), -1, -1)));
+        }
+
+        private static LoadFlowResult.ComponentResult.Status toLoadFlowStatus(Status status) {
+            switch (status) {
+                case SUCCESS -> {
+                    return LoadFlowResult.ComponentResult.Status.CONVERGED;
+                }
+                case NO_IMPACT -> {
+                    return LoadFlowResult.ComponentResult.Status.NO_CALCULATION;
+                }
+                default -> {
+                    return LoadFlowResult.ComponentResult.Status.FAILED;
+                }
+            }
+        }
+
+        public SensitivityStateStatus addComponentLoadFlowStatus(LoadFlowStatus loadFlowStatus, int numCC, int numCS) {
+            componentsLoadFlowStatusList.add(new ComponentStatus(loadFlowStatus, numCC, numCS));
+            return this;
+        }
+
+        public SensitivityStateStatus addComponentLoadFlowStatus(LoadFlowResult.ComponentResult.Status status, int numCC, int numCS) {
+            componentsLoadFlowStatusList.add(new ComponentStatus(new LoadFlowStatus(status, ""), numCC, numCS));
+            return this;
         }
 
         public static void writeJson(JsonGenerator jsonGenerator, SensitivityStateStatus stateStatus) {
-            writeJson(jsonGenerator, stateStatus.state, stateStatus.status);
+            writeJson(jsonGenerator, stateStatus.state, stateStatus.componentsLoadFlowStatusList);
         }
 
-        public static void writeJson(JsonGenerator jsonGenerator, SensitivityState state, Status status) {
+        public static void writeJson(JsonGenerator jsonGenerator, SensitivityState state,
+                                     List<ComponentStatus> componentsLoadFlowStatusList) {
             try {
                 jsonGenerator.writeStartObject();
                 if (state.contingencyId() != null) {
@@ -103,7 +180,18 @@ public class SensitivityAnalysisResult {
                 if (state.operatorStrategyId() != null) {
                     jsonGenerator.writeStringField("operatorStrategyId", state.operatorStrategyId());
                 }
-                jsonGenerator.writeStringField("status", status.name());
+                if (componentsLoadFlowStatusList != null && !componentsLoadFlowStatusList.isEmpty()) {
+                    jsonGenerator.writeArrayFieldStart(COMPONENTS_LOADFLOW_STATUSES);
+                    for (ComponentStatus componentStatus : componentsLoadFlowStatusList) {
+                        jsonGenerator.writeStartObject();
+                        jsonGenerator.writeStringField(LOAD_FLOW_STATUS, componentStatus.status().status().toString());
+                        jsonGenerator.writeStringField(LOAD_FLOW_STATUS_DESCRIPTION, componentStatus.status().statusText());
+                        jsonGenerator.writeNumberField(NUM_CC, componentStatus.numCC());
+                        jsonGenerator.writeNumberField(NUM_CS, componentStatus.numSC());
+                        jsonGenerator.writeEndObject();
+                    }
+                    jsonGenerator.writeEndArray();
+                }
                 jsonGenerator.writeEndObject();
             } catch (IOException e) {
                 throw new UncheckedIOException(e);
@@ -114,6 +202,7 @@ public class SensitivityAnalysisResult {
             private String contingencyId;
             private String operatorStrategyId;
             private Status status;
+            private List<ComponentStatus> componentsLoadFlowStatusList;
         }
 
         public static SensitivityStateStatus parseJson(JsonParser parser, String version) {
@@ -126,8 +215,14 @@ public class SensitivityAnalysisResult {
                     if (token == JsonToken.FIELD_NAME) {
                         parseJson(parser, context, version == null ? VERSION : version);
                     } else if (token == JsonToken.END_OBJECT) {
-                        return new SensitivityStateStatus(new SensitivityState(context.contingencyId, context.operatorStrategyId),
-                                                          context.status);
+                        if (version != null && version.compareTo("1.1") <= 0) {
+                            return new SensitivityStateStatus(
+                                    new SensitivityState(context.contingencyId, context.operatorStrategyId), context.status);
+                        } else {
+                            return new SensitivityStateStatus(
+                                    new SensitivityState(context.contingencyId, context.operatorStrategyId),
+                                    context.componentsLoadFlowStatusList != null ? context.componentsLoadFlowStatusList : Collections.emptyList());
+                        }
                     }
                 }
             } catch (IOException e) {
@@ -154,13 +249,50 @@ public class SensitivityAnalysisResult {
                     context.status = Status.valueOf(parser.getValueAsString());
                     break;
                 case "status":
-                    JsonUtil.assertGreaterOrEqualThanReferenceVersion(CONTEXT_NAME, "Tag: status", version, "1.1");
+                    JsonUtil.assertLessThanOrEqualToReferenceVersion(CONTEXT_NAME, "Tag: status", version, "1.1");
                     parser.nextToken();
                     context.status = Status.valueOf(parser.getValueAsString());
+                    break;
+                case COMPONENTS_LOADFLOW_STATUSES:
+                    JsonUtil.assertGreaterOrEqualThanReferenceVersion(CONTEXT_NAME, "Tag: " + COMPONENTS_LOADFLOW_STATUSES, version, "1.2");
+                    context.componentsLoadFlowStatusList = parseComponentLoadFlowStatuses(parser);
                     break;
                 default:
                     throw new PowsyblException("Unexpected field: " + fieldName);
             }
+        }
+
+        private static List<ComponentStatus> parseComponentLoadFlowStatuses(JsonParser parser) throws IOException {
+            if (parser.nextToken() != JsonToken.START_ARRAY) {
+                throw new PowsyblException("Expected start of array for component loadflow statuses");
+            }
+            List<ComponentStatus> statuses = new ArrayList<>();
+            while (parser.nextToken() != JsonToken.END_ARRAY) {
+                if (parser.currentToken() == JsonToken.START_OBJECT) {
+                    statuses.add(parseSingleComponentStatus(parser));
+                }
+            }
+            return statuses;
+        }
+
+        private static ComponentStatus parseSingleComponentStatus(JsonParser parser) throws IOException {
+            String statusStr = null;
+            String descStr = null;
+            int numCC = 0;
+            int numCS = 0;
+            while (parser.nextToken() != JsonToken.END_OBJECT) {
+                String fieldName = parser.currentName();
+                parser.nextToken();
+                switch (fieldName) {
+                    case LOAD_FLOW_STATUS -> statusStr = parser.getText();
+                    case LOAD_FLOW_STATUS_DESCRIPTION -> descStr = parser.getText();
+                    case NUM_CC -> numCC = parser.getIntValue();
+                    case NUM_CS -> numCS = parser.getIntValue();
+                    default -> parser.skipChildren();
+                }
+            }
+            LoadFlowStatus lfs = new LoadFlowStatus(LoadFlowResult.ComponentResult.Status.valueOf(statusStr), descStr);
+            return new ComponentStatus(lfs, numCC, numCS);
         }
     }
 
@@ -174,6 +306,20 @@ public class SensitivityAnalysisResult {
      */
     public SensitivityAnalysisResult(List<SensitivityFactor> factors, List<SensitivityStateStatus> stateStatuses, List<String> contingencyIds,
                                      List<String> operatorStrategyIds, List<SensitivityValue> values) {
+        this(factors, stateStatuses, contingencyIds, operatorStrategyIds, values, true);
+    }
+
+    /**
+     * Sensitivity analysis result
+     * @param factors the list of sensitivity factors that have been computed.
+     * @param stateStatuses the list of states and their associated computation status.
+     * @param contingencyIds the list of contingency IDs that have been considered during the sensitivity analysis.
+     * @param operatorStrategyIds the list of operator strategy IDs that have been considered during the sensitivity analysis.
+     * @param values result values of the sensitivity analysis in pre-contingency state and post-contingency states.
+     * @param computationComplete whether the sensitivity computation fully or partially completed
+     */
+    public SensitivityAnalysisResult(List<SensitivityFactor> factors, List<SensitivityStateStatus> stateStatuses, List<String> contingencyIds,
+                                     List<String> operatorStrategyIds, List<SensitivityValue> values, boolean computationComplete) {
         this.factors = Collections.unmodifiableList(Objects.requireNonNull(factors));
         this.stateStatuses = Collections.unmodifiableList(Objects.requireNonNull(stateStatuses));
         this.contingencyIds = Collections.unmodifiableList(Objects.requireNonNull(contingencyIds));
@@ -193,6 +339,7 @@ public class SensitivityAnalysisResult {
         for (SensitivityStateStatus stateStatus : stateStatuses) {
             this.statusByState.put(stateStatus.getState(), stateStatus);
         }
+        this.computationComplete = computationComplete;
     }
 
     /**
@@ -635,13 +782,34 @@ public class SensitivityAnalysisResult {
     }
 
     /**
-     * Get the status associated to a state
+     * Get the status associated to a state.
      *
-     * @param state The state
-     * @return The associated status.
+     * @param state the considered state.
+     * @return the associated status.
+     * @deprecated Use {@link SensitivityAnalysisResult#getStateComponentStatus(SensitivityState)} instead.
      */
+    @Deprecated(since = "7.4.0")
     public Status getStateStatus(SensitivityState state) {
         Objects.requireNonNull(state);
         return statusByState.get(state).getStatus();
+    }
+
+    /**
+     * Get the status associated to a state for all components
+     *
+     * @param state the considered state.
+     * @return the components' status.
+     */
+    public List<SensitivityStateStatus.ComponentStatus> getStateComponentStatus(SensitivityState state) {
+        Objects.requireNonNull(state);
+        return statusByState.get(state).getComponentsLoadFlowStatusList();
+    }
+
+    /**
+     * Return true if the computation was fully completed, false if it was partially completed
+     * @return the computation complete boolean
+     */
+    public boolean isComputationComplete() {
+        return computationComplete;
     }
 }
--- a/sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisTool.java
+++ b/sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisTool.java
@@ -189,6 +189,11 @@ public class SensitivityAnalysisTool implements Tool {
             if (line.hasOption(SINGLE_OUTPUT)) {
                 throw new PowsyblException("Unsupported " + SINGLE_OUTPUT + " option does not support csv file as argument of " + OUTPUT_FILE_OPTION + ". Must be json.");
             }
+        } else {
+            // json format
+            if (line.hasOption(OUTPUT_STATE_STATUS_FILE_OPTION)) {
+                throw new PowsyblException(OUTPUT_STATE_STATUS_FILE_OPTION + " file is not supported in json");
+            }
         }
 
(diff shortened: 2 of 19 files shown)


sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityResultCsvWriter.java
```diff
@@ -68,12 +87,33 @@ public void writeSensitivityValue(int factorIndex, int contingencyIndex, double
...
+            formatterPreContingency.writeCell(loadFlowStatus.status().toString());
+            formatterPreContingency.writeCell(loadFlowStatus.statusText());
         } catch (IOException e) {
             throw new UncheckedIOException(e);
         }
     }
+
+    @Override
+    public void computationComplete() {
+        // WHAT TO DO HERE ??
```
alicecaron: set to computationComplete to true like for the json writer?

snoopfab: Hello Alice. 

The goal of the writer is to write the information on several csv files (unlike the json one that can handle all information in one single file). I'm a little bit reluctant to create a computation.csv with just a computationComplete column with a true value in one line hence my question. Maybe adding some other global statistics on the computation for the future in the computationComplete( Map<String,Double> computationStats) would have sense in a future PR ?

snoopfab:
I rebased the `sensitivity-result-writer-enhancements` branches of **powsybl-core** and **powsybl-open-loadflow** onto `main`.

The rebases were really not trivial (an automatic merge was impossible).

While my branches were under review, Geoffroy’s PR **#3670** (operator strategies in the sensitivity API) was merged into core, along with its OLF counterpart **#1326**. These two PRs touched the same `SensitivityResultWriter` interface, the same `SensitivityAnalysisResult` class, and the same provider implementations as my work on multi‑component reporting — so we ended up with two parallel refactorings of the same surface area.

Rather than keeping the two APIs side by side, I merged my additions into the interface redesigned by Geoffroy. His design is centered around the new `SensitivityState` (a contingency + an optional operator strategy).

My additions — per‑component load‑flow status and partial result reporting in case of interruption — are a cross‑cutting concern that integrates naturally into this state.

So concretely, for the writer interface:

*   `writeSensitivityValue` gained the operator strategy index via #3670; nothing new on my side.
*   `writeContingencyStatus` and the post‑#3670 3‑argument `writeStateStatus` have both disappeared. They are replaced by:
        writeStateStatus(int contingencyIndex, int operatorStrategyIndex,
                         Status, LoadFlowStatus, int numCC, int numCS)
    This is a merge of his “which state failed?” with my “which load‑flow status on which component?”  
    Pre‑contingency reporting is done with both indices set to `-1`, which replaces my former `writeSynchronousComponentStatus`.
*   `computationComplete()` was added.

The final proposed contract is therefore:

```java
/**
 * @author Fabrice Buscaylet {@literal <fabrice.buscaylet at artelys.com}
 */
public interface SensitivityResultWriter {

    /**
     * In the case of multi-component computation, a factor has a variable
     * that is linked to an equipment that belongs to a single component.
     * Called once per factor and contingency in the factor context that
     * belongs to the same component as the factor equipment.
     * NaN for invalid factors whose equipment does not belong to a component
     * that has been run.
     *
     * @param factorIndex the factor index
     * @param contingencyIndex the contingency index, -1 for pre-contingency state
     * @param operatorStrategyIndex the operator strategy index, -1 if none
     * @param value the sensitivity value
     * @param functionReference the function reference
     */
    void writeSensitivityValue(int factorIndex, int contingencyIndex,
                               int operatorStrategyIndex, double value,
                               double functionReference);

    /**
     * Reports the status for a given state (contingency + optional operator strategy)
     * and a given component (numCC/numCS).
     * <p>
     * Called for every state and every numCC/numCS where the state has an impact.
     * For pre-contingency reporting, both {@code contingencyIndex} and
     * {@code operatorStrategyIndex} are -1.
     * For contingencies that are never run, called once in the end with NO_IMPACT
     * and numCC and numCS set to -1.
     *
     * @param contingencyIndex the contingency index, -1 for pre-contingency state
     * @param operatorStrategyIndex the operator strategy index, -1 if none
     * @param status the sensitivity analysis status
     * @param loadFlowStatus the load flow status for this component
     * @param numCC index of connected component, -1 if not applicable
     * @param numCS index of synchronous component, -1 if not applicable
     */
    void writeStateStatus(int contingencyIndex, int operatorStrategyIndex,
                          SensitivityAnalysisResult.Status status,
                          SensitivityAnalysisResult.LoadFlowStatus loadFlowStatus,
                          int numCC, int numCS);

    /**
     * Called at the end of the computation if the computation has not been interrupted.
     */
    void computationComplete();
}
```

On the result model side:

*   `SensitivityContingencyStatus` has disappeared and is replaced by `SensitivityStateStatus` from #3670, and I moved the per‑component load‑flow list onto this class.
*   `SensitivityPreContingencyStatus` has also disappeared; the pre‑contingency status is now a normal entry in `stateStatuses`, keyed by `SensitivityState.PRE_CONTINGENCY`.
*   The result JSON is bumped to version **1.2**; readers for v1.0 and v1.1 continue to work via the existing version‑aware deserializer.

On the OLF side:

*   AC analysis received the full multi‑component restructuring (loop per component via `getNetworksToSimulate`, aggregation of `NO_IMPACT` across all components, proper reporting of pre‑contingency load‑flow failure).
*   DC analysis received the same treatment. Operator strategies remain fully supported on both sides.

The tests pass: **61/61** on core, **284/284** on the OLF sensi package.

Two unrelated failures appeared, which look like upstream issues related to the core bump **7.2.0 → 7.3.0‑SNAPSHOT**:

*   `ZeroImpedanceFlowsTest` (the load flow gives 2.9 MW where 1.4 MW is expected, possibly due to the switch to JGraphT 1.5.3),
*   `DcSensitivityAnalysisContingenciesTest.testDebug` (XML IIDM format drift; the reference file just needs to be regenerated).

Practically speaking, since the commits were squashed and the branches had already been pushed to GitHub, the remote history no longer matched what I had locally. I therefore force‑pushed on the same branch.

sensitivity-analysis-api/src/test/resources/SensitivityAnalysisResultRefV1.2.json
```diff
@@ -0,0 +1,57 @@
...
+    "value" : 3.0,
+    "functionReference" : 4.0
+  } ],
+  "stateStatus" : [ {
+    "contingencyId" : "NHV1_NHV2_2",
+    "status" : "SUCCESS"
+  } ],
+  "contingencyIds" : [ "NHV1_NHV2_2" ],
+  "operatorStrategyIds" : [ ]
+}
```
alicecaron, line 117: Shouldn't we also add a test `testCanReadResult11()` to check that we can still read a 1.1 version also? => if yes maybe create a `@ParameterizedTest`

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
```diff
@@ -12,6 +12,7 @@
 import com.fasterxml.jackson.core.JsonToken;
```
alicecaron, line 12: We can update the documentation about the output [here](https://powsybl.readthedocs.io/projects/powsybl-core/en/stable/simulation/sensitivity/index.html#outputs)

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
```diff
@@ -73,10 +83,21 @@ public enum Status {
...
+        static final String NUM_CS = "numCS";
+
         private final SensitivityState state;
 
         private final Status status;
 
+        /**
+         * Per-component load flow status (one entry per (numCC, numCS) for which a load flow has run).
+         */
+        private final List<Triple<LoadFlowStatus, Integer, Integer>> componentsLoadFlowStatusList;
```
vidaldid-rte: Maybe use a record instead of a triple, to store status, ccnum,csnum.
Would be more convenient to use.

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
```diff
@@ -85,16 +106,30 @@ public Status getStatus() {
             return status;
         }
 
+        public List<Triple<LoadFlowStatus, Integer, Integer>> getComponentsLoadFlowStatusList() {
+            return componentsLoadFlowStatusList;
+        }
+
         public SensitivityStateStatus(SensitivityState state, Status status) {
```
vidaldid-rte: Because the Status is possibly in contradiction with the more detailed componentsLoadFlowStatusList,
this constructor should be deprecated (but might be kept to minimize breaking changes)
It would add to the componentsLoadFlowStatus a record equivalent to the status (maybe with numcc and numcs set to -1 indicating they were not provided)

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
```diff
@@ -73,10 +83,21 @@ public enum Status {
...
 
+        static final String COMPONENTS_LOADFLOW_STATUSES = "componentsLoadFlowStatuses";
+        static final String LOAD_FLOW_STATUS = "loadFlowStatus";
+        static final String LOAD_FLOW_STATUS_DESCRIPTION = "loadFlowStatusDescription";
+        static final String NUM_CC = "numCC";
+        static final String NUM_CS = "numCS";
+
         private final SensitivityState state;
 
         private final Status status;
```
vidaldid-rte: That field is redundant with the componentLoadFlowStatusList. 
It can be a source of conflict with the componentLoadFlowStatusList, providing contradictory information.

I think it should be removed.
In this case, getter could be deprecated, but return a value derived from the component status list.

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
vidaldid-rte, line 108: If we decide to remove the Status field (that might be in contradiction with the componentsLoadFlowStatusList,
this method should be deprecated. And return a Status derived from the list
(for example derived from the first element in componentsLoadFlowStatusList. In this case this would be equivalent to the single component security analysis computation, and ensure bacward compatibility)

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/json/SensitivityAnalysisResultDeserializer.java
```diff
@@ -42,7 +42,7 @@ public SensitivityAnalysisResult deserialize(JsonParser parser, DeserializationC
         while (parser.nextToken() != JsonToken.END_OBJECT) {
             switch (parser.currentName()) {
                 case "version":
-                    parser.nextToken(); // skip
+                    parser.nextToken();
```
vidaldid-rte, line 45: Why did you remove the comment ? Is it no longer relevant ?

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisTool.java
```diff
@@ -187,12 +186,17 @@ public void run(CommandLine line, ToolRunningContext context) throws Exception {
             if (line.hasOption(SINGLE_OUTPUT)) {
                 throw new PowsyblException("Unsupported " + SINGLE_OUTPUT + " option does not support csv file as argument of " + OUTPUT_FILE_OPTION + ". Must be json.");
             }
+        } else {
+            // json format
+            if (line.hasOption(OUTPUT_STATE_STATUS_FILE_OPTION)) {
+                throw new PowsyblException(OUTPUT_STATE_STATUS_FILE_OPTION + " file is not supported in json");
```
vidaldid-rte, line 195: Is this a new limitation ?
If yes, why ?

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityResultCsvWriter.java
```diff
@@ -57,7 +58,11 @@ public static TableFormatter createContingencyStatusTableFormatter(Writer writer
         return factory.create(writer, "Sensitivity analysis status result", tfc,
                 new Column("Contingency ID"),
                 new Column("Operator strategy ID"),
-                new Column("Status"));
+                new Column("Status"),
```
vidaldid-rte: Do we still need Status now that we have LoadFlowStatus ?

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityResultCsvWriter.java
```diff
@@ -76,13 +81,28 @@ public void writeSensitivityValue(int factorIndex, int contingencyIndex, int ope
...
 
     @Override
-    public void writeStateStatus(int contingencyIndex, int operatorStrategyIndex, SensitivityAnalysisResult.Status status) {
+    public void writeStateStatus(int contingencyIndex, int operatorStrategyIndex, SensitivityAnalysisResult.Status status,
+                                 SensitivityAnalysisResult.LoadFlowStatus loadFlowStatus, int numCC, int numCS) {
         try {
             formatterStatus.writeCell(contingencyIndex != -1 ? contingencies.get(contingencyIndex).getId() : "");
             formatterStatus.writeCell(operatorStrategyIndex != -1 ? operatorStrategies.get(operatorStrategyIndex).getId() : "");
             formatterStatus.writeCell(status.name());
+            if (loadFlowStatus != null) {
```
vidaldid-rte: If we remove the status (that can conflict with the LoadFlowStatus list for all component) the loadFlow status should be mandatory.

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityResultWriter.java
```diff
@@ -9,10 +9,49 @@
...
+     * and numCC and numCS set to -1.
+     *
+     * @param contingencyIndex the contingency index, -1 for pre-contingency state
+     * @param operatorStrategyIndex the operator strategy index, -1 if none
+     * @param status the sensitivity analysis status
+     * @param loadFlowStatus the load flow status for this component
+     * @param numCC index of connected component, -1 if not applicable
+     * @param numCS index of synchronous component, -1 if not applicable
+     */
+    void writeStateStatus(int contingencyIndex, int operatorStrategyIndex, SensitivityAnalysisResult.Status status,
```
vidaldid-rte: The SensitivityAnalysisResult.Status is redundant with the more detailed LoadFlowStatus. It should be removed from the argument list.

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityResultModelWriter.java
```diff
@@ -38,15 +41,30 @@ public List<SensitivityAnalysisResult.SensitivityStateStatus> getStateStatuses()
...
-                                                      operatorStrategyIndex != -1 ? operatorStrategies.get(operatorStrategyIndex).getId() : null);
-        stateStatuses.put(state, new SensitivityAnalysisResult.SensitivityStateStatus(state, status));
+    public void writeStateStatus(int contingencyIndex, int operatorStrategyIndex, SensitivityAnalysisResult.Status status,
+                                 SensitivityAnalysisResult.LoadFlowStatus loadFlowStatus, int numCC, int numCS) {
+        SensitivityState state = new SensitivityState(
+                contingencyIndex != -1 ? contingencies.get(contingencyIndex).getId() : null,
+                operatorStrategyIndex != -1 ? operatorStrategies.get(operatorStrategyIndex).getId() : null);
+        SensitivityAnalysisResult.SensitivityStateStatus stateStatus = stateStatuses.computeIfAbsent(
+                state, k -> new SensitivityAnalysisResult.SensitivityStateStatus(k, status));
+        if (loadFlowStatus != null) {
```
vidaldid-rte: Why do you authorize a null value ?
And the

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityResultModelWriter.java
```diff
@@ -38,15 +41,30 @@ public List<SensitivityAnalysisResult.SensitivityStateStatus> getStateStatuses()
...
-    public void writeStateStatus(int contingencyIndex, int operatorStrategyIndex, SensitivityAnalysisResult.Status status) {
-        SensitivityState state = new SensitivityState(contingencyIndex != -1 ? contingencies.get(contingencyIndex).getId() : null,
-                                                      operatorStrategyIndex != -1 ? operatorStrategies.get(operatorStrategyIndex).getId() : null);
-        stateStatuses.put(state, new SensitivityAnalysisResult.SensitivityStateStatus(state, status));
+    public void writeStateStatus(int contingencyIndex, int operatorStrategyIndex, SensitivityAnalysisResult.Status status,
+                                 SensitivityAnalysisResult.LoadFlowStatus loadFlowStatus, int numCC, int numCS) {
+        SensitivityState state = new SensitivityState(
+                contingencyIndex != -1 ? contingencies.get(contingencyIndex).getId() : null,
+                operatorStrategyIndex != -1 ? operatorStrategies.get(operatorStrategyIndex).getId() : null);
+        SensitivityAnalysisResult.SensitivityStateStatus stateStatus = stateStatuses.computeIfAbsent(
```
vidaldid-rte: This should not be done as it puts a general status independant of the numCC/numCS

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
vidaldid-rte, line 307: The SensitivityAnalysisResult should contain the computationComplete flag, so that the user can determine if the securityAnalysis computation is a paertial result (interrupted) or a complete result.

vidaldid-rte (changes requested):
Thanks for the implementation.
My comments are mostly about the the old SensitivityStateStatus that is potentially conflicting with the more detailed LoadFlowStatus per CC/CS.
This field should be removed.
Suggestions to minimize API breaking changes and ensuring data consistency.

Also, is there a branch in OpenLoadFlow that tests this implementation and that proves that nothing is missing ?

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisTool.java
obrix: It seems it is described as csv only in the options. So not a new limitation but an added check.

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/json/SensitivityAnalysisResultDeserializer.java
jeandemanged: i think it is not relevant

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
```diff
@@ -635,13 +782,34 @@ public double getBusVoltageFunctionReferenceValue(String functionId) {
...
+     * @return the components' status.
+     */
+    public List<SensitivityStateStatus.ComponentStatus> getStateComponentStatus(SensitivityState state) {
+        Objects.requireNonNull(state);
+        return statusByState.get(state).getComponentsLoadFlowStatusList();
+    }
+
+    /**
+     * Return true if the computation was fully completed, false if it was partially completed
+     * @return the computation complete boolean
```
jeandemanged, line 810:
```suggestion
     * Return true if the computation was fully completed, false if it was partially completed because of interruption. Interrupted calculations contain only partial results.
     * @return true if the computation completed without interruption, i.e. the results are complete.
```

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
```diff
@@ -65,6 +67,16 @@ public class SensitivityAnalysisResult {
...
+
+    /**
+     * The load flow status reported for a given component.
+     * @param status the load flow component status
+     * @param statusText the human-readable description of the status
+     */
+    public record LoadFlowStatus(LoadFlowResult.ComponentResult.Status status, String statusText) {
+    }
+
     public enum Status {
```
jeandemanged, line 80: In the discussions it was decided to deprecate this status and use LoadFlowResult.ComponentResult.Status instead. This enum should be marked as deprecated with description of the new way of doing.

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
```diff
@@ -65,6 +67,16 @@ public class SensitivityAnalysisResult {
...
     private final Map<SensitivityState, SensitivityStateStatus> statusByState = new HashMap<>();
 
+    private final boolean computationComplete;
+
+    /**
+     * The load flow status reported for a given component.
+     * @param status the load flow component status
+     * @param statusText the human-readable description of the status
+     */
+    public record LoadFlowStatus(LoadFlowResult.ComponentResult.Status status, String statusText) {
```
jeandemanged, line 77: I think it would be much simpler if this gets merged in SensitivityStateStatus.ComponentStatus below directly.
Also I would put CC and SC as the first arguments since these are the main attributes.
Finally for consistency wwith LoadFlow Results API use `connectedComponentNum`  and `synchronousComponentNum`.

something like:

```java
public record ComponentStatus(
    int connectedComponentNum,
    int synchronousComponentNum,
    LoadFlowResult.ComponentResult.Status status, String statusText) { }
```

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
```diff
@@ -73,28 +85,93 @@ public enum Status {
 
     public static class SensitivityStateStatus {
 
+        public record ComponentStatus(LoadFlowStatus status, int numCC, int numSC) { }
+
+        static final String COMPONENTS_LOADFLOW_STATUSES = "componentsLoadFlowStatuses";
```
jeandemanged, line 90: I'd remove 'loadflow', and the double plural is confusing


```suggestion
        static final String COMPONENT_STATUSES = "componentStatuses";
```

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
```diff
@@ -73,28 +85,93 @@ public enum Status {
...
+        static final String LOAD_FLOW_STATUS = "loadFlowStatus";
+        static final String LOAD_FLOW_STATUS_DESCRIPTION = "loadFlowStatusDescription";
```
jeandemanged, line 92: elsewhere it is statusText, not statusDescirption


```suggestion
        static final String LOAD_FLOW_STATUS_TEXT = "loadFlowStatusText";
```

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
```diff
@@ -73,28 +85,93 @@ public enum Status {
...
+        static final String NUM_CC = "numCC";
+        static final String NUM_CS = "numCS";
```
jeandemanged, line 94: Unlike previous comment, here it is for JSON and I am fine using short string and not connectedComponentNum and synchronousComponentNum: this will keep JSON more compact.

However CS sounds like french Composante Synchrone and must be fixed to be english Synchronous Component. This must be fixed a little bit everywhere in the PR.

```suggestion
        static final String NUM_CC = "numCC";
        static final String NUM_SC = "numSC";
```

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
```diff
@@ -73,28 +85,93 @@ public enum Status {
...
+
+        public SensitivityStateStatus addComponentLoadFlowStatus(LoadFlowStatus loadFlowStatus, int numCC, int numCS) {
+            componentsLoadFlowStatusList.add(new ComponentStatus(loadFlowStatus, numCC, numCS));
+            return this;
+        }
+
+        public SensitivityStateStatus addComponentLoadFlowStatus(LoadFlowResult.ComponentResult.Status status, int numCC, int numCS) {
+            componentsLoadFlowStatusList.add(new ComponentStatus(new LoadFlowStatus(status, ""), numCC, numCS));
+            return this;
         }
```
jeandemanged, line 167: In LoadFlow and Security Analysis Results API, and here too before this PR, the XxxxResult object is immutable. But with this change this is not true anymore. I think this should be done differently.

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
```diff
@@ -103,7 +180,18 @@ public static void writeJson(JsonGenerator jsonGenerator, SensitivityState state
...
+                if (componentsLoadFlowStatusList != null && !componentsLoadFlowStatusList.isEmpty()) {
+                    jsonGenerator.writeArrayFieldStart(COMPONENTS_LOADFLOW_STATUSES);
+                    for (ComponentStatus componentStatus : componentsLoadFlowStatusList) {
+                        jsonGenerator.writeStartObject();
+                        jsonGenerator.writeStringField(LOAD_FLOW_STATUS, componentStatus.status().status().toString());
+                        jsonGenerator.writeStringField(LOAD_FLOW_STATUS_DESCRIPTION, componentStatus.status().statusText());
+                        jsonGenerator.writeNumberField(NUM_CC, componentStatus.numCC());
+                        jsonGenerator.writeNumberField(NUM_CS, componentStatus.numSC());
+                        jsonGenerator.writeEndObject();
+                    }
```
jeandemanged, line 192: see also other comments on java part, this to be reviewed accordingly:
- https://github.com/powsybl/powsybl-core/pull/3773/changes#r4045996552: will fix the weird `status().status()`, write NUM_CC and NUM_SC first
- https://github.com/powsybl/powsybl-core/pull/3773/changes#r4046063536 typo CS -> SC

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
```diff
@@ -126,8 +215,14 @@ public static SensitivityStateStatus parseJson(JsonParser parser, String version
                     if (token == JsonToken.FIELD_NAME) {
                         parseJson(parser, context, version == null ? VERSION : version);
                     } else if (token == JsonToken.END_OBJECT) {
-                        return new SensitivityStateStatus(new SensitivityState(context.contingencyId, context.operatorStrategyId),
-                                                          context.status);
+                        if (version != null && version.compareTo("1.1") <= 0) {
```
jeandemanged, line 218: use `JsonUtil.compareVersions`, string comparison will break after 1.10

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
```diff
@@ -154,14 +249,51 @@ private static void parseJson(JsonParser parser, SensitivityStateStatus.ParsingC
                     context.status = Status.valueOf(parser.getValueAsString());
                     break;
                 case "status":
-                    JsonUtil.assertGreaterOrEqualThanReferenceVersion(CONTEXT_NAME, "Tag: status", version, "1.1");
+                    JsonUtil.assertLessThanOrEqualToReferenceVersion(CONTEXT_NAME, "Tag: status", version, "1.1");
```
jeandemanged, line 252: I think this should accept only 1.1 and not 1.0 ?

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisResult.java
```diff
@@ -635,13 +782,34 @@ public double getBusVoltageFunctionReferenceValue(String functionId) {
...
         return statusByState.get(state).getStatus();
     }
+
+    /**
+     * Get the status associated to a state for all components
+     *
+     * @param state the considered state.
+     * @return the components' status.
+     */
+    public List<SensitivityStateStatus.ComponentStatus> getStateComponentStatus(SensitivityState state) {
```
jeandemanged, line 803: similar to a previous comment, also the javadoc seems incorrect


```suggestion
    public List<SensitivityStateStatus.ComponentStatus> getStateComponentStatuses(SensitivityState state) {
```

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisTool.java
jeandemanged: that branch is hit for anything not CSV, this includes JSON but also XML or whatever. Exception message to be adjusted to say that only CSV is supported.

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityResultCsvWriter.java
```diff
@@ -57,7 +58,10 @@ public static TableFormatter createContingencyStatusTableFormatter(Writer writer
         return factory.create(writer, "Sensitivity analysis status result", tfc,
                 new Column("Contingency ID"),
                 new Column("Operator strategy ID"),
-                new Column("Status"));
+                new Column("Loadflow Status"),
+                new Column("Loadflow Status Description"),
+                new Column("Connected component"),
+                new Column("Synchronous component"));
```
jeandemanged, line 64: Loadflow -> Load flow
Status Description -> Status Text
Connected component and Synchronous component columns before LF status

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityResultWriter.java
```diff
@@ -9,10 +9,42 @@
...
+     * and numCC and numCS set to -1.
+     *
+     * @param contingencyIndex the contingency index, -1 for pre-contingency state
+     * @param operatorStrategyIndex the operator strategy index, -1 if none
+     * @param loadFlowStatus the load flow status for this component
+     * @param numCC index of connected component, -1 if not applicable
+     * @param numCS index of synchronous component, -1 if not applicable
+     */
+    void writeStateStatus(int contingencyIndex, int operatorStrategyIndex,
+                          SensitivityAnalysisResult.LoadFlowStatus loadFlowStatus, int numCC, int numCS);
```
jeandemanged, line 49: see other comments
```
    void writeStateStatus(int contingencyIndex, int operatorStrategyIndex,
                          int connectedComponentNum, int synchronousComponentNum,
                         etc status, statusText);
```

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityAnalysisTool.java
alicecaron: Should the check throw an exception?

jeandemanged: Indeed it could be just a warning that this option is ignored in the case of JSON

obrix:
@jeandemanged @alicecaron @So-Fras  Thanks for the review. I mainly resolved all the conflicts and the early comments here. Please note that the existing prototype of implementation in olf needs to be stabilized before trying to merge this. I will take into account the comments as I continue to try to make all of this working.

PR #3776, opened 2026-02-17
Cgmes export doesnt alter the network

### LINKED ISSUE #3632: CGMES export shouldn't alter the network

CGMES export relies on the presence of aliases and properties to identify the exported CGMES objects. However, these aliases and properties are not always present, for instance if the network doesn't come from a CGMES import, or if the network has been expanded with the addition of new elements. That's why currently, a method: `addIidmMappings` is called at the very begin of the export process to ensure these properties and aliases are present. In the case they are not present, they are added, which constitutes a modification of the network and that is something that needs to be avoided.

CGMES export doesn't alter the network. If properties or aliases to identify CGMES objects are missing, they should be generated in the export classes themselves on the fly and not be stored in the network. Ultimately, the `addIidmMappings` method should be removed.

### Describe the motivation

The network should remain unchanged by any export or serialization process.

### Extra Information

_No response_

rcourtier:

Fixes #3632.

Refactoring.

CGMES export adds properties, aliases and extensions to the network.

CGMES export doesn't alter the network.

There is also one functional change: in case there is no `CgmesTapChangers` extension to an IIDM `TapChanger`, the CGMES TapChanger.normalStep is now defaulted to the neutralStep instead of the current step.

- [x] Yes
- [ ] No

Some constants names(`public static final String`) have been renamed in `Conversion` class of `com.powsybl.cgmes.conversion` package:
- `PROPERTY_CGMES_GOVERNOR_SCD` is now `PROPERTY_GOVERNOR_SCD`
- `PROPERTY_CGMES_SYNCHRONOUS_MACHINE_TYPE` is now `PROPERTY_SYNCHRONOUS_MACHINE_TYPE`
- `PROPERTY_CGMES_REGULATION_CAPABILITY` is now `PROPERTY_REGULATION_CAPABILITY`
- `PROPERTY_CGMES_DC_CONVERTER_UNIT` is now `PROPERTY_DC_CONVERTER_UNIT`

The constant values remain identical.

 The following public methods were removed:
 - `CgmesExportContext.addIidmMappings(Network network)`
 - `CgmesExportContext.getBaseVoltageByNominalVoltage(double nominalV)`
 - `CgmesExportContext.getRegionsIds()`
 - `CgmesExportContext.getRegionName(String regionId)`
 - `CgmesExportContext.getSubRegionName(String subRegionId)`
 - `CgmesExportUtil.cgmesTapChangerType(C eq, String tcId)`
 - `CgmesExportUtil.setCgmesTapChangerType(C eq, String tapChangerId, String type)`
 - `CgmesExportUtil.getTerminalId(Terminal t, CgmesExportContext context)`

`RegulatingControlEq.writeRegulatingControlEq(...)` changed:
- signature:
  - its return type is now `void` instead of `String`
  - its `Set<String> regulatingControlsWritten` was replaced by `String regulatingControlId`
- behavior:
  - the `regulatingControlId` is no more retrieved from the naming strategy but given as a parameter
 

**Other information**:

2 tests were conducted on a large network (RTE France):
* Starting from a network with no property/alias/extension
  * Export it to XIIDM
  * Export it to CGMES
  * Export it again to XIIDM and compare to the first IIDM export
  * The networks were identical: no property/alias/extension has been created (idempotency)
* Starting from a network with no property/alias/extension
  * Export from main branch
  * Export from this branch
  * Compare the CGMES files
  * The generated EQ (except for TapChanger.normalStep), SSH, TP profiles were identical (IDs persistency)


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
```diff
@@ -1138,18 +1138,68 @@ public Config setUseDetailedDcModel(boolean useDetailedDcModel) {
...
+    public static final String PROPERTY_SYNCHRONOUS_MACHINE_TYPE = CGMES_PREFIX_ALIAS_PROPERTIES + "synchronousMachineType";
+    public static final String PROPERTY_TERMINAL = CGMES_PREFIX_ALIAS_PROPERTIES + TERMINAL;
+    public static final String PROPERTY_TERMINAL_BOUNDARY = CGMES_PREFIX_ALIAS_PROPERTIES + TERMINAL_BOUNDARY;
+    public static final String PROPERTY_TERMINAL_SIGN = CGMES_PREFIX_ALIAS_PROPERTIES + TERMINAL_SIGN;
+    public static final String PROPERTY_TERMINAL_SIGN1 = CGMES_PREFIX_ALIAS_PROPERTIES + TERMINAL_SIGN1;
+    public static final String PROPERTY_TERMINAL_SIGN2 = CGMES_PREFIX_ALIAS_PROPERTIES + TERMINAL_SIGN2;
+    public static final String PROPERTY_TERMINAL_SIGN3 = CGMES_PREFIX_ALIAS_PROPERTIES + TERMINAL_SIGN3;
+    public static final String PROPERTY_TOPOLOGICAL_NODE_BOUNDARY = CGMES_PREFIX_ALIAS_PROPERTIES + TOPOLOGICAL_NODE_BOUNDARY;
+    public static final String PROPERTY_TP_BD_ID = CGMES_PREFIX_ALIAS_PROPERTIES + "TP_BD_ID";
+    public static final String PROPERTY_WIND_GEN_UNIT_TYPE = CGMES_PREFIX_ALIAS_PROPERTIES + "windGenUnitType";
```
olperr1, line 1206: In this class, you can also add static methods that generates the properties that depends on variables. This way, all your property names will be handled by this class.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/naming/NamingStrategy.java
```diff
@@ -41,6 +52,46 @@ default String getCgmesId(String identifier) {
 
     String getCgmesId(CgmesObjectReference... refs);
```
olperr1, line 53: Since this method is not called once per network element at the beginning of the export, but every time the id is needed, the result should always be the same when called several times with the same arguments.

It could be useful to mention this fact as a Javadoc comment in this interface, in the PR's description, and in the migration guide.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractTransformerConversion.java
```diff
@@ -188,7 +186,8 @@ static <C extends Connectable<C>> void updatePhaseTapChanger(Connectable<C> tw,
     }
 
     private static <C extends Connectable<C>> void updateRatioTapChanger(Connectable<C> tw, RatioTapChanger rtc, String end, Context context, boolean isRegulatingAllowed) {
-        String ratioTapChangerId = findTapChangerId(tw, Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.RATIO_TAP_CHANGER + end);
+        String propertyTag = "".equals(end) ? CgmesNames.RATIO_TAP_CHANGER : getRatioTapChangerAliasType(end);
+        String ratioTapChangerId = findTapChangerId(tw, propertyTag);
```
marqueslanauja: The `CGMES` prefix is not added to the propertyTag when the `end` is empty. 
Maybe we could define a `PROPERTY_RATIO_TAP_CHANGER`, or extend the `getRatioTapChangerAliasType` method to support an empty `end`, as you did in the `getTerminalSignPropertyName` method.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractTransformerConversion.java
```diff
@@ -254,7 +253,8 @@ private static void setRegulation(RatioTapChanger rtc, double targetV, double ta
     }
 
     private static <C extends Connectable<C>> void updatePhaseTapChanger(Connectable<C> tw, PhaseTapChanger ptc, String end, Context context, boolean isRegulatingAllowed) {
-        String phaseTapChangerId = findTapChangerId(tw, Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.PHASE_TAP_CHANGER + end);
+        String propertyTag = "".equals(end) ? CgmesNames.PHASE_TAP_CHANGER : getPhaseTapChangerAliasType(end);
+        String phaseTapChangerId = findTapChangerId(tw, propertyTag);
```
marqueslanauja: The `CGMES` prefix is not added to the propertyTag when the `end` is empty.
Maybe we could define a `PROPERTY_PHASE_TAP_CHANGER`, or extend the `getPhaseTapChangerAliasType` method to support an empty `end`, as you did in the `getTerminalSignPropertyName` method.

marqueslanauja (commented):
I haven’t completed the full review yet, but here are my comments on the CGMES conversion classes. The next step will be to review the export classes.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
```diff
@@ -231,142 +222,103 @@ private void computeBoundaryIds(ExportParameters exportParameters, Network netwo
...
-            region = Pair.of(regionId, regionName);
-        }
-        return region;
-    }
+    private String getOrCreateSubRegion(Substation substation, Region region) {
+        Set<String> geoTags = substation.getGeographicalTags();
+        String subRegionName = geoTags.size() == 1 ? geoTags.iterator().next() : region.name;
+        String subRegionId = substation.hasProperty(PROPERTY_SUB_REGION_ID) ?
+                namingStrategy.getCgmesIdFromProperty(substation, PROPERTY_SUB_REGION_ID) :
+                namingStrategy.getCgmesId(ref(subRegionName), SUB_GEOGRAPHICAL_REGION);
```
marqueslanauja, line 286: Since the existence of the property has already been checked, we could use `substation.getProperty(PROPERTY_SUB_REGION_ID)` instead of `namingStrategy.getCgmesIdFromProperty(substation, PROPERTY_SUB_REGION_ID)`.

This would avoid checking the property twice andand improve code clarity.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportUtil.java
```diff
@@ -358,110 +352,102 @@ public static <C extends Connectable<C>> void setCgmesTapChangerType(C eq, Strin
...
-        } else {
-            int sequenceNumber = getTerminalSequenceNumber(t);
-            aliasType = Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.TERMINAL + sequenceNumber;
+    public static String getDanglingLineBoundaryTerminalId(DanglingLine danglingLine, CgmesExportContext context) {
+        if (danglingLine.getAliasFromType(ALIAS_TERMINAL_BOUNDARY).isEmpty()
+            && danglingLine.getTieLine().flatMap(tl -> tl.getAliasFromType(ALIAS_TERMINAL_BOUNDARY)).isPresent()) {
+            return context.getNamingStrategy().getCgmesIdFromAlias(danglingLine.getTieLine().orElseThrow(), ALIAS_TERMINAL_BOUNDARY);
         }
-        return context.getNamingStrategy().getCgmesIdFromAlias(c, aliasType);
+        return context.getNamingStrategy().getCgmesIdFromAlias(danglingLine, ALIAS_TERMINAL_BOUNDARY);
```
marqueslanauja: I don’t think any properties are recorded directly on the `TieLine`. This code is probably obsolete and may have been valid before the `TieLine` was managed as two `DanglingLines`.
We can probably call  `context.getNamingStrategy().getCgmesIdFromAlias(danglingLine, ALIAS_TERMINAL_BOUNDARY` directly.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
```diff
@@ -232,30 +227,34 @@ private static void writeSubGeographicalRegion(String subGeographicalRegionId, S
         }
     }
 
-    private static void writeVoltageLevels(Network network, String cimNamespace, XMLStreamWriter writer, CgmesExportContext context, Set<Double> exportedBaseVoltagesByNominalV) throws XMLStreamException {
+    private static void writeBaseVoltages(String cimNamespace, XMLStreamWriter writer, CgmesExportContext context) throws XMLStreamException {
+        for (CgmesExportContext.BaseVoltageSource baseVoltageSource : context.getBaseVoltageSources()) {
```
marqueslanauja: I ran some tests to verify ID persistence and noticed that the `baseVoltages` are written in a different order each time the network is exported. We could eliminate these differences and improve the code by sorting the `baseVoltages `beforehand.

```
List<CgmesExportContext.BaseVoltageSource> baseVoltageSources = context.getBaseVoltageSources().stream().sorted(Comparator.comparing(CgmesExportContext.BaseVoltageSource::id)).toList();
for (CgmesExportContext.BaseVoltageSource baseVoltageSource : baseVoltageSources) {
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
```diff
@@ -844,32 +822,46 @@ private static void writeThreeWindingsTransformerEnd(ThreeWindingsTransformer tw
...
+            }
+            String tapChangerId = context.getNamingStrategy().getCgmesIdFromAlias(eq, aliasType);
             int neutralStep = getClosestNeutralStep(ptc);
             int normalStep = getNormalStep(eq, tapChangerId).orElse(neutralStep);
-            Optional<String> regulatingControlId = getTapChangerControlId(eq, tapChangerId);
-            String cgmesRegulatingControlId = null;
-            if (regulatingControlId.isPresent() && CgmesExportUtil.regulatingControlIsDefined(ptc)) {
-                cgmesRegulatingControlId = context.getNamingStrategy().getCgmesId(regulatingControlId.get());
-                if (!regulatingControlsWritten.contains(cgmesRegulatingControlId)) {
+            String regulatingControlId = null;
```
marqueslanauja: To get the `normalStep`, we need to use the tapChangerId recorded as an alias, rather than the generated through the `namingStrategy`.

One possible approach is:

```
int normalStep = eq.getAliasFromType(aliasType).flatMap(cgmesTapChangerId -> {
                OptionalInt optionalNormalStep = getNormalStep(eq, cgmesTapChangerId);
                return optionalNormalStep.isPresent() ? Optional.of(optionalNormalStep.getAsInt()) : Optional.empty();
            }).orElse(neutralStep);
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
```diff
@@ -886,16 +878,16 @@ private static <C extends Connectable<C>> void writePhaseTapChanger(C eq, PhaseT
...
+                    regulatingControlsWritten.add(regulatingControlId);
                 }
             }
             String phaseTapChangerTableId = context.getNamingStrategy().getCgmesId(refTyped(eq), ref(endNumber), PHASE_TAP_CHANGER_TABLE);
             // If we write the EQ, we will always write the Tap Changer as tabular
             // We reset the phase tap changer type stored in the extensions
             String typeTabular = CgmesNames.PHASE_TAP_CHANGER_TABULAR;
             CgmesExportUtil.setCgmesTapChangerType(eq, tapChangerId, typeTabular);
-            TapChangerEq.writePhase(typeTabular, cgmesTapChangerId, twtName + "_PTC", endId, ptc.getLowTapPosition(), ptc.getHighTapPosition(), neutralStep, normalStep, neutralU, ptc.hasLoadTapChangingCapabilities(), phaseTapChangerTableId, cgmesRegulatingControlId, cimNamespace, writer, context);
+            TapChangerEq.writePhase(typeTabular, tapChangerId, twtName + "_PTC", endId, ptc.getLowTapPosition(), ptc.getHighTapPosition(), neutralStep, normalStep, neutralU, ptc.hasLoadTapChangingCapabilities(), phaseTapChangerTableId, regulatingControlId, cimNamespace, writer, context);
```
marqueslanauja: I don’t think we need to change the initial type recorded in the `cgmeTapChangerExtension` during the import process.

We can remove the `CgmesExportUtil.setCgmesTapChangerType` method and we have to adjust the unit test `CgmesExportTest.testPhaseTapChangerType` by replacing 

			`assertEquals(CgmesNames.PHASE_TAP_CHANGER_TABULAR, typeEqAndSsh);`

with
			`assertEquals("PhaseTapChangerAsymmetrical", typeEqAndSsh);`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
```diff
@@ -904,22 +896,27 @@ private static <C extends Connectable<C>> void writePhaseTapChanger(C eq, PhaseT
...
+                // - If the original CGMES contained both a tap changer at end 1 and at end 2, they have been combined,
+                // and we only export back one tap changer with the id of the original one from end 1.
+                // - If the original CGMES contained only a tap changer at end 2, it has been moved to end 1 during import
+                // and is exported back at end 1, but with its original id.
+                aliasType = ALIAS_RATIO_TAP_CHANGER2;
+            }
+            String tapChangerId = context.getNamingStrategy().getCgmesIdFromAlias(eq, aliasType);
 
             int neutralStep = getClosestNeutralStep(rtc);
             int normalStep = getNormalStep(eq, tapChangerId).orElse(neutralStep);
```
marqueslanauja: To get the `normalStep`, we need to use the tapChangerId recorded as an alias, rather than the generated through the `namingStrategy`.

One possible approach is:

```
int normalStep = eq.getAliasFromType(aliasType).flatMap(cgmesTapChangerId -> {
                OptionalInt optionalNormalStep = getNormalStep(eq, cgmesTapChangerId);
                return optionalNormalStep.isPresent() ? Optional.of(optionalNormalStep.getAsInt()) : Optional.empty();
            }).orElse(neutralStep);
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
```diff
@@ -551,37 +541,36 @@ private static void writeTapSteps(Network network, String cimNamespace, XMLStrea
...
         if (twt.hasPhaseTapChanger()) {
-            PhaseTapChanger phaseTapChanger = twt.getPhaseTapChanger();
-            int endNumber = twt.getAliasFromType(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.PHASE_TAP_CHANGER + 1).isPresent() ? 1 : 2;
-            String ptcId = context.getNamingStrategy().getCgmesIdFromAlias(twt, Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.PHASE_TAP_CHANGER + endNumber);
-            writeSvTapStep(ptcId, phaseTapChanger.findSolvedTapPosition().orElse(phaseTapChanger.getTapPosition()), cimNamespace, writer, context);
+            String aliasType = twt.getAliasFromType(ALIAS_PHASE_TAP_CHANGER2).isPresent() && twt.getAliasFromType(ALIAS_PHASE_TAP_CHANGER1).isEmpty() ?
+                ALIAS_PHASE_TAP_CHANGER2 : ALIAS_PHASE_TAP_CHANGER1;
+            String ptcId = context.getNamingStrategy().getCgmesIdFromAlias(twt, aliasType);
+            writeSvTapStep(ptcId, twt.getPhaseTapChanger().findSolvedTapPosition().orElse(twt.getPhaseTapChanger().getTapPosition()), cimNamespace, writer, context);
             writeSvTapStepHidden(twt, ptcId, cimNamespace, writer, context);
```
marqueslanauja: To get the hidden step we need to use the phaseTapChangerId recorded as an alias, rather than the generated through the namingStrategy.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
```diff
@@ -551,37 +541,36 @@ private static void writeTapSteps(Network network, String cimNamespace, XMLStrea
...
         if (twt.hasRatioTapChanger()) {
-            RatioTapChanger ratioTapChanger = twt.getRatioTapChanger();
-            int endNumber = twt.getAliasFromType(Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.RATIO_TAP_CHANGER + 1).isPresent() ? 1 : 2;
-            String rtcId = context.getNamingStrategy().getCgmesIdFromAlias(twt, Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.RATIO_TAP_CHANGER + endNumber);
-            writeSvTapStep(rtcId, ratioTapChanger.findSolvedTapPosition().orElse(ratioTapChanger.getTapPosition()), cimNamespace, writer, context);
+            String aliasType = twt.getAliasFromType(ALIAS_RATIO_TAP_CHANGER2).isPresent() && twt.getAliasFromType(ALIAS_RATIO_TAP_CHANGER1).isEmpty() ?
+                ALIAS_RATIO_TAP_CHANGER2 : ALIAS_RATIO_TAP_CHANGER1;
+            String rtcId = context.getNamingStrategy().getCgmesIdFromAlias(twt, aliasType);
+            writeSvTapStep(rtcId, twt.getRatioTapChanger().findSolvedTapPosition().orElse(twt.getRatioTapChanger().getTapPosition()), cimNamespace, writer, context);
             writeSvTapStepHidden(twt, rtcId, cimNamespace, writer, context);
```
marqueslanauja: To get the hidden step we need to use the ratioTapChangerId recorded as an alias, rather than the generated through the namingStrategy.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
```diff
@@ -551,37 +541,36 @@ private static void writeTapSteps(Network network, String cimNamespace, XMLStrea
...
     private static void writeTapStepsThreeWindingsTransformer(ThreeWindingsTransformer twt, String cimNamespace, XMLStreamWriter writer, CgmesExportContext context) throws XMLStreamException {
-        int endNumber = 1;
         for (ThreeWindingsTransformer.Leg leg : Arrays.asList(twt.getLeg1(), twt.getLeg2(), twt.getLeg3())) {
+            String endNumber = Integer.toString(leg.getSide().getNum());
             if (leg.hasPhaseTapChanger()) {
-                String ptcId = context.getNamingStrategy().getCgmesIdFromAlias(twt, Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.PHASE_TAP_CHANGER + endNumber);
+                String ptcId = context.getNamingStrategy().getCgmesIdFromAlias(twt, getPhaseTapChangerAliasType(endNumber));
                 PhaseTapChanger phaseTapChanger = leg.getPhaseTapChanger();
                 writeSvTapStep(ptcId, phaseTapChanger.findSolvedTapPosition().orElse(phaseTapChanger.getTapPosition()), cimNamespace, writer, context);
                 writeSvTapStepHidden(twt, ptcId, cimNamespace, writer, context);
```
marqueslanauja: To get the hidden step we need to use the phaseTapChangerId recorded as an alias, rather than the generated through the namingStrategy.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
```diff
@@ -551,37 +541,36 @@ private static void writeTapSteps(Network network, String cimNamespace, XMLStrea
...
             }
             if (leg.hasRatioTapChanger()) {
-                String rtcId = context.getNamingStrategy().getCgmesIdFromAlias(twt, Conversion.CGMES_PREFIX_ALIAS_PROPERTIES + CgmesNames.RATIO_TAP_CHANGER + endNumber);
+                String rtcId = context.getNamingStrategy().getCgmesIdFromAlias(twt, getRatioTapChangerAliasType(endNumber));
                 RatioTapChanger ratioTapChanger = leg.getRatioTapChanger();
                 writeSvTapStep(rtcId, ratioTapChanger.findTapPosition().orElse(ratioTapChanger.getTapPosition()), cimNamespace, writer, context);
                 writeSvTapStepHidden(twt, rtcId, cimNamespace, writer, context);
```
marqueslanauja: To get the hidden step we need to use the ratioTapChangerId recorded as an alias, rather than the generated through the namingStrategy.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
```diff
@@ -188,63 +170,56 @@ private static void writeEquivalentInjections(Network network, String cimNamespa
...
+                    writeTapChanger(twt, rtcId, tcCtrlId, leg.getRatioTapChanger(), CgmesNames.RATIO_TAP_CHANGER, regulatingControlViews, cimNamespace, writer, context);
                 }
-                i++;
             }
         }
     }
 
-    private static <C extends Connectable<C>> void writeTapChanger(C eq, String tcId, TapChanger<?, ?, ?, ?> tc, String defaultType, Map<String, List<RegulatingControlView>> regulatingControlViews, String cimNamespace, XMLStreamWriter writer, CgmesExportContext context) throws XMLStreamException {
+    private static <C extends Connectable<C>> void writeTapChanger(C eq, String tcId, String tcCtrlId, TapChanger<?, ?, ?, ?> tc, String defaultType, Map<String, List<RegulatingControlView>> regulatingControlViews, String cimNamespace, XMLStreamWriter writer, CgmesExportContext context) throws XMLStreamException {
         String type = CgmesExportUtil.cgmesTapChangerType(eq, tcId).orElse(defaultType);
```
marqueslanauja: To get the `cgmesTypeChangertype` we need to use the tapChangerId recorded as an alias, rather than the generated through the namingStrategy. We cannot use `tcId`.

The same applies when selecting selecting the `cgmesTapChanger` to be used in the `writeHiddenTapChanger` method.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/EquipmentExportTest.java
```diff
@@ -616,7 +616,7 @@ void testExportEquivalentInjectionBaseVoltage() throws IOException {
                 .setVoltageRegulatorOn(true).setTargetV(400)
                 .setTargetP(0).setMinP(0).setMaxP(10)
                 .add();
-        g.setProperty(Conversion.PROPERTY_CGMES_ORIGINAL_CLASS, "EquivalentInjection");
+        g.setProperty(PROPERTY_CGMES_ORIGINAL_CLASS, "EquivalentInjection");
```
marqueslanauja: Better

`g.setProperty(PROPERTY_CGMES_ORIGINAL_CLASS, CgmesNames.EQUIVALENT_INJECTION);`

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/naming/NamingStrategy.java
rcourtier: Is it ok for you if I write the javadoc in https://github.com/powsybl/powsybl-core/pull/3779?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/Conversion.java
rcourtier: done!

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractTransformerConversion.java
rcourtier: You are right.
I have added an alias `ALIAS_RATIO_TAP_CHANGER`.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractTransformerConversion.java
rcourtier: You are right.
I have added an alias `ALIAS_PHASE_TAP_CHANGER`.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
rcourtier: I don't think that is correct.

Depending on the chosen `namingStrategy`, the result from `getCgmesIdFromProperty` will be different. Yes, in the case of the `identity` naming strategy, the property value will indeed be exported as it is, and yes a `getProperty` would be simpler. However, in the case of the `cgmes-fix-all-invalid-ids` naming strategy, the property value will be processed by `getCgmesId` so that if the property value is not valid (is not compliant with IEC 61970-552), it will be made valid.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/transformers/AbstractTransformerConversion.java
marqueslanauja: ok

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
marqueslanauja: You are right

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
marqueslanauja: Fixed during my review. It is correct.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
marqueslanauja: Fixed during my review. Your proposal is correct and simplifies the code. My only observation is that assigning a not null default value (`tapChangerId`) when the alias is missing could make the code less clear, as that value will never be useful.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
marqueslanauja: Fixed during my review. Your proposal is correct and simplifies the code. My only observation is that assigning a not null default value (`tapChangerId`) when the alias is missing could make the code less clear, as that value will never be useful.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
rcourtier: done!

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
rcourtier: That is right. It bothered me too. I will change it.

rcourtier: done!

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/EquipmentExport.java
rcourtier: done!

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/EquipmentExportTest.java
rcourtier: done!

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/SteadyStateHypothesisExport.java
rcourtier: done!

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
rcourtier: done!

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
rcourtier: done!

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
rcourtier: done!

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/StateVariablesExport.java
rcourtier: done!

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportUtil.java
rcourtier: That's right, properties and alias are not recorded on the `TieLine` anymore. But we want to keep that for legacy reason in order to make sure that old xiidm files with the alias on the TieLine are correctly exported to CGMES.

rcourtier: I've added a comment in `getDanglingLineBoundaryTerminalId`.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/naming/NamingStrategy.java
olperr1: Yes, it's OK!

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
```diff
@@ -241,36 +240,42 @@ public ReferenceDataProvider getReferenceDataProvider() {
...
         String defaultRegionName = substation.hasProperty(PROPERTY_REGION_NAME) ?
                 substation.getProperty(PROPERTY_REGION_NAME) :
                 substation.getCountry().map(Country::name).orElse(DEFAULT_REGION);
+
         if (substation.hasProperty(PROPERTY_REGION_ID)) {
-            // Add this geographical region id from property if it is not already mapped.
             regionId = namingStrategy.getCgmesIdFromProperty(substation, PROPERTY_REGION_ID);
-            regionName = regionsNameById.computeIfAbsent(regionId, k -> defaultRegionName);
+            region = new Region(regionId, defaultRegionName, Source.IGM);
+        } else if (regionRef != null) {
```
marqueslanauja, line 267: It may happen that the `regionId` recorded as a property of the substation is the `regionRef.id()`, in which case a new region should not be created.

rcourtier: In that case, a new region is not created. See comment below (line 275).

marqueslanauja: Sorry, I haven't read the comment, but I think we have to consider the case where the `namingStrategy` changes the initial cgmes region id.

rcourtier: The initial cgmes region id doesn't matter. The only thing to consider is the generated region id. If that id matches the one from the reference data, we don't export it. The purpose is to not write in `EQ` a region that's already in `EQBD`. As long as the generated id is different, we can export it in `EQ`.

marqueslanauja: ok, understood.

PR #3783, opened 2026-02-18
Make cgmes tap changer control id unique

rcourtier:

No.

Bug fix.

CGMES TapChangerControl ID is not unique. It is named after the CGMES PowerTransformer.

CGMES TapChangerControl ID is globally unique. It is named after the CGMES PowerTransformerEnd.

- [ ] Yes
- [X] No

**Other information**:


cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/TapChangerConversionTest.java
```diff
@@ -129,6 +130,29 @@ void phaseTapChangerLinearTest() {
         assertTapStep(steps2.get(2), 2.0, 10.0);
     }
 
+    @Test
+    void tapChangerControlIdTest() throws IOException {
```
zamarrenolm, line 134: The test checks that a power transformer can have more than one ratio tap changers. Maybe we should check that with this change we are also adding support for transformers with many phase tap changers

rcourtier: Done

zamarrenolm (approved):

olperr1:
Regarding the sonar issue, the problem was previously here.
We could have introduced a new object to contain `String cimNamespace, XMLStreamWriter writer, CgmesExportContext context`, this would make sense since they are always coming together. But it won't solve the issue (and could be done in a dedicated PR).

PR #3784, opened 2026-02-18
Fix terminalRef (de)serialization in old BIIDM versions

olperr1:

No

Bug fix

With `powsybl-core 7.1.1`, the import of a previous BIIDM version (< 1.15) and the BIIDM export in a previous version are broken.
The problem is that in the resulting file, the "number" attribute was exported for Terminal even in versions <1.15. Since null-valued attributes are not exported in XIIDM and JIIDM, these formats are not impacted.

The "number" attribute is not exported for versions < 1.15, in BIIDM, JIIDM and XIIDM.
The only exception of this rule is when the `iidmVersionIncompatibilityBehavior` export option set to `LOG_ERROR`. In this case, the `terminal` field is also exported when it is not null.
Note that in this specific case, this leads to an inconsistent file in BIIDM. Since this format is still a BETA feature and the problematic case is very restricted, we consider that this is not a real problem.

- [ ] Yes
- [X] No

**Other information**:


iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/TerminalRefSerDe.java
```diff
@@ -62,7 +76,15 @@ public static void writeTerminalRefAttribute(Terminal terminal, NetworkSerialize
 
         writer.writeStringAttribute(ID, connectableId);
         writer.writeEnumAttribute(SIDE, tSide);
-        writer.writeEnumAttribute(NUMBER, tNumber);
+        if (context.getVersion().compareTo(IidmVersion.V_1_15) >= 0 || tNumber != null) {
```
flo-dup: Writing a number field in a terminalRef for a iidm version <1.15 would produce an invalid file, whatever the format is, so we'd better remove that condition.
```suggestion
        if (context.getVersion().compareTo(IidmVersion.V_1_15) >= 0) {
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/TerminalRefSerDe.java
```diff
@@ -36,16 +40,26 @@ public static void writeTerminalRef(Terminal t, NetworkSerializerContext context
...
+    public static void writeTerminalRefAttribute(Terminal t, NetworkSerializerContext context, String parentElementName) {
+        writeTerminalRefAttribute(t, context, parentElementName, context.getWriter());
+    }
+
+    @Deprecated(since = "7.2.0")
     public static void writeTerminalRefAttribute(Terminal terminal, NetworkSerializerContext context, TreeDataWriter writer) {
+        writeTerminalRefAttribute(terminal, context, null, writer);
+    }
+
+    public static void writeTerminalRefAttribute(Terminal terminal, NetworkSerializerContext context, String parentElementName, TreeDataWriter writer) {
```
flo-dup: You can now remove parentElementName here and propagate the removal

olperr1: You are right!

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/VoltageLevelSerDe.java
```diff
@@ -359,8 +359,16 @@ protected void readSubElements(VoltageLevel vl, NetworkDeserializerContext conte
                 case VscConverterStationSerDe.ROOT_ELEMENT_NAME -> VscConverterStationSerDe.INSTANCE.read(vl, context);
                 case LccConverterStationSerDe.ROOT_ELEMENT_NAME -> LccConverterStationSerDe.INSTANCE.read(vl, context);
                 case GroundSerDe.ROOT_ELEMENT_NAME -> GroundSerDe.INSTANCE.read(vl, context);
-                case LineCommutatedConverterSerDe.ROOT_ELEMENT_NAME -> LineCommutatedConverterSerDe.INSTANCE.read(vl, context);
-                case VoltageSourceConverterSerDe.ROOT_ELEMENT_NAME -> VoltageSourceConverterSerDe.INSTANCE.read(vl, context);
+                case LineCommutatedConverterSerDe.ROOT_ELEMENT_NAME -> {
+                    IidmSerDeUtil.assertMinimumVersion(ROOT_ELEMENT_NAME, LineCommutatedConverterSerDe.ROOT_ELEMENT_NAME,
```
flo-dup, line 363: Not directly related to that PR, but ok! If we go that direction we should probably do the same with Ground just above, and others...

PR #3786, opened 2026-02-18
Fix unstable connectable ordering

### LINKED ISSUE #3674: Unstable connectable ordering

When getting connectables the ordering may be changed on a file roundtrip for bus-breaker voltage levels

When getting connectables the ordering should remain the same on any file roundtrip, or the documentation should clearly state it might be altered by a Network.copy (or writing/reading).

### Describe the steps

- Checkout 268ae366fe62f5d857fdeee6ef3d7facf710379d
- Run the unit test ConnectableOrderingXmlTest::test

### Environment

_No response_

### Relevant Log Output

_No response_

### Extra Information

_No response_

quinarygio:

Fixes #3674 

- [x] Yes
- [ ] No

- Both `sort(...)` methods of `IidmSerDeUtil` take an additional `Network network` parameter.
- `NetworkSerDe.copy(Network, NetworkFactory, ExecutorService, TreeDataFormat`) takes an additional `boolean preserveConnectableCreationOrder` parameter.
  - When set to true, the behavior is the same as before.
  - When set to false, the connectables' creation order is preserved, but the copy takes slightly more time.
  - The `NetworkSerDe.copy(...)` methods that don't have this parameter keep the same behavior as before (the connectables' creation order is not preserved, but it is more performant).

**Other information**:

- At export, you can set `iidm.export.xml.use-connectable-creation-order` to true to order the connectable in the resulting IIDM file (XML, JSON or Binary).
- By default the connectable order is not preserved at copy in order to avoid degrading the performances. You can call `NetworkSerDe.copy(...)` with `preserveConnectableCreationOrder` to true if you want to preserve the order.


iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/util/IidmSerDeUtil.java
```diff
@@ -351,7 +351,9 @@ public static <T extends Identifiable> Iterable<T> sorted(Iterable<T> identifiab
         return exportOptions.isSorted() ? StreamSupport.stream(identifiables.spliterator(), false)
                 .sorted(Comparator.comparing(Identifiable::getId))
                 .collect(Collectors.toList())
-                : identifiables;
+                : StreamSupport.stream(identifiables.spliterator(), false)
+                .sorted(Comparator.comparing(Identifiable::getSortIndex))
+                .collect(Collectors.toList());
```
olperr1: Only the comparator is different between the 2 alternatives. It would be cleaner to choose the comparator first and then to execute the processing with it.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Identifiable.java
```diff
@@ -36,6 +36,13 @@ default Network getParentNetwork() {
      */
     String getId();
 
+    /**
+     * Get the sort index of the object.
+     */
+    long getSortIndex();
+
+    void setSortIndex(long index);
```
olperr1: This method must be removed from the interface. Users should not be allowed to change it.

iidm/iidm-serde/src/test/resources/twtOrdering.xiidm
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<iidm:network xmlns:iidm="http://www.powsybl.org/schema/iidm/1_15" id="sim1" caseDate="2013-01-15T18:45:00.000+01:00" forecastDistance="0" sourceFormat="test" minimumValidationLevel="STEADY_STATE_HYPOTHESIS">
```
alicecaron: set version `1_16` as IIDM current version in main is now 1.16

iidm/iidm-serde/src/test/resources/twtOrderingOk.xiidm
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
alicecaron, line 1: this file is no longer used and can be deleted

iidm/iidm-serde/src/test/resources/twtOrdering.xiidm
```diff
@@ -0,0 +1,22 @@
...
+            <iidm:generator id="GEN" energySource="OTHER" minP="-9999.99" maxP="9999.99" voltageRegulatorOn="true" targetP="607.0" targetV="24.5" targetQ="301.0" bus="NGEN1" connectableBus="NGEN1">
+                <iidm:minMaxReactiveLimits minQ="-9999.99" maxQ="9999.99"/>
+            </iidm:generator>
+        </iidm:voltageLevel>
+        <iidm:voltageLevel id="VLHV1" nominalV="380.0" topologyKind="BUS_BREAKER">
+            <iidm:busBreakerTopology>
+                <iidm:bus id="NHV1"/>
+            </iidm:busBreakerTopology>
+        </iidm:voltageLevel>
+        <iidm:twoWindingsTransformer id="NGEN_NHV1" r="0.2" x="11.1" g="0.0" b="0.0" ratedU1="24.0" ratedU2="400.0" bus1="NGEN1" connectableBus1="NGEN1" voltageLevelId1="VLGEN" bus2="NHV1" connectableBus2="NHV1" voltageLevelId2="VLHV1"/>
```
alicecaron: to make the round trip work just re order the attribute? It would make the test pass, the attribute order is not important for a functionnal use I think

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Identifiable.java
```diff
@@ -36,6 +36,13 @@ default Network getParentNetwork() {
      */
     String getId();
 
+    /**
+     * Get the sort index of the object.
+     */
+    long getSortIndex();
```
olperr1: A test should be added in the TCK to ensure that the custom IIDM implementations handle it correctly.

olperr1 (changes requested):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkIndex.java
```diff
@@ -43,6 +44,7 @@ void checkAndAdd(Identifiable<?> obj) {
                     + ") '" + obj.getId() + "' already exists");
         }
         objectsById.put(obj.getId(), obj);
+        obj.setSortIndex(sortIndex++);
```
alicecaron:
```suggestion
        ((AbstractIdentifiable<?>) obj).setSortIndex(sortIndex++);
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiable.java
```diff
@@ -219,6 +220,16 @@ public Set<String> getPropertyNames() {
         return properties.getPropertyNames();
     }
 
+    @Override
+    public long getSortIndex() {
+        return sortIndex;
+    }
+
+    @Override
```
alicecaron: Consequence of Olivier comment, as the method is not in the interface: remove `@Override`, the implementation is directly here. Then to be able to use it:
- in `NetworkIndex` => cast there
- the `IdentifiableMock` test class should now extend directly the `AbstracIdenifiable` and no longer the `AbstractExtendable`

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/VariantManagerImplTest.java
```diff
@@ -142,6 +143,16 @@ public void allocateVariantArrayElement(int[] indexes, int sourceIndex) {
...
+
+        @Override
+        public long getSortIndex() {
+            return sortIndex;
+        }
+
+        @Override
+        public void setSortIndex(long index) {
+            this.sortIndex = index;
+        }
```
alicecaron, line 149: remove those methods

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/VariantManagerImplTest.java
alicecaron:
```suggestion
    private static final class IdentifiableMock extends AbstractIdentifiable<IdentifiableMock> implements Identifiable<IdentifiableMock>, MultiVariantObject {
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/util/IidmSerDeUtil.java
alicecaron: done

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Identifiable.java
alicecaron: done

alicecaron (dismissed):

alicecaron (dismissed):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiable.java
```diff
@@ -21,7 +21,7 @@
 /**
  * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
  */
-abstract class AbstractIdentifiable<I extends Identifiable<I>> extends AbstractExtendable<I> implements Identifiable<I>, Validable, MultiVariantObject {
+public abstract class AbstractIdentifiable<I extends Identifiable<I>> extends AbstractExtendable<I> implements Identifiable<I>, Validable, MultiVariantObject {
```
olperr1: There's no need to change the visibility of this class:

```suggestion
abstract class AbstractIdentifiable<I extends Identifiable<I>> extends AbstractExtendable<I> implements Identifiable<I>, Validable, MultiVariantObject {
```

iidm/iidm-modification/src/test/resources/eurostag-create-twt-feeder-bays.xml
```diff
@@ -14,8 +14,8 @@
                 <iidm:bus id="NHV1"/>
             </iidm:busBreakerTopology>
         </iidm:voltageLevel>
-        <iidm:twoWindingsTransformer id="NGEN_NHV1" r="0.26658461538461536" x="11.104492831516762" g="0.0" b="0.0" ratedU1="24.0" ratedU2="400.0" bus1="NGEN" connectableBus1="NGEN" voltageLevelId1="VLGEN" bus2="NHV1" connectableBus2="NHV1" voltageLevelId2="VLHV1"/>
-        <iidm:twoWindingsTransformer id="twtTest" r="1.0" x="1.0" g="0.0" b="0.0" ratedU1="220.0" ratedU2="400.0" bus1="NGEN" connectableBus1="NGEN" voltageLevelId1="VLGEN" bus2="NHV1" connectableBus2="NHV1" voltageLevelId2="VLHV1"/>
+        <iidm:twoWindingsTransformer id="NGEN_NHV1" r="0.26658461538461536" x="11.104492831516762" g="0.0" b="0.0" ratedU1="24.0" ratedU2="400.0" voltageLevelId1="VLGEN" bus1="NGEN" connectableBus1="NGEN" voltageLevelId2="VLHV1" bus2="NHV1" connectableBus2="NHV1"/>
+        <iidm:twoWindingsTransformer id="twtTest" r="1.0" x="1.0" g="0.0" b="0.0" ratedU1="220.0" ratedU2="400.0" voltageLevelId1="VLGEN" bus1="NGEN" connectableBus1="NGEN" voltageLevelId2="VLHV1" bus2="NHV1" connectableBus2="NHV1"/>
```
samirromdhani: Not sure that these changes are needed, here, the connectable order still the same, only the attribute order changed => no tests are affected by this changes, unless we switch to assertText instead of assertXml.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/util/IidmSerDeUtil.java
```diff
@@ -348,12 +350,19 @@ public static String getAttributeName(String oldName, String newName, IidmVersio
...
-                .sorted(comparator)
-                .collect(Collectors.toList());
+
+        List<T> listIdentifiables = StreamSupport.stream(identifiables.spliterator(), false).collect(Collectors.toList());
+        if (listIdentifiables.isEmpty()) {
+            return listIdentifiables;
+        }
+        Network n = listIdentifiables.getFirst().getNetwork();
+
+        if (exportOptions.isSorted() || n.getIdentifiableNaturalOrderComparator().isPresent()) {
```
alicecaron: so juste to be sure, if `exportOptions.isSorted() == true` and `n.getIdentifiableNaturalOrderComparator().isPresent() == true` we suppose that `exportOptions.isSorted()`  takes precedence over the presence of an implementation of `getIdentifiableNaturalOrderComparator()` we need to:  
- precise this in the javadoc/comment on the `sorted` method
- add a unit test to cover this case

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/util/IidmSerDeUtil.java
```diff
@@ -348,10 +350,19 @@ public static String getAttributeName(String oldName, String newName, IidmVersio
...
+            listIdentifiables.sort(exportOptions.isSorted()
+                    ? Comparator.comparing(Identifiable::getId)
+                    : n.getIdentifiableNaturalOrderComparator().get());
+        }
+        return listIdentifiables;
```
alicecaron: For more readable clarity you can simplify the code like that:
```suggestion
        if (exportOptions.isSorted()) {
            listIdentifiables.sort(Comparator.comparing(Identifiable::getId));
        } else if (n.getIdentifiableNaturalOrderComparator().isPresent()) {
            listIdentifiables.sort(n.getIdentifiableNaturalOrderComparator().get());
        }
        return listIdentifiables;
```

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractIdentifiableTest.java
```diff
@@ -0,0 +1,31 @@
...
+import com.powsybl.iidm.network.Network;
+import com.powsybl.iidm.network.test.EurostagTutorialExample1Factory;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * @author Alice Caron {@literal <alice.caron at rte-france.com>}
+ */
+public abstract class AbstractIdentifiableTest {
```
olperr1, line 22: This class must be removed.

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/tck/IdentifiableTest.java
```diff
@@ -0,0 +1,15 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+package com.powsybl.iidm.network.impl.tck;
+
+import com.powsybl.iidm.network.tck.AbstractIdentifiableTest;
+
+/**
+ * @author Alice Caron {@literal <alice.caron at rte-france.com>}
+ */
+class IdentifiableTest extends AbstractIdentifiableTest { }
```
olperr1: This class must be removed.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Identifiable.java
olperr1: This method must be removed from this interface.

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/VariantManagerImplTest.java
```diff
@@ -32,7 +31,7 @@
  */
 class VariantManagerImplTest {
 
-    private static final class IdentifiableMock extends AbstractExtendable<IdentifiableMock> implements Identifiable<IdentifiableMock>, MultiVariantObject {
+    private static final class IdentifiableMock extends AbstractIdentifiable<IdentifiableMock> implements Identifiable<IdentifiableMock>, MultiVariantObject {
```
olperr1, line 34: I think that the changes of this class could (should?) be reverted.

olperr1 (changes requested):

olperr1:
@lejeana
Another point on this PR: all commits must have a "signed-off-by" clause in their messages to respect the DCO requirement.
You can find more detail on this subject [here](https://github.com/powsybl/.github/blob/main/CONTRIBUTING.md#dco).

Could you alter your two last commits to add this clause and push-force their modified version?
If you need some assistance for this, you can reach me on slack.

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/VariantManagerImplTest.java
lejeana: This change is due to change in function `checkAndAdd(Identifiable<?> obj)` from class NetworkIndex.
Without this change, some tests will fail to pass.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/util/IidmSerDeUtil.java
lejeana: done
- javadoc updated on the sorted method
- a new unit test cover the case when `exportOptions.isSorted() == true` takes precedence over `getIdentifiableNaturalOrderComparator`

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/util/IidmSerDeUtil.java
lejeana: done

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiable.java
lejeana: done

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractIdentifiableTest.java
lejeana: done

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/tck/IdentifiableTest.java
lejeana: done

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/VariantManagerImplTest.java
olperr1: OK, so we keep it like this.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Network.java
```diff
@@ -1773,6 +1773,13 @@ default Network setMinimumAcceptableValidationLevel(ValidationLevel validationLe
         return this;
     }
 
+    /**
+     * Return a comparator to sort element in the natural order
+     */
+    default Optional<Comparator> getIdentifiableNaturalOrderComparator() {
```
olperr1:
```suggestion
    default Optional<Comparator<Identifiable>> getIdentifiableNaturalOrderComparator() {
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractNetwork.java
```diff
@@ -82,6 +80,11 @@ protected String getTypeDescription() {
         return "Network";
     }
 
+    @Override
+    public Optional<Comparator> getIdentifiableNaturalOrderComparator() {
```
olperr1:
```suggestion
    public Optional<Comparator<Identifiable>> getIdentifiableNaturalOrderComparator() {
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/util/IidmSerDeUtil.java
```diff
@@ -367,15 +345,26 @@ public static String getAttributeName(String oldName, String newName, IidmVersio
...
      */
     public static <T extends Identifiable> Iterable<T> sorted(Iterable<T> identifiables, ExportOptions exportOptions) {
         Objects.requireNonNull(identifiables);
         Objects.requireNonNull(exportOptions);
-        return exportOptions.isSorted() ? StreamSupport.stream(identifiables.spliterator(), false)
-                .sorted(Comparator.comparing(Identifiable::getId))
-                .collect(Collectors.toList())
-                : identifiables;
+
+        List<T> listIdentifiables = StreamSupport.stream(identifiables.spliterator(), false).collect(Collectors.toList());
```
olperr1: To improve performance, we can process the case where the order is not changed before executing this instruction:
```java
if (!exportOptions.isSorted() && network.getIdentifiableNaturalOrderComparator().isEmpty()) {
    return identifiables;
}
// ...
```

To do this, the `network` should be given as a parameter and not retrieved from `identifiable` (since it is an Iterable, its first element cannot be directly used).

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/util/IidmSerDeUtil.java
olperr1: This method must also be changed to use the comparator.

olperr1 (changes requested):

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/util/IidmSerDeUtil.java
lejeana: I was afraid that the function signature modification would be undesirable.
On second thought, the impact on perfomance isn't small, I will apply this change.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Network.java
lejeana: done

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractNetwork.java
lejeana: done

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/util/IidmSerDeUtil.java
lejeana: done

alicecaron (dismissed):

lejeana:
With this PR, function `NetworkSerDe.copy()` set the option for `NaturalOrder` in `ExportOtions` to `true` ([see change here](https://github.com/powsybl/powsybl-core/pull/3786/changes#diff-433e3fd968e64fb9ef502067b6f229fba646919091da99a062a3a736d13688c0R1299))

This change involves sorting the given network to function *copy* and may affect performance.

I ran two tests locally, with and without call to `setNaturalOrder(true)`.

### First: impact on existing unit test
I did run the whole lifecycle to compile and run tests for the entire project.
In both cases, this took about 7 minutes, with no noticeable difference.

### Second: impact on the function itself 
I did run a test that calls the `copy()` method 1,000 times on network `disMeasRef.xiidm`, chosen for its size.
In both cases, this took around 5 seconds.

NathanDissoubray:
Running `NetworkSerializationBenchmark` from `powsybl-benchmark` on this PR, with the RealGrid network, all results in ms/op:
```
| Benchmark Operation  | XML (XIIDM) | JSON (JIIDM) | Binary (BIIDM) | CGMES   |
|----------------------|-------------|--------------|----------------|---------|
| Deserialization      | 372,17      | 201,88       | 133,59         | 4037,42 |
| Stream Serialization | 335,29      | 283,84       | 213,09         | —       |
| File Serialization   | 748,98      | 304,64       | 226,92         | 1970,24 |
| Copy                 | 21896,00    | 379,90       | 340,49         | —       |
```
Main for reference (same benchmark and network):
```
| Benchmark Operation  | XML (XIIDM) | JSON (JIIDM) | Binary (BIIDM) | CGMES   |
|----------------------|-------------|--------------|----------------|---------|
| Deserialization      | 386,05      | 208,62       | 122,43         | 4004,56 |
| Stream Serialization | 326,13      | 289,77       | 211,29         | —       |
| File Serialization   | 741,74      | 299,97       | 227,56         | 1996,08 |
| Copy                 | 21520,24    | 372,13       | 339,20         | —       |
```
No noticeable difference. I'll run it on a few other networks just in case this is an exception.

NathanDissoubray:
IEEE 300
this PR
```
| Benchmark Operation  | XML (XIIDM) | JSON (JIIDM) | Binary (BIIDM) | CGMES |
|----------------------|-------------|--------------|----------------|-------|
| Deserialization      | 3,16        | 2,21         | 1,50           | 91,26 |
| Stream Serialization | 3,44        | 3,33         | 2,83           | —     |
| File Serialization   | 6,06        | 3,45         | 3,03           | 33,08 |
| Copy                 | 132,41      | 4,68         | 5,01           | —     |
```

main
```
| Benchmark Operation  | XML (XIIDM) | JSON (JIIDM) | Binary (BIIDM) | CGMES |
|----------------------|-------------|--------------|----------------|-------|
| Deserialization      | 3,12        | 2,17         | 1,50           | 93,43 |
| Stream Serialization | 3,42        | 3,42         | 2,81           | —     |
| File Serialization   | 6,19        | 3,54         | 3,05           | 33,04 |
| Copy                 | 132,66      | 3,96         | 4,44           | —     |
```

NathanDissoubray:
RTE 1888
this PR
```
| Benchmark Operation  | XML (XIIDM) | JSON (JIIDM) | Binary (BIIDM) | CGMES  |
|----------------------|-------------|--------------|----------------|--------|
| Deserialization      | 36,86       | 19,36        | 12,01          | 444,15 |
| Stream Serialization | 33,21       | 31,01        | 25,16          | —      |
| File Serialization   | 68,22       | 31,95        | 25,79          | 184,27 |
| Copy                 | 1750,53     | 42,59        | 42,58          | —      |
```

main
```
| Benchmark Operation  | XML (XIIDM) | JSON (JIIDM) | Binary (BIIDM) | CGMES  |
|----------------------|-------------|--------------|----------------|--------|
| Deserialization      | 37,96       | 19,36        | 12,22          | 456,92 |
| Stream Serialization | 33,69       | 30,01        | 24,40          | —      |
| File Serialization   | 66,86       | 31,53        | 25,18          | 183,67 |
| Copy                 | 1727,04     | 37,03        | 38,72          | —      |
```

NathanDissoubray:
RTE 6515
this PR
```
| Benchmark Operation  | XML (XIIDM) | JSON (JIIDM) | Binary (BIIDM) | CGMES   |
|----------------------|-------------|--------------|----------------|---------|
| Deserialization      | 99,73       | 60,68        | 43,49          | 1681,53 |
| Stream Serialization | 121,13      | 108,18       | 92,69          | —       |
| File Serialization   | 198,56      | 111,72       | 93,87          | 701,21  |
| Copy                 | 4399,23     | 147,95       | 152,30         | —       |
```

main
```
| Benchmark Operation  | XML (XIIDM) | JSON (JIIDM) | Binary (BIIDM) | CGMES   |
|----------------------|-------------|--------------|----------------|---------|
| Deserialization      | 104,46      | 62,24        | 43,90          | 1693,97 |
| Stream Serialization | 115,57      | 110,53       | 92,97          | —       |
| File Serialization   | 201,69      | 110,61       | 94,18          | 697,14  |
| Copy                 | 4226,81     | 128,19       | 135,70         | —       |
```

NathanDissoubray:
We can notice on bigger networks that it is pretty consistently slightly slower on serialization and copy, but not by much.
Around 1 to 5% for IIDM formats. CGMES not much of an impact.
There seems to be about a 10-15% impact on the copy for the RTE 6515 (but absolute times remain low)

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/util/IidmSerDeUtil.java
```diff
@@ -424,24 +425,56 @@ public static String getAttributeName(String oldName, String newName, IidmVersio
...
     /**
-     * Sort identifiables by their ids.
+     * Sort identifiables by their ids if given export option is activated,
+     * otherwise, by their natural order if the network defines one.
+     * In all other cases, do not change the identifiables order.
      */
-    public static <T extends Identifiable<T>> Stream<T> sorted(Stream<T> stream, ExportOptions exportOptions) {
+    public static <T extends Identifiable<T>> Stream<T> sorted(Stream<T> stream, ExportOptions exportOptions, Network network) {
+        return sorted(network, stream, exportOptions);
+    }
```
olperr1: What is the interest of this method?
It is called only once, and we can use the next method instead...

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ExportOptions.java
```diff
@@ -78,6 +78,13 @@ public enum BusBranchVoltageLevelIncompatibilityBehavior {
      */
     private boolean sorted = false;
 
+    /**
+     * Sort IIDM objects so that generated XML does depend on data model object order.
+     * Require object exported to define a natural order.
+     * Remark: Export option 'sorted' takes priority over 'naturalOrder'.
+     */
+    private boolean naturalOrder = false;
```
olperr1: `useConnectableCreationOrder` would be more representative of the behavior.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/ConnectableOrderingXmlTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+import com.powsybl.iidm.network.Network;
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * @author Florian Dupuy {@literal <florian.dupuy at rte-france.com>}
```
olperr1, line 18: Could you also add your authorship?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiable.java
```diff
@@ -33,6 +33,7 @@ abstract class AbstractIdentifiable<I extends Identifiable<I>> extends AbstractE
 
     private final Set<String> aliasesWithoutType = new HashSet<>();
     private final Map<String, String> aliasesByType = new HashMap<>();
+    private long sortIndex;
```
olperr1: Maybe `creationOrder` instead?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Network.java
```diff
@@ -1789,6 +1789,13 @@ default Network setMinimumAcceptableValidationLevel(ValidationLevel validationLe
         return this;
     }
 
+    /**
+     * Return a comparator to sort element in the natural order
+     */
+    default Optional<Comparator<Identifiable<?>>> getIdentifiableNaturalOrderComparator() {
```
olperr1: `getIdentifiableCreationOrderComparator()`?

olperr1 (changes requested):

olperr1:
@NathanDissoubray:
Regarding your metrics:
- The CGMES export is not impacted by this PR
- The ordering is conditioned to an option which is only set during the network copy. It is consistent with your tests.

@flo-dup: do you think it is OK to have a network copy 10-15% longer in order to preserve the creation order?

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/util/IidmSerDeUtil.java
lejeana: This method has been removed, a mistake on my part.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/ConnectableOrderingXmlTest.java
lejeana: done

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ExportOptions.java
lejeana: i agree, change has been done, also impact related methods' name

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiable.java
lejeana: change has been done, also impact related methods' name

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Network.java
lejeana: change has been done

iidm/iidm-serde/src/test/resources/twtOrdering.xiidm
lejeana, line 1: Should we move this file to another location? Like in V1_18/?

iidm/iidm-serde/src/test/resources/twtOrderingSorted.xiidm
lejeana, line 1: Should we move this file to another location? Like in V1_18/?

lejeana (commented):
@olperr1 
I applied suggested changes and updated the doc about the new export option.

The newly added unit test use created ressources files and will work if the current powsybl version is set to v1.8.
Any upgrade in powsybl version will cause the test to fail.

Do we need to change where created references files are stored?

docs/grid_exchange_formats/iidm/export.md
```diff
@@ -132,6 +132,12 @@ Depending on object types the following sorting key has been chosen :
 
 By default, the network components are not sorted.
 
+**iidm.export.xml.use-connectable-creation-order**<br>
+The `iidm.export.xml.use-connectable-creation-order` property is an optional property that defines whether the XIIDM objects in file generated by the XIIDM exporter will be sorted using the creation order defined.
```
olperr1:
```suggestion
The `iidm.export.xml.use-connectable-creation-order` property is an optional property that defines whether the XIIDM objects in file generated by the XIIDM exporter will be sorted using the creation order.
```

iidm/iidm-serde/src/test/resources/twtOrdering.xiidm
olperr1: I don't think this is necessary. We already have tests that are only performed on the last version and thus test files that must be updated every time we increase the IIDM version.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -1296,7 +1296,7 @@ public static Network copy(Network network, NetworkFactory networkFactory, Execu
             Pipe pipe = Pipe.open();
             executor.execute(() -> {
                 try (Pipe.SinkChannel sinkChannel = pipe.sink()) {
-                    write(network, new ExportOptions().setFormat(format), Channels.newOutputStream(sinkChannel));
+                    write(network, new ExportOptions().setFormat(format).setConnectableCreationOrder(true), Channels.newOutputStream(sinkChannel));
```
olperr1, line 1299: Regarding the performances' decrease, I wonder if we shouldn't add an additional parameter to the `copy` methods. In some cases, preserving the order may not be as important as having better performances.
By default, we can preserve the order (set `connectableCreationOrder` to true), but we may skip this if the caller prefers better performances.

olperr1 (changes requested):

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
NathanDissoubray: can't the caller just set `connectableCreationOrder` to `false` to have the performance increase ? Because if we add a `increaseCopyPerf` boolean, what will it do apart from turning the ordering off ?

lejeana: @NathanDissoubray the `copy` methods have no parameter to configure the `ExportOptions` used ; hence no way to interact with `connectableCreationOrder` value.
I believe the suggestion is to add such a parameter in `copy` methods, not to add a new boolean, such as `increaseCopyPerf`, in `ExportOptions` that would indeed do nothing apart turning `connectableCreationOrder` off.

lejeana: How many overaloading methods do we want while adding a new parameter in `copy` methods?

As of now, many overloading methods exist and adding it to the most basic overloadded method `copy(Network network)` imply adding lot of new overloading methods. Not really a problem, just wanted to inquire about the changes before doing it to avoid rollback or more changes afterward.

olperr1: Good question!
I think I would:
- add the boolean to this signature (breaking change):
  - `copy(Network, NetworkFactory, ExecutorService, TreeDataFormat)`
- add a new `copy(Network, NetworkFactory, boolean)` signature

lejeana: i added the new signature and update the signature `copy(Network, NetworkFactory, ExecutorService, TreeDataFormat)`

this PR migration steps has been added in consequence

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -1284,19 +1296,29 @@ public static Network copy(Network network, TreeDataFormat format) {
...
     }
 
+    /**
+     * Deep copy of the network using the specified converter
+     *
+     * @param network                     the network to copy
+     * @param networkFactory              the network factory to use for the copy
+     * @param executor                    the executor service to use for the copy
+     * @param format                      the converter to use to export/import the network
+     * @param useConnectableCreationOrder if `true`, keep connectables in the same ordering in data structures, has a low impact on performance.
```
olperr1:
```suggestion
     * @param useConnectableCreationOrder if `true`, keep connectables in the same ordering in data structures. If you are aiming for performance, consider setting it to false.
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -1284,19 +1296,29 @@ public static Network copy(Network network, TreeDataFormat format) {
...
+     * @return the copy of the network
+     */
     @SuppressWarnings("checkstyle:IllegalCatchWarning") // Any kind of Exception shall be managed here
-    public static Network copy(Network network, NetworkFactory networkFactory, ExecutorService executor, TreeDataFormat format) {
+    public static Network copy(Network network, NetworkFactory networkFactory, ExecutorService executor, TreeDataFormat format, boolean useConnectableCreationOrder) {
```
olperr1: I think `preserveConnectableCreationOrder` provides a more precise view of its interest.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -1260,8 +1260,20 @@ public static Network copy(Network network, NetworkFactory networkFactory) {
         return copy(network, networkFactory, ForkJoinPool.commonPool());
     }
 
+    /**
+     * Deep copy of the network using XML converter.
+     *
+     * @param network        the network to copy
+     * @param networkFactory the network factory to use for the copy
+     * @param useConnectableCreationOrder if `true`, keep connectables in the same ordering in data structures, has a low impact on performance.
```
olperr1: Same comment as for the other method's javadoc.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -1260,8 +1260,20 @@ public static Network copy(Network network, NetworkFactory networkFactory) {
         return copy(network, networkFactory, ForkJoinPool.commonPool());
     }
 
+    /**
+     * Deep copy of the network using XML converter.
```
olperr1: Can you change the javadoc comments to replace `using XML converter` by `using serialization/deserialization`?

Several javadoc comments are affected.

olperr1 (changes requested):

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
olperr1: Thanks

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -1260,8 +1260,20 @@ public static Network copy(Network network, NetworkFactory networkFactory) {
...
+     * @param preserveConnectableCreationOrder if `true`, keep connectables in the same ordering in data structures. If you are aiming for performance, consider setting it to false.
+     * @return the copy of the network
+     */
+    public static Network copy(Network network, NetworkFactory networkFactory, boolean preserveConnectableCreationOrder) {
+        return copy(network, networkFactory, ForkJoinPool.commonPool(), TreeDataFormat.JSON, preserveConnectableCreationOrder);
+    }
+
     public static Network copy(Network network, NetworkFactory networkFactory, ExecutorService executor) {
-        return copy(network, networkFactory, executor, TreeDataFormat.JSON);
+        return copy(network, networkFactory, executor, TreeDataFormat.JSON, true);
```
olperr1: Seen with @flo-dup: it would make more sense to have the most performant option by default (hence `preserveConnectableCreationOrder` set to false).

Could you make this change in all the methods without the `preserveConnectableCreationOrder` argument?
Also, could you update the migration guide to reflect this change?

olperr1 (changes requested):

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
lejeana: Default value set to false

The migration guide is up to date

olperr1 (dismissed):

olperr1:
@lejeana: Could you update the PR and solve the conflict?

olperr1 (approved):

PR #3795, opened 2026-02-26
Add warning when connecting voltage levels with different nominalV

### LINKED ISSUE #2526: [Load flow] Add warning when connecting voltage levels with different nominal voltages

Feature

Currently, it is possible to accidentally connect, say a 400kV bus to a 220kV one which can lead to weird load flow results (in my case, I also obtain different results when using itools and pypowsybl, maybe because of different versions/default parameters).

Output a warning when computing a load flow for such invalid cases (I guess no warning should be given for connections between 380 and 400KV buses).

  - PowSyBl Version: ...
  - OS Version: Ubuntu 22.04

samirromdhani:

Fixes #2526 

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/commons/src/main/resources/com/powsybl/commons/reports_fr.properties
+++ b/commons/src/main/resources/com/powsybl/commons/reports_fr.properties
@@ -214,4 +214,4 @@ core.ucte.UcteReading = Lecture du fichier de réseau UCTE.
 core.ucte.UnsupportedTTBlock = Block TT non supporté.
 core.ucte.wrongAngleRegulationDu = ${transfoId} - Pour les LTCs, la tension de régulation de l'angle d'un transformateur par prise ne doit pas être nulle. Sa valeur absolue ne doit pas être supérieure à 6% (${du} %).
 core.ucte.wrongPhaseRegulationDu = ${transfoId} - Pour les LTCs, la tension de régulation de phase d'un transformateur par prise ne doit pas être nulle. Sa valeur absolue ne doit pas être supérieure à 6% (${du} %).
-core.iidm.network.lineNominalVoltageDifferent = La ligne '${lineId}' connecte les niveaux de tension '${voltageLevelId1}' (${nominalV1}) et '${voltageLevelId2}' (${nominalV2}) avec un �cart important de tension nominale (${gapPercent}%).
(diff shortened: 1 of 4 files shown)


samirromdhani:
After implementing ReportNode logging, I think this may no longer be a generic log, it is implementation specific behavior of a LoadFlow API and should be implemented in OpenLoadFlow.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java
```diff
@@ -944,6 +944,23 @@ public static ValidationLevel validate(Collection<Identifiable<?>> identifiables
         return validationLevel;
     }
 
+    public static void checkLineNominalVoltages(String lineId, VoltageLevel voltageLevel1, VoltageLevel voltageLevel2, double threshold, ReportNode reportNode) {
```
olperr1, line 947: Could you replace `lineId` by a `Validable`? This would be more consistent with the existing.
The ID could then be retrieved via `validable.getMessageHeader().getId()`.

geofjamg:
You can also look at IEEE14 matpower, there are many lines between different nominal voltages and this is no a problem and admitted by the community.

geofjamg:
> I don't have your test cases, so I don't know how many warnings this would generate.
> 
> For the IEEE14 (https://matpower.org/docs/ref/matpower5.0/case14.html), I see all buses have the same basekV (of 0 but anyway).
> 
> In powsybl, it does not matter since branch parameters are in natural units, but in matpower they are in pu, so the base voltage must be coherent (the docs do no specify from which side it is taken).
> 
> ```
>    Branch Data Format
>        1   f, from bus number
>        2   t, to bus number
>    (-)     (circuit identifier)
>        3   r, resistance (p.u.)
>        4   x, reactance (p.u.)
>        5   b, total line charging susceptance (p.u.)
> ```

Original model is indeed in PU, but there are usual additional base voltages taken for the IEEE14 like here https://github.com/powsybl/powsybl-core/blob/2ec235c70a0bca674a08692f63efd1ac4ab1c167/ieee-cdf/ieee-cdf-converter/src/main/java/com/powsybl/ieeecdf/converter/IeeeCdfNetworkFactory.java#L55

samirromdhani:
As suggested, and based on the current doc, it looks that connecting two voltage levels with different nominal voltages is valid. 
Also, as noted in the discussion the IEEE14 import shows that nominal voltage values does not follow any convention rules,
> the nominal voltage provider given here follows no convention rules but is an assumption

This PR will be closed

PR #3800, opened 2026-03-02
fix create line on line modification when with limits sets

EtienneLt:

no

Bug fix

only selected Operational limits group is copied to the split lines when using modification CreateLineOnLine

copy all operational limits groups on both side on both new lines

- [ ] Yes
- [x] No


olperr1:
Automatically closed by GitHub when the GridSuite PR was closed. This was not on purpose!

NathanDissoubray:
> Why not using
> 
> https://github.com/powsybl/powsybl-core/blob/f15566b4fd7b268156c51776eabddfbf54494bc7/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/LoadingLimitsUtil.java#L129
> directly ?

`LoadingLimitsUtil.copyOperationalLimits(copiedBranch, branch)` assumes that both branches exist when performing the copy operation.
This is not true here, since we need the delete the original branch before the two other branches can be created. Otherwise the nodes for the two new branches will not be available when creating them (since they will still be used by the original branch).

PR #3801, opened 2026-03-02
Create schema specific to the version to validate

### LINKED ISSUE #2818: Create schema specific to the version to validate

We use Suppliers.memoize to construct a Schema to validate a file against **any** iidm version

We use Suppliers.memoize with a map to construct a Schema to validate a file against a **specific** iidm version.

### Describe the motivation

In the unit tests we know against which version the file should be validated, so there's no need to load all the schema for validating against a specific version.

A user might also need to check the file he's using is validating a specific version xsd.

With such a map we also reduce the memory cost of Suppliers.memoize, as only the validated versions would be loaded.

### Extra Information

_No response_

samirromdhani:

Fixes #2818 

- [ ] Yes
- [x] No

**Other information**:

- Some extension XSD files reference IIDM XSD files from other versions (`xs:import` + `schemaLocation`), which can impact schema resolution . To handle this, `extractSchemaLocations` and `getSupportedExtensionsByIIdmVersion` were added.
- To make version validation strict, we verify the root namespace first, then run XSD validation, this is handled by `readRootNamespace` and `checkNamespace`


iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -85,7 +92,15 @@ public final class NetworkSerDe {
     static final byte[] BIIDM_MAGIC_NUMBER = {0x42, 0x69, 0x6e, 0x61, 0x72, 0x79, 0x20, 0x49, 0x49, 0x44, 0x4d};
 
     private static final Supplier<Schema> DEFAULT_SCHEMA_SUPPLIER = Suppliers.memoize(() -> NetworkSerDe.createSchema(DefaultExtensionsSupplier.getInstance()));
+    private static final Supplier<ConcurrentMap<IidmVersion, Schema>> DEFAULT_SCHEMAS_SUPPLIER = Suppliers.memoize(ConcurrentHashMap::new);
```
samirromdhani, line 98: - memoize stores the map instance once, the map starts empty, when a version is requested `computeIfAbsent` build it's Schema  After that, the same Schema is reused from the map.
==> caching is lazy by version

commons/src/main/java/com/powsybl/commons/xml/XmlUtil.java
```diff
@@ -180,6 +182,21 @@ public static XMLStreamWriter initializeWriter(boolean indent, String indentStri
         return initializeWriter(indent, indentString, xmlWriter);
     }
 
+    public static SchemaFactory newSchemaFactory() {
+        SchemaFactory factory = SchemaFactory.newInstance(XMLConstants.W3C_XML_SCHEMA_NS_URI);
+        try {
+            factory.setProperty(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "");
+        } catch (SAXException e) {
+            LOGGER.info("- Property unsupported by SchemaFactory implementation: {}", XMLConstants.ACCESS_EXTERNAL_SCHEMA);
+        }
```
rolnico, line 191: you can factorise this code in a small utility method used for the two properties

commons/src/main/java/com/powsybl/commons/xml/XmlUtil.java
```diff
@@ -180,6 +182,21 @@ public static XMLStreamWriter initializeWriter(boolean indent, String indentStri
         return initializeWriter(indent, indentString, xmlWriter);
     }
 
+    public static SchemaFactory newSchemaFactory() {
```
rolnico, line 185: Rename to `createSchemaFactoryInstance` to mirror `createXMLInputFactoryInstance`

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -151,6 +168,198 @@ private static Schema createSchema(ExtensionsSupplier extensionsSupplier) {
         }
     }
 
+    public static void validate(InputStream is, IidmVersion version, ExtensionsSupplier extensionsSupplier) {
```
rolnico, line 171: Add Javadoc for this method to explain its specificities.

Also, add a warning that it loads the full network file in memory.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -151,6 +168,198 @@ private static Schema createSchema(ExtensionsSupplier extensionsSupplier) {
...
+        }
+        return extensions;
+    }
+
+    private static void checkNamespace(byte[] xmlBytes, IidmVersion validationVersion) {
+        String actualNs = readRootNamespace(xmlBytes);
+        boolean matches = actualNs.equals(validationVersion.getNamespaceURI())
+                || validationVersion.supportEquipmentValidationLevel() && actualNs.equals(validationVersion.getNamespaceURI(false));
+        if (!matches) {
+            throw new PowsyblException("Namespace mismatch: expected validation version " + validationVersion.toString(".") + ", found namespace " + actualNs);
```
rolnico, line 270:
```suggestion
        String actualNamespace = readRootNamespace(xmlBytes);
        boolean matches = actualNamespace.equals(validationVersion.getNamespaceURI())
                || validationVersion.supportEquipmentValidationLevel() && actualNamespace.equals(validationVersion.getNamespaceURI(false));
        if (!matches) {
            throw new PowsyblException("Namespace mismatch: expected validation version " + validationVersion.toString(".") + ", found namespace " + actualNamespace);
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -151,6 +168,198 @@ private static Schema createSchema(ExtensionsSupplier extensionsSupplier) {
...
+     * Extract {@code xs:import/@schemaLocation} from XSD document
+     *
+     * <p>XSD document snippet:</p>
+     * <pre>{@code
+     * ...
+     *     targetNamespace="http://www.powsybl.org/schema/iidm/ext/extension-name/1_0"
+     *     xmlns:iidm="http://www.powsybl.org/schema/iidm/1_10">
+     *     <xs:import namespace="http://www.powsybl.org/schema/iidm/1_10" schemaLocation="iidm_V1_10.xsd"/>
+     * </xs:schema>
+     * }</pre>
```
rolnico, line 284: Good idea to illustrate with a small example!
Could you do the same with the other methods (when possible and useful)?

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -151,6 +168,198 @@ private static Schema createSchema(ExtensionsSupplier extensionsSupplier) {
...
+    public static void validate(InputStream is, IidmVersion version, ExtensionsSupplier extensionsSupplier) {
+        Objects.requireNonNull(is);
+        Objects.requireNonNull(version);
+        Objects.requireNonNull(extensionsSupplier);
+
+        // check version namespace
+        byte[] xmlBytes;
+        try {
+            xmlBytes = is.readAllBytes();
+            checkNamespace(xmlBytes, version);
```
rolnico, line 180: This does could be outside the try/catch as it does not throw an IOException

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -151,6 +168,198 @@ private static Schema createSchema(ExtensionsSupplier extensionsSupplier) {
...
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+        // XSD validation
+        Schema schema;
+        if (extensionsSupplier == DefaultExtensionsSupplier.getInstance()) {
+            schema = DEFAULT_SCHEMAS_SUPPLIER.get().computeIfAbsent(version, v -> createSchema(DefaultExtensionsSupplier.getInstance(), v));
+        } else {
+            schema = createSchema(extensionsSupplier, version);
+        }
```
rolnico, line 190:
```suggestion

        // XSD validation
        Schema schema = extensionsSupplier == DefaultExtensionsSupplier.getInstance() ?
            DEFAULT_SCHEMAS_SUPPLIER.get().computeIfAbsent(version, v -> createSchema(DefaultExtensionsSupplier.getInstance(), v)) :
            createSchema(extensionsSupplier, version);
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -151,6 +168,198 @@ private static Schema createSchema(ExtensionsSupplier extensionsSupplier) {
...
+        try {
+            schema.newValidator().validate(new StreamSource(new ByteArrayInputStream(xmlBytes)));
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        } catch (SAXException e) {
+            throw new UncheckedSaxException(e);
+        }
+    }
+
+    private static Schema createSchema(ExtensionsSupplier extensionsSupplier, IidmVersion version) {
```
rolnico, line 200: This method could be factorized with the existing one.
You might want to keep working on arrays instead of lists for efficiency (to be tested if possible)

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
rolnico, line 180: Is this check even necessary? There might be a better way to do this, avoiding loading the whole file in memory (which you are doing by creating the array of bytes).

For example something like this:

```java
    public static void validate(InputStream is, IidmVersion version, ExtensionsSupplier extensionsSupplier) {
        Objects.requireNonNull(is);
        Objects.requireNonNull(version);
        Objects.requireNonNull(extensionsSupplier);

        // XSD validation
        Schema schema = extensionsSupplier == DefaultExtensionsSupplier.getInstance() ?
            DEFAULT_SCHEMAS_SUPPLIER.get().computeIfAbsent(version, v -> createSchema(DefaultExtensionsSupplier.getInstance(), v)) :
            createSchema(extensionsSupplier, version);
        try {
            XMLFilter xmlFilter = getXMLFilter(version);
            SAXSource saxSource = new SAXSource(xmlFilter, new InputSource(is));
            schema.newValidator().validate(saxSource);
        } catch (IOException e) {
        } catch (SAXException e) {
            throw new UncheckedSaxException(e);
        } catch (ParserConfigurationException e) {
            throw new UncheckedParserConfigurationException(e);
        }
    }

    private static XMLFilter getXMLFilter(IidmVersion validationVersion) throws ParserConfigurationException, SAXException {
        XMLFilter filter = new XMLFilterImpl() {
            private boolean rootSeen = false;

            @Override
            public void startElement(String uri, String localName, String qName, Attributes atts)
                throws SAXException {
                if (!rootSeen) {
                    checkNamespace(uri, validationVersion);
                    rootSeen = true;
                }
                super.startElement(uri, localName, qName, atts);
            }
        };

        XMLReader xmlReader = SAXParserFactory.newNSInstance().newSAXParser().getXMLReader();
        filter.setParent(xmlReader);
        return filter;
    }

    private static void checkNamespace(String actualNamespace, IidmVersion validationVersion) {
        boolean matches = actualNamespace.equals(validationVersion.getNamespaceURI())
            || validationVersion.supportEquipmentValidationLevel() && actualNamespace.equals(validationVersion.getNamespaceURI(false));
        if (!matches) {
            throw new PowsyblException("Namespace mismatch: expected validation version " + validationVersion.toString(".") + ", found namespace " + actualNamespace);
        }
    }
```

Note: this is an example, this has to be checked/improved/validated!

rolnico (changes requested):

commons/src/main/java/com/powsybl/commons/xml/XmlUtil.java
```diff
@@ -180,6 +182,21 @@
         return initializeWriter(indent, indentString, xmlWriter);
     }
 
+    public static SchemaFactory createSchemaFactoryInstance() {
+        SchemaFactory factory = SchemaFactory.newInstance(XMLConstants.W3C_XML_SCHEMA_NS_URI);
+        setProperty(factory, XMLConstants.ACCESS_EXTERNAL_SCHEMA, "");
+        setProperty(factory, XMLConstants.ACCESS_EXTERNAL_DTD, "");
```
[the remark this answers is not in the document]
samirromdhani: Fixed when undo the proposed code factorization

commons/src/main/java/com/powsybl/commons/xml/XmlUtil.java
samirromdhani: done

commons/src/main/java/com/powsybl/commons/xml/XmlUtil.java
samirromdhani: done

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
samirromdhani: done

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
samirromdhani: done

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
samirromdhani: done

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
samirromdhani: done

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
samirromdhani: done

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
samirromdhani: done

commons/src/main/java/com/powsybl/commons/xml/XmlUtil.java
olperr1: Can you check `createXMLInputFactoryInstance()` and `createDocumentBuilderFactoryInstance()` methods from `XmlUtil`? We previously had vulnerabilities because some properties weren't set (see [this commit](https://github.com/powsybl/powsybl-core/commit/e6c7c4997ae8758b54a2f23ce1a499e25113acdc) and [this advisory](https://github.com/advisories/GHSA-qpj9-qcwx-8jv2)).

samirromdhani: I think nothing to change for `createXMLInputFactoryInstance()` and `createDocumentBuilderFactoryInstance()` but added new properties for `createXMLReader`

olperr1: Sorry, I was not precise enough.
What I meant was that these methods have multiple options that were carefully set to prevent vulnerabilities. We should check if some of these options must also be set in your new method.

samirromdhani: Yes, options should be present in the new method, i added them for `createXMLReader` and defined constants for those used in multiple places.

- xmlutil: adding security properties for XMLReader: https://github.com/powsybl/powsybl-core/pull/3801/changes/bed7452488fd30f3105a84754ed4a1a861e61f45
- xmlutil: define a constants: https://github.com/powsybl/powsybl-core/pull/3801/changes/f6b3b2300fdadbf479b33b770a2c3fbbebc2d201

commons/src/main/java/com/powsybl/commons/xml/XmlUtil.java
```diff
@@ -182,6 +194,30 @@ public static XMLStreamWriter initializeWriter(boolean indent, String indentStri
         return initializeWriter(indent, indentString, xmlWriter);
     }
 
+    public static XMLReader createXMLReader() throws ParserConfigurationException, SAXException {
```
samirromdhani, line 196: [XML Parser Security Features Matrix](https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#xml-parser-security-features-matrix)

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
samirromdhani: I agree to avoid loading the whole file into memory.
I added an XMLReader instance in XMLUtil with security properties.
```java
XMLReader xmlReader = SAXParserFactory.newNSInstance().newSAXParser().getXMLReader();
```

commons/src/main/java/com/powsybl/commons/xml/XmlUtil.java
samirromdhani: Hello, I added tests for new methods and make sure that options are present: SAXParser, XMLReader

commons/src/main/java/com/powsybl/commons/xml/XmlUtil.java
```diff
@@ -48,6 +55,11 @@ public final class XmlUtil {
     private static final Supplier<XMLOutputFactory> XML_OUTPUT_FACTORY_SUPPLIER = Suppliers.memoize(XmlUtil::createXMLOutputFactoryInstance);
     private static final Supplier<DocumentBuilderFactory> DOCUMENT_BUILDER_FACTORY_SUPPLIER = Suppliers.memoize(XmlUtil::createDocumentBuilderFactoryInstance);
     private static final Supplier<CSVFormat> CSV_FORMAT_SUPPLIER = Suppliers.memoize(XmlUtil::createCsvFormatInstance);
+    private static final Supplier<SchemaFactory> SCHEMA_FACTORY_SUPPLIER = Suppliers.memoize(XmlUtil::createSchemaFactoryInstance);
+
```
rolnico:
```suggestion
```

commons/src/main/java/com/powsybl/commons/xml/XmlUtil.java
```diff
@@ -182,6 +194,35 @@ public static XMLStreamWriter initializeWriter(boolean indent, String indentStri
...
+            saxParser.setProperty(XMLConstants.ACCESS_EXTERNAL_DTD, "");
+        } catch (SAXException e) {
+            logUnsupportedSaxParserProperty(XMLConstants.ACCESS_EXTERNAL_DTD);
+        }
+        // Create XMLReader from parser
+        XMLReader xmlReader = saxParser.getXMLReader();
+        xmlReader.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+        xmlReader.setFeature(XML_DISALLOW_DOCTYPE, true);
+        xmlReader.setFeature(SAX_EXTERNAL_GENERAL_ENTITIES, false);
+        xmlReader.setFeature(SAX_EXTERNAL_PARAMETER_ENTITIES, false);
```
rolnico, line 221: Is it necessary to add the features on both the factory and the xmlReader?

commons/src/main/java/com/powsybl/commons/xml/XmlUtil.java
```diff
@@ -260,23 +301,39 @@ private static DocumentBuilderFactory createDocumentBuilderFactoryInstance() {
...
-        setFeature(factory, "http://apache.org/xml/features/disallow-doctype-decl", true);
-        setFeature(factory, "http://xml.org/sax/features/external-general-entities", false);
-        setFeature(factory, "http://xml.org/sax/features/external-parameter-entities", false);
+        setFeatures(factory);
         factory.setXIncludeAware(false);
         factory.setExpandEntityReferences(false);
         return factory;
     }
 
+    private static void setFeatures(DocumentBuilderFactory factory) {
```
rolnico: Why moving this code to a specific method? It seems to be only used once

commons/src/main/java/com/powsybl/commons/xml/XmlUtil.java
```diff
@@ -288,4 +345,33 @@ public static CSVFormat getCsvFormat() {
...
+        }
+        try {
+            factory.setProperty(XMLConstants.ACCESS_EXTERNAL_DTD, "");
+        } catch (SAXException e) {
+            logUnsupportedSchemaFactoryProperty(XMLConstants.ACCESS_EXTERNAL_DTD);
+        }
+        return factory;
+    }
+
+    private static void logUnsupportedSchemaFactoryProperty(String property) {
```
rolnico: regroup the other `logUnsupported...` with these one

commons/src/main/java/com/powsybl/commons/xml/XmlUtil.java
```diff
@@ -288,4 +345,33 @@ public static CSVFormat getCsvFormat() {
...
+
+    public static SchemaFactory createSchemaFactoryInstance() {
+        SchemaFactory factory = SchemaFactory.newInstance(XMLConstants.W3C_XML_SCHEMA_NS_URI);
+        setFeature(factory, XMLConstants.FEATURE_SECURE_PROCESSING, true);
+        setFeature(factory, XML_DISALLOW_DOCTYPE, true);
+        try {
+            factory.setProperty(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "");
+        } catch (SAXException e) {
+            logUnsupportedSchemaFactoryProperty(XMLConstants.ACCESS_EXTERNAL_SCHEMA);
+        }
```
rolnico, line 352: Maybe use a small `setProperty` utility method like in the other cases? Same thing for `createXMLReader()`.
And regroup all those small private utility methods at the bottom of the class for consistency

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -151,6 +177,208 @@ private static Schema createSchema(ExtensionsSupplier extensionsSupplier) {
...
+            }
+            return factory.newSchema(sources);
+        } catch (SAXException e) {
+            throw new UncheckedSaxException(e);
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
+
+     /**
```
rolnico:
```suggestion
    /**
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -151,6 +177,208 @@ private static Schema createSchema(ExtensionsSupplier extensionsSupplier) {
...
+     * try (InputStream is = getClass().getResourceAsStream(path)) {
+     *     NetworkSerDe.validate(is, IidmVersion.V_1_16, DefaultExtensionsSupplier.getInstance());
+     * }
+     * }</pre>
+     *
+     * @param is the IIDM network input stream to validate
+     * @param version the IIDM version used for validation
+     * @param extensionsSupplier the extension supplier used to resolve extension schemas
+     */
+    public static void validate(InputStream is, IidmVersion version, ExtensionsSupplier extensionsSupplier) {
```
rolnico, line 196: You finally managed to validate both the namespace and the content without loading the whole network in memory? That's very nice!

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/NetworkSerDeTest.java
```diff
@@ -421,4 +425,103 @@ void testExportWithoutFlatten() {
         assertEquals(2, merged.getSubnetworks().size());
         assertEquals(2, readNetwork.getSubnetworks().size());
     }
+
+    @Test
+    void testValidateByVersionWhenMatchingOldNetwork() throws IOException {
+        try (InputStream is = getClass().getResourceAsStream("/V1_2/shuntRoundTripRef.xml")) {
+            assertNotNull(is);
```
rolnico: Is it really needed?

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/NetworkSerDeTest.java
```diff
@@ -421,4 +425,103 @@ void testExportWithoutFlatten() {
         assertEquals(2, merged.getSubnetworks().size());
         assertEquals(2, readNetwork.getSubnetworks().size());
     }
+
+    @Test
+    void testValidateByVersionWhenMatchingOldNetwork() throws IOException {
```
rolnico:
```suggestion
    void testValidateByVersionOnIidm102() throws IOException {
```

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/NetworkSerDeTest.java
```diff
@@ -421,4 +425,103 @@ void testExportWithoutFlatten() {
...
+        try (InputStream is = getClass().getResourceAsStream("/V1_2/shuntRoundTripRef.xml")) {
+            assertNotNull(is);
+            assertDoesNotThrow(() -> NetworkSerDe.validate(is, IidmVersion.V_1_2));
+        }
+    }
+
+    @Test
+    void testValidateByVersionWhenMatchingRecentNetwork() throws IOException {
```
rolnico:
```suggestion
    void testValidateByVersionOnIidm116() throws IOException {
```

rolnico (changes requested):

commons/src/main/java/com/powsybl/commons/xml/XmlUtil.java
samirromdhani: I grouped them to make the features easier to review in a single block. I will inline that now i think there is no changes in DocumentBuilderFactory's feaures

commons/src/main/java/com/powsybl/commons/xml/XmlUtil.java
samirromdhani: If I wrap these calls in utility methods, Sonar will trigger and report vulnerabilities.

commons/src/main/java/com/powsybl/commons/xml/XmlUtil.java
samirromdhani: Yes, I think it is safer to set them on both, as mentionned here https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html

> These features can either be set on the factory or the underlying XMLReader setFeature method.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/NetworkSerDeTest.java
samirromdhani: It is not needed, but it gives a clearer failure message if the test resource is missing.

rolnico: I understand, but here the assertions are usually done on the elements you want to test.
You could replace it with:

```java
Objects.requireNonNull(getClass().getResourceAsStream("/V1_2/shuntRoundTripRef.xml"))
```

(same in other tests)

rolnico (changes requested):

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/NetworkSerDeTest.java
samirromdhani: done

rolnico (approved):

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -151,6 +177,208 @@ private static Schema createSchema(ExtensionsSupplier extensionsSupplier) {
...
+     *
+     * @param extensionsSupplier extension provider used to discover available extension
+     * @param version IIDM version used to filter compatible extensions
+     * @return list of additional schema sources required by extension
+     */
+    private static List<Source> getExtensionSources(ExtensionsSupplier extensionsSupplier, IidmVersion version) throws IOException {
+        List<Source> sources = new ArrayList<>();
+        for (InputStream is : getSupportedExtensionByIIdmVersion(extensionsSupplier, version)) {
+            byte[] extensionXsd = is.readAllBytes();
+            // required iidm xsd in extension's xsd: source
```
samirromdhani, line 262: Note that required iidm xsd in extension's xsd force us to load additional resources, to avoid that we can discuss this feature in #3811.

rolnico (approved):

PR #3804, opened 2026-03-02
Rewrite binary serialization

clementleclercRTE:

  Feature — redesign of the binary IIDM serialization format (BinReader / BinWriter).

  The binary format uses a positional (sequential) attribute encoding: attributes are written in a fixed order and read back in the same order. The name parameter passed to read*Attribute("name") is
  completely ignored — the reader blindly consumes bytes from the stream. As a result:
  - Reading attributes out of order silently produces corrupted values with no exception.
  - skipNode() is not implemented and throws PowsyblException, making it impossible to skip unknown or private extensions.
  - Absent values (NaN, absentValue, null) are still written to the stream, wasting space.

 Attributes are now named: a global dictionary maps attribute names to compact 1-byte indices. Each attribute occurrence in the body is encoded as [byte: index][byte: type][value]. Before a node's children are processed, all its attributes are pre-read into a Map<String, Object> keyed by name. 

- [ ] Yes
- [ ] No

  The binary file format  is not backward compatible. Binary files produced before this PR cannot be read by the new code, and vice versa. Since the binary format was already marked as BETA with no stability guarantee across releases, no migration path is provided. Users should re-export any existing .biidm files using the new version.

**Other information**:


commons/src/main/java/com/powsybl/commons/binary/BinReader.java
```diff
@@ -14,20 +14,24 @@
...
     private final DataInputStream dis;
-    private final Map<Integer, String> dictionary = new HashMap<>();
     private final byte[] binaryMagicNumber;
 
+    private final Map<Integer, String> nodeDictionary = new HashMap<>();
+
+    private String[] attrNames;
+    private byte[] attrTypes;
+
+    private Map<String, Object> currentAttrs = Collections.emptyMap();
```
flo-dup: Don't buffer attributes, just read them on the fly

commons/src/main/java/com/powsybl/commons/binary/BinWriter.java
```diff
@@ -113,115 +83,220 @@ public void writeStartNode(String namespace, String name) {
...
+        }
+    }
+
+    private void writeAttrIndex(String name, byte type) {
+        int index = attrNamesIndex.computeIfAbsent(new AttrKey(name, type), k -> 1 + attrNamesIndex.size());
+        if (index > 255) {
+            throw new PowsyblException("Binary format: too many distinct attribute (name, type) pairs (max 255)");
+        }
+        try {
+            attrDos.writeByte(index);
```
NathanDissoubray: you are checking that the index is not more than 255, but in java Byte is signed, shouldn't it be 127 instead ? How about using Byte.MAX_VALUE instead ?

clementleclercRTE: I think the check index > 255 is correct here because BinReader uses readUnsignedByte(). So the full range of 255 distinct attribute pairs is usable.

NathanDissoubray: you are right ! I did not see the `readUnsignedByte` in the reader

commons/src/main/java/com/powsybl/commons/binary/BinReader.java
```diff
@@ -61,194 +69,226 @@ public Map<String, String> readExtensionVersions() throws IOException {
...
+        attrNames = new String[nbEntries + 1];
+        attrTypes = new byte[nbEntries + 1];
+        for (int i = 0; i < nbEntries; i++) {
+            attrNames[i + 1] = readString();
+            attrTypes[i + 1] = dis.readByte();
         }
     }
 
-    private float readFloat() {
+    private Object readAttrValue(String name) {
```
flo-dup: Please delete this method to avoid boxing the values and avoid factorizing methods which should not be factorized. When you're in readDouble, call readDouble directly! rather than calling a method which will switch case, box to Double and then instanceof Double to switch back to double.

commons/src/main/java/com/powsybl/commons/binary/BinReader.java
```diff
@@ -61,194 +69,226 @@ public Map<String, String> readExtensionVersions() throws IOException {
...
 
-    private int readInt() {
-        try {
-            return dis.readInt();
-        } catch (IOException e) {
-            throw new UncheckedIOException(e);
+    private void skipRemainingAttributes() throws IOException {
+        while (nextAttrIdx != END_ATTRS) {
+            readTypedValue(attrTypes[nextAttrIdx]);
+            nextAttrIdx = dis.readUnsignedByte();
```
flo-dup: Attribute index is coded on a byte, I think that's might be soon not enough, if not already? How many attributes exist in the real network files you tested?

commons/src/main/java/com/powsybl/commons/binary/BinUtil.java
```diff
@@ -16,6 +16,18 @@ private BinUtil() {
...
+    static final byte TYPE_FLOAT = 1;
+    static final byte TYPE_INT = 2;
+    static final byte TYPE_BOOLEAN = 3;
+    static final byte TYPE_STRING = 4;
+    static final byte TYPE_ENUM = 5;
+    static final byte TYPE_INT_ARRAY = 6;
+    static final byte TYPE_STRING_ARRAY = 7;
+
+    // NUL prefix ensures this name cannot collide with any real XML attribute name
+    static final String CONTENT_ATTR_NAME = "\u0000content";
```
flo-dup: you should avoid this hack, this does not really ensure anything... In JSON, the content is written under a "content" field, which is not that nice either, but do we need to consider this as an attribute here? Couldn't we use a different type, like TYPE_STRING_CONTENT?

commons/src/main/java/com/powsybl/commons/binary/BinWriter.java
flo-dup: why not writing directly to tmpDos?

commons/src/main/java/com/powsybl/commons/binary/BinWriter.java
```diff
@@ -113,115 +83,220 @@ public void writeStartNode(String namespace, String name) {
...
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
     }
 
     @Override
     public void writeOptionalIntAttribute(String name, Integer value) {
+        if (value == null) {
+            return;
+        }
```
flo-dup: I think you could avoid all these redundant checks by inheriting from AbstractTreeDataWriter

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -319,7 +319,6 @@ private static JsonWriter createJsonWriter(OutputStream os, ExportOptions option
     }
 
     private static TreeDataWriter createBinWriter(OutputStream os, ExportOptions options) {
-        LOGGER.warn("BETA feature, the resulting binary file is not guaranteed to still be readable in the next releases");
```
flo-dup, line 322: 🎉

flo-dup (changes requested):

commons/src/test/java/com/powsybl/commons/binary/BinWriterReaderTest.java
```diff
@@ -0,0 +1,239 @@
...
+        reader.readChildNodes(nodeName -> {
+            visited.add(nodeName);
+            reader.readEndNode();
+        });
+        assertEquals(List.of("child1", "child2"), visited);
+        reader.close();
+    }
+
+    @Test
+    void testSkipNode() {
```
flo-dup, line 195: as mentioned by sonar, could you add an assert? detecting that we're at the end of the stream as we've skipped everything

commons/src/main/java/com/powsybl/commons/binary/BinReader.java
```diff
@@ -14,31 +14,39 @@
...
     private final byte[] binaryMagicNumber;
 
+    private String[] nodeNames;
+
+    private String[] attrNames;
+    private byte[] attrTypes;
+
+    private int nextAttrIdx = END_ATTRS;
+
+    private static final int BUFFER_SIZE = 1024 * 1024;
```
flo-dup: This optimization is interesting! but this is something out of current PR's scope

commons/src/main/java/com/powsybl/commons/binary/BinUtil.java
```diff
@@ -16,6 +16,17 @@ private BinUtil() {
     }
 
     static final int END_NODE = 0;
-    static final int NULL_ENUM = -1;
+    static final int END_ATTRS = 0;
```
flo-dup: This indicates the end of attributes, but for bin and json there's no big difference between a child node and an attribute. I think you could give a try in this PR to merge for bin writer/reader the logic of nodes and attributes, that is:
- having a single dictionary for both attributes and nodes names
- having a TYPE_OBJECT for nodes

That would avoid the end attributes complexity, and the "flush" method in writing.

commons/src/main/java/com/powsybl/commons/binary/BinReader.java
```diff
@@ -14,31 +14,39 @@
...
-import java.util.function.Supplier;
 
-import static com.powsybl.commons.binary.BinUtil.END_NODE;
-import static com.powsybl.commons.binary.BinUtil.NULL_ENUM;
+import static com.powsybl.commons.binary.BinUtil.*;
 
 /**
  * @author Florian Dupuy {@literal <florian.dupuy at rte-france.com>}
  */
 public class BinReader implements TreeDataReader {
```
flo-dup: Similarly to BinWriter, it now makes sense that BinReader extends AbstractTreeDataReader.

commons/src/main/java/com/powsybl/commons/binary/BinReader.java
```diff
@@ -61,194 +69,302 @@ public Map<String, String> readExtensionVersions() throws IOException {
...
+            nextAttrIdx = dis.readUnsignedShort();
         }
     }
 
-    private int readInt() {
-        try {
-            return dis.readInt();
-        } catch (IOException e) {
-            throw new UncheckedIOException(e);
+    private void skipTypedValue(byte typeTag) throws IOException {
```
flo-dup, line 118: You were right to rename this in skip! Indeed we don't need the result anymore. But you need to go further now: `dis.readDouble()` is now doing too much for what's asked by this method, just replace that by `dis.skipNBytes(8)` for instance. That would avoid the extra logic to transform bytes into values/objects. Skipping string and arrays needs specific skipping, but that remains easy

commons/src/main/java/com/powsybl/commons/binary/BinReader.java
```diff
@@ -61,194 +69,302 @@ public Map<String, String> readExtensionVersions() throws IOException {
...
+            return val;
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
     }
 
     @Override
     public int readIntAttribute(String name) {
-        return readInt();
+        return readIntAttribute(name, 0);
```
flo-dup: Warning, with this new binary design, we should handle the missing attributes like in JSON and XML. If we don't find the attribute for ints and booleans we're suppose to throw an exception. For other types for some reason there's a fallback on Double.NaN or Float.NaN or empty list or null. This probably needs to be uniform for all types, but let's keep something coherent betwen formats for now

commons/src/main/java/com/powsybl/commons/binary/BinWriter.java
```diff
@@ -113,115 +80,178 @@ public void writeStartNode(String namespace, String name) {
...
+        attrBlockTerminated = false;
     }
 
     @Override
     public void writeEndNode() {
+        flushCurrentNodeAttrsIfNeeded();
         writeIndex(END_NODE, tmpDos);
     }
 
+    private void flushCurrentNodeAttrsIfNeeded() {
```
flo-dup: Now that there's no attributes buffering anymore, I think you should be able to remove this method. I'm not so sure you really need the END_ATTRS, maybe I'm missing something? In binary (and json) there's no difference between attributes and child nodes, hence my proposition of merging the two names maps.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/NetworkSerDeTest.java
```diff
@@ -91,7 +91,7 @@ void writeMultipleSelectedOperationalLimitsGroupToOlderFormat() throws IOExcepti
     }
 
     @ParameterizedTest
-    @EnumSource(value = TreeDataFormat.class, names = {"XML", "JSON"})
```
flo-dup, line 94: well spotted 👍

flo-dup (changes requested):

NathanDissoubray (dismissed):
I can't suggest a change on a file that was not changed in the PR, but please modify the Read the Doc which specifies that BIIDM is in beta (since with this PR, it won't be beta anymore). It's in `docs/grid_exchange_formats/iidm/index.md`

commons/src/test/java/com/powsybl/commons/binary/BinWriterReaderTest.java
```diff
@@ -0,0 +1,240 @@
...
+            writer.writeIntAttribute("x", 42);
+            writer.writeStartNode(null, "grandchild");
+            writer.writeBooleanAttribute("flag", true);
+            writer.writeEndNode();
+            writer.writeEndNode();
+        });
+
+        // skipNode must recursively skip attrs and children at all depths
+        reader.readChildNodes(nodeName -> reader.skipNode());
+        assertTrue(reader.readEndOfStream());
```
flo-dup: You added a new method only for this unit test. I think we can test the reader without that method. To test you went through all the nodes, you can pile up the node names encountered in a list, something like:
```java
reader.readChildNodes(nodeName -> {
  reader.skipNode():
  nodesNameSkipped.add(nodeName);
});
```
And then assert on that list.

flo-dup (changes requested):

commons/src/main/java/com/powsybl/commons/binary/BinReader.java
```diff
@@ -85,172 +186,223 @@ private String readString() {
...
+        readChildNodes(nodeName -> skipNode());
     }
 
     @Override
     public void readChildNodes(ChildNodeReader childNodeReader) {
         try {
-            int nodeNameIndex;
-            while ((nodeNameIndex = dis.readShort()) != END_NODE) {
-                String nodeName = dictionary.get(nodeNameIndex);
+            skipRemainingAttributes();
```
flo-dup: If there are remaining attributes, they are unexpected so we should throw an exception
```suggestion
            if (nextNameIdx != END_NODE && nextType != TYPE_OBJECT) {
                throw new PowsyblException("Binary parsing: expected child node but got attribute " + names[nextNameIdx]);
            }
```

commons/src/main/java/com/powsybl/commons/binary/BinReader.java
```diff
@@ -259,10 +411,11 @@ public void readChildNodes(ChildNodeReader childNodeReader) {
     @Override
     public void readEndNode() {
         try {
-            int nextIndex = dis.readShort();
-            if (nextIndex != END_NODE) {
-                throw new PowsyblException("Binary parsing: expected end node but got " + nextIndex);
+            skipRemainingAttributes();
```
flo-dup: We should throw an exception if there are any remaining attributes

commons/src/main/java/com/powsybl/commons/binary/BinReader.java
```diff
@@ -52,31 +65,119 @@ private void readMagicNumber() throws IOException {
...
-        return versions;
     }
 
-    private void readDictionary() throws IOException {
-        int nbEntries = dis.readShort();
+    private void readNamesDictionary() throws IOException {
+        int nbEntries = dis.readUnsignedShort();
+        names = new String[nbEntries + 1];
+        types = new byte[nbEntries + 1];
         for (int i = 0; i < nbEntries; i++) {
```
flo-dup, line 87: Add a comment saying that the 0 entry is for the end node

commons/src/main/java/com/powsybl/commons/binary/BinReader.java
```diff
@@ -85,172 +186,223 @@ private String readString() {
...
+            return val;
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
     }
 
     @Override
     public void skipNode() {
-        throw new PowsyblException("Binary format does not support skipping child nodes");
+        readChildNodes(nodeName -> skipNode());
```
flo-dup, line 422:
```suggestion
        skipRemainingAttributes();
        readChildNodes(nodeName -> skipNode());
```

commons/src/main/java/com/powsybl/commons/binary/BinReader.java
```diff
@@ -52,31 +65,119 @@ private void readMagicNumber() throws IOException {
...
+            return true;
+        }
+        String entryName = names[nextNameIdx];
+        if (entryName == null) {
+            throw new PowsyblException("Cannot read attribute: unknown name index " + nextNameIdx);
+        }
+        return !name.equals(entryName);
+    }
+
+    private void skipRemainingAttributes() throws IOException {
```
flo-dup: Catch the IOException within the method skipRemainingAttributes method
```suggestion
    private void skipRemainingAttributes() {
       try {
```

commons/src/main/java/com/powsybl/commons/binary/BinReader.java
```diff
@@ -52,31 +65,119 @@ private void readMagicNumber() throws IOException {
...
     }
 
+    private void peekNextEntry() throws IOException {
+        try {
+            nextNameIdx = dis.readUnsignedShort();
+            if (nextNameIdx != END_NODE) {
+                nextType = types[nextNameIdx];
+            }
+        } catch (EOFException e) {
+            nextNameIdx = END_NODE;
```
flo-dup: The EOF exception doesn't mean it's an END_NODE, otherwise a missing END_NODE would be ok. So you should rather add another type END_OF_FILE
```suggestion
            nextNameIdx = END_OF_FILE;
```

commons/src/test/java/com/powsybl/commons/binary/BinWriterReaderTest.java
```diff
@@ -0,0 +1,249 @@
...
+            writer.writeFloatAttribute("c", 3.0f);
+            writer.writeBooleanAttribute("d", true);
+            writer.writeStringAttribute("e", "skip");
+            writer.writeEnumAttribute("f", Thread.State.RUNNABLE);
+            writer.writeIntArrayAttribute("g", List.of(1, 2));
+            writer.writeStringArrayAttribute("h", List.of("x"));
+        });
+
+        assertEquals(1, reader.readIntAttribute("a"));
+        // All remaining attrs skipped by readEndNode
```
flo-dup: No, this should throw an exception if there are remaining attributes!

commons/src/test/java/com/powsybl/commons/binary/BinWriterReaderTest.java
```diff
@@ -0,0 +1,249 @@
...
+            writer.writeEndNode();
+            writer.writeStartNode(null, "child2");
+            writer.writeStringAttribute("y", "hello");
+            writer.writeEndNode();
+        });
+
+        List<String> visited = new ArrayList<>();
+        reader.readChildNodes(nodeName -> {
+            visited.add(nodeName);
+            reader.readEndNode();
```
flo-dup:
```suggestion
            reader.skipNode();
```

commons/src/main/java/com/powsybl/commons/binary/BinReader.java
```diff
@@ -52,31 +65,119 @@ private void readMagicNumber() throws IOException {
...
+        }
+    }
+
+    private boolean isAttrAbsent(String name) {
+        if (nextNameIdx == END_NODE || nextType == TYPE_OBJECT) {
+            return true;
+        }
+        String entryName = names[nextNameIdx];
+        if (entryName == null) {
```
flo-dup: entryName cannot be null

commons/src/main/java/com/powsybl/commons/binary/BinReader.java
```diff
@@ -85,172 +186,223 @@ private String readString() {
...
         try {
-            int nodeNameIndex;
-            while ((nodeNameIndex = dis.readShort()) != END_NODE) {
-                String nodeName = dictionary.get(nodeNameIndex);
+            skipRemainingAttributes();
+            while (nextNameIdx != END_NODE) {
+                String nodeName = names[nextNameIdx];
                 if (nodeName == null) {
-                    throw new PowsyblException("Cannot read child node: unknown element name index " + nodeNameIndex);
+                    throw new PowsyblException("Cannot read child node: unknown name index " + nextNameIdx);
```
flo-dup: you can remove it here also

flo-dup (changes requested):

commons/src/main/java/com/powsybl/commons/binary/BinReader.java
```diff
@@ -52,31 +65,119 @@ private void readMagicNumber() throws IOException {
...
 
+    private void peekNextEntry() throws IOException {
+        try {
+            nextNameIdx = dis.readUnsignedShort();
+            if (nextNameIdx != END_NODE) {
+                nextType = types[nextNameIdx];
+            }
+        } catch (EOFException e) {
+            nextNameIdx = END_NODE;
+        }
```
flo-dup: Also catch index out of bound:

```suggestion
        } catch (ArrayIndexOutOfBoundsException e) {
            throw new PowsyblException("Corrupted binary file: invalid index " + nextNameIdx
                    + " (max " + (names.length - 1) + ")");
        }
```

flo-dup (changes requested):

commons/src/main/java/com/powsybl/commons/binary/BinReader.java
```diff
@@ -90,31 +91,46 @@ private void readNamesDictionary() throws IOException {
...
+        int b1 = dis.read();
+        if (b1 < 0) {
+            nextNameIdx = END_OF_FILE;
+            return;
+        }
+        int b2 = dis.read();
+        if (b2 < 0) {
+            throw new PowsyblException("Corrupted binary file: truncated name index");
+        }
+        nextNameIdx = (b1 << 8) | b2;
```
flo-dup: Encapsulate this in a readNameIndex method

commons/src/main/java/com/powsybl/commons/binary/BinReader.java
```diff
@@ -90,31 +91,46 @@ private void readNamesDictionary() throws IOException {
...
             }
-        } catch (EOFException e) {
-            nextNameIdx = END_NODE;
+            nextType = types[nextNameIdx];
         }
     }
 
     private boolean isAttrAbsent(String name) {
-        if (nextNameIdx == END_NODE || nextType == TYPE_OBJECT) {
+        if (nextNameIdx == END_NODE || nextNameIdx == END_OF_FILE || nextType == TYPE_OBJECT) {
```
flo-dup: if == END_OF_FILE should lead to a unexpected end of file PowsyblException

flo-dup (changes requested):

commons/src/main/java/com/powsybl/commons/binary/BinReader.java
```diff
@@ -52,217 +65,372 @@ private void readMagicNumber() throws IOException {
...
     public <T extends Enum<T>> T readEnumAttribute(String name, Class<T> clazz, T defaultValue) {
-        return readEnum(clazz);
+        try {
+            if (isAttrAbsent(name)) {
+                return defaultValue;
+            }
+            int ordinal = dis.readUnsignedShort();
+            peekNextEntry();
+            T[] constants = clazz.getEnumConstants();
+            return ordinal < constants.length ? constants[ordinal] : defaultValue;
```
flo-dup: if ordinal is too big you should throw an exception:
```java
throw new PowsyblException("Invalid enum ordinal for " + clazz.getSimpleName() + ": "
        + ordinal + " (max " + (constants.length - 1) + ")");
```

flo-dup (approved):

PR #3805, opened 2026-03-03
Fix voltage limit import

### LINKED ISSUE #3787: CGMES import shouldn't allow VoltageLevel's LowVoltageLimit greater than HighVoltageLimit

In the case of a CGMES VoltageLevel with 2 ConductingEquipment (such as BusbarSection) having VoltageLimit as follows:
|Set|Low|High|
|----|-----|----|
|Set 1|20|40|
|Set 2|60|80|

The most restrictive limits are kept: the lowest high VoltageLimit and the highest low VoltageLimit:
|Low|High|
|-----|----|
|60|40|

which leads to having a LowVoltageLimit being higher than the HighVoltageLimit.

This situation shouldn't be allowed. A control should be implemented to verify that the retained LowVoltageLimit is lower than the retained HighVoltageLimit. In the case of inconsistent limits, the VoltageLevel's own limits should be kept if they exist and they are consistent, else no limit should be set.

### Describe the steps

_No response_

### Environment

_No response_

### Relevant Log Output

_No response_

### Extra Information

_No response_

rcourtier:

Fixes #3787.

Bug fix.

Incorrect order of low/high voltage limits makes the CGMES import crash.

* `VoltageLevel` `highVoltageLimit` and `lowVoltageLimit` aren't imported if they are not correctly ordered.
* `VoltageLimit` EQ `value` (CGMES 2.4.15) or `normalValue` (CGMES 3.0) is not applied on the fly anymore. Instead, the most restrictive values inside the `VoltageLevel` limits (which are saved in a property) are now considered after all `VoltageLimit` have been processed. If the limit order is respected, the values are applied, else the properties holding the values are removed.
* Same happens for `VoltageLimit` SSH `value` (CGMES 3.0) : the most restrictve values inside the `VoltageLevel` limits are applied only if the order is respected, else the default values are applied (these can be the previous values or the normal EQ value depending on the import options).

- [X] Yes
- [ ] No

Public class `LoadingLimitsMapping` from `com.powsybl.cgmes.conversion` package has been renamed `LimitsMapping`.

As a consequence, `Context.loadingLimitsMapping()` has been renamed `limitsMapping()`.

**Other information**:


cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
```diff
@@ -462,8 +487,26 @@ private static String className(Identifiable<?> o) {
...
+
+        String lowLimitPropertyName = voltageLevel.hasProperty(PROPERTY_NORMAL_VALUE_LOW_VOLTAGE_LIMIT) ?
+            PROPERTY_NORMAL_VALUE_LOW_VOLTAGE_LIMIT : PROPERTY_LOW_VOLTAGE_LIMIT;
+        Double normalLowLimitValue = getNormalLimitValue(voltageLevel, lowLimitPropertyName);
+        Double defaultLowLimitValue = getDefaultValue(normalLowLimitValue, voltageLevel.getLowVoltageLimit(), context);
+        Double lowLimitValue = getUpdatedVoltageLimitValue(voltageLevel, PROPERTY_OPERATIONAL_LIMIT_LOW_VOLTAGE_LIMIT, Double::max, context);
+
+        // Update value if it's defined and if the order is respected, otherwise use default value.
+        voltageLevel.setHighVoltageLimit(Double.isNaN(highLimitValue) || lowLimitValue > highLimitValue ? defaultHighLimitValue : highLimitValue);
+        voltageLevel.setLowVoltageLimit(Double.isNaN(lowLimitValue) || lowLimitValue > highLimitValue ? defaultLowLimitValue : lowLimitValue);
```
marqueslanauja: I’m wondering whether it is possible for the `defaultHighLimitValue` and the `lowLimitValue` to be selected in the wrong order. The same applies to `highLimitValue` and `defaultLowLimitValue`.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
```diff
@@ -514,38 +557,23 @@ private static double getDefaultValue(Double normalValue, double previousValue,
...
-
-    private static Double getNormalValue(VoltageLevel voltageLevel, String propertyName) {
-        return voltageLevel.getProperty(propertyName) != null ? Double.parseDouble(voltageLevel.getProperty(propertyName)) : null;
+            .map(operationalLimitId -> updatedValue(operationalLimitId, context))
+            .filter(OptionalDouble::isPresent)
+            .mapToDouble(OptionalDouble::getAsDouble)
+            .filter(value -> (Double.isNaN(voltageLevelHighLimit) || value < voltageLevelHighLimit)
+                && (Double.isNaN(voltageLevelLowLimit) || value > voltageLevelLowLimit))
+            .reduce(operator)
+            .orElse(Double.NaN);
```
marqueslanauja: According to your explanations, I understand that the updated limit values must be within the VoltageLevel limits (`voltageLevelLowLimit`, `voltageLevelHighLimit`) if they are defined in the EQ file. Is this a CGMES rule?

rcourtier: I haven't found any such rule in IEC 61970-600-2 or in QoCDC. So since nothing contradicts it, my opinion is that the PowSyBl limit is the most restrictive of the CGMES limits:
- In case of an EQ import, we pick the most restrictive limit between the `VoltageLevel` limit and the `VoltageLimit` values.
- In case of a SSH update, the same applies. Only the SSH `VoltageLimit.value` supersedes the EQ `VoltageLimit.normalValue`.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
rcourtier: I have changed it a little bit to make sure that limits are consistent and updated in the correct order.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/elements/OperationalLimitConversion.java
marqueslanauja: ok

olperr1 (approved):

PR #3809, opened 2026-03-04
Fix V / Angle set on wrong bus in busbreakerview after view invalidation

### LINKED ISSUE #3200: V / Angle set on wrong bus in busbreakerview after view invalidation

        vl.getNodeBreakerView().newBusbarSection().setId("BBS1").setNode(1).add();
        vl.getNodeBreakerView().newBreaker().setId("BR1").setRetained(true).setOpen(false).setNode1(1).setNode2(2).add();
        System.out.println(vl.getBusBreakerView().getBus("VL_1").setV(123).getV());
        System.out.println(vl.getBusBreakerView().getBus("VL_2").getV());
        vl.getNodeBreakerView().getSwitch("BR1").setOpen(true);
        System.out.println(vl.getBusBreakerView().getBus("VL_1").setV(456).getV());
        System.out.println(vl.getBusBreakerView().getBus("VL_2").getV());

123.0
123.0
456.0
456.0

When we setV(456), the switch is open so VL_2 should not be set to 456

This is because there is no equipment associated with the bus, so to get a terminal to store the voltage/angle we findEquivalentTerminal but findEquivalementTerminal doesn't stop at retained switch so the equivalent terminal of vl2 is the one from the busbarsection of vl1 (because the switch is closed when the findEquivalementTerminal happens). Then we open the switch but because it is retained we don't invalidate the busbreakerbuses (normal because toggling retained switch should not affect busbreakerview buses because they should stop at retained switch anyway !) and so we keep the old equivalent terminal but now it's no longer equivalent 

should print
123.0
123.0
456.0
123.0

### Describe the steps

_No response_

### Environment

_No response_

### Relevant Log Output

_No response_

### Extra Information

_No response_

samirromdhani:

Fixes #3200

- [ ] Yes
- [x] No

**Other information**:

This pull try to resolve #3200, and do not invalidate the BusBreakerView buses cache when possible.

As stated and implemented before (like here PR 2439)
> If switch is retained, bus breaker view cache not invalidated when switch opened/closed
> 
> when a retained switch is opened/closed, we do not invalidate the BusBreakerView buses cache, since retained switch toggling should not affect BusBreakerView topology.

To resolve this issue (3200), recompute equivalent terminal in getters (getV, getAngle, etc.) only if the version of the equivalentTerminal has changed: we increment equivalentTerminal version when a retained switch is opened/closed
In `CalculatedBusImpl`, `terminalRef` is recomputed only if that version changed, else the cached value is reused.


rolnico (commented):
I have another proposition of implementation on this branch: https://github.com/powsybl/powsybl-core/tree/nro/issue3200

Tell me what you uthink of it!
It might be interesting to create more complex cases to check the validiity of the solutions, and to do some benchmark to see the impact on the global performances of the lib in both cases.

samirromdhani:
> I have another proposition of implementation on this branch: https://github.com/powsybl/powsybl-core/tree/nro/issue3200
> 
> Tell me what you uthink of it! It might be interesting to create more complex cases to check the validiity of the solutions, and to do some benchmark to see the impact on the global performances of the lib in both cases.

Thanks for the proposition,
I think some cases does not work as expected in the proposition, like: `closed -> opened -> then closed`. I added a test for this case.

Also i think add new traverser to find equivalent terminal, gives kind of duplication with `Networks.getEquivalentTerminal` that become used only in test.

rolnico (commented):
I updated my proposal to fix the issue you raised correctly. What do you think of it?

samirromdhani:
> I updated my proposal to fix the issue you raised correctly. What do you think of it?

Thanks ! the proposed version passes the tests. I think both solutions solve the issue. 
One thing I note that in the proposed fix `checkValidityAndUpdateTerminalRefOnSwitchModification()` (previously `checkValidity()`) was called from almost every method of CalculatedBusImpl, even that they have nothing to do with terminalRef. May be there is extra checks on unrelated operations.
Otherwise, what do you think about the PR fix ?

PR #3820, opened 2026-03-12
Add private extensions validation (loadflow-validation)

### LINKED ISSUE #696: Loadflow validation : being able to validate extensions behaviour

Feature.

The loadflow validation tool does not know about extensions and cannot validate the behaviour of computations with respect with those extensions.

It would be nice to be able to plug in the validation of "private" extensions into the validation tool. For example, if a loadflow handles a specific type of HVDC regulation, we could check that the result conforms to that regulation.

Being able to fully validating results on networks which uses specific extensions for computations (HVDC regulations ...).

samirromdhani:

Fixes #696

- [ ] Yes
- [x] No

**Other information**:


loadflow/loadflow-validation/src/main/java/com/powsybl/loadflow/validation/ValidationTool.java
olperr1: The problem with the unit test comes from these lines.
`Paths.get(...)` retrieves the file from the OS file system, but in the unit test we use a virtual FS.

See for instance how it is done in [`DynamicSimulationTool`](https://github.com/powsybl/powsybl-core/blob/be01387a2c5901fadef716d085ad42325df58802/dynamic-simulation/dynamic-simulation-tool/src/main/java/com/powsybl/dynamicsimulation/tool/DynamicSimulationTool.java#L125): the FS that is used is retrieved from the context.

samirromdhani: Thanks, that’s helpful! I see how using the FS from context's tool.

docs/user/itools/loadflow-validation.md
```diff
@@ -86,6 +87,9 @@ Use the `--load-flow` parameter to run a load-flow before the validation. This o
 `--output-format`<br>
 Use the `--output-format` parameter to specify the format of the output files. The available output formats are `CSV` or `CSV_MULTILINE`.
 
+`--with-extensions-validation`<br>
+Use the `--with-extensions-validation` parameter to perform extensions validation.
+
```
rolnico, line 92: Wrong place, you put it between an argument and the examples linked to that argument

rolnico (changes requested):
Seems good except for a fix needed in the documentation.
To be sure that everything works fine, it could be tested with an OLF extension for example?

docs/user/itools/loadflow-validation.md
samirromdhani: Done

loadflow/loadflow-validation/src/main/java/com/powsybl/loadflow/validation/extension/ExtensionsValidation.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public static List<String> getExtensionsNames() {
+        return getExtensions().stream().map(ExtensionValidation::getName).toList();
+    }
+
+    public static Optional<ExtensionValidation> getExtension(String name) {
+        return getExtensions().stream().filter(v -> v.getName().equals(name)).findFirst();
+    }
+
+    public void runExtensionValidations(Network network, ValidationConfig config, ToolRunningContext context) {
```
rolnico: Is there a reason this method is not static?

samirromdhani: There is no reason, i changed it to static.

samirromdhani: Good catch, this was initially implemented with a test first approach, the test is adjusted as well.

rolnico (approved):

PR #3822, opened 2026-03-16
Fix name serialization problem in XML with special characters

NathanDissoubray:

Fix a bug where names containing special characters (in particular, a comma) would be incorrectly serialized

The commas in operational limits groups are exported as is. Since names are separated by commas, that means when it is read again, we incorrectly read group names and number.

Special characters are properly quoted.

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/cgmes/cgmes-extensions/src/main/java/com/powsybl/cgmes/extensions/BaseVoltageMappingSerDe.java
+++ b/cgmes/cgmes-extensions/src/main/java/com/powsybl/cgmes/extensions/BaseVoltageMappingSerDe.java
@@ -13,8 +13,12 @@ import com.powsybl.commons.extensions.AbstractExtensionSerDe;
 import com.powsybl.commons.extensions.ExtensionSerDe;
 import com.powsybl.commons.io.DeserializerContext;
 import com.powsybl.commons.io.SerializerContext;
+import com.powsybl.commons.io.TreeDataReader;
 import com.powsybl.commons.io.TreeDataWriter;
 import com.powsybl.iidm.network.Network;
+import com.powsybl.iidm.serde.IidmVersion;
+import com.powsybl.iidm.serde.NetworkDeserializerContext;
+import com.powsybl.iidm.serde.NetworkSerializerContext;
 
 import java.util.LinkedHashMap;
 import java.util.Map;
@@ -41,7 +45,8 @@ public class BaseVoltageMappingSerDe extends AbstractExtensionSerDe<Network, Bas
 
     @Override
     public void write(BaseVoltageMapping extension, SerializerContext context) {
-        TreeDataWriter writer = context.getWriter();
+        NetworkSerializerContext networkContext = (NetworkSerializerContext) context;
+        TreeDataWriter writer = networkContext.getWriter();
         Map<Double, BaseVoltageMapping.BaseVoltageSource> sortedBaseVoltages = extension.getBaseVoltages().entrySet().stream()
                 .sorted(Map.Entry.comparingByKey())
                 .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (oldValue, newValue) -> oldValue, LinkedHashMap::new));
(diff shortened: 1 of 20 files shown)


commons/src/main/java/com/powsybl/commons/xml/XmlReader.java
```diff
@@ -149,7 +154,13 @@ private String[] readAndSplitStringArray(String name) {
         if (arrayString == null) {
             return new String[0];
         }
-        return arrayString.split(",");
+        try (CSVParser parser = CSVParser.parse(new StringReader(arrayString), CSVFormat.DEFAULT)) {
```
rolnico:
```suggestion
        try (CSVParser parser = CSVParser.parse(new StringReader(arrayString), XmlUtil.getCsvFormat())) {
```

commons/src/main/java/com/powsybl/commons/xml/XmlWriter.java
```diff
@@ -182,7 +187,15 @@ public void writeIntArrayAttribute(String name, Collection<Integer> values) {
...
-            writeStringAttribute(name, String.join(",", values));
+            CSVFormat format = CSVFormat.DEFAULT.builder().setRecordSeparator("").get();
+            StringWriter arrayWriter = new StringWriter();
+            try (CSVPrinter printer = new CSVPrinter(arrayWriter, format)) {
+                printer.printRecord(values);
+            } catch (IOException e) {
+                //Shouldn't happen, StringWriter doesn't throw
+                throw new UncheckedIOException(e);
+            }
+            writeStringAttribute(name, arrayWriter.toString());
```
rolnico:
```suggestion
            writeStringAttribute(name, XmlUtil.getCsvFormat().format(values.toArray()));
```

commons/src/main/java/com/powsybl/commons/xml/XmlReader.java
rolnico: For this to work, you need to add this in XmlUtil:
```java
    private static final Supplier<CSVFormat> CSV_FORMAT_SUPPLIER = Suppliers.memoize(XmlUtil::createCSVFormatInstance);

    public static CSVFormat getCsvFormat() {
        return CSV_FORMAT_SUPPLIER.get();
    }

    private static CSVFormat createCSVFormatInstance() {
        return CSVFormat.DEFAULT.builder().setRecordSeparator("").get();
    }
```

commons/src/main/java/com/powsybl/commons/xml/XmlReader.java
```diff
@@ -149,7 +154,13 @@ private String[] readAndSplitStringArray(String name) {
...
+            //get single csv value
```
rolnico:
```suggestion
            // Get a single CSV value
```

commons/src/main/java/com/powsybl/commons/xml/XmlReader.java
```diff
@@ -149,7 +154,13 @@ private String[] readAndSplitStringArray(String name) {
...
+            return parser.iterator().next().values();
+        } catch (IOException e) {
+            // should not happen as StringReader doesn't throw
```
rolnico:
```suggestion
            // Should not happen as StringReader doesn't throw
```

commons/src/main/java/com/powsybl/commons/xml/XmlReader.java
NathanDissoubray: Ah yes, it's better to have it there indeed. Also used CSVFormat.parse instead of CSVParser.parse to look more like the writer (CSVFormat.format)
Done in 1c49a285af85

commons/src/main/java/com/powsybl/commons/xml/XmlWriter.java
NathanDissoubray: It's much better thay way, thanks done in 1c49a285af85 (did not directly use the suggestion because import needed to be changed)

commons/src/main/java/com/powsybl/commons/xml/XmlReader.java
NathanDissoubray: I reworded it a bit to explain why there is a single value in 908c636c07bd

olperr1:
The code coverage cannot be increased: the uncovered lines are unreachable "catch" cases (the corresponding exceptions cannot be thrown).

PR #3824, opened 2026-03-17
Fix find available positions

rolnico:

No

Bug fix

When a voltage level has an internal line, the position of both sides of the line are added to the list of occupied positions. The position range is then wrong.

Only the correct side of the branch is considered, and the traverser does not continue on the other side.

- [ ] Yes
- [x] No

**Other information**:


iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/TopologyModificationUtils.java
```diff
@@ -573,52 +598,85 @@ private static List<Integer> getOrderPositions(ConnectablePosition<?> position,
...
+            Optional.ofNullable(getFeederBySide(position, ThreeSides.TWO)).ifPresent(feeders::add);
         }
         if (twt.getLeg3().getTerminal().getVoltageLevel() == voltageLevel) {
-            Optional.ofNullable(position.getFeeder3()).ifPresent(feeders::add);
+            Optional.ofNullable(getFeederBySide(position, ThreeSides.THREE)).ifPresent(feeders::add);
         }
         return feeders;
     }
 
+    private static ConnectablePosition.Feeder getFeederBySide(ConnectablePosition<?> position, ThreeSides side) {
```
colineplqt, line 669: I wonder if this method should not be somewhere else, we already have the same in other repositories because it is pretty useful. Maybe we could make it public and move it somewhere?

colineplqt: Talked about it, we will do that in another PR.

PR #3826, opened 2026-03-17
Skip writing default attribute values in IIDM serialization

clementleclercRTE:

- [X] A PR or issue has been opened in all impacted repositories (if any)

Performance feature: reduction of serialized file size by omitting attributes equal to their default value.

Every attribute is always written during serialization, even when its value equals the defined default. This produces unnecessarily verbose output, especially for tap changer steps (4 attributes
  r, x, g, b almost always zero), switches (retained/open typically false), and transformer/line shunt parameters (g, b typically zero). On large networks this has a measurable impact on file size
  and I/O performance, particularly for the binary serialization format.

 Attributes whose value equals their default are omitted from the serialized output. The read side is unaffected: all readers already use readDoubleAttribute(name,defaultValue) / readBooleanAttribute(name, defaultValue), which transparently return the default when an attribute is absent.
Versions ≤ 1.16 are unchanged — all attributes continue to be written unconditionally, preserving full backward compatibility.
 For the binary format, this PR requires PR #3804, which makes the binary reader order-independent and tolerant of absent attributes.

  Full list of attributes now omitted when equal to their default :      

Attribute | Type | Changed to
-- | -- | --
retained | Switch | optional default false
r, x, g, b | RatioTapChangerStep | optional default 0
r, x, g, b | PhaseTapChangerStep | optional default 0
lowTapPosition | RatioTapChanger, PhaseTapChanger | optional default 0
g1, g2, b1, b2 | Line | optional default 0
g, b | TwoWindingsTransformer | optional default 0
g1, g2, g3, b1, b2, b3 | ThreeWindingsTransformer | optional default 0
                                                                                                                                                                      
- [ ] Yes
- [X] No

**Other information**:

The internal method writeTapChangerStep in AbstractTransformerSerDe had its signature changed from (TapChangerStep, TreeDataWriter) to (TapChangerStep, NetworkSerializerContext) in order to access
   the IIDM version inside that method for the version guard. This is a package-private refactoring with no impact on the public API.


flo-dup:
The attribute default values should be in line with the adder (see #2417)

olperr1:
Could you update the PR to **exclude** the following attributes?
We discussed about this PR with some developers and we don't think their selected default values are really obvious.

Attribute | Type | Changed to
-- | -- | --
~open~ | ~Switch~ | ~optional default false~ 
~loadTapChangingCapabilities~ | ~RatioTapChanger only~ | ~optional default false~
~minP~ | ~Generator~ | ~optional default 0~
~voltageRegulatorOn~ | ~Generator~ | ~optional default false~

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTransformerSerDe.java
```diff
@@ -46,10 +46,18 @@ protected static void writeTapChangerStep(TapChangerStep<?> tcs, NetworkSerializ
 
     private static TreeDataWriter readCommonTapChangerAttributes(TapChangerStep<?> tcs, NetworkSerializerContext context) {
```
olperr1: Could you rename this method that **writes** attributes?

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractTransformerSerDe.java
```diff
@@ -324,10 +335,10 @@ protected static void readPhaseTapChanger(int leg, ThreeWindingsTransformer.Leg
     }
 
     private static double[] readCommonDoubleAttributesForAdder(NetworkDeserializerContext context) {
-        double r = context.getReader().readDoubleAttribute("r");
-        double x = context.getReader().readDoubleAttribute("x");
-        double g = context.getReader().readDoubleAttribute("g");
-        double b = context.getReader().readDoubleAttribute("b");
+        double r = context.getReader().readDoubleAttribute("r", 0.0);
```
flo-dup: For the whole PR, the attributes which are now optional remain mandatory for the previous version, so you should separate the read exactly like the write

flo-dup (dismissed):

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractSwitchSerDe.java
```diff
@@ -29,9 +29,9 @@ protected String getRootElementName() {
     @Override
     protected void writeRootElementAttributes(Switch s, VoltageLevel vl, NetworkSerializerContext context) {
         context.getWriter().writeEnumAttribute("kind", s.getKind());
-        context.getWriter().writeBooleanAttribute("retained", s.isRetained());
+        IidmSerDeUtil.runUntilMaximumVersion(IidmVersion.V_1_16, context, () -> context.getWriter().writeBooleanAttribute("retained", s.isRetained()));
+        IidmSerDeUtil.runFromMinimumVersion(IidmVersion.V_1_17, context, () -> context.getWriter().writeBooleanAttribute("retained", s.isRetained(), false));
```
NathanDissoubray, line 33: Technically not necessary for the PR, but could you add a bit of javadoc for the writeAttribute / readAttribute in TreeDataReader / TreeDataWriter ? There is none actually, and since you are using it here, I thought it would be a good occasion to add some.

NathanDissoubray (changes requested):
If there is no time to do it, you can dismiss this.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ThreeWindingsTransformerSerDe.java
```diff
@@ -39,24 +41,52 @@ protected String getRootElementName() {
...
+                    "b2", IidmSerDeUtil.ErrorMessage.NOT_DEFAULT_NOT_SUPPORTED, IidmVersion.V_1_1, context);
+        });
+        IidmSerDeUtil.runInBetweenTwoVersions(IidmVersion.V_1_1, IidmVersion.V_1_16, context, () -> {
+            context.getWriter().writeDoubleAttribute("g2", g2);
+            context.getWriter().writeDoubleAttribute("b2", b2);
+        });
+        IidmSerDeUtil.runFromMinimumVersion(IidmVersion.V_1_17, context, () -> {
+            context.getWriter().writeDoubleAttribute("g2", g2, 0);
+            context.getWriter().writeDoubleAttribute("b2", b2, 0);
+        });
```
NathanDissoubray: could you add something like writeFormerlyMandatoryDoubleAttribute but in between two versions ? It's very similar, but it's something like writeMandatoryBetweenVersionsDoubleAttribute(minVersion, optionalSince)
And it could be used in a few places
A similar read would be good too

NathanDissoubray (changes requested):

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ThreeWindingsTransformerSerDe.java
flo-dup: It's only for the three-winding transformers, or is there some other equipments?

I don't think I may write such a method for the reading part. How did you think writing it?

flo-dup:
I cannot approve this PR, as I did the changes myself, but current state looks good to me. The changes I did were:
- extending the "optionality" of g, b for three-winding transformer & boundary lines
- renaming and moving the writeDoubleAttribute method from ConnectableSerDeUtil to IidmSerDeUtil

NathanDissoubray:
~~Looking at the implementations of the `TreeDataReader` / `TreeDataWriter` with the default value in `BinReader` / `BinWriter`, they do not take into account the default value at all, and always read / write the values. Is this on purpose, or does it need modification ? It works correctly for json and xml.~~

Edit: I was on the wrong branch, disregard this

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ThreeWindingsTransformerSerDe.java
NathanDissoubray:
```java
public static void writeMandatoryBetweenVersionsDoubleAttribute(String name, double value, IidmVersion minVersion, IidmVersion optionalSince, NetworkSerializerContext context) {
  if (context.getVersion().compareTo(minVersion) >= 0 {
      if (context.getVersion().compareTo(optionalSince) < 0) {
          context.getWriter().writeDoubleAttribute(name, value);
      } else {
          context.getWriter().writeDoubleAttribute(name, value, 0.0);
      }
  }
}
```

flo-dup: Yes, that's for the writing part, there's no problem there. I was speaking about the reading one

NathanDissoubray (approved):

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ThreeWindingsTransformerSerDe.java
NathanDissoubray: It's more difficult on the reading part I think

olperr1 (approved):

PR #3833, opened 2026-03-19
CGMES: cim:GeographicalRegion.name DE-* converted to Country.DE

jeandemanged:

No

Feature

cim:GeographicalRegion.name with values `D1`, `D2`, `D4`, `D7`, `D8` map to `Country.DE` for IIDM substations.

In addition, cim:GeographicalRegion.name starting with `DE-` map to `Country.DE` for IIDM substations.

- [x] No

**Other information**:


rcourtier:
This will work, however don't you think we should retrieve the country from the GeographicalRegion.shortName instead of its name?

jeandemanged:
> This will work, however don't you think we should retrieve the country from the GeographicalRegion.shortName instead of its name?

This would work for Germany, however maybe this would break others...  CGMES does not have any country concept. Longer term probably we would need some configuration for this instead of current hardcodes. Also GeographicalRegion of DC IGMs are going to be an issue.

PR #3754, opened 2026-02-03
Optimize getFictitiousP0() and getFictitiousQ0() to compute over every nodes only when needed

SylvestreSakti:
**Please check if the PR fulfills these requirements**

- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] A PR or issue has been opened in all impacted repositories (if any)

**Does this PR already have an issue describing the problem?**
No

**What kind of change does this PR introduce?**
Performance issue is observed when using `getFictitiousP0()` and `getFictitiousQ0()` even when there are no fictitious injection. This PR optimizes this behavior.

**What is the current behavior?**
`getFictitiousP0()` and `getFictitiousQ0()`  always search all nodes of the bus to compute fictitious injection.

**What is the new behavior (if this is a feature change)?**
The search is done only if there is at least one non-zero fictitious injection in the voltage level.

**Does this PR introduce a breaking change or deprecate an API?**
- [ ] Yes
- [x] No

PR #3834, opened 2026-03-19
Fix and optimize fictitious P0 and Q0 values management

rolnico:

#3766 and #3754

Bugfix + feature

See refered PR

Save and reuse the list of nodes of the `CalculatedBusImpl` + add hasFictitiousP0 and hasFictitiousQ0 methods to avoid looking at the nodes values if not necessary + refactor ficititious values management regarding variants to follow usual behavior.

- [ ] Yes
- [x] No

**Other information**:


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java
```diff
@@ -984,21 +937,48 @@ public void traverse(int node, TopologyTraverser t) {
...
+            return fictitiousValueByNodes;
+        }
+
+        private boolean hasFictitiousInjection(ArrayList<TIntDoubleMap> fictitiousValueByNodes) {
+            return fictitiousValueByNodes.get(getNetwork().getVariantIndex()) != null
+                && !fictitiousValueByNodes.get(getNetwork().getVariantIndex()).isEmpty();
+        }
+
+        private double getFictitiousInjection(ArrayList<TIntDoubleMap> fictitiousValueByNodes, int node) {
+            TIntDoubleMap fictitiousP0Variant = fictitiousValueByNodes.get(getNetwork().getVariantIndex());
```
olperr1: `fictitiousP0Variant` -> `fictitiousValueVariant`

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java
```diff
@@ -984,21 +937,48 @@ public void traverse(int node, TopologyTraverser t) {
...
+            // Return the value corresponding to the node, or 0.0 if the node is not in the keyset (default in TIntDoubleHashMap)
+            return fictitiousP0Variant.get(node);
+        }
+
+        private void setFictitiousInjection(ArrayList<TIntDoubleMap> fictitiousValueByNodes, int node, double value,
+                                            String modifiedVariable) {
+            int variantIndex = getNetwork().getVariantIndex();
+
+            // Get the map corresponding to the current variant from the array
+            TIntDoubleMap fictitiousP0Variant = fictitiousValueByNodes.get(variantIndex);
```
olperr1: Same here: `fictitiousP0Variant` -> `fictitiousValueVariant`

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NodeBreakerTopologyModel.java
```diff
@@ -984,21 +937,48 @@ public void traverse(int node, TopologyTraverser t) {
...
+                                            String modifiedVariable) {
+            int variantIndex = getNetwork().getVariantIndex();
+
+            // Get the map corresponding to the current variant from the array
+            TIntDoubleMap fictitiousValueVariant = fictitiousValueByNodes.get(variantIndex);
+
+            // If it does not exist yet, create a new one
+            if (fictitiousValueVariant == null) {
+                fictitiousValueVariant = new TIntDoubleHashMap();
+                fictitiousValueByNodes.set(variantIndex, fictitiousValueVariant);
```
SylvestreSakti, line 975: More a question than a remark : should we use `synchronized` keyword to avoid thread-unsafe issues ? It was used  in the previous code.

rolnico: No because threads using different variants will operate on different array elements, so there will be no concurrency. It's the same behaviour for all network element internal variables (see GeneratorImpl for example)

rolnico: The previous code code was a bit inverted, which made it necessary to synchronize the treatments, but now it's back to usual behavior

PR #3847, opened 2026-03-25
Deprecate `LimitViolation` constructors and use `LimitViolationBuilder` instead

NathanDissoubray:

Deprecate and replace some code.

There are a lot of constructors in `LimitViolation`, `LimitViolationBuilder` is not really used.

Deprecate most constructors of `LimitViolation` and add a call to create a `LimitViolationBuilder` directly from `LimitViolation`

- [x] Yes
- [ ] No

If you are using the following constructors:
- `LimitViolation(String subjectId, @Nullable String subjectName, String operationalLimitsGroupId, LimitViolationType limitType, @Nullable String limitName, int acceptableDuration, double limit, double limitReduction, double value, @Nullable ThreeSides side, @Nullable ViolationLocation voltageLocation)`
- `LimitViolation(String subjectId, LimitViolationType limitType, double limit, double limitReduction, double value, ViolationLocation voltageLocation)`

Then no change is required.

Otherwise you can:
- use `LimitViolationBuilder`, either with `new LimitViolationBuilder()` or with `LimitViolation.builder()`, add the values and then use `.build()`
- use the constructor of `LimitViolation` with all the variables (mentionned above), and provide null values for all the variables you do not have a value for

Note that if you were using a constructor and were providing null values to it, you do not need to do that with the Builder. Also note that the default `limitReduction` is 1 and the default `acceptableDuration` is `Integer.MAX_VALUE`.

**Other information**:

### DIFF
--- a/contingency/contingency-api/src/main/java/com/powsybl/contingency/violations/LimitViolation.java
+++ b/contingency/contingency-api/src/main/java/com/powsybl/contingency/violations/LimitViolation.java
@@ -77,10 +77,6 @@ public class LimitViolation extends AbstractExtendable<LimitViolation> {
         this.voltageLocation = voltageLocation;
     }
 
-    public static LimitViolationBuilder builder() {
-        return new LimitViolationBuilder();
-    }
-
     /**
      * Create a new LimitViolation.
      *
@@ -278,6 +274,10 @@ public class LimitViolation extends AbstractExtendable<LimitViolation> {
     @Deprecated(since = "7.3.0", forRemoval = true)
     public LimitViolation(String subjectId, LimitViolationType limitType, double limit, double limitReduction, double value) {
         this(subjectId, null, limitType, null, Integer.MAX_VALUE, limit, limitReduction, value, null, null);
+    }
+
+    public static LimitViolationBuilder builder() {
+        return new LimitViolationBuilder();
     }
 
     /**
--- a/contingency/contingency-api/src/main/java/com/powsybl/contingency/violations/LimitViolation.java
+++ b/contingency/contingency-api/src/main/java/com/powsybl/contingency/violations/LimitViolation.java
@@ -76,7 +76,12 @@ public class LimitViolation extends AbstractExtendable<LimitViolation> {
         this.voltageLocation = voltageLocation;
     }
 
+    public static LimitViolationBuilder builder() {
+        return new LimitViolationBuilder();
+    }
+
     /**
+     * <p>@deprecated use {@link LimitViolationBuilder} or {@link LimitViolation#LimitViolation(String, String, String, LimitViolationType, String, int, double, double, double, ThreeSides, ViolationLocation)}</p>
      * Create a new LimitViolation.
      *
      * <p>According to the violation type, all parameters may not be mandatory. See constructor overloads for particular types.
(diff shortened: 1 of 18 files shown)


contingency/contingency-api/src/main/java/com/powsybl/contingency/violations/LimitViolation.java
```diff
@@ -76,7 +77,12 @@ public LimitViolation(String subjectId, @Nullable String subjectName, String ope
         this.voltageLocation = voltageLocation;
     }
 
+    public static LimitViolationBuilder builder() {
+        return new LimitViolationBuilder();
+    }
+
     /**
+     * <p>@deprecated use {@link LimitViolationBuilder} or {@link LimitViolation#LimitViolation(String, String, String, LimitViolationType, String, int, double, double, double, ThreeSides, ViolationLocation)}</p>
```
rolnico: The @deprecated part should be moved to the end of each Javadoc comment, with the other @ parts. Also, there is no need to add the `<p>` here

rolnico (changes requested):

contingency/contingency-api/src/main/java/com/powsybl/contingency/violations/LimitViolation.java
NathanDissoubray: Done in d43dc022d2dc

contingency/contingency-api/src/main/java/com/powsybl/contingency/violations/LimitViolation.java
```diff
@@ -76,6 +77,10 @@ public LimitViolation(String subjectId, @Nullable String subjectName, String ope
         this.voltageLocation = voltageLocation;
     }
 
+    public static LimitViolationBuilder builder() {
+        return new LimitViolationBuilder();
+    }
```
rolnico: Move this method right after the constructors

contingency/contingency-api/src/main/java/com/powsybl/contingency/violations/LimitViolation.java
```diff
@@ -206,7 +223,7 @@ public LimitViolation(String subjectId, LimitViolationType limitType, String lim
      * @param voltageLocation    Detailed information about the location of the violation.
      */
     public LimitViolation(String subjectId, LimitViolationType limitType, double limit, double limitReduction, double value, ViolationLocation voltageLocation) {
```
rolnico, line 221: Why is this constructor not deprecated?

rolnico (changes requested):

contingency/contingency-api/src/main/java/com/powsybl/contingency/violations/LimitViolation.java
NathanDissoubray: We discussed this with @olperr1 and if I understood correctly, it's because the constructor using `ViolationLocation` is simple and that would force people to use the builder.
I think it should be deprecated though.

contingency/contingency-api/src/main/java/com/powsybl/contingency/violations/LimitViolation.java
NathanDissoubray: done in 1489e074a8e9

rolnico (approved):

PR #3856, opened 2026-03-31
Add LimitReduction with a value above 1

### LINKED ISSUE #3706: Limit reduction API: be able to increase or ignore a limit

#2751 describes the `LimitReduction` API, for which the goal is to apply reduction on operational limits value in order to raise limit violation even under the physical authorized limit defined in the nework for an element.

The limit value is reduced as a percentage of the initial limit, so the limit is defined as a `double` in the interval [0; 1]

It is possible to apply an increase to a limit value or to avoid raising violation for a limit.

### Describe the motivation

Open to discussion to handle different needs regarding the applicability of limits:
- As proposed in the original issue here #2751 some operators want to ignore a limit.
- Another need is to be able to increase the limit, like having a reduction value strictly superior to 1.

### Extra Information

- to ignore a limit: a convention could be to handle the 0 value as a way to ignore the limit. Another way could be to have an attribute (boolean?) in the `LimitReduction` determining whether to apply a reduction with the given value or to ignore the limit (naming proposal: `shouldIgnoreLimit`).
- to increase a limit: we should allow to pass a value superior to 1.
- In `DefaultLimitsReducer::reduce()` (that calls `applyReduction`): we should take into account the new developement as it can be used to apply the reduction
  - if we decide 0 is a convention to not raise a violation in any case then it should be taken into account in this method
  - check also for possible other usages elsewhere
  - check also for impacts when increasing a limit, maybe in the implementations uising the limit reduction (SA implementations such as open-loadflow repository)

NathanDissoubray:

Closes #3706 

Add ability to use a limit reduction with a value above 1.

Currently, a limit reduction is with a value between 0 and 1.

The value can now be above 1.

- [ ] Yes
- [x] No

The LimitReduction JSON format has been bump to version **1.2**. This is not a breaking change _per se_, but you should be aware of this if your project communicates with another one based on an older version of PowSyBl:
- LimitReduction JSON files generated by an older version of PowSyBl could still be read;
- but LimitReduction JSON files generated by this version of PowSyBl won't be readable by older versions.

**Other information**:

This will need a renaming PR, as reduction is not really suited when the value can increase the limit instead. This PR only introduces the functionnality changes, the renaming is left in another PR in order to make this PR smaller and more manageable.
The RTD has not been modified as no functionnality change needs to be reported in the doc. The changes are purely naming related, and will thus be done with the renaming PR


security-analysis/security-analysis-api/src/main/java/com/powsybl/security/limitreduction/LimitReduction.java
```diff
@@ -231,8 +231,8 @@ private LimitReduction(LimitType limitType, double value, boolean monitoringOnly
         } else {
             throw new PowsyblException(limitType + " is not a supported limit type for limit reduction");
         }
-        if (value > 1. || value < 0.) {
-            throw new PowsyblException("Limit reduction value should be in [0;1]");
+        if (value < 0.) {
+            throw new PowsyblException("Limit reduction value should be equal or more than 0");
```
olperr1:
```suggestion
            throw new PowsyblException("Limit reduction value should be equal or greater than 0");
```

security-analysis/security-analysis-api/src/test/java/com/powsybl/security/limitreduction/LimitReductionListTest.java
```diff
@@ -126,11 +126,8 @@ void unsupportedLimitType() {
 
     @Test
     void unsupportedLimitReductionValues() {
-        String expectedMessage = "Limit reduction value should be in [0;1]";
+        String expectedMessage = "Limit reduction value should be equal or more than 0";
```
olperr1:
```suggestion
        String expectedMessage = "Limit reduction value should be equal or greater than 0";
```

security-analysis/security-analysis-api/src/test/java/com/powsybl/security/LimitViolationDetectionTest.java
```diff
@@ -125,6 +125,54 @@ void testVoltageViolationDetection() {
...
+        LimitViolation violation = violationsCollector.getFirst();
+        // the limit at 1200 is raised to 1320
+        assertEquals(LimitViolationUtils.PERMANENT_LIMIT_NAME, violation.getLimitName());
+        assertEquals(600, violation.getAcceptableDuration());
+        assertEquals(1100, violation.getLimit());
+        assertEquals(1, violation.getLimitReduction());
+    }
+
+    @Test
+    void testTempLimitInterveredWithIncrease() {
```
olperr1: Did you mean "Inverted", here?

security-analysis/security-analysis-api/src/test/java/com/powsybl/security/LimitViolationDetectionTest.java
```diff
@@ -125,6 +125,54 @@ void testVoltageViolationDetection() {
...
+            line, TwoSides.ONE, 1300., LimitType.CURRENT, Set.of(LoadingLimitType.TATL, LoadingLimitType.PATL), applier, violationsCollector::add
+        );
+        assertEquals(1, violationsCollector.size());
+        LimitViolation violation = violationsCollector.getFirst();
+        // the limit at 1200 is raised to 1560 above the limit at 1500
+        assertEquals(LimitViolationUtils.PERMANENT_LIMIT_NAME, violation.getLimitName());
+        assertEquals(60, violation.getAcceptableDuration());
+        assertEquals(1100, violation.getLimit());
+        assertEquals(1, violation.getLimitReduction());
+    }
```
olperr1, line 187: I think there's a problem in the detection.

I added the following section in the test:
```java
        LimitViolationDetection.checkLimitViolation(
                line, TwoSides.ONE, 1600., LimitType.CURRENT, Set.of(LoadingLimitType.TATL, LoadingLimitType.PATL), applier, violationsCollector::add
        );
        assertEquals(1, violationsCollector.size());
        violation = violationsCollector.getFirst();
        assertEquals("1'", violation.getLimitName());
        assertEquals(0, violation.getAcceptableDuration());
        assertEquals(1500, violation.getLimit());
        assertEquals(1, violation.getLimitReduction());
```

The detection should return a single violation corresponding to the most critical temporary limit that is exceeded. At 1600.A, both the "10'" and the "1'" are exceeded, so the returned violation should be for the "1'" limit.
But this test returns 2 violations, one for each temporary limit.

olperr1 (changes requested):

security-analysis/security-analysis-api/src/test/java/com/powsybl/security/LimitViolationDetectionTest.java
NathanDissoubray: Indeed, since the IT10 is raised from 1200 to 1560 (factor 1.3) it goes above the IT1 at 1500. So detecting at 1600 should ignore the IT10 and consider only the IT1 for the limit value. I'll investigate this.

NathanDissoubray: Ah I think I know why you have two detections. `violationsCollector` is only cleared before the test. If you make two detections in the same test, then you need to clear the collector between the detections, or it will keep the previous detection.
Adding `violationsCollector.clear();` before the check at 1600 solves the issue.
I'll still add the code you gave above as it's a good thing to test being above a raised limit too.

Added in f4e08cc05142

security-analysis/security-analysis-api/src/test/java/com/powsybl/security/LimitViolationDetectionTest.java
NathanDissoubray: Indeed, done in f4e08cc05142

security-analysis/security-analysis-api/src/test/java/com/powsybl/security/LimitViolationDetectionTest.java
olperr1: Oh sorry, you are right. I was not careful enough!

SylvestreSakti:
Hello, for your information, this PR has been tested in open-loadflow (https://github.com/powsybl/powsybl-open-loadflow/pull/1393) and gives the expected behaviour :+1:

olperr1 (approved):

PR #3455, opened 2025-05-27
Add debugDir parameter to dynamic-simulation, dynamic-security-analysis, shortcircuit and ampl-executor

etiennehomer:
**Please check if the PR fulfills these requirements**

- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)
- [x] A PR or issue has been opened in all impacted repositories (if any)

**Does this PR already have an issue describing the problem?**
No

**What kind of change does this PR introduce?**
Feature

**What is the current behavior?**
We can enable debug mode only with itools config.

**What is the new behavior (if this is a feature change)?**
We give a debugDir parameter to the run() method to indicate the path where the computation should dump execution files.
If debugDir is null, nothing is dumped. 

**Does this PR introduce a breaking change or deprecate an API?**
- [ ] Yes
- [x] No

Linked PR:
https://github.com/powsybl/powsybl-optimizer/pull/115
https://github.com/powsybl/powsybl-dynawo/pull/458

PR #3862, opened 2026-04-02
Add debugDir parameter in Loadflow, Security and Sensitivity Analysis

Lisrte:

No

Feature

`debugDir` parameter is only for dynamic-simulation, dynamic-security-analysis, shortcircuit and ampl-executor

`debugDir` parameter available in `LoadFlowParameters`, `SecurityAnalysisParameters` and `SensitivityAnalysisParameters`.

- [x] Yes
- [ ] No

A new version of **load flow parameters** files, the 1.11, was introduced. Differences with the previous version are:
- a new `debugDir` element was added.

A new version of **security analysis parameters** files, the 1.3, was introduced. Differences with the previous version are:
- a new `debug-dir` element was added.

A new version of **sensitivity analysis parameters** files, the 1.3, was introduced. Differences with the previous version are:
- a new `debug-dir` element was added.

These are not breaking changes _per se_, but you should be aware of this if your project communicates with another one based on an older version of PowSyBl:
- JSON files for these elements generated by an older version of PowSyBl could still be read;
- but JSON files for these elements generated by this version of PowSyBl won't be readable by older versions.

**Other information**:

Same as PR #3455 for LF, SA and Sensi
:warning: `SecurityAnalysisParameters` does not handle `ModuleConfig` properly (see the issue #3947). Thus, the new `debugDir` parameter cannot be loaded from `PlatformConfig` until this issue is fixed.


loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowParameters.java
```diff
@@ -454,8 +472,11 @@ public Map<String, Object> toMap() {
                 .put("countriesToBalance", countriesToBalance)
                 .put("computedComponentMode", componentMode)
                 .put("hvdcAcEmulation", hvdcAcEmulation)
-                .put("dcPowerFactor", dcPowerFactor)
-                .build();
+                .put("dcPowerFactor", dcPowerFactor);
+        if (debugDir != null) {
```
olperr1: What is the purpose of this test?
For instance, in `DynamicSimulationParameters`, `debugDir` is always added.

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisParameters.java
```diff
@@ -173,6 +176,7 @@ public static SecurityAnalysisParameters load(PlatformConfig platformConfig) {
                 .ifPresent(config -> {
                     parameters.setIntermediateResultsInOperatorStrategy(config.getBooleanProperty("intermediate-results-in-operator-strategy", DEFAULT_INTERMEDIATE_RESULTS_IN_OPERATOR_STRATEGY));
                     parameters.getIncreasedViolationsParameters().load(config);
+                    parameters.setDebugDir(config.getStringProperty("debugDir", DEFAULT_DEBUG_DIR));
```
olperr1: To prevent problems if one day the default value is not null anymore:
```java
config.getOptionalStringProperty("debugDir").ifPresent(parameters::setDebugDir);
```
(and for consistency too)

olperr1 (changes requested):

loadflow/loadflow-api/src/main/java/com/powsybl/loadflow/LoadFlowParameters.java
Lisrte: The `ImmutableMap` doesn't accept null value. Moreover in `LoadFlowParametersSerializer` `debugDir` is written only if the value is not null. We have the same process here.

olperr1: Then this method is not consistent with the existing `DynamicSimulationParameters.toMap()`, and there's no valid reason for this:
- In `DynamicSimulationParametersSerializer`, `debugDir` is also only written when the value is not null
- The way this map is build in this method can be changed to allow null values (just like in `DynamicSimulationParameters.toMap()`).

jeandemanged:
Hello @Lisrte , add also this debugDir parameter to Sensitivity Analysis parameters ?
In Open Load Flow, we have this parameter for all three: LF, SA, Sensi in respective parameters extensions.
cc @SylvestreSakti

Lisrte:
> Hello @Lisrte , add also this debugDir parameter to Sensitivity Analysis parameters ? In Open Load Flow, we have this parameter for all three: LF, SA, Sensi in respective parameters extensions. cc @SylvestreSakti

Hello @jeandemanged I just added the `debugDir` parameter to `SensitivityAnalysisParameters` .
But just so you know, `SecurityAnalysisParameters` does not handle `ModuleConfig` properly (see the issue #3947). Thus, the new `debugDir` parameter cannot be loaded from `PlatformConfig` until this issue is fixed.

sensitivity-analysis-api/src/test/resources/SensitivityAnalysisParametersV1.3.json
```diff
@@ -0,0 +1,28 @@
+{
+  "version" : "1.3",
+  "load-flow-parameters" : {
+    "version" : "1.10",
```
olperr1:
```suggestion
    "version" : "1.11",
```

olperr1 (changes requested):

olperr1 (approved):

alicecaron (approved):

Issue #3827, opened 2026-03-17
Enhanced DC Circuit Breaker


SGI-Landry:
### Describe the current behavior

There is a DcSwitch class, with on/off states but no resistance.

### Describe the expected behavior

A representation of DC circuit breakers with switching capabilities and built-in resistance. It must be compatible with the following file formats
- IIDM (import / export).
- CGMES (import).
- DGS (import).

Related issues to be created in Open Load Flow and PyPowSyBl.

### Describe the motivation

Resistance of breakers are essential to analyze losses as well as security in AC/DC load flow analyses.

### Extra Information

Limitation: no further feature (e.g. no reactance).

Proposed implementation:
- add a resistance parameter in class DcSwitch
- import of ElmCoup and TypSwitch for DC circuits from DGS files.
- IIDM and CGMES: to be further specified.

SGI-Landry:
### Outcome of discussion with @jeandemanged and @geofjamg on 2026-03-20
A choice was needed regarding the way to add a resistance to the disconnecting capability. Considered possibilities
1. Add a resistance attribute to DcSwitch. This keeps the consistency with the AC side regarding the `kind` attribute to determine whether the switch represents a disconnector or a breaker. This requires however to change the behavior of the topology computation, because when the resistance is non-zero, both ends of a closed switch have different voltage values. They can no longer be on the same bus.
2. Create a specific DcCircuitBreaker component. The `kind` attribute would no longer be relevant. This would break the consistency with the AC switch.
3. It is possible in principle to model a DC breaker as a switch connected to a line to model its resistance. While this is mathematically equivalent, it is less faithful to the reality of the breaker. This would likely cause pain to analyze simulation results.

@jeandemanged pointed out that this activity would require a change in the IIDM format, thus prompting for a discussion during a TSC.

Solution 1 is preferred for now. The decision must be finalized with the TSC.

PR #3869, opened 2026-04-10
DGS import dc switch

SGI-Landry:

Part of #3827

DGS import of ElmCoup elements with both ends in  the DC network. Possible resistance is read from a related DcSwitch.

I used this opportunity to do a small refactoring. I replaced AbstractHvdcConverter::isDcLink by AbstractHvdcConverter::isDcObject to generalize the approach and altogether bypass processing by AC DGS importer for equipment part of DC subnetworks.

- [x] Yes
- [ ] No

**Other information**:

- Import of ElmCoup in the AC network seems untested!

### DIFF
--- a/powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/AbstractHvdcConverter.java
+++ b/powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/AbstractHvdcConverter.java
@@ -25,6 +25,8 @@ abstract class AbstractHvdcConverter extends AbstractConverter {
         super(importContext, network);
     }
 
+    abstract boolean isDcObject(DataObject obj);
+
     /**
      * Check if a given DGS terminal is in a DC subgrid.
      *
--- a/powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/AbstractHvdcConverter.java
+++ b/powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/AbstractHvdcConverter.java
@@ -36,14 +36,6 @@ abstract class AbstractHvdcConverter extends AbstractConverter {
     abstract boolean isDcNode(DataObject elmTerm);
 
     /**
-     * Check if a given DGS line is in a DC subgrid.
-     *
-     * @param elmLne Line to check.
-     * @return true iff the line is in a DC subgrid.
-     */
-    abstract boolean isDcLink(DataObject elmLne);
-
-    /**
      * Create the DC subgrids in the network.
      */
     abstract void create();
--- a/powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/AbstractHvdcConverter.java
+++ b/powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/AbstractHvdcConverter.java
@@ -25,6 +25,13 @@ abstract class AbstractHvdcConverter extends AbstractConverter {
         super(importContext, network);
     }
 
+    /**
+     * Check if a DGS DataObject belongs to the DC part of the network.
+     *
+     * @param obj DataObject from the DGS data model.
+     * @return true iff the object is part of the DC network, and can therefore be omitted by the
+     *         AC importer.
+     */
     abstract boolean isDcObject(DataObject obj);
 
     /**
--- a/powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/DetailedHvdcConverter.java
+++ b/powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/DetailedHvdcConverter.java
@@ -64,7 +64,7 @@ final class DetailedHvdcConverter extends AbstractHvdcConverter {
             Set<DataObject> dcElmLnes = new HashSet<>();
             List<DataObject> elmLines = elmNets.stream().flatMap(elmNet -> elmNet.search(".*.ElmLne").stream()).toList();
             for (DataObject elmLne : elmLines) {
-                DataObjectRef typLneRef = elmLne.getObjectAttributeValue("typ_id");
+                DataObjectRef typLneRef = elmLne.getObjectAttributeValue(TYP_ID);
                 DataObject typLne = typLneRef.resolve().orElseThrow(() -> new PowerFactoryException("Missing line type in TypLne for ElmLne " + elmLne.getId() + "."));
                 int lineSysType = typLne.findIntAttributeValue("systp").orElseThrow(() -> new PowerFactoryException("Missing systp for TypLne " + typLne.getId() + "."));
                 if (lineSysType == 1) {
@@ -92,6 +92,12 @@ final class DetailedHvdcConverter extends AbstractHvdcConverter {
             return new DcGridData(dcElmLnes, dcElmTerms, usedVscs, elmGndswt, dcElmCoup);
         }
 
+        /**
+         * Gather data of DC ElmCoup from the PowerFactory data model.
+         * @param elmNets network elements where to find the ElmCoup.
+         * @param dcElmTerms DC terminals.
+         * @return set of ElmCoup data objects that are connected to 2 DC terminals.
+         */
         private static @NonNull Set<DataObject> getDcSwitchObjs(List<DataObject> elmNets, Set<DataObject> dcElmTerms) {
             assert dcElmTerms != null;
             assert dcElmTerms.isEmpty() || ELMTERM.equals(dcElmTerms.iterator().next().getDataClassName());
@@ -265,9 +271,9 @@ final class DetailedHvdcConverter extends AbstractHvdcConverter {
      * or zero if absent or if no related TypSwitch.
      */
     private static double getSwitchResistance(DataObject elmCoup) {
-        Optional<DataObject> typSwitch = elmCoup.findObjectAttributeValue("typ_id")
+        Optional<DataObject> typSwitch = elmCoup.findObjectAttributeValue(TYP_ID)
                 .flatMap(DataObjectRef::resolve);
-        return (double) typSwitch.map(t -> t.findFloatAttributeValue("R_on").orElse(0.0f)).orElse(0.0f);
+        return typSwitch.map(t -> t.findFloatAttributeValue("R_on").orElse(0.0f)).orElse(0.0f);
     }
 
     /**
@@ -428,7 +434,7 @@ final class DetailedHvdcConverter extends AbstractHvdcConverter {
         double lineLength = line.getFloatAttributeValue("dline"); // km
         int numberOfParallelLines = line.findIntAttributeValue("nlnum").orElse(1);
 
-        DataObject lineType = line.getObjectAttributeValue("typ_id").resolve().orElseThrow(() -> new PowerFactoryException("Missing line type referenced by DC line " + line.getId() + "."));
+        DataObject lineType = line.getObjectAttributeValue(TYP_ID).resolve().orElseThrow(() -> new PowerFactoryException("Missing line type referenced by DC line " + line.getId() + "."));
 
         int lineSysType = lineType.findIntAttributeValue("systype").orElse(1);
         if (lineSysType != 1) {
--- a/powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/DetailedHvdcConverter.java
+++ b/powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/DetailedHvdcConverter.java
@@ -141,15 +141,16 @@ final class DetailedHvdcConverter extends AbstractHvdcConverter {
 
         }
 
+        int getCount() {
+            return dcElmLnes.size() + dcElmTerms.size() + acDcConverters.size() + elmGndswt.size() + elmCoup.size();
+        }
+
     } // private record DcGridData
 
     DetailedHvdcConverter(ImportContext importContext, Network network, List<DataObject> elmNets) {
         super(importContext, network);
         gridData = DcGridData.createGridData(elmNets);
-        final int nRead = gridData.acDcConverters.size() // @todo update with all elements
-                + gridData.dcElmLnes.size()
-                + gridData.dcElmTerms.size();
-        LOGGER.info("{} data objects read from the DGS file for detailed DC data .", nRead);
+        LOGGER.info("{} data objects read from the DGS file for detailed DC data .", gridData.getCount());
     }
 
     @Override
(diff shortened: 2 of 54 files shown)


SGI-Landry:
Hello @bperr 
Can you please review my code?

powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/AbstractHvdcConverter.java
```diff
@@ -26,20 +26,21 @@ abstract class AbstractHvdcConverter extends AbstractConverter {
     }
 
     /**
-     * Check if a given DGS terminal is in a DC subgrid.
+     * Check if a DGS DataObject belongs to the DC part of the network.
      *
-     * @param elmTerm Terminal to check.
-     * @return true iff the terminal is in a DC subgrid.
+     * @param obj DataObject from the DGS data model.
+     * @return true iff the object is part of the DC network, and can therefore be omitted by the
```
bperr, line 32: Typo: iff -> if

powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/AbstractHvdcConverter.java
```diff
@@ -26,20 +26,21 @@ abstract class AbstractHvdcConverter extends AbstractConverter {
...
+    abstract boolean isDcObject(DataObject obj);
 
     /**
-     * Check if a given DGS line is in a DC subgrid.
+     * Check if a given DGS terminal is in a DC subgrid.
      *
-     * @param elmLne Line to check.
-     * @return true iff the line is in a DC subgrid.
+     * @param elmTerm Terminal to check.
+     * @return true iff the terminal is in a DC subgrid.
```
bperr, line 41: Typo: iff -> if

powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/DetailedHvdcConverter.java
```diff
@@ -85,33 +85,81 @@ static DcGridData createGridData(List<DataObject> elmNets) {
...
+            return new DcGridData(dcElmLnes, dcElmTerms, usedVscs, elmGndswt, dcElmCoup);
+        }
+
+        /**
+         * Gather data of DC ElmCoup from the PowerFactory data model.
+         * @param dcElmTerms DC terminals.
+         * @return set of ElmCoup data objects that are connected to 2 DC terminals.
+         */
+        private static @NonNull Set<DataObject> getDcSwitchObjs(Set<DataObject> dcElmTerms) {
+            assert dcElmTerms.isEmpty() || ELMTERM.equals(dcElmTerms.iterator().next().getDataClassName());
```
bperr: Why asserting the set is empty ? I believe it should be the opposite !
Also, I think the second condition should be with a `&&` instead of `||`

powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/DetailedHvdcConverter.java
```diff
@@ -85,33 +85,81 @@ static DcGridData createGridData(List<DataObject> elmNets) {
...
+            List<DataObject> connectedCubic = dcElmTerms.stream().flatMap(e -> e.getChildrenByClass("StaCubic").stream()).toList();
+            // List of ElmCoup that are connected to at least one DC terminal, with repetitions
+            List<DataObject> connectedElmCoup = connectedCubic.stream()
+                    .flatMap(cubic -> cubic.findObjectAttributeValue(OBJ_ID).flatMap(DataObjectRef::resolve).stream()) // all equipment connected to at least one DC terminal
+                    .filter(obj -> "ElmCoup".equals(obj.getDataClassName())) // Keep only ElmCoup
+                    .toList();
+
+            Set<DataObject> dcElmCoup = new HashSet<>(); // ElmCoup which are connected to DC terminals at both ends
+            Set<DataObject> dcElmCoup1Term = new HashSet<>(); // ElmCoup where 1 DC terminal was already encountered
+            for (DataObject elmCoup : connectedElmCoup) {
```
bperr: Is it possible to have an ill-conditioned case with a elmCoup connected to more than 2 terminals ?

powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/DetailedHvdcConverter.java
rolnico: If the set is empty, the method will return an empty set as it loops on the elements of `dcElmTerms`. Therefore it is not an issue if it's empty.

rolnico: Why using specifically an assertion here and not a usual check+PowsyblException?

powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/AbstractHvdcConverter.java
SGI-Landry: In English iff is a shorthand for "if and only if", i.e. logical equivalence.

powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/DetailedHvdcConverter.java
SGI-Landry: This algorithm will detect any odd number of terminals. This is a good point, I will tackle this edge case.

powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/DetailedHvdcConverter.java
SGI-Landry: An empty set is ok. This test reads
"If there is any element in the set, then the first element should be of type "ElmTerm".

I will change and document this.

SGI-Landry: This is a private method and should only be called with validated data, so there is no point raising an exception.

I mixed up the DataObject arguments several times between ElmTerm, ElmCoup, etc. when developing the detailed DC DGS importer. The assertion is a safeguard for programming errors.

Is this acceptable in the PowSyBl coding conventions?

The test on dcElmTerms.isEmpty() avoids failing with dcElmTerms.iterator().next(). I intend to replace this with a test on all elements of dcElmTerms (if any), which should better convey the intent and cover the issue better as well.

SGI-Landry:
Hello @marqueslanauja 
Would you please review this PR on the update of DGS import to include resistance in DC switches?

marqueslanauja:
> Hello @marqueslanauja Would you please review this PR on the update of DGS import to include resistance in DC switches?

Yes, I’ll do it

powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/DetailedHvdcConverter.java
```diff
@@ -85,33 +86,78 @@ static DcGridData createGridData(List<DataObject> elmNets) {
...
+         * @param dcElmTerms DC terminals.
+         * @return set of ElmCoup data objects that are connected to 2 DC terminals.
+         */
+        private static @NonNull Set<DataObject> getDcSwitchObjs(Set<DataObject> dcElmTerms) {
+            assert dcElmTerms.stream().allMatch(e -> ELMTERM.equals(e.getDataClassName()));
+
+            // List of cubicles connected to DC terminals
+            List<DataObject> connectedCubic = dcElmTerms.stream().flatMap(e -> e.getChildrenByClass("StaCubic").stream()).toList();
+            // List of ElmCoup that are connected to at least one DC terminal, with repetitions
+            List<DataObject> connectedElmCoup = connectedCubic.stream()
```
marqueslanauja: Maybe we can avoid using the intermediate and temporary `connectedElmCoup` List

```
Map<DataObject, Long> terminalCount = connectedCubic.stream()
                    .flatMap(cubic -> cubic.findObjectAttributeValue(OBJ_ID).flatMap(DataObjectRef::resolve).stream()) // all equipment connected to at least one DC terminal
                    .filter(obj -> "ElmCoup".equals(obj.getDataClassName())) // Keep only ElmCoup
                    .collect(Collectors.groupingBy(Function.identity(), Collectors.counting()));
```

powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/DetailedHvdcConverter.java
```diff
@@ -85,33 +86,78 @@ static DcGridData createGridData(List<DataObject> elmNets) {
...
+                    .flatMap(cubic -> cubic.findObjectAttributeValue(OBJ_ID).flatMap(DataObjectRef::resolve).stream()) // all equipment connected to at least one DC terminal
+                    .filter(obj -> "ElmCoup".equals(obj.getDataClassName())) // Keep only ElmCoup
+                    .toList();
+
+            // Each ElmCoup appears once per DC terminal it is connected to, so counting occurrences gives
+            // the number of DC terminals connected to each switch.
+            Map<DataObject, Long> terminalCount = connectedElmCoup.stream()
+                    .collect(Collectors.groupingBy(Function.identity(), Collectors.counting()));
+
+            for (Map.Entry<DataObject, Long> entry : terminalCount.entrySet()) {
```
marqueslanauja: Maybe we can simplify it, my proposal is to check only whether the terminal count differs from 2.

```
terminalCount.forEach((elmCoup, count) -> {
                if (count != 2) {
                    throw new PowerFactoryException("ElmCoup " + elmCoup.getId() + " is connected to " + count + " DC terminals, expected exactly 2.");
                }
            });
```

powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/DetailedHvdcConverter.java
```diff
@@ -152,14 +198,67 @@ void create() {
...
+     * @param elmCoup data object with DC switch.
+     * @param network where to add the DcSwitch.
+     * @param objIdDcNodeRef map from objects to DC nodes.
+     */
+    private static void addDcSwitch(DataObject elmCoup, Network network, Map<Long, List<DcNodeRef>> objIdDcNodeRef) {
+        // Find the nodes
+        List<String> dcNodeIds = getAndCheckDcNodes(elmCoup, 2, objIdDcNodeRef);
+
+        // disconnector unless explicitly a breaker
+        String aUsage = elmCoup.findStringAttributeValue("aUsage").orElse("dct");
```
marqueslanauja: Maybe we could define a private method for obtaining the kind, but as you prefer

powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/DetailedHvdcConverter.java
```diff
@@ -152,14 +198,67 @@ void create() {
...
+                .add();
+    }
+
+    /**
+     * get resistance of DC switch if present in a related TypSwitch.
+     * @param elmCoup data object of the DC switch
+     * @return resistance of the switch if present in the related TypSwitch
+     * or zero if absent or if no related TypSwitch.
+     */
+    private static double getSwitchResistance(DataObject elmCoup) {
```
marqueslanauja: Maybe rename it to getDcSwitchResistance; it can also be simplified.

```
private static double getDcSwitchResistance(DataObject elmCoup) {
        return elmCoup.findObjectAttributeValue(TYP_ID)
                .flatMap(DataObjectRef::resolve)
                .flatMap(t -> t.findFloatAttributeValue("R_on"))
                .orElse(0.0f);
    }
```

marqueslanauja (commented):
Looks good. Only some minor simplifications suggested, but as you prefer

powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/DetailedHvdcConverter.java
SGI-Landry: Absolutely, there is no point in creating a list from a stream and then immediately restreaming it.

powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/DetailedHvdcConverter.java
SGI-Landry: Yes, simpler indeed.

powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/DetailedHvdcConverter.java
SGI-Landry: Good call, thanks.

powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/DetailedHvdcConverter.java
SGI-Landry: Much better indeed, thanks!

SGI-Landry:
Thanks @marqueslanauja. Your comments helped simplify my code.

@marqueslanauja @rolnico do you think this is ready to be merged?

powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/DetailedHvdcConverter.java
```diff
@@ -152,14 +189,80 @@ void create() {
...
+    }
+
+    /**
+     * Get the kind of DcSwitch from the string in elmCoup aUsage field. If aUsage is absent, default to
+     * disconnector.
+     * @param elmCoup ElmCoup object from the DGS file.
+     * @return DcSwitchKind associated with the ElmCoup DGS object.
+     * @throws PowerFactoryException in case of unsupported string in aUsage field.
+     */
+    private static DcSwitchKind getKind(DataObject elmCoup) {
```
marqueslanauja, line 236: You can return this expression directly instead of assigning it to the temporary variable kind. This will eliminate the issue reported in the Quality Gate.


```
private static DcSwitchKind getKind(DataObject elmCoup) {
        // disconnector unless explicitly a breaker
        String aUsage = elmCoup.findStringAttributeValue("aUsage").orElse("dct");
        return switch (aUsage) {
            case "dct" -> DcSwitchKind.DISCONNECTOR;
            case "cbk" -> DcSwitchKind.BREAKER;
            default -> throw new PowerFactoryException("Unsupported aUsage value '" + aUsage + "' for ElmCoup " + elmCoup.getId()
                                                        + ". Supported values: 'dct' (disconnector), 'cbk' (breaker).");
        };
    }
```

powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/DetailedHvdcConverter.java
```diff
@@ -52,18 +53,18 @@ private record DcGridData(Set<DataObject> dcElmLnes,
          * Selection is done based on the systp attribute of ElmTerm and of TypLne.
          */
         static DcGridData createGridData(List<DataObject> elmNets) {
-            assert elmNets.isEmpty() || ELMNET.equals(elmNets.getFirst().getDataClassName());
+            assert elmNets.stream().allMatch(e -> ELMNET.equals(e.getDataClassName()));
```
rolnico: As the assertions in the files `DetailedHvdcConverter` and `PowerFactoryImporter` are the only such assertions in powsybl-core production code, and since they were really useful only during developement, I would remove them.

rolnico (changes requested):

SGI-Landry:
@marqueslanauja @rolnico  I have taken your reviews into account.

marqueslanauja (approved):

rolnico (approved):
Thanks for the work!

PR #3871, opened 2026-04-13
Two-pass security analysis

RiadBRD:

feature

This PR introduces a new Two pass Security Analysis feature. It adds a new SecurityAnalysisProvider that orchestrates a two-pass workflow:

1. First Pass: Runs a fast analysis on all specified contingencies using a configurable primary provider (e.g., OpenLoadFlow).

2. Criteria Evaluation: Evaluates the results of the first pass to identify contingencies that did not converge (status is neither CONVERGED nor NO_IMPACT).

3. Second Pass: For any contingencies that meet the switch criteria, automatically triggers a more detailed analysis using a configurable secondary provider (e.g., DynaFlow).

4. Result Merging: Merges the results from both passes, with the second pass results taking precedence for any re-analyzed contingencies.

Providers are discovered via ServiceLoader and configured by name through TwoPassSecurityAnalysisParameters.

- [ ] Yes
- [x] No


security-analysis/mixed-security-analysis-api/src/main/java/com/powsybl/mixed/security/analysis/MixedSecurityAnalysis.java
```diff
@@ -0,0 +1,204 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+/**
+ * Business logic for mixed-mode security analysis: static pass on all contingencies,
+ * then dynamic pass on those that meet the switch criteria, results merged.
+ *
+ * @author Riad Benradi {@literal <riad.benradi at rte-france.com>}
+ */
+public class MixedSecurityAnalysis {
```
geofjamg, line 30: HybridSecurityAnalysis would be better

security-analysis/mixed-security-analysis-api/src/main/java/com/powsybl/mixed/security/analysis/MixedSecurityAnalysisProvider.java
```diff
@@ -0,0 +1,108 @@
...
+
+        return analysis.run();
+    }
+
+    @Override
+    public Optional<Extension<SecurityAnalysisParameters>> loadSpecificParameters(PlatformConfig config) {
+        MixedModeParametersExtension ext = new MixedModeParametersExtension();
+
+        // Default values — used if the key is absent from the config file
+        ext.setStaticSimulator("load-flow");
```
geofjamg, line 94: Maybe not a good idea to have a default value, as in powsybl core we don't know anything about OLF and DynaFlow

security-analysis/mixed-security-analysis-api/src/main/java/com/powsybl/mixed/security/analysis/MixedSecurityAnalysisProvider.java
```diff
@@ -0,0 +1,108 @@
...
+        ext.setDynamicSimulator("dynaflow");
+        ext.setSwitchCriteria(List.of("NOT_CONVERGED"));
+
+        // Override with values from config file if the section exists
+        // Reads from a YAML block named "mixed-mode-analysis"
+        config.getOptionalModuleConfig(MixedModeParametersExtension.NAME).ifPresent(module -> {
```
geofjamg, line 100: What are you doing here, you set to the extension what you read in the extension???

security-analysis/mixed-security-analysis-api/src/main/java/com/powsybl/mixed/security/analysis/MixedSecurityAnalysisProvider.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @Override
+    public Optional<Extension<SecurityAnalysisParameters>> loadSpecificParameters(PlatformConfig config) {
+        MixedModeParametersExtension ext = new MixedModeParametersExtension();
+
+        // Default values — used if the key is absent from the config file
+        ext.setStaticSimulator("load-flow");
+        ext.setDynamicSimulator("dynaflow");
+        ext.setSwitchCriteria(List.of("NOT_CONVERGED"));
```
geofjamg, line 96: Use a Constant from the criteria class

security-analysis/mixed-security-analysis-api/src/main/java/com/powsybl/mixed/security/analysis/MixedSecurityAnalysisProvider.java
```diff
@@ -0,0 +1,108 @@
...
+
+    /** @return {@code "MixedSecurityAnalysis"} */
+    @Override
+    public String getName() {
+        return PROVIDER_NAME;
+    }
+
+    /** @return the provider version */
+    @Override
+    public String getVersion() {
```
geofjamg, line 43: We would expect here either:
 - the core version ?
 - the version of found simulators ?

security-analysis/mixed-security-analysis-api/src/main/java/com/powsybl/mixed/security/analysis/criteria/AnalysisSwitchCriteria.java
```diff
@@ -0,0 +1,70 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.Objects;
+/**
+ * Evaluates whether a contingency result should trigger a switch to dynamic analysis.
+ *
+ * @author Riad Benradi {@literal <riad.benradi at rte-france.com>}
+ */
+
+public class AnalysisSwitchCriteria {
```
geofjamg, line 16: My advice here: before trying to create an abstraction on static / dynamic simulation contingency selection, try to code it simple and make it working with a simple processing: from static analysis, get contingencies from which an automaton has been triggered or a divergence and the re-run the dynamic one. It could be coded in a very simple way.

security-analysis/mixed-security-analysis-api/src/main/java/com/powsybl/mixed/security/analysis/parameters/MixedModeParametersExtension.java
```diff
@@ -0,0 +1,61 @@
...
+ * PowSybl will handle reading a configuration file (e.g., YAML) and populate
+ * the fields of this class automatically.
+ */
+public class MixedModeParametersExtension extends AbstractExtension<SecurityAnalysisParameters> {
+
+    public static final String NAME = "mixed-mode-analysis";
+    /**
+     * The name of the static simulator to use for the first pass.
+     */
+    private String staticSimulator;
```
geofjamg, line 21: The right term for a SecurityAnalysis is provider so here you need the name of the static security analysis provider and the name of the dynamic security analysis provider.

security-analysis/mixed-security-analysis-api/src/main/java/com/powsybl/mixed/security/analysis/parameters/MixedModeParametersExtension.java
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * The name of the dynamic simulator to use for the second pass (for complex cases).
+     */
+    private String dynamicSimulator;
```
geofjamg, line 26: Maybe also another question from myself: could dynamic term here be a little bit confusing as we also have a DynamicSecurityAnalysis API (impl for dynawaltz)? But here we a looking for the classical implementation of SecurityAnalysis but implemented by DynaFlow which is is also a dynamic simulator... this is very confusing.

security-analysis/mixed-security-analysis-api/pom.xml
```diff
@@ -0,0 +1,104 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.powsybl</groupId>
+            <artifactId>powsybl-security-analysis-api</artifactId>
+        </dependency>
+        <dependency>
```
geofjamg, line 31: You are creating a cycling dependency! We cannot have a dependency on OLF in powsybl core

security-analysis/security-analysis-api/src/main/java/com/powsybl/security/PostContingencyComputationStatus.java
```diff
@@ -15,5 +15,6 @@ public enum PostContingencyComputationStatus {
     MAX_ITERATION_REACHED,
     SOLVER_FAILED,
     FAILED,
-    NO_IMPACT
+    NO_IMPACT,
+    SPS_TRIGGERED
```
geofjamg, line 19: There is probably something to do like this, but for me it should be done in another PR before with everything regarding upgrading the security analysis result API to store automation systems triggering.

geofjamg (changes requested):

security-analysis/hybrid-security-analysis-api/src/main/java/com/powsybl/hybrid/security/analysis/parameters/HybridModeParametersExtension.java
```diff
@@ -0,0 +1,46 @@
+package com.powsybl.hybrid.security.analysis.parameters;
```
olperr1: This copyright header must be added to **all** the new files:
```java
/**
 * Copyright (c) 2026, RTE (http://www.rte-france.com)
 */
```

security-analysis/hybrid-security-analysis-api/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+    <artifactId>powsybl-hybrid-security-analysis-api</artifactId>
+    <name>Hybrid Security Analysis API</name>
+    <description>An API for running hybrid (static and dynamic) security analyses.</description>
+
+    <properties>
+        <maven.compiler.source>21</maven.compiler.source>
+        <maven.compiler.target>21</maven.compiler.target>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    </properties>
+
```
olperr1: These properties are not needed. The java version to use and the encoding are already defined in `powsybl-parent` (used by the project main `pom.xml`).
Can you remove them?

security-analysis/hybrid-security-analysis-api/src/main/java/com/powsybl/hybrid/security/analysis/HybridSecurityAnalysisProvider.java
```diff
@@ -0,0 +1,81 @@
...
+    /** @return {@code "HybridSecurityAnalysis"} */
+    @Override
+    public String getName() {
+        return PROVIDER_NAME;
+    }
+
+    /** @return the provider version */
+    @Override
+    public String getVersion() {
+        return getClass().getPackage().getImplementationVersion();
```
olperr1: With the following line, you would have more information (git version, build timestamp, ...), and this would be homogenous with the other providers (such as the OLF's one):
```java
return new PowsyblCoreVersion().toString();
```

security-analysis/hybrid-security-analysis-api/src/main/java/com/powsybl/hybrid/security/analysis/HybridSecurityAnalysisProvider.java
```diff
@@ -0,0 +1,81 @@
...
+        Objects.requireNonNull(workingVariantId);
+        Objects.requireNonNull(contingenciesProvider);
+        Objects.requireNonNull(runParameters);
+        LOGGER.info("Starting hybrid-mode security analysis for network: {}", network.getId());
+
+        HybridModeParametersExtension extension = runParameters.getSecurityAnalysisParameters().getExtension(HybridModeParametersExtension.class);
+        if (extension == null) {
+            LOGGER.warn("HybridModeParametersExtension configuration is missing from SecurityAnalysisRunParameters, using defaults.");
+            extension = (HybridModeParametersExtension) loadSpecificParameters(PlatformConfig.defaultConfig()).orElseThrow(() ->
+                    new IllegalArgumentException("Failed to load HybridModeParametersExtension configuration"));
```
olperr1: Since you haven't redefined `loadSpecificParameters` in this provider, it will call the default implementation of `SecurityAnalysisProvider` (which always returns `Optional.empty()`) when no extension is found on the SA parameters. In this case, this instruction will always throw the exception.

olperr1: You can take inspiration on what was done in OLF for this case:

- redefine the method
```java
    @Override
    public Optional<Extension<SecurityAnalysisParameters>> loadSpecificParameters(PlatformConfig platformConfig) {
        return Optional.of(HybridModeParametersExtension.load(platformConfig));
    }
```

- add methods similar to the 3 `load(...)` and the `update(...)` methods of OLF's [`OpenSecurityAnalysisParameters`](https://github.com/powsybl/powsybl-open-loadflow/blob/f7a4c223256299dad4c3fa14b1778412c1997879/src/main/java/com/powsybl/openloadflow/sa/OpenSecurityAnalysisParameters.java#L125-L161) in `HybridModeParametersExtension`.

This will allow to switch to the hybrid mode by only changing the configuration file (without having to change the caller code).

security-analysis/hybrid-security-analysis-api/src/main/java/com/powsybl/hybrid/security/analysis/HybridSecurityAnalysis.java
```diff
@@ -0,0 +1,176 @@
...
+ * @author Riad Benradi {@literal <riad.benradi at rte-france.com>}
+ */
+public class HybridSecurityAnalysis {
+    private static final Logger LOGGER = LoggerFactory.getLogger(HybridSecurityAnalysis.class);
+
+    private final Network network;
+    private final String workingVariantId;
+    private final ContingenciesProvider contingenciesProvider;
+    private final SecurityAnalysisRunParameters runParameters;
+    private final HybridModeParametersExtension extension;
```
olperr1: It would be clearer to rename all the `extension` variables to `hybridModeParameters`.

security-analysis/hybrid-security-analysis-api/src/main/java/com/powsybl/hybrid/security/analysis/HybridSecurityAnalysis.java
```diff
@@ -0,0 +1,176 @@
...
+                .filter(p -> p.getName().equals(providerName))
+                .findFirst()
+                .orElseThrow(() -> new IllegalArgumentException("Security analysis provider '" + providerName + "' not found by ServiceLoader."));
+    }
+
+    /**
+     * Executes the full hybrid-mode analysis workflow.
+     */
+    public CompletableFuture<SecurityAnalysisReport> run() {
+        LOGGER.info("Starting hybrid-mode security analysis");
```
olperr1: All these messages are added to the logger, but it may be useful to also have messages in the `ReportNode` (which are business messages processable by the caller. They are internationalized and could be displayed to the end-user).
For instance, OLF and dynaFlow both add messages to the `ReportNode` defined in the run parameters (returned by `runParameters.getReportNode()`).
I think that the following messages could be added to the `ReportNode`:
- "Hybrid-mode security analysis"
- "Total contingencies to analyze: {}"
- "First pass using {} provider"
- "Contingencies requiring second pass analysis: {}"
- "Second pass using {} provider" (if necessary)

security-analysis/hybrid-security-analysis-api/src/main/java/com/powsybl/hybrid/security/analysis/HybridSecurityAnalysis.java
```diff
@@ -0,0 +1,176 @@
...
+                LOGGER.info("No contingencies require second pass analysis, returning first pass results");
+                return CompletableFuture.completedFuture(firstReport);
+            }
+
+            // Run second pass analysis on filtered contingencies
+            return runSecondPassAnalysis(contingenciesForSecondPass)
+                .thenApply(secondReport -> mergeResults(firstReport, secondReport));
+        }).exceptionally(ex -> {
+            LOGGER.error("Error during hybrid-mode security analysis", ex);
+            return new SecurityAnalysisReport(SecurityAnalysisResult.empty());
```
olperr1: I don't know if returning an empty result is the best option when an error is encountered.
We should check with OLF's and dynaFlow's developers what policy they use.

security-analysis/hybrid-security-analysis-api/src/main/java/com/powsybl/hybrid/security/analysis/HybridSecurityAnalysis.java
```diff
@@ -0,0 +1,176 @@
...
+        // Step 1: Get all contingencies
+        List<Contingency> allContingencies = contingenciesProvider.getContingencies(network);
+        LOGGER.info("Total contingencies to analyze: {}", allContingencies.size());
+
+        // Step 2: Run first pass analysis
+        CompletableFuture<SecurityAnalysisReport> firstAnalysisFuture = firstProvider.run(
+            network, workingVariantId, contingenciesProvider, runParameters);
+
+        // Step 3: Chain second pass analysis based on first pass results
+        return firstAnalysisFuture.thenCompose(firstReport -> {
```
olperr1: Could you extract the content of this lambda in a new method to improve readability?

security-analysis/hybrid-security-analysis-api/src/main/java/com/powsybl/hybrid/security/analysis/HybridSecurityAnalysis.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    /**
+     * Selects contingencies that should be re-analyzed with the second provider:
+     * those that did not converge or triggered an automaton.
+     */
+    private List<Contingency> selectContingenciesForSecondPass(SecurityAnalysisResult firstResult,
+                                                                     List<Contingency> allContingencies) {
+        Map<String, Contingency> contingencyById = allContingencies.stream()
+            .collect(Collectors.toMap(Contingency::getId, c -> c));
```
olperr1: To avoid an unnecessary object creation:

```java
            .collect(Collectors.toMap(Contingency::getId, Function.identity()));
```

security-analysis/hybrid-security-analysis-api/src/main/java/com/powsybl/hybrid/security/analysis/HybridSecurityAnalysis.java
```diff
@@ -0,0 +1,176 @@
...
+
+    /**
+     * Runs second pass analysis on a subset of contingencies.
+     */
+    private CompletableFuture<SecurityAnalysisReport> runSecondPassAnalysis(List<Contingency> contingencies) {
+        LOGGER.info("Starting second pass analysis for {} contingencies", contingencies.size());
+
+        ContingenciesProvider filteredProvider = n -> contingencies;
+
+        return secondProvider.run(network, workingVariantId, filteredProvider, runParameters);
```
olperr1:
```suggestion
        LOGGER.info("Starting second pass analysis for {} contingencies", contingencies.size());
        ContingenciesProvider filteredProvider = n -> contingencies;
        return secondProvider.run(network, workingVariantId, filteredProvider, runParameters);
```

security-analysis/hybrid-security-analysis-api/src/main/java/com/powsybl/hybrid/security/analysis/HybridSecurityAnalysis.java
```diff
@@ -0,0 +1,176 @@
...
+     * For each contingency, the second pass result takes precedence over the first pass one.
+     */
+    private SecurityAnalysisReport mergeResults(SecurityAnalysisReport firstReport, SecurityAnalysisReport secondReport) {
+        LOGGER.info("Merging first and second pass analysis results");
+        SecurityAnalysisResult firstResult = firstReport.getResult();
+        SecurityAnalysisResult secondResult = secondReport.getResult();
+
+        Map<String, PostContingencyResult> secondResultsMap = secondResult.getPostContingencyResults()
+            .stream()
+            .collect(Collectors.toMap(r -> r.getContingency().getId(), r -> r));
```
olperr1:
```java
.collect(Collectors.toMap(r -> r.getContingency().getId(), Function.identity()));
```

security-analysis/hybrid-security-analysis-api/src/main/java/com/powsybl/hybrid/security/analysis/HybridSecurityAnalysis.java
```diff
@@ -0,0 +1,176 @@
...
+                    LOGGER.debug("Using first result for contingency {}", contingencyId);
+                    return firstResultItem;
+                }
+            })
+            .collect(Collectors.toList());
+
+        SecurityAnalysisResult finalResult = new SecurityAnalysisResult(
+            firstResult.getPreContingencyResult(),
+            mergedResults,
+            firstResult.getOperatorStrategyResults());
```
olperr1: Operator strategies can be applied after a contingency (see [the documentation](https://powsybl.readthedocs.io/projects/powsybl-core/en/stable/simulation/security/index.html#operator-strategies)). Therefore, we cannot just take the operator strategy results of the 1st analysis but we should merge them with the ones of the second analysis.

security-analysis/hybrid-security-analysis-api/src/main/java/com/powsybl/hybrid/security/analysis/HybridSecurityAnalysis.java
```diff
@@ -0,0 +1,176 @@
...
+
+        if (firstResult.getNetworkMetadata() != null) {
+            finalResult.setNetworkMetadata(firstResult.getNetworkMetadata());
+        }
+
+        LOGGER.info("Merge complete: {} post-contingency results", mergedResults.size());
+        return new SecurityAnalysisReport(finalResult)
+            .setLogBytes(firstReport.getLogBytes().orElse(null));
```
olperr1: I don't know how these log bytes are used, but I guess some users / projects may want the ones of the 2 analyses. I think this could be achieved by adding the 2 log bytes to a new Zip archive via [`ZipPackager.archiveBytesByNameToZipBytes(Map<String, byte[]> bytesByName)`](https://github.com/powsybl/powsybl-core/blob/ae8897cf4fdb76ddb64610cca42f640cbe1ebd14/commons/src/main/java/com/powsybl/commons/compress/ZipPackager.java#L165).

security-analysis/hybrid-security-analysis-api/src/test/java/com/powsybl/hybrid/security/analysis/parameters/HybridModeParametersExtensionTest.java
```diff
@@ -0,0 +1,40 @@
+package com.powsybl.hybrid.security.analysis.parameters;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+/** @author Riad Benradi {@literal <riad.benradi at rte-france.com>}*/
+class HybridModeParametersExtensionTest {
```
olperr1:
```suggestion

/** 
 * @author Riad Benradi {@literal <riad.benradi_externe at rte-france.com>}
 */
class HybridModeParametersExtensionTest {
```

security-analysis/hybrid-security-analysis-api/src/main/java/com/powsybl/hybrid/security/analysis/parameters/HybridModeParametersExtension.java
```diff
@@ -0,0 +1,46 @@
...
+
+import com.powsybl.commons.extensions.AbstractExtension;
+import com.powsybl.security.SecurityAnalysisParameters;
+
+/**
+ * This class contains configuration parameters specific to
+ * hybrid-mode security analysis.
+ * @author Riad Benradi {@literal <riad.benradi at rte-france.com>}
```
olperr1: In the "literal" section, your valid e-mail address is expected:

```suggestion
 * @author Riad Benradi {@literal <riad.benradi_externe at rte-france.com>}
```

(same thing for **all** the new files)

security-analysis/hybrid-security-analysis-api/src/test/java/com/powsybl/hybrid/security/analysis/HybridSecurityAnalysisProviderTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Test
+    void testProviderName() {
+        assertEquals("HybridSecurityAnalysis", provider.getName());
+    }
+
+    @Test
+    void testProviderVersion() {
+        String version = provider.getVersion();
+        assertNull(version);
```
olperr1: `assertEquals(new PowsyblCoreVersion().toString(), provider.getVersion());`

security-analysis/hybrid-security-analysis-api/src/test/java/com/powsybl/hybrid/security/analysis/HybridSecurityAnalysisProviderTest.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Test
+    void testRunMissingExtensionThrows() {
+        SecurityAnalysisParameters mockSaParams = mock(SecurityAnalysisParameters.class);
+        when(runParameters.getSecurityAnalysisParameters()).thenReturn(mockSaParams);
+        when(mockSaParams.getExtension(HybridModeParametersExtension.class)).thenReturn(null);
+        // This will try to load from PlatformConfig.defaultConfig() which should fail in test env if not found
```
olperr1: A passing test (with a proper config file) must be added.

security-analysis/hybrid-security-analysis-api/src/test/java/com/powsybl/hybrid/security/analysis/HybridSecurityAnalysisTest.java
```diff
@@ -0,0 +1,339 @@
...
+        when(secondProvider.getName()).thenReturn("SecondProvider");
+
+        HybridModeParametersExtension extension = new HybridModeParametersExtension();
+        extension.setFirstProviderName("FirstProvider");
+        extension.setSecondProviderName("SecondProvider");
+        contingency1 = mock(Contingency.class);
+
+        when(contingency1.getId()).thenReturn("contingency-1");
+        contingency2 = mock(Contingency.class);
+        when(contingency2.getId()).thenReturn("contingency-2");
```
olperr1: You can use real objects here:

```suggestion
        contingency1 = new Contingency("contingency-1");
        contingency2 = new Contingency("contingency-2");
```

security-analysis/hybrid-security-analysis-api/src/test/java/com/powsybl/hybrid/security/analysis/HybridSecurityAnalysisTest.java
```diff
@@ -0,0 +1,339 @@
...
+        verify(secondProvider, never()).run(any(), any(), any(), any());
+    }
+
+    @Test
+    void testRunSecondPassOnlyForDivergingContingencies() {
+        List<Contingency> allContingencies = Arrays.asList(contingency1, contingency2);
+        when(contingenciesProvider.getContingencies(network)).thenReturn(allContingencies);
+
+        Contingency contingency3 = mock(Contingency.class);
+        when(contingency3.getId()).thenReturn("contingency-3");
```
olperr1:
```suggestion
        Contingency contingency3 = new Contingency("contingency-3");
```

security-analysis/hybrid-security-analysis-api/src/test/java/com/powsybl/hybrid/security/analysis/HybridSecurityAnalysisTest.java
```diff
@@ -0,0 +1,339 @@
...
+                        Collections.emptyList())
+        );
+        when(secondProvider.run(any(), any(), any(), any())).thenReturn(CompletableFuture.completedFuture(secondReport));
+
+        SecurityAnalysisReport report = hybridSecurityAnalysis.run().join();
+
+        List<PostContingencyResult> results = report.getResult().getPostContingencyResults();
+        assertEquals(2, results.size());
+        PostContingencyResult result1 = results.stream()
+                .filter(r -> r.getContingency().getId().equals("contingency-1"))
```
olperr1: When comparing a variable to a String constant, it is a good practice to call `equals` on the constant:
```suggestion
                .filter(r -> "contingency-1".equals(r.getContingency().getId()))
```

security-analysis/hybrid-security-analysis-api/src/test/java/com/powsybl/hybrid/security/analysis/HybridSecurityAnalysisTest.java
```diff
@@ -0,0 +1,339 @@
...
+                .findFirst()
+                .orElseThrow();
+        // second result overrides first: CONVERGED status and 2 violations win
+        assertSame(PostContingencyComputationStatus.CONVERGED, result1.getStatus());
+        assertEquals(2, result1.getLimitViolationsResult().getLimitViolations().size());
+
+        PostContingencyResult result2 = results.stream()
+                .filter(r -> r.getContingency().getId().equals("contingency-2"))
```
olperr1: Same here

security-analysis/hybrid-security-analysis-api/src/test/java/com/powsybl/hybrid/security/analysis/HybridSecurityAnalysisTest.java
```diff
@@ -0,0 +1,339 @@
...
+        assertEquals(1, results.get(1).getLimitViolationsResult().getLimitViolations().size());
+        assertSame(PostContingencyComputationStatus.SOLVER_FAILED, results.get(2).getStatus());
+    }
+
+    @Test
+    void testConstructorWithServiceLoader() {
+        // This test verifies that the constructor that uses ServiceLoader works.
+        HybridModeParametersExtension extension = createExtension("NonExistentProvider1", "NonExistentProvider2");
+
+        assertThrows(IllegalArgumentException.class, () ->
```
olperr1: You should also check the exception message.

security-analysis/hybrid-security-analysis-api/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>com.powsybl</groupId>
+        <artifactId>powsybl-security-analysis</artifactId>
+        <version>7.3.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>powsybl-hybrid-security-analysis-api</artifactId>
+    <name>Hybrid Security Analysis API</name>
+    <description>An API for running hybrid (static and dynamic) security analyses.</description>
```
olperr1: This is not an API but a new implementation of the Security Analysis.
Can you remove the "api" term in the module name, the directory name, ... ?

olperr1 (changes requested):

olperr1:
I suggest to use the term `two-pass security analysis`. It would be more precise.

geofjamg:
> I suggest to use the term `two-pass security analysis`. It would be more precise.

Or may TwoLevelsSecurityAnalysis

olperr1:
> > I suggest to use the term `two-pass security analysis`. It would be more precise.
> 
> Or may TwoLevelsSecurityAnalysis

OK for `TwoLevelSecurityAnalysis`... but without the `s`.

jeandemanged:
I see the name of the new provider already changed many times, and therefore my apologies for what is coming...

"contingency screening" is a term well known in the industry, and refers specifically to the process of identifying whether a contingency is worth or not going through more complex calculations. From the PR description it corresponds only to the step 2 of the workflow: "Criteria Evaluation". Hence I think "Contingency screening SA" is not appropriate, I liked better the "Two-Pass SA" proposition. Or similarly, "Two-Stages SA", or why not "Multi-Stage SA" if in the future we want more than 2 passes/stages. More ideas: "ChainedSecurityAnalysisProvider", "SecurityAnalysisPipelineProvider"

RiadBRD:
@jeandemanged 
No need to apologize, and thanks a lot for the remark, it is actually very helpful. I did not know that “contingency screening” had such a specific meaning in the industry terminology, and I realize now that my PR description was probably not clear enough regarding the intended workflow.

What is currently implemented is a two-stage workflow with a first and second provider currently we use :

* a first pass using OLF for standard/static security analysis,
* followed by a second pass using Dynaflow for contingencies requiring dynamic assessment.

The initial idea behind the naming was that the OLF pass acts as a screening stage to determine which contingencies need escalation to Dynaflow.

With that interpretation in mind, would you still consider the term misleading, or do you think it could still fit even if the provider responsibility goes beyond the screening step itself?

jeandemanged:
Hi @RiadBRD ,
I was confused by the TSC slide communicated yesterday showing screening in between provider 1 and 2, indeed the screening is what is performed by the provider 1 (e.g. OLF static SA).

What I do not like about the name "Contingency Screening Security Analysis" is that it suggests it performs only screening (only provider 1), and not a complete SA (provider 1 -> filtering(and ranking?) -> provider 2).
The "Contingency Screening Security Analysis" name suggests that a it computes via approximations only a list a contingencies to be further processed, and this is not what is being built here - because there is provider 2 / Dynaflow providing accurate results wherever needed.

image

RiadBRD:
@jeandemanged So if I understand correctly, the name "Contingency Screening Security Analysis" would only be appropriate for the first part of the workflow — i.e. the provider used to determine which contingencies require dynamic analysis — but not for the complete workflow itself. Am I right?

geofjamg:
SecurityAnalysisPipeline would make sens but only if the 3 parts (sa1, screening, sa2) is kept generic et replaceable

CamilleHamon:
Let's compile the names proposed so far :
- HybridSecurityAnalysis
- ContingencyScreeningSecurityAnalysis
- TwoPassSecurityAnalysis
- MultiStageSecurityAnalysis
- TwoStageSecurityAnalysis
- SecurityAnalysisPipeline
- SecurityAnalysisPipelineProvider
- ChainedSecurityAnalysisProvider

What we want to convey in the name :
- It is, in the end, a security analysis.
- It uses contingency screening
- Contingency screening is performed by using a provider 1 that acts as a proxy for provider 2 (i.e. provider 2 is supposed to be more precise than provider 1).

For me two-stage or multi-stage is connoted with stochastic optimisation or preventive/corrective actions so I would prefer to avoid it. Of all the names, TwoPassSecurityAnalysis is probably the one that ticks most of the boxes although it doesn't convey the meaning that there is a contingency screening at play but it's probably ok.

security-analysis/hybrid-security-analysis-api/src/main/java/com/powsybl/hybrid/security/analysis/parameters/HybridModeParametersExtension.java
olperr1: The header is missing in `ContingencyScreeningSecurityAnalysisReportsTest`.

security-analysis/contingency-screening-security-analysis/src/main/java/com/powsybl/contingencyscreening/security/analysis/parameters/ContingencyScreeningSecurityAnalysisParameters.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public static ContingencyScreeningSecurityAnalysisParameters load() {
+        return load(PlatformConfig.defaultConfig());
+    }
+
+    public static ContingencyScreeningSecurityAnalysisParameters load(PlatformConfig platformConfig) {
+        ContingencyScreeningSecurityAnalysisParameters parameters = new ContingencyScreeningSecurityAnalysisParameters();
+        platformConfig.getOptionalModuleConfig(NAME).ifPresent(config -> {
+            parameters.setFirstProviderName(config.getStringProperty("firstProviderName"));
+            parameters.setSecondProviderName(config.getStringProperty("secondProviderName"));
```
olperr1: Could you change the setters to return this? It will allow to chain the calls.

security-analysis/contingency-screening-security-analysis/src/main/java/com/powsybl/contingencyscreening/security/analysis/ContingencyScreeningSecurityAnalysisReports.java
```diff
@@ -0,0 +1,56 @@
...
+package com.powsybl.contingencyscreening.security.analysis;
+
+import com.powsybl.commons.report.ReportNode;
+import com.powsybl.commons.report.TypedValue;
+
+/**
+ * @author Riad Benradi {@literal <riad.benradi_externe at rte-france.com>}
+ * */
+
+public final class ContingencyScreeningSecurityAnalysisReports {
```
olperr1: The corresponding `reports.properties` and `reports_fr.properties` files are missing.

security-analysis/contingency-screening-security-analysis/src/main/java/com/powsybl/contingencyscreening/security/analysis/ContingencyScreeningSecurityAnalysisHandler.java
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    private byte[] mergeLogBytes(SecurityAnalysisReport firstReport, SecurityAnalysisReport secondReport) {
+        java.util.Optional<byte[]> firstLogBytes = firstReport.getLogBytes();
+        java.util.Optional<byte[]> secondLogBytes = secondReport.getLogBytes();
+
+        if (firstLogBytes.isPresent() && secondLogBytes.isEmpty()) {
+            return firstLogBytes.get();
+        }
+        if (secondLogBytes.isPresent() && firstLogBytes.isEmpty()) {
```
olperr1: I find it more readable the other way. This makes it easier to compare with the previous `if` condition:

```suggestion
        if (firstLogBytes.isEmpty() && secondLogBytes.isPresent()) {
```

security-analysis/contingency-screening-security-analysis/src/main/java/com/powsybl/contingencyscreening/security/analysis/ContingencyScreeningSecurityAnalysis.java
```diff
@@ -0,0 +1,242 @@
...
+        if (firstLogBytes.isPresent() && secondLogBytes.isEmpty()) {
+            return firstLogBytes.get();
+        }
+        if (secondLogBytes.isPresent() && firstLogBytes.isEmpty()) {
+            return secondLogBytes.get();
+        }
+
+        if (firstLogBytes.isEmpty()) {
+            return new byte[0];
+        }
```
olperr1, line 213:
```suggestion
        if (firstLogBytes.isEmpty()) {
            // Both logs are empty
            return new byte[0];
        }
```

security-analysis/contingency-screening-security-analysis/src/test/java/com/powsybl/contingencyscreening/security/analysis/parameters/ContingencyScreeningSecurityAnalysisParametersTest.java
```diff
@@ -0,0 +1,94 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.*;
+/** * @author Riad Benradi {@literal <riad.benradi_externe at rte-france.com>}*/
+
+class ContingencyScreeningSecurityAnalysisParametersTest {
```
olperr1: Could you adopt this format which is the mostly used in our base code, **for all the new files**?
- an empty line between the imports and the javadoc comment
- A new line for the beginning and the end
- Only one star on the author and end lines
- No empty line before the class definition


```suggestion
import static org.junit.jupiter.api.Assertions.*;

/**
 * @author Riad Benradi {@literal <riad.benradi_externe at rte-france.com>}
 */
class ContingencyScreeningSecurityAnalysisParametersTest {
```

security-analysis/contingency-screening-security-analysis/src/main/java/com/powsybl/contingencyscreening/security/analysis/ContingencyScreeningSecurityAnalysisHandler.java
```diff
@@ -0,0 +1,242 @@
...
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * Business logic for contingency screening security analysis: first pass on all contingencies,
+ * then second pass on those that did not converge or triggered an automaton, results merged.
+ *
+ * @author Riad Benradi {@literal <riad.benradi_externe at rte-france.com>}
+ */
+public class ContingencyScreeningSecurityAnalysisHandler {
```
olperr1: I prefer without the "handler" part. We sometimes use it, but mostly for classes implementing the `ExecutionHandler` interface.

olperr1 (changes requested):

CamilleHamon:
Hello @jeandemanged @geofjamg @olperr1 @RiadBRD ,
We still haven't fully aligned on the name. I agree with @jeandemanged that contingency screening is a term used in the industry for a very particular task, which is not what we do here. 
I propose TwoPassSecurityAnalysis, as already proposed by @jeandemanged.

Any disagreement and if so, please provide alternatives that you prefer.

security-analysis/twopass-security-analysis/src/main/resources/com/powsybl/twopasssecurity/security/analysis/reports_fr.properties
olperr1, line 1: In this file, `contingency`/`contingencies` must be replaced by `défaut`/`défauts`.

security-analysis/twopass-security-analysis/src/main/resources/com/powsybl/twopasssecurity/security/analysis/reports.properties
olperr1, line 1: The contents of this file and the corresponding "fr" file must be moved in the 2 report files of `powsybl-commons`.

security-analysis/twopass-security-analysis/src/test/java/com/powsybl/twopasssecurity/security/analysis/parameters/TwoPassSecurityAnalysisParametersTest.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @Test
+    void testDefaultValues() {
+        assertNull(twoPassSecurityAnalysisParameters.getFirstProviderName());
+        assertNull(twoPassSecurityAnalysisParameters.getSecondProviderName());
+    }
+
+    @Test
+    void testSetBothProviders() {
```
olperr1: You can remove this test. It is already covered by `testCompleteConfiguration`.

security-analysis/twopass-security-analysis/src/test/java/com/powsybl/twopasssecurity/security/analysis/TwoPassSecurityAnalysisReportsTest.java
```diff
@@ -0,0 +1,78 @@
+package com.powsybl.twopasssecurity.security.analysis;
```
olperr1, line 8: The header is missing in this file:

```suggestion
/**
 * Copyright (c) 2026, RTE (http://www.rte-france.com)
 * This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/.
 * SPDX-License-Identifier: MPL-2.0
 */
package com.powsybl.twopasssecurity.security.analysis;
```

security-analysis/twopass-security-analysis/src/test/java/com/powsybl/twopasssecurity/security/analysis/TwoPassSecurityAnalysisReportsTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+import com.powsybl.commons.report.ReportNode;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * @author Riad Benradi {@literal <riad.benradi_externe at rte-france.com>} */
```
olperr1:
```suggestion
/**
 * @author Riad Benradi {@literal <riad.benradi_externe at rte-france.com>}
 */
```

security-analysis/twopass-security-analysis/src/main/resources/com/powsybl/twopasssecurity/security/analysis/reports_fr.properties
```diff
@@ -0,0 +1,5 @@
+twoPassSecurityAnalysis = Ex�cution de l'analyse de s�curit� avec filtrage des contingences sur le r�seau '${networkId}'.
```
olperr1: `Exécution de l'analyse de sécurité en deux passages sur le réseau '${networkId}'.`

security-analysis/twopass-security-analysis/src/main/java/com/powsybl/twopasssecurity/security/analysis/TwoPassSecurityAnalysisReports.java
```diff
@@ -0,0 +1,55 @@
...
+ * @author Riad Benradi {@literal <riad.benradi_externe at rte-france.com>}
+ */
+public final class TwoPassSecurityAnalysisReports {
+
+    private TwoPassSecurityAnalysisReports() {
+    }
+
+    public static ReportNode createTwoPassSecurityAnalysisReportNode(ReportNode reportNode, String networkId) {
+        return reportNode.newReportNode()
+                .withMessageTemplate("twoPassSecurityAnalysis")
```
olperr1: For consistency with the existing report keys, could you use the following keys instead?

- `twoPassSecurityAnalysis` -> `core.securityAnalysis.twoPass.start`
- `twoPassSecurityAnalysisTotalContingencies` -> `core.securityAnalysis.twoPass.totalContingencies`
- `twoPassSecurityAnalysisFirstPassStarted` -> `core.securityAnalysis.twoPass.firstPassStarted`
- `twoPassSecurityAnalysisSecondPassRequired` -> `core.securityAnalysis.twoPass.secondPassRequired`
- `twoPassSecurityAnalysisSecondPassStarted` -> `core.securityAnalysis.twoPass.secondPassStarted`

security-analysis/twopass-security-analysis/src/main/java/com/powsybl/twopasssecurity/security/analysis/TwoPassSecurityAnalysis.java
```diff
@@ -0,0 +1,243 @@
...
+
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * Business logic for two-pass security analysis: first pass on all contingencies,
+ * then second pass on those that did not converge or triggered an automaton, results merged.
```
olperr1: Could you remove the parts where the automata are mentioned?
In this PR, it is not yet handled.

security-analysis/twopass-security-analysis/src/main/java/com/powsybl/twopasssecurity/security/analysis/TwoPassSecurityAnalysis.java
```diff
@@ -0,0 +1,243 @@
...
+     */
+    private List<Contingency> selectContingenciesForSecondPass(SecurityAnalysisResult firstResult,
+                                                                     List<Contingency> allContingencies) {
+        Map<String, Contingency> contingencyById = allContingencies.stream()
+            .collect(Collectors.toMap(Contingency::getId, Function.identity()));
+
+        return firstResult.getPostContingencyResults().stream()
+            .filter(this::requiresSecondPass)
+            .map(r -> contingencyById.get(r.getContingency().getId()))
+            .filter(java.util.Objects::nonNull)
```
olperr1: In this class, the full `java.util.` package is used for several classes (just like here).
Can you adapt the class to remove these mentions?

olperr1 (changes requested):

commons/src/main/resources/com/powsybl/commons/reports_fr.properties
```diff
@@ -200,6 +200,11 @@ core.iidm.serde.importedNetwork = Le réseau ${networkId} est importé à partir
 core.iidm.serde.validationWarnings = Avertissements sur la validation.
 core.iidm.serde.xiidmImportDone = Import XIIDM terminé.
 core.loadflow.loadflowTool = Outils de calcul de répartition.
+core.securityAnalysis.twoPass.firstPassStarted = Premier passage d�marr� avec le fournisseur '${providerName}'.
```
olperr1: You have an encoding problem on this file.

olperr1 (changes requested):

olperr1 (approved):
Could you create another PR with the documentation on this feature?
This other PR could be merged between the RC and the final release.

security-analysis/twopass-security-analysis/pom.xml
```diff
@@ -0,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
olperr1, line 1:
```suggestion
<?xml version="1.0" encoding="UTF-8"?>
<!--
    Copyright (c) 2026, RTE (http://www.rte-france.com)
    This Source Code Form is subject to the terms of the Mozilla Public
    License, v. 2.0. If a copy of the MPL was not distributed with this
    file, You can obtain one at http://mozilla.org/MPL/2.0/.
-->
```

olperr1 (changes requested):

security-analysis/twopass-security-analysis/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+    <parent>
+        <groupId>com.powsybl</groupId>
+        <artifactId>powsybl-security-analysis</artifactId>
+        <version>7.4.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>powsybl-twopass-security-analysis</artifactId>
+    <name>TwoPass Security Analysis</name>
+    <description>A module for running two-pass security analysis.</description>
+
```
olperr1: Could you add this section here?


```suggestion

    <build>
        <plugins>
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-jar-plugin</artifactId>
                <configuration>
                    <archive>
                        <manifestEntries>
                            <Automatic-Module-Name>com.powsybl.iidm.twopass.security.analysis</Automatic-Module-Name>
                        </manifestEntries>
                    </archive>
                </configuration>
            </plugin>
        </plugins>
    </build>

```

security-analysis/twopass-security-analysis/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+            <groupId>com.google.auto.service</groupId>
+            <artifactId>auto-service-annotations</artifactId>
+            <version>${autoservice.version}</version>
+        </dependency>
+
+        <!-- Test dependencies -->
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <version>${junit-jupiter.version}</version>
```
olperr1:
```suggestion
            <artifactId>junit-jupiter</artifactId>
```

olperr1 (changes requested):
I missed some points in the `pom.xml`:

Could you add this dependency in the "dependencyManagement" section of the main `pom.xml`, just after `powsybl-security-analysis-api`?

```xml
            <dependency>
                <groupId>${project.groupId}</groupId>
                <artifactId>powsybl-twopass-security-analysis</artifactId>
            </dependency>
```

olperr1 (approved):

PR #3883, opened 2026-04-22
Powerfactory: loadflow results comparison

marqueslanauja:

Feature

ElmTerms are converted to busbarSections based on the iUsage attribute. When iUsage is zero, a busbarSection is created and associated with the corresponding ElmTerm.

An import parameter, `PowerFactoryImporter.FORCE_ALL_ELMTERMS_AS_BUSBARS`, has been added. When this parameter is set to true, a busbarSection is created and associated with each ElmTerm.

When the attribute is true, all the calculated buses have at least a busbarSection and the id of the busbarSection is built by the concatenation of the voltageLevel Id and the `loc_name` of the ElmTerm. As the powerfactory results are associated with the ElmTerms will be possible to compare with the PowSyBl results.

- [ ] Yes
- [x] No

**Other information**:


geofjamg (approved):
Thanks, this is a very useful option to compare to PF load flow values.

PR #3798, opened 2026-02-27
Do not allow connect/disconnect to modify fictionnal breaker by default

NathanDissoubray:
**Please check if the PR fulfills these requirements**

- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] A PR or issue has been opened in all impacted repositories (if any)

**Does this PR already have an issue describing the problem?**

Fixes #3790 

**What kind of change does this PR introduce?**

Make connect / disconnect symmetrical by default.

**What is the current behavior?**

Disconnect can operate on fictitious breaker by default, but connect cannot.

**What is the new behavior (if this is a feature change)?**

Both disconnect and connect cannot operate on fictitious breaker by default.

**Does this PR introduce a breaking change or deprecate an API?**
- [ ] Yes
- [x] No, but custom IIDM maintainers will have to slightly adapt their code

**If yes, please check if the following requirements are fulfilled**

- [ ] The *Breaking Change* or *Deprecated* label has been added
- [ ] The migration steps are described in the following section

**What changes might users need to make in their application due to this PR? (migration steps)**

Custom IIDM implementations' maintainers should modify their `Connectable.disconnect()` to only change the state of the **non-fictional** breakers, instead of all the breakers.

**Other information**:

Though it is not possible to use disconnect / connect to operate on fictitious breaker by default, it is still possible to operate those by using a `SwitchPredicates` as such:
- `connectable.connect(SwitchPredicates.IS_BREAKER_OR_DISCONNECTOR)`
- `connectable.disconnect(SwitchPredicates.IS_CLOSED_BREAKER)`

PR #3885, opened 2026-04-23
Fix asymmetric connect / disconnect terminal, tie line, hvdc line and add default methods

### LINKED ISSUE #3884: Connect/disconnect default predicates not symmetrical on Terminal

When disconnecting a connected terminal and reconnecting it, the terminal might be stuck in disconnected state.

```java
terminal.disconnect();
terminal.connect();
```
should result in the same connected state as initially. If the disconnection is allowed to manipulate fictitious breakers, the connection should also.

### Describe the steps

Step 1: Disconnect a connected terminal connected on one side to a busbar with several breakers, the first one being a fictitious one: the fictitious breaker is opened to disconnect the terminal.
Step 2: Reconnect that disconnected line: the reconnection fails.

### Environment

_No response_

### Relevant Log Output

_No response_

### Extra Information

This is similar to issue https://github.com/powsybl/powsybl-core/issues/3790 which was fixed for Connectable but not for Terminal which is still asymmetrical. 
See the associated code https://github.com/powsybl/powsybl-core/blob/main/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractTerminal.java#L168-L198

NathanDissoubray:

Fixes #3884 
Also same fix for tie line and hvdc line

Make the connect / disconnect default function symmetric.

Disconnect can operate on fictitious breakers.

Both connect and disconnect cannot operate on fictitious breakers by default. It is still possible to do so using the connect/Disconnect function taking a `SwitchPredicates`

- [ ] Yes
- [x] No, but custom IIDM maintainers will have to slightly adapt their code

Custom IIDM implementations' maintainers should modify:
- `Terminal.disconnect()`
- `TIeLine.disconnectBoundaryLines()`
- `HvdcLine.disconnectConverterStations()`

 to only change the state of the **non-fictional** breakers, instead of all the breakers.

For users relying on the previous default to disconnect fictitious breakers, to disconnect fictitious breakers, use:
- `Terminal.disconnect(SwitchPredicates.IS_CLOSED_BREAKER)`
- `TIeLine.disconnectBoundaryLines(SwitchPredicates.IS_CLOSED_BREAKER)`
- `HvdcLine.disconnectConverterStations(SwitchPredicates.IS_CLOSED_BREAKER)`

**Other information**:

Missing some TCK tests


iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractHvdcLineTest.java
```diff
@@ -253,6 +253,29 @@ void testConnectDisconnect() {
...
+        assertHvdcLineConnection(hvdcLine, false, false);
+
+        //Fails since connect cannot operate fictitious breakers by default
+        assertFalse(hvdcLine.connectConverterStations());
+        assertHvdcLineConnection(hvdcLine, false, false);
+
+        //Force closed
+        assertTrue(hvdcLine.connectConverterStations(SwitchPredicates.IS_BREAKER_OR_DISCONNECTOR));
+        assertHvdcLineConnection(hvdcLine, true, true);
+    }
```
samirromdhani, line 277: I think some checks are too specific to be in the TCK, checks like `hvdcLine.connectConverterStations()` and  `hvdcLine.disconnectConverterStations()` depend on the implementation used.

Have you tried to run this test on powsybl-network-store ?
If the goal is to enforce this behavior in every implementations, then it is a good idea to keep this tests in the TCK.

Same remark for `AbstractTieLineTest`

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/HvdcLine.java
```diff
@@ -206,12 +206,24 @@ default HvdcConverterStation<?> getConverterStation(TwoSides side) {
      */
     void remove();
 
+    /**
+     * Connect the converter station.<br/>
+     * By default, this method does not change the state of fictitious breakers. If you wish to do that,
+     * use {@link #connectConverterStations(Predicate)} with {@link com.powsybl.iidm.network.util.SwitchPredicates#IS_BREAKER_OR_DISCONNECTOR}
```
rolnico: If we are to add this comment/expectation here, the a default method should be given.

```java
    /**
     * Connect the converter station.<br/>
     * By default, this method does not change the state of fictitious breakers. If you wish to do that,
     * use {@link #connectConverterStations(Predicate)} with {@link com.powsybl.iidm.network.util.SwitchPredicates#IS_BREAKER_OR_DISCONNECTOR}
     * @return true if the station has been connected, false otherwise
     */
    default boolean connectConverterStations() {
        return connectConverterStations(SwitchPredicates.IS_NONFICTIONAL_BREAKER, null);
    }
```
Else, the tests on those methods should be implementation-dependent and be relocated from the TCK to the implementation-specific tests

(this comment is the same for each method in the different classes)

rolnico (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/HvdcLine.java
NathanDissoubray: will do
I also put the same kind of comment on Connectable in #3798, should I provide a default implementation for Connectable too ?

rolnico: Yes this change has to be done everywhere (we have to be consistent)

NathanDissoubray: done in da5119ff6855

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Terminal.java
```diff
@@ -151,11 +152,14 @@ public static interface BusView {
 
     /**
      * Try to connect the terminal.<br/>
-     * Depends on the working variant.
+     * Depends on the working variant. By default, this method does not change the state of fictitious breakers. If you wish to do that,
+     * use {@link #connect(Predicate)} with {@link com.powsybl.iidm.network.util.SwitchPredicates#IS_BREAKER_OR_DISCONNECTOR}
```
rolnico:
```suggestion
     * use {@link #connect(Predicate)} with {@link com.powsybl.iidm.network.util.SwitchPredicates#IS_BREAKER}
```

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/TieLine.java
```diff
@@ -131,16 +133,43 @@ default IdentifiableType getType() {
      */
     void remove(boolean updateBoundaryLines);
 
-    boolean connectBoundaryLines();
+    /**
+     * Try to connect the two boundary lines of the tie line.<br/>
+     * By default, this method does not change the state of fictitious breakers. If you wish to do that,
+     * use {@link #connectBoundaryLines(Predicate)} with {@link com.powsybl.iidm.network.util.SwitchPredicates#IS_BREAKER_OR_DISCONNECTOR}
```
rolnico:
```suggestion
     * use {@link #connectBoundaryLines(Predicate)} with {@link com.powsybl.iidm.network.util.SwitchPredicates#IS_BREAKER}
```

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Connectable.java
```diff
@@ -29,7 +31,9 @@ public interface Connectable<I extends Connectable<I>> extends Identifiable<I> {
      * By default, this method does not change the state of fictitious breakers. If you wish to do that, use
      * {@link #connect(Predicate)} with {@link com.powsybl.iidm.network.util.SwitchPredicates#IS_BREAKER_OR_DISCONNECTOR}.
```
rolnico:
```suggestion
     * {@link #connect(Predicate)} with {@link com.powsybl.iidm.network.util.SwitchPredicates#IS_BREAKER}.
```

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/HvdcLine.java
```diff
@@ -206,16 +208,44 @@ default HvdcConverterStation<?> getConverterStation(TwoSides side) {
      */
     void remove();
 
-    boolean connectConverterStations();
+    /**
+     * Connect the converter stations.<br/>
+     * By default, this method does not change the state of fictitious breakers. If you wish to do that,
+     * use {@link #connectConverterStations(Predicate)} with {@link com.powsybl.iidm.network.util.SwitchPredicates#IS_BREAKER_OR_DISCONNECTOR}
```
rolnico:
```suggestion
     * use {@link #connectConverterStations(Predicate)} with {@link com.powsybl.iidm.network.util.SwitchPredicates#IS_BREAKER}
```

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Connectable.java
```diff
@@ -29,7 +31,9 @@ public interface Connectable<I extends Connectable<I>> extends Identifiable<I> {
...
      */
-    boolean connect();
+    default boolean connect() {
+        return connect(SwitchPredicates.IS_NONFICTIONAL_BREAKER);
+    }
 
     boolean connect(Predicate<Switch> isTypeSwitchToOperate);
```
rolnico: While we are at it, add the default implementation here too?
I think the TCK also has tests on those methods

rolnico: and rename the PR something like "Add default and symetric connect / disconnect methods to API"?

rolnico (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Connectable.java
NathanDissoubray: in our implementation, we are using the fact that passing null to connect(Predicate, TwoSides) operates on all sides (for Hvdc and TieLine). If we want to make a default based on this assumption, we should therefore add this requirement in the javadoc of those functions that take a side, is that acceptable ?

rolnico: Is it not already what is expected in the TCK tests?

NathanDissoubray: > and rename the PR something like "Add default and symetric connect / disconnect methods to API"?

I kept the "fix" in the PR title

NathanDissoubray: > Is it not already what is expected in the TCK tests?

there is no TCK test with the side at `null`

rolnico: Yes there are: every test where you don't provide argument (like `tieLine.connectBoundaryLines()` in `AbstractTieLineTest#testConnectDisconnect`) calls the method with the side as "null", as it was the implemented default method in iidm-impl.
In this test we try at least once all 6 methods.

NathanDissoubray: Isn't the side as null only in the implementation ? In the TCK test, I do not see any call in `testConnectDisconnect` where we explicitly call with null.
Indeed it calls with null in the impl, but looking at the tck test, no guarantee is made regarding the behaviour of those functions with the null argument for the side. If we want to make guarantees, shouldn't we directly call with null inside the TCK test, instead of relying on the fact that is done like that in our impl ?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Connectable.java
```diff
@@ -27,22 +29,44 @@ public interface Connectable<I extends Connectable<I>> extends Identifiable<I> {
...
-     * {@link #connect(Predicate)} with {@link com.powsybl.iidm.network.util.SwitchPredicates#IS_BREAKER_OR_DISCONNECTOR}.
+     * {@link #connect(Predicate)} with {@link com.powsybl.iidm.network.util.SwitchPredicates#IS_BREAKER}.
      */
-    boolean connect();
+    default boolean connect() {
+        return connect(SwitchPredicates.IS_NONFICTIONAL_BREAKER);
+    }
 
-    boolean connect(Predicate<Switch> isTypeSwitchToOperate);
+    default boolean connect(Predicate<Switch> isTypeSwitchToOperate) {
```
rolnico, line 44: Add the Javadoc for this method too

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Connectable.java
```diff
@@ -27,22 +29,44 @@ public interface Connectable<I extends Connectable<I>> extends Identifiable<I> {
...
      * By default, this method does not change the state of fictitious breakers. If you wish to do that, use
      * {@link #disconnect(Predicate)} with {@link com.powsybl.iidm.network.util.SwitchPredicates#IS_CLOSED_BREAKER}.
      */
-    boolean disconnect();
+    default boolean disconnect() {
+        return disconnect(SwitchPredicates.IS_NONFICTIONAL_CLOSED_BREAKER);
+    }
 
-    boolean disconnect(Predicate<Switch> isSwitchOpenable);
+    default boolean disconnect(Predicate<Switch> isSwitchOpenable) {
```
rolnico, line 72: Add the Javadoc for this method too

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/HvdcLine.java
```diff
@@ -206,16 +208,60 @@ default HvdcConverterStation<?> getConverterStation(TwoSides side) {
      */
     void remove();
 
-    boolean connectConverterStations();
+    /**
+     * Connect the converter stations.<br/>
+     * By default, this method does not change the state of fictitious breakers. If you wish to do that,
+     * use {@link #connectConverterStations(Predicate)} with {@link com.powsybl.iidm.network.util.SwitchPredicates#IS_BREAKER}
```
rolnico:
```suggestion
     * By default, this method only operates on non-fictitious breakers. If you wish to operate on other switches,
     * use {@link #connectConverterStations(Predicate)} with another specific {@link com.powsybl.iidm.network.util.SwitchPredicates}
```
Or something like this (it's a more positive way to express what is meant). This should be applied to other similar methods too.

rolnico (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Connectable.java
NathanDissoubray: I did it the other day but forgot to push, done in 2c2d8ed2b24f

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Connectable.java
NathanDissoubray: done in 2c2d8ed2b24f

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/HvdcLine.java
NathanDissoubray: done in f446c99cd958
I left the reference to the specific SwitchPredicates for connect / disconnect since it's what was used before to operate on fictitious breakers

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Connectable.java
```diff
@@ -26,23 +28,57 @@ public interface Connectable<I extends Connectable<I>> extends Identifiable<I> {
 
     /**
      * Connects the connectable by operating real breakers.<br>
-     * By default, this method does not change the state of fictitious breakers. If you wish to do that, use
-     * {@link #connect(Predicate)} with {@link com.powsybl.iidm.network.util.SwitchPredicates#IS_BREAKER_OR_DISCONNECTOR}.
+     * By default, this method only operates on non-fictitious breakers. If you wish to operate on other switches, use
+     * {@link #connect(Predicate)} with another specific {@link com.powsybl.iidm.network.util.SwitchPredicates} such as {@link com.powsybl.iidm.network.util.SwitchPredicates#IS_BREAKER}.
+     * @return true if the connectable has been connected by this operation, false otherwise (this fails if it was already connected on any side)
```
rolnico:
```suggestion
     * @return true if the connectable has been connected by this operation, false otherwise (the connectable could not be connected or it was already connected)
```

(to be changed on the other methods)

rolnico (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Connectable.java
NathanDissoubray: done in 6925751e51ab

rolnico (approved):

PR #3890, opened 2026-04-28
Added test for Anonymize network id

### LINKED ISSUE #2113: Anonymize network id
Bug

Network id is not anonymized

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
1. Export with anonymized data a network which contains an extension on the network 
2. Try to import the anonymized generated file and get the `PowsyblException` `"Identifiable xxx not found"`
3. Look the generated file and see the network id in the `iidm:network` tag and the corresponding anonymized id in the `iidm:extension` tag

Network id is replaced with generated id everywhere

**Other informations**
`NetworkXml::writeMainAttributes` is done too early in a method called `initializeWriter` which initializes the writer indeed, but also does some other things.

samirromdhani:

Not a fix , just test added to prove that bug not reproduced 
Fixes #2113

Test

- [ ] Yes
- [x] No

**Other information**:

This is not a fix, only a test added to confirm the bug cannot be reproduced.

### DIFF
--- a/iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/IdentifiableExtensionSerDeTest.java
+++ b/iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/IdentifiableExtensionSerDeTest.java
@@ -24,10 +24,14 @@ import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
 import java.util.Properties;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import static com.powsybl.commons.test.ComparisonUtils.assertXmlEquals;
 import static com.powsybl.iidm.serde.IidmSerDeConstants.CURRENT_IIDM_VERSION;
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
 import static org.junit.jupiter.api.Assertions.*;
 
 /**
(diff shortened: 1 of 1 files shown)


iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/IdentifiableExtensionSerDeTest.java
```diff
@@ -189,4 +196,49 @@ void testThrowErrorUnsupportedExtensionVersion2() {
...
     }
+
+    @Test
+    void testAnonymizedNetworkIdWithExtension() throws ParserConfigurationException, SAXException, IOException {
+        // Given Network with extension
+        Network network = EurostagTutorialExample1Factory.create();
+        network.addExtension(NetworkSourceExtension.class, new NetworkSourceExtensionImpl("source"));
+        String originalNetworkId = network.getId();
+        // When export with anonymized data
+        ByteArrayOutputStream os = new ByteArrayOutputStream();
```
rolnico: Would it not be easier to export the network in a file, and then re-import the network with and without the anonymizer to compare the two networks?

Something like this, with additional comparisons that you can add on the extension:

```java
        Path exportedFile = tmpDir.resolve("anonymized-network-with-extension.xiidm");
        Anonymizer anonymizer = NetworkSerDe.write(network, new ExportOptions().setAnonymized(true), exportedFile);
        Network importedNetwork = NetworkSerDe.read(exportedFile);
        assertEquals(anonymizer.anonymizeString(originalNetworkId), importedNetwork.getId());
        Network importedNetworkDeanonymized = NetworkSerDe.read(exportedFile, new ImportOptions(), anonymizer, NetworkFactory.findDefault(), ReportNode.NO_OP);
        assertEquals(originalNetworkId, importedNetworkDeanonymized.getId());
```

rolnico (changes requested):

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/IdentifiableExtensionSerDeTest.java
samirromdhani: Hello @rolnico, yes, it’s easier.
I modified the test as you suggested, it still prove that the bug is not reproducible.

import with anonymizer is possible for Network as Path argument, but with 2 extra arguments, that’s why I kept using the import with an input stream which doesn’t need unnecessary arguments in my previous commit.

rolnico (approved):

PR #3891, opened 2026-04-29
More detailed error messages at temporary limits creation

EtienneLt:

No

Feature

The error messages at temporary limits' creation don't contain the name of the limit nor the concerned operational limits group.

Add more information in the temporary limits' creation error messages.

- [x] Yes
- [ ] No

The constructors of the following classes, of the in-memory IIDM implementation, take an additional `String operationalLimitGroupId` argument:
- `AbstractLoadingLimitsAdder`
- `ActivePowerLimitsAdderImpl`
- `ApparentPowerLimitsAdderImpl`
- `CurrentLimitsAdderImpl`

Custom IIDM implementations' maintainers should change the exception messages that may be encountered at the temporary limits creation:
- When the value is `NaN`:
  - `temporary limit value is not set for '<name>' within limit set '<operationalGroupId>'`
- When the value is `< 0`:
  - `temporary limit value must be >= 0 for '<name>' within limit set '<operationalGroupId>'`
- When the acceptable duration is null:
  - `acceptable duration is not set for '<name>' within limit set '<operationalGroupId>'`
- When the acceptable duration is `< 0`:
  - `acceptable duration must be >= 0 for '<name>' within limit set '<operationalGroupId'>`


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractLoadingLimitsAdder.java
```diff
@@ -114,6 +115,8 @@ private void checkAndGetUniqueName() {
                     i++;
                 }
                 name = uniqueName;
+            } else if (nameExists(name)) {
+                throw new ValidationException(validable, "temporary limit name '" + name + "' should be unique within limit set '" + operationalGroupId + "'");
```
olperr1: This change must be reverted. We don't want to forbid having the same name for different temporary limits in the same group.
Note that the attribute is named `name` (and not `id`) to reflect this characteristic.

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractCurrentLimitsTest.java
```diff
@@ -468,9 +468,9 @@ public void ensureNameUnicity() {
     }
 
     @Test
-    public void testNameDuplicationIsAllowed() {
+    public void testNameDuplicationIsNotAllowed() {
```
olperr1: This unit test change must be reverted (see the previous comment).

olperr1 (changes requested):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractLoadingLimitsAdder.java
olperr1: The PR's title and description must also be updated accordingly.

olperr1 (approved):

Issue #3757, opened 2026-02-04
Method to transport group or load to the voltage level above


gautierbureau:
### Describe the current behavior
In some networks, the generators are in a separate voltage level, therefore hidden behind a transformer from the "core" voltage level.

### Describe the expected behavior
Two visions for this method:
- vizualization
- computation

In the first case it would just be easier in network diagrams to see groups on the vl above.

In the second we would need to compute the transported values to the voltage level above. The only complicated case is when we have two groups behind the same transformer. Otherwise the new value is just the transformer injected value. In this vision we would create a new network.

gautierbureau:
@flo-dup

PR #3893, opened 2026-04-30
Detect specific group of elements that could be replaced by non-electrically equivalent generator

### LINKED ISSUE #3878: Detect and merge voltage levels containing only generator, loads into voltage level above - Visualization

In some voltage levels, there are only generators and loads linked to the voltage level above by one or more two-winding transformer.
They are "hidden" from the core voltage level behind a transformer.

Be able to calculate an equivalent generator / load on the other side of the transformer. The voltage level above would directly have an equivalent group.
We also need to be able to detect those groups.

### Describe the motivation

This would allow for a more succinct view of a given network.

### Extra Information

Calculating an equivalent should be done by using the value of the P/Q on the terminal on the core voltage level (as this is what the core voltage level sees).
If a breaker is open on one of the equipment we are trying to merge, we will not make the merge to not display incorrect information.

NathanDissoubray:

Fixes #3878 

Allow detection of elements behind a two winding transformer that can be reduced to an equivalent generator.
The criteria are:
- the voltage level must contain only reducible elements
- all breakers in the voltage level must be closed (to prevent issue where a part of the network is incorrectly reduced)
- a set of reducible elements must be linked to the voltage level above through a single two-winding transformer (but there can be multiple transformer per voltage level)
- there must be at least one generator per set of reducible elements

This does not modify the network, it only does detection.

N/A

- [ ] Yes
- [x] No

**Other information**:

This is to be used only for visualisation purpose, by replacing a two-winding transformer and the elements behind it by a generator, and using the active / reactive power at the terminal of the transformer as an equivalent P/Q injection. This supposes a loadflow has been performed first.
This would not work for calculations.

### DIFF
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/Networks.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/Networks.java
@@ -481,6 +481,7 @@ public final class Networks {
     public static Stream<ReducibleTransformerData> getReducibleTransformerDataStream(Network network) {
         return network.getVoltageLevelStream()
             .filter(v -> StreamSupport.stream(v.getConnectables().spliterator(), false).allMatch(c -> isReducibleElement(c.getType())))
+            .filter(v -> StreamSupport.stream(v.getSwitches().spliterator(), false).noneMatch(Switch::isOpen))
             .flatMap(v -> v.getTwoWindingsTransformerStream()
                 .map(t -> new ReducibleTransformerData(
                     t,
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/Networks.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/Networks.java
@@ -479,15 +479,23 @@ public final class Networks {
     }
 
     public static Stream<ReducibleTransformerData> getReducibleTransformerDataStream(Network network) {
-        return network.getVoltageLevelStream()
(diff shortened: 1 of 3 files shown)


NathanDissoubray:
I'm not sure if the test networks I wrote should be in Eurostag, but I don't know where else I could put those.

geofjamg (changes requested):
I don't understand the purpose of this PR. I f you really want to have a more compact visualization of the grid it is not the right place to do it. You should do it on SLD repo side on SLD internal model.
From an electrical point of view you cannot remove the transformer and move the generator. It is not equivalent!!! Even on base case, it is not equivalent. You would need to adapt active and reactive limits of the generator to take into account active and reactive losses of the transformer. And even doing that N-K would not be the same. We cannot let a buggy code done on IIDM just for visu purposes.
Also notice that visualizing a network should never change the network. It should only be a read operation on the network otherwise you will get awful side effects

NathanDissoubray:
I was actually hesitant on putting this detection directly in SLD. The original issue (#3757) asked for an equivalent for both visualisation and computation, so it made sense to find this equivalent in core. As you pointed out, for computing purposes, one would need to adapt the active and reactive limits (which was something I noticed after I started working on it, and it was more complex than I first expected). It's not equivalent at all to just replace stuff like that.

After the issue was split into 2 (one for visualisation, one for computation), I just continued on core, but I should probably have done this part on diagram directly.

NathanDissoubray:
Closing this as stale since there were no news about the demand.

PR #3901, opened 2026-05-05
Add new types of sensibility variable

gautierbureau:

No

Feature

We added new types of sensibility variable:
- `SHUNT_COMPENSATOR_SUSCEPTANCE`
- `BRANCH_RESISTANCE`
- `BRANCH_REACTANCE`
- `BRANCH_ADMITTANCE`
- `SVC_PILOT_POINT_TARGET_VOLTAGE`

- [X] Yes
- [ ] No

A new version of sensitivity analysis result files, the 1.2, was introduced.
It allows to have factors with new sensitivity variable types:
- `SHUNT_COMPENSATOR_SUSCEPTANCE`
- `BRANCH_RESISTANCE`
- `BRANCH_REACTANCE`
- `BRANCH_ADMITTANCE`
- `SVC_PILOT_POINT_TARGET_VOLTAGE`

**Other information**:


geofjamg (approved):

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityVariableType.java
```diff
@@ -43,7 +43,8 @@ public enum SensitivityVariableType {
     /** increase in degrees */
     TRANSFORMER_PHASE_2(2),
     /** increase in degrees */
-    TRANSFORMER_PHASE_3(3);
+    TRANSFORMER_PHASE_3(3),
+    SHUNT_B;
```
geofjamg: Could you describe the unit of the sensitivity?

geofjamg (dismissed):

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityVariableType.java
gautierbureau: done

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityVariableType.java
olperr1, line 1: The serialization version of the objects containing attributes of the `SensitivityVariableType` type should be increased.
This type is used in the following objects that are serialized:
- `SensitivityFactor`
- `SensitivityAnalysisResult`

For the latter, you can simply change `SensitivityAnalysisResult.VERSION` from `1.1` to `1.2` + add unit tests in `SensitivityAnalysisResultTest` (no need to change the corresponding serializer / deserializer).

But for `SensitivityFactor`, it's a bit trickier. When serialized (for instance to be used via the itools command), they are directly serialized in an array: no dedicated format was defined.

Should we let it like this, at least for now, or should we introduce a proper format? This should be discussed...
Note that it seems that open-loadflow also serialize/deserialize these factors directly in arrays (see for instance [in the `SensitivityFactoryJsonRecored` class](https://github.com/powsybl/powsybl-open-loadflow/blob/ced5a027773fdffb348110c13ba869cd3a1ca372/src/main/java/com/powsybl/openloadflow/sensi/SensitivityFactoryJsonRecorder.java#L36)), so adaptations might be necessary in the simulators.

olperr1 (changes requested):

sensitivity-analysis-api/src/main/java/com/powsybl/sensitivity/SensitivityVariableType.java
gautierbureau: I tried to do it for SensitivityAnalysisResult let me know if it seems ok ! For the OLF part I know that it might need adaptation and will see to it thanks ! 
For SensitivityFactor I wait on you so we can discuss it more.

olperr1: For `SensitivityFactor`, even if we introduce a proper format, it will be a breaking change since the simulators won't know it. I think this is something that must be done to prevent each simulator to handle the (de)serialization and to allow a better management of potential future changes in the factors' data.
But this could be delayed for now.

olperr1 (approved):

PR #3903, opened 2026-05-05
Security analysis - Adding parameters to filter monitored elements results

HugoKulesza:

No

Feature

When running security analysis, if an element is monitored in N and N-1 state, both states are written in the security analysis result even if nothing changed for the monitored element.
When there is a lot of monitored elements / a lot of contingencies, it can produce way too much of not useful memory usage.

There is now a new parameters object (mirroring IncreasedViolationsParameters) that can be used by simulators as thresholds to filter N-1 monitored element results. 
Branches and 3 windings transformers can be filtered on a proportional power threshold.
Buses can be filtered on a proportional or absolute voltage threshold. 

- [X] Yes
- [ ] No

A new version of security analysis parameters files, the 1.3, was introduced. Differences with the previous version are:
- a new `monitored-element-modification-threshold` element was added, containing thresholds configuration allowing to filter monitored elements results.


security-analysis/security-analysis-api/src/main/java/com/powsybl/security/SecurityAnalysisParameters.java
```diff
@@ -30,10 +30,12 @@ public class SecurityAnalysisParameters extends AbstractExtendable<SecurityAnaly
     // VERSION = 1.0
     // VERSION = 1.1 IncreasedViolationsParameters adding.
     // VERSION = 1.2 intermediateResultsInOperatorStrategy
-    public static final String VERSION = "1.2";
+    // VERSION = 1.3 ModifiedMonitoredElementsParameters
+    public static final String VERSION = "1.3";
 
     private LoadFlowParameters loadFlowParameters = new LoadFlowParameters();
     private IncreasedViolationsParameters increasedViolationsParameters = new IncreasedViolationsParameters();
+    private ModifiedMonitoredElementsParameters modifiedMonitoredElementsParameters = new ModifiedMonitoredElementsParameters();
```
alicecaron, line 38: Add documentation about this parameter in `docs/simulation/security/configuration.md`

security-analysis/security-analysis-api/src/test/resources/SecurityAnalysisParametersV1.3.json
```diff
@@ -0,0 +1,35 @@
...
+    "low-voltage-proportional-threshold" : 0.0,
+    "low-voltage-absolute-threshold" : 0.0,
+    "high-voltage-proportional-threshold" : 0.0,
+    "high-voltage-absolute-threshold" : 0.0
+  },
+  "monitored-element-modification-threshold" : {
+    "power-modification-threshold" : 0.0,
+    "voltage-modification-proportional-threshold" : 0.0,
+    "voltage-modification-absolute-threshold" : 0.0
+  },
```
olperr1, line 14: Could you use non-default values in this file to assert that they are indeed correctly serialized/deserialized?

docs/simulation/security/configuration.md
```diff
@@ -55,6 +55,19 @@ Same as before but for high-voltage violations.
...
+
+**power-modification-threshold**<br>
+After a contingency, all the monitored branches and 3 windings transformers whose power has not changed by more than the threshold (in percentage) compared to pre-contingency state are filtered from the post-contingency results. It is set at 0 by default, so it doesn't filter anything.
+
+**voltage-modification-proportional-threshold**<br>
+After a contingency, all the monitored buses whose voltage has not changed by more than the threshold (in percentage) compared to pre-contingency state are filtered from the post-contingency results. It is set at 0 by default, so it doesn't filter anything.
+
+**voltage-modification-absolute-threshold**<br>
+After a contingency, all the monitored buses whose voltage has not changed by more than the threshold (in kV) compared to pre-contingency state are filtered from the post-contingency results. It is set at 0 by default, so it doesn't filter anything.
+
```
olperr1, line 79: Here is a suggestion to clarify the 2 categories and the fact that the 2 parameters on voltage are not exclusive:

```suggestion
(monitoring-modification-thresholds)=
### Monitoring modification thresholds
When monitoring elements both before and after contingencies, it is possible to define thresholds that limit the number of element states stored in the post-contingency state.

#### Thresholds on power
**power-modification-threshold**<br>
After a contingency, all the monitored branches and 3 windings transformers whose power has not changed by more than the threshold (in percentage) compared to pre-contingency state are filtered from the post-contingency results. It is set at 0 by default, so it doesn't filter anything.

#### Thresholds on voltage
These two following parameters are not exclusive. When both are defined, the threshold in kV that will be used will be the minimum between the value of the absolute parameter and the result of the proportional parameter applied to the pre-contingency value.

**voltage-modification-proportional-threshold**<br>
After a contingency, all the monitored buses whose voltage has not changed by more than the threshold (in percentage) compared to pre-contingency state are filtered from the post-contingency results. It is set at 0 by default, so it doesn't filter anything.

**voltage-modification-absolute-threshold**<br>
After a contingency, all the monitored buses whose voltage has not changed by more than the threshold (in kV) compared to pre-contingency state are filtered from the post-contingency results. It is set at 0 by default, so it doesn't filter anything.

```

olperr1 (changes requested):

olperr1 (approved):

PR #3905, opened 2026-05-06
Fix BUS_BRANCH export failure with keepOriginalTopology

### LINKED ISSUE #3851: Avoid error related to a know unsupported topology in bus branch export when using "keep original topology" incompatibility behavior

the export  of a network in a IIDM format with BUS_BRANCH topology level and BUS_BRANCH_VOLTAGE_LEVEL_INCOMPATIBILITY_BEHAVIOR=KEEP_ORIGINAL_TOPOLOGY sometime fails with the following error: 

Exception in thread "main" com.powsybl.commons.PowsyblException: Terminal ref should not point to a busbar section (here <BBS Name>). Try to export in node-breaker or delete this terminal ref.

I would expect the keep original topology option to automatically export the voltage level in node breaker in this case

### Describe the steps

_No response_

### Environment

_No response_

### Relevant Log Output

_No response_

### Extra Information

_No response_

cphili:

https://github.com/powsybl/powsybl-core/issues/3851

Bug fix

When exporting a network in IIDM format with BUS_BRANCH topology level and BUS_BRANCH_VOLTAGE_LEVEL_INCOMPATIBILITY_BEHAVIOR = KEEP_ORIGINAL_TOPOLOGY,
the export may fail in some cases where the chosen topology leads to an invalid IIDM configuration. In particular, when a voltage level originally in NODE_BREAKER topology is exported in BUS_BRANCH, some terminal references may become invalid (e.g. referencing elements that are not representable in BUS_BRANCH), leading to a runtime exception during serialization.

The export logic now consistently relies on checkVoltageLevelExportTopology to detect invalid BUS_BRANCH configurations. When such an incompatibility is detected:  
If the behavior is THROW_EXCEPTION, an exception is still thrown
If the behavior is KEEP_ORIGINAL_TOPOLOGY, the export automatically falls back to the original topology (typically NODE_BREAKER)

- [ ] Yes
- [X] No

None. The change is backward-compatible and only affects cases that previously resulted in invalid exports or runtime exceptions.  

**Other information**:


iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/TopologyLevelTest.java
```diff
@@ -52,4 +56,124 @@ private void testConversion(Network network) throws IOException {
         testWriteVersionedXml(network, options.setTopologyLevel(TopologyLevel.BUS_BREAKER), "fictitiousSwitchRef-bbk.xml", CURRENT_IIDM_VERSION);
         testWriteVersionedXml(network, options.setTopologyLevel(TopologyLevel.BUS_BRANCH), "fictitiousSwitchRef-bbr.xml", CURRENT_IIDM_VERSION);
     }
+
```
olperr1: These tests succeed also directly on main (without the `TopologyLevelUtil` changes). Therefore none of them covers the new case.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/util/TopologyLevelUtil.java
```diff
@@ -26,13 +26,24 @@ private TopologyLevelUtil() {
     public static TopologyLevel determineTopologyLevel(VoltageLevel vl, NetworkSerializerContext context) {
         TopologyLevel exportTopologyLevel = context.getVoltageLevelExportTopologyLevel(vl.getId());
         if (exportTopologyLevel == null) {
-            TopologyLevel configTopologyLevel = Objects.requireNonNullElse(context.getOptions().getVoltageLevelTopologyLevel(vl.getId()), context.getOptions().getTopologyLevel());
-            exportTopologyLevel = checkVoltageLevelExportTopology(vl, context, TopologyLevel.min(vl.getTopologyKind(), configTopologyLevel));
+            if (context.getOptions().getBusBranchVoltageLevelIncompatibilityBehavior()
+                    == ExportOptions.BusBranchVoltageLevelIncompatibilityBehavior.KEEP_ORIGINAL_TOPOLOGY && requiresNodeBreakerExport(vl)) {
+                exportTopologyLevel = TopologyLevel.NODE_BREAKER;
```
olperr1: This is not the right place to add your new check. It should be added to `checkVoltageLevelExportTopology`.

You can change the first condition of this method to look like:
```java
if (topologyLevel == TopologyLevel.BUS_BRANCH && nodeBreakerWithBusBarSection(vl)) {
```

With 
```java
private static boolean withNullConnectableBus(VoltageLevel vl) {
    return vl.getConnectableStream()
        .map(Connectable::getTerminals)
        .flatMap(List<Terminal>::stream)
        .anyMatch(t -> t.getBusView().getConnectableBus() == null);
    }
```

Then you can update it with your additional condition:
```java
if (topologyLevel == TopologyLevel.BUS_BRANCH
    && (nodeBreakerWithBusBarSection(vl) || requiresNodeBreakerExport(vl)) {
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/util/TopologyLevelUtil.java
```diff
@@ -26,13 +26,24 @@ private TopologyLevelUtil() {
...
+            } else {
+                TopologyLevel configTopologyLevel = Objects.requireNonNullElse(context.getOptions().getVoltageLevelTopologyLevel(vl.getId()), context.getOptions().getTopologyLevel());
+                exportTopologyLevel = checkVoltageLevelExportTopology(vl, context, TopologyLevel.min(vl.getTopologyKind(), configTopologyLevel));
+            }
             context.addVoltageLevelExportTopologyLevel(vl.getId(), exportTopologyLevel);
         }
         return exportTopologyLevel;
     }
 
+    private static boolean requiresNodeBreakerExport(VoltageLevel vl) {
```
olperr1: This is not a suitable name. There are other cases when we need to export in node/breaker.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/util/TopologyLevelUtil.java
```diff
@@ -26,13 +26,24 @@ private TopologyLevelUtil() {
...
+        return vl.getTopologyKind() == TopologyKind.NODE_BREAKER
+                && vl.getConnectableStream()
+                .anyMatch(BusbarSection.class::isInstance);
```
olperr1: The exception mentioned by the issue indicates that there's a terminalRef pointing to a busbar section.
If the VL is in node/breaker and has busbar section but no terminalRef pointing to them, the export is possible in bus/branch.

For instance, [`fictitiousSwitchRef.xml`](https://github.com/powsybl/powsybl-core/blob/155ab7ded83551ed9766c5e525350f35ec990de5/iidm/iidm-serde/src/test/resources/V1_17/fictitiousSwitchRef.xml) has a VL in node/breaker with a busbar section (vl="C", busbarSection="D") and can be exported in bus/branch (see [`fictitiousSwitchRef-bbr.xml`](https://github.com/powsybl/powsybl-core/blob/155ab7ded83551ed9766c5e525350f35ec990de5/iidm/iidm-serde/src/test/resources/V1_17/fictitiousSwitchRef-bbr.xml)). Since no terminalRef references "D", there's no problem.

olperr1: Note that the example is covered by the `testConversion()` unit test (by last line of `testConversion(Network network)`).
With this PR current code, the test succeed because the behavior is not set to `KEEP_ORIGINAL_TOPOLOGY`, then `else` branch is executed. since no error was detected with the previous check, the VL is exported in bus/branch as expected by the unit test.
This demonstrates that `requiresNodeBreakerExport(VoltageLevel vl)` detects as problematic situations that are not.

olperr1 (changes requested):

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/util/TopologyLevelUtil.java
cphili: The issue cannot be easily fixed in TopologyLevelUtil, since the error (a terminalRef linked to a busbar) cannot be detected at that stage. To properly fall back to exporting the original topology (NODE_BREAKER), these cases would need to be identified before serialization through a network pre-processing step, which adds complexity.
As an alternative, we can ensure that the export process does not fail, even if such inconsistencies are present.

olperr1: The alternative is not viable: the export does not fail, but the exported file cannot be imported.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/util/TopologyLevelUtil.java
```diff
@@ -33,6 +35,49 @@ public static TopologyLevel determineTopologyLevel(VoltageLevel vl, NetworkSeria
...
     }
 
+    private static boolean withNullConnectableBus(VoltageLevel vl) {
+        return vl.getConnectableStream()
+                .map(Connectable::getTerminals)
+                .flatMap(List<Terminal>::stream)
+                .anyMatch(t -> t.getBusView().getConnectableBus() == null);
+    }
+
+    private static boolean isBusbarTerminalRefIssue(VoltageLevel vl) {
```
olperr1: I have **not** tested it, but I think you can simply check if there is at least one busbar section referenced with this method:
```java
    private static boolean isBusbarTerminalRefIssue(VoltageLevel vl) {
        if (vl.getTopologyKind() != TopologyKind.NODE_BREAKER) {
            return false;
        }
        return vl.getNodeBreakerView().getBusbarSectionStream()
                .anyMatch(bbs -> !bbs.getTerminal().getReferrers().isEmpty());
    }
```

Note that it would be better to have a more descriptive name for the method. Maybe something like "hasReferencedBusbarSections"?

olperr1 (changes requested):

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/util/TopologyLevelUtil.java
cphili: Thanks for the suggestion.
I tried this approach and added a test, but it fails: getReferrers() is still empty in this scenario.
This seems to indicate that the terminalRef relationships are not populated yet and this method cannot reliably detect the issue at this stage.

olperr1: The problem you encountered is that your unit test network is not problematic. There's no terminalRef pointing to the busbar section.

If you want to check this, you just have to checkout `main` and try to export your network in bus/branch then import it. Without any changes, this test should fail at the export with the error message described in the issue (` com.powsybl.commons.PowsyblException: Terminal ref should not point to a busbar section (here ...). Try to export in node-breaker or delete this terminal ref.`). **This is not the case.**

<br/>

Here is a unit test that check the problem:
```java
    @Test
    void shouldKeepOriginalTopologyWhenDetectingReferenceToBusbarTerminal() {
        Network network = Network.create("n1", "test");
        Substation substation = network.newSubstation().setId("S1").add();
        VoltageLevel vl = substation.newVoltageLevel()
                .setId("VL1")
                .setNominalV(225.0)
                .setTopologyKind(TopologyKind.NODE_BREAKER)
                .add();

        vl.getNodeBreakerView().newBusbarSection().setId("BBS1").setNode(0).add();
        BusbarSection bbs2 = vl.getNodeBreakerView().newBusbarSection()
                .setId("BBS2")
                .setNode(1)
                .add();

        vl.newGenerator()
                .setId("GEN1")
                .setNode(2)
                .setRegulatingTerminal(bbs2.getTerminal())
                .setMinP(-9999.99)
                .setMaxP(9999.99)
                .setVoltageRegulatorOn(true)
                .setTargetV(24.5)
                .setTargetP(607.0)
                .setTargetQ(301.0)
                .add();

        vl.getNodeBreakerView().newSwitch()
                .setId("SW1")
                .setNode1(0)
                .setNode2(1)
                .setKind(SwitchKind.BREAKER)
                .setOpen(true)
                .add();

        vl.getNodeBreakerView().newSwitch()
                .setId("SW2")
                .setNode1(0)
                .setNode2(2)
                .setKind(SwitchKind.BREAKER)
                .setOpen(false)
                .add();

        ExportOptions options = new ExportOptions();
        options.setTopologyLevel(TopologyLevel.BUS_BRANCH);
        options.setBusBranchVoltageLevelIncompatibilityBehavior(ExportOptions.BusBranchVoltageLevelIncompatibilityBehavior.KEEP_ORIGINAL_TOPOLOGY);
        Path file = tmpDir.resolve("data");

        NetworkSerDe.write(network, options, file);
        assertDoesNotThrow(() -> NetworkSerDe.read(file));
    }
```
If you try to execute it on `main`, you will have the following error:
`com.powsybl.commons.PowsyblException: Terminal ref should not point to a busbar section (here BBS2). Try to export in node-breaker or delete this terminal ref.`

Here, the sensible point is the `setRegulatingTerminal(...)` using the busbar section terminal.


Could you add this unit test **in `BusBranchExportTest`**, which is the place where test on the bus/branch export are centralized?

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/TopologyLevelTest.java
olperr1, line 1: Could you revert these changes?

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/TerminalRefSerDe.java
```diff
@@ -76,10 +76,17 @@ private static void checkTerminal(Terminal t, NetworkSerializerContext context)
...
                 && t.getConnectable() instanceof BusbarSection) {
-            throw new PowsyblException(String.format("Terminal ref should not point to a busbar section (here %s). Try to export in node-breaker or delete this terminal ref.",
+
+            if (context.getOptions().getBusBranchVoltageLevelIncompatibilityBehavior() == ExportOptions.BusBranchVoltageLevelIncompatibilityBehavior.KEEP_ORIGINAL_TOPOLOGY) {
+                return;
+            }
+
+            throw new PowsyblException(String.format(
+                    "Terminal ref should not point to a busbar section (here %s). Try to export in node-breaker or delete this terminal ref.",
                     t.getConnectable().getId()));
```
olperr1, line 83: These changes are no more needed and must be reverted.
The call to `TopologyLevelUtil.determineTopologyLevel` now returns `NODE_BREAKER` so we don't enter in this block in the case reported by the issue.

olperr1 (changes requested):

iidm/iidm-serde/src/test/resources/V1_17/voltageLevelFallbackNodeBreaker.xml
olperr1, line 1: Sorry, I missed this file. It is not needed anymore too.

olperr1 (changes requested):

iidm/iidm-serde/src/test/resources/V1_17/voltageLevelFallbackNodeBreaker.xml
cphili: Thanks, good catch! It's now fixed.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/BusBranchExportTest.java
```diff
@@ -55,4 +54,74 @@ void roundTripTestOK() {
...
+        options.setTopologyLevel(TopologyLevel.BUS_BRANCH);
+        options.setBusBranchVoltageLevelIncompatibilityBehavior(ExportOptions.BusBranchVoltageLevelIncompatibilityBehavior.KEEP_ORIGINAL_TOPOLOGY);
+        Path file = tmpDir.resolve("data");
+
+        NetworkSerDe.write(network, options, file);
+        assertDoesNotThrow(() -> NetworkSerDe.read(file));
+    }
+
+    @Test
+    void shouldReturnFalseWhenTopologyIsBusBranch() {
```
olperr1:
```suggestion
    void shouldReturnFalseWhenTopologyIsBusBreaker() {
```

olperr1 (changes requested):

olperr1 (approved):

PR #3906, opened 2026-05-06
Fix bus voltage and angle stability

### LINKED ISSUE #2509: Stability of bus voltage and angle
Bug

Voltage and angle of bus taken from first terminal encountered

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
On a node breaker network, after computing a loadflow, with powsybl-iidm-impl:
- Add a load to an existing bus of the bus view
   - either with a low node number (lower than the lowest node number), then the terminal ref is the terminal of the new load, hence the bus voltage and angle are NaN
   - or with a high node number (higher than the lowest node number), then the terminal ref remains the same terminal, hence the bus voltage and angle are defined

The bus voltage and angle should be the same in both cases.

**What is the motivation / use case for changing the behavior?**
Stable V and angle, not based on traversing order

**Other information**
Unit test:
```java
Network network = FictitiousSwitchFactory.create();
LoadFlow.run(network);

VoltageLevel vlN = network.getVoltageLevel("N");
Terminal.BusView loadBusView = network.getLoad("CE").getTerminal().getBusView(); // bus containing nodes 3, 4, 5

// loadBusView.getBus().getV() now is equal to 234kV
vlN.getNodeBreakerView().newInternalConnection().setNode1(3).setNode2(23).add();
vlN.newLoad().setId("newLoad").setP0(1).setQ0(1).setNode(23).add();
// loadBusView.getBus().getV() remains equal to 234kV

vlN.newLoad().setId("newLoad").setP0(1).setQ0(1).setNode(3).add();
// loadBusView.getBus().getV() is now equal to NaN
```

samirromdhani:

Fixes #2509 

- [ ] Yes
- [x] No

**Other information**:


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/CalculatedBusImpl.java
```diff
@@ -39,22 +39,30 @@ class CalculatedBusImpl extends AbstractBus implements CalculatedBus {
...
+            return terminals.stream()
+                    .filter(CalculatedBusImpl::hasDefinedVoltageOrAngle)
+                    .findFirst()
+                    .orElse(terminals.getFirst());
         }
         return (NodeTerminal) Networks.getEquivalentTerminal(voltageLevel, nodes.getQuick(0));
     }
 
+    private static boolean hasDefinedVoltageOrAngle(NodeTerminal t) {
+        return !Double.isNaN(t.getV()) || !Double.isNaN(t.getAngle());
```
rolnico, line 64: @olperr1 should we check that both values are not NaN or having only one is enough?

PR #3923, opened 2026-05-19
optimize Network#getConnectables(Class), avoid full scan

### LINKED ISSUE #1413: Network.getConnectables() doesn't work as expected
Bug

We recently introduce methods to get the connectables from the network. There is a method to get connectables from a certain type, but it doesn't work properly with intermediate classes.

The call to `network.getConnectables(Injection.class)` returns an empty set, whereas we would like to get all injections from the network.

The method `getConnectables(Class<?> clazz)` should be fixed to work with all valid classes. An optimization could be to fix the `NetworkIndex.getAll` method:
```
    <T extends Identifiable> Set<T> getAll(Class<T> clazz) {
        Set<Identifiable<?>> all = objectsByClass.get(clazz);
        if (all == null) {
            return Collections.emptySet();
        }
        return (Set<T>) all;
    }
```
Instead of returning an empty set, we could iterate over the keys of `objectsByClass`, testing if the given class is a base class of the key, and concatenate the sets.

Get efficiently the injections of a network

  - PowSyBl Version: ...
  - OS Version: ...

samirromdhani:

Fixes #1413

- [ ] Yes
- [ ] No

**Other information**:

- Network connectables Benchmark here : [powsybl-benchmark / TEST_CASE_CORE_1413.md](https://github.com/powsybl/powsybl-benchmark/blob/test/benchmark_core_1413_networkgetconnectables_implementation/src/main/java/com/powsybl/benchmark/iidm/TEST_CASE_CORE_1413.md#test-case-core-1413)

### DIFF
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
@@ -954,7 +954,7 @@ public class NetworkImpl extends AbstractNetwork implements VariantManagerHolder
 
     @Override
     public <C extends Connectable> Stream<C> getConnectableStream(Class<C> clazz) {
-        return index.getAll().stream().filter(clazz::isInstance).map(clazz::cast);
+        return index.streamAssignable(clazz);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


samirromdhani:
blocked by #3927

PR #3836, opened 2026-03-19
feat: add observability area extension on voltage levels 

TheMaskedTurtle:
**Please check if the PR fulfills these requirements**

- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] A PR or issue has been opened in all impacted repositories (if any)

**What kind of change does this PR introduce?**

This PR introduces a new extension on voltage level that describes observability on network equipments.

**What is the current behavior?**

**What is the new behavior (if this is a feature change)?**

**Does this PR introduce a breaking change or deprecate an API?**
- [ ] Yes
- [x] No

PR #3928, opened 2026-05-22
Add setId to identifiable

### LINKED ISSUE #3844: Be able to update an Identifiable id

An `Identifiable` id cannot be updated.

An `Identifiable` id can be updated.

### Describe the motivation

For some use cases, for instance to follow a naming strategy, when a voltage level structure is changed, some identifiable id should be updated to follow a given naming strategy. For that we need a function to be able to update its id. 

It is needed for any `Identifiable` class.

We can also think about a utility method that could be called to clean the naming of the identifiables according to a given naming strategy (or the default one). For instance to be called after a voltage level topology is completely changed.

- at least add a `setId` method (this behavior is already implemented in BoundaryLineImpl - see `replaceId`)
- add TCK tests for the custom IIDM implementations (such as powsybl-network-store)

### Extra Information

A `AbstractIdentifiable::replaceId` method already exists but is only implemented in `BoundaryLineImpl`.
We can replace this method by:
- renaming it to `setId`
- move it to the Identifiable interface
- implement it in the `AbstractIdentifiable` class

Need to add TCK tests to be sure to implement it also in the network-store or other implementation.

obrix:

Yes https://github.com/powsybl/powsybl-core/issues/3844

Add a setId function to Identifiable

No setId available

Id can be updated through setId

- [x] Yes
- [ ] No

In the IIDM in-memory implementation (provided by `iidm-impl`), the `BoundaryLine#replaceId(String newId)` method was removed. You can use the new `Identifiable#setId(String id)` instead.

The `DefaultNetworkListener` abstract class was removed; its empty implementation methods are now defined as default methods in the `NetworkListener` interface. If one of your classes extended `DefaultNetworkListener`, you just have to make it implement `NetworkListener` instead.

Here are some examples of adaptations:
```patch
- public class MyListener extends DefaultNetworkListener {
+ public class MyListener implement NetworkListener {
```

```patch
- NetworkListener listener = new DefaultNetworkListener() {
+ NetworkListener listener = new NetworkListener() {
      @Override
      public void onCreation(Identifiable identifiable) {
          listenerCalled.setTrue();
      }
}
```

```patch
- Mockito.mock(DefaultNetworkListener.class);
+ Mockito.mock(NetworkListener.class);
```

<br>

Custom IIDM implementations' maintainers should implement the following method:
- `Identifiable#setId(String id)`

**Other information**:


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiable.java
```diff
@@ -71,6 +67,20 @@ public I setName(String name) {
         return (I) this;
     }
 
+    @Override
+    public I setId(String id) {
+        NetworkIndex.checkId(id);
```
olperr1: Before any other operation, you can directly exit when the new ID is equal to the old one.

olperr1 (changes requested):

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractIdentifiableTest.java
```diff
@@ -0,0 +1,39 @@
...
+        assertEquals("LOAD", load.getId());
+        load.setId("NEW_LOAD_ID");
+        assertEquals(load.getId(), network.getIdentifiable("NEW_LOAD_ID").getId());
+        assertEquals("NEW_LOAD_ID", load.getId());
+
+        Generator gen = network.getGenerator("GEN");
+        assertThrows(PowsyblException.class, () -> gen.setId("NEW_LOAD_ID"), "Object with id (NEW_LOAD_ID) already exists");
+
+        load.setId("NEW_LOAD_ID");
+        assertEquals("NEW_LOAD_ID", load.getId());
```
olperr1, line 41: Can you also check that no identifiable exists with the old ID?

olperr1 (changes requested):

olperr1:
Changing the identifiers may cause inconsistencies with some extensions, if element ids are stored in the extension at its creation instead of references to the elements.

I have spotted at least one case in the powsybl-core extensions:
- in the `SecondaryVoltageControl` extension
  - interface `PilotPoint`: `busbarSectionsOrBusesIds`
  - interface `ControlUnit`: `id` (it is generally the element id, but it could be something else, I think)

(And maybe in the `ConnectablePosition` extension.)

#### Potential solutions
- fix the `SecondaryVoltageControlExtension` to use references + write guidelines to clearly indicate that no id should be directly stored in the extensions
- create a callback mechanism to change the content of the extension storing an element id when this id changes

In both cases, downstream projects will have to review all their private extensions and adapt the one storing ids.

olperr1:
There's also the extension introduced by #3836 (it contains busIds).

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/ControlUnitImpl.java
```diff
@@ -42,6 +42,10 @@ protected int getVariantIndex() {
         return controlZone.getSecondaryVoltageControl().getVariantManagerHolder().getVariantIndex();
     }
 
+    public void setId(String newId) {
```
olperr1: It would be safer to set this method protected:

```suggestion
    protected void setId(String newId) {
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/ControlZoneImpl.java
```diff
@@ -58,6 +58,10 @@ public List<ControlUnit> getControlUnits() {
         return Collections.unmodifiableList(controlUnits);
     }
 
+    public void setControlUnits(List<ControlUnit> newControlUnits) {
```
olperr1:
```suggestion
    protected void setControlUnits(List<ControlUnit> newControlUnits) {
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/PilotPointImpl.java
```diff
@@ -53,6 +53,10 @@ public List<String> getBusbarSectionsOrBusesIds() {
         return Collections.unmodifiableList(busbarSectionsOrBusesIds);
     }
 
+    public void setBusbarSectionsOrBusesIds(List<String> newBusbarSectionsOrBusesIds) {
```
olperr1:
```suggestion
    protected void setBusbarSectionsOrBusesIds(List<String> newBusbarSectionsOrBusesIds) {
```

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/extensions/AbstractSecondaryVoltageControlTest.java
```diff
@@ -27,9 +27,9 @@
  */
 public abstract class AbstractSecondaryVoltageControlTest {
 
-    private Network network;
+    public Network network;
 
-    private SecondaryVoltageControl control;
+    public SecondaryVoltageControl control;
```
olperr1: It doesn't seem that these changes are necessary. Could you revert them?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DefaultNetworkListener.java
```diff
@@ -12,74 +12,74 @@
 /**
  * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
  */
-public class DefaultNetworkListener implements NetworkListener {
+public interface DefaultNetworkListener extends NetworkListener {
```
olperr1: I think it is better to keep the existing signature to prevent an unnecessary breaking change.
You can create a new listener class (which can be an anonymous one) instead of defining the whole `SecondaryVoltageControlImpl` extension as a listener.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/extensions/SecondaryVoltageControlImpl.java
```diff
@@ -65,4 +66,36 @@ public void allocateVariantArrayElement(int[] indexes, int sourceIndex) {
             ((ControlZoneImpl) controlZone).allocateVariantArrayElement(indexes, sourceIndex);
         }
     }
+
+    @Override
+    public void afterRemoval(String identifiable) {
```
olperr1: In this method and in the updating one, you retrieve the list, generate another one with the modification, then set it. Since this is not atomic, this may cause problems in multi-threading context.

I think the solution can be to create methods in `ControlZoneImpl`, `PilotPointImpl`, ... that will directly alter the inner collection with methods such as `List.replaceAll(UnaryOperator<E>)` or `List.remove(Object)`, or that will call a concurrent-safe method.

olperr1 (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DefaultNetworkListener.java
olperr1: After a second thought (and a discussion with @obrix), this class/interface isn't really useful.
The default implementation methods could be moved to `NetworkListener`, and `DefaultNetworkListener` could be removed.

olperr1 (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/NetworkListener.java
```diff
@@ -14,31 +14,59 @@
...
+        // empty default implementation
+    }
 
-    void beforeRemoval(Identifiable<?> identifiable);
+    default void beforeRemoval(Identifiable<?> identifiable) {
+        // empty default implementation
+    }
 
-    void afterRemoval(String id);
+    default void afterRemoval(String identifiable) {
```
olperr1: The previous parameter's name in the interface was more suitable than the one in the abstract class:

```suggestion
    default void afterRemoval(String id) {
```

olperr1 (changes requested):

olperr1 (approved):

PR #3930, opened 2026-05-27
Report messages indicating if there are remaining branches when deleting a VL on revert modification

EtienneLt:

No

Feature: more detailed reports

When deleting a voltage level for a revert modification, the report message doesn't indicate if there is still branches on the voltage level.

When deleting a voltage level for a revert modification, the report message is different whether there is still branches on the voltage level or not.

- [x] Yes
- [ ] No

The `voltageLevelRemovingEquipmentsLeftReport(...)` method of `ModificationReports` was replaced by the followings:
- `voltageLevelNotRemovedWithRemainingBranches(...)`, that should be called when branches are still present on the voltage level to remove;
- `voltageLevelNotRemovedWithNoBranch(...)`, that should be called when there are no remaining branches on the voltage level to remove.

Also, in the report property files, the key `core.iidm.modification.voltageLevelRemovingEquipmentsLeft` was replaced by the followings:
- `core.iidm.modification.voltageLevelNotRemovedWithRemainingBranches`, used when branches are still present;
- `core.iidm.modification.voltageLevelNotRemovedWithNoBranch`, used when there are no remaining branches.


iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/TopologyModificationUtils.java
```diff
@@ -202,7 +202,14 @@ static void removeVoltageLevelAndSubstation(VoltageLevel voltageLevel, ReportNod
         String vlId = voltageLevel.getId();
         boolean noMoreEquipments = voltageLevel.getConnectableStream().noneMatch(c -> c.getType() != IdentifiableType.BUSBAR_SECTION);
         if (!noMoreEquipments) {
-            voltageLevelRemovingEquipmentsLeftReport(reportNode, vlId);
+            boolean noMoreBranch = voltageLevel.getConnectableStream().noneMatch(c -> c.getType() == IdentifiableType.LINE
+                    || c.getType() == IdentifiableType.TWO_WINDINGS_TRANSFORMER || c.getType() == IdentifiableType.THREE_WINDINGS_TRANSFORMER ||
+                    c.getType() == IdentifiableType.HVDC_CONVERTER_STATION);
```
NathanDissoubray: how about something like:
```java
c -> switch(c.getType()) {
   case LINE, TWO_WINDINGS_TRANSFORMER, THREE_WINDINGS_TRANSORMER, HVDC_CONVERTER_STATION -> true;
   default -> false;
}
```
to avoid repetition ?

iidm/iidm-modification/src/test/resources/reportNode/revert-create-line-on-line-keeping-vl-with-branches.txt
```diff
@@ -0,0 +1,7 @@
++ Testing reportNode for reverting create line on line and not removing the voltage level
+   Line CJ_1 removed
+   Line CJ_2 removed
+   Line testLine removed
+   Line CJ_NEW created
+   Voltage level CJ_VL still contains equipments including branches, it is not removed.
+   Voltage level VLTEST still contains equipments including branches, it is not removed.
```
NathanDissoubray: is doing a partial removal what we want to happen ? if the goal is just to remove a teePoint, why is it possible to delete voltage levels ?

commons/src/main/resources/com/powsybl/commons/reports.properties
```diff
@@ -135,7 +135,8 @@ core.iidm.modification.voltageConnectedOnLine = Voltage level ${voltageLevelId}
 core.iidm.modification.voltageLevelCreated = VoltageLevel ${voltageLevelId} created
 core.iidm.modification.voltageLevelNotFound = Voltage level ${voltageLevelId} is not found
 core.iidm.modification.voltageLevelRemoved = Voltage level ${vlId} removed
-core.iidm.modification.voltageLevelRemovingEquipmentsLeft = Voltage level ${vlId} still contains equipments and it is not removed.
+core.iidm.modification.voltageLevelNotRemovedWithRemainingBranches = Voltage level ${vlId} still contains equipments including branches, it is not removed.
+core.iidm.modification.voltageLevelNotRemovedWithNoBranch = Voltage level ${vlId} still contains equipments but no branch, it is not removed.
```
NathanDissoubray:
```suggestion
core.iidm.modification.voltageLevelNotRemovedWithNoBranch = Voltage level ${vlId} contains no branches but it still contains equipment, it is not removed.
```

commons/src/main/resources/com/powsybl/commons/reports.properties
```diff
@@ -135,7 +135,8 @@ core.iidm.modification.voltageConnectedOnLine = Voltage level ${voltageLevelId}
 core.iidm.modification.voltageLevelCreated = VoltageLevel ${voltageLevelId} created
 core.iidm.modification.voltageLevelNotFound = Voltage level ${voltageLevelId} is not found
 core.iidm.modification.voltageLevelRemoved = Voltage level ${vlId} removed
-core.iidm.modification.voltageLevelRemovingEquipmentsLeft = Voltage level ${vlId} still contains equipments and it is not removed.
+core.iidm.modification.voltageLevelNotRemovedWithRemainingBranches = Voltage level ${vlId} still contains equipments including branches, it is not removed.
```
NathanDissoubray:
```suggestion
core.iidm.modification.voltageLevelNotRemovedWithRemainingBranches = Voltage level ${vlId} still contains equipment including branches, it is not removed.
```

commons/src/main/resources/com/powsybl/commons/reports_fr.properties
```diff
@@ -135,7 +135,8 @@ core.iidm.modification.voltageConnectedOnLine = Le poste ${voltageLevelId} conne
 core.iidm.modification.voltageLevelCreated = Poste ${voltageLevelId} créé.
 core.iidm.modification.voltageLevelNotFound = Poste ${voltageLevelId} introuvable.
 core.iidm.modification.voltageLevelRemoved = Le poste ${vlId} a été supprimé.
-core.iidm.modification.voltageLevelRemovingEquipmentsLeft = Le poste ${vlId} contient toujours des équipements, il n'est donc pas supprimé.
+core.iidm.modification.voltageLevelNotRemovedWithRemainingBranches = Le poste ${vlId} contient toujours des équipements dont des quadripôles, il n'est pas supprimé.
+core.iidm.modification.voltageLevelNotRemovedWithNoBranch =Le poste ${vlId} contient toujours des équipements mais aucun quadripôles, il n'est pas supprimé.
```
NathanDissoubray:
```suggestion
core.iidm.modification.voltageLevelNotRemovedWithNoBranch = Le poste ${vlId} ne contient aucun quadripôles mais il reste toujours des équipements, il n'est pas supprimé.
```

NathanDissoubray (changes requested):

commons/src/main/resources/com/powsybl/commons/reports.properties
NathanDissoubray: equipment is not countable, no "s" should be used. The plural form should be used after "no" (thus branches)
I also think it's clearer to formulate it with a "and", ending with the "but no branches" makes it seem like we are not deleting the voltage level because there are no branches

NathanDissoubray:
Could you also update the migration guide to explain which methods should be used in place of the deleted method, and in which cases ?

iidm/iidm-modification/src/test/resources/reportNode/revert-create-line-on-line-keeping-vl-with-branches.txt
EtienneLt: image
It was decided to delete the attachment point and the voltage level if they contain nothing

commons/src/main/resources/com/powsybl/commons/reports_fr.properties
EtienneLt: I prefer "Le poste ${vlId} contient encore des équipements mais plus aucun quadripôle, il n'est pas supprimé"

NathanDissoubray: "et plus aucun quadripôle" ?
the "mais" makes it feel like it's because of this that we don't delete (but it's because there are still equipment left). It should also have the same meaning as the english report node

EtienneLt: good point

NathanDissoubray (approved):

olperr1 (approved):

PR #3937, opened 2026-06-03
Add a method to remove all loadflow output values

HugoKulesza:

No

Feature

There is now a method that can remove from a network all the loadflow output information.
It can be useful to invalidate loadflow results, or lighten an export.

- [ ] Yes
- [x] No

**Additional information**
To remove all loadflow output values on a network, you can call `Networks.unsetSolvedValues(network);`.


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ThreeWindingsTransformer.java
```diff
@@ -469,4 +469,9 @@ default void setRatioTapPositionToSolvedTapPosition() {
     default void setPhaseTapPositionToSolvedTapPosition() {
         this.getLegStream().forEach(leg -> leg.getOptionalPhaseTapChanger().ifPresent(TapChanger::applySolvedValues));
     }
+
+    default void unsetSolvedTapChangerValues() {
+        this.getLegStream().forEach(leg -> leg.getOptionalRatioTapChanger().ifPresent(TapChanger::unsetSolvedTapPosition));
+        this.getLegStream().forEach(leg -> leg.getOptionalPhaseTapChanger().ifPresent(TapChanger::unsetSolvedTapPosition));
```
olperr1:
```suggestion
        this.getLegStream().forEach(leg -> {
            leg.getOptionalRatioTapChanger().ifPresent(TapChanger::unsetSolvedTapPosition);
            leg.getOptionalPhaseTapChanger().ifPresent(TapChanger::unsetSolvedTapPosition);
        });
```

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcConnectable.java
```diff
@@ -25,4 +25,11 @@ public interface DcConnectable<I extends DcConnectable<I>> extends Identifiable<
      * Remove the DC connectable from the network.
      */
     void remove();
+
+    /**
+     * Removes P and I obtained by a computation on each DC terminal of the connectable.
+     */
+    default void unsetDcSolvedValues() {
```
olperr1: `unsetDcSolvedValues` -> `unsetSolvedValues`:
There's no need to specify the DC here. Also, in DcTerminal the method is called `unsetSolvedValues`, without the "Dc" part.

olperr1 (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcConnectable.java
HugoKulesza: I had to do this because of AcDcConverter that is both a Connectable and a DcConnectable, and it was an issue that it inherited both "unsetSolvedValues" default methods from its interfaces. 
Is there another way to bypass this issue ?

olperr1: You can define this default method in `AcDcConverter`:
```java
    default void unsetSolvedValues() {
        Connectable.super.unsetSolvedValues();
        DcConnectable.super.unsetSolvedValues();
    }
```

But since objects implementing `AcDcConverter` are also `Connectable` and `DcConnectable`, you should slightly adapt in `Networks.unsetSolvedValues` to prevent calling this method twice on them.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcConnectable.java
```diff
@@ -25,4 +25,11 @@ public interface DcConnectable<I extends DcConnectable<I>> extends Identifiable<
      * Remove the DC connectable from the network.
      */
     void remove();
+
+    /**
+     * Removes P and I obtained by a computation on each DC terminal of the connectable.
```
olperr1: Could you adapt the comment to be more generic?
For instance:

```suggestion
     * Removes all loadflow output values of this DC connectable (e.g., P and I on DC terminals).
```

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Connectable.java
```diff
@@ -81,4 +81,11 @@ default boolean disconnect(Predicate<Switch> isSwitchOpenable) {
      * @return true if the disconnection succeeded, false otherwise (the connectable on the given <code>side</code> could not be disconnected, or it was already disconnected)
      */
     boolean disconnect(Predicate<Switch> isSwitchOpenable, ThreeSides side);
+
+    /**
+     * Removes all loadflow output values of this AC connectable (e.g., P and Q on terminals).
```
olperr1:
```suggestion
     * Removes all loadflow output values of this connectable (e.g., P and Q on terminals).
```

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcConnectable.java
olperr1: I think having a single "unsetSolvedValues" is better for AcDcConverter instead of having 2. It doesn't really make sense to unset only the AC or the DC solved values on these elements.
(And calling the method twice on the AC/DC converters is maybe more performant than checking the type of all the DcConnectables.)

olperr1 (changes requested):

olperr1 (approved):

PR #3938, opened 2026-06-05
Add option to import/export only the active operational limits groups

### LINKED ISSUE #3911: Add option to import/export only currently active operational limits groups

Be able to chose (with an option) when importing or exporting an IIDM network, if we want all the operational limits groups or only the active ones.

Currently when importing or exporting an IIDM network, all the operational limits groups are always imported/exported.

### Describe the motivation

Lighten the file at export, and import only useful operational limits groups(the active ones) 

### Extra Information

_No response_

NathanDissoubray:

Closes #3911

Allow to import / export only the active operational limits groups instead of all the groups.
This will reduce the size of serialized networks for the export, and of in memory network for the import if used.

All groups are exported.

Using `new ImportOptions().setOnlySelectedOperationalLimitsGroups(true)` (or the same with the export options), and giving this to NetworkSerDe, only the active groups will be read / exported.

- [x] Yes
- [ ] No

In `iidm-api`, the `getAllSelectedOperationalLimitsGroups` methods of `Branch` and `FlowsLimitsHolder` now return a `List<OperationalLimitsGroup>` instead of a `Collection<OperationalLimitsGroup>`. A stable ordering when adding or deleting selected group is now required for these methods, and enforced by a TCK test.

Inside `iidm-serde`, all the changed methods are package-protected (no public method was changed). Therefore those changes only concern you if you are working directly inside `iidm-serde`.
In `ConnectableSerDeUtil`:
- `readLoadingLimitsGroup(Function<String, OperationalLimitsGroup> groupBuilder, String groupElementName, NetworkDeserializerContext context)` became `readLoadingLimitsGroup(Branch<?> branch, TwoSides side, String groupElementName, NetworkDeserializerContext context)`. The `groupBuilder` generally requires both the branch and the side, so pass that directly to the function instead of creating the builder.
- likewise, `readLoadingLimitsGroups(FlowsLimitsHolder h, String groupElementName, NetworkDeserializerContext context)` became `readLoadingLimitsGroups(FlowsLimitsHolder h, String identifiableId, ThreeSides side, String groupElementName, NetworkDeserializerContext context)`. Directly pass the `FlowsLimitsHolder`. For the side, if this is a tie-line or a boundary line, the side should be ONE. For a three-winding transformer, use the number of the leg on the transformer as the side.
- `readAllSelectedGroupIds(Integer index, Consumer<Collection<String>> consumer, NetworkDeserializerContext context)` was renamed to `readAndGetAllSelectedGroupIds`, it now returns the collection of the selected Ids (it returned nothing before).

**Other information**:


iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ExportOptions.java
```diff
@@ -51,6 +51,8 @@ public enum BusBranchVoltageLevelIncompatibilityBehavior {
 
     private boolean flatten = false;
 
+    private boolean onlySelectedOperationalLimitsGroups = false;
```
samirromdhani: It can take place in `AbstractOptions` as it is shared by both export and import options.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/OperationalLimitsGroupsImpl.java
```diff
@@ -28,7 +28,7 @@ class OperationalLimitsGroupsImpl implements FlowsLimitsHolder {
     private final String attributeName;
     private final LinkedHashSet<String> selectedLimitsIds = new LinkedHashSet<>();
 
-    private final Map<String, OperationalLimitsGroupImpl> operationalLimitsGroupById = new LinkedHashMap<>();
+    private final LinkedHashMap<String, OperationalLimitsGroupImpl> operationalLimitsGroupById = new LinkedHashMap<>();
```
samirromdhani: Can stay as an interface: Map

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ExportOptions.java
NathanDissoubray: Good point !

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/NetworkSerDeTest.java
```diff
@@ -95,6 +97,31 @@ void writeMultipleSelectedOperationalLimitsGroupToOlderFormat() throws IOExcepti
...
+        Network n = EurostagTutorialExample1Factory.createWithMultipleSelectedFixedCurrentLimits();
+        n.getLine(EurostagTutorialExample1Factory.NHV1_NHV2_1).deselectOperationalLimitsGroups(TwoSides.ONE, EurostagTutorialExample1Factory.ACTIVATED_ONE_ONE);
+        //network with some inactive groups, export only active
+        Network networkRead = allFormatsRoundTripTest(n, "eurostag-tutorial-only-active-groups.xml", CURRENT_IIDM_VERSION, new ExportOptions().setOnlySelectedOperationalLimitsGroups(true));
+        assertEquals(Set.of("DEFAULT", EurostagTutorialExample1Factory.ACTIVATED_ONE_TWO), networkRead.getLine(EurostagTutorialExample1Factory.NHV1_NHV2_1).getAllSelectedOperationalLimitsGroupIds(TwoSides.ONE));
+    }
+
+    @Test
+    void testImportOnlyActiveGroups() throws IOException {
+        ExportOptions exportOptions = new ExportOptions();
```
olperr1: This variable is not necessary:
`NetworkSerDe.write(n, new ExportOptions(), p)` can be replaced by `NetworkSerDe.write(n, p)`.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ExportOptions.java
NathanDissoubray: Done in 16079af3c064

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/OperationalLimitsGroupsImpl.java
NathanDissoubray: Done in 16079af3c064

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/NetworkSerDeTest.java
```diff
@@ -95,6 +97,31 @@ void writeMultipleSelectedOperationalLimitsGroupToOlderFormat() throws IOExcepti
...
+    void roundTripTestExportOnlyActiveGroups() throws IOException {
+        Network n = EurostagTutorialExample1Factory.createWithMultipleSelectedFixedCurrentLimits();
+        n.getLine(EurostagTutorialExample1Factory.NHV1_NHV2_1).deselectOperationalLimitsGroups(TwoSides.ONE, EurostagTutorialExample1Factory.ACTIVATED_ONE_ONE);
+        //network with some inactive groups, export only active
+        Network networkRead = allFormatsRoundTripTest(n, "eurostag-tutorial-only-active-groups.xml", CURRENT_IIDM_VERSION, new ExportOptions().setOnlySelectedOperationalLimitsGroups(true));
+        assertEquals(Set.of("DEFAULT", EurostagTutorialExample1Factory.ACTIVATED_ONE_TWO), networkRead.getLine(EurostagTutorialExample1Factory.NHV1_NHV2_1).getAllSelectedOperationalLimitsGroupIds(TwoSides.ONE));
+    }
+
+    @Test
+    void testImportOnlyActiveGroups() throws IOException {
```
olperr1: Instead of writing the network, importing the file then re-exporting it, you can just import `eurostag-tutorial-multiple-selected-op-lim-group.xml` and check that the inactive groups are not imported. It would be lighter.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ConnectableSerDeUtil.java
```diff
@@ -414,13 +431,21 @@ static void readAllSelectedGroupIds(Branch<?> branch, TwoSides side, NetworkDese
...
+            side,
+            readAllSelectedGroupIds(
+                side.getNum(),
+                c -> transformer.getLeg(side).addSelectedOperationalLimitsGroups(c.toArray(String[]::new)),
+                context
+            ));
     }
 
-    static void readAllSelectedGroupIds(Integer index, Consumer<Collection<String>> consumer, NetworkDeserializerContext context) {
+    static Collection<String> readAllSelectedGroupIds(Integer index, Consumer<Collection<String>> consumer, NetworkDeserializerContext context) {
```
samirromdhani: Can be renamed to reflect its role as getter 

```suggestion
    static Collection<String> getAllSelectedGroupIds(Integer index, Consumer<Collection<String>> consumer, NetworkDeserializerContext context) {
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkDeserializerContext.java
```diff
@@ -36,6 +37,7 @@ public class NetworkDeserializerContext extends AbstractNetworkSerDeContext<Impo
     private ValidationLevel networkValidationLevel;
 
     private Set<String> ignoredEquipments = new HashSet<>();
+    private final Map<String, IdentifiableSelectedOperationalLimitGroups> selectedOperationalLimitGroupsByIdentifiableId = new HashMap<>();
```
samirromdhani: Selected group always fetched by IdentifiableId then by side, it would be simpler if the key become (identifiableId + side) and the value just a list:
```java
private final Map<SelectedGroupsKey, List<String>> selectedOperationalLimitGroupsByKey = new HashMap<>();
//
record SelectedGroupsKey(String identifiableId, ThreeSides side) { }

```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkDeserializerContext.java
```diff
@@ -103,8 +105,36 @@ public boolean isIgnoredEquipment(String equipment) {
...
     }
 
+    public void addSelectedGroupIds(String identifiableId, ThreeSides side, Collection<String> selectedGroupIds) {
+        IdentifiableSelectedOperationalLimitGroups selectedGroupsBySide = selectedOperationalLimitGroupsByIdentifiableId.get(identifiableId);
+        if (selectedGroupsBySide == null) {
+            selectedGroupsBySide = new IdentifiableSelectedOperationalLimitGroups(new ArrayList<>(), new ArrayList<>(), new ArrayList<>());
+            selectedOperationalLimitGroupsByIdentifiableId.put(identifiableId, selectedGroupsBySide);
+        }
+        selectedGroupsBySide.getSelectedFromSide(side).addAll(selectedGroupIds);
+    }
```
samirromdhani, line 112: `addSelectedGroupIds` and `getSelectedGroupIds` can be simplified:
```java
//addSelectedGroupIds
selectedOperationalLimitGroupsByIdentifiableId
        .computeIfAbsent(identifiableId, id -> new IdentifiableSelectedOperationalLimitGroups(new ArrayList<>(), new ArrayList<>(), new ArrayList<>()))
        .getSelectedFromSide(side)
        .addAll(selectedGroupIds);

//getSelectedGroupIds
return Optional.ofNullable(selectedOperationalLimitGroupsByIdentifiableId.get(identifiableId))
       .map(groups -> groups.getSelectedFromSide(side))
       .orElseThrow(() -> new NoSuchElementException("No identifiable of ID " + identifiableId + " has had its selected groups stored when deserializing the network"));
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/OperationalLimitsGroupsImpl.java
```diff
@@ -172,12 +172,12 @@ public List<String> getAllSelectedOperationalLimitsGroupIdsOrdered() {
...
 
     @Override
-    public Collection<OperationalLimitsGroup> getAllSelectedOperationalLimitsGroups() {
+    public List<OperationalLimitsGroup> getAllSelectedOperationalLimitsGroups() {
         return operationalLimitsGroupById.entrySet()
                 .stream()
                 .filter(e -> selectedLimitsIds.contains(e.getKey()))
                 .map(Map.Entry::getValue)
-                .collect(Collectors.toUnmodifiableSet());
+                .collect(Collectors.toList());
```
samirromdhani: Changes in the API contract, mutable list can be avoided ?

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractOperationalLimitsGroupsTest.java
```diff
@@ -578,6 +578,19 @@ public void operationalLimitsGroupOrdering() {
...
+        Network n = EurostagTutorialExample1Factory.createWithMultipleSelectedFixedCurrentLimits();
+        Line line = n.getLine(EurostagTutorialExample1Factory.NHV1_NHV2_1);
+        List<OperationalLimitsGroup> expectedGroups = line.getAllSelectedOperationalLimitsGroups(TwoSides.ONE);
+        expectedGroups.remove(1);
+        line.deselectOperationalLimitsGroups(TwoSides.ONE, EurostagTutorialExample1Factory.ACTIVATED_ONE_ONE);
+        assertEquals(expectedGroups, line.getAllSelectedOperationalLimitsGroups(TwoSides.ONE));
+        line.addSelectedOperationalLimitsGroups(TwoSides.ONE, EurostagTutorialExample1Factory.NOT_ACTIVATED);
+        expectedGroups.add(line.getOperationalLimitsGroup1(EurostagTutorialExample1Factory.NOT_ACTIVATED).orElseThrow());
+        assertEquals(expectedGroups, line.getAllSelectedOperationalLimitsGroups(TwoSides.ONE));
+    }
```
samirromdhani, line 597: The test is not very clear, it build the expected result by mutating a list, mutable list can be avoided (remove and add method not needed)

Here is a suggestion.
```suggestion
    public void getOperationalLimitsGroupsOrdering() {
        Network n = EurostagTutorialExample1Factory.createWithMultipleSelectedFixedCurrentLimits();
        Line line = n.getLine(EurostagTutorialExample1Factory.NHV1_NHV2_1);
        assertArrayEquals(new String[]{"DEFAULT", "activated_1_1", "activated_1_2"},
                line.getAllSelectedOperationalLimitsGroups(TwoSides.ONE).stream().map(OperationalLimitsGroup::getId).toArray());
        // Act
        line.deselectOperationalLimitsGroups(TwoSides.ONE, EurostagTutorialExample1Factory.ACTIVATED_ONE_ONE);
        // Assert
        assertArrayEquals(new String[]{"DEFAULT", "activated_1_2"},
                line.getAllSelectedOperationalLimitsGroups(TwoSides.ONE).stream().map(OperationalLimitsGroup::getId).toArray());
        // Act
        line.addSelectedOperationalLimitsGroups(TwoSides.ONE, EurostagTutorialExample1Factory.NOT_ACTIVATED);
        // Assert
        assertArrayEquals(new String[]{"DEFAULT", "activated_1_2", "not_activated"},
                line.getAllSelectedOperationalLimitsGroups(TwoSides.ONE).stream().map(OperationalLimitsGroup::getId).toArray());
    }
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/OperationalLimitsGroupsImpl.java
NathanDissoubray: The order was unstable with a set, discussed this with @olperr1 and we agreed to change the specification and the API to require a stable ordering (which is why it was changed from set to list). Also added a TCK test to reflect this change.
But it's true it was unmodifiable before, and it's modifiable now. However, the API does not make such guarantees regarding the returned collection, so I'm not sure it's necessary to make it unmodifiable.
There is no direct Collectors.toUnmodifiableList, but it's doable with a :
```java
Collectors.collectingAndThen(
                Collectors.toList(),
                Collections::unmodifiableList
        )
```

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkDeserializerContext.java
NathanDissoubray: when using this in code, one would still have to do a concatenation of the ids of the identifiable and the side. I don't think it's simpler than passing the identifiable and the side to the getter

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractOperationalLimitsGroupsTest.java
NathanDissoubray: you are right, since it's a mutable list and not a copy, there is no guarantee I'm not modifying the list directly in the line I'm taking this from.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ConnectableSerDeUtil.java
NathanDissoubray: mmh, it's not just a getter, it also creates a context task to add it to the network. `readAndGetAllSelectedGroupIds` ?

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkDeserializerContext.java
NathanDissoubray: It feels simpler simpler for add, but for get I think the original version is simpler
Before the logic was:
get it, it exists ? good, return it, otherwise throw
now it's, Make an optional of a value, map that (so if it's empty do nothing and if exists get it), then treat the case of it being null and return. It feels longer.

Ultimately, it's a choice of using stream or if/else

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ConnectableSerDeUtil.java
NathanDissoubray: Done in 5c4474b3a15b

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkDeserializerContext.java
NathanDissoubray: I misunderstood, I thought you meant that the id is a string concatenation of identifiableId + String.of(side), but you meant to use a record of the key. That avoids having empty list in the record, which is better, I'll do that

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/NetworkSerDeTest.java
```diff
@@ -95,6 +97,31 @@ void writeMultipleSelectedOperationalLimitsGroupToOlderFormat() throws IOExcepti
         );
     }
 
+    @Test
+    void roundTripTestExportOnlyActiveGroups() throws IOException {
```
samirromdhani, line 103: import and export tests can be done in `LoadingLimitsXmlTest` test class, what do you think ?

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkDeserializerContext.java
NathanDissoubray: Done in ad3ca8415347

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/NetworkSerDeTest.java
NathanDissoubray: the export test is a round trip on all formats so I would keep it in NetworkSerDeTest. And since the export is there, I would also keep the import, even though it's only on XIIDM.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/OperationalLimitsGroupsImpl.java
olperr1: `Collectors.toUnmodifiableList` [does exist in Java 21](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/stream/Collectors.html#toUnmodifiableList()).

NathanDissoubray: I searched in the javadoc but didn't find it, might have been looking at the wrong java version, I'll use that then

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/NetworkSerDeTest.java
NathanDissoubray: Indeed, since I rewrote the test according to your other comment, it doesn't use the export anyway now

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/NetworkSerDeTest.java
NathanDissoubray: done in 7ab92dece9e4

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/OperationalLimitsGroupsImpl.java
NathanDissoubray: done in 7ab92dece9e4

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkDeserializerContext.java
NathanDissoubray: I did it for the `add` in dfb6f91d04c4
I'm not convinced it's simpler with the `get` though

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/TieLineSerDe.java
```diff
@@ -195,7 +197,7 @@ protected void readSubElements(TieLine tl, NetworkDeserializerContext context) {
                 }
                 case LIMITS_GROUP_2 -> {
                     IidmSerDeUtil.assertMinimumVersion(ROOT_ELEMENT_NAME, LIMITS_GROUP_2, IidmSerDeUtil.ErrorMessage.NOT_SUPPORTED, IidmVersion.V_1_12, context);
-                    IidmSerDeUtil.runFromMinimumVersion(IidmVersion.V_1_12, context, () -> readLoadingLimitsGroups(tl.getBoundaryLine2(), LIMITS_GROUP_2, context));
+                    IidmSerDeUtil.runFromMinimumVersion(IidmVersion.V_1_12, context, () -> readLoadingLimitsGroups(tl.getBoundaryLine2(), tl.getId(), ThreeSides.ONE, LIMITS_GROUP_2, context)); //use side one since side is relative to the boundary line, not the tie line
```
olperr1: I was wondering if it is really `tl.getId()` that must be used here…
But it doesn't really matter since tie-lines don't have limits starting from IIDM v1.9! Therefore, this code is never used.
I think we should consider removing this dead code, but in another PRs. Could you write an issue on this subject?

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ConnectableSerDeUtil.java
```diff
@@ -412,7 +422,14 @@ static void readSelectedGroupId(Integer index, Consumer<String> selectedGroupIdS
...
      */
     static void readAllSelectedGroupIds(Branch<?> branch, TwoSides side, NetworkDeserializerContext context) {
-        readAllSelectedGroupIds(side.getNum(), c -> branch.addSelectedOperationalLimitsGroups(side, c.toArray(String[]::new)), context);
+        context.addSelectedGroupIds(
+            branch.getId(),
+            side.toThreeSides(),
+            readAndGetAllSelectedGroupIds(
+                side.getNum(),
+                c -> branch.addSelectedOperationalLimitsGroups(side, c.toArray(String[]::new)),
+                context
```
olperr1: i think it would be more readable to call `readAndGetAllSelectedGroupIds` and store its result in a variable first, and then to call `readAndGetAllSelectedGroupIds` using this variable.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ConnectableSerDeUtil.java
```diff
@@ -423,13 +440,21 @@ static void readAllSelectedGroupIds(Branch<?> branch, TwoSides side, NetworkDese
      * @param context     to deserialize the data
      */
     static void readAllSelectedGroupIds(ThreeWindingsTransformer transformer, ThreeSides side, NetworkDeserializerContext context) {
-        readAllSelectedGroupIds(side.getNum(), c -> transformer.getLeg(side).addSelectedOperationalLimitsGroups(c.toArray(String[]::new)), context);
+        context.addSelectedGroupIds(
```
olperr1, line 450: Same here, could you call `readAndGetAllSelectedGroupIds` first?

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ConnectableSerDeUtil.java
```diff
@@ -438,7 +463,14 @@ static void readAllSelectedGroupIds(Integer index, Consumer<Collection<String>>
      * @param context to deserialize the data
      */
     static void readAllSelectedGroupIds(BoundaryLine boundaryLine, NetworkDeserializerContext context) {
-        readAllSelectedGroupIds(null, c -> boundaryLine.addSelectedOperationalLimitsGroups(c.toArray(String[]::new)), context);
+        context.addSelectedGroupIds(
+            boundaryLine.getId(),
+            ThreeSides.ONE,
+            readAndGetAllSelectedGroupIds(
```
olperr1: And here too.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/NetworkSerDeTest.java
```diff
@@ -96,6 +99,37 @@ void writeMultipleSelectedOperationalLimitsGroupToOlderFormat() throws IOExcepti
...
```
olperr1, line 103: Could you also add tests on the other types?
- TwoWindingsTransformer
- ThreeWindingsTransformer
- BoundaryLine

TieLine could be omitted (see my comment on `TieLineSerDe.java`).

olperr1 (changes requested):

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/TieLineSerDe.java
NathanDissoubray: It should be the id of the boundary line instead of the tie line yes. Tie lines don't have limits, but boundary line can have limits, right ?
I'll open an issue

olperr1: Sorry, it was "starting from IIDM **v1.10**":
- [IIDM 1.9](https://github.com/powsybl/powsybl-core/blob/f9fca0b333e606174d9339c173d0947ae87dc151/iidm/iidm-serde/src/main/resources/xsd/iidm_V1_9.xsd#L557)
- [IIDM 1.10](https://github.com/powsybl/powsybl-core/blob/f9fca0b333e606174d9339c173d0947ae87dc151/iidm/iidm-serde/src/main/resources/xsd/iidm_V1_10.xsd#L575)

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ConnectableSerDeUtil.java
NathanDissoubray: done in 249250db5c04

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ConnectableSerDeUtil.java
NathanDissoubray: done in 249250db5c04

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/ConnectableSerDeUtil.java
NathanDissoubray: done in 249250db5c04

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/NetworkSerDeTest.java
NathanDissoubray: done in fe5200bfb7fb
I ended up testing tie line because the eurostag network which had boundary lines had them in a tie line

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/TieLineSerDe.java
NathanDissoubray: Fixed the ID in 271b7b81128c

olperr1 (approved):

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/TieLineSerDe.java
NathanDissoubray: Issue open #3961

PR #3939, opened 2026-06-05
Add ObservableList + retain TimeSeries values array to avoid recomputation when not necessary

rolnico:

No

Feature

When a TimeSeries is computed, if you call `.toArray()` on it twice in a row, the array is computed twice, even if no modification has been made on the TimeSeries in between.

When a TimeSeries is computed, if you call `.toArray()` on it, the array is computed once and then saved in memory for future use.
If any modification is done on the TimeSeries, the saved array is discarded so that it is recomputed at the next call.
To help, a new interface `ObservableList` and its implementation `ObservableArrayList` were added, so that you can add listeners (via a new class `ListChangeListener`) on any list modification

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/commons/src/main/java/com/powsybl/commons/list/ObservableArrayList.java
+++ b/commons/src/main/java/com/powsybl/commons/list/ObservableArrayList.java
@@ -1,0 +1,213 @@
+/*
+ * Copyright (c) 2026, RTE (https://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+package com.powsybl.commons.list;
+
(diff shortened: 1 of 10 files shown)


commons/src/main/java/com/powsybl/commons/list/ObservableArrayList.java
```diff
@@ -0,0 +1,213 @@
+/*
+ * Copyright (c) 2026, RTE (https://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+package com.powsybl.commons.list;
```
samirromdhani, line 8: I'm wondering if we can keep this local to time series and align with the existing `TimeSeriesStoreListener` or something similar, instead of introducing a generic Observable List API in commons.

PR #3942, opened 2026-06-08
DC connectable connect & disconnect method

bperr:

Add a connect and disconnect method for DC connectable, allowing to act directly on all their terminals rather than accessing each one separately.

To (dis)connect an DC element, one must iterate over its DC terminals and (dis)connect them individually

the process is encapsulated in a single method

- [x] Yes
- [ ] No

If you have your own implementation of `DcConnectable`, you must add:
- `boolean connectDc()`
- `boolean disconnectDc()`

**Other information**:

`AcDcConverter` have both (dis)connect and (dis)connectDc. The first one acts on the AC side only and the second one on the DC side only.


docs/grid_model/network_subnetwork.md
```diff
@@ -942,7 +942,7 @@ LCC and VSC share the following characteristics.
 | $SwitchingLoss$ | MW / A   | Switching losses                                                      |
 | $ResistiveLoss$ | $\Omega$ | Resistive losses                                                      |
 | $PccTerminal$   |          | Point of common coupling (PCC) AC terminal                            |
-| $ControlMode$   |          | The converter's control mode: P_PCC, V_DC or V_DC_DROOP               |
+| $ControlMode$   |          | The converter's control mode: P_PCC, V_DC or P_PCC_DROOP              |
```
SGI-Landry, line 945: Thanks for the fix!

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractDcConnectableTest.java
```diff
@@ -0,0 +1,110 @@
...
+import com.powsybl.iidm.network.test.DcDetailedNetworkFactory;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * @author Baptiste Perreyon {@literal <baptiste.perreyon at supergrid-institute.com>}
+ */
+public abstract class AbstractDcConnectableTest {
```
SGI-Landry, line 19: This is systematic and very thorough, thanks!

If anything, it could be interesting in another test (maybe elsewhere?) to check that connecting/disconnecting the AC side does not modify the DC side of a converter.

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/tck/DcConnectableTest.java
```diff
@@ -0,0 +1,16 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+package com.powsybl.iidm.network.impl.tck;
+
+import com.powsybl.iidm.network.tck.AbstractDcConnectableTest;
+
+/**
+ * @author Baptiste Perreyon {@literal <baptiste.perreyon at supergrid-institute.com>}
+ */
+class DcConnectableTest extends AbstractDcConnectableTest {
```
SGI-Landry, line 15: I am wondering if all test classes of DcConnectable (e.g. AbstractDcLineTest) should also inherit AbstractDcConnectableTest. This might require AbstractDcConnectableTest to be changed into an interface. This would check that the interface is satisfied for all implementations.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
```diff
@@ -148,4 +146,85 @@ static boolean disconnectAllTerminals(Identifiable<?> identifiable, List<? exten
...
+     * @param dcTerminals list of the DC terminals that should be connected. It should be the DC terminals of the DC
+     *                    connectable.
+     * @param reportNode report node
+     * @return {@code true} if all the specified DC terminals have been connected, else {@code false}.
+     */
+    static boolean connectAllDcTerminals(Identifiable<?> identifiable, List<? extends DcTerminal> dcTerminals, ReportNode reportNode) {
+
+        // Booleans
+        boolean isAlreadyConnected = true;
+        boolean isNowConnected = true;
```
SGI-Landry: This can be simplified
- Remove `isNowConnected`.
- Return true at the end of the method.

DC equipment cannot fail to connect, unlike AC equipment.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
```diff
@@ -148,4 +146,85 @@ static boolean disconnectAllTerminals(Identifiable<?> identifiable, List<? exten
...
+    /**
+     * Disconnect the specified DC terminals.
+     *
+     * @param identifiable network element to disconnect. It should be a DC connectable.
+     * @param dcTerminals list of the DC terminals that should be disconnected. It should be the DC terminals of the DC
+     *                    connectable.
+     * @param reportNode report node
+     * @return {@code true} if all the specified DC terminals have been disconnected, else {@code false}.
+     */
+    static boolean disconnectAllDcTerminals(Identifiable<?> identifiable, List<? extends DcTerminal> dcTerminals, ReportNode reportNode) {
```
SGI-Landry, line 196: Same comment as above.

SGI-Landry (commented):
Thanks @bperr for this activity. A few minor comments but you are almost good to go.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
bperr: Yes, much simpler that wat

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/tck/DcConnectableTest.java
bperr: I understand the idea, but i think that would lead to unecessary compexity. As an example, `AbstractDcLineTest` has a beforeEach method that creates memebers that becomes accessible in `AbstractDcConnectableTest` and which name can clash with normal variables.
I believe it is better to keep the current structure, which is also the one used for the rest of the code base (e.g. `AbstractConnectableTest` has only one inherithor: `ConnectableTest`)

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractDcConnectableTest.java
bperr: I updated the `AbstractConnectableTest` (AC) to include this verification

bperr:
Hi @olperr1 and @NathanDissoubray. Could you take a look at this PR ? Also, if you can think of anyone else who might be better placed to review it, please ask them; I wasn’t sure who to ask to review this

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/tck/DcConnectableTest.java
```diff
@@ -0,0 +1,16 @@
+/**
+ * Copyright (c) 2026, SuperGrid Institute (http://www.rte-france.com)
```
olperr1:
```suggestion
 * Copyright (c) 2026, SuperGrid Institute (https://www.supergrid-institute.com)
```

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractDcConnectableTest.java
```diff
@@ -0,0 +1,110 @@
+/**
+ * Copyright (c) 2026, SuperGrid Institute (http://www.rte-france.com)
```
olperr1:
```suggestion
 * Copyright (c) 2026, SuperGrid Institute (https://www.supergrid-institute.com)
```

NathanDissoubray:
I'll take a look :+1:

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcConnectable.java
```diff
@@ -25,4 +25,16 @@ public interface DcConnectable<I extends DcConnectable<I>> extends Identifiable<
...
+     * Try to connect the DC connectable
+     * @return true if the connection by this operation succeeded, false otherwise (the DC connectable was already connected)
+     */
+    boolean connectDc();
+
+    /**
+     * Try to disconnect the DC connectable
+     * @return true if the disconnection by this operation succeeded, false otherwise (the DC connectable was already disconnected)
+     */
+    boolean disconnectDc();
```
NathanDissoubray, line 39: why not call it `connect` and `disconnect` like in `Connectable` ?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
```diff
@@ -148,4 +146,79 @@ static boolean disconnectAllTerminals(Identifiable<?> identifiable, List<? exten
...
+    /**
+     * Connect the specified DC terminals.
+     *
+     * @param identifiable network element to connect. It should be a DC connectable.
+     * @param dcTerminals list of the DC terminals that should be connected. It should be the DC terminals of the DC
+     *                    connectable.
+     * @param reportNode report node
+     * @return {@code true} if all the specified DC terminals have been connected, else {@code false}.
+     */
+    static boolean connectAllDcTerminals(Identifiable<?> identifiable, List<? extends DcTerminal> dcTerminals, ReportNode reportNode) {
```
NathanDissoubray, line 159: why doesn't this operate switches too ?

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractDcConnectableTest.java
```diff
@@ -0,0 +1,110 @@
...
+        assertTrue(dcline.connectDc());
+        assertTrue(dcline.getDcTerminal1().isConnected());
+        assertTrue(dcline.getDcTerminal2().isConnected());
+
+        // Partially disconnect DC side, disconnectDc should return true
+        dcline.getDcTerminal1().setConnected(false);
+        assertTrue(dcline.disconnectDc());
+        assertFalse(dcline.getDcTerminal1().isConnected());
+        assertFalse(dcline.getDcTerminal2().isConnected());
+    }
```
NathanDissoubray, line 59:
```suggestion
    public void connectAndDisconnect() {
        Network network = DcDetailedNetworkFactory.createVscAsymmetricalMonopole();
        DcLine dcline = network.getDcLine("dcLinePos");

        // Check dcLine is fully connected
        assertDcLineStatus(true, dcline);

        // Disconnect dcLine, which should return true
        assertTrue(dcline.disconnectDc());
        assertDcLineStatus(false, dcline);

        // Disconnect dcLine again, which should return false
        assertFalse(dcline.disconnectDc());
        assertDcLineStatus(false, dcline);

        // Connect dcLine, which should return true
        assertTrue(dcline.connectDc());
        assertDcLineStatus(true, dcline);

        // Connect dcLine again, which should return false
        assertFalse(dcline.connectDc());
        assertDcLineStatus(true, dcline);

        // Partially disconnect DC side, connectDc should return true
        dcline.getDcTerminal1().setConnected(false);
        assertTrue(dcline.connectDc());
        assertDcLineStatus(true, dcline);

        // Partially disconnect DC side, disconnectDc should return true
        dcline.getDcTerminal1().setConnected(false);
        assertTrue(dcline.disconnectDc());
        assertDcLineStatus(false, dcline);
    }
    
    private void assertDcLineStatus(boolean expectedStatus, DcLine dcLine) {
        assertEquals(expectedStatus, dcline.getDcTerminal1().isConnected());
        assertEquals(expectedStatus, dcline.getDcTerminal2().isConnected());
    }
```

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractDcConnectableTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @Test
+    public void connectAndDisconnectAcDcConverter() {
```
NathanDissoubray, line 62: same suggestion by using the proposed `assertDcLineStatus` function

NathanDissoubray (changes requested):

olperr1 (changes requested):

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcConnectable.java
bperr: The problem here is the `AcDcConverter` interface which extends both `Connectable` and `DcConnectable`.
- Firstly, from a physical perspective, you might want to disconnect only the DC part or only the AC part of a converter.
- Secondly, from a sofware perspective, this double inheritance causes conflicts in the `AcDcConverter` interface which would have a default implementation of `connect()` from the `Connectable` and an abstract method connect() from the `DcConnectable`. While this can be overcommed by overriding `connect()` in the `AcDcConverter` interface, this eventually leads to a more ambiguous code that is harder to understand and maintain.

Therefore, I believe it is better to name this methods `connectDc` and `diconnectDc` even though it can seem a little redundant for other `DcConnectable`

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
bperr: To my understanding, the DC detailed model is closer to the bus/breaker view of AC, and one can assume there  are implicit disconnectors at the end of each DC connectable. Therefore, there is no need to search for switches here.

NathanDissoubray: I see, but there are still DC breakers though https://github.com/powsybl/powsybl-core/blob/921437736ae300557bcd4831451399a288065339/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcSwitchKind.java#L10-L18 what about those ?

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractDcConnectableTest.java
```diff
@@ -22,43 +21,41 @@ public abstract class AbstractDcConnectableTest {
...
-        dcline.getDcTerminal1().setConnected(false);
-        assertTrue(dcline.disconnectDc());
-        assertFalse(dcline.getDcTerminal1().isConnected());
-        assertFalse(dcline.getDcTerminal2().isConnected());
+        dcLine.getDcTerminal1().setConnected(false);
+        assertTrue(dcLine.disconnectDc());
+        assertDcLineStatus(false, dcLine);
+    }
+
+    private void assertDcLineStatus(boolean expectedStatus, DcLine dcLine) {
```
NathanDissoubray: I would maybe rename it to `assertDcLineConnectionStatus` (similarly for the converter)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
bperr: There are DC switches (breakers/disconnectors) in DC. But to me they are used in case of contingencies (for breakers) and to modifiy the DC topology (i.e. merge two DC nodes into the same DC bus when closing a DC switch). This is why I was comparing it to the AC bus/breaker view: you still have switches to modify the topology, but you can consider that each connectable can be disconnected safely.

If it is required to operate DC switch to disconnect a DC connectable, that implies the user is required to model *explicitely* every DC switch in the system.
I am not sure how this is handle by other data model such as CGMES

NathanDissoubray: @jeandemanged do you have some input regarding this ? i'm not well versed enough regarding DC, and it seems strange to me that we can disconnect anything without operating breakers.

jeandemanged: Hi @NathanDissoubray ,
@bperr is correct. DC topology model is built like AC Voltage Level Bus/Breaker topology model, not the Node/Breaker topology model.
In DC topology model (just like AC Bus/Breaker topology model), the terminals are connected/disconnect using a simple boolean at the terminal itself, not via switches/breakers manipulations.

NathanDissoubray: Thank you for the input @jeandemanged and sorry for the confusion on my part @bperr  
If it's built like AC bus/breaker topology, don't you ever need to operate breakers though (independently of the connect/disconnect here) ?

bperr: We need to operate breakers, for example in case of contingencies. From a physical point of view, there are disconnectors at the end of the DC line but to clear a fault you need a DCCB (i.e. not a simple disconnector).

NathanDissoubray (dismissed):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
```diff
@@ -148,4 +146,79 @@ static boolean disconnectAllTerminals(Identifiable<?> identifiable, List<? exten
...
+
+        // Booleans
+        boolean isAlreadyConnected = true;
```
olperr1: Can you rename this variable `isAlreadyFullyConnected`?
This would be more readable.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
```diff
@@ -148,4 +146,79 @@ static boolean disconnectAllTerminals(Identifiable<?> identifiable, List<? exten
...
+
+        // Check if the element is already connected
+        for (DcTerminal dcTerminal : dcTerminals) {
+            if (dcTerminal.isConnected()) {
+                NetworkReports.alreadyConnectedIdentifiableDcTerminal(reportNode, identifiable.getId());
```
olperr1, line 167: It would be clearer to have this report in the `if (isAlreadyFullyConnected)` block.
In the current state, we could generate several report messages for the same identifiable. And a message will be added even when the connect method does something (when one terminal is connected but not the other). This could be misleading.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
```diff
@@ -148,4 +146,79 @@ static boolean disconnectAllTerminals(Identifiable<?> identifiable, List<? exten
...
+     * @param identifiable network element to disconnect. It should be a DC connectable.
+     * @param dcTerminals list of the DC terminals that should be disconnected. It should be the DC terminals of the DC
+     *                    connectable.
+     * @param reportNode report node
+     * @return {@code true} if all the specified DC terminals have been disconnected, else {@code false}.
+     */
+    static boolean disconnectAllDcTerminals(Identifiable<?> identifiable, List<? extends DcTerminal> dcTerminals, ReportNode reportNode) {
+
+        // Booleans
+        boolean isAlreadyDisconnected = true;
```
olperr1: Same, could you rename it `isAlreadyFullyDisconnected`?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
olperr1: Also, the message should be slightly adapted in the `reports....properties` files to indicate that the identifiable is already **fully** connected.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
```diff
@@ -148,4 +146,79 @@ static boolean disconnectAllTerminals(Identifiable<?> identifiable, List<? exten
...
+
+        // Check if the element is already disconnected
+        for (DcTerminal dcTerminal : dcTerminals) {
+            if (!dcTerminal.isConnected()) {
+                NetworkReports.alreadyDisconnectedIdentifiableDcTerminal(reportNode, identifiable.getId());
```
olperr1, line 204: Here, I think that it is interesting to have a report when at least one terminal is already disconnected. But it's maybe better to have it only once.

olperr1 (changes requested):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
bperr: Should I update the util for AC disconnection too ? Because it creates a report for each AC terminal already connected. This is why I chose to follow the same logic here

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
bperr: Same question here. Should I also modify the reports behavior for AC, leave both as it is now or modify only DC (I would not recommend it, I believe it is clearer if both behavior are aligned)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
olperr1: Sorry, I missed this point. You are right it's better to be consistent.
I think will can keep it like this for now.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ConnectDisconnectUtil.java
olperr1: Same, we can keep it like this for now.

olperr1 (approved):

NathanDissoubray (approved):

PR #3612, opened 2025-10-16
Implement Model a PQU diagram for reactive limits

snoopfab:
**Please check if the PR fulfills these requirements**

- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)
- [x] A PR or issue has been opened in all impacted repositories (if any)

**Does this PR already have an issue describing the problem?**

Implement #3511

**What kind of change does this PR introduce?**

feature : Implement Model a PQU diagram for reactive limits

**What is the current behavior?**

https://github.com/powsybl/powsybl-core/issues/3511

**What is the new behavior (if this is a feature change)?**

**Does this PR introduce a breaking change or deprecate an API?**
- [ ] Yes
- [x] No

**If yes, please check if the following requirements are fulfilled**

- [ ] The *Breaking Change* or *Deprecated* label has been added
- [ ] The migration steps are described in the following section

**What changes might users need to make in their application due to this PR? (migration steps)**

**Other information**:

PR #3949, opened 2026-06-12
Implement Model a PQU diagram for reactive limits - Without serialization

### LINKED ISSUE #3511: Model a PQU diagram for reactive limits

Generators, dangling lines and battery at least have reactive limits. These reactive limits can be of two types currently: 

- `MinMaxReactiveLimits`: determined by a minQ and maxQ
- `ReactiveCapabilityCurve`: determined by a minQ and maxQ per given P (active power)

We add a new implementation `ReactiveShape` that is able to model a 3D (P, Q, V) volume. 
The volume is defined by a list of planes.
Each plane is described by an equation `q + alpha * v + beta * p < OR > gamma`
A detection algorithm is implemented to check if a given point (p, q, v) is inside the shape. (!!! there is probably some existing Java geometry libraries for that, to check otherwise to fully recode)

### Describe the motivation

Model a volume of reactive limits for a generator for instance, depending on both the voltage and the active power.

### Extra Information

Create a new `ReactiveLimits` implemenation allowing to compute minQ and maxQ according to given active power AND voltage. This may imply a refactor of the `ReactiveLimits` implementations + modify the iidm, allowng to define new kinds of reactive limit (a volume instead of only a reactive capability curve).

In addition to existing `double getMinQ(double p)` and `double getMaxQ(double p)` of `ReactiveLimits`, we add `double getMinQ(double p, double v)` and `double getMaxQ(double p, double v)`. In case of calling its with just p with a ReactiveShape we use the current calculated v of the generator (bus.getV()). In case of no calculated v, an exception is thrown.

snoopfab:

Implement #3511

feature : Implement Model a PQU diagram for reactive limits
Split up into two branches of PR  Implement Model a PQU diagram for reactive limits  #361 (https://github.com/powsybl/powsybl-core/pull/3612#issuecomment-4563442874) between PoJo Model classes and serialisation classes in IIDM XML
This one just add PoJo classes with no serialization 
See #3950 for serialization implementation on top of this one

https://github.com/powsybl/powsybl-core/issues/3511

- [X] Yes, for custom IIDM implementations' maintainers only
- [ ] No

Custom IIDM implementation maintainers must implement the following interfaces:
- `ReactiveCapabilityShape`
- `ReactiveCapabilityShapeAdder`
- `ReactiveCapabilityShapePlane`

And the following method in their `ReactiveLimitsHolder` implementations
 - `ReactiveCapabilityShapeAdder newReactiveCapabilityShape()`

Note that:
- this feature is in BETA for now. No TCK are available for now, since the format may change a little in the future.
- these elements are not handled by the IIDM (de)serialization

**Other information**:


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/MinMaxReactiveLimitsAdder.java
```diff
@@ -11,7 +11,7 @@
  *
  * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
  */
-public interface MinMaxReactiveLimitsAdder {
+public interface MinMaxReactiveLimitsAdder extends PropertiesHolder {
```
olperr1: Having the ability to add properties to the adders is a great feature, but I think this should be introduced in a dedicated PR and not inside this one.

Could you report this class changes in another PR and revert this class changes in this PR?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ReactiveCapabilityCurveAdder.java
```diff
@@ -11,7 +11,7 @@
  *
  * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
  */
-public interface ReactiveCapabilityCurveAdder extends BasePropertiesHolder {
+public interface ReactiveCapabilityCurveAdder extends PropertiesHolder {
```
olperr1: Related to my previous comment on properties.
This class changes should be reported in the property dedicated PR, and be reverted here.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/MinMaxReactiveLimitsAdderImpl.java
```diff
@@ -16,7 +16,7 @@
  * @author Geoffroy Jamgotchian {@literal <geoffroy.jamgotchian at rte-france.com>}
  * @author Mathieu Bague {@literal <mathieu.bague at rte-france.com>}
  */
-class MinMaxReactiveLimitsAdderImpl<O extends ReactiveLimitsOwner & Validable> implements MinMaxReactiveLimitsAdder {
+class MinMaxReactiveLimitsAdderImpl<O extends ReactiveLimitsOwner & Validable> extends AbstractPropertiesHolder implements MinMaxReactiveLimitsAdder {
```
olperr1: Related to my previous comment on properties.
This class changes should be reported in the property dedicated PR, and be reverted here.

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractMinMaxReactiveLimitsTest.java
```diff
@@ -9,6 +9,7 @@
 
 import com.powsybl.iidm.network.Generator;
 import com.powsybl.iidm.network.MinMaxReactiveLimits;
+import com.powsybl.iidm.network.MinMaxReactiveLimitsAdder;
```
olperr1: Related to my previous comment on properties.
This class changes should be reported in the property dedicated PR, and be reverted here.

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractReactiveCapabilityCurveTest.java
```diff
@@ -10,6 +10,7 @@
 import com.powsybl.iidm.network.Generator;
 import com.powsybl.iidm.network.Network;
 import com.powsybl.iidm.network.ReactiveCapabilityCurve;
+import com.powsybl.iidm.network.ReactiveCapabilityCurveAdder;
```
olperr1: Related to my previous comment on properties.
This class changes should be reported in the property dedicated PR, and be reverted here.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ReactiveLimitsHolder.java
```diff
@@ -45,4 +51,13 @@ default ReactiveCapabilityCurveAdder newReactiveCapabilityCurve(ReactiveCapabili
...
         return adder;
     }
+
+    default double getMinP() {
+        return Double.NaN;
+    }
+
+    default double getMaxP() {
+        return Double.NaN;
+    }
```
olperr1: This changes are not needed. Can you revert them?

olperr1 (changes requested):

docs/grid_model/additional.md
```diff
@@ -64,6 +72,30 @@ generator.newReactiveCapabilityCurve()
...
+ReactiveCapabilityShapePlane p3 = ReactiveCapabilityShapePlane.build(0.0, 1.0).lessOrEqual(120.0);
+// Q + 0*U + 1*P ≥ -50
+ReactiveCapabilityShapePlane p4 = ReactiveCapabilityShapePlane.build(0.0, 1.0).greaterOrEqual(-50.0);
+// Q + 1*U + 0*P ≤ 410
+ReactiveCapabilityShapePlane p5 = ReactiveCapabilityShapePlane.build(1.0, 0.0).lessOrEqual(410.0);
+// Q + 1*U + 0*P ≥ 390    → U ≥ 390
+ReactiveCapabilityShapePlane p6 = ReactiveCapabilityShapePlane.build(1.0, 0.0).greaterOrEqual(390.0);
+
+Generator generator = network.getGenerator("G");
+generator.newReactiveCapabilityShape().setPolyhedron(
```
olperr1: Unresolved comment from @colineplqt on the previous PR:

> I wanted to test this new feature and I noticed that setPolyhedron does not exist. Can you fix this documentation?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ReactiveCapabilityShapePolyhedron.java
```diff
@@ -0,0 +1,356 @@
...
+    /**
+     * Logger
+     */
+    private static final Logger LOGGER = LoggerFactory.getLogger(ReactiveCapabilityShapePolyhedron.class);
+    /**
+     * Variables indexes
+     */
+    private static final int IDX_Q = 0;
+    private static final int IDX_P = 1;
+    private static final int IDX_U = 2;
```
olperr1: Unresolved comment from @colineplqt on the previous PR:

> I am not sure but I think there is an inconsistency between these indexes and the constraints.
> Here you have Q=0, P=1 and U=2 but the constraints are defined as:
> `new double[]{plane.getDelta(), plane.getAlpha(), plane.getBeta()}, plane.isLessOrEqual() ? Relationship.LEQ : Relationship.GEQ, plane.getGamma()`
> 
> So first delta, then alpha then beta ie first Q then U then P.
> So should it not be Q = 0, U = 1 and P = 2?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ReactiveCapabilityShapeImpl.java
```diff
@@ -0,0 +1,104 @@
...
+     */
+    private ReactiveCapabilityShapeImpl(final ReactiveCapabilityShapePolyhedron polyhedron) {
+        this.polyhedron = polyhedron;
+    }
+
+    /**
+     * Builder
+     * @param polyhedron the reactive capability shape polyhedron
+     */
+    public static ReactiveCapabilityShapeImpl build(final ReactiveCapabilityShapePolyhedron polyhedron) {
```
olperr1, line 90: Unresolved comment from @colineplqt on the previous PR:
> I wonder if this `build` method could not be defined in the interface so that you can do `ReactiveCapabilityShapePlane.build()` like it is mentionned in the documentation.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ReactiveCapabilityShapePlaneImpl.java
```diff
@@ -0,0 +1,179 @@
...
+        return new ReactiveCapabilityShapePlaneImpl(alphaU, betaP, deltaQ);
+    }
+
+    /**
+     * Set the hyperplane constraint to a less or equal ≤ inequality with gamma right hand side
+     * @param gamma the gamma right hand side
+     * @return this
+     */
+    @Override
+    public ReactiveCapabilityShapePlane lessOrEqual(double gamma) {
```
olperr1, line 107: Unresolved comment from @colineplqt on the previous PR:
> I tested using these new limits and I was wondering why you chose to set gamma and the inequality type outside of the builder?

docs/grid_model/additional.md
```diff
@@ -18,6 +18,14 @@ With the reactive capability curve limits, the reactive power limitation depends
 The curve is defined as a set of points that associate, to each active power value, a minimum and maximum reactive power value.
 In between the defined points of the curve, the reactive power limits are computed through a linear interpolation.
 
+(reactive-capability-shape)=
+### Reactive capability shape
+With the reactive capability shape limits, the reactive power limitation depends on a 3D (P, Q, V) convex volume.
+The volume is defined by a list of planes provided by the user
+Each plane is described by an equation q + alpha * v + beta * p ≤ or ≥ gamma.
+Additionally, some upper and lower bounds can be defined for each of the P,Q and V variables 
+
```
olperr1, line 40: Can you add the following comment here:

\`\`\`{note}
Currently, this type of limits is in BETA and only available in the IIDM representation. There is no import or export of this kind of limit with any exchange format.
The model is subject to change and support for downstream projects (\`powsybl-open-loadflow\`, \`powsybl-dynawo\`, etc.) may vary.
Please consult the documentation of each project to verify support. In general, lack of explicit mention means no support.

If you're unsure, feel free to reach out to the PowSyBl community \[here\]\(https://www.powsybl.org/pages/community/contact.html \)
\`\`\`

docs/grid_model/additional.md
```diff
@@ -18,6 +18,14 @@ With the reactive capability curve limits, the reactive power limitation depends
...
+
```
olperr1, line 41: On the same subject, we should throw an error then someone tries to export a network having this new kind of reactive limits.

Can you add this exception in that case?

```java
import org.apache.commons.lang3.NotImplementedException;


if (!exportOptions.isForceExportNetworkWithBetaFeatures()) {
    throw new NotImplementedException("The network contains Reactive capability shapes, export of this kind of reactive limits is not yet supported. " +
        "To force the export of the network and ignore those limits, either use the config parameter iidm.export.xml.force-export-network-with-beta-features, " +
        "or ExportOptions.setForceExportNetworkWithBetaFeatures");
}
```

olperr1 (dismissed):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ReactiveCapabilityShapePolyhedron.java
snoopfab: you were right, the index were wrong. Fixed, and added tests to catch this

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ReactiveCapabilityShapeImpl.java
snoopfab: the documentation example that suggested ReactiveCapabilityShapePlane.build(...) was stale and has been corrected. Planes are not meant to be constructed directly by API users: the public construction path is the fluent adder, generator.newReactiveCapabilityShape().addPlane(alpha, beta, delta, isGreaterOrEqual, gamma)...add(), which takes the coefficients, the inequality direction and gamma in a single call. The build(...) static factories therefore stay on the impl classes (iidm-impl), which an iidm-api interface cannot reference anyway without extra plumbing. The docs now show the real addPlane API, so there's no longer a public-facing build() expectation to satisfy.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ReactiveCapabilityShapePlaneImpl.java
snoopfab: the public-API level gamma and the inequality direction are part of the builder: addPlane(...) takes them in one call. The two-step build(coeffs) + lessOrEqual/greaterOrEqual(gamma) only exists internally; it keeps the plane's geometric coefficients separate from its relational orientation, which mirrors how the half-space is later assembled into a LinearConstraint. I've documented this on ReactiveCapabilityShapePlaneImpl.

docs/grid_model/additional.md
snoopfab: done

docs/grid_model/additional.md
snoopfab: done

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ReactiveCapabilityShapePolyhedron.java
colineplqt: Thank you!

docs/grid_model/additional.md
olperr1: Thanks!

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ReactiveCapabilityShapeImpl.java
colineplqt: Thank you

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/ReactiveCapabilityShapePlaneImpl.java
colineplqt: Thank you ! It looks good for me.

colineplqt (approved):

PR #3951, opened 2026-06-12
Features/parallel switchflow indicator

nicow-elia:

Until now, the switch flow functionality does not report to the user whether a switch has other parallel elements or not. If there are no parallel switches, then the results of the switchflow computation are physically meaningful, i.e. the values reported should match roughly what is measured in the grid. On parallel switches, the current functionality does still report p/q values that at least satisfy kirchhoff (i.e. sum to the correct amount) but the distribution of the switch flow is based on switch order in the grid model, where the switch that the algorithm first iterates on take all the flow. This implementation choice is not physically meaningful in that configuration but simplifies the logic, as it does not require any switch impedances or distribution mechanism. However, to make this functionality more usable, the limitations of the algorithm should be presented explicitely. Namely, if there is a case of parallel switchers, a user of the switchflow API would like to know that the results are not to be interpreted physicallyi

This PR introduces two new methods to the Switchflow interface, `hasParallelSwitch` and `getFirstParallelSwitch`. With the first one, the user can determine if there is a parallel switch - with the second one you can see the id of one of the parallel switches in case the user wants to investigate which switch is offending. I was considering - for the case of 3 or more switches - to return a set of parallel switches in a method `getParallelSwitches` but that would increase the runtime requirements slightly and complicate the code, and is not strictly needed for debugging as you can trivially reverse-engineer the graph for the offending switch. If you think this is a better solution though, I can include this as well, maybe in a separate function `getParallelSwitches` that performs a graph reconstruction.

Only p and q are exposed for each switch

Information about existing parallel switches is exposed as well.

- [ ] Yes
- [x] No


iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/util/SwitchesFlowTest.java
```diff
@@ -61,6 +66,31 @@ void switchesFlowNodeBreaker() {
         assertSwitchFlowValue(switchesFlow31, "S3VL1-SW-0-2", 71.0, 11.0);
     }
 
+    @Test
+    void switchesFlowParallelSwitchMetadata() {
```
olperr1, line 66: Could you add a test for the bus/breaker topology too?
For instance, you can slightly change the network returned by `createBusBreaker()`:
- add a second bus on "S1VL2"
- connect "S1VL2-Load" to this new bus instead
- add a switch between the 2 busses of "S1VL2".

Then in your unit test, you can add 2 new switches between the 2 busses.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/util/SwitchesFlow.java
```diff
@@ -140,12 +162,22 @@ private void buildGraphFromBusBreaker(Map<String, SwNode> swNodeInjection, Simpl
...
     }
 
+    private void addSwitchEdge(SimpleWeightedGraph<SwNode, SwEdge> graph, Switch sw, SwNode swNode1, SwNode swNode2) {
+        graph.addVertex(swNode1);
+        graph.addVertex(swNode2);
+        SwEdge existingEdge = graph.getEdge(swNode1, swNode2);
+        if (existingEdge == null) {
+            graph.addEdge(swNode1, swNode2, new SwEdge(sw, swNode1, swNode2));
+        } else if (existingEdge.isSwitch()) {
+            parallelSwitchIds.put(sw.getId(), existingEdge.getSwitchId());
```
olperr1, line 177:
```suggestion
            // No need to add a new edge on the graph:
            // since SimpleWeightedGraph does not allow parallel edges, it would not change the graph anyway.
            parallelSwitchIds.put(sw.getId(), existingEdge.getSwitchId());
```

olperr1 (changes requested):
Thank you for the contribution, @nicow-elia! 

There are checkstyle errors due to the new rules applied on the project (introduced by #3924. See also powsybl/powsybl-parent#90). Could you fix them?

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/util/SwitchesFlowTest.java
nicow-elia: Done, had to add a new VL so the isEmpty test still does something.

olperr1: Thanks!

olperr1 (approved):

nicow-elia:
Thanks for the review! Just FYI - I am not authorized to merge, even with approval.

PR #3960, opened 2026-06-15
Powerfactory: import gps coordinates

### LINKED ISSUE #2405: PowerFactory: import GPS coordinate
Feature

We should take GPS coordinates and create SubstationPostion, LinePosition IIDM extensions.
ElmSubstat, ElmTrfstat, or ElmSite should have GPSLat and GPSLon attributes.
ElmLne has GPSCoords attribute.

  - PowSyBl Version: ...
  - OS Version: ...

marqueslanauja:

Fixes #2405

Feature

GPS coordinates are not imported. 

GPS coordinates are imported through configuration. The `powerfactory.import.geodata` property must be set to true to enable geodata import for substations and lines.

For lines, the` GPScoords` of `ElmLne` attribute is used to fill the line position coordinates.
For substations, the `GPSlat` and `GPSlon` attributes of `ElmSite`, `ElmSubstat`, `ElmTrfstat`, or `ElmTerm` are used to define the substation geodata.

`ElmSite` is used when the substation ID is created from it.
`ElmSubstat` and `ElmTrfstat` are used when the substation has only one voltage level and the voltage level ID is created from them.
Finally, when the substation ID is created from an `ElmTerm`, its coordinates are used to define the substation coordinates.

- [ ] Yes
- [x] No

**Other information**:


powerfactory/powerfactory-converter/src/main/java/com/powsybl/powerfactory/converter/PowerFactoryImporter.java
```diff
@@ -70,11 +70,19 @@ public class PowerFactoryImporter implements Importer {
             Boolean.FALSE
     );
 
+    public static final Parameter IMPORT_GEODATA_PARAMETER = new Parameter(
```
zamarrenolm, line 73: We should write some documentation for this new parameter. The user must be aware that geodata can be imported. I agree with the default value set to false (focus on analysis data first).

marqueslanauja: Done

zamarrenolm (approved):

PR #3962, opened 2026-06-15
Merge networks: improve exception message when duplicates are found

### LINKED ISSUE #3721: Specify which networks contain the same element when merging more than 2 networks with duplicate elements

As we can see Network.merge(String id, Network... networks) can takes multiple networks (which means it's possible to merge 3 networks with this.
When merging three networks, I get the error message:
`com.powsybl.commons.PowsyblException: The following object(s) of type GeneratorImpl exist(s) in both networks: [GEN]`
However, since I have three networks, I do not know which networks are referenced by "both networks".
This error is obtained with the following code:
```
Network network1 = NetworkTest1Factory.create();
Network network2 = EurostagTutorialExample1Factory.create();
Network network3 = SecurityAnalysisTestNetworkFactory.create();
Network fullNetwork = Network.merge("full", network1, network2, network3);
```

Modify the error message of `com.powsybl.iidm.network.impl.NetworkImpl.checkMergeability` so that it says which two networks that are being checked; for example by giving the ID of the two networks.

### Describe the motivation

Make it clearer which networks are in conflict when performing a merge (as this issue could also arise when merging more than 3 networks, in which case it could quickly become difficult to find which combination of two networks is causing the issue)

### Extra Information

_No response_

zamarrenolm:

Fixes #3721 

- [X] Yes, for custom IIDM implementation maintainers only
- [ ] No

The error message when elements with the same ID are found during network merge was changed to indicate which networks are concerned.
- Before: `The following object(s) of type <type> exist(s) in both networks: [<duplicate_ids>]`
- After: `The following object(s) of type <type> already exist in merged network when trying to merge network '<network_id_2>': [<duplicate_ids>]`

Since some TCK unit tests check this message, custom IIDM implementation maintainers should adapt there merge method to have the same error message.

**Other information**:


iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractMergeNetworkTest.java
```diff
@@ -616,7 +616,7 @@ public void failMergeWithCommonAreaConflict() {
 
         // Merge should fail, because area "bza" is present in both network but with different attribute values
         PowsyblException e = assertThrows(PowsyblException.class, () -> Network.merge(n1, n2));
-        assertEquals("The following object(s) of type AreaImpl exist(s) in both networks: [bza]", e.getMessage());
+        assertEquals("The following object(s) of type AreaImpl exist(s) in networks [n1+n2] and [n2] : [bza]", e.getMessage());
```
olperr1: In the previous message, the `[` and `]` were used to indicate a list (the duplicate ids one).
It would be more consistent to use other separators to introduce the network ids, for instance `"` ou `'`.

olperr1 (changes requested):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
```diff
@@ -1298,8 +1298,10 @@ private void checkMergeability(NetworkImpl otherNetwork) {
             Class<? extends Identifiable> clazz = entry.getKey();
             Collection<String> objs = entry.getValue();
             if (!objs.isEmpty()) {
+                String thisNetworkId = this.getId();
+                String otherNetworkId = otherNetwork.getId();
                 throw new PowsyblException("The following object(s) of type "
-                        + clazz.getSimpleName() + " exist(s) in both networks: "
+                        + clazz.getSimpleName() + " exist(s) in networks [" + thisNetworkId + "] and [" + otherNetworkId + "] : "
```
olperr1: Could you also remove the space character before `:`?

olperr1 (changes requested):

iidm/iidm-impl/src/test/java/com/powsybl/iidm/network/impl/MergeTest.java
```diff
@@ -28,6 +30,18 @@
...
+    @Test
+    void merge3NetworksWithDuplicates() {
+        Network network1 = NetworkTest1Factory.create();
+        Network network2 = EurostagTutorialExample1Factory.create();
+        Network network3 = SecurityAnalysisTestNetworkFactory.create();
+        PowsyblException x = assertThrows(PowsyblException.class, () -> {
+            Network.merge("full", network1, network2, network3);
+        });
+        // Assert that the message contains the identifiers of networks where duplicates are found
+        assertTrue(x.toString().contains("exist(s) in networks [full] and [fictitious]"));
```
NathanDissoubray: "full" is the name of the network we are trying to merge into. I would expect the error message to give ids for the networks that are merged (so ids of network1, network2 or network3), not for the network that is created

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractMergeNetworkTest.java
```diff
@@ -57,7 +57,7 @@ public void failMergeWithSameObj() {
         addSubstation(n1, "P1");
         addSubstation(n2, "P1");
         PowsyblException e = assertThrows(PowsyblException.class, () -> Network.merge(n1, n2));
-        assertEquals("The following object(s) of type SubstationImpl exist(s) in both networks: [P1]", e.getMessage());
+        assertEquals("The following object(s) of type SubstationImpl exist(s) in networks [n1+n2] and [n2] : [P1]", e.getMessage());
```
NathanDissoubray: why not "in networks [n1] and [n2]" ?

NathanDissoubray (changes requested):

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractMergeNetworkTest.java
olperr1: This is directly linked to the way the merge works: it creates a new network, then merge the networks one by one in it.
The steps are therefore the following:
- create a new empty network "n1+n2"
- merge "n1" in "n1+n2"
- merge "n2" in "n1+n2"

NathanDissoubray: but that means that the error message gives you the name of the destination network, not the name of the two source networks in which the element is duplicate. That's actually more confusing

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractMergeNetworkTest.java
zamarrenolm: done

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
zamarrenolm: done

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractMergeNetworkTest.java
zamarrenolm: One way to solve it without changing the overall merge strategy would be to "look back" (individually) all the duplicated elements found in the previously merged networks and report the current one and the ones where they are found ... That would require a little bit more of work, and there could be potentially different messages for each of the "found" duplicated elements.

For the current release maybe we could stay with this solution and discuss improvements. 

For this release what we could also do is to report only the "other network", and avoid reporting the destination network in the error message.

What do you think ?

NathanDissoubray: > One way to solve it without changing the overall merge strategy would be to "look back" (individually) all the duplicated elements found in the previously merged networks and report the current one and the ones where they are found ... That would require a little bit more of work, and there could be potentially different messages for each of the "found" duplicated elements.

What you do you mean by "look back" ? Memorize the elements and find them in the networks that are merged ?
For me, the ideal message would be something like that:
When you merge n1 and n2 into "full", get a message like "battery myBatteryName exists in both network1Name and network2Name"

> For this release what we could also do is to report only the "other network", and avoid reporting the destination network in the error message.

It's a good idea, since the destination network is created for the merge it's more confusing to have a duplicate element in a network that starts empty.

zamarrenolm: For the "ideal" message to be put, we would need to introduce that relatively big change on the processing of duplicates.

I've pushed a change in the message to show only the "source" network where duplicates are found, during the merge process.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
```diff
@@ -1298,8 +1298,10 @@ private void checkMergeability(NetworkImpl otherNetwork) {
             Class<? extends Identifiable> clazz = entry.getKey();
             Collection<String> objs = entry.getValue();
             if (!objs.isEmpty()) {
-                throw new PowsyblException("The following object(s) of type "
-                        + clazz.getSimpleName() + " exist(s) in both networks: "
+                String thisNetworkId = this.getId();
```
NathanDissoubray: can you remove this please ?

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractMergeNetworkTest.java
```diff
@@ -57,7 +57,7 @@ public void failMergeWithSameObj() {
         addSubstation(n1, "P1");
         addSubstation(n2, "P1");
         PowsyblException e = assertThrows(PowsyblException.class, () -> Network.merge(n1, n2));
-        assertEquals("The following object(s) of type SubstationImpl exist(s) in both networks: [P1]", e.getMessage());
+        assertEquals("The following duplicated object(s) of type SubstationImpl were found when trying to merge network 'n2': [P1]", e.getMessage());
```
NathanDissoubray: should it specify that the duplicate is with another network of the merge ? Because as it stands, it could be interpreted as "the network 'fictitious' contains duplicate elements"

NathanDissoubray (changes requested):

iidm/iidm-tck/src/test/java/com/powsybl/iidm/network/tck/AbstractMergeNetworkTest.java
zamarrenolm: What about writing the message this way:

`The following object(s) of type " + clazz.getSimpleName() + " already exist in merged network when trying to merge network '" + otherNetworkId + "': " + objs`

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/NetworkImpl.java
zamarrenolm: removed, sorry

NathanDissoubray (approved):

olperr1 (approved):

PR #3919, opened 2026-05-18
Use of channels for biidm and network copy

clementleclercRTE:
**Please check if the PR fulfills these requirements**

- [X] The commit message follows our guidelines
- [X] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] A PR or issue has been opened in all impacted repositories (if any)

**Does this PR already have an issue describing the problem?**

Fixes #3881

**What kind of change does this PR introduce?**

 Migrates the binary reader and writer from `DataInputStream`/`DataOutputStream` over
  `BufferedInputStream`/`BufferedOutputStream` to direct `ByteBuffer` + `java.nio.channels.ReadableByteChannel`/`WritableByteChannel`.

**What is the current behavior?**

  - `BinReader` wraps the caller's `InputStream` into a `BufferedInputStream` + `DataInputStream`.
  - `BinWriter` wraps the caller's `OutputStream` into a `BufferedOutputStream` + `DataOutputStream`.
  - `NetworkSerDe.copy()` uses `PipedInputStream`/`PipedOutputStream`.
  - When the underlying source is a `Files.newInputStream(path)` / `Files.newOutputStream(path)` (themselves wrapping a `FileChannel`), the data path goes `Channel → Stream → DataStream`, with
  redundant adapters.

**What is the new behavior (if this is a feature change)?**

- `BinReader` now reads through a `BufferedChannelReader` backed by a direct `ByteBuffer` and a `ReadableByteChannel`.
  - `BinWriter` now stages body bytes in a direct `ByteBuffer` (`GrowingByteBuffer`) and drains them straight to a `WritableByteChannel` on close.
  - New public constructors `BinReader(ReadableByteChannel, byte[])` / `BinReader(Path, byte[])` and `BinWriter(WritableByteChannel, byte[], String)` / `BinWriter(Path, byte[], String)` avoid the
  `Channel → Stream → Channel` double-wrapping when the caller already has a channel (or a `Path`).
  - `NetworkSerDe.read(Path, …)` and `NetworkSerDe.write(Network, ExportOptions, Path)` now dispatch on format: for BIN, they open a `Files.newByteChannel(path)` directly and feed the channel
  constructors — no `OutputStream`/`InputStream` adapter.
[...]

PR #3968, opened 2026-06-18
Fix use of channels for biidm and network copy

flo-dup:

No, but issues brought by #3919 were written [there](https://github.com/powsybl/powsybl-core/pull/3919#pullrequestreview-4525324306) 

Bug fix

- Duplicate code + code smell for Network copy
- Reading direct memory buffer size oversized
- isEndOfStream method mainly added for unit tests
- GrowingByteBuffer badly named

- Network copy code unified
- Reading direct memory buffer size reduced
- isEndOfStream method removed, specific method added to detect end of stream when reading next name index
- GrowingByteBuffer renamed

- [ ] Yes
- [x] No

### DIFF
--- a/commons/src/main/java/com/powsybl/commons/binary/BinReader.java
+++ b/commons/src/main/java/com/powsybl/commons/binary/BinReader.java
@@ -94,10 +94,7 @@ public class BinReader extends AbstractTreeDataReader {
     }
 
     private int readNameIndex() {
-        if (in.isEndOfStream()) {
-            return END_OF_FILE;
-        }
-        return in.readUnsignedShort();
+        return in.readOptionalUnsignedShort();
     }
 
     private boolean isAttrAbsent(String name) {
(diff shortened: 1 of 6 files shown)


flo-dup:
Edited the PR description: `GrowingByteBuffer` is not allocating direct memory, but heap memory (it's using ByteBuffer::allocate method). The performance are good in the benchmarks, and that is explained by several reasons I'll detail just after

flo-dup:
Two arguments to keep the the GrowingByteBuffer (renamed to SegmentedByteBuffer) instead of a ByteArrayOutputStream:
- It writes primitives directly to the buffer, whereas the proposed logic with ByteArrayOutputStream first writes to a small direct buffer then copies it again to the ByteArrayOutputStream. Besides that copy is checking the bounds, the capacity, then clearing the small buffer.
- When growing it just adds a new buffer to the list, whereas the ByteArray doubles the size of the backing array and copies the previous content

olperr1 (approved):

PR #3969, opened 2026-06-18
Biidm format natively ZSTD-compressed

flo-dup:

No

Feature

The biidm file for a given 8000 voltage levels network weighs about 30Mb

The biidm file for a given 8000 voltage levels network weighs about 6Mb

- [ ] Yes
- [x] No

### DIFF
--- a/commons/src/main/java/com/powsybl/commons/binary/BinReader.java
+++ b/commons/src/main/java/com/powsybl/commons/binary/BinReader.java
@@ -7,6 +7,7 @@
  */
 package com.powsybl.commons.binary;
 
+import com.github.luben.zstd.ZstdInputStream;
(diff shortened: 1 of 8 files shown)


rolnico:
It would be interesting to compare the performances with and without this compression (for example using powsybl-benchmark, as it contains a serialization benchmark)

flo-dup:
I did compare the performances on several real network files, and I couldn't see much difference in computation time, maybe a few percent changes with the compression level chosen.

flo-dup:
Indeed the best would be to use powsybl-benchmark to compare it for different compression levels, here I only tested it with my own benchmark test

geofjamg (dismissed):

rolnico (approved):

olperr1 (approved):

PR #3973, opened 2026-06-24
Apply fix for GHSA-jqvf-j3ww-r8c7

olperr1:
> [!CAUTION]
> This PR should be merged manually with ff-only

The purpose of this PR is to run the CI on the advisory fix before merging it in main.


rolnico (approved):

olperr1:
OK for the coverage. It is computed on an Unix OS, but the code detected as "non-covered" is the one for Windows (which is covered by Windows unit test, running only on Windows).

PR #3981, opened 2026-06-28
Fix split for time series with gaps

### LINKED ISSUE #3941: [TimeSeries] split handle non-successive chunks inconsistently

`split()` works for some chunk size but fails for others when chunks are not successive.

Example
```yaml
chunk1: offset = 0, values = [1, 2, 3]
chunk2: offset = 6, values = [7, 8]
# timeSeries
timeSeries (chunk1, chunk2): [1d, 2d, 3d, NaN, NaN, NaN, 7d, 8d]
```
- `split(2)` works and returns non successive chunks:
```yaml
offset = 0, values = [1, 2]
offset = 2, values = [3]
offset = 6, values = [7, 8]
```
These chunks are not successive: after offset 2 with size 1, the next expected offset should be 3, not 6.

- `split(3)` works also.

- but `split(4)` fails with:

```shell
java.lang.IllegalArgumentException: Chunks are not successive. First offset is 0 and first size is 3; second offset should be 3but is 6
```

The behavior should be consistent.

Either:

- non successive chunks are supported, and split() should work for all valid split sizes, or
- non successive chunks are not supported, and split() should fail consistently.

### Describe the steps

1. Create a time series with two non successive chunks: 
Example
```java
DoubleDataChunk chunk1 = new UncompressedDoubleDataChunk(0, new double[]{1d, 2d, 3d});
DoubleDataChunk chunk2 = new UncompressedDoubleDataChunk(6, new double[]{7d, 8d});
StoredDoubleTimeSeries timeSeries = new StoredDoubleTimeSeries(metadata, chunk1, chunk2);
```
2. Call split(2)
 this works
3. Call split(3)
 this works also
4. Call split(4) 
this fails

### Environment

Version: 7.2.1

### Relevant Log Output

_No response_

### Extra Information

_No response_

Chandu0501:
Fixes #3941

## Summary
- avoid merging a partial output chunk with the next chunk when there is a gap between their offsets
- keep sparse/non-successive chunks split consistently for all valid chunk sizes
- add regression coverage for `StoredDoubleTimeSeries.split(4)` on a time series containing a gap

## Verification
- `./mvnw.cmd -pl time-series/time-series-api -am "-Dtest=StoredDoubleTimeSeriesTest" "-Dsurefire.failIfNoSpecifiedTests=false" test`

### DIFF
--- a/time-series/time-series-api/src/main/java/com/powsybl/timeseries/AbstractTimeSeries.java
+++ b/time-series/time-series-api/src/main/java/com/powsybl/timeseries/AbstractTimeSeries.java
@@ -170,7 +170,10 @@ public abstract class AbstractTimeSeries<P extends AbstractPoint, C extends Data
         C previousChunk = splitChunks.isEmpty() ? null : splitChunks.get(splitChunks.size() - 1);
         int previousChunkSize = 0;
         //We can complete the previous chunk if 1) it is uncomplete and 2) the current offset is not a multiple of newChunkSize
-        if (previousChunk != null && previousChunk.getLength() < newChunkSize && chunkToSplit.getOffset() % newChunkSize != 0) {
+        if (previousChunk != null
+                && previousChunk.getLength() < newChunkSize
+                && previousChunk.getOffset() + previousChunk.getLength() == chunkToSplit.getOffset()
+                && chunkToSplit.getOffset() % newChunkSize != 0) {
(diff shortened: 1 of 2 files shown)


So-Fras:
Hi again @Chandu0501 👋
Is your PR ready to be reviewed? If this is the case, could you take it out from the draft mode? Thanks a lot!

PR #3986, opened 2026-07-01
Improve NodeCalcPrinter performance

### LINKED ISSUE #878: NodeCalcPrinter O(n^2) performance
bug

NodeCalcPrinter O(n^2) performance:
```
print tree with 1000 nodes 19 ms.
print tree with 2000 nodes 44 ms.
print tree with 4000 nodes 46 ms.
print tree with 8000 nodes 249 ms.
print tree with 16000 nodes 801 ms.
print tree with 32000 nodes 2284 ms.
print tree with 64000 nodes 8611 ms.
```

```
    private void timetester(int n) {
        long start = System.currentTimeMillis();
        NodeCalc node = new IntegerNodeCalc(0);
        for (int i = 0; i < n; i++) {
            node = BinaryOperation.plus(new IntegerNodeCalc(0), node);
        }
        NodeCalcPrinter.print(node);
        long end = System.currentTimeMillis();
        System.out.println("print tree with " + n  + " nodes " + (end - start) + " ms.");
    }
    
    @Test
    public void test1() {
        timetester(1000);
        timetester(2000);
        timetester(4000);
        timetester(8000);
        timetester(16000);
        timetester(32000);
        timetester(64000);
    }
```

O(n) performance

faster

samirromdhani:

Fixes #878 

- [ ] Yes
- [x] No

**Other information**:

##### Benchmark
```shell
Benchmark                   Mode  Cnt     Score   Error  Units
     Time
                    print    avgt    2  1046,929          us/op   <- before performance improvement
                    print    avgt    2   225,151          us/op   <- after performance improvement
      Memory
                    print    avgt    2  12216808,768      B/op  <- before performance improvement
                    print    avgt    2    403712,167      B/op  <- after performance improvement
```


time-series/time-series-api/src/test/java/com/powsybl/timeseries/NodeCalcTooManyRecursionExceptionTest.java
```diff
@@ -53,4 +54,24 @@ void testRight() {
...
         }
     }
+
+    @Test
+    void shouldPrintDeepTreeInLinearTime() {
+        long t1 = timeToPrint(100_000);
+        long t2 = timeToPrint(200_000);
+        double ratio = (double) t2 / t1;
+        assertTrue(ratio < 3);
+    }
```
NathanDissoubray: I'm not sure that this test is reliable, especially in a CI

NathanDissoubray (changes requested):
It seems to me that this is still a quadratic complexity. This changes the time constant by using a string builder instead of concatenating strings (which is faster), but it seems to me that it does the same kind of operation (thus the complexity would be the same)
Feel free to correct me if I missed something that is different

MatthieuSAUR:
> It seems to me that this is still a quadratic complexity. This changes the time constant by using a string builder instead of concatenating strings (which is faster), but it seems to me that it does the same kind of operation (thus the complexity would be the same) Feel free to correct me if I missed something that is different

Before this PR
print tree with  1000 nodes:     0.401 ms -   400.676 (time/n) -     0.401 (time/n2)
print tree with  2000 nodes:     1.427 ms -   713.585 (time/n) -     0.357 (time/n2)
print tree with  4000 nodes:     4.013 ms -  1003.374 (time/n) -     0.251 (time/n2)
print tree with  8000 nodes:    15.453 ms -  1931.622 (time/n) -     0.241 (time/n2)
print tree with 16000 nodes:    72.808 ms -  4550.515 (time/n) -     0.284 (time/n2)
print tree with 32000 nodes:   306.190 ms -  9568.422 (time/n) -     0.299 (time/n2)
print tree with 64000 nodes:  1406.612 ms - 21978.313 (time/n) -     0.343 (time/n2)
 
With this PR
print tree with  1000 nodes:     0.462 ms -   461.658 (time/n) -     0.462 (time/n2)
print tree with  2000 nodes:     0.494 ms -   247.147 (time/n) -     0.124 (time/n2)
print tree with  4000 nodes:     1.637 ms -   409.166 (time/n) -     0.102 (time/n2)
print tree with  8000 nodes:     3.406 ms -   425.690 (time/n) -     0.053 (time/n2)
print tree with 16000 nodes:     3.527 ms -   220.448 (time/n) -     0.014 (time/n2)
print tree with 32000 nodes:     8.673 ms -   271.026 (time/n) -     0.008 (time/n2)
print tree with 64000 nodes:    15.721 ms -   245.636 (time/n) -     0.004 (time/n2)

NathanDissoubray:
Thanks @MatthieuSAUR for running the test, it's indeed linear, I'll have to take a look at the PR more closely to understand why that's the case (seems like it's not just using a string builder that makes it faster)

time-series/time-series-api/src/test/java/com/powsybl/timeseries/NodeCalcTooManyRecursionExceptionTest.java
samirromdhani: I agree, test removed

time-series/time-series-api/src/test/java/com/powsybl/timeseries/NodeCalcTooManyRecursionExceptionTest.java
```diff
@@ -53,4 +53,5 @@ void testRight() {
             fail(e);
         }
     }
+
```
NathanDissoubray:
```suggestion
```

time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/Printable.java
```diff
@@ -0,0 +1,17 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+package com.powsybl.timeseries.ast;
+
+import java.util.Deque;
+
+/**
+ * @author Samir Romdhani {@literal <samir.romdhani at rte-france.com>}
+ */
+public interface Printable {
```
NathanDissoubray: I would rename this to not overlap with the exisiting [java.awt.print.Printable](https://docs.oracle.com/en/java/javase/21/docs/api/java.desktop/java/awt/print/Printable.html)

time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/NodeCalcPrinter.java
```diff
@@ -9,83 +9,131 @@
...
+        Deque<Object> stack = new ArrayDeque<>();
+        stack.push(root);
+
+        while (!stack.isEmpty()) {
+            Object node = stack.pop();
+            if (node instanceof String text) {
+                builder.append(text);
+            } else if (node instanceof Printable printable) {
+                printable.pushTo(stack);
+            }
```
NathanDissoubray, line 35: To avoid having the instanceof cast, how about:
- pushTo takes the stack and the builder
- add to `Printable`
```java
static Printable builderAppend(String s) {
        return (stack, builder) -> builder.append(s);
    }
```
- in the `visit`
  - if you only need to add a string, directly append the string to the builder instead of adding the string to the stack
  - if you need to push nodes, use the `builderAppend` to wrap the string
```java
public Printable visit(BinaryOperation nodeCalc, Void arg, Printable left, Printable right) {
    return (stack, builder) -> {
        stack.push(builderAppend(")"));
        stack.push(right);
        stack.push(builderAppend(" " + nodeCalc.getOperator() + " "));
        stack.push(left);
        stack.push(builderAppend("("));
    };
}
```
I'm not sure if this would be equivalent (or slower / faster) or take more memory, but it would make the stack of a single type (and not a mix of String and Printable).
Could you test this and see what results you have in time / memory ?

time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/Printable.java
NathanDissoubray: You can add the `@FunctionalInterface` annotation

NathanDissoubray (changes requested):

time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/NodeCalcPrinter.java
samirromdhani: This way still linear but the mixed String / Printable stack is slightly faster and use less memory than wrapping every plain string in builderAppend.

samirromdhani: Technically passing the builder introduce extra lambda allocations.

NathanDissoubray: I was wondering if not having all the cast would gain time, but it seems like the extra-allocation are not worth it, thanks for testing

time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/Printable.java
samirromdhani: Updated to `NodeCalcPrintable`

samirromdhani: done

NathanDissoubray (dismissed):

time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/NodeCalcPrinter.java
olperr1, line 1: Could you add a few comments to explain the main idea of your algorithm and why the computation is linear now?

olperr1 (changes requested):

time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/NodeCalcPrinter.java
samirromdhani: Done, I added comments explaining how the algorithm avoids repeated string copying

time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/NodeCalcPrinter.java
```diff
@@ -23,6 +23,9 @@ public static String print(NodeCalc nodeCalc) {
         Deque<Object> stack = new ArrayDeque<>();
         stack.push(root);
 
+        // Build the output iteratively, each printable pushes its children and its expression parts onto the stack in reverse order,
+        // Then process the stored expressions in their printing order.
+        // Unlike recursive string concatenation, this does not repeatedly copy the strings of entire subtree, resulting linear time complexity.
```
NathanDissoubray: I think it could go a bit more into the details, as I think that would be clearer. To my understanding, it goes from quadratic to linear because:
a
copy to
ab
copy to
abc

Meaning you need to allocate and copy the start of the string multiple times. If n is the length of the final string, a is copied n times, b is copied n-1 times, c n-2 times... And $\sum_{i=1}^n i = \frac{n(n-1)}2$ so a quadratic time

But the string builder doesn't need to copy, since it's a growing buffer, so each part of the string is written only once -> linear time

samirromdhani: The main idea is to use an iterative approach instead of a recursive one.

Before this PR there is a recursive string concatenation (the strings are not stored, there are repeatedly copied)
With this PR, the string concatenation is handled iteratively using a stack, so we stored the expressions parts in a stack which makes it easier to process them once.

NathanDissoubray: I think even if you use a stack but don't use a string builder, the performance won't be good. The fact you don't need to repeatedly copy is thanks to using the builder to process the stack in linear time (otherwise you would still need to perform copies)

samirromdhani: Yes, there is something that could be improved regarding my comments ?

NathanDissoubray: your comment only mentions the stack but not the importance of the String builder. We could have a stack but still repeatedly copy strings (which would result in a quadratic performance), so I think it would be good to mention the string builder as well.

samirromdhani: Mentionned the use of StringBuilder

time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/NodeCalcPrinter.java
```diff
@@ -9,83 +9,134 @@
...
     public static String print(NodeCalc nodeCalc) {
-        return nodeCalc.accept(new NodeCalcPrinter(), null, 0);
+        NodeCalcPrintable root = nodeCalc.accept(new NodeCalcPrinter(), null, 0);
+        StringBuilder builder = new StringBuilder();
+        Deque<Object> stack = new ArrayDeque<>();
+        stack.push(root);
+
+        // Build the output iteratively using a StringBuilder, each printable pushes its children and its expression parts onto the stack in reverse order,
+        // Then process the stored expressions in their printing order, and the String parts are appended once to the shared StringBuilder,
+        // Unlike recursive string concatenation, this does not repeatedly copy the strings of entire subtree, resulting linear time complexity.
```
NathanDissoubray:
```suggestion
        // Unlike recursive string concatenation, this does not repeatedly copy the strings of entire subtree, resulting in linear time complexity.
```

time-series/time-series-api/src/main/java/com/powsybl/timeseries/ast/NodeCalcPrinter.java
olperr1: Thanks!

NathanDissoubray (approved):

MatthieuSAUR (approved):

olperr1 (approved):

PR #3995, opened 2026-07-06
Allow print a Matrix with a specified value of decimal places

### LINKED ISSUE #3543: Allowing to print a Matrix with a specified value of decimal places

When calling one of the `Matrix.print(...)` methods, the values are printed with a lot of decimals, which may be not easily read.

When printing a `Matrix`, it should be possible (but not mandatory) to precise the number of decimal places to use to render the `double` values.

### Describe the motivation

To improve values readability.

### Extra Information

_No response_

samirromdhani:

Fixes #3543

- [X] Yes
- [ ] No

If you have defined implementations of the `Matrix` interface, you must implement the following methods:
- `void print(PrintStream out, List<String> rowNames, List<String> columnNames, PrintConfig config)`
- `void print(PrintStream out, PrintConfig config)`

**Other information**:


math/src/main/java/com/powsybl/math/matrix/DenseMatrix.java
```diff
@@ -431,12 +452,21 @@ public void print(PrintStream out, List<String> rowNames, List<String> columnNam
                 out.print(Strings.padStart(rowNames.get(i), rowNamesWidth + 1, ' '));
             }
             for (int j = 0; j < getColumnCount(); j++) {
-                out.print(Strings.padStart(Double.toString(get(i, j)), width[j] + 1, ' '));
+                if (defaultDecimalDigits.isPresent()) {
+                    out.print(Strings.padStart(decimalFormat.format(get(i, j)), width[j] + 1, ' '));
+                } else {
+                    out.print(Strings.padStart(Double.toString(get(i, j)), width[j] + 1, ' '));
+                }
```
alicecaron:
```suggestion
out.print(Strings.padStart(
      DEFAULT_DECIMAL_DIGITS.isPresent() ? decimalFormat.format(get(i, j)) : Double.toString(get(i, j)),
      width[j] + 1,
      ' '
));
```

math/src/main/java/com/powsybl/math/matrix/DenseMatrix.java
```diff
@@ -460,6 +490,19 @@ private int[] getMaxWidthForEachColumn(List<String> columnNames) {
         return width;
     }
 
+    private int[] getMaxWidthForEachColumn(List<String> columnNames, DecimalFormat decimalFormat) {
```
alicecaron:
```suggestion
    private int[] getMaxWidthPerColumn(List<String> columnNames, DecimalFormat decimalFormat) {
```

math/src/main/java/com/powsybl/math/matrix/DenseMatrix.java
samirromdhani: done

math/src/main/java/com/powsybl/math/matrix/DenseMatrix.java
samirromdhani: done

alicecaron (dismissed):

samirromdhani:
Hello @alicecaron ,

To document the added configuration, I was thinking of adding a new Math or Matrix section under the existing Configuration/Modules documentation: https://powsybl.readthedocs.io/projects/powsybl-core/en/stable/user/configuration/modules.html, we can document the current optional configuration:

```yaml
matrix:
  print-decimal-digits: 10 
```
Or if we expect math related configuration in the future, we can introduce math as a parent module:
```yaml
math:
  matrix:
    print-decimal-digits: 10
```
What do you think ?

math/src/main/java/com/powsybl/math/matrix/MatrixConfig.java
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * @author Samir Romdhani {@literal <samir.romdhani at rte-france.com>}
+ */
+public final class MatrixConfig {
+
+    private final OptionalInt printDecimalDigits;
+
+    private MatrixConfig(OptionalInt decimalDigits) {
+        this.printDecimalDigits = decimalDigits;
+    }
```
jeandemanged: - "decimal places" is I think better understandable than "decimal digits"
- using Optional as field or parameter is generally discouraged. Optional is mean to be used only as return type. Use rather a `Integer` here
- there should be some validation performed, negative values should not be accepted - or should fallback to empty/null with warning

math/src/main/java/com/powsybl/math/matrix/DenseMatrix.java
```diff
@@ -31,6 +33,19 @@ public class DenseMatrix extends AbstractMatrix {
...
+        this.matrixConfig = matrixConfig;
+    }
+
+    protected MatrixConfig getConfig() {
+        if (matrixConfig == null) {
+            matrixConfig = MatrixConfig.load();
+        }
+        return matrixConfig;
+    }
+
```
jeandemanged: I find it odd to store the config in the matrix. I think it would be best to provide the config to the print method via new overloads taking the config as parameter. Existing print methods without argument would just use PlatformConfig.load().

math/src/main/java/com/powsybl/math/matrix/DenseMatrix.java
```diff
@@ -431,12 +452,21 @@ public void print(PrintStream out, List<String> rowNames, List<String> columnNam
...
+                        width[j] + 1,
+                        ' '
+                ));
             }
             out.println();
         }
     }
 
+    private static DecimalFormat createFormatter(int maxDecimals) {
+        String pattern = "0.0" + "#".repeat(Math.max(0, maxDecimals - 1));
```
jeandemanged: hence we have always at least one decimal place, worth mentioning in the documentation

jeandemanged:
> Hello @alicecaron ,
> 
> To document the added configuration, I was thinking of adding a new Math or Matrix section under the existing Configuration/Modules documentation: https://powsybl.readthedocs.io/projects/powsybl-core/en/stable/user/configuration/modules.html, we can document the current optional configuration:
> 
> ```yaml
> matrix:
>   print-decimal-digits: 10 
> ```
> 
> Or if we expect math related configuration in the future, we can introduce math as a parent module:
> 
> ```yaml
> math:
>   matrix:
>     print-decimal-digits: 10
> ```
> 
> What do you think ?

The new module configuration must indeed be added in the documentation. I do not think we support "parent" modules in powsybl config YAML, there is only a single module key and single property key, your second example would not work, even if you name your module `math.matrix`. Taking this into account, `math-matrix` could be a better module name then.

math/src/main/java/com/powsybl/math/matrix/DenseMatrix.java
olperr1: I agree with @jeandemanged: it is weird to store the print config inside the matrix.

I suggest the following, to complete/detail Damien's comment:
- rename `MatrixConfig` in `PrintConfig`
- in the `Matrix` interface:
  - add the following signatures:
    - `void print(PrintStream out, List<String> rowNames, List<String> columnNames, PrintConfig config)`
    - `void print(PrintStream out, PrintConfig config)`
  - add a default implementations for the 2 existing `print(...)` methods: they should call their counterpart having a `PrintConfig` argument by passing `PrintConfig.load()` for the extra argument.
- add the `PrintConfig` argument to the `print` method implementations of `SparseMatrix` and `DenseMatrix`.

olperr1:
> The new module configuration must indeed be added in the documentation. I do not think we support "parent" modules in powsybl config YAML, there is only a single module key and single property key, your second example would not work, even if you name your module `math.matrix`. Taking this into account, `math-matrix` could be a better module name then.

`math-matrix` is fine for me.

math/src/main/java/com/powsybl/math/matrix/DenseMatrix.java
samirromdhani: Added to the documentation

math/src/main/java/com/powsybl/math/matrix/DenseMatrix.java
samirromdhani: Implemented, thanks for the suggestion.

math/src/main/java/com/powsybl/math/matrix/MatrixConfig.java
samirromdhani: I agree with the naming "decimal places"

samirromdhani: Other suggestions done.

docs/user/configuration/math-matrix.md
```diff
@@ -0,0 +1,27 @@
+# math-matrix
+The `math-matrix` module is used to configure the matrix print formatting.
+
+- `print-decimal-places` (optional integer, `>= 0`): maximum number of decimal places.
+- If missing: default `Double.toString` formatting is used.
+- If invalid (`< 0`): ignored, with a warning.
```
olperr1: Could you adopt the same formatting as in the other modules' pages?
(add a "Required properties" title, use the property name as a subsection title, ...)

math/src/main/java/com/powsybl/math/matrix/AbstractMatrix.java
```diff
@@ -60,4 +61,16 @@ public Matrix add(Matrix other) {
...
+        if (matrixConfig != null && matrixConfig.getPrintDecimalPlaces() != null) {
+            return createFormatter(matrixConfig.getPrintDecimalPlaces());
+        }
+        return null;
+    }
+
+    protected static DecimalFormat createFormatter(int maxDecimals) {
+        String pattern = "0.0" + "#".repeat(Math.max(0, maxDecimals - 1));
+        return new DecimalFormat(pattern);
+    }
```
olperr1, line 74: These methods can be defined in `Printconfig` instead.

math/src/main/java/com/powsybl/math/matrix/DenseMatrix.java
```diff
@@ -408,14 +409,27 @@ public void removeSmallValues(double epsilonValue) {
...
     @Override
     public void print(PrintStream out, List<String> rowNames, List<String> columnNames) {
-        int rowNamesWidth = getMaxWidthAmongRowNames(rowNames);
+        print(out, rowNames, columnNames, PrintConfig.load());
+    }
+
+    @Override
+    public void print(PrintStream out, PrintConfig config) {
+        print(out, null, null, config);
+    }
```
olperr1: The implementation of these methods can be defined in `AbstractMatrix` to avoid code duplication.

math/src/main/java/com/powsybl/math/matrix/DenseMatrix.java
```diff
@@ -460,6 +478,19 @@ private int[] getMaxWidthForEachColumn(List<String> columnNames) {
...
+        for (int i = 0; i < getRowCount(); i++) {
+            for (int j = 0; j < getColumnCount(); j++) {
+                width[j] = Math.max(width[j], decimalFormat.format(get(i, j)).length());
+                if (columnNames != null) {
+                    width[j] = Math.max(width[j], columnNames.get(j).length());
+                }
+            }
+        }
+        return width;
+    }
```
olperr1: These 2 methods have the same algorithm. They can be reworked to have a single method instead.

math/src/main/java/com/powsybl/math/matrix/SparseMatrix.java
```diff
@@ -522,4 +537,11 @@ public static SparseMatrix read(Path file) {
             throw new UncheckedIOException(e);
         }
     }
+
+    private TDoubleArrayListHack formatValues(DecimalFormat decimalFormat) {
+        for (int i = 0; i < values.size(); i++) {
+            values.set(i, Double.parseDouble(decimalFormat.format(values.get(i))));
+        }
+        return values;
+    }
```
olperr1: This method should not change the content of `values`. Could you return an array of Strings instead?

math/src/main/java/com/powsybl/math/matrix/Matrix.java
```diff
@@ -297,4 +295,23 @@ interface ElementHandler {
      * @param out the stream
      */
     void print(PrintStream out);
+
+    /**
+     * Print the matrix to a stream including the row, the column and the print configuration
```
olperr1:
```suggestion
     * Print the matrix to a stream including the row and the column, using the given print configuration
```

math/src/main/java/com/powsybl/math/matrix/Matrix.java
```diff
@@ -297,4 +295,23 @@ interface ElementHandler {
...
+     *
+     * @param out         the stream
+     * @param rowNames    the row names
+     * @param columnNames the column names
+     * @param config      the print configuration
+     */
+    void print(PrintStream out, List<String> rowNames, List<String> columnNames, PrintConfig config);
+
+    /**
+     * Print the matrix to a stream including the print configuration
```
olperr1:
```suggestion
     * Print the matrix to a stream, using the given print configuration
```

math/src/test/java/com/powsybl/math/matrix/AbstractMatrixTest.java
```diff
@@ -476,4 +491,60 @@ void testTranspose() {
...
+    }
+
+    @Test
+    void printWhenDecimalPlacesPropertyNotDefined() throws IOException {
+        Matrix matrix = getMatrixFactory().create(1, 3, 1);
+        matrix.set(0, 0, 1.0 / 3.0);
+        matrix.set(0, 1, Math.PI);
+        matrix.set(0, 2, 2);
+        PlatformConfig platformConfig = new PlatformConfig(name -> Optional.empty(), null);
+        String result = print(matrix, null, null, PrintConfig.load(platformConfig));
```
olperr1: If you set the `PrintConfig` constructor public, you won't have to create a `PlatformConfig` object to have a `PrintConfig` object.

math/src/test/java/com/powsybl/math/matrix/AbstractMatrixTest.java
```diff
@@ -476,4 +491,60 @@ void testTranspose() {
...
+        matrix.set(0, 1, Math.PI);
+        matrix.set(0, 2, 2);
+        matrix.set(0, 3, 1.23456);
+        String result = print(matrix);
+        assertThat(result).contains("0.33")
+                .doesNotContain(Double.toString(1.0 / 3.0));
+        assertThat(result).contains("3.14")
+                .doesNotContain(Double.toString(Math.PI));
+        assertThat(result).contains("2.0");
+        assertThat(result).contains("1.235");
```
olperr1:
```suggestion
        assertThat(result)
                .contains("0.33").doesNotContain("0.3333")
                .contains("3.14").doesNotContain("3.1415").doesNotContain("3.1416")
                .contains("2.0").doesNotContain("2.00");
                .contains("1.235").doesNotContain("1.2345");
```

math/pom.xml
```diff
@@ -96,6 +96,10 @@
             <artifactId>junit-jupiter</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+        </dependency>
```
olperr1, line 103: This dependency is only needed for the tests!
```suggestion
        <dependency>
            <groupId>org.assertj</groupId>
            <artifactId>assertj-core</artifactId>
            <scope>test</scope>
        </dependency>
```

olperr1 (changes requested):

math/src/test/java/com/powsybl/math/matrix/AbstractMatrixTest.java
samirromdhani: This will introduce adding the same verification on negative value in the constructor too. I think it is not needed to make the constructor public if only for testing purposes.

olperr1: You can also set it in package-protected scope, but we can keep it like this.

math/src/main/java/com/powsybl/math/matrix/SparseMatrix.java
samirromdhani: Replaced by
```java
    public String formatValues(DecimalFormat decimalFormat) {
        return Arrays.stream(getValues())
                .mapToObj(decimalFormat::format)
                .collect(Collectors.joining(", ", "{", "}"));
    }
```
Here return just a String, i found the array of Strings  does not match the expected print result.

math/src/main/java/com/powsybl/math/matrix/PrintConfig.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public static PrintConfig load() {
+        return load(PlatformConfig.defaultConfig());
+    }
+
+    public Integer getPrintDecimalPlaces() {
+        return printDecimalPlaces;
+    }
+
+    public static DecimalFormat getFormatter(PrintConfig config) {
```
olperr1: It would be clearer to return an `optional<DecimalFormat>`.

In `DenseMatrix.print(...)`, you can then:
- compute `width` with:
```java
        DoubleFunction<String> formattingFunction = getFormatter(config)
                .map(decimalFormat -> (DoubleFunction<String>) (decimalFormat::format))
                .orElse(Double::toString);
        int[] width = getMaxWidthPerColumn(columnNames, formattingFunction);
```
- use `formattingFunction.apply(get(i, j))` to print the values.

olperr1 (changes requested):

math/src/main/java/com/powsybl/math/matrix/PrintConfig.java
samirromdhani: Good point, done.

olperr1 (approved):

jeandemanged (approved):

PR #3984, opened 2026-06-30
Bump IIDM to 1.18

NathanDissoubray:
**Please check if the PR fulfills these requirements**

- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] A PR or issue has been opened in all impacted repositories (if any)

**Does this PR already have an issue describing the problem?**

None

**What kind of change does this PR introduce?**

Feature

**What is the current behavior?**

IIDM in 1.17

**What is the new behavior (if this is a feature change)?**

IIDM in 1.18

**Does this PR introduce a breaking change or deprecate an API?**
- [x] Yes
- [ ] No

**If yes, please check if the following requirements are fulfilled**

- [x] The *Breaking Change* or *Deprecated* label has been added
- [x] The migration steps are described in the following section

**What changes might users need to make in their application due to this PR? (migration steps)**

Downstream projects should:

- Check in their versionable extensions' serializers/deserializers (classes extending `AbstractVersionableNetworkExtensionSerDe`) that a `SerDeVersion` value support the new version (1.18) (usually, at least a `SerDeVersion` has its `maxIidmVersionExcluded` attribute set to `null`);

- Upgrade the version in their XIIDM / JIIDM unit test files corresponding to the current version:
    - For ".xiidm" or ".xml" files: replace:
        - `xmlns:iidm="http://www.powsybl.org/schema/iidm/1_17"` with `xmlns:iidm="http://www.powsybl.org/schema/iidm/1_18"`;
        - `xmlns:iidm="http://www.powsybl.org/schema/iidm/equipment/1_17"` with `xmlns:iidm="http://www.powsybl.org/schema/iidm/equipment/1_18"`;

    - For ".json" or ".jiidm" files: replace `"version" : "1.17"` with `"version" : "1.18"`.

**Other information**:

PR #4013, opened 2026-07-23
Make LoadType optional in the equipment .xsd files of the network.

### LINKED ISSUE #3687: The `LoadType`attribute is optional in the IIDM xsd but required in the equipment xsd

According to the IIDM XSD (for all versions), the load type is optional. However, it is required at equipment level.
When attempting to set it to null in an SSH-level network, a ValidationException is raised.

I believe the XSD file should be updated to ensure this attribute is mandatory, in line with the Equipment XSD and validation checks.
Is a new version of IIDM required for that? 

### Describe the steps

_No response_

### Environment

_No response_

### Relevant Log Output

_No response_

### Extra Information

_No response_

colineplqt:

Fixes #3687

### DIFF
--- a/docs/grid_model/network_subnetwork.md
+++ b/docs/grid_model/network_subnetwork.md
@@ -238,8 +238,8 @@ A load is a passive equipment representing a delivery point that consumes or pro
     - Consumptions are positive.
 
 **Metadata**
-In the grid model, loads comprise the following metadata:
-- The load type, which can be:
+In the grid model, the following metadata is included for loads:
+- The load type, which is required and by default `UNDEFINED`. It can be:
     - `UNDEFINED`
     - `AUXILIARY`
     - `FICTITIOUS`
--- a/docs/grid_model/network_subnetwork.md
+++ b/docs/grid_model/network_subnetwork.md
@@ -239,10 +239,11 @@ A load is a passive equipment representing a delivery point that consumes or pro
 
 **Metadata**
 In the grid model, the following metadata is included for loads:
-- The load type, which is required and by default `UNDEFINED`. It can be:
+- The load type, which can be:
     - `UNDEFINED`
     - `AUXILIARY`
     - `FICTITIOUS`
+By default, it is `UNDEFINED`.
 - The load model, which can be:
     - `ZIP` (or polynomial), following equations:
 
--- a/computation-local/src/main/java/com/powsybl/computation/local/UnixLocalCommandExecutor.java
+++ b/computation-local/src/main/java/com/powsybl/computation/local/UnixLocalCommandExecutor.java
@@ -42,13 +42,19 @@ public class UnixLocalCommandExecutor extends AbstractLocalCommandExecutor {
         for (Map.Entry<String, String> entry : env2.entrySet()) {
             String name = entry.getKey();
             String value = entry.getValue();
+            // A variable name sits before the '=' assignment and after '$', so it cannot be
+            // single-quoted like the value. Reject anything but alphanumerics and underscores
+            // to keep a hostile name from breaking out of the command.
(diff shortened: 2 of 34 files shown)


colineplqt:
@olperr1, I open this pull request to fix https://github.com/powsybl/powsybl-core/issues/3687. The validation indicates that the `LoadType` is required, and it is impossible to create a SSH network with a `Load` that has a `null` `LoadType`. The equipment level XSD files also indicates that the `LoadType` is required. I thus fix all the XSD files, can you confirm that it is the right way to do it? Thanks

NathanDissoubray (dismissed):
Checked if there would be no problem with older PowSyBl versions. The code requiring the `LoadType` exists since at least April 2016 (see this commit https://github.com/powsybl/powsybl-core/commit/18c88baec92608ea3ea6aad6c70312207a30eb4c#diff-b5458f134a4bd47978b42eb6a1f26642e8c3872bd66d3321c8625c719b66f383R37), so no issue there.
Will add it in the [IIDM 1.18 bump](https://github.com/powsybl/powsybl-core/pull/3984) as well

olperr1:
> @olperr1, I open this pull request to fix #3687. The validation indicates that the `LoadType` is required, and it is impossible to create a SSH network with a `Load` that has a `null` `LoadType`. The equipment level XSD files also indicates that the `LoadType` is required. I thus fix all the XSD files, can you confirm that it is the right way to do it? Thanks

It's a little tricky...
The XSDs describe the XIIDM format, and the load type can be absent of the file. In that case, the `UNDEFINED` value is used by the SerDe:
https://github.com/powsybl/powsybl-core/blob/4e8024eaf07a43673c68226aefc8b57dae4c4ffb/iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/LoadSerDe.java#L88

Also note that in the adder, the type can be omitted; in that case, the default value applies. But if the type is specified, the value must be valid... and `null` is not.

I wonder if we shouldn't change the equipment XSDs instead, and set the loadType as `optional` in them...

olperr1:
I've also checked the differences between the 2 XSDs of the IIDM 1.17. This attribute is the only one being declared as required in the equipment level and optional in the ssh one.

olperr1:
@colineplqt: The IIDM 1.18 PR is now merged in main.
Could you update this PR with main and apply the same changes to the IIDM 1.18 XSDs?

colineplqt:
> @colineplqt: The IIDM 1.18 PR is now merged in main. Could you update this PR with main and apply the same changes to the IIDM 1.18 XSDs?

Done! On my side, the PR is ready.

olperr1 (dismissed):

docs/grid_model/network_subnetwork.md
```diff
@@ -238,11 +238,12 @@ A load is a passive equipment representing a delivery point that consumes or pro
     - Consumptions are positive.
 
 **Metadata**
-In the grid model, loads comprise the following metadata:
+In the grid model, the following metadata is included for loads:
 - The load type, which can be:
     - `UNDEFINED`
     - `AUXILIARY`
     - `FICTITIOUS`
+By default, it is `UNDEFINED`.
```
NathanDissoubray, line 247:
```suggestion

By default, it is `UNDEFINED`.
```
Without newline
Image

And with the newline

Image

NathanDissoubray (changes requested):

docs/grid_model/network_subnetwork.md
colineplqt: Thanks!

NathanDissoubray (approved):

olperr1 (approved):

PR #4019, opened 2026-07-27
[CGMES] CGM quick export also writes TP file 

colineplqt:

- [ ] Docs have been added / updated (for bug fixes / features) **Documentation to do once we agree on what should be done**

No

Arguable

CGM quick export only exports SSh for each IGM and a common SV.

CGM quick export exports SSH and TP for each IGM and a common SV. This is to take into account the export of CGMs with node/breaker topology: the TP is an output and not an input in that case.

- [ ] Yes
- [ ] No

**Other information**:


colineplqt:
I am not sure this is the right approach. As the SV is dependent on the TP profiles, users cannot discard the exported TPs and only use the SSHs and SV as before. I think it is good for node/breaker export but I wonder if there should be an option, or a check performed on the network topology (if one node/breaker voltage level, then export also the TP).

zamarrenolm:
Checking the documentation about CGM building (latest is available at ENTSO-E CGMES Library: https://eepublicdownloads.entsoe.eu/clean-documents/CIM_documents/Grid_Model_CIM/CGM%20BUILDING%20PROCESS%20IMPLEMENTATION%20GUIDE_v2.0.pdf), after assembling a CGM model the only outputs for export should be the common SV and the list of (eventually updated) IGM SSHs. Assembling a CGM should not modify the topology, so there should be no need to export TPs.

If we want to support TP export on CGM I agree with you that it should be configurable, and disabled by default, to follow the specification.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -213,6 +213,7 @@ private void exportCGM(Network network, DataSource dataSource, CgmesExportContex
         for (Network subnetwork : network.getSubnetworks()) {
             IgmModelsForCgm igmModelsForCgm = new IgmModelsForCgm(
                     initializeModelForExport(subnetwork, CgmesSubset.STEADY_STATE_HYPOTHESIS, context, false, true),
+                    initializeModelForExport(subnetwork, CgmesSubset.TOPOLOGY, context, false, true),
```
zamarrenolm, line 218: be aware that the model for the TOPOLOGY subset is initialized also below with different flags

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -227,13 +229,18 @@ private void exportCGM(Network network, DataSource dataSource, CgmesExportContex
             updateDependenciesCGM(igmModels.values(), updatedCgmSvModel, context.getBoundaryTpId());
         }
 
-        // Export the SSH for the IGMs and the SV for the CGM
+        // Export the SSH and TP for the IGMs and the SV for the CGM
```
zamarrenolm: TP optionally:
```suggestion
        // Export the SSH and optionally the TP for the IGMs and the SV for the CGM
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -340,9 +348,16 @@ private void updateDependenciesCGM(Collection<IgmModelsForCgm> igmModels, CgmesM
         igmModels.forEach(m -> m.updatedSsh.clearSupersedes());
         igmModels.forEach(m -> m.updatedSsh.addSupersedes(m.originalSsh.getId()));
 
-        // Updated SV model depends on updated SSH models and original TP and TP_BD models
+        // Each TP model depends on the original EQ model and on the original TP_BP model
```
zamarrenolm: There was a typo in the original comment:
```suggestion
        // Each TP model depends on the original EQ model and on the original TP_BD model
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -561,15 +577,15 @@ private static class IgmModelsForCgm {
         CgmesMetadataModel updatedSsh;
         CgmesMetadataModel originalEq;
         CgmesMetadataModel originalSsh;
-        CgmesMetadataModel originalTp;
+        CgmesMetadataModel originalOrUpdatedTp;
```
zamarrenolm: I would prefer to have two separate models here: `originalTp` and `updatedTp`, the same way this is done for SSH.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -561,15 +577,15 @@ private static class IgmModelsForCgm {
...
         CgmesMetadataModel originalTpBd;
 
         IgmModelsForCgm(CgmesMetadataModel updatedSsh, CgmesMetadataModel originalEq, CgmesMetadataModel originalSsh,
-                               CgmesMetadataModel originalTp, CgmesMetadataModel originalTpBd) {
+                               CgmesMetadataModel originalOrUpdatedTp, CgmesMetadataModel originalTpBd) {
             this.updatedSsh = updatedSsh;
             this.originalEq = originalEq;
             this.originalSsh = originalSsh;
-            this.originalTp = originalTp;
+            this.originalOrUpdatedTp = originalOrUpdatedTp;
```
zamarrenolm: Receive  and keep `updatedTp` and `originalTp` separately.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -227,13 +229,18 @@ private void exportCGM(Network network, DataSource dataSource, CgmesExportContex
...
         String baseName = getBaseName(context, dataSource, network);
         for (Network subnetwork : network.getSubnetworks()) {
             String country = getCountry(subnetwork);
             String igmName = country != null ? country : subnetwork.getNameOrId();
             String igmSshFileName = baseName + "_" + igmName + "_" + CgmesSubset.STEADY_STATE_HYPOTHESIS.getIdentifier() + ".xml";
             subsetExport(subnetwork, CgmesSubset.STEADY_STATE_HYPOTHESIS, igmSshFileName, dataSource, context, igmModels.get(subnetwork).updatedSsh);
+
+            if (exportTP) {
+                String igmTpFileName = baseName + "_" + igmName + "_" + CgmesSubset.TOPOLOGY.getIdentifier() + ".xml";
+                subsetExport(subnetwork, CgmesSubset.TOPOLOGY, igmTpFileName, dataSource, context, igmModels.get(subnetwork).originalOrUpdatedTp);
```
zamarrenolm: Use `updatedTp` here. Because we have separated `originalTp` and `updatedTp`, this line will be read in a less confusing way, following the same scheme used for SSH.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -340,9 +348,16 @@ private void updateDependenciesCGM(Collection<IgmModelsForCgm> igmModels, CgmesM
         igmModels.forEach(m -> m.updatedSsh.clearSupersedes());
         igmModels.forEach(m -> m.updatedSsh.addSupersedes(m.originalSsh.getId()));
 
-        // Updated SV model depends on updated SSH models and original TP and TP_BD models
+        // Each TP model depends on the original EQ model and on the original TP_BP model
+        igmModels.forEach(m -> m.originalOrUpdatedTp.addDependentOn(m.originalEq.getId()));
```
zamarrenolm: In the comment we say that updated TP depends on original TP_BD but we are not explicitly adding the dependency. It will be present only if there was an extension created from original CGMES import. I think we could add it explicitly (check if it is duplicated in the output). I think we can add something like `m.updatedTp.addDependentOn(m.originalTpBd.getId())`.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -340,9 +348,16 @@ private void updateDependenciesCGM(Collection<IgmModelsForCgm> igmModels, CgmesM
...
+
+        // Each updated TP model supersedes the original one
+        igmModels.forEach(m -> m.originalOrUpdatedTp.clearSupersedes());
+        igmModels.forEach(m -> m.originalOrUpdatedTp.addSupersedes(m.originalOrUpdatedTp.getId()));
```
zamarrenolm: We can always prepare the `updatedTp` superseding the `originalTp`. Because we have split the two models, the intention here will be more clear.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -340,9 +348,16 @@ private void updateDependenciesCGM(Collection<IgmModelsForCgm> igmModels, CgmesM
...
+
+        // Updated SV model depends on updated SSH models, original TP_BD models and origianl or updated TP models
```
zamarrenolm: Typo:
```suggestion
        // Updated SV model depends on updated SSH models, original TP_BD models and original or updated TP models
```

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -340,9 +348,16 @@ private void updateDependenciesCGM(Collection<IgmModelsForCgm> igmModels, CgmesM
...
         updatedCgmSvModel.addDependentOn(igmModels.stream().map(m -> m.updatedSsh.getId()).collect(Collectors.toSet()));
-        updatedCgmSvModel.addDependentOn(igmModels.stream().map(m -> m.originalTp.getId()).collect(Collectors.toSet()));
+        updatedCgmSvModel.addDependentOn(igmModels.stream().map(m -> m.originalOrUpdatedTp.getId()).collect(Collectors.toSet()));
```
zamarrenolm: Here we should add original or updated TP as a dependency, depending on the flag exportTP. Again, the intention will become more clear if we separate the two models and explicitly use the flag.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CommonGridModelExportTest.java
```diff
@@ -580,6 +579,46 @@ void testFaraoUseCaseManualExport() throws IOException {
...
+        assertEquals("2021-02-03T04:30:00Z", getFirstMatch(beTpXml, REGEX_SCENARIO_TIME));
+        assertEquals("2021-02-03T04:30:00Z", getFirstMatch(nlTpXml, REGEX_SCENARIO_TIME));
+
+        // Profiles should be consistent with the instance files
+        assertEquals("http://entsoe.eu/CIM/SteadyStateHypothesis/1/1", getFirstMatch(updatedBeSshXml, REGEX_PROFILE));
+        assertEquals("http://entsoe.eu/CIM/SteadyStateHypothesis/1/1", getFirstMatch(updatedNlSshXml, REGEX_PROFILE));
+        assertEquals("http://entsoe.eu/CIM/StateVariables/4/1", getFirstMatch(updatedCgmSvXml, REGEX_PROFILE));
+        assertEquals("http://entsoe.eu/CIM/Topology/4/1", getFirstMatch(beTpXml, REGEX_PROFILE));
+        assertEquals("http://entsoe.eu/CIM/Topology/4/1", getFirstMatch(nlTpXml, REGEX_PROFILE));
+
```
zamarrenolm: We could check also that the dependencies are correct.

zamarrenolm (changes requested):
I think the code will be more clear if we separate `originalTp` and `updatedTP` models as two different attributes in `IgmModelsForCgm`.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
colineplqt: Actually I think the code I wrote there did not work. You are right, it is much clearer with keeping the original and updated profiles separated! And also more correct.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -325,24 +333,41 @@ public static CgmesMetadataModel initializeModelForExport(
...
 
         // Each updated SSH model supersedes the original one
         igmModels.forEach(m -> m.updatedSsh.clearSupersedes());
         igmModels.forEach(m -> m.updatedSsh.addSupersedes(m.originalSsh.getId()));
 
-        // Updated SV model depends on updated SSH models and original TP and TP_BD models
+        // Each updated TP model depends on the original EQ model and on the original TP_BD model
+        igmModels.forEach(m -> m.updatedTp.addDependentOn(m.originalEq.getId()));
+        if (boundaryTpId != null) {
+            igmModels.forEach(m -> m.updatedTp.addDependentOn(m.originalEq.getId()));
```
zamarrenolm: It should be:
```suggestion
            igmModels.forEach(m -> m.updatedTp.addDependentOn(boundaryTpId));
```

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CommonGridModelExportTest.java
```diff
@@ -580,6 +579,70 @@ void testFaraoUseCaseManualExport() throws IOException {
...
+        assertEquals(originalNlTpId, getFirstMatch(updatedNlTpXml, REGEX_SUPERSEDES));
+
+        // When CGM_EXPORT_WITH_TP is false (the default), no TP files should be exported
+        Properties exportParamsNoTp = new Properties();
+        exportParamsNoTp.put(CgmesExport.CGM_EXPORT, true);
+        String basenameNoTp = "test_cgm_without_tp";
+        network.write("CGMES", exportParamsNoTp, tmpDir.resolve(basenameNoTp));
+        assertFalse(Files.exists(tmpDir.resolve(basenameNoTp + "_BE_TP.xml")));
+        assertFalse(Files.exists(tmpDir.resolve(basenameNoTp + "_NL_TP.xml")));
+    }
```
zamarrenolm: I propose to add a test with explicit identifier for the boundaries, to also check the small fix mentioned above, when setting the dependencies. We can reuse your code for checking the output:

```suggestion
        // Perform a CGM export with TP profile enabled and check the exported files
        Properties exportParamsTp = new Properties();
        exportParamsTp.put(CgmesExport.CGM_EXPORT, true);
        exportParamsTp.put(CgmesExport.CGM_EXPORT_WITH_TP, true);
        String basenameTp = "test_cgm_with_tp";
        network.write("CGMES", exportParamsTp, tmpDir.resolve(basenameTp));
        checkCGMExportWithTp(basenameTp, Optional.empty());

        // Perform a CGM export with TP profile enabled and explicit boundary TP id and read the exported files
        String explicitBoundaryTpBdId = "myBoundaryTpId";
        Properties exportParamsTpTpBdExplicit = new Properties();
        exportParamsTpTpBdExplicit.put(CgmesExport.CGM_EXPORT, true);
        exportParamsTpTpBdExplicit.put(CgmesExport.CGM_EXPORT_WITH_TP, true);
        exportParamsTpTpBdExplicit.put(CgmesExport.BOUNDARY_TP_ID, explicitBoundaryTpBdId);
        String basenameTpTpBdExplicit = "test_cgm_with_tp_and_tp_bd_explicit";
        network.write("CGMES", exportParamsTpTpBdExplicit, tmpDir.resolve(basenameTpTpBdExplicit));
        checkCGMExportWithTp(basenameTpTpBdExplicit, Optional.of(explicitBoundaryTpBdId));

        // When CGM_EXPORT_WITH_TP is false (the default), no TP files should be exported
        Properties exportParamsNoTp = new Properties();
        exportParamsNoTp.put(CgmesExport.CGM_EXPORT, true);
        String basenameNoTp = "test_cgm_without_tp";
        network.write("CGMES", exportParamsNoTp, tmpDir.resolve(basenameNoTp));
        assertFalse(Files.exists(tmpDir.resolve(basenameNoTp + "_BE_TP.xml")));
        assertFalse(Files.exists(tmpDir.resolve(basenameNoTp + "_NL_TP.xml")));
    }

    private void checkCGMExportWithTp(String basename, Optional<String> explicitTpBdId) throws IOException {
        String updatedBeSshXml = Files.readString(tmpDir.resolve(basename + "_BE_SSH.xml"));
        String updatedNlSshXml = Files.readString(tmpDir.resolve(basename + "_NL_SSH.xml"));
        String updatedCgmSvXml = Files.readString(tmpDir.resolve(basename + "_SV.xml"));
        String updatedBeTpXml = Files.readString(tmpDir.resolve(basename + "_BE_TP.xml"));
        String updatedNlTpXml = Files.readString(tmpDir.resolve(basename + "_NL_TP.xml"));

        // Scenario time should be the same for all models
        assertEquals("2021-02-03T04:30:00Z", getFirstMatch(updatedBeSshXml, REGEX_SCENARIO_TIME));
        assertEquals("2021-02-03T04:30:00Z", getFirstMatch(updatedNlSshXml, REGEX_SCENARIO_TIME));
        assertEquals("2021-02-03T04:30:00Z", getFirstMatch(updatedCgmSvXml, REGEX_SCENARIO_TIME));
        assertEquals("2021-02-03T04:30:00Z", getFirstMatch(updatedBeTpXml, REGEX_SCENARIO_TIME));
        assertEquals("2021-02-03T04:30:00Z", getFirstMatch(updatedNlTpXml, REGEX_SCENARIO_TIME));

        // Profiles should be consistent with the instance files
        assertEquals("http://entsoe.eu/CIM/SteadyStateHypothesis/1/1", getFirstMatch(updatedBeSshXml, REGEX_PROFILE));
        assertEquals("http://entsoe.eu/CIM/SteadyStateHypothesis/1/1", getFirstMatch(updatedNlSshXml, REGEX_PROFILE));
        assertEquals("http://entsoe.eu/CIM/StateVariables/4/1", getFirstMatch(updatedCgmSvXml, REGEX_PROFILE));
        assertEquals("http://entsoe.eu/CIM/Topology/4/1", getFirstMatch(updatedBeTpXml, REGEX_PROFILE));
        assertEquals("http://entsoe.eu/CIM/Topology/4/1", getFirstMatch(updatedNlTpXml, REGEX_PROFILE));

        // Dependency check
        // The updated TPs should depend on the original EQ model and on the original TP_BD model
        // The updated SV model should depend on the updated TP models
        // Here the version number part of the id 1 for original models and 2 for updated ones
        String originalBeEqId = "urn:uuid:Network_BE_N_EQUIPMENT_2021-02-03T04:30:00Z_1_1D__FM";
        String originalNlEqId = "urn:uuid:Network_NL_N_EQUIPMENT_2021-02-03T04:30:00Z_1_1D__FM";
        String originalBeTpBdId = "urn:uuid:Network_BE_N_TOPOLOGY_BOUNDARY_2021-02-03T04:30:00Z_1_1D__FM";
        String originalNlTpBdId = "urn:uuid:Network_NL_N_TOPOLOGY_BOUNDARY_2021-02-03T04:30:00Z_1_1D__FM";
        Set<String> expectedDependenciesBeTp = Set.of(originalBeEqId, explicitTpBdId.orElse(originalBeTpBdId));
        assertEquals(expectedDependenciesBeTp, getUniqueMatches(updatedBeTpXml, REGEX_DEPENDENT_ON));
        Set<String> expectedDependenciesNlTp = Set.of(originalNlEqId, explicitTpBdId.orElse(originalNlTpBdId));
        assertEquals(expectedDependenciesNlTp, getUniqueMatches(updatedNlTpXml, REGEX_DEPENDENT_ON));

        String updatedBeTpId = "urn:uuid:Network_BE_N_TOPOLOGY_2021-02-03T04:30:00Z_2_1D__FM";
        String updatedNlTpId = "urn:uuid:Network_NL_N_TOPOLOGY_2021-02-03T04:30:00Z_2_1D__FM";
        assertTrue(getUniqueMatches(updatedCgmSvXml, REGEX_DEPENDENT_ON).contains(updatedBeTpId));
        assertTrue(getUniqueMatches(updatedCgmSvXml, REGEX_DEPENDENT_ON).contains(updatedNlTpId));

        // TP should supersede the original model
        String originalBeTpId = "urn:uuid:Network_BE_N_TOPOLOGY_2021-02-03T04:30:00Z_1_1D__FM";
        String originalNlTpId = "urn:uuid:Network_NL_N_TOPOLOGY_2021-02-03T04:30:00Z_1_1D__FM";
        assertEquals(originalBeTpId, getFirstMatch(updatedBeTpXml, REGEX_SUPERSEDES));
        assertEquals(originalNlTpId, getFirstMatch(updatedNlTpXml, REGEX_SUPERSEDES));
    }

```

zamarrenolm (changes requested):

zamarrenolm (approved):

rcourtier:
I think we should export only one TP for the CGM.

The SV contains the result of the power flow calculation. It is an output, there is no updated version of it (unlike SSH profiles), and there is only one for the whole CGM.
Likewise, the TP contains the result of the topology processing. It is an output (in that case where IGMs EQ inputs files are in node-breaker topology level), there is no updated version of it (unlike SSH profiles) and there should be only one for the whole CGM.

zamarrenolm:
> I think we should export only one TP for the CGM.
> 
> The SV contains the result of the power flow calculation. It is an output, there is no updated version of it (unlike SSH profiles), and there is only one for the whole CGM. Likewise, the TP contains the result of the topology processing. It is an output (in that case where IGMs EQ inputs files are in node-breaker topology level), there is no updated version of it (unlike SSH profiles) and there should be only one for the whole CGM.

Agree, although this feature is being added to deal for an exceptional situation. As stated in previous comments, assembling a CGM should not modify the topology, so there should be no need to export any result of topology processing. I think the output (global TP or TP by IGM) should be provided in the way that best fits the use-case at hand.

zamarrenolm:
Maybe we should add a configuration export to let the user decide if a single TP (for the whole CGM) or multiple TPs (one for each IGM) should be output.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
```diff
@@ -325,29 +364,59 @@ public static CgmesMetadataModel initializeModelForExport(
...
+        } else {
+            updatedCgmSvModel.addDependentOn(igmModels.stream().map(m -> m.originalTp.getId()).collect(Collectors.toSet()));
+        }
         if (boundaryTpId != null) {
             updatedCgmSvModel.addDependentOn(boundaryTpId);
         } else {
             updatedCgmSvModel.addDependentOn(igmModels.stream().map(m -> m.originalTpBd.getId()).collect(Collectors.toSet()));
         }
+
+        // Updated TP model depends on the original EQ model and on the original TP_BD model
```
zamarrenolm, line 404: I think we should make TP dependent on SSH (original or updated) and TP_BD, not directly the EQ. You can check this in the ENTSO-E test configurations we keep in cgmes-conformity module. Examples: 
- In relicap_grid IGM Belgovia, its TP depends on its SSH.
- In relicap_grid, the common TP depends on all updated SSHs.
- In cgmes3-test-models/MicroGrid the ASSEMBLED TP depends on the updated SSHs.

checks for MicroGrid:
```bash
cgmes3-test-models/MicroGrid % grep DependentOn *ASSE*TP*
    <md:Model.DependentOn rdf:resource="urn:uuid:3a2a0416-c6a8-49a9-9f23-b8ec7cb29bae" />
    <md:Model.DependentOn rdf:resource="urn:uuid:1e239780-b5f9-4ed1-ad73-041bc406bcfc" />

cgmes3-test-models/MicroGrid % grep 1e239780-b5f9-4ed1-ad73-041bc406bcfc *|grep -v DependentOn
20210209T1930Z_1D_NL_SSH_9.xml:  <md:FullModel rdf:about="urn:uuid:1e239780-b5f9-4ed1-ad73-041bc406bcfc">

cgmes3-test-models/MicroGrid % grep 3a2a0416-c6a8-49a9-9f23-b8ec7cb29bae *|grep -v DependentOn
20210209T1930Z_1D_BE_SSH_9.xml:  <md:FullModel rdf:about="urn:uuid:3a2a0416-c6a8-49a9-9f23-b8ec7cb29bae">
```

zamarrenolm (changes requested):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesExport.java
colineplqt: I am not sure, in QoCDC, the rule TPEQTPEQDIFF which is an error states that TP should depend on an EQ or EQDIFF file.

zamarrenolm: You are right, there is that explicit rule in the QoCDC. 

Although it is not respected in neither the MicroGrid not the relicap-grid examples. They only contain explicit DependenOn references to the updated SSH models, that in turn depend on the EQ.

The CGM implementation guide v2.0 () offers some complementary info:
- The figure 8 on page 39 depicts only the dependency TP -> EQ,
- But the figures 9 and 10 on pages 39 and 40 seem to indicate the additional TP -> SSH dependent on relationship.

Agree on keeping the TP->EQ dependency to comply with QoCDC. 

For me it is interesting to keep showing that this TP has been built based on information contained in the corresponding SSH (like the examples in relicap-grid and MicroGrid show). So adding additional dependencies to SSH files would clearly state that. (QoCDC allows to add more dependencies, the TP->EQ is the minimum one, required to resolve the instance references).

zamarrenolm (approved):
Agree on the change about exported TP follow the QoCDC rule for dependency TP -> EQ.
We should investigate the "generic way" of respecting the rule stated for CGMES 3.0 in the QoCDC.

rcourtier (approved):

PR #4020, opened 2026-07-28
[CGMES] Use network ref to generate UUIDs for OperationalLimitType and LoadGroup

colineplqt:

No

Bug fix.

When exporting the EQ for each IGM of a CGM, some equipment objects are exported with the same UUID for each IGM (LoadGroup, OperationalLimitType). These duplicates can be an issue.

The naming strategy now takes into account the network for the export of these objects. 

- [ ] Yes
- [ ] No

**Other information**:


colineplqt:
@zamarrenolm @marqueslanauja this PR is for discussion, feel free to challenge it.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
```diff
@@ -98,8 +99,9 @@ public void setFictitiousContainerFor(Identifiable<?> id, String containerId) {
         fictitiousContainers.put(id.getId(), containerId);
     }
 
-    public CgmesExportContext() {
+    public CgmesExportContext() { //TODO: delete?
```
zamarrenolm: You can label it as deprecated and then remove it in the next release

zamarrenolm (changes requested):
It would be nice to have a unit test that checks that the exported IGMs from a CGM now have separate ids for the LoadGroup, OperationalLimitType objects.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
colineplqt: Do you think it would make sense to have a constructor that only takes the network instead? I am not sure how useful it is to have a constructor that creates a context with a null network then.

zamarrenolm: For me it is perfectly correct to delete the empty constructor, but because it is declared `public` we should not remove it directly in this release. First we have to deprecate it. I propose to label it as deprecated now and then remove it in the next release.

To be consistent with the deprecation, we should remove the test `emptyConstructor` in the module `CgmesExportContextTest.java`. We can modify the `getSet` test in the same module to use a dummy network:

```java
    @Test
    void getSet() {
        Network network = Network.create("dummy", "test");
        CgmesExportContext context = new CgmesExportContext(network)
``` 

These are the only points in the powsybl code base where the empty constructor is used.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CommonGridModelExportTest.java
```diff
@@ -708,6 +708,94 @@ private void addModelsForSubnetworks(Network network, int version) {
...
+                .add();
+        // Add a line with current limits within the subnetwork
+        Substation s1b = network1.newSubstation().setId("Substation_BE2").setCountry(Country.BE).add();
+        VoltageLevel vl1b = s1b.newVoltageLevel().setId("VL_BE2").setNominalV(400.0).setTopologyKind(TopologyKind.BUS_BREAKER).add();
+        vl1b.getBusBreakerView().newBus().setId("Bus_BE2").add();
+        network1.newLine().setId("Line_BE")
+                .setVoltageLevel1("VL_BE").setBus1("Bus_BE")
+                .setVoltageLevel2("VL_BE2").setBus2("Bus_BE2")
+                .setR(1.0).setX(10.0).setG1(0.0).setB1(0.0).setG2(0.0).setB2(0.0).add()
+                .newCurrentLimits1().setPermanentLimit(1000.0).add();
```
zamarrenolm: You could create the limits using `newOperationalLimitsGroup1` to avoid the deprecated `newCurrentLimits1`.

Also, we could add temporary limits to extend a little bit the coverage (the network `ref` has been added to both types, permanent and temporary).

My suggestion (also create the same temporary limits for NL network):

```suggestion
                .newOperationalLimitsGroup1("OLG_BE_1")
                .newCurrentLimits()
                .setPermanentLimit(1000.0)
                .beginTemporaryLimit()
                .setName("OLG_BE_T_1")
                .setAcceptableDuration(10)
                .setValue(1100.0)
                .endTemporaryLimit()
                .beginTemporaryLimit()
                .setName("OLG_BE_T_2")
                .setAcceptableDuration(5)
                .setValue(1200.0)
                .endTemporaryLimit()
                .add();
```

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CommonGridModelExportTest.java
```diff
@@ -708,6 +708,94 @@ private void addModelsForSubnetworks(Network network, int version) {
...
+        // Each EQ file must contain one OperationalLimitType
+        Pattern regexOperationalLimitType = Pattern.compile("cim:OperationalLimitType rdf:ID=\"(_[^\"]*)\"");
+        Set<String> beOpLimitTypeIds = getUniqueMatches(beEqXml, regexOperationalLimitType);
+        Set<String> nlOpLimitTypeIds = getUniqueMatches(nlEqXml, regexOperationalLimitType);
+        assertFalse(beOpLimitTypeIds.isEmpty());
+        assertFalse(nlOpLimitTypeIds.isEmpty());
+
+        // The LoadGroup IDs and OperationalLimitType IDs must be different between the two IGMs
+        assertNotEquals(beLoadGroupIds, nlLoadGroupIds);
+        assertNotEquals(beOpLimitTypeIds, nlOpLimitTypeIds);
```
zamarrenolm: If we add temporary limits in the test network we will end up having operational limit types. 
If we "forget" to add the network `ref` only to TATLs, the sets will be different, but both will contain common types (you can check it using the ref only for PATLs).
So here we should check that the intersection of the two sets is empty, they do not have any common element:

```suggestion
        assert(Collections.disjoint(beLoadGroupIds, nlLoadGroupIds));
        assert(Collections.disjoint(beOpLimitTypeIds, nlOpLimitTypeIds));
```

zamarrenolm (changes requested):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
colineplqt: Sounds like a good strategy to me! I will do that. Thank you

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CommonGridModelExportTest.java
```diff
@@ -708,6 +709,107 @@ private void addModelsForSubnetworks(Network network, int version) {
...
+                .add();
+        // Add a line with current limits within the subnetwork
+        Substation s2b = network2.newSubstation().setId("Substation_NL2").setCountry(Country.NL).add();
+        VoltageLevel vl2b = s2b.newVoltageLevel().setId("VL_NL2").setNominalV(400.0).setTopologyKind(TopologyKind.BUS_BREAKER).add();
+        vl2b.getBusBreakerView().newBus().setId("Bus_NL2").add();
+        network2.newLine().setId("Line_NL")
+                .setVoltageLevel1("VL_NL").setBus1("Bus_NL")
+                .setVoltageLevel2("VL_NL2").setBus2("Bus_NL2")
+                .setR(1.0).setX(10.0).setG1(0.0).setB1(0.0).setG2(0.0).setB2(0.0).add()
+                .newCurrentLimits1().setPermanentLimit(1000.0).add();
```
zamarrenolm: Please add a block using `newOperationalLimitsGroup1` and temporary limits with the same scheme used the "BE" network. 

This way we will remove the sonar warning about using deprecated `newCurrentLimits1` and force two temporary limit types that could end up having the same id.

The PR ensures that temporary limit types also receive the network ref, so they are exported without potential duplications. With this addition to test we will covering also that the change for temporary limits is correct.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/export/CgmesExportContext.java
```diff
@@ -98,8 +99,10 @@ public void setFictitiousContainerFor(Identifiable<?> id, String containerId) {
         fictitiousContainers.put(id.getId(), containerId);
     }
 
+    @Deprecated(since = "7.4.0")
```
zamarrenolm: We can get rid of the sonar warning updating the deprecation tag like this:
```suggestion
    /**
     * Creates an export context that is not bound to any network.
     *
     * @deprecated Use {@link #CgmesExportContext(Network)} instead.
     *             Building a context without parameters is no longer supported:
     *             all constructors will require at least the network being exported.
     */
    @Deprecated(since = "7.4.0", forRemoval = true)
```

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CommonGridModelExportTest.java
```diff
@@ -708,6 +709,120 @@ private void addModelsForSubnetworks(Network network, int version) {
...
+        vl2b.getBusBreakerView().newBus().setId("Bus_NL2").add();
+        network2.newLine().setId("Line_NL")
+                .setVoltageLevel1("VL_NL").setBus1("Bus_NL")
+                .setVoltageLevel2("VL_NL2").setBus2("Bus_NL2")
+                .setR(1.0).setX(10.0).setG1(0.0).setB1(0.0).setG2(0.0).setB2(0.0).add()
+                .newOperationalLimitsGroup1("OLG_NL_1")
+                .newCurrentLimits()
+                .setPermanentLimit(1000.0)
+                .beginTemporaryLimit()
+                .setName("OLG_BE_T_1")
```
zamarrenolm: to avoid confusion in the example, use "OLG_NL_T_1" instead of "OLG_BE_T_1".

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CommonGridModelExportTest.java
```diff
@@ -708,6 +709,120 @@ private void addModelsForSubnetworks(Network network, int version) {
...
+                .setAcceptableDuration(10)
+                .setValue(1100.0)
+                .endTemporaryLimit()
+                .beginTemporaryLimit()
+                .setName("OLG_BE_T_2")
```
zamarrenolm: to avoid confusion in the example, use "OLG_NL_T_2" instead of "OLG_BE_T_2".

colineplqt: Sorry, I forgot to modify my copy/paste

zamarrenolm (dismissed):

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CommonGridModelExportTest.java
```diff
@@ -708,6 +709,120 @@ private void addModelsForSubnetworks(Network network, int version) {
...
+        Set<String> nlLoadGroupIds = getUniqueMatches(nlEqXml, regexNonConformLoadGroup);
+        assertEquals(1, beLoadGroupIds.size());
+        assertEquals(1, nlLoadGroupIds.size());
+
+        // Each EQ file must contain one OperationalLimitType
+        Pattern regexOperationalLimitType = Pattern.compile("cim:OperationalLimitType rdf:ID=\"(_[^\"]*)\"");
+        Set<String> beOpLimitTypeIds = getUniqueMatches(beEqXml, regexOperationalLimitType);
+        Set<String> nlOpLimitTypeIds = getUniqueMatches(nlEqXml, regexOperationalLimitType);
+        assertFalse(beOpLimitTypeIds.isEmpty());
+        assertFalse(nlOpLimitTypeIds.isEmpty());
```
NathanDissoubray, line 742: why not test that there is exactly 1 `OperationalLimitType` (instead of "at least one") ?

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CommonGridModelExportTest.java
```diff
@@ -708,6 +709,120 @@ private void addModelsForSubnetworks(Network network, int version) {
...
+
+        // Each EQ file must contain exactly one NonConformLoadGroup
+        Pattern regexNonConformLoadGroup = Pattern.compile("cim:NonConformLoadGroup rdf:ID=\"(_[^\"]*)\"");
+        Set<String> beLoadGroupIds = getUniqueMatches(beEqXml, regexNonConformLoadGroup);
+        Set<String> nlLoadGroupIds = getUniqueMatches(nlEqXml, regexNonConformLoadGroup);
+        assertEquals(1, beLoadGroupIds.size());
+        assertEquals(1, nlLoadGroupIds.size());
+
+        // Each EQ file must contain one OperationalLimitType
+        Pattern regexOperationalLimitType = Pattern.compile("cim:OperationalLimitType rdf:ID=\"(_[^\"]*)\"");
```
NathanDissoubray: I think it would be clearer to have `\"(_[^\"]*)\"` as a single string injected in the two regex.
With a comment for people that might not be too familiar with regex, and an example:
ID has following shape: "_blablabla-32-something"
so we match `"_` literally, then `^\"` matches everything that is not `"` as many times as needed with `*`

NathanDissoubray (dismissed):

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CommonGridModelExportTest.java
colineplqt: Actually in that case there are 3 operational limit type -> 1 PATL and two TATL. I am going to modify the test to make it clearer!

colineplqt: If you can tell me if you find it clearer :)

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CommonGridModelExportTest.java
colineplqt: I added a comment of how to understand the regex, and I did two, one for PATL and one for PATL, linked to the first comment. I think it's clearer like that.

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CommonGridModelExportTest.java
NathanDissoubray: i'm not sure to understand why there is one type for the PATL and two for the TATL, but at least I understand the test :+1:

colineplqt: Each tatl type is for one duration, but what is important is that the IDs are different from one IGM to another.

NathanDissoubray: ah I thought it was always 2 for TATL, but it's just because there are 2 TATL in this particular case

colineplqt: Yes exactly

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CommonGridModelExportTest.java
```diff
@@ -708,6 +709,138 @@ private void addModelsForSubnetworks(Network network, int version) {
...
+        exportParams.put(CgmesExport.CGM_EXPORT, false);
+        exportParams.put(CgmesExport.PROFILES, List.of("EQ"));
+
+        String basenameBeEq = "test_loads_limits_be";
+        String basenameNlEq = "test_loads_limits_nl";
+        network.getSubnetwork("Network_BE").write("CGMES", exportParams, tmpDir.resolve(basenameBeEq));
+        network.getSubnetwork("Network_NL").write("CGMES", exportParams, tmpDir.resolve(basenameNlEq));
+
+        String beEqXml = Files.readString(tmpDir.resolve(basenameBeEq + "_EQ.xml"));
+        String nlEqXml = Files.readString(tmpDir.resolve(basenameNlEq + "_EQ.xml"));
```
rcourtier: You can use `writeCgmesProfile` from `ConversionUtil`:

```java
String beEqXml = writeCgmesProfile(network.getSubnetwork("Network_BE"), "EQ", tmpDir, exportParams);
String nlEqXml = writeCgmesProfile(network.getSubnetwork("Network_NL"), "EQ", tmpDir, exportParams);
```

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CommonGridModelExportTest.java
```diff
@@ -708,6 +709,138 @@ private void addModelsForSubnetworks(Network network, int version) {
...
+
+        // Each EQ file must contain exactly one NonConformLoadGroup
+        Pattern regexNonConformLoadGroup = Pattern.compile("cim:NonConformLoadGroup rdf:ID=\"(_[^\"]*)\"");
+        Set<String> beLoadGroupIds = getUniqueMatches(beEqXml, regexNonConformLoadGroup);
+        Set<String> nlLoadGroupIds = getUniqueMatches(nlEqXml, regexNonConformLoadGroup);
+        assertEquals(1, beLoadGroupIds.size());
+        assertEquals(1, nlLoadGroupIds.size());
```
rcourtier: In order to not use regex, I suggest to create a new method in `ConversionUtil` to retrieve all elements from a certain class, e.g.:
```java
    public static Set<String> getElements(String xmlFile, String className) {
        String regex = "(<cim:" + className + " rdf:.*?</cim:" + className + ">)";
        Pattern pattern = Pattern.compile(regex, Pattern.DOTALL);
        return getUniqueMatches(xmlFile, pattern);
    }
```

That way, you can refactor this code into:
```java
        Set<String> beLoadGroups = getElements(beEqXml, "NonConformLoadGroup");
        Set<String> nlLoadGroups = getElements(nlEqXml, "NonConformLoadGroup");
        assertEquals(1, beLoadGroups.size());
        assertEquals(1, nlLoadGroups.size());
```

cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/export/CommonGridModelExportTest.java
```diff
@@ -708,6 +709,138 @@ private void addModelsForSubnetworks(Network network, int version) {
...
+        Set<String> nlTatlOpLimitTypeIds = getUniqueMatches(nlEqXml, regexTatlOperationalLimitType);
+        assertEquals(1, bePatlOpLimitTypeIds.size());
+        assertEquals(2, beTatlOpLimitTypeIds.size());
+        assertEquals(1, nlPatlOpLimitTypeIds.size());
+        assertEquals(2, nlTatlOpLimitTypeIds.size());
+
+        // The LoadGroup IDs and OperationalLimitType IDs must be different between the two IGMs
+        assertTrue(Collections.disjoint(beLoadGroupIds, nlLoadGroupIds));
+        assertTrue(Collections.disjoint(bePatlOpLimitTypeIds, nlPatlOpLimitTypeIds));
+        assertTrue(Collections.disjoint(beTatlOpLimitTypeIds, nlTatlOpLimitTypeIds));
```
rcourtier: In order to not use regex, I suggest to create a new method in ConversionUtil to retrieve the identifier of an element, e.g.:
```java
    public static String getIdentifier(String element) {
        String regex = "<cim:.*? (?:rdf:ID=\"_|rdf:about=\"#_)(.*?)\">";
        Pattern pattern = Pattern.compile(regex);
        return getFirstMatch(element, pattern);
    }
```

That way, you can refactor this code into:
```java
        // Each EQ file must contain one OperationalLimitType for the PATL and two OperationalLimitType for the TATL
        Set<String> beOpLimitTypes = getElements(beEqXml, "OperationalLimitType");
        Set<String> bePatlOpLimitTypes = beOpLimitTypes.stream()
            .filter(e -> getIdentifier(e).startsWith("PATL_"))
            .collect(Collectors.toSet());
        Set<String> beTatlOpLimitTypes = beOpLimitTypes.stream()
            .filter(e -> getIdentifier(e).startsWith("TATL_"))
            .collect(Collectors.toSet());
        Set<String> nlOpLimitTypes = getElements(nlEqXml, "OperationalLimitType");
        Set<String> nlPatlOpLimitTypes = nlOpLimitTypes.stream()
            .filter(e -> getIdentifier(e).startsWith("PATL_"))
            .collect(Collectors.toSet());
        Set<String> nlTatlOpLimitTypes = nlOpLimitTypes.stream()
            .filter(e -> getIdentifier(e).startsWith("TATL_"))
            .collect(Collectors.toSet());

        assertEquals(1, bePatlOpLimitTypes.size());
        assertEquals(2, beTatlOpLimitTypes.size());
        assertEquals(1, nlPatlOpLimitTypes.size());
        assertEquals(2, nlTatlOpLimitTypes.size());

        // The LoadGroup IDs and OperationalLimitType IDs must be different between the two IGMs
        assertTrue(Collections.disjoint(beLoadGroups, nlLoadGroups));
        assertTrue(Collections.disjoint(bePatlOpLimitTypes, nlPatlOpLimitTypes));
        assertTrue(Collections.disjoint(beTatlOpLimitTypes, nlTatlOpLimitTypes));
```

rcourtier (changes requested):

rcourtier (approved):

zamarrenolm (approved):

PR #4036, opened 2026-08-08
prevent zip path traversal in LocalComputationManager unzip

naruto-lgtm:
Repro: register a Command input file with FilePreProcessor.ARCHIVE_UNZIP whose archive holds an entry named ../evil.txt; preProcess writes it into the parent of the execution working directory.
Cause: the ARCHIVE_UNZIP branch resolves ze.getName() against workingDir with no containment check, so ../ segments and absolute entry names land outside the sandboxed working directory.
Fix: normalize each resolved target and reject any entry that does not stay under the working directory before copying.


olperr1:
Thanks for your contribution.
This is a problem we are aware of, and we are already working on it (the patch is under review in a specific private repository). We plan to publish a corrective release on this subject very soon.

PR #4037, opened 2026-08-10
fix locale-dependent base voltage mRID in CGMES export

naruto-lgtm:
The BaseVoltage mRID in computeBaseVoltageMapping is derived from a DecimalFormat created without an explicit locale, so on a JVM whose default locale uses a comma decimal separator the same network yields a different id (220,5_BV instead of 220.5_BV) and puts a comma inside the identifier. Pinning the formatter to Locale.US matches CgmesExportUtil, which already fixes the decimal symbols for the numeric values it writes. The added regression test exports under Locale.FRANCE and checks the id stays 220.5_BV.


zamarrenolm:
Instead of referencing an explicit local (Locale.US) you could explicitly set the decimal separator, to clarify the intention. No need to set also the the grouping separator because we won't use it in our format.

```java
DecimalFormatSymbols customSymbols = new DecimalFormatSymbols();
customSymbols.setDecimalSeparator('.');
DecimalFormat noTrailingZerosFormat = new DecimalFormat("0.##", customSymbols);
```

naruto-lgtm:
Makes sense, the intent is clearer that way. Switched to a DecimalFormatSymbols with only the decimal separator set to '.', left the grouping separator alone since the pattern doesn't use it. Test still passes under Locale.FRANCE.

zamarrenolm (approved):

colineplqt (approved):

PR #4040, opened 2026-08-11
quote program name and validate env var name in unix executor

naruto-lgtm:

No

Security fix

On Unix systems, the local command executor doesn't escape the program name, nor the envvar names.

On Unix systems, the local command executor **does** escape the program name and the envvar names.

- [ ] Yes
- [X] No

**Other information**:

Repro: on a unix host, hand `LocalCommandExecutor` an environment variable whose name is `FOO=1; touch /tmp/pwned; BAR`, or a program name like `echo $(touch /tmp/pwned)` — the injected command runs.

Cause: the GHSA-jqvf-j3ww-r8c7 fix single-quoted the env values and the args in `UnixLocalCommandExecutor`, but the variable name and the program are still concatenated into the `bash -c` string raw. `WindowsLocalCommandExecutor` already guards both (name regex, program quote/space checks), so this is the unix side of that same fix.

Fix: reject names outside `[a-zA-Z_]\w*` with the same message the windows executor uses — a name sits before `=` and after `$`, so it can't be single-quoted like a value — and single-quote the program the way the values and args already are. Added the two matching unix cases to `LocalCommandExecutorSecurityTest`.

### DIFF
--- a/computation-local/src/main/java/com/powsybl/computation/local/UnixLocalCommandExecutor.java
+++ b/computation-local/src/main/java/com/powsybl/computation/local/UnixLocalCommandExecutor.java
@@ -42,13 +42,19 @@ public class UnixLocalCommandExecutor extends AbstractLocalCommandExecutor {
         for (Map.Entry<String, String> entry : env2.entrySet()) {
             String name = entry.getKey();
             String value = entry.getValue();
+            // A variable name sits before the '=' assignment and after '$', so it cannot be
+            // single-quoted like the value. Reject anything but alphanumerics and underscores
+            // to keep a hostile name from breaking out of the command.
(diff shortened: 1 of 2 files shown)


olperr1 (approved):

olperr1:
Thank you for this valuable contribution!

Can you check [our security policy](https://github.com/powsybl/powsybl-core/security/policy)? For any important or critical security vulnerability, be sure to reach us via e-mail.
For this current PR, we evaluate the risk as moderate, so a public PR is acceptable. But if you have any doubt, it would be safer to contact us privately.

(On a side note, for your future contributions, it would be great if you could use the project's PR template.)

PR #4043, opened 2026-08-12
Memory improvement

### LINKED ISSUE #4038: Improve memory use for big networks

When creating a big network, for instance with 80 000 switches, `Identifiable` objects are created and with them a lot of empty maps and sets that are not necessary filled afterwards (aliases, extensions...). This causes big memory use for no real reason and can be improved.

Empty maps and set of `Identifiable` objects are set to `null` if possible to avoid consuming memory for no use.

### Describe the motivation

Enlighten the memory use of network `Identiiable` objecs to improve memory us for very big networks. and avoid Java Heap Space exceptions.

### Extra Information

Exemple of a memory analyze for a network with around 80 000 switches for which a set of aliases, some empty map/set are created for the extensions for instance and empty Properties object are created and use a lot of memory provoiking Java Hap Space excepions:

Image

alicecaron:

Fixes #4038 

Change initialization of Network Identifiable objects to avoid creating unnecessary objects that increase memory use.

- [ ] Yes
- [x] No

This is not a breaking change, but in order to limit memory usage when using the in-memory IIDM implementation, it is recommended:
- to call `hasProperties()` before calling `getProperties()` on:
  - the `PropertiesHolder` objects (and then the `Identifiable` objects)
  - the `PropertiesContainer` objects.
- to call `hasExtensions()` before calling `getExtensions()` on:
  - the `Identifiable` objects.

**Other information**:

### DIFF
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/ReportTapChangers.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/ReportTapChangers.java
@@ -27,7 +27,7 @@ public class ReportTapChangers {
     public ReportTapChangers(CgmesModel cgmes, Consumer<String> output) {
         this.cgmes = cgmes;
         this.output = output;
-        Path p = cgmes.hasProperties() ? (Path) cgmes.getProperties().get("dataSource") : null;
+        Path p = cgmes.getProperties() != null ? (Path) cgmes.getProperties().get("dataSource") : null;
         String ps = "";
         if (p != null) {
             ps = p.toString().replaceAll("../data", "");
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/ReportTapChangers.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/ReportTapChangers.java
@@ -27,7 +27,7 @@ public class ReportTapChangers {
     public ReportTapChangers(CgmesModel cgmes, Consumer<String> output) {
         this.cgmes = cgmes;
         this.output = output;
-        Path p = cgmes.getProperties() != null ? (Path) cgmes.getProperties().get("dataSource") : null;
+        Path p = cgmes.hasProperties() ? (Path) cgmes.getProperties().get("dataSource") : null;
         String ps = "";
         if (p != null) {
             ps = p.toString().replaceAll("../data", "");
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/ReportTapChangers.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/ReportTapChangers.java
@@ -27,7 +27,7 @@ public class ReportTapChangers {
     public ReportTapChangers(CgmesModel cgmes, Consumer<String> output) {
         this.cgmes = cgmes;
         this.output = output;
-        Path p = (Path) cgmes.getProperties().get("dataSource");
+        Path p = cgmes.getProperties() != null ? (Path) cgmes.getProperties().get("dataSource") : null;
         String ps = "";
         if (p != null) {
             ps = p.toString().replaceAll("../data", "");
--- a/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/ReportTapChangers.java
+++ b/cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/ReportTapChangers.java
@@ -27,7 +27,7 @@ public class ReportTapChangers {
     public ReportTapChangers(CgmesModel cgmes, Consumer<String> output) {
         this.cgmes = cgmes;
         this.output = output;
-        Path p = cgmes.getProperties() != null ? (Path) cgmes.getProperties().get("dataSource") : null;
+        Path p = (Path) cgmes.getProperties().get("dataSource");
         String ps = "";
         if (p != null) {
             ps = p.toString().replaceAll("../data", "");
--- a/cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesModel.java
+++ b/cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesModel.java
@@ -30,6 +30,8 @@ public interface CgmesModel {
 
     Properties getProperties();
 
+    boolean hasProperties();
+
     default PropertyBags fullModels() {
         return new PropertyBags();
     }
--- a/cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesModel.java
+++ b/cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesModel.java
@@ -30,8 +30,6 @@ public interface CgmesModel {
 
     Properties getProperties();
 
-    boolean hasProperties();
-
     default PropertyBags fullModels() {
         return new PropertyBags();
     }
--- a/commons/src/main/java/com/powsybl/commons/extensions/AbstractExtendable.java
+++ b/commons/src/main/java/com/powsybl/commons/extensions/AbstractExtendable.java
@@ -72,11 +72,14 @@ public abstract class AbstractExtendable<T> implements Extendable<T> {
         return removed;
     }
 
+    /**
+     * {@inheritDoc}
+     * <p>To limit memory usage, it is recommended to use {@link #hasExtensions()} before calling this method.
+     * (or to call {@link #getExtensionsStream()} instead).</p>
+     * @return the properties
+     */
     @Override
     public Collection<Extension<T>> getExtensions() {
-        //if (extensionsByName == null) {
-        //    return Collections.emptyList();
-        //}
         initiateExtensions();
         return extensionsByName.values();
     }
@@ -99,7 +102,7 @@ public abstract class AbstractExtendable<T> implements Extendable<T> {
 
     @Override
     public boolean hasExtensions() {
-        return !(extensions == null || extensions.isEmpty());
+        return extensions != null && !extensions.isEmpty();
     }
 
     @Override
--- a/commons/src/main/java/com/powsybl/commons/extensions/AbstractExtendable.java
+++ b/commons/src/main/java/com/powsybl/commons/extensions/AbstractExtendable.java
@@ -17,7 +17,7 @@ import java.util.Objects;
  */
 public abstract class AbstractExtendable<T> implements Extendable<T> {
 
-    private final Map<Class<?>, Extension<T>> extensions = new HashMap<>();
+    private Map<Class<?>, Extension<T>> extensions = null;
 
     private final Map<String, Extension<T>> extensionsByName = new HashMap<>();
 
@@ -28,6 +28,9 @@ public abstract class AbstractExtendable<T> implements Extendable<T> {
         // remove any existing, which will trigger extension own cleanup if any
         this.removeExtension((Class<E>) type);
         extension.setExtendable((T) this);
+        if (extensions == null) {
+            extensions = new HashMap<>();
+        }
         extensions.put(type, extension);
         extensionsByName.put(extension.getName(), extension);
     }
@@ -35,6 +38,9 @@ public abstract class AbstractExtendable<T> implements Extendable<T> {
     @Override
     public <E extends Extension<T>> E getExtension(Class<? super E> type) {
         Objects.requireNonNull(type);
+        if (extensions == null) {
+            return null;
+        }
         return (E) extensions.get(type);
     }
 
@@ -45,7 +51,9 @@ public abstract class AbstractExtendable<T> implements Extendable<T> {
     }
 
     protected <E extends Extension<T>> void removeExtension(Class<E> type, E extension) {
-        extensions.remove(type);
+        if (extensions != null) {
+            extensions.remove(type);
+        }
         extensionsByName.remove(extension.getName());
         extension.setExtendable(null);
     }
--- a/commons/src/main/java/com/powsybl/commons/extensions/AbstractExtendable.java
+++ b/commons/src/main/java/com/powsybl/commons/extensions/AbstractExtendable.java
@@ -7,10 +7,7 @@
  */
 package com.powsybl.commons.extensions;
 
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Objects;
+import java.util.*;
 
 /**
  * @author Mathieu Bague {@literal <mathieu.bague at rte-france.com>}
@@ -19,7 +16,7 @@ public abstract class AbstractExtendable<T> implements Extendable<T> {
 
     private Map<Class<?>, Extension<T>> extensions = null;
 
-    private final Map<String, Extension<T>> extensionsByName = new HashMap<>();
+    private Map<String, Extension<T>> extensionsByName = null;
 
     @Override
     public <E extends Extension<T>> void addExtension(Class<? super E> type, E extension) {
@@ -32,6 +29,9 @@ public abstract class AbstractExtendable<T> implements Extendable<T> {
             extensions = new HashMap<>();
         }
         extensions.put(type, extension);
+        if (extensionsByName == null) {
+            extensionsByName = new HashMap<>();
+        }
         extensionsByName.put(extension.getName(), extension);
     }
 
@@ -47,6 +47,9 @@ public abstract class AbstractExtendable<T> implements Extendable<T> {
     @Override
     public <E extends Extension<T>> E getExtensionByName(String name) {
         Objects.requireNonNull(name);
+        if (extensionsByName == null) {
+            return null;
+        }
         return (E) extensionsByName.get(name);
     }
 
@@ -54,7 +57,9 @@ public abstract class AbstractExtendable<T> implements Extendable<T> {
         if (extensions != null) {
             extensions.remove(type);
         }
-        extensionsByName.remove(extension.getName());
+        if (extensionsByName != null) {
+            extensionsByName.remove(extension.getName());
+        }
         extension.setExtendable(null);
     }
 
@@ -73,6 +78,9 @@ public abstract class AbstractExtendable<T> implements Extendable<T> {
 
     @Override
     public Collection<Extension<T>> getExtensions() {
+        if (extensions == null) {
+            return Collections.emptyList();
+        }
         return extensionsByName.values();
     }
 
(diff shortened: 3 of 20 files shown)


iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/PropertiesContainer.java
```diff
@@ -16,39 +16,58 @@
  * @author Olivier Perrin {@literal <olivier.perrin at rte-france.com>}
  */
 public class PropertiesContainer {
-    protected final Properties properties = new Properties();
+    private Properties properties = null;
 
     public Properties getProperties() {
+        if (properties == null) {
+            properties = new Properties();
```
phiedw, line 29: Do we want to allow the user to set properties by doing getProperties().put(k,v) or only by doing setProperty(k,v)?
I'm not sure if the first way is even doable but it would allow us to avoid creating the empty Properties when a user just wants to check what Properties an object has (for exports for instance).

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/PropertiesContainer.java
```diff
@@ -16,39 +16,58 @@
...
     public String removeProperty(String key) {
+        if (properties == null) {
+            return null;
+        }
         return (String) properties.remove(key);
     }
 
     public Set<String> getPropertyNames() {
+        if (properties == null) {
+            properties = new Properties();
```
phiedw: Why do we need to create an empty Properties here? Can we not simply return Collections.emptySet()

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/PropertiesContainer.java
alicecaron: This would imply that for instance the `IdentifiableExtensionGroovyTest::testBooleanProperty` should be adapted and that it introduces a breaking change in case some user set property directly with the Properties direct access `properties.my_property = "value"` because till now it was possible

geofjamg:
Could you add a memory profiling to figure out how much to we decrease memory consomption on a real case ? (a % of initial size)

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/PropertiesContainer.java
geofjamg: Be aware that it would be a breaking change very difficult to catch because only seen at runtime

alicecaron:
A test has been done with a network having more than 80 000 switches:

Before: 
before
After: 
after


The memory use goes from 280M to 226M (54M to 31M for the switches) so a improvement of around 20% for the total  network.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/PropertiesContainer.java
alicecaron: For those two comments: the goal of this MR is to improve the import/network memory use. So maybe we can keep the code like it was a the beginning (I can revert my last commit that changes to return an empy collection). It avoids breaking change and improve memory use anyway. Then later if needed we can think about returning the empty collection. What do you think @geofjamg  @phiedw ?

commons/src/main/java/com/powsybl/commons/extensions/AbstractExtendable.java
```diff
@@ -65,6 +78,9 @@ public <E extends Extension<T>> boolean removeExtension(Class<E> type) {
 
     @Override
     public Collection<Extension<T>> getExtensions() {
+        if (extensions == null) {
+            return Collections.emptyList();
```
olperr1: `Map.values()` is backed by the map (see [the Javadoc](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Map.html#values())).
I don't think someone relies on this specificity since we haven't described it in our Javadoc and it is clearly dependant on the implementation, but it may also be considered as a breaking change.

Something that can be done is to:
- create the map here when it is null
- introduce a new `hasExtension()` method
- call this new method in our code before calling `getExtensions()`

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/PropertiesContainer.java
```diff
@@ -16,39 +17,58 @@
...
     public String removeProperty(String key) {
+        if (properties == null) {
+            return null;
+        }
         return (String) properties.remove(key);
     }
 
     public Set<String> getPropertyNames() {
+        if (properties == null) {
+            return Collections.emptySet();
```
olperr1: To avoid a breaking change here, you can use `return new HashSet<>();`
This way, the returned set will be mutable, just like the result of `... .collect(Collectors.toSet());` (see [the javadoc](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/stream/Stream.html#collect(java.util.stream.Collector))).

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractIdentifiable.java
```diff
@@ -146,17 +156,23 @@ public void addAlias(String alias, String aliasType, boolean ensureAliasUnicity)
...
         if (Strings.isNullOrEmpty(type)) {
-            aliasesWithoutType.remove(alias);
+            if (aliasesWithoutType != null) {
+                aliasesWithoutType.remove(alias);
+            }
         } else {
-            aliasesByType.remove(type);
+            if (aliasesByType != null) {
+                aliasesByType.remove(type);
+            }
```
olperr1:
```suggestion
            // aliasesByType cannot be null here
            aliasesByType.remove(type);
```

olperr1 (changes requested):

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/util/RegulatedTerminalControllers.java
```diff
@@ -99,8 +99,10 @@ private static void add(List<TerminalRef> regulatedTerminals, Terminal regulated
 
     private List<TerminalRef> findRegulatedTerminalsInExtensions(Identifiable<?> identifiable) {
         List<TerminalRef> regulatedTerminals = new ArrayList<>();
-        identifiable.getExtensions().stream().map(Extension::getName).forEach(extensionName ->
-                add(regulatedTerminals, findRegulatedTerminalInExtension(identifiable, extensionName)));
+        if (identifiable.hasExtensions()) {
+            identifiable.getExtensions().stream().map(Extension::getName).forEach(extensionName ->
+                    add(regulatedTerminals, findRegulatedTerminalInExtension(identifiable, extensionName)));
+        }
```
olperr1:
```suggestion
        identifiable.getExtensionsStream().map(Extension::getName).forEach(extensionName ->
                add(regulatedTerminals, findRegulatedTerminalInExtension(identifiable, extensionName)));
```

commons/src/main/java/com/powsybl/commons/extensions/AbstractExtendable.java
```diff
@@ -65,9 +74,34 @@ public <E extends Extension<T>> boolean removeExtension(Class<E> type) {
...
+            extensions = new HashMap<>();
+        }
+        if (extensionsByName == null) {
+            extensionsByName = new HashMap<>();
+        }
+    }
+
+    @Override
+    public boolean hasExtensions() {
+        return !(extensions == null || extensions.isEmpty());
```
olperr1: This would be more readable like this:
```suggestion
        return extensions != null && !extensions.isEmpty();
```

commons/src/main/java/com/powsybl/commons/extensions/AbstractExtendable.java
```diff
@@ -65,9 +74,34 @@ public <E extends Extension<T>> boolean removeExtension(Class<E> type) {
 
     @Override
     public Collection<Extension<T>> getExtensions() {
```
olperr1, line 82:
```suggestion
    /**
     * {@inheritDoc}
     * <p>To limit memory usage, it is recommended to use {@link #hasExtensions()} before (or to call {@link #getExtensionsStream()} instead).</p>
     */
    @Override
    public Collection<Extension<T>> getExtensions() {
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/PropertiesContainer.java
```diff
@@ -16,39 +17,58 @@
  * @author Olivier Perrin {@literal <olivier.perrin at rte-france.com>}
  */
 public class PropertiesContainer {
-    protected final Properties properties = new Properties();
+    private Properties properties = null;
 
     public Properties getProperties() {
+        if (properties == null) {
+            return null;
```
olperr1: I think we must not return `null` here, but initialize the object instead.

It would be the same as for the extensions. When there are no properties:
1. calling directly `getProperties()` will initialize the object and return it
2. we can call `hasProperty()` to check if there are any properties before calling `getProperties()`. In this case, the object will not be initialized if not needed.

With this policy, the existing code will still work without adaptations. It is recommanded to use the 2nd to save memory, but it is not required.

If we return null, existing code MUST be adapted to prevent NullPointerException. This is a (dangerous) breaking change.

Also could you add the following comment on the method?
```java
    /**
     * <p>Returns the properties.</p>
     * <p>To limit memory usage, it is recommended to use {@link #hasProperty()} before calling this method.</p>
     * @return the properties
     */
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractNetwork.java
```diff
@@ -146,15 +148,17 @@ protected static void transferProperties(AbstractNetwork fromNetwork, AbstractNe
...
-        fromNetwork.getProperties().forEach((key, value) -> {
-            if (ignoreAlreadyPresent && toNetwork.hasProperty(key.toString())) {
-                LOGGER.warn("Property \"{}\" was not transferred from \"{}\" to \"{}\": it already exists in the destination network.",
-                        key, fromNetwork.getId(), toNetwork.getId());
-            } else {
-                toNetwork.setProperty(key.toString(), value.toString());
-                fromNetwork.removeProperty(key.toString());
-            }
-        });
+        if (fromNetwork.getProperties() != null) {
```
olperr1:
```suggestion
        if (fromNetwork.hasProperty()) {
```

commons/src/main/java/com/powsybl/commons/extensions/AbstractExtendable.java
```diff
@@ -65,9 +74,34 @@ public <E extends Extension<T>> boolean removeExtension(Class<E> type) {
...
+        //if (extensionsByName == null) {
+        //    return Collections.emptyList();
+        //}
```
olperr1:
```suggestion
```

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/PropertiesContainer.java
olperr1: Also, can you add this comment on `AbstractIdentifiable#getProperties()` and `AbstractPropertiesHolder#getProperties()`?

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/ReportTapChangers.java
```diff
@@ -27,7 +27,7 @@ public class ReportTapChangers {
     public ReportTapChangers(CgmesModel cgmes, Consumer<String> output) {
         this.cgmes = cgmes;
         this.output = output;
-        Path p = (Path) cgmes.getProperties().get("dataSource");
+        Path p = cgmes.getProperties() != null ? (Path) cgmes.getProperties().get("dataSource") : null;
```
olperr1:
```suggestion
        Path p = cgmes.hasProperty() ? (Path) cgmes.getProperties().get("dataSource") : null;
```

iidm/iidm-scripting/src/test/groovy/com/powsybl/iidm/network/scripting/IdentifiableExtensionGroovyTest.groovy
olperr1, line 1: If `PropertiesContainer#getProperties()` initializes the object, this test file changes must be reverted (no breaking change).

olperr1 (changes requested):
For the migration guide:
This is not a breaking change, but it is recommanded to use `hasProperty()` and `hasExtensions()` before calling respectively `getProperties()` and `getExtensions()`, in order to limit memory usage.

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesModel.java
```diff
@@ -30,6 +30,8 @@ public interface CgmesModel {
 
     Properties getProperties();
 
+    boolean hasProperties();
```
olperr1: I think we should not alter `CgmesModel` in this PR. Adding this method doesn't improve the memory usage.

olperr1 (changes requested):

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesModel.java
olperr1: I have just found that you have added this method because of one of my comments... which was irrelevant.
Sorry!

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/ReportTapChangers.java
olperr1: It's irrelevant: the CgmesModel is not concerned by the memory optimization (cf. other comments).

iidm/iidm-scripting/src/test/groovy/com/powsybl/iidm/network/scripting/IdentifiableExtensionGroovyTest.groovy
olperr1, line 1: These changes are not relevant in the context of this PR (only white spaces added).

olperr1 (changes requested):

cgmes/cgmes-model/src/main/java/com/powsybl/cgmes/model/CgmesModel.java
olperr1: My comment is maybe unclear, so just to sum-up: changes on `ReportTapChangers`, `AbstractCgmesModel`, `CgmesModel` and `InMemoryCgmesModel` must be reverted.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/ReportTapChangers.java
olperr1, line 1: I think this file can be reverted too (or am I missing something?)

olperr1 (changes requested):

commons/src/main/java/com/powsybl/commons/extensions/Extendable.java
```diff
@@ -56,6 +57,18 @@ public interface Extendable<O> {
...
+     * Get all extensions associated with this extendable object as a stream.
+     * @return
+     */
+    Stream<Extension<O>> getExtensionsStream();
+
+    /**
+     * Returns whether this extendable has at least an extension.
+     * @return
+     */
+    boolean hasExtensions();
```
olperr1, line 70: To avoid a breaking change, these methods could have a default implementation here:
```java
    default Stream<Extension<O>> getExtensionsStream() {
        Collection<Extension<O>> extensions = getExtensions();
        return extensions != null ? extensions.stream() : Stream.empty();
    }

    /**
     * Returns whether this extendable has at least an extension.
     * @return
     */
    default boolean hasExtensions() {
        Collection<Extension<O>> extensions = getExtensions();
        return extensions != null && !extensions.isEmpty();
    }
```

olperr1 (changes requested):

commons/src/main/java/com/powsybl/commons/extensions/Extendable.java
olperr1: In your last commit ([549f58a](https://github.com/powsybl/powsybl-core/pull/4043/commits/549f58afc51ac5c87b9bca993a9121432aba9645)), you removed the overrides for these 2 methods in `AbstractExtendable`.
They must be kept or there won't be any performance gain since in the default implementations we call `getExtensions()` which initialized the maps if they are null.
Can you restore them?

olperr1 (changes requested):

commons/src/main/java/com/powsybl/commons/extensions/Extendable.java
alicecaron: Yes you're right! Done

olperr1: Thanks!

olperr1 (approved):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/AbstractNetwork.java
```diff
@@ -113,21 +113,23 @@ protected static void transferExtensions(Network from, Network to) {
...
-                                    clazz.getName(), from.getId(), to.getId());
-                        } else {
-                            ((AbstractIdentifiable<Network>) from).removeExtension((Class<? extends Extension<Network>>) clazz, false);
-                            to.addExtension((Class<? super Extension<Network>>) clazz, (Extension<Network>) e);
-                        }
-                    })
-        );
+        if (from.hasExtensions()) {
+            // Only well-defined extensions (with an interface) are transferred
+            new ArrayList<Extension<?>>(from.getExtensions()).forEach(e ->
```
phiedw, line 118: I'm not sure I understand the logic here, can we not just do from.getExtensionsStream().forEach(...)?

olperr1: This is exactly the same code as before, but it was moved in the `if (from.hasExtensions())` block. The only changes here are the indents.
Before this PR, the `getExtensionsStream()` method didn't exist. This explains why it wasn't used.
But you are right! Now, we can just use this new method to have a cleaner code.

olperr1: In fact, some extensions are removed in this block (l.127). Therefore, it is not possible to use directly the stream and a copy is necessary, or a `ConcurrentModification` exception is encountered.

olperr1 (approved):

phiedw (approved):

PR #4045, opened 2026-08-12
Add documentation explaining why negative R values are allowed on AC lines and AC transformers

NathanDissoubray:

No, in link with #3527 to avoid further confusion on AC lines, transformer and DC line (detailled model)

Docs and javadoc update

It is not clear why negative values are accepted for lines and two-windings transformer, but not for DCLine

Add some doc to explain why it is allowed for lines and transformers but not for DCLine

- [ ] Yes
- [x] No

**Other information**:

Informations taken from #3576 
@jeandemanged could you confirm I wrote the correct information regarding equivalent models ?


jeandemanged (commented):
Thanks for the doc.

OK for AC components. For DC I would like community feedback.

The PR is set to close issue #3527 which requested to forbid negative R in the case of HvdcLine (simplified DC model). I agree with @rcourtier that HvdcLine (simplified DC model) negative resistance is indeed something that should be blocked.

For DcLine (detailed model) negative resistances are blocked, at the time this was made by mimicking CGMES [QoCDC](https://eepublicdownloads.azureedge.net/clean-documents/digital/Quality%20Of%20CGMES%20Datasets%20and%20Calculations%20v4.1.4_clean.pdf) rule `DCLineSegmentR`. On my side I have no use case for DcLine negative resistance, but maybe in a DC detailed model we could also have equivalents with negative resistance, although CGMES does not support it? Any opinion on this @SGI-Landry ? Should we allow negative DC Line R in detailed model, or is it OK as it is now?

NathanDissoubray:
> The PR is set to close issue https://github.com/powsybl/powsybl-core/issues/3527 which requested to forbid negative R in the case of HvdcLine (simplified DC model). I agree with @rcourtier that HvdcLine (simplified DC model) negative resistance is indeed something that should be blocked.

According to the PR #3576, [some comments](https://github.com/powsybl/powsybl-core/pull/3576#pullrequestreview-3302513335) from @geofjamg and @JB-H mention that negative R values should be accepted for HVDC lines, since it is possible for a network reduction to yield negative R values (which is currently accepted by our model).

The original issue has thus been decided as "invalid", as negative values should not be blocked for HVDC lines. This PR still "fixes" the issue, because rather than doing what the issues asks, I add documentation to explain why it should not be done.

Do you agree, or do you think negative R values should be blocked for HVDC lines ?

SGI-Landry:
> Thanks for the doc.
> 
> OK for AC components. For DC I would like community feedback.
> 
> The PR is set to close issue #3527 which requested to forbid negative R in the case of HvdcLine (simplified DC model). I agree with @rcourtier that HvdcLine (simplified DC model) negative resistance is indeed something that should be blocked.
> 
> For DcLine (detailed model) negative resistances are blocked, at the time this was made by mimicking CGMES [QoCDC](https://eepublicdownloads.azureedge.net/clean-documents/digital/Quality%20Of%20CGMES%20Datasets%20and%20Calculations%20v4.1.4_clean.pdf) rule `DCLineSegmentR`. On my side I have no use case for DcLine negative resistance, but maybe in a DC detailed model we could also have equivalents with negative resistance, although CGMES does not support it? Any opinion on this @SGI-Landry ? Should we allow negative DC Line R in detailed model, or is it OK as it is now?

I trust @JB-H and @geofjamg on the issue of network equivalents. I see no urgency in changing the behavior for the detailed DC model created by @jeandemanged . We may want to use the detailed DC model for equivalent networks in the future, so the comment may be a bit awkward.

jeandemanged:
Hi @NathanDissoubray, @SGI-Landry,

I fully agree with @geofjamg and @JB-H comments from #3576, and I shared the exact same concerns in that PR, **but these comments in #3576 are related to AC Lines and AC transformers, not DcLine, not HvdcLine**.

My question is about HvdcLine. @rcourtier requests to forbid negative R for simplified DC model HvdcLine, and I also totally agree there, I think this should be done, and it is not done currently in this PR. Simplified model does not mean equivalent model, and I do no see any use case for negative R in the simplified DC model.

Regarding DC detailed model, I think we can leave as it is (negative R forbidden).

NathanDissoubray:
Hello @jeandemanged and @SGI-Landry, thank you for your comments.
I think I got confused between the DC and AC line since the original issue talked about those too (thus the other PR one AC lines and transformer which was closed). I'll modify the doc regarding HVDC line on this PR and I'll remove the close of the issue #3527 

I'll open another PR to forbid negative R values for HVDC lines (which will close #3527). That way, this current PR only stays as a doc PR and doesn't try to do too many things at once. I'll request your review once it's open @jeandemanged if it's okay with you.

NathanDissoubray:
I removed the doc on the HVDC line that I added (that said negative R values are allowed). I'll add the doc in the other PR at the same time as the code that forbids negative R values

NathanDissoubray:
Opened #4062 to forbid negative R values on HVDC lines

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcLineAdder.java
```diff
@@ -12,6 +12,12 @@
  */
 public interface DcLineAdder extends IdentifiableAdder<DcLine, DcLineAdder> {
 
+    /**
+     * Set the resistance of the DC line in &#937. The resistance value should be positive, since this is a detailed
+     * model, it is a direct physical representation.
+     * @param r the resistance value of the DC line in &#937, should be positive
+     * @return this adder, for chaining
+     */
```
jeandemanged, line 20: it is a strong requirement, should -> must


```suggestion
    /**
     * Set the resistance of the DC line in &#937. The resistance value must be positive, since this is a detailed
     * model, it is a direct physical representation.
     * @param r the resistance value of the DC line in &#937, must be positive
     * @return this adder, for chaining
     */
```

jeandemanged (changes requested):

jeandemanged (approved):

PR #4046, opened 2026-08-12
Use ToolOptions in remaining CLI tools

### LINKED ISSUE #687: Generalize the use of ToolOptions utility in Tools

None. The goal is to improve code readability and mutualization by using the `ToolOptions` utility in command line `Tool`s. It facilitates conversion of command line arguments to java objects, and provides `Optional`s for better semantics in the handling of optional arguments.

Currently command line options are requested and parsed into objects with repeated code in each `Tool` implementation.

The motivation is code quality.

PowSyBl Version: 2.4.0-SNAPSHOT

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)

anishtilekar:
Closes #687.

Went through the list of Tools that Alice pointed out in the issue and switched them over to `ToolOptions` instead of pulling values straight off the `CommandLine`. Same behavior everywhere, just less repeated `line.hasOption(...) ? line.getOptionValue(...) : ...` boilerplate.

Tools updated:
- ActionSimulatorTool
- DynamicSimulationTool
- ListDynamicSimulationModelsTool
- ConversionTool
- RunLoadFlowTool
- ValidationTool
- RunScriptTool
- CompareSecurityAnalysisResultsTool
- SensitivityAnalysisTool

Didn't touch `PluginsInfoTool` and `VersionTool` even though they were on the list - they don't actually declare any command line options, so there's nothing for `ToolOptions` to wrap. Happy to add something there if I'm missing the point though.

One side effect: `ActionSimulatorToolTest` mocked `CommandLine.getOptionValue()` for the required `case-file`/`dsl-file` options but never stubbed `hasOption()` for them. `ToolOptions.getValue()` checks `hasOption()` first, so 3 tests started failing until I added the missing stubs (same pattern already used in `SecurityAnalysisToolTest`).

Ran `mvn test` on all the touched modules, everything green, checkstyle happy.

### DIFF
--- a/action-ial/action-ial-simulator/src/main/java/com/powsybl/action/ial/simulator/tools/ActionSimulatorTool.java
+++ b/action-ial/action-ial-simulator/src/main/java/com/powsybl/action/ial/simulator/tools/ActionSimulatorTool.java
@@ -27,8 +27,8 @@ import com.powsybl.security.Security;
 import com.powsybl.security.SecurityAnalysisResult;
 import com.powsybl.security.converter.SecurityAnalysisResultExporters;
 import com.powsybl.tools.Command;
-import com.powsybl.tools.CommandLineUtil;
 import com.powsybl.tools.Tool;
+import com.powsybl.tools.ToolOptions;
 import com.powsybl.tools.ToolRunningContext;
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.Option;
@@ -42,7 +42,6 @@ import java.io.IOException;
 import java.io.OutputStreamWriter;
 import java.io.PrintStream;
 import java.io.Writer;
-import java.nio.file.FileSystem;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.*;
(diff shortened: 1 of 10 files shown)


action-ial/action-ial-simulator/src/main/java/com/powsybl/action/ial/simulator/tools/ActionSimulatorTool.java
```diff
@@ -194,61 +193,61 @@ private static LoadFlowActionSimulatorObserver createCaseExporter(Path outputCas
...
     public void run(CommandLine line, ToolRunningContext context) throws Exception {
-        Path caseFile = context.getFileSystem().getPath(line.getOptionValue(CASE_FILE));
-        Path dslFile = context.getFileSystem().getPath(line.getOptionValue(DSL_FILE));
-        List<String> contingencies = line.hasOption(CONTINGENCIES) ? Arrays.stream(line.getOptionValue(CONTINGENCIES).split(",")).collect(Collectors.toList())
-                                                                     : Collections.emptyList();
-        boolean verbose = line.hasOption(VERBOSE);
-        boolean applyIfSolved = line.hasOption(APPLY_IF_SOLVED_VIOLATIONS);
-        boolean isSubTask = line.hasOption(TASK_COUNT);
+        ToolOptions options = new ToolOptions(line, context);
+        Path caseFile = options.getPath(CASE_FILE).orElseThrow(IllegalStateException::new);
```
alicecaron: for required options throw a missing required option ParseException like in other tools implementation.
```suggestion
        Path caseFile = options.getPath(CASE_FILE).orElseThrow(() -> new ParseException("Missing required option: " + CASE_FILE));
```
Maybe it could be interesting, to handle this option check homogeneously, to have in the ToolOptions class some `getRequiredPath`, `getRequiredValue`, `getRequiredFloat` etc...  methods that return the option value or throws this error ? or to add a parameter `isRequired` in the existing methods (then it would imply a breaking change in tool options usages)
What do you think @olperr1 ?

alicecaron: and rebase this branch on this one `Partharsid:tool-options-conversion-tool` once it is merged to also handle th refacto on it

anishtilekar: good call, swapped those over to ParseException with the missing-option message so it matches the rest of the codebase.

also pulled my ConversionTool changes back out since #4047 already covers that file, no point having two PRs fight over it. will rebase this branch once that one lands.

PR #4048, opened 2026-08-15
bound declared matrix dimensions before allocation in powerfactory

naruto-lgtm:

No

Security fix

The PowerFactory readers size a `BlockRealMatrix` from a row/column count taken from the input before checking that the input actually carries that many values. In `DgsParser.ElementDataMatrix.read` the `SIZEROW` field of a matrix data row flows straight into `new BlockRealMatrix(actualRows, actualCols)`. The vector reader validates its length and the matrix column count is validated too, but the row count is only logged on mismatch, so a small `.dgs` declaring a huge `SIZEROW` allocates (and overflows) an arbitrarily large matrix. `DataObject.parseMatrixJson` (study-case JSON, reachable from `StudyCase.readJson` / `Project.parseJson`) and `DataObjectBuilder.setDoubleMatrixAttributeValue` (native DB reader) have the same gap.

Each of the three sites checks the declared `rows * cols` against the values actually present before allocating, mirroring the length check the vector reader already performs. Valid files behave exactly as before; a matrix declaring more cells than it carries is skipped (DGS, consistent with the reader's existing empty-on-missing-cell handling) or rejected (JSON/DB) instead of driving the allocation. Regression tests cover the DGS and JSON paths.

- [ ] Yes
- [x] No

### DIFF
--- a/powerfactory/powerfactory-db/src/main/java/com/powsybl/powerfactory/db/DataObjectBuilder.java
+++ b/powerfactory/powerfactory-db/src/main/java/com/powsybl/powerfactory/db/DataObjectBuilder.java
@@ -121,6 +121,10 @@ public class DataObjectBuilder {
 
     public void setDoubleMatrixAttributeValue(long objectId, String attributeName, int rowCount, int columnCount, List<Double> value) {
         DataObject object = getObjectById(objectId);
+        if (rowCount < 0 || columnCount < 0 || (long) rowCount * columnCount > value.size()) {
+            throw new PowerFactoryException("Matrix data holds " + value.size()
+                    + " values, less than the declared " + rowCount + "x" + columnCount);
+        }
         RealMatrix matrix = new BlockRealMatrix(rowCount, columnCount);
         for (int row = 0; row < rowCount; row++) {
             for (int col = 0; col < columnCount; col++) {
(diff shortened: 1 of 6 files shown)


naruto-lgtm:
any update?

marqueslanauja (commented):
We have reviewed the code, and here are our comments:

- We propose centralizing the validation logic in a single function within the `DataObject` class and separating the validation into two steps to make the exception messages clearer.
- We also consider 0 to be an invalid dimension.

The validation code we propose to include in the `DataObject` class could look like thiss:

```
   public static void checkValidRealMatrix(int rowCount, int columnCount, int dataSize) {
        if (rowCount <= 0 || columnCount <= 0) {
            throw new PowerFactoryException(
                    "Matrix dimensions must be greater than zero: "
                            + rowCount + "x" + columnCount);
        }
        if (!isDataSufficientForRealMatrix(rowCount, columnCount, dataSize)) {
            throw new PowerFactoryException(
                    "Matrix data holds " + dataSize
                            + " values, less than the declared "
                            + rowCount + "x" + columnCount);
        }
    }

    public static boolean isDataSufficientForRealMatrix(int rowCount, int columnCount, int dataSize) {
        long expectedData = (long) rowCount * columnCount;
        return expectedData > 0 && expectedData <= dataSize;
    }
```

Then, in `DataObject` and `DataObjectBuilder`, we can use `checkValidRealMatrix`, while in `DgsParser `we can use `isDataSufficientForRealMatrix`

Also, you should address the two issues reported by the Sonar check. The complexity issue will probably disappear once the validation is extracted into a separate method.

PR #4049, opened 2026-08-18
Fix action-ial and psse builds on a fresh clone

jletroui:

No.

Deactivate tests on 2 modules that have no tests and will probably never have.

On a fresh clone, getting this error for `powsybl-action-ial-dsl-spi` and `psse-model-test`:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.5.5:test (default-test) on project powsybl-action-ial-dsl-spi: groups/excludedGroups require TestNG, JUnit48+ or JUnit 5 (a specific engine required on classpath) on project test classpath -> [Help 1]
```

No error, the build passes.

- [ ] Yes
- [x] No

** Possible alternative **

The downside of the current approach is that if a test is added in the module later, it will be ignored by the build. This would happen if the contributor is not paying attention to the maven output and do not realize the tests have been skipped.

A possible alternative is to manually add the test engine dependency in those 2 modules. This is quieting surefire, and if we add a test later, it will be picked up by the build. But this would add a dependency that is literally not used today. If and when a test would be added, people would need to be careful to realize they have to cleanup the dependencies.

I went with the proposed solution as the semantics seems more aligned with the problem and less surprising.

### DIFF
--- a/action-ial/action-ial-dsl-spi/pom.xml
+++ b/action-ial/action-ial-dsl-spi/pom.xml
@@ -35,6 +35,14 @@
                     </archive>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <!-- This is addressing https://github.com/apache/maven-surefire/issues/2795 -->
+                    <skipTests>true</skipTests>
+                </configuration>
+            </plugin>
         </plugins>
     </build>
 
--- a/psse/psse-model-test/pom.xml
+++ b/psse/psse-model-test/pom.xml
@@ -33,6 +33,14 @@
                     </archive>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <!-- This is addressing https://github.com/apache/maven-surefire/issues/2795 -->
+                    <skipTests>true</skipTests>
+                </configuration>
+            </plugin>
         </plugins>
     </build>
 


olperr1:
Thank you for your contribution.
Could you provide us with more details on your OS and the command you used?

We run a full compilation with the unit tests from a fresh clone every night on Linux, Windows and MacOS, and we haven't encountered this problem yet. I also just made a test on Linux and everything was fine.

Could you also try to call maven with the `-Dmaven.repo.local=<path>` option, with `<path>` being an empty directory?
For instance, on Linux:
```shell
mvn -Dmaven.repo.local=/tmp/m2_repository clean install
```

This will use the specified path as the local maven repository. Since it is empty, all the dependencies will be downloaded. This will ensure that the maven environment is clean.

jletroui:
Sure, here is the information (from main:0c26bacb4d1219b88336b0a40c21cfe8b041ed9e).

The command I used was `./mvnw package` on a fresh linux install (no prior maven installation).

My system:

```
    Operating System: Nobara Linux 44 (KDE Plasma Desktop Edition)
         CPE OS Name: cpe:/o:nobaraproject:nobara:44
      OS Support End: Wed 2026-12-02
OS Support Remaining: 3month 1w 6d
              Kernel: Linux 7.1.4-200.nobara.fc44.x86_64
        Architecture: x86-64
     Hardware Vendor: Micro-Star International Co., Ltd.
      Hardware Model: MS-7D91
    Hardware Version: 4.0
    Firmware Version: H.C1
       Firmware Date: Tue 2024-05-14
        Firmware Age: 2y 3month 4d     
```

Attached, the log content when running `./mvnw -Dmaven.repo.local=/tmp/m2_repository clean install >clean_install.log`.

[clean_install.log](https://github.com/user-attachments/files/31190625/clean_install.log)

olperr1:
In the log file you've shared, the error is not the same:
`[ERROR] Failed to execute goal org.apache.maven.plugins:maven-clean-plugin:3.5.0:clean (default-clean) on project powsybl-psse-converter: Failed to clean project: Failed to delete [...]/powsybl-core/psse/psse-converter/target -> [Help 1]`

Here, no test failed, but maven could not delete a "target" directory. My guess is that this directory could have been locked by another process (IDE, antivirus, ...).
Could you try to clone the repository in `/tmp` and directly run the maven compilation/test on it?

Also, I made a test using the maven wrapper (`mvnw`), and the command worked too.

jletroui:
> Could you try to clone the repository in /tmp and directly run the maven compilation/test on it?

Sure. Even a simple `./mvn package` is not reproducing the issue on that new fresh clone (even with system wide Maven cache).


> the error is not the same

Indeed. However, `./mvnw -Dmaven.repo.local=/tmp/m2_repository3 package >package.log` is yielding the same one:

[package.log](https://github.com/user-attachments/files/31231201/package.log)

Comparing the 2 copy directories after `./mvnw clean`, I remarked that the `/target` directories where held open by VSCode, as you correctly guessed. This would also explain the error on `./mvnw clean install`. After closing VSCode and cleaning again, this time, I could not reproduce the issue, even on the original clone.

So it seems a transient issue caused by some interaction with VSCode. Sorry for the noise, and thank you for exploring the issue.

PR #4051, opened 2026-08-20
fix naming strategy when creating busbarSection

EtienneLt:

Bug fix

when inserting a new busbar between two busbar the switch kind where the busbar will be inserted is send to the naming strategy but it shouldn't as the modification will probably change this switch kind

fix

- [ ] Yes
- [x] No


iidm/iidm-modification/src/test/java/com/powsybl/iidm/modification/topology/CreateVoltageLevelSectionsTest.java
```diff
@@ -433,4 +433,41 @@ void testWithNamingStrategy() {
                 .toList();
         assertTrue(busbarSectionIds.containsAll(List.of("VL1_1_LOAD_BREAKER_2_BREAKER", "VL1_2_LOAD_BREAKER_2_BREAKER", "VL1_3_LOAD_BREAKER_2_BREAKER")));
     }
+
+    @Test
+    void testNamingStrategy() {
```
colineplqt: Can you rename this test because there is already a `testWithNamingStrategy`?
I would suggest something like `testNamingStrategyWhenReplacingBreakerWithDisconnectors` for instance.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/CreateVoltageLevelSections.java
```diff
@@ -229,7 +229,12 @@ private void createSection(CreateSectionParameters createSectionParameters) {
         SwitchKindsBetweenBusbarSectionsTraverser switchKindsBetweenBusbarSectionsTraverser = new SwitchKindsBetweenBusbarSectionsTraverser(busbarSection);
         busbarSection.getTerminal().traverse(switchKindsBetweenBusbarSectionsTraverser);
         leftSwitchesBetweenBusbar.addAll(switchKindsBetweenBusbarSectionsTraverser.getLeftSwitchesBetweenBusbar());
-        rightSwitchesBetweenBusbar.addAll(switchKindsBetweenBusbarSectionsTraverser.getRightSwitchesBetweenBusbar());
+        List<SwitchKind> rightSwitchesBetweenBusbarFromTraverser = switchKindsBetweenBusbarSectionsTraverser.getRightSwitchesBetweenBusbar();
+        // need to remove the first switch because it will be replaced by rightSwitchKind only if there is a left busbar
```
colineplqt: I think the comment could be changed to something like that :
```suggestion
        // The first switch kind is the existing connection to the neighbouring busbar. This connection is removed when inserting the new busbar, so it must not be passed to the naming strategy.
```

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/CreateVoltageLevelSections.java
```diff
@@ -229,7 +229,12 @@ private void createSection(CreateSectionParameters createSectionParameters) {
         SwitchKindsBetweenBusbarSectionsTraverser switchKindsBetweenBusbarSectionsTraverser = new SwitchKindsBetweenBusbarSectionsTraverser(busbarSection);
         busbarSection.getTerminal().traverse(switchKindsBetweenBusbarSectionsTraverser);
         leftSwitchesBetweenBusbar.addAll(switchKindsBetweenBusbarSectionsTraverser.getLeftSwitchesBetweenBusbar());
-        rightSwitchesBetweenBusbar.addAll(switchKindsBetweenBusbarSectionsTraverser.getRightSwitchesBetweenBusbar());
+        List<SwitchKind> rightSwitchesBetweenBusbarFromTraverser = switchKindsBetweenBusbarSectionsTraverser.getRightSwitchesBetweenBusbar();
```
colineplqt: I think it would be better to create a new object from the right switches returned by the traverser.
Two reasons: 
- The object is modified, so if we rely later on it, it won't be correct.
- it makes it clearer that this list is only to be used for naming purposes.

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/CreateVoltageLevelSections.java
```diff
@@ -229,7 +229,12 @@ private void createSection(CreateSectionParameters createSectionParameters) {
...
+        // need to remove the first switch because it will be replaced by rightSwitchKind only if there is a left busbar
+        if (!rightSwitchesBetweenBusbarFromTraverser.isEmpty() && leftSwitchKind != null) {
```
colineplqt: I don't understand why it is linked to `leftSwitchKind` here

colineplqt: I'm also wondering if this processing should not be done closer to the `createBusbarSection` call. What do you think?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/CreateVoltageLevelSections.java
```diff
@@ -229,7 +229,8 @@ private void createSection(CreateSectionParameters createSectionParameters) {
         SwitchKindsBetweenBusbarSectionsTraverser switchKindsBetweenBusbarSectionsTraverser = new SwitchKindsBetweenBusbarSectionsTraverser(busbarSection);
         busbarSection.getTerminal().traverse(switchKindsBetweenBusbarSectionsTraverser);
         leftSwitchesBetweenBusbar.addAll(switchKindsBetweenBusbarSectionsTraverser.getLeftSwitchesBetweenBusbar());
-        rightSwitchesBetweenBusbar.addAll(switchKindsBetweenBusbarSectionsTraverser.getRightSwitchesBetweenBusbar());
+        List<SwitchKind> rightSwitchesBetweenBusbarFromTraverser = getRightSwitchesBetweenBusbarFromTraverser(switchKindsBetweenBusbarSectionsTraverser);
```
etiennehomer: give only switchKindsBetweenBusbarSectionsTraverser.getRightSwitchesBetweenBusbar() as a parameter to this method

etiennehomer (dismissed):
Works fine in gridsuite

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/CreateVoltageLevelSections.java
```diff
@@ -269,6 +271,16 @@ private void createSection(CreateSectionParameters createSectionParameters) {
         }
     }
 
+    private List<SwitchKind> getRightSwitchesBetweenBusbarFromTraverser(List<SwitchKind> rightSwitchesBetweenBusbarFromTraverser) {
```
colineplqt: What about something like `removeExistingConnectionSwitchKind` ?

iidm/iidm-modification/src/main/java/com/powsybl/iidm/modification/topology/CreateVoltageLevelSections.java
```diff
@@ -269,6 +271,16 @@ private void createSection(CreateSectionParameters createSectionParameters) {
...
+        List<SwitchKind> rightSwitchesBetweenBusbar = new ArrayList<>(rightSwitchesBetweenBusbarFromTraverser);
+        // The first switch kind is the existing connection to the neighboring busbar.
+        // This connection is removed when inserting the new busbar, so it must not be passed to the naming strategy.
+        if (!rightSwitchesBetweenBusbar.isEmpty() && leftSwitchKind != null) {
```
colineplqt: I have a minor comment, tell me if you disagree. 
It was not clear to me why you checked `leftSwitchKind != null` and I am wondering if extracting a little helper method would not help with understanding. Something like:

```java
private boolean isCreatingSectionsAfterReferenceBusbar() {
    return leftSwitchKind != null;
}
```

colineplqt (commented):
Two minor comments, I don't think it's really necessary if you disagree with them. Otherwise, it looks good to me.

colineplqt (approved):

etiennehomer (approved):

PR #4053, opened 2026-08-20
Deprecate and add a remplacement function to get country name

### LINKED ISSUE #3571: Confuse `Country` API
The [`com.powsybl.iidm.network.Country`](//github.com/powsybl/powsybl-core/blob/v6.8.1/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Country.java) define a [`name`](//github.com/powsybl/powsybl-core/blob/v6.8.1/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Country.java#L267) field for the name of the country with a getter [`getName()`](//github.com/powsybl/powsybl-core/blob/v6.8.1/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Country.java#L273-L275).

But as it is an [`Enum`](//docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Enum.html), it already have a `name` attribute accessible with methods [`name()`](//docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Enum.html#name()) and [`toString()`](//docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Enum.html#toString()) which confuse me and I found error prone when using the API.

What do you think of using a more explicit or distinct attribute name like "countryName"?

lejeana:

Fixes #3571 

Deprecate the function getName() as a function called name() already exist which lead to confusion. Added a new function with the same behaviour and whose name is more explicit : getFullName().

#3571

Deprecated a function and add a remplacement function with the same behaviour.

- [x] Yes
- [ ] No

Change call to function `Country::getName()` to `Country::getFullName()` as `Country::getName()` is now deprecated and will later be removed.  
Both functions have the same behaviour, hence no more change is required.

### DIFF
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Country.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Country.java
@@ -274,7 +274,7 @@ public enum Country {
      * @deprecated This method is deprecated because not explicit enough as function name() exists.
      * Use {@link #getFullName()} instead.
      */
-    @Deprecated
+    @Deprecated(forRemoval=false)
     public String getName() {
         return name;
     }
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Country.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/Country.java
@@ -270,8 +270,16 @@ public enum Country {
         this.name = name;
     }
 
+    /**
+     * @deprecated This method is deprecated because not explicit enough as function name() exists.
+     * Use {@link #getFullName()} instead.
+     */
+    @Deprecated
     public String getName() {
         return name;
     }
 
+    public String getFullName() {
+        return name;
+    }
 }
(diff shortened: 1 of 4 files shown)


alicecaron (dismissed):

colineplqt (commented):
Thank you for this pull request. Could you modify the code so that the calls to getName() are replaced by getFullName?

lejeana:
calls to getName() are now replaced by the new function getFullName()

also, fixture _deprecated_ allows to precise if the function will be removed in futher version, should we set it to true?

colineplqt:
> calls to getName() are now replaced by the new function getFullName()
> 
> also, fixture _deprecated_ allows to precise if the function will be removed in futher version, should we set it to true?

I think you can add the version since when the method is deprecated (7.4.0). I would say that it is a function that could be removed after some time, but both is fine.

lejeana:
> I think you can add the version since when the method is deprecated (7.4.0). I would say that it is a function that could be removed after some time, but both is fine.

Changes have been done and the _fixture_ precise that the function will be removed.

alicecaron (approved):

colineplqt (approved):

PR #4056, opened 2026-08-21
Fix performance issue for copy of network in XML

NathanDissoubray:

None

Performance

Following bump of core version in powsybl-benchmark from 7.2.1 to 7.3.0 leads to degraded copy performance when using XML (about 10 times slower):

Core 7.2.1
```
| Network name | XML (XIIDM) | JSON (JIIDM) | Binary (BIIDM) |
|--------------|-------------|--------------|----------------|
| IEEE 14      | 0,61        | 0,47         | 0,31           |
| IEEE 118     | 4,93        | 3,85         | 2,61           |
| IEEE 300     | 10,98       | 8,52         | 5,76           |
| RTE 1888     | 124,44      | 81,32        | 53,53          |
| RTE 6515     | 348,57      | 251,73       | 178,85         |
| RealGrid     | 1381,19     | 873,97       | 539,26         |
```
Core 7.3.0
```
| Network name | XML (XIIDM) | JSON (JIIDM) | Binary (BIIDM) |
|--------------|-------------|--------------|----------------|
| IEEE 14      | 6,13        | 0,27         | 0,32           |
| IEEE 118     | 59,61       | 1,83         | 2,01           |
| IEEE 300     | 129,55      | 3,92         | 4,53           |
| RTE 1888     | 1650,76     | 37,60        | 39,45          |
| RTE 6515     | 4085,79     | 131,93       | 137,54         |
| RealGrid     | 20910,20    | 381,44       | 326,14         |
```

Added use of buffered channel for formats other than binary, here is the resulting benchmark:
```
| Network name | XML (XIIDM) | JSON (JIIDM) | Binary (BIIDM) |
|--------------|-------------|--------------|----------------|
| IEEE 14      | 0,44        | 0,32         | 0,32           |
| IEEE 118     | 3,05        | 1,80         | 1,99           |
| IEEE 300     | 6,68        | 4,04         | 4,34           |
| RTE 1888     | 76,39       | 37,09        | 39,21          |
| RTE 6515     | 220,80      | 130,42       | 135,88         |
| RealGrid     | 821,72      | 364,11       | 314,06         |
```
It's about 40% faster than core 7.2.1 and about 25 times faster than core 7.3.0. It's still slower than JSON and IIDM which both got their performance improved by a factor 2 when going from core 7.2.1 to core 7.3.0

- [ ] Yes
- [x] No

**Other information**:

I investigated a bit the cause of this slow down.
#3919 introduced use of channels for binary, and using buffered channel for json and xml, which is where we see the improvement (result similar to after this fix)
```
| Network name | XML (XIIDM) | JSON (JIIDM) | Binary (BIIDM) |
|--------------|-------------|--------------|----------------|
| IEEE 14      | 0,42        | 0,32         | 0,22           |
| IEEE 118     | 3,01        | 1,85         | 1,79           |
| IEEE 300     | 6,64        | 4,11         | 4,01           |
| RTE 1888     | 74,59       | 37,60        | 37,50          |
| RTE 6515     | 214,57      | 129,65       | 133,84         |
| RealGrid     | 814,54      | 371,41       | 301,48         |
```

The PR #3968 cleans up the code and uses a non buffered channel sink for all formats, which is where the slow down happens (results similar to bumping to core 7.3.0)
```
| Network name | XML (XIIDM) | JSON (JIIDM) | Binary (BIIDM) |
|--------------|-------------|--------------|----------------|
| IEEE 14      | 0,42        | 0,32         | 0,22           |
| IEEE 118     | 3,01        | 1,85         | 1,79           |
| IEEE 300     | 6,64        | 4,11         | 4,01           |
| RTE 1888     | 74,59       | 37,60        | 37,50          |
| RTE 6515     | 214,57      | 129,65       | 133,84         |
| RealGrid     | 814,54      | 371,41       | 301,48         |
```

So the fix of this PR is to use buffered channel or sink depending on the format (but still keeping the clean code of #3968)

The main speedup comes from the write, below is a benchmark where only the write part of the copy uses buffered channel:
```
| Network name | XML (XIIDM) | JSON (JIIDM) | Binary (BIIDM) |
|--------------|-------------|--------------|----------------|
| IEEE 14      | 0,43        | 0,32         | 0,32           |
| IEEE 118     | 2,94        | 1,82         | 1,97           |
| IEEE 300     | 6,51        | 4,03         | 4,36           |
| RTE 1888     | 75,05       | 37,59        | 38,91          |
| RTE 6515     | 213,62      | 127,07       | 136,14         |
| RealGrid     | 832,32      | 358,27       | 325,87         |
```
And here is the current fix, where both write and read use buffered channels for XML:
```
| Network name | XML (XIIDM) | JSON (JIIDM) | Binary (BIIDM) |
|--------------|-------------|--------------|----------------|
| IEEE 14      | 0,44        | 0,32         | 0,32           |
| IEEE 118     | 3,05        | 1,80         | 1,99           |
| IEEE 300     | 6,68        | 4,04         | 4,34           |
| RTE 1888     | 76,39       | 37,09        | 39,21          |
| RTE 6515     | 220,80      | 130,42       | 135,88         |
| RealGrid     | 821,72      | 364,11       | 314,06         |
```
As we can see, the results are similar, so we could choose to not use buffered channel for read, but it was done this way here to have a symmetric behavior.

### DIFF
--- a/iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
+++ b/iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
@@ -1295,14 +1295,18 @@ public final class NetworkSerDe {
         try {
             Pipe pipe = Pipe.open();
             executor.execute(() -> {
-                try (Pipe.SinkChannel sinkChannel = pipe.sink()) {
-                    write(network, new ExportOptions().setFormat(format), Channels.newOutputStream(sinkChannel));
+                try (OutputStream tmp = Channels.newOutputStream(pipe.sink());
+                     //using buffered stream is about 20 times more effective for xml and json formats
+                     OutputStream os = format == TreeDataFormat.BIN ? tmp : new BufferedOutputStream(tmp)) {
+                    write(network, new ExportOptions().setFormat(format), os);
                 } catch (Exception t) {
                     LOGGER.error(t.toString(), t);
                 }
             });
-            try (Pipe.SourceChannel sourceChannel = pipe.source()) {
-                return read(Channels.newInputStream(sourceChannel),
+            try (InputStream tmp = Channels.newInputStream(pipe.source());
+                 //using buffered stream for read has little impact, but it mimics the write behavior
+                 InputStream is = format == TreeDataFormat.BIN ? tmp : new BufferedInputStream(tmp)) {
+                return read(is,
                         new ImportOptions().setFormat(format), null, networkFactory, ReportNode.NO_OP);
             }
         } catch (IOException e) {
--- a/iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
+++ b/iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
@@ -1296,7 +1296,7 @@ public final class NetworkSerDe {
             Pipe pipe = Pipe.open();
             executor.execute(() -> {
                 try (OutputStream tmp = Channels.newOutputStream(pipe.sink());
-                     //using buffered stream is about 20 times more effective for xml and json formats
+                     //using buffered stream is about 20 times more effective for xml
                      OutputStream os = format == TreeDataFormat.BIN ? tmp : new BufferedOutputStream(tmp)) {
                     write(network, new ExportOptions().setFormat(format), os);
                 } catch (Exception t) {
--- a/iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
+++ b/iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
@@ -1297,7 +1297,7 @@ public final class NetworkSerDe {
             executor.execute(() -> {
                 try (OutputStream tmp = Channels.newOutputStream(pipe.sink());
                      //using buffered stream is about 20 times more effective for xml
-                     OutputStream os = format == TreeDataFormat.BIN ? tmp : new BufferedOutputStream(tmp)) {
+                     OutputStream os = format == TreeDataFormat.XML ? new BufferedOutputStream(tmp) : tmp) {
                     write(network, new ExportOptions().setFormat(format), os);
                 } catch (Exception t) {
                     LOGGER.error(t.toString(), t);
@@ -1305,7 +1305,7 @@ public final class NetworkSerDe {
             });
             try (InputStream tmp = Channels.newInputStream(pipe.source());
                  //using buffered stream for read has little impact, but it mimics the write behavior
-                 InputStream is = format == TreeDataFormat.BIN ? tmp : new BufferedInputStream(tmp)) {
+                 InputStream is = format == TreeDataFormat.XML ? new BufferedInputStream(tmp) : tmp) {
                 return read(is,
                         new ImportOptions().setFormat(format), null, networkFactory, ReportNode.NO_OP);
             }


skrcode:
I ran JAIPilot Cloud against this exact PR head. It confirmed that the new write-side buffering is the material XML fix, then found one small cleanup: the additional read-side BufferedInputStream has no reproducible benefit and allocates its own 8 KB buffer for every copy.

The same 37 focused tests passed before and after, with byte-identical BIN, XML, and JSON round trips. Thirty warmups plus 40 measured iterations, repeated three times per format, found no reproducible regression, and the iidm-serde module suite with dependencies passed.

PR directly onto this source branch: https://github.com/powsybl/powsybl-core/pull/4061  
Cloud-generated draft and full evidence: https://github.com/skrcode/powsybl-core/pull/2

The write buffer remains untouched. This is an optional resource cleanup; no throughput improvement is claimed.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -1295,14 +1295,18 @@ public static Network copy(Network network, NetworkFactory networkFactory, Execu
         try {
             Pipe pipe = Pipe.open();
             executor.execute(() -> {
-                try (Pipe.SinkChannel sinkChannel = pipe.sink()) {
-                    write(network, new ExportOptions().setFormat(format), Channels.newOutputStream(sinkChannel));
+                try (OutputStream tmp = Channels.newOutputStream(pipe.sink());
+                     //using buffered stream is about 20 times more effective for xml and json formats
```
flo-dup: Isn't this only for xml? at least from the benchmarks table you mentioned in the PR description

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -1295,14 +1295,18 @@ public static Network copy(Network network, NetworkFactory networkFactory, Execu
...
+                     OutputStream os = format == TreeDataFormat.BIN ? tmp : new BufferedOutputStream(tmp)) {
```
flo-dup: Why don't you get some slight improvement for binary too?

flo-dup: Why isn't this handled later by createXmlWriter/createJsonWriter/createBinWriter? If that's improving so much performance we should probably ensure it's always done with a buffered output stream.

flo-dup (changes requested):

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
NathanDissoubray: yes you are right, I forgot to update the comment after I did more testing

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
NathanDissoubray: I'm not sure of the reason. This might be due to the way we write xml compared to json/binary. Or it has something to do with the format itself

NathanDissoubray: I changed it here only because I noticed a regression with the copy when bumping core version in powsybl-benchmark. As such, I have only made changes for the copy.

I could check to see if a bigger refactoring of the logic in `createXmlWriter` and other related writer creator would also have an impact on the performance of other operations.

However, the aim of this PR is to make a correction for a regression, not to improve the performance of all serde operations, so I think it would be better left for another PR.

NathanDissoubray: I wasn't sure I tested using buffered channel for binary too, so I did, here are the results:
```
| Network name | XML (XIIDM) | JSON (JIIDM) | Binary (BIIDM) |
|--------------|-------------|--------------|----------------|
| IEEE 14      | 0,44        | 0,34         | 0,32           |
| IEEE 118     | 2,96        | 1,85         | 2,02           |
| IEEE 300     | 6,63        | 4,08         | 4,31           |
| RTE 1888     | 75,25       | 39,43        | 39,16          |
| RTE 6515     | 221,70      | 128,78       | 140,88         |
| RealGrid     | 828,00      | 367,99       | 318,11         |
```

For reference:
```
| Network name | XML (XIIDM) | JSON (JIIDM) | Binary (BIIDM) |
|--------------|-------------|--------------|----------------|
| IEEE 14      | 0,44        | 0,32         | 0,32           |
| IEEE 118     | 3,05        | 1,80         | 1,99           |
| IEEE 300     | 6,68        | 4,04         | 4,34           |
| RTE 1888     | 76,39       | 37,09        | 39,21          |
| RTE 6515     | 220,80      | 130,42       | 135,88         |
| RealGrid     | 821,72      | 364,11       | 314,06         |
```
As we can see, using buffered channel on the channel for binary doesn't change anything

NathanDissoubray: I also tested for JSON, it doesn't make a difference if we use a buffered channel, it's only relevant for XML

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
NathanDissoubray: changed in 1f3f4e16995c

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
NathanDissoubray: I used the buffered channel only for XML in 8c94bc1b7e3c

flo-dup: We could have changed the aim of this PR, to avoid coming back to it later, as all the callers add a BufferedOutputStream I thought. All but one in fact, as NSD::gzip does not. Ok to keep it that way, and let's see if we come back to it one day.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
```diff
@@ -1295,14 +1295,18 @@ public static Network copy(Network network, NetworkFactory networkFactory, Execu
...
+                     OutputStream os = format == TreeDataFormat.XML ? new BufferedOutputStream(tmp) : tmp) {
+                    write(network, new ExportOptions().setFormat(format), os);
                 } catch (Exception t) {
                     LOGGER.error(t.toString(), t);
                 }
             });
-            try (Pipe.SourceChannel sourceChannel = pipe.source()) {
-                return read(Channels.newInputStream(sourceChannel),
+            try (InputStream tmp = Channels.newInputStream(pipe.source());
+                 //using buffered stream for read has little impact, but it mimics the write behavior
```
flo-dup: If we want to stick to the regression correction in this PR, there's no use of mimicing. Besides symetry's nice to look at, but as it has little impact we should forget it.

NathanDissoubray: Changed it in e121dad8b08e
Kept a comment to explain why it's not symmetric with the write

flo-dup (approved):

PR #4061, opened 2026-08-23
Remove redundant read buffering from NetworkSerDe.copy

skrcode:
This is a one-commit JAIPilot Cloud follow-up on the exact current head of [powsybl/powsybl-core#4056](https://github.com/powsybl/powsybl-core/pull/4056). It targets that PR branch directly.

### Change

Keep the material write-side buffering from #4056 and remove the read-side BufferedInputStream wrapper. BIN already used the channel stream directly, and the wrapper was documented as having little impact.

### Proof

- Exact parent: fc86ce540bb6fe8716ebbddf917cfd54e7a34a24, the current #4056 head.
- The same 37 focused tests passed before and after; BIN, XML, and JSON copies remain byte-identical.
- Thirty warmups plus 40 measured iterations, repeated three times per format, found no reproducible throughput regression from removing the read wrapper.
- The iidm-serde module suite with dependencies passed.
- One wrapper and its 8 KB buffer allocation are removed per copy.

### Boundary and disclosure

The write buffer remains the important XML improvement and is untouched. This is a small resource cleanup; no throughput improvement is claimed.

JAIPilot Cloud generated and validated the patch in [skrcode/powsybl-core#2](https://github.com/skrcode/powsybl-core/pull/2). I reviewed the exact ancestry, complete two-file diff, format-equivalence evidence, repeated timing boundary, and limitation before offering it here.

### DIFF
--- a/iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
+++ b/iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/NetworkSerDe.java
@@ -1303,9 +1303,7 @@ public final class NetworkSerDe {
                     LOGGER.error(t.toString(), t);
                 }
             });
-            try (InputStream tmp = Channels.newInputStream(pipe.source());
-                 //using buffered stream for read has little impact, but it mimics the write behavior
-                 InputStream is = format == TreeDataFormat.BIN ? tmp : new BufferedInputStream(tmp)) {
+            try (InputStream is = Channels.newInputStream(pipe.source())) {
                 return read(is,
                         new ImportOptions().setFormat(format), null, networkFactory, ReportNode.NO_OP);
             }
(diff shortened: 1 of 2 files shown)


NathanDissoubray:
The PR #4056 explicitly mentions:
> As we can see, the results are similar [between using buffered for read or direct channel], so we could choose to not use buffered channel for read, but it was done this way here to have a symmetric behavior.

I'm not sure an 8 kB improvement makes much of an impact, considering that a network can be thousands of time this size. The initial aim to have symmetric behavior between read and write was meant for code clarity. We can of course debate if it is really clearer that way; maybe it's not, in which case we could remove the buffered read for the read side of the copy.

skrcode:
Closing after reconsidering the tradeoff. The candidate showed no reproducible throughput benefit, and removing one 8 KB read buffer does not justify giving up the clearer symmetric read/write structure in #4056. Thanks @NathanDissoubray for the context.

PR #4023, opened 2026-07-29
Enforce valid reactive capability curve minQ/maxQ ordering in IIDM

cphili:
**Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

**Does this PR already have an issue describing the problem?**

https://github.com/powsybl/powsybl-core/issues/3648 

**What kind of change does this PR introduce?**

Feature.

**What is the current behavior?**

Reactive Capability Curve points can be defined in IIDM with invalid `minQ > maxQ`.

**What is the new behavior (if this is a feature change)?**

Reactive Capability Curve points are enforced by IIDM validation to have `minQ <= maxQ`.

To support re-importing older serialized IIDM files having invalid values, a new IIDM import option (`iidm.import.xml.repair-invalid-reactive-curve-limits`) is introduced.

When this option is disabled (default behavior), reactive capability curve points with minQ > maxQ remain invalid and the import fails with a ValidationException.

When this option is enabled, the XIIDM importer detects reversed minQ/maxQ values in reactive capability curve points and automatically swaps them before IIDM validation occurs. The import can therefore succeed even when the source file contains reversed values.

For users of other formats:
- **UCTE-DEF**: not applicable since UCTE-DEF only uses min/max Q limits where proper ordering was already enforced by IIDM.
- **CGMES**: [QoCDC](https://eepublicdownloads.azureedge.net/clean-documents/digital/Quality%20Of%20CGMES%20Datasets%20and%20Calculations%20v4.1.4_clean.pdf) rule `RCCYValues` enforces proper minQ/maxQ ordering with level error. Hence rejecting within PowSyBl is valid.
- Other formats (PowerFactory, PSS/E, etc...): to be assessed. In principle the IIDM validation introduced here is legitimate, however specific importers may need to also implement an auto-repairing option as well in the case these formats allow invalid values. This is not covered in this PR.

**Does this PR introduce a breaking change or deprecate an API?**
- [x] Yes
- [ ] No

**What changes might users need to make in their application due to this PR? (migration steps)**

Reactive capability curve points with minQ > maxQ are now rejected by IIDM validation.

Applications importing IIDM files containing reversed values should either fix the source data or enable the `iidm.import.repair-invalid-reactive-curve-limits` import option to automatically reorder minQ and maxQ during import.

[...]

PR #4062, opened 2026-08-24
Forbid negative R values on HVDC line

### LINKED ISSUE #3527: IIDM shouldn't accept negative r values for HVDCLines

NathanDissoubray:

Closes #3527 

Feature, doc

Negative R values are accepted for HVDC lines (both in adder and setter).

Negative R values are not accepted for HVDC lines (both in adder and setter). Using a negative a negative value will throw an exception.

- [x] Yes
- [ ] No

For implementations of IIDM api:
- ensure that all implementations of `HvdcLineAdder` check that there is no negative r value (either check when setting r or when creating the `HvdcLine` with `add()`)
- ensure that all implementations of `HvdcLine` forbid negative r values with using `setR`

For users:
- if you are using a negative R value for an `HvdcLine`, it is now not allowed in the IIDM model.

**Other information**:


iidm/iidm-api/src/main/java/com/powsybl/iidm/network/HvdcLineAdder.java
```diff
@@ -15,6 +15,12 @@
  */
 public interface HvdcLineAdder extends IdentifiableAdder<HvdcLine, HvdcLineAdder> {
 
+    //TODO explain why negative values are not allowed ?
```
jeandemanged: remove this line

jeandemanged (changes requested):
Similar comment to #4045, change all "should" (weak requirement, suggesting negative works) to "must" (strong constraint).

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/HvdcLineAdder.java
NathanDissoubray: is there a particular reason why negative values are not allowed ? I would like to add this at least in the read the doc
Removed the TODO

jeandemanged: not sure it is worth mentioning in the doc. Negative resistance means negative losses i.e. the DC line generates power, this does not make sense physically. (AC equivalents is a whole different story).

jeandemanged (dismissed):
Good for me, thank you. Only wondering if we need a repair mechanism for older serialized IIDM models with invalid values as was done in #4023. I do not feel it is necessary here, but maybe @olperr1 has another opinion.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/HvdcLineAdder.java
NathanDissoubray: I think in the javadoc maybe not, but in the read the doc it could be interesting

NathanDissoubray:
If we choose to "repair", we would need to choose a value for negative R, and I'm not sure there is a good default choice for that.

jeandemanged:
> If we choose to "repair", we would need to choose a value for negative R, and I'm not sure there is a good default choice for that.

In #4023 too there is no good choice (swapping the values), this "repair" is arbitrary and not perfect at all. Here we could make also a (not good) choice of just flipping the sign, same this would be completely arbitrary. But I think this is not needed.

olperr1:
> Good for me, thank you. Only wondering if we need a repair mechanism for older serialized IIDM models with invalid values as was done in #4023. I do not feel it is necessary here, but maybe @olperr1 has another opinion.

I talked about it with a few colleagues and we don't think it is  necessary too.
@NathanDissoubray: could you update the PR and merge the conflicts?

jeandemanged (approved):

olperr1 (approved):

PR #4064, opened 2026-08-24
Sanitize string fields in UCTE record writer

naruto-lgtm:

No

Bug fix

Repro: import a network whose `geographicalName`/`elementName` property (or a `##C` comment) holds a line break, then export it to UCTE.
Cause: `writeString` is the only fixed-width field writer with no bound on its input. `buffer.replace(beginIndex, endIndex, value)` writes the value as-is, so an overlong value shifts every following column and an embedded line break splits the single record; the exporter feeds imported-network property values straight through.

Fix: sanitize inside `writeString` - replace control characters with a space so a value cannot inject a line break, and truncate to the field width so it cannot shift the following columns. This mirrors the guards `writeInteger`/`writeDouble`/`writeEnumValue` already apply.

- [x] No


naruto-lgtm:
Added the DCO sign-off, so that check is green now. The change itself is unchanged.

jeandemanged (approved):
Thank you!

PR #4070, opened 2026-08-27
IIDM Tool. Replace IDs with names

### LINKED ISSUE #3899: Option to set a human readable id on IIDM elements

On CGMES import, CGMES rdfid is used to define the IIDM id elements.

An option exists in CGMES import to chose whether the IIDM element id should be populated by the CGMES rdfid or with another rule more human readable and ensuring the unicity of the element id within the network.

### Describe the motivation

On CGMES import the IIDM elements id is populated with the CGMES rdfid. This way the id unicity is guaranteed but is not human readable.

### Extra Information

- add a CGMES import option (for instance: `iidm.import.cgmes.replace-rdfid`)
- default value should imply to put the CGMES rdfid into the IIDM id object (current behavior)
- at creation a check is done to ensure the element doses not have the same name as another one in the network. If it is the case, instead of putting only the `<CGMES element name>` as the id, set the `<CGMES element name>_<CGMES rdfid>`. For this case add a WARNing message.
- when the CGMES import option is activated, store the rdfid in an alias with alias type (`cgmes-id` for instance) so that:
  - for multiple SSH updaes afterwards it will be easy to retreive he element to be updated and .
  - at CGMES export we replace the id by the `cgmes-id` alias value if present (round trip use case)

zamarrenolm:

Fixes #3899.
This is just a demonstration of the proposed tool in the issue.
It is implemented as a unit test in CGMES conversion, but the tool, if approved, should be placed in the `iidm` module.

- [ ] Yes
- [ ] No

**Other information**:


zamarrenolm:
if we apply the change as it is to the Svedala example, there is a lot of suffixed entries, coming from all fictitious switches created for disconnected terminals, because all terminals are named "Terminal".

That may happen for real-world cases. 

To avoid it, we could rename only names that are unique. An alternative implementation with this strategy is also provided

Issue #3954, opened 2026-06-12
AC Active Power bounds for AcDcConverter


SGI-Landry:
### Describe the current behavior

Unlike AC generators, AC-DC converters have no power bounds on their AC sides.

### Describe the expected behavior

IIDM representation of power bounds for AC-DC converters. Manage import/export of CGMES and DGS for this as well if available, and IIDM ser/deser-ialization.

Expected behavior
- properties minP and maxP.
- default values -Double.MAX_VALUE and Double.MAX_VALUE to keep the bounds optional.
- we must always have minP <= maxP.

Note that
- ideally we would use -inf and inf for default values but this pattern is not used elsewhere. Non-finite real values are also unsupported in JSON.
- we arguably refrain from requesting that minP <= targetP <= maxP for fear that it may be inconvenient when setting these values.

### Describe the motivation

Power bounds on the AC side of AC-DC converters is needed for some load flow analysis. This is similar to power bounds for AC generators.

PR #3940, opened 2026-06-05
Add active power limits to AC/DC converters

SGI-Landry:
**Please check if the PR fulfills these requirements**

- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)
- [x] A PR or issue has been opened in all impacted repositories (if any)

**Does this PR already have an issue describing the problem?**
https://github.com/powsybl/powsybl-core/issues/3954

**What kind of change does this PR introduce?**
Introduce AC power limits for AC-DC converters.

**What is the current behavior?**
No power bounds.

**What is the new behavior (if this is a feature change)?**
It is possible to define AC power bounds for AC-DC converters.
**Does this PR introduce a breaking change or deprecate an API?**
- [x] Yes
- [ ] No

**If yes, please check if the following requirements are fulfilled**

- [x] The *Breaking Change* or *Deprecated* label has been added
- [x] The migration steps are described in the following section

**What changes might users need to make in their application due to this PR? (migration steps)**
- Implementations of `AcDcConverterAdder` will need to implement methods `setMinP` and `setMaxP` to set power limits of ACDC converters.
- Implementations of `AcDcConverter` will need to implement methods `getMinP`, `setMinP`, `getMaxP` and `setMaxP`.
- Default values are -Double.MAX_VALUE (resp. Double.MAX_VALUE) for minP (resp. maxP) so you only need to assign these bounds if you actually need them.

**Other information**:
The intention is to carry on with the implementation in Open Load Flow after the 2026.1 release.

Serialization and deserialization are not implemented due to the short time before the release train. They are intended to be implemented in IIDM 1.18, in the 2026.2 release. This is supposed to avoid increased risk.

PR #3340, opened 2025-03-03
Deprecate creation of new LoadingLimits from Branch and FlowsLimitsHolder

quinarygio:
**Please check if the PR fulfills these requirements**

- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**Does this PR already have an issue describing the problem?**

#3212 

**What kind of change does this PR introduce?**

**What is the current behavior?**
The creation of new `LoadingLimits` are done from Branch/FlowsLimitsHolder.

**What is the new behavior (if this is a feature change)?**
The creation of new `LoadingLimits` from the Branch/FlowsLimitsHolder is now deprecated and should not be used. The creation is now done from a Branch/FlowsLimitsHolder `OperationalLimitsGroup` directly.

If the methods `getOrCreateSelectedOperationalLimitsGroupXXXX` are called from a limits holder but no `OperationalLimitsGroup` exist a default one is created and selected by default so that the `LoadingLimits` can be created on it.

**Does this PR introduce a breaking change or deprecate an API?**
- [x] Yes
- [ ] No

**If yes, please check if the following requirements are fulfilled**

- [x] The *Breaking Change* or *Deprecated* label has been added
- [ ] The migration steps are described in the following section

**What changes might users need to make in their application due to this PR? (migration steps)**

**For all:**

The methods `newCurrentLimits`, `newApparentPowerLimits` and `newActivePowerLimits` on `Branch` and `FlowsLimitsHolder` are now deprecated.

Their use should be replaced by the call to the new method `getOrCreateSelectedOperationalLimitsGroup()` (that creates an`OperationalLimitsGroup` and set it as selected, if needed) followed by a call to the limits' creation on the group.

Basically, replace `myDanglingLine.newCurrentLimits()` by  `myDanglingLine.getOrCreateSelectedOperationalLimitsGroup().newCurrentLimits()`

Note that previously, the deprecated methods only created the default group - if needed - at the limits' creation (during the call to `add()` on the resulting limits adder). Now, the default group is created by the deprecated methods. Therefore, the group is created even if `add()` is never called on the resulting adder.

**For custom IIDM implementation maintainers:**

Custom IIDM implementations' maintainers should implement the following methods:

- in their `Branch` implementations:
  - `OperationalLimitsGroup getOrCreateSelectedOperationalLimitsGroup1();`
[...]

PR #4079, opened 2026-08-28
Add limits to detailed DC components

### LINKED ISSUE #4024: Limits for detailed DC components

**Detailed DC components carry no operational limits, while their AC counterparts all do.**

- DcLine exposes its resistance, its connection status and its computed current and power but no current rating.
- DcNode exposes its nominal voltage and its solved voltage but no voltage limits.
- AcDcConverter does have minP/maxP, and they are readable and writable in memory. However a finite value cannot be serialized: exporting the network raises an exception rather than writing the limit out.

On the AC side the equivalents all exist: branches carry operational limit groups, generators carry minP/maxP, and VoltageLevel carries lowVoltageLimit/highVoltageLimit.

Because there is nothing to set, any tool that needs a DC limit has to invent a placeholder value in its own code. 

Three changes:

1. A current rating on DcLine, so a DC cable can declare what it is rated for. We would suggest reusing the existing OperationalLimitsGroup mechanism that AC branches already use, for consistency and so that temporary limits come along for free.
2. Voltage limits on DcNode, directly analogous to VoltageLevel.lowVoltageLimit and highVoltageLimit, so a DC busbar can declare its operating band.
3. Serialization support for AcDcConverter minP/maxP, so that a network carrying converter limits survives a file round trip.

### Describe the motivation

DC limits are needed for the same reasons AC limits are.

- After a load flow, an overloaded DC cable or an out-of-band DC busbar should be reportable the same way an overloaded AC line is. Today there is nothing to compare the result against.
- An OPF on a detailed DC network has to bound DC currents, DC voltages and converter power somehow.  Real attributes let the OPF respect the actual equipment ratings, and let the user see and change them.
- Limits should travel inside the network file, so a case keeps its operating limits when it is exchanged or re-loaded.

As detailed DC modelling matures, the gap between what can be described on the AC side and what can be described on the DC side should reduce.

### Extra Information

Reproducing the serialization failure:

```python
import pypowsybl.network as pn

n = pn.create_ac_dc_bipolar_network()
ids = list(n.get_voltage_source_converters().index)
n.update_voltage_source_converters(id=ids, min_p=[-400.0] * len(ids), max_p=[400.0] * len(ids))

# the limits are set correctly in memory
print(n.get_voltage_source_converters(attributes=['min_p', 'max_p']))

# but the export raises
n.save('out.xiidm', 'XIIDM')
```

thomas-durand-SGI:
- [x] A PR or issue has been opened in all impacted repositories (if any): powsybl/powsybl-network-store#675

Fixes [#4024](https://github.com/powsybl/powsybl-core/issues/4024).

It also completes part of [#3954](https://github.com/powsybl/powsybl-core/issues/3954) (open, assigned to
[@SGI-Landry](https://github.com/SGI-Landry)): the IIDM serialization of the AC/DC converter active power
limits, which [#3940](https://github.com/powsybl/powsybl-core/pull/3940) deliberately left out
(*"intended to be implemented in IIDM 1.18, in the 2026.2 release"*). The CGMES and DGS half of
[#3954](https://github.com/powsybl/powsybl-core/issues/3954) is not touched here.

Depends on [#3984](https://github.com/powsybl/powsybl-core/pull/3984) (Bump IIDM to 1.18) and is based on
`bump_iidm_18` rather than `main`, so the diff shows only this work. Draft until
[#3984](https://github.com/powsybl/powsybl-core/pull/3984) merges.

Feature.

The detailed DC model has no limits:

- `AcDcConverter.getMinP()` / `getMaxP()` exist but cannot be written or read: `AbstractAcDcConverterSerDe`
  throws `NotImplementedException`.
- `DcNode` has no voltage limits.
- `DcLine` has no current limits, so a DC cable has nowhere to record its rating.

Three additions, all gated on IIDM 1.18.

1. **`AcDcConverter` `minP`/`maxP` are serialized**, as two optional attributes.

2. **`DcNode` gains `lowVoltageLimit` and `highVoltageLimit`**, both optional, `NaN` when unset.

3. **`DcLine` becomes a [`FlowsLimitsHolder`](https://github.com/powsybl/powsybl-core/blob/main/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/FlowsLimitsHolder.java)**,
   so it carries operational limits groups the same way
   [`BoundaryLine`](https://github.com/powsybl/powsybl-core/blob/main/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/BoundaryLine.java)
   does:

```java
dcLine.newOperationalLimitsGroup("summer").newCurrentLimits().setPermanentLimit(2106.).add();
dcLine.setSelectedOperationalLimitsGroup("summer");
```

No DC-specific limits type is introduced; `OperationalLimitsGroup` and `CurrentLimits` are reused
unchanged.

**Design choices**

The rest of the diff follows existing AC or DC patterns.

*1. A `DcLine` has one set of limits for the whole line, not one set per side.*

An AC branch has two current limits because its shunt admittance (`g1/b1/g2/b2`) makes the currents at its
two terminals differ. A `DcLine`'s only electrical attribute is `r`, so the current entering one terminal
is the current leaving the other, and one limit describes both.

Two sets per side would mean writing the same rating twice:

```java
dcLine.newOperationalLimitsGroup1("summer").newCurrentLimits().setPermanentLimit(2106.).add();
dcLine.newOperationalLimitsGroup2("summer").newCurrentLimits().setPermanentLimit(2106.).add();
```

Nothing would stop 2106 on side 1 and 2400 on side 2, leaving the model claiming that one physical current
must stay under both. 

So `DcLine` follows
[`BoundaryLine`](https://github.com/powsybl/powsybl-core/blob/main/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/BoundaryLine.java),
which has a single set of limits, rather than `Branch`, which has one per side. `FlowsLimitsHolderBranchAdapter` in
[`MatpowerExporter`](https://github.com/powsybl/powsybl-core/blob/main/matpower/matpower-converter/src/main/java/com/powsybl/matpower/converter/MatpowerExporter.java)
already wraps one side of a branch into a single `FlowsLimitsHolder`, so the interface describes one set of
limits and not one terminal.

*2. A `DcNode`'s voltage limits are signed, and only their ordering is validated.*

A `DcNode`'s `v` is a signed potential, not an RMS magnitude: a node at negative polarity sits near
−500 kV and needs a band like [−520, −480].
[`ValidationUtil.checkVoltageLimits`](https://github.com/powsybl/powsybl-core/blob/main/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/ValidationUtil.java)
rejects limits below zero, so a new `checkDcVoltageLimits` validates only `low <= high`.

Signed rather than magnitude because of what a detector will do with it:
[`LimitViolationDetection.checkVoltage`](https://github.com/powsybl/powsybl-core/blob/main/security-analysis/security-analysis-api/src/main/java/com/powsybl/security/LimitViolationDetection.java)
compares `value <= lowVoltageLimit` and `value >= highVoltageLimit`. A DC detector copying those lines
works with signed limits; with magnitude limits `-500 <= 480` is true, so it would report `LOW_VOLTAGE` on
a healthy node.
[`VoltageAngleLimit`](https://github.com/powsybl/powsybl-core/blob/main/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/VoltageAngleLimit.java)
was used as a precedent for a signed low/high pair validated for ordering only.

*3. Violation detection is deliberately out of scope.*

Limits are reachable through `OperationalLimitsGroup` / `LoadingLimits`, so adding detection later is a
matter of new cases. `LimitViolationType` already has `CURRENT`, `LOW_VOLTAGE` and `HIGH_VOLTAGE`, so it
likely needs no new entries.

- [x] Yes
- [ ] No

Action is needed only if you **implement** `DcLine`, `DcNode` or `DcNodeAdder` yourself, as
[powsybl-network-store](https://github.com/powsybl/powsybl-network-store) does. This is tracked in
powsybl/powsybl-network-store#675. Those implementations will stop compiling, because the interfaces
gain abstract methods:

- **`DcLine`** now extends `FlowsLimitsHolder`: 17 methods. Hold one `OperationalLimitsGroupsImpl` and
  delegate to it, as
  [`BoundaryLineImpl`](https://github.com/powsybl/powsybl-core/blob/main/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/BoundaryLineImpl.java)
  does.
- **`DcNode`**: `getLowVoltageLimit`, `setLowVoltageLimit`, `getHighVoltageLimit`, `setHighVoltageLimit`.
- **`DcNodeAdder`**: `setLowVoltageLimit`, `setHighVoltageLimit`.

`DcLineImpl` and `DcNodeImpl` in this repository are updated accordingly and can be used as the reference.

Two further points, neither of which affects code that only *uses* these interfaces:

- **A `DcLine` now refuses active and apparent power limits.** Both fail with a `ValidationException`, and
  a file carrying one is rejected on import. The rule is asserted in the TCK, so an implementation of
  `DcLine` has to enforce it as well: hold the group collection with the two kinds marked unsupported, as
  `DcLineImpl` does.
- **Exporting `AcDcConverter.minP`/`maxP` to IIDM 1.17 or earlier now fails.** Until now these values had
  no serialization at all, and 7.3.0 shipped [#3954](https://github.com/powsybl/powsybl-core/issues/3954) let a network carrying them be exported by ignoring the values, either with the `iidm.export.xml.force-export-network-with-beta-features` config parameter or `ExportOptions.setForceExportNetworkWithBetaFeatures`. That escape pattern is removed: from 1.18 the values are written properly, and exporting them to an older version follows the normal `IidmVersionIncompatibilityBehavior`, so it raises an error by default or logs and drops them on request. Anyone on 7.3.0 who was relying on the force flag to export converters with `minP`/`maxP` to 1.17 should target 1.18 or set the incompatibility behaviour to `LOG_ERROR`.

**Other information**:

- The two sentinel conventions differ on purpose. `AcDcConverter.minP`/`maxP` use `∓Double.MAX_VALUE`, per
  the reasoning in [#3954](https://github.com/powsybl/powsybl-core/issues/3954) (*"ideally we would use
  -inf and inf for default values but this pattern is not used elsewhere. Non-finite real values are also
  unsupported in JSON."*), while the `DcNode` limits follow `VoltageLevel` and use `NaN`. This is why the
  two serializations use different version-gating helpers.
- One change reaches AC code: `ConnectableSerDeUtil.readAllSelectedGroupIds` is widened from
  `BoundaryLine` to `FlowsLimitsHolder` plus an id, and its single `BoundaryLineSerDe` call site is
  updated.
- Writing the new data to a pre-1.18 file raises an error, or logs and drops it, per
  `IidmVersionIncompatibilityBehavior`. Each piece has a test asserting that, and that an element
  *without* the new data still exports cleanly at those versions.

### DIFF
--- a/docs/grid_model/network_subnetwork.md
+++ b/docs/grid_model/network_subnetwork.md
@@ -916,12 +916,19 @@ DC nodes are points where DC terminals of DC conducting equipment are connected 
 
 **Characteristics**
 
-| Attribute  | Unit | Description                          |
-|------------|------|--------------------------------------|
-| $NominalV$ | kV   | The nominal voltage, always positive |
+| Attribute           | Unit | Description                          |
+|---------------------|------|--------------------------------------|
+| $NominalV$          | kV   | The nominal voltage, always positive |
+| $LowVoltageLimit$   | kV   | The low voltage limit, optional      |
+| $HighVoltageLimit$  | kV   | The high voltage limit, optional     |
 
 Although the nominal voltage of DC nodes must always be specified as a positive value,
 the solved voltages can be negative - for example, in the case of an LCC monopole operating in reverse polarity.
+
+The voltage limits describe the operating band of the DC node. Like the solved voltage, they are signed: they bound
+the voltage itself and not its magnitude, so both limits are negative on a DC node operating at negative polarity,
+and a DC node that may operate at either polarity has a band spanning both signs. Only the ordering of the two
+limits is checked: $LowVoltageLimit$ must not be greater than $HighVoltageLimit$.
 
 **Available extensions**
 - [Dynamic Model Info](extensions.md#dynamic-model-info)
--- a/docs/grid_model/network_subnetwork.md
+++ b/docs/grid_model/network_subnetwork.md
@@ -947,6 +947,12 @@ A DC Line has two DC Terminals.
 |-----------|----------|----------------------------------------|
 | $R$       | $\Omega$ | The series resistance, always positive |
 
+**Metadata**
+
+- A DC Line can have [loading limits](./additional.md#loading-limits), in a single collection for the whole line
+rather than one per side: a DC Line has no shunt admittance, so the current entering one DC Terminal is the current
+leaving the other. Current limits are the meaningful kind here; apparent power limits have no meaning on DC.
+
 **Available extensions**
 - [Dynamic Model Info](extensions.md#dynamic-model-info)
 
--- a/docs/grid_model/network_subnetwork.md
+++ b/docs/grid_model/network_subnetwork.md
@@ -949,9 +949,10 @@ A DC Line has two DC Terminals.
 
 **Metadata**
 
-- A DC Line can have [loading limits](./additional.md#loading-limits), in a single collection for the whole line
-rather than one per side: a DC Line has no shunt admittance, so the current entering one DC Terminal is the current
-leaving the other. Current limits are the meaningful kind here; apparent power limits have no meaning on DC.
+- A DC Line can have [loading limits](./additional.md#loading-limits), in which case they are stored in a single
+collection for the whole line rather than one per side: a DC Line has no shunt admittance, so the current entering
+one DC Terminal is the current leaving the other. Only current limits are accepted. Active and apparent power limits
+are refused, because a DC Line carries no reactive power and its two DC Terminals carry different active power.
 
 **Available extensions**
 - [Dynamic Model Info](extensions.md#dynamic-model-info)
--- a/docs/grid_model/network_subnetwork.md
+++ b/docs/grid_model/network_subnetwork.md
@@ -1066,7 +1066,8 @@ same load sign convention as `TargetP`.
 
 `MinP` must be less than or equal to `MaxP`. Both attributes are optional; if not set, the converter is considered with unlimited active power capability.
 
-`MinP` and `MaxP` are not serializable as of today. Trying to serialize AC-DC converters with non-default values of `MinP` or `MaxP` will raise an error.
+`MinP` and `MaxP` are serialized from IIDM version 1.18 onwards, and are omitted from the serialized file when left unset.
+Exporting a converter with a non-default `MinP` or `MaxP` to an earlier IIDM version raises an error.
 
 **Available extensions**
 - [Dynamic Model Info](extensions.md#dynamic-model-info)
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcLine.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcLine.java
@@ -20,13 +20,14 @@ package com.powsybl.iidm.network;
  * than one collection per side as a {@link Branch} does. A DC Line has no shunt admittance, so the current
  * entering one DC Terminal is the current leaving the other, and one limit describes both.
  *
- * <p>{@link CurrentLimits} is the meaningful limit on a DC Line: it is the rating of the conductor, and it
- * is unambiguous because both DC Terminals carry the same current.
+ * <p>{@link CurrentLimits} is the only kind of limit a DC Line accepts: it is the rating of the conductor,
+ * and it is unambiguous because both DC Terminals carry the same current.
  *
- * <p>The other two kinds come with {@link FlowsLimitsHolder} and should not be relied on here.
- * {@link ApparentPowerLimits} has no meaning on DC. An {@link ActivePowerLimits} is ambiguous: the two DC
- * Terminals carry different active power, since the series resistance consumes some of it, and a single
- * limit does not say which of the two it bounds.
+ * <p>The other two kinds come with {@link FlowsLimitsHolder}, but a DC Line refuses them, because neither
+ * describes anything physical on DC. An {@link ApparentPowerLimits} has no meaning where there is no
+ * reactive power. An {@link ActivePowerLimits} would not say which DC Terminal it bounds, since the series
+ * resistance makes the two carry different active power. Setting either one fails with a
+ * {@link ValidationException}, and so does importing a network that contains one.
  *
  * <p>
  *  Characteristics
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcLine.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcLine.java
@@ -13,6 +13,22 @@ package com.powsybl.iidm.network;
  * <p> To create a DcLine, see {@link DcLineAdder}
  *
  * <p>
+ *  Operational limits
+ * </p>
+ *
+ * <p>A DC Line carries a single collection of {@link OperationalLimitsGroup} for the whole line, rather
+ * than one collection per side as a {@link Branch} does. A DC Line has no shunt admittance, so the current
+ * entering one DC Terminal is the current leaving the other, and one limit describes both.
+ *
+ * <p>{@link CurrentLimits} is the meaningful limit on a DC Line: it is the rating of the conductor, and it
+ * is unambiguous because both DC Terminals carry the same current.
+ *
+ * <p>The other two kinds come with {@link FlowsLimitsHolder} and should not be relied on here.
+ * {@link ApparentPowerLimits} has no meaning on DC. An {@link ActivePowerLimits} is ambiguous: the two DC
+ * Terminals carry different active power, since the series resistance consumes some of it, and a single
+ * limit does not say which of the two it bounds.
+ *
+ * <p>
  *  Characteristics
  * </p>
  *
@@ -57,7 +73,7 @@ package com.powsybl.iidm.network;
  *
  * @author Damien Jeandemange {@literal <damien.jeandemange at artelys.com>}
  */
-public interface DcLine extends DcConnectable<DcLine> {
+public interface DcLine extends DcConnectable<DcLine>, FlowsLimitsHolder {
 
     @Override
     default IdentifiableType getType() {
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcNode.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcNode.java
@@ -52,8 +52,30 @@ package com.powsybl.iidm.network;
  *             <td style="border: 1px solid black"> - </td>
  *             <td style="border: 1px solid black">The nominal voltage, always positive</td>
  *         </tr>
+ *         <tr>
+ *             <td style="border: 1px solid black">LowVoltageLimit</td>
+ *             <td style="border: 1px solid black">double</td>
+ *             <td style="border: 1px solid black">kV</td>
+ *             <td style="border: 1px solid black">no</td>
+ *             <td style="border: 1px solid black"> - </td>
+ *             <td style="border: 1px solid black">The low voltage limit, may be negative</td>
+ *         </tr>
+ *         <tr>
+ *             <td style="border: 1px solid black">HighVoltageLimit</td>
+ *             <td style="border: 1px solid black">double</td>
+ *             <td style="border: 1px solid black">kV</td>
+ *             <td style="border: 1px solid black">no</td>
+ *             <td style="border: 1px solid black"> - </td>
+ *             <td style="border: 1px solid black">The high voltage limit, may be negative</td>
+ *         </tr>
  *     </tbody>
  * </table>
+ *
+ * <p>
+ * The voltage limits bound the <b>signed</b> voltage {@link #getV()}, not its magnitude, so that the operating band
+ * of a DC node on the negative pole can be expressed. On such a node both limits are negative and crossing the low
+ * voltage limit means that the voltage <i>magnitude</i> is too large.
+ * </p>
  *
  * @author Damien Jeandemange {@literal <damien.jeandemange at artelys.com>}
  */
@@ -74,6 +96,36 @@ public interface DcNode extends Identifiable<DcNode>, DcTopologyVisitable {
      * @return self for method chaining
      */
     DcNode setNominalV(double nominalV);
+
+    /**
+     * Get the low voltage limit in kV. The limit bounds the signed voltage {@link #getV()} and may be negative.
+     *
+     * @return the low voltage limit or NaN if undefined
+     */
+    double getLowVoltageLimit();
+
+    /**
+     * Set the low voltage limit in kV. The limit bounds the signed voltage {@link #getV()} and may be negative.
+     *
+     * @param lowVoltageLimit new low voltage limit in kV
+     * @return self for method chaining
+     */
+    DcNode setLowVoltageLimit(double lowVoltageLimit);
+
+    /**
+     * Get the high voltage limit in kV. The limit bounds the signed voltage {@link #getV()} and may be negative.
+     *
+     * @return the high voltage limit or NaN if undefined
+     */
+    double getHighVoltageLimit();
+
+    /**
+     * Set the high voltage limit in kV. The limit bounds the signed voltage {@link #getV()} and may be negative.
+     *
+     * @param highVoltageLimit new high voltage limit in kV
+     * @return self for method chaining
+     */
+    DcNode setHighVoltageLimit(double highVoltageLimit);
 
     /**
      * Get the voltage of the DC node in kV.
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcNode.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcNode.java
@@ -72,9 +72,10 @@ package com.powsybl.iidm.network;
  * </table>
  *
  * <p>
- * The voltage limits bound the <b>signed</b> voltage {@link #getV()}, not its magnitude, so that the operating band
- * of a DC node on the negative pole can be expressed. On such a node both limits are negative and crossing the low
- * voltage limit means that the voltage <i>magnitude</i> is too large.
+ * The voltage limits bound the <b>signed</b> voltage {@link #getV()}, not its magnitude. The rule is always
+ * {@code LowVoltageLimit <= V <= HighVoltageLimit}: a DC Node at a positive potential has a band such as
+ * {@code [480, 520]}, and one at a negative potential has {@code [-520, -480]}, with both limits negative. On the
+ * latter, crossing the low voltage limit means the voltage <i>magnitude</i> is too large.
  * </p>
  *
  * @author Damien Jeandemange {@literal <damien.jeandemange at artelys.com>}
--- a/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcNode.java
+++ b/iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcNode.java
@@ -72,10 +72,38 @@ package com.powsybl.iidm.network;
  * </table>
  *
  * <p>
- * The voltage limits bound the <b>signed</b> voltage {@link #getV()}, not its magnitude. The rule is always
- * {@code LowVoltageLimit <= V <= HighVoltageLimit}: a DC Node at a positive potential has a band such as
- * {@code [480, 520]}, and one at a negative potential has {@code [-520, -480]}, with both limits negative. On the
- * latter, crossing the low voltage limit means the voltage <i>magnitude</i> is too large.
+ * These limits apply to the signed value returned by {@link #getV()}, not to how big that number is once you
+ * ignore its sign. The rule is always {@code LowVoltageLimit <= V <= HighVoltageLimit}. This matters because
+ * the same check already written for {@link VoltageLevel} can be reused as is: a healthy DC Node on the
+ * negative side reads something like {@code -500}. If the limits were stored ignoring the sign, as
+ * {@code [480, 520]}, that check would see {@code -500 <= 480} and wrongly say the voltage is too low. Stored
+ * signed instead, as {@code [-520, -480]}, the same check sees {@code -500} sits between them and reports
+ * nothing, which is correct.
+ * </p>
+ *
+ * <p>
+ * A DC Node on the positive side has a band such as {@code [480, 520]}. One on the negative side has a band
+ * such as {@code [-520, -480]}. On the negative side, going below {@code -520} (the low limit) means the
+ * voltage is too big once you ignore the sign, and going above {@code -480} (the high limit) means it is too
+ * small: the opposite of the positive side. So "low" always means the smaller number and "high" always means
+ * the bigger number, but which one is "too big" and which one is "too little" flips depending on which side of
+ * zero the band is on.
+ * </p>
+ *
+ * <p>
+ * A band can even sit around zero, such as {@code [-10, 10]}. There, {@code V = -15} crosses the low limit and
+ * {@code V = 15} crosses the high limit, but both are the same problem: the voltage moved too far from zero, in
+ * one direction or the other. Neither crossing means "too little", because every value close to zero is inside
+ * the band and allowed. So code that wants to report "voltage too high" or "voltage too low" cannot decide it
+ * from which field, low or high, was crossed. It first has to check where zero falls relative to the band.
+ * </p>
+ *
+ * <p>
+ * There are too many different bands for one meaning of "low" and "high" to fit them all: a positive-side node,
+ * a negative-side node, and a node centered on zero each need a different answer. So this class does not try
+ * to pick one. It only checks that {@code LowVoltageLimit <= HighVoltageLimit}
+ * ({@link ValidationUtil#checkDcVoltageLimits}), the one rule that holds in every case, and leaves it to
+ * whoever reads {@code V} and the band to decide what "too high" or "too low" means for that particular node.
  * </p>
  *
  * @author Damien Jeandemange {@literal <damien.jeandemange at artelys.com>}
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcLineImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcLineImpl.java
@@ -104,6 +104,7 @@ public class DcLineImpl extends AbstractDcConnectable<DcLine> implements DcLine 
         return this;
     }
 
+    // A removed DcLine's limits cannot be accessed, like r above and like DcNodeImpl does for its voltage limits.
     private void checkAccess() {
         ValidationUtil.checkAccessOfRemovedEquipment(this.id, this.removed, LIMITS_ATTRIBUTE);
     }
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcLineImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcLineImpl.java
@@ -13,12 +13,14 @@ import com.powsybl.iidm.network.ApparentPowerLimitsAdder;
 import com.powsybl.iidm.network.CurrentLimitsAdder;
 import com.powsybl.iidm.network.DcLine;
 import com.powsybl.iidm.network.DcTerminal;
+import com.powsybl.iidm.network.LimitType;
 import com.powsybl.iidm.network.OperationalLimitsGroup;
 import com.powsybl.iidm.network.TwoSides;
 import com.powsybl.iidm.network.ValidationUtil;
 
 import java.util.Collection;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 
@@ -31,6 +33,10 @@ public class DcLineImpl extends AbstractDcConnectable<DcLine> implements DcLine 
 
     private static final String LIMITS_ATTRIBUTE = "limits";
 
+    private static final Map<LimitType, String> UNSUPPORTED_LIMITS = Map.of(
+            LimitType.ACTIVE_POWER, "active power limits are not supported: the two DC terminals carry different active power",
+            LimitType.APPARENT_POWER, "apparent power limits are not supported: a DC line carries no reactive power");
+
     private double r;
 
     private final OperationalLimitsGroupsImpl operationalLimitsGroups;
@@ -38,7 +44,7 @@ public class DcLineImpl extends AbstractDcConnectable<DcLine> implements DcLine 
     DcLineImpl(Ref<NetworkImpl> ref, Ref<SubnetworkImpl> subnetworkRef, String id, String name, boolean fictitious, double r) {
         super(ref, subnetworkRef, id, name, fictitious);
         this.r = r;
-        this.operationalLimitsGroups = new OperationalLimitsGroupsImpl(this, LIMITS_ATTRIBUTE);
+        this.operationalLimitsGroups = new OperationalLimitsGroupsImpl(this, LIMITS_ATTRIBUTE, UNSUPPORTED_LIMITS);
     }
 
     @Override
@@ -212,6 +218,8 @@ public class DcLineImpl extends AbstractDcConnectable<DcLine> implements DcLine 
     @Override
     public ActivePowerLimitsAdder newActivePowerLimits() {
         checkModification();
+        // before getOrCreate, or a refusal leaves an empty group behind
+        operationalLimitsGroups.checkSupported(LimitType.ACTIVE_POWER);
         return operationalLimitsGroups.getOrCreateSelectedOperationalLimitsGroup().newActivePowerLimits();
     }
 
@@ -222,6 +230,7 @@ public class DcLineImpl extends AbstractDcConnectable<DcLine> implements DcLine 
     @Override
     public ApparentPowerLimitsAdder newApparentPowerLimits() {
         checkModification();
+        operationalLimitsGroups.checkSupported(LimitType.APPARENT_POWER);
         return operationalLimitsGroups.getOrCreateSelectedOperationalLimitsGroup().newApparentPowerLimits();
     }
 }
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcLineImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcLineImpl.java
@@ -104,11 +104,6 @@ public class DcLineImpl extends AbstractDcConnectable<DcLine> implements DcLine 
         return this;
     }
 
-    /**
-     * A removed DC equipment must refuse access, like the {@code r} and terminal accessors above and like
-     * {@code DcNodeImpl} does for its voltage limits. {@code BoundaryLineImpl}, whose delegations this class
-     * otherwise copies, has no such rule and delegates bare; ours add one of these two calls.
-     */
     private void checkAccess() {
         ValidationUtil.checkAccessOfRemovedEquipment(this.id, this.removed, LIMITS_ATTRIBUTE);
     }
--- a/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcLineImpl.java
+++ b/iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcLineImpl.java
@@ -8,12 +8,19 @@
 package com.powsybl.iidm.network.impl;
 
 import com.powsybl.commons.ref.Ref;
+import com.powsybl.iidm.network.ActivePowerLimitsAdder;
+import com.powsybl.iidm.network.ApparentPowerLimitsAdder;
+import com.powsybl.iidm.network.CurrentLimitsAdder;
 import com.powsybl.iidm.network.DcLine;
 import com.powsybl.iidm.network.DcTerminal;
+import com.powsybl.iidm.network.OperationalLimitsGroup;
 import com.powsybl.iidm.network.TwoSides;
 import com.powsybl.iidm.network.ValidationUtil;
 
+import java.util.Collection;
+import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 
 /**
  * @author Damien Jeandemange {@literal <damien.jeandemange at artelys.com>}
@@ -22,11 +29,16 @@ public class DcLineImpl extends AbstractDcConnectable<DcLine> implements DcLine 
 
     public static final String R_ATTRIBUTE = "r";
 
+    private static final String LIMITS_ATTRIBUTE = "limits";
+
     private double r;
+
+    private final OperationalLimitsGroupsImpl operationalLimitsGroups;
 
     DcLineImpl(Ref<NetworkImpl> ref, Ref<SubnetworkImpl> subnetworkRef, String id, String name, boolean fictitious, double r) {
         super(ref, subnetworkRef, id, name, fictitious);
         this.r = r;
+        this.operationalLimitsGroups = new OperationalLimitsGroupsImpl(this, LIMITS_ATTRIBUTE);
     }
 
     @Override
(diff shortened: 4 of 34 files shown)


docs/grid_model/network_subnetwork.md
```diff
@@ -940,6 +947,12 @@ A DC Line has two DC Terminals.
 |-----------|----------|----------------------------------------|
 | $R$       | $\Omega$ | The series resistance, always positive |
 
+**Metadata**
+
+- A DC Line can have [loading limits](./additional.md#loading-limits), in a single collection for the whole line
+rather than one per side: a DC Line has no shunt admittance, so the current entering one DC Terminal is the current
+leaving the other. Current limits are the meaningful kind here; apparent power limits have no meaning on DC.
```
NathanDissoubray: it's not very clear what you mean by "the meaningful kind here". People could wonder "so there are limits that are not useful" ?
I think you could phrase it as: "only limits of kind X are used for DC lines, as limits of kind Y do no represent anything physically correct", or another similar formulation

docs/grid_model/network_subnetwork.md
```diff
@@ -940,6 +947,12 @@ A DC Line has two DC Terminals.
 |-----------|----------|----------------------------------------|
 | $R$       | $\Omega$ | The series resistance, always positive |
 
+**Metadata**
+
+- A DC Line can have [loading limits](./additional.md#loading-limits), in a single collection for the whole line
```
NathanDissoubray: this could be read as "it is possible to have a single collection of limits for the entire line", when what we should understand is "limits are always stored in a single collection for the whole line". You can rephrase like that:
"A DC line can have loading limits, in which case those limits are stored in a single collection for the whole line"

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcLine.java
```diff
@@ -13,6 +13,22 @@
...
  * <p>
+ *  Operational limits
+ * </p>
+ *
+ * <p>A DC Line carries a single collection of {@link OperationalLimitsGroup} for the whole line, rather
+ * than one collection per side as a {@link Branch} does. A DC Line has no shunt admittance, so the current
+ * entering one DC Terminal is the current leaving the other, and one limit describes both.
+ *
+ * <p>{@link CurrentLimits} is the meaningful limit on a DC Line: it is the rating of the conductor, and it
+ * is unambiguous because both DC Terminals carry the same current.
```
NathanDissoubray: "meaningful" seems a bit strange to me. Do you mean to say that a current limit is the only kind of limit that makes sense to use for a DC line ?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcLine.java
```diff
@@ -13,6 +13,22 @@
...
+ *
+ * <p>The other two kinds come with {@link FlowsLimitsHolder} and should not be relied on here.
+ * {@link ApparentPowerLimits} has no meaning on DC. An {@link ActivePowerLimits} is ambiguous: the two DC
+ * Terminals carry different active power, since the series resistance consumes some of it, and a single
+ * limit does not say which of the two it bounds.
```
NathanDissoubray: this should be enforced by code when trying to add an `ApparentPowerLimits` or an `ActivePowerLimits` to a `DcLine` rather than being only a javadoc comment.

I'm also wondering if it makes sense to extend from `FlowsLimitHolder` if two thirds of the functions provided by it cannot be used, would it make more sense to separate the `CurrentLimits` into a `CurrentLimitsHolder` so that DC components extend only from that ? And `FlowsLimitsHolder` would be something that extends from the `CurrentLimitsHolder` and adds active and reactive limits.

That way, we don't need to forbid usage of other kinds of limits in DC components, since by construction of the interface, only current limits could be added to it.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcNode.java
```diff
@@ -52,9 +52,31 @@
...
+ *         </tr>
  *     </tbody>
  * </table>
  *
+ * <p>
+ * The voltage limits bound the <b>signed</b> voltage {@link #getV()}, not its magnitude, so that the operating band
+ * of a DC node on the negative pole can be expressed. On such a node both limits are negative and crossing the low
+ * voltage limit means that the voltage <i>magnitude</i> is too large.
+ * </p>
+ *
```
NathanDissoubray, line 108: so if I understand correctly, you have:
-|maxV| < V < -|minV|
or
|minV| < V < |maxV|
?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcLineImpl.java
```diff
@@ -85,4 +97,131 @@ public DcLine setR(double r) {
         getNetwork().getListeners().notifyUpdate(this, R_ATTRIBUTE, oldValue, r);
         return this;
     }
+
+    /**
+     * A removed DC equipment must refuse access, like the {@code r} and terminal accessors above and like
+     * {@code DcNodeImpl} does for its voltage limits. {@code BoundaryLineImpl}, whose delegations this class
+     * otherwise copies, has no such rule and delegates bare; ours add one of these two calls.
```
NathanDissoubray: that last sentence is a bit difficult to read, could you clarify please ?

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractAcDcConverterSerDe.java
```diff
@@ -80,6 +80,14 @@ protected void readRootElementCommonAttributes(final A adder, final VoltageLevel
...
         double targetP = context.getReader().readDoubleAttribute("targetP");
         double targetVdc = context.getReader().readDoubleAttribute("targetVdc");
+
+        // unbounded active power as default value for IIDM version < 1.18 and for converters declaring no limit
+        double[] minP = {-Double.MAX_VALUE};
+        double[] maxP = {Double.MAX_VALUE};
+        IidmSerDeUtil.runFromMinimumVersion(IidmVersion.V_1_18, context, () -> {
+            minP[0] = context.getReader().readDoubleAttribute(MIN_P, -Double.MAX_VALUE);
+            maxP[0] = context.getReader().readDoubleAttribute(MAX_P, Double.MAX_VALUE);
+        });
```
NathanDissoubray: if the value minP/maxP doesn't exist (so for IIDM version 1.17 and below), `readDoubleAttribute` would return the default value of +-`Double.MAX_VALUE`
you could do something like that:
```java
double minP = context.getReader().readDoubleAttribute(MIN_P, -Double.MAX_VALUE);
double maxP = context.getReader().readDoubleAttribute(MAX_P, Double.MAX_VALUE);
```
though I guess if someone edits an old IIDM files to include a non default value and we read it, that would lead to a crash, so it's safer this way against IIDM files that do not match the scheme of their version 🤔 what do you think ?

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractAcDcConverterSerDe.java
```diff
@@ -41,16 +44,13 @@ protected void writeRootElementAttributes(final T converter, final VoltageLevel
...
-                "To force the export of the network and ignore this value, either use the config parameter iidm.export.xml.force-export-network-with-beta-features, " +
-                "or ExportOptions.setForceExportNetworkWithBetaFeatures");
-        }
+
+        IidmSerDeUtil.assertMinimumVersionAndRunIfNotDefault(converter.getMinP() != -Double.MAX_VALUE, getRootElementName(), MIN_P,
+                IidmSerDeUtil.ErrorMessage.NOT_DEFAULT_NOT_SUPPORTED, IidmVersion.V_1_18, context,
+                () -> context.getWriter().writeDoubleAttribute(MIN_P, converter.getMinP()));
+        IidmSerDeUtil.assertMinimumVersionAndRunIfNotDefault(converter.getMaxP() != Double.MAX_VALUE, getRootElementName(), MAX_P,
+                IidmSerDeUtil.ErrorMessage.NOT_DEFAULT_NOT_SUPPORTED, IidmVersion.V_1_18, context,
+                () -> context.getWriter().writeDoubleAttribute(MAX_P, converter.getMaxP()));
```
NathanDissoubray, line 53: the code to allow export of a network by ignoring the serde of maxP / minP is in the release 7.3.0 of powsybl (and version 1.17 of IIDM).
If somebody updates to the version of powsybl that includes this code, they won't be able to export to IIDM 1.17 if their network has limits.
In itself, it's not too much of an issue since it's a beta feature (and as thus, it was advertised as breakable), but this should be reported in the migration guide section of this PR

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/DcLineSerDeTest.java
```diff
@@ -45,6 +50,55 @@ void testNotSupported() throws IOException {
...
+        Network network = createBaseNetwork();
+
+        ExportOptions options = new ExportOptions().setOnlySelectedOperationalLimitsGroups(true);
+        Path path = tmpDir.resolve("onlySelectedGroups.xml");
+        NetworkSerDe.write(network, options, path);
+        DcLine dcLine = NetworkSerDe.read(path).getDcLine("dcLineWithSolvedV");
+
+        assertEquals(1, dcLine.getOperationalLimitsGroups().size());
+        assertEquals("summer", dcLine.getSelectedOperationalLimitsGroupId().orElseThrow());
+        assertEquals(2106.0, dcLine.getCurrentLimits().orElseThrow().getPermanentLimit());
```
NathanDissoubray: you seem to be doing a round trip test. We have functions to do that in `AbstractIidmSerDeTest`, which allows you to test for all versions since a given version (1.18 in this case) and for all the file formats (json and binary too).
In particular, you should take a look at `allFormatsRoundTripFromVersionedXmlFromMinVersionTest`.

If you want to check that the result is correct, use `testForAllVersionsBetween` coupled with `allFormatsRoundTripTest` which will allow you to check that everything is correct for each network that you read (so you can check that the group is called "summer" and the value is correct)

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/DcLineSerDeTest.java
```diff
@@ -45,6 +50,55 @@ void testNotSupported() throws IOException {
...
+    @Test
+    void testLimitsNotSupported() {
+        Network network = createBaseNetwork();
+
+        // versions 1.15 to 1.17 support the DC line but not its operational limits, so exporting a DC line
+        // that declares limits must fail rather than silently drop them
+        testForAllPreviousVersions(IidmVersion.V_1_18, version -> {
+            if (version.compareTo(IidmVersion.V_1_15) < 0) {
+                return;
+            }
```
NathanDissoubray: You could use `testForAllVersionsBetween(minVersion, maxVersion, test)` instead

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/DcLineSerDeTest.java
```diff
@@ -45,6 +50,55 @@ void testNotSupported() throws IOException {
...
+        assertDoesNotThrow(() -> NetworkSerDe.write(noLimits, options, tmpDir.resolve("noLimits.xml")));
+
+        // exporting only the selected groups leaves nothing to write when none is selected, so the
+        // version error must not fire either
+        network.getDcLine("dcLineWithSolvedV").cancelSelectedOperationalLimitsGroup();
+        ExportOptions selectedOnly = new ExportOptions()
+                .setVersion(IidmVersion.V_1_17.toString("."))
+                .setOnlySelectedOperationalLimitsGroups(true);
+        assertDoesNotThrow(() -> NetworkSerDe.write(network, selectedOnly, tmpDir.resolve("noSelected.xml")));
+    }
```
NathanDissoubray: this should be in a separate test, since here you are checking that the export succeeds, in a test called "limitsNotSupported" where just above we are expecting the export to fail

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/DcLineSerDeTest.java
```diff
@@ -79,6 +133,20 @@ private static Network createBaseNetwork() {
...
+                .setName("20'")
+                .setAcceptableDuration(20 * 60)
+                .setValue(2350.0)
+                .endTemporaryLimit()
+                .add();
+        dcLine2.newOperationalLimitsGroup("winter")
+                .newCurrentLimits()
+                .setPermanentLimit(2400.0)
+                .add();
+        dcLine2.setSelectedOperationalLimitsGroup("summer");
```
NathanDissoubray, line 188: you should add more than one temporary limit to your use case to check that all are exported. It might also be good to check that everything works as expected when multiple limit are selected, in a separate test (using `addSelectedOperationalLimitsGroup` instead of `set`

NathanDissoubray (changes requested):
Thank you for the extensive doc, it helped understand what you were doing.

docs/grid_model/network_subnetwork.md
thomas-durand-SGI: Rephrased. It also needed updating rather than just rewording: the paragraph mentioned only apparent power, while active power limits are now refused as well. It now says only current limits are accepted, with the reason for each of the other two.

docs/grid_model/network_subnetwork.md
thomas-durand-SGI: Fixed, it now reads "in which case they are stored in a single collection for the whole line".

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcLine.java
thomas-durand-SGI: Reworded and changed the code. Current limits are now described as the only kind a DC Line accepts, and the other two as refused.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcLine.java
thomas-durand-SGI: Understood.
I asked myself this same question. The initial reason I extended `FlowsLimitsHolder `is to inherit the loading limits (which it deserves, as a DC line is a line with loading limits as well): `FlowsLimitsHolder `is what gives access to an `OperationalLimitsGroup`, which is where seasonal groups and temporary limits actually live. The tradeoff is that it comes with the unwanted apparent and active power limits.
Separating `CurrentLimits `into a `CurrentLimitsHolder `wouldn't stop a DC line from getting an apparent or active power limit, because the limits don't live on the holder, they live in the `OperationalLimitsGroup `you get back from `newOperationalLimitsGroup(String)`, and that group always has all three slots: current, active power, apparent power. A `CurrentLimitsHolder `still has to hand out an `OperationalLimitsGroup `with all three slots, or a DC line loses its seasonal groups and temporary limits.

 I checked #3340, which deprecated `newCurrentLimits()`, `newActivePowerLimits()` and `newApparentPowerLimits()` on the holder in favour of the equivalents on `OperationalLimitsGroup`, so the group is now the only supported place to create a limit from. So with a current-only holder, this still compiles:

```java
dcLine.newOperationalLimitsGroup("summer").newApparentPowerLimits().setPermanentLimit(1000.).add();
```

The way I see it there are two solutions:

1. Refuse these limits by throwing a `ValidationException` (the one iidm already uses for invalid equipment state, so the message comes out prefixed with the equipment) when an active or apparent power limit is set. It is rejected on import. Concretely:

- `newActivePowerLimits()` and `newApparentPowerLimits()` fail on a DC line's group, and so do the deprecated ones on the DcLine itself:

```
DC Line 'dcLine': active power limits are not supported: the two DC terminals carry different active power
DC Line 'dcLine': apparent power limits are not supported: a DC line carries no reactive power
```

- a refused call leaves nothing behind, the group it would have created is not created
- current limits are untouched: seasonal groups and temporary limits work exactly as before
- the test is in the TCK, so the rule holds for every implementation and not only iidm-impl

I have implemented this. It is in 45e60db731, "Reject apparent and active power limits on a DcLine".

The tradeoff is that the getters stay on the API and just return empty, and the 1.18 XSD still declares `dcLine` with the shared `OperationalLimitsGroup` type, so a file carrying an active or apparent power limit on a `dcLine` is schema valid and only the reader rejects it. Do you think it is worth giving `dcLine` its own group type in the 1.18 XSD, one that only allows `currentLimits`, instead of the shared `OperationalLimitsGroup` type? Is 1.18 still open for this change?

2. The second option is to make it impossible to even try, not just refused. That needs two changes: split `OperationalLimitsGroup` so a DC line's group has only a current slot, no active or apparent power slot at all, and split `FlowsLimitsHolder` so `DcLine` doesn't expose `getApparentPowerLimits()` and the rest.

The `FlowsLimitsHolder` split alone changes nothing, that's the one I already ruled out above, since the group would still have all three slots. The `OperationalLimitsGroup` split is the one that actually removes a slot, and it's the expensive one: it touches the read and write paths in `ConnectableSerDeUtil`, the XSD, CGMES, Matpower, and every implementor of `OperationalLimitsGroup`, including powsybl-network-store, for whom it's a breaking change. It also means editing AC code, which I'd rather not do here unless you want it.
This is why I opted for option 1 in this PR review. However, this is an open subject and I more than happy to implement option 2 if you prefer.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcNode.java
thomas-durand-SGI: There is one single rule and it covers both cases: `LowVoltageLimit <= V <= HighVoltageLimit`, applied to the signed voltage.
So a DC Node at a positive potential has a band like` [480, 520`], and one at a negative potential has` [-520, -480`]. On the negative one, crossing the low voltage limit means` V < -520`, in other words the magnitude is too large.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/DcLineSerDeTest.java
thomas-durand-SGI: Done. It now uses `allFormatsRoundTripTest `against a versioned reference, so JSON and binary are covered too, and the assertions on the group id and the value are kept. Fixed in b0df7bce95

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/DcLineSerDeTest.java
thomas-durand-SGI: Fixed, it is now `testForAllVersionsBetween(V_1_15, V_1_17, ...)` with no early return.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/DcLineSerDeTest.java
thomas-durand-SGI: Fixed, they are now their own test, `testExportToOlderVersionSucceedsWithNoLimitToDrop`.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/DcLineSerDeTest.java
thomas-durand-SGI: Fixed. The summer group now has two temporary limits, and there is a new `testMultipleSelectedGroups `using `addSelectedOperationalLimitsGroups`, round tripped against its own reference. Fixed in b0df7bce95

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractAcDcConverterSerDe.java
thomas-durand-SGI: I added it to the migration guide. Anyone on 7.3.0 who used` force-export-network-with-beta-features` to export a converter with `minP/maxP` to 1.17 will find that export path gone: the escape hatch is removed, `minP/maxP` are written properly from 1.18 on, and exporting to an older version with limits set now follows the normal `IidmVersionIncompatibilityBehavior`.

thomas-durand-SGI:
Hi @NathanDissoubray,
Thank you for your PR review comments. I have added two commits implementing the requested changes:
- 45e60db731 : Reject apparent and active power limits on a DcLine.
    - Raises `ValidationException ` for apparent and active power limits, as discussed in the reply
- b0df7bce95 : Address review comments 1 on DC limits
   - Contains all the other test/documentation change requested in this PR review

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcLine.java
NathanDissoubray: You are right, the different limits are stored in an `OperationalLimitsGroup`. `FlowsLimitHolder` allows to access those limits.
Regarding your first solution, you could also add a separate `DcOperationalLimitsGroup` (name might change) in the XSD. Even if we don't split the interface of `FlowsLimitHolder` and `OperationalLimitsGroup`, that can still prevent incorrect XML from being read.

I'll see if refactoring to split is not too big of a change, in which case it would be ok to make the change. I'll keep you informed.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/OperationalLimitsGroupImpl.java
```diff
@@ -49,12 +63,28 @@ public class OperationalLimitsGroupImpl extends AbstractPropertiesHolder impleme
...
+     * @param identifiable on what to create the group
+     * @param listeners the listeners on the changes of this group
+     * @param validable used for exception mechanism when validating the network, provides a label for the error
+     * @param attributeName prefix used for referencing the limits (used when notifying listeners)
+     * @param isSelected a predicate telling this group if it is selected or not
+     * @param unsupportedLimitTypes limit types this group refuses, each with the message to throw
+     */
+    public OperationalLimitsGroupImpl(String id, Identifiable<?> identifiable, NetworkListenerList listeners,
+                                      Validable validable, String attributeName, Predicate<String> isSelected,
+                                      Map<LimitType, String> unsupportedLimitTypes) {
```
NathanDissoubray, line 80: I'm not a fan of having the "unsupportedLimitTypes" like that, but that joins what we were discussing regarding splitting the interface of `FlowsLimitHolder` and `OperationalLimitsGroup` which would make this cleaner.

docs/grid_model/network_subnetwork.md
```diff
@@ -949,9 +949,10 @@ A DC Line has two DC Terminals.
 
 **Metadata**
 
-- A DC Line can have [loading limits](./additional.md#loading-limits), in a single collection for the whole line
-rather than one per side: a DC Line has no shunt admittance, so the current entering one DC Terminal is the current
-leaving the other. Current limits are the meaningful kind here; apparent power limits have no meaning on DC.
+- A DC Line can have [loading limits](./additional.md#loading-limits), in which case they are stored in a single
+collection for the whole line rather than one per side: a DC Line has no shunt admittance, so the current entering
+one DC Terminal is the current leaving the other. Only current limits are accepted. Active and apparent power limits
+are refused, because a DC Line carries no reactive power and its two DC Terminals carry different active power.
```
NathanDissoubray, line 955: I fail to understand how both terminals having different active power prevents having limits. On the contrary, it could justify having one active power limit per side (rather than one for the whole line). Why is it not allowed then ?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcNode.java
NathanDissoubray: you mention in that paragraph "The voltage limits bound the <b>signed</b> voltage {@link #getV()}, not its magnitude", then "crossing the low voltage limit means the voltage <i>magnitude</i> is too large". Isn't that contradictory ?

In the case `[480, 520]`, 520 is the high limit but in the case `[-520, -480]` -480 is the high limit ?

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/DcLineSerDeTest.java
```diff
@@ -51,19 +52,31 @@ void testNotSupported() throws IOException {
...
-    void testExportOnlySelectedGroups() {
+    void testExportOnlySelectedGroups() throws IOException {
         Network network = createBaseNetwork();
 
-        ExportOptions options = new ExportOptions().setOnlySelectedOperationalLimitsGroups(true);
-        Path path = tmpDir.resolve("onlySelectedGroups.xml");
-        NetworkSerDe.write(network, options, path);
-        DcLine dcLine = NetworkSerDe.read(path).getDcLine("dcLineWithSolvedV");
+        Network read = allFormatsRoundTripTest(network, "/dcLineOnlySelectedGroupsRef.xml", CURRENT_IIDM_VERSION,
+                new ExportOptions().setOnlySelectedOperationalLimitsGroups(true));
```
NathanDissoubray: this should test for all versions since IIDM 1.18 rather than only the current version. With only the current version, If we bump IIDM to 1.19, it means version 1.18 is not tested by this anymore and won't catch regressions.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/DcLineSerDeTest.java
```diff
@@ -51,19 +52,31 @@ void testNotSupported() throws IOException {
...
         assertEquals("summer", dcLine.getSelectedOperationalLimitsGroupId().orElseThrow());
         assertEquals(2106.0, dcLine.getCurrentLimits().orElseThrow().getPermanentLimit());
     }
 
+    @Test
+    void testMultipleSelectedGroups() throws IOException {
+        Network network = createBaseNetwork();
+        network.getDcLine("dcLineWithSolvedV").addSelectedOperationalLimitsGroups("winter");
+
+        Network read = allFormatsRoundTripTest(network, "/dcLineMultipleSelectedGroupsRef.xml", CURRENT_IIDM_VERSION);
```
NathanDissoubray: this should test for all versions since 1.18

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/DcLineSerDeTest.java
```diff
@@ -51,19 +52,31 @@ void testNotSupported() throws IOException {
...
+
+        DcLine dcLine = read.getDcLine("dcLineWithSolvedV");
+        assertEquals(List.of("summer", "winter"), dcLine.getAllSelectedOperationalLimitsGroupIdsOrdered());
+        assertEquals(2, dcLine.getAllSelectedCurrentLimits().size());
+        assertEquals(2400.0, dcLine.getCurrentLimits().orElseThrow().getPermanentLimit());
```
NathanDissoubray: this should check both limits. `getCurrentLimits` only gets the last selected limit, check that both selected limits are correct.
You can use something like that:
```
org.assertj.core.api.Assertions.assertThat(List.of())
            .hasSize()
            .extracting(//Extract the values you want to test)
                .containsExactlyInAnyOrder(//what you expect)
```

NathanDissoubray (changes requested):

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcLineImpl.java
NathanDissoubray: I notice you removed this doc, but what did it mean ?

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcLine.java
NathanDissoubray: I created a PR targeting your repo (since this is a fork, I can't directly pull request on this pull request, I must pull request your fork) https://github.com/thomas-durand-SGI/powsybl-core/pull/1
Please check in the fork PR if you are ok with it. On our side, we will also try to have someone look at it before merging into your branch, so you don't have to change it after.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcNode.java
thomas-durand-SGI: These limits apply to the signed value returned by `getV(`), not to how big that number is once you ignore its sign. The rule is always `LowVoltageLimit <= V <= HighVoltageLimit`. This matters because the same check already written for `VoltageLevel` can be reused as is: a "normal-functioning" DC Node on the negative side reads something like `-500`. If the limits were stored ignoring the sign, as `[480, 520]`, that check would see` -500 <= 480` and wrongly say the voltage is too low. Stored signed instead, as` [-520, -480]`, the same check sees `-500` sits between them and reports nothing, which is correct.

A DC Node on the positive side has a band such as `[480, 520]`. One on the negative side has a band such as `[-520, -480]`. On the negative side, going below `-520` (the low limit) means the voltage is too big once you ignore the sign, and going above `-480` (the high limit) means it is too small: the opposite of the positive side. So "low" always means the smaller number and "high" always means the bigger number, but which one is "too big" and which one is "too little" flips depending on which side of zero the band is on.

A band can even sit around zero, such as ` [-10, 10]`. There, `V = -15` crosses the low limit and `V = 15` crosses the high limit, but both are the same problem: the voltage moved too far from zero, in one direction or the other. Neither crossing means "too little", because every value close to zero is inside the band and allowed. So code that wants to report "voltage too high" or "voltage too low" cannot decide it from which field, low or high, was crossed. It first has to check where zero falls relative to the band.

Essentially, there are too many different bands for one meaning of "low" and "high" to fit them all: a positive-side node, a negative-side node, and a node centered on zero each need a different answer. So this class does not try to pick one. It only checks that `LowVoltageLimit <= HighVoltageLimit` (ValidationUtil#checkDcVoltageLimits), the one rule that holds in every case, and leaves it to whoever reads V and the band to decide what "too high" or "too low" means for that particular node.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcLineImpl.java
thomas-durand-SGI: It meant this: once you remove a `DcLine`, trying to read or change its limits should give an error, the same way trying to read or change its `r` already fails, and the same way `DcNodeImpl` already fails for its voltage limits. `checkAccess()`/`checkModification()` are what make that error happen.

The confusing part was comparing it to `BoundaryLineImpl`. Here's the actual difference, in code:

```java
// DcLineImpl
public Collection<OperationalLimitsGroup> getOperationalLimitsGroups() {
    checkAccess();   // this line throws if the DcLine was removed
    return operationalLimitsGroups.getOperationalLimitsGroups();
}

// BoundaryLineImpl
public Collection<OperationalLimitsGroup> getOperationalLimitsGroups() {
    return operationalLimitsGroups.getOperationalLimitsGroups();   // no check at all
}
```

So today, on a removed `BoundaryLine`, you can still read its limits and get a normal answer, no error. On a removed `DcLine`, you get an error instead. We made `DcLine` stricter than `BoundaryLine` already is. Sorry the original comment didn't just say that. The code is self explanatory, which is why i deleted this from the doc.

iidm/iidm-serde/src/main/java/com/powsybl/iidm/serde/AbstractAcDcConverterSerDe.java
thomas-durand-SGI: If that's ok, I would recommend keeping the check on the version: without it, a hand-edited 1.17 file carrying `minP` would get read and accepted, and then written back out on the next export as if it had always been valid. With the check, the attribute just doesn't exist below 1.18, which matches what the file format itself says.

I have implemented the change for the arrays. They were only there because Java doesn't let a lambda change a variable declared outside it, so the old code used a one-item array as a workaround, changing what's inside the array instead of the variable itself. Once the lambda is gone, so is the need for it:

```java
double minP = -Double.MAX_VALUE;
double maxP = Double.MAX_VALUE;
if (context.getVersion().compareTo(IidmVersion.V_1_18) >= 0) {
    minP = context.getReader().readDoubleAttribute(MIN_P, -Double.MAX_VALUE);
    maxP = context.getReader().readDoubleAttribute(MAX_P, Double.MAX_VALUE);
}
```

Same pattern `PropertiesSerDe` and `TerminalRefSerDe` already use. Fixed in b0df7bc

Does that match what you had in mind, or is there a case the version check still misses?

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/OperationalLimitsGroupImpl.java
thomas-durand-SGI: I replied in the other thread.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcLine.java
thomas-durand-SGI: I tried your split to check it actually blocks the call we were talking about:

```java
CurrentOperationalLimitsGroup g = dcLine.newOperationalLimitsGroup("summer");
g.newApparentPowerLimits();
```

```
error: cannot find symbol
  symbol:   method newApparentPowerLimits()
  location: variable g of type CurrentOperationalLimitsGroup
```
It does! I'll go review the fork PR properly there. Once DcLine moves onto it, unsupportedLimitTypes has nothing left to do and I'll delete it.

iidm/iidm-impl/src/main/java/com/powsybl/iidm/network/impl/DcLineImpl.java
NathanDissoubray: It was the last part that was a bit confusing, I think a little comment like `A removed DC equipment's limit cannot be accessed, like the {@code r} {@code DcNodeImpl} does for its voltage limits` is fine

docs/grid_model/network_subnetwork.md
thomas-durand-SGI: You're right. Btw this is already how this works for AC. CGMES attaches active power limits per terminal on a Branch (`OperationalLimitConversion:104-124`), because the two ends genuinely see different power, same reason you gave.

For current on a `DcLine `we kept one collection for the whole line, because I1 and I2 are the same current, so two numbers would just risk disagreeing with each other for no reason. Active power doesn't have that problem, so it was never blocked by that argument, we just didn't build a per-side accessor for it in this PR.

`DcLine's` current limit doesn't change: it keeps the one `LimitsGroupsHolder<CurrentOperationalLimitsGroup>` for the whole line, exactly as it is now. Active power would sit next to it as its own thing: a new narrow group type alongside `CurrentOperationalLimitsGroup`, say `ActivePowerOperationalLimitsGroup`, with just `newActivePowerLimits()`/`getActivePowerLimits()` on it. DcLine would hold two `LimitsGroupsHolder<ActivePowerOperationalLimitsGroup>` instances built from that, one per side, the same way `Branch` already holds two separate holders for its own limits. All of that builds directly on what you're adding, nothing on your interfaces would need to change.

If we agree that this is needed, I'd wait for your PR to land, then build it myself as part of mine, current per side already fits inside what I'm doing here, and active power per side would just sit next to it the same way. Happy to help on your PR too if that's useful.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/DcLineSerDeTest.java
thomas-durand-SGI: Fixed, it now loops `testForAllVersionsSince(V_1_18, ...)` instead of only the current version.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/DcLineSerDeTest.java
thomas-durand-SGI: Fixed. Same loop.

iidm/iidm-serde/src/test/java/com/powsybl/iidm/serde/DcLineSerDeTest.java
thomas-durand-SGI: Fixed, it now checks both selected groups' limits via `getAllSelectedCurrentLimits`() instead of only the last selected one.

docs/grid_model/network_subnetwork.md
NathanDissoubray: So we would need one list of group of currents for the whole line, then 2 more lists of groups for active power for the DC line ?

thomas-durand-SGI:
Hi @NathanDissoubray ,
I have finished the second pass of your comments. Thank you for the precious feedback.

- b2744f9: "Address review comments 2 on DC limits" contains the DcNode javadoc rewrite for the signed-limits question, and the two remaining DcLineSerDeTest coverage fixes.
- 8e6d331: "Restore short comment on DcLineImpl.checkAccess" puts back the one-line comment you asked for on the last thread.

I have gone through your thomas-durand-SGI#1 and from my side I confirm it now does what we discussed, apparent and active power limits on a DcLine become impossible to express at compile time, rather than compiling and being refused at runtime like they are now.

docs/grid_model/network_subnetwork.md
thomas-durand-SGI: You're reading it right, one collection for current, two more for active power because current is the same number at both ends while active power is not.

I checked whether that's the standard shape: CIM's `OperationalLimitSet`, the class that groups limits like `ActivePowerLimit` into seasonal sets, attaches to `ACDCTerminal`, explicitly AC **or** DC. So the standard already allows this, powsybl's importer just hasn't wired DC into it yet, same reason `DcLine` couldn't hold any limit before this PR either.

Same plan as before: current stays one collection for the whole line. I'll wait for your split to land, then add the per-side active power limit myself inside this PR in the same shape `Branch` already uses for its own two holders.

NathanDissoubray: but in that case we would need to split active and reactive power limits group too (since currently it's in the same group)

thomas-durand-SGI: You're right, I ran it to be sure: `OperationalLimitsGroup` still adds active and apparent power in one step, there's no type today that has current and active but not apparent.

Are you adding a type between `CurrentLimitsGroup `and `OperationalLimitsGroup `for active power alone, or should I skip `LimitsGroupsHolder `for active power and just give `DcLine `plain `minP1`/`maxP1`, `minP2`/`maxP2 `fields per side, like `AcDcConverter `has?

NathanDissoubray: If we want to separate completely, we would need to make the current, active power and reactive power group interface all independant. I'm not sure of the impact of that.
I don't think using minP / maxP is a good idea, because it doesn't take time into account with temporary limits. I don't have a solution right now.

iidm/iidm-api/src/main/java/com/powsybl/iidm/network/DcNode.java
thomas-durand-SGI: Hi @NathanDissoubray,
You were right on your first read. You asked whether the rule was on |V|, and I now think it should be. Two things I got wrong.

1. The names come out backwards. On a node at −500 with a signed band [-520, -480], a drift to −530 is an overvoltage, and -530 <= -520 reports it as `LOW_VOLTAGE `as you pointed out. A drift to −450 is an undervoltage and reports as `HIGH_VOLTAGE`. This is indeed contradictory.
2. Signed breaks existing code. `Security.java:345` is `Math.abs(violation.getValue()) / violation.getLimit() * 100`. Abs on the measured value, none on the limit. A node at `−530 kV` against a `−520` limit prints `−101.9 %` overload in the violation table. Against 520 it prints `101.9 %` (which is correct).

There is also a tell I ignored: signed forced me to add checkDcVoltageLimits to ValidationUtil purely to drop the >= 0 rule every other rating obeys. Magnitude deletes it and reuses checkVoltageLimits unchanged.

What we give up is detecting a polarity flip, which on a VSC cannot happen because the polarity is modeled by the current, and on an LCC is normal operation. The one rule the docs then have to add is: compare Math.abs(getV()), never getV(), which `Security.java` is already doing.

Still WIP, I will run this by colleagues and will confirm tomorrow. Not worth reviewing the current DcNode javadoc until then.

olperr1:
@thomas-durand-SGI: 
@NathanDissoubray has told me that this PR was closed. Sorry about that! It was not done on purpose, but automatically when I merged the PR based on "bump_iidm_18" which was used as a base for this PR (the branch was automatically deleted, then I restored it manually).

I reopen this PR.
Could you rebase your commits on the actual state of powsybl-core's main branch and change the base of this PR?

PR #4080, opened 2026-08-28
CGMES: Deterministic CGM subnetworks import order

jeandemanged:

No

Feature

When importing a CGM with subnetworks, the subnetworks are in a random order. The same import when repeated leads to different IIDM and reports because of different subnetworks loading order.

Subnetworks are always imported in deterministic order, depending on `iidm.import.cgmes.cgm-with-subnetworks-defined-by` parameter:
- `FILENAME`: IGM name used as sorting key
- `MODELING_AUTHORITY`: modeling authority used as sorting key

The PR also fixes a bug with the `FILENAME` method: an IGM name that's a substring of another (e.g. `Galia` vs. `HVDC-Nordheim-Galia`) caused name.contains(igmName) to pull the composite IGM's files into the simple IGM's subnetwork too, producing a merge failure (equipment already exists).

- [x] No

**Other information**:


cgmes/cgmes-conversion/src/test/java/com/powsybl/cgmes/conversion/test/conformity/CgmesConformity3ConversionTest.java
```diff
@@ -96,14 +93,17 @@ void microGridBaseCaseAssembledSeparatingByModelingAuthority() {
         params.put(CgmesImport.IMPORT_CGM_WITH_SUBNETWORKS_DEFINED_BY, CgmesImport.SubnetworkDefinedBy.MODELING_AUTHORITY.name());
         Network n = Network.read(CgmesConformity3Catalog.microGridBaseCaseAssembled().dataSource(), params);
```
NathanDissoubray, line 97: since both the IGM name and modeling authority lead to the same ordering, it's not clear if there is really an impact from the key on the sorting order, could you add a test where the ordering changes between the filename and the modeling authority to show this ?
Or is it not necessary ?

jeandemanged: I added the same tests for ReliCapGrid, where the order differs for the two methods. Note that this revealed a bug in the `FILENAME` method that has been fixed.

NathanDissoubray: Was the bug that HVDC-Espheim-Svedala and Espheim were considered equal since one contains the other ?

jeandemanged: yes, this caused the `HVDC-Espheim-Svedala` files to be included with the `Espheim` files. Then the final merge fails because of duplicate equipment.

NathanDissoubray (dismissed):

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
```diff
@@ -420,28 +438,31 @@ private Set<ReadOnlyDataSource> separateByIgmName() {
...
         private static boolean isBoundary(String name) {
             return CgmesSubset.EQUIPMENT_BOUNDARY.isValidName(name) || CgmesSubset.TOPOLOGY_BOUNDARY.isValidName(name);
         }
 
+        private static boolean matchesIgmName(String name, String igmName) {
+            // Match the sourcingActor segment exactly, not just as a substring,
+            // so that an IGM name that is a substring of another one (for example "Galia" in "HVDC-Nordheim-Galia")
+            // does not wrongly pull files of the other IGM into this one.
+            String[] parts = name.split("_");
+            return parts.length > SOURCING_ACTOR_INDEX && parts[SOURCING_ACTOR_INDEX].equals(igmName);
```
colineplqt, line 459: I am wondering if it would not be good to fail if the name is incorrect and the second field is not a sourcing actor with an exception. What do you think?

colineplqt (commented):
Thank you for your work on this!
I have one comment on the code.
Also, could you add one sentence on the documentation (grid_exchange_formats/cgmes/import.md) to specify that the import is deterministic?
But otherwise the PR looks good to me.

cgmes/cgmes-conversion/src/main/java/com/powsybl/cgmes/conversion/CgmesImport.java
jeandemanged: Why not, this validation should be done elsewhere though, checking upfront all files in presence and if they conform to expected pattern. I would lack availabilities to do that though. I propose to leave that as future work / future PR. Currently the PR just fixes a pre-existing issue and is iso-functional otherwise.

colineplqt: Yes you are right, it can be in another PR. Thanks !

colineplqt (approved):
Thank you for the documentation! It is very clear and helpful for CGM import!

NathanDissoubray (approved):

